PR #217, opened 2014-12-18
Enable ACU laser from water

aeoncleanse:
This came up in this thread: http://forums.faforever.com/forums//viewtopic.php?f=2&t=9032

This capability was removed ages ago because the laser would fire while the ACU couldn't be shot. The ACU has new above water targetbones now, so this shouldn't be a problem.

This is a balance change.


Sheeo:
Closing as it will be succeeded by a branch targeting 3641.

PR #263, opened 2014-12-29
Some fiddling with defaultunits.lua

aeoncleanse:
In addition to comment and whitespace changes, and a bit of shifting things around, I also moved the Overcharge into its own weapon class to cut clutter in the unit script files.

There may be an issue with line endings being messed up in this one, something which won't happen again as I'm now set up properly for it.


Sheeo:
This going into a new PR?

Issue #158, opened 2014-12-08
Investigate GC rapid-fire claw bug


aeoncleanse:

aeoncleanse:
This issue can go poof at some point.

MrRowey:
Dulicated isssuse

PR #275, opened 2015-01-02
Fix and improve GC Tractor claws

aeoncleanse:
This fixes issue #158 and cleans up the associated files for comments etc.

WARNING: The bug fix also makes the beam behave properly. Combined with some of the changes in commit 3f1e... this adds up to a SIGNIFICANT increase in the power of the GC against armies. I still think this should be merged as in my testing the claws now behave the way I think people have always expected them to, and the bug fixed is a major one.


ResinSmoker:
AeonCleanse, have you considered using the GC code I wrote for this?

http://forums.faforever.com/forums/viewtopic.php?f=45&t=8296

ResinSmoker:
In 85 aeonweapons.lua lines 79 to 94, it's used to prevent both claws from locking onto the same target!

aeoncleanse:
I've moved and enhanced it, not REmoved it ;) I looked at your code, but as my experiments had already worked I decided not to change too much at once. This GC works in the same way which won't cause uproar.

ResinSmoker:
Don't forget to adjust the firing arcs as all to often the CG will attack units outside of the yaw / pitch limits.

Resin

aeoncleanse:
WARNING: Error running lua script: c:\github\fa\lua\shield.lua(582): Game object has been destroyed
         stack traceback:
         	[C]: in function `GetArmy'
         	c:\github\fa\lua\shield.lua(582): in function <c:\github\fa\lua\shield.lua:581>

This comes up every time the claws kill a Harbinger (I think)

ResinSmoker:
I will give this a look. Is this with my modded version or the FAF version?

Kevin

ResinSmoker:
Ok after re-reading your comment... I'll suggest to "disable" the tractored units shields, weapons abilities, make it un-selectable and un-targetable. Now should the CG die while the Unit is still attached, simply re-enable everything and detach the Unit. This way you'll avoid some of the drama caused by having an active Unit held in limbo via the tractor event. If i recall correctly, the unit transports do something similar when loading / unloading units. Besides, none of this should be available while the Unit is held with the tractor anyway.

Resin

ResinSmoker:
After further research i believe that my initial assessment is correct. The shield.lua CreateImpactEffects line 582 is using "self" to get the shields owners army. If the unit is dead / destroyed just a tick before the shield is removed, you'll see something like this. Just remove / disable shields on all tractored units to avoid the possibility of this event taking place.

Resin

ResinSmoker:
Oh last comment then i'll shut-up... After some research i believe that its possible to have a unit "arc" towards the GC's claw rather then appear to be attached to rods. In a nutshell you'd use a projectile to attach the unit to, then give the GC's claw (bone) as the target for the projectile. This projectile being of the homing / tracking type will swing around and move towards the claw, accelerating as it does so. Now as the projectile gets closer to the claw, its turn rate will increase, thus ensuring that the missile will not overshoot the claw. Once the is in range of the claw it should of triggered an OnImpact / OnCollision event on either the CG or the projectile. At that point release / kill the unit / and update the GC's kill counter.

Resin

Sheeo:
Needs rebasing, closing for now.

Issue #276, opened 2015-01-03
System message broken


Xinnony:
With merge this https://github.com/FAForever/fa/pull/223

Now the 'Show system message' in 'Lobby Options' not work if i uncheck this.
Revert please.

aeoncleanse:
It's not that commit which has done it, because those xngine strings are not called or referenced anywhere. Not saying it's not broken (Have not checked) but I don't see how it can be this one.

aeoncleanse:
I checked now, system messages are working my end, auto teams settings are displaying properly.

Oh, and I'm gonna be frank this time: Instead of acting like an indolent, petulant arsewipe, try explaining why it is you think that the removal of these strings, which are not referenced at all, is going to have any effect? That's how adults behave when they have an issue.

Xinnony:
System Message not worked for Connect/Disconnect message.

Sheeo:
He means the system messages can't be _hidden_. I have yet to verify this myself.

aeoncleanse:
Oh yeah... Odd that. Still, it's not the commit he pointed to.

Xinnony:
I create the script for hide the Connect/Disconnect message, i know whereof I speak as well !

ChrisKitching:
After a quick glance at this, it looks like the problem might be something to do with the way this:
https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/lobby.lua#L4514

seems not to be checking SystemMessagesEnabled before printing. Thoughts?

Let's simplify how we check for enabledness, properly define what a "system message" really is (and which ones we want to suppress) and do it all from ShowSystemMessage. This probably warrants some discussion with the rest of the team: Do we really want to muffle events like "user timed out"? People plausibly want to be able to observe the difference between someone disconnecting and someone being kicked (though I realise that this presently is impossible either way)

#279 is probably in about the right direction. Haven't tested it, but, hey, it's 3 in the morning. Hopefully it'll at least push the discussion in a productive direction.

ChrisKitching:
@Xinnony: In response to your comment about trust: nobody should be a "trusted developer". Everyone makes mistakes, especially in software design: that's why we have code review. Even highly experienced developers still submit code to each other for review because they're aware that they make mistakes (really, really often, it turns out.).
I've probably got at least one amusing mistake in my latest PR. These errors are a big part of why they're "requests": it allows reviewers to spot the mistae before it starts ruining their life. This is just how things go.
[...]

PR #278, opened 2015-01-03
Push connectivity-related events through DisplaySystemMessage

ChrisKitching:
According to the "discussion" over in #276, these count as system messages, too. The likely cause of that bug is that those messages were just being printed with AddChatText instead of being subjected to something that checks SystemMessagesEnabled.
It seems that DisplaySystemMessage is the "Display a system message if it's enabled" function. Let's push the problematic messages through it so they'll be muffled if the tickbox is set.

There was also logic in DisplaySystemMessage that causes some messages to be displayed irrespective of the setting. Either the nomenclature, or this behaviour, needs to be changed. It's either a "system message" (in which case it should be shown only if those are enabled), or it's some other sort of message that we just display. It's very strange for a function called DisplaySystemMessage to in some cases show a message given that SystemMessagesEnabled() is false.

To sum up:

- The cause of #276 is that the problematic messages have never been printed in a way that checks SystemMessagesEnabled.
- We need to consider precisely what we mean by "system message" and decide what, if any, messages should be exempt from filtering. Do we even want this feature at all? It seems like you definitely want to hear about players being disconnected from your lobby...

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -798,9 +798,9 @@ function SetSlotInfo(slot, playerInfo)
             if not table.find(ConnectionEstablished, playerInfo.PlayerName) then
                 if playerInfo.Human and not isLocallyOwned then
                     if table.find(ConnectedWithProxy, playerInfo.OwnerID) then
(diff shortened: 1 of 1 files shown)


ChrisKitching:
... Aaaand this is why we don't write code at 0300.

PR #283, opened 2015-01-03
fix: cartographic view all black if map file doesn't contain cartographi...

anihilnine:
lots of maps incorrectly have all black settings for the cartographic view, resulting in the view being all black.
this occurs when someone saves a map using the current buggy map editor.
this fix detects if there is no colour data and makes some colours up at runtime instead


Sheeo:
This is very nice!

I'd like this in 3639 personally.

Sheeo:
I'll split these into seperate PR's for discussion.

Sheeo:
Split into:

 - [Bomber category for notha](https://github.com/FAForever/fa/pull/285)
 - [Cartographic map preview improvement](https://github.com/FAForever/fa/pull/286)

PR #286, opened 2015-01-03
Always generate cartographic view

Sheeo:
Before this, if a map file doesn't contain cartographic colours, the preview
will not be generated.

Since the shaders are not being passed the water height level, the cartographic
preview cannot include the water level of the map, however.

There's a discussion thread on the forums about patching the engine for this
[here](http://forums.faforever.com/forums/viewtopic.php?f=45&t=9141).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,9 @@ Changelog for patch 3639
 *Exploit fixes*
 - Fixed a regression of the free ACU upgrade exploit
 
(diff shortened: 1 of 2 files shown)


aeoncleanse:
Yup, I think a merge is in order. A view without water is better than no view at all.

Sheeo:
Indeed. This is great work @anihilnine!

PR #298, opened 2015-01-04
Remove gaps from button bar

### LINKED ISSUE #297: Gap around the "rerun CPU benchmark button"
Strangely, there's an ugly gap around this button. Layout bug. Blurgh.

ChrisKitching:
Fixes #297. Badly.

We'll presumably be torching this next cycle anyway, so I'm not too worried.


aeoncleanse:
:+1:

PR #300, opened 2015-01-04
Shrink lobby options menu

ChrisKitching:
It had lots of blank space. I killed some of it.

@aeoncleanse: You may want to further tweak the background image of the dialog. We can deal with giving it a background in a way that makes us want to cry less a little later on.


lua/ui/lobby/lobby.lua
```diff
@@ -5773,30 +5773,6 @@ function CreateOptionLobbyDialog()
             Prefs.SetToCurrentProfile('LobbySystemMessagesEnabled', 'false')
         end
     end
-    ----------------------
-    -- Developer box --
-    --[[
-    local devsHeader = UIUtil.CreateText(dialog2, LOC("<LOC lobui_0399>"), 17, 'Arial Gras', true)
```
Xinnony: Humm not.

Sheeo: Why not remove these comments?

Xinnony: I'm not agree about line 5782 to 5799.
And Moritz too i thinks.

If I add names here it is because there a reasons.

Sheeo: So you want this label back in, is what I gather.

I wouldn't mind a "contributor" section somewhere in the lobby UI, but it shouldn't be visible all the time: The list will clog the interface since we have so many people.

ChrisKitching: We have thing thing called `git blame`...

visionik: I think lines 5782-5799, giving credit to people who have worked on the lobby, is fine in concept.  With the lobby we have a precedent of Xinnony giving credit to others.

We have a broader issue though - why do lobby developers get credit displayed "in game" but not other developers?  

So the solutions are either:  we don't display in-game credit for anyone, or we need to come up with a way to display in-game credit for *everyone* that works on FAF:  the client, the server processes, the server operations, the graphics, the mods, etc.

I propose that instead of having in-game credits all over the moving forward, we create a page on the web site, http://www.faforever.com/credits, and list everyone there in various categories.  We can then link to it from wherever we want.

Xinnony would you be OK with this compromise?

Sheeo: While `git blame` serves the purpose of finding the instigator of flurry rainbows in code; it does not serve as a "Thank these people" functionality shown to users who do not use the VCS.

While we do post these regularly around the place, we actually have no central "Board of Apreciation", which is what I think @Xinnony is getting at.

ChrisKitching: Sprinkling your name all over the UI just gets in the way. How many of the programs you use day-to-day have this?
None.

You might find an author list on the `about` page, but that's usually about it. Putting a large blob of text in the settings UI is just unpleasant.


I find it slightly amusing that you didn't speak up when @aeoncleanse commented the code out, but you whinge when I delete the comments. :P

Xinnony: There have been problems in the past I let you talk to Moritz and Ze_Pilot.

visionik: Please see my proposal above.   Thank you!

Xinnony: But you can remove the description and hold just the nickname.

ChrisKitching: I think visionik's suggestion is an excellent idea. It should be all contributors, or none at all, and it should be in a less irritating place than right in the UI people are trying to use.

What problems have there been in the past, @Xinnony?

Xinnony: What you do not understand is that people contact me because they see my nickname to resolve bugs or improve a feature.

People are lazy and do not post their logs come so easily !

ChrisKitching: Instead of having reports directed to a single person (who then must be relied upon to push them into the issue tracker), it's probably more useful to direct people to the support forum (or possibly IRC where they can get immediate help).
The forum may well be able to resolve their problem without any involvement from us. If not, it'll get their issue on our radar so we can pick it up and suck it into our ticketing system.

We definitely don't want to miss an otherwise useful report because it was sent to the contributor who was on holiday. Or asleep. Or just forgot.
Equally, a report that one of us thinks isn't interesting or useful might turn out to be the missing piece to a puzzle someone else on the team is trying to solve.

aeoncleanse: A central credits page is the best solution, but I still think Xinnony should have a line of thanks in the Lobby Options because, unlike any one feature in game or in the main lobby, Xinnony came up with the current lobby practically single-handed. People helped, but the bulk of the work is his. Nobody else can claim anything close to that other than PilOt.

Sheeo: If we do that, Ze_PilOt should have a line of thanks somewhere too.

aeoncleanse:
Yes, he should, on the main lobby somewhere

aeoncleanse:
:+1:

PR #301, opened 2015-01-04
Add smart selection

anihilnine:
Adds select-by-category hotkey support for specifying categories to exclude.
Fixes the "select fighters" hotkey to not select t2 bombers anymore.

### DIFF
--- a/lua/keymap/smartSelection.lua
+++ b/lua/keymap/smartSelection.lua
@@ -1,0 +1,102 @@
+-- example usage: select all idle air units, excluding transports and bombers
+-- smartSelect("AIR MOBILE +idle -TRANSPORTATION -BOMBER")
+
+-- to bind this as a hotkey in your game.prefs make an action like this:
+-- UI_Lua import("/lua/keymap/smartSelection.lua").smartSelect("AIR MOBILE +idle -TRANSPORTATION -BOMBER")
+
+
+-- sets selection as per string expression
+function smartSelect(strExpression)
+  
+  -- some of these calls crash game if bad category name, so wrap in pcall
+  local success, value = pcall(function() 
+	local expression = compile(strExpression)
+	LOG(repr(expression))  
+	setSelection(expression)
+	return expression
+  end)
+
+  if success == false then
+     LOG("smartSelect failed: " .. strExpression)
+  end
+end
+
+
+-- sets selection as per compiled expression
+function setSelection(expression)
+  local others = join(expression.others, " ")
+  ConExecute("Ui_SelectByCategory " .. others)
+  local units = GetSelectedUnits()
+  
+  for k,v in expression.negatives do
+	if units ~= nil then
+		units = EntityCategoryFilterOut(categories[v], units)
+	end
+  end
+
+  SelectUnits(units)
+end
+
+
+-- convert a string expression into tokens
+function compile(strExpression)
+  local result = {}
+  result.others = {};
+  result.negatives = {}
+
+  local tokens = strSplit(strExpression, "%s")
+  for k,v in tokens do
+   
+    -- tokens with minus symbol are "negative"
+    if startsWith(v, "-") then
+       local withoutSymbol = string.sub(v,2)
+       table.insert(result.negatives, withoutSymbol)
+   
+    -- everything else is an "other". This includes categories (eg: AIR) and modifiers (eg: +idle)
+    else
+       table.insert(result.others, v)
+    end
+
+  end
+  return result
+end
+
+
+-- join strings together
+function join(items, delimiter)
+   local str = "";
+   for k,v in items do
+     str = str .. v .. delimiter
+   end
+   return str
+end
+
+
+-- split strings - from internet
+function strSplit(str, pat)
+   local t = {}  
+   local fpat = "(.-)" .. pat
+   local last_end = 1
+   local s, e, cap = str:find(fpat, 1)
+   while s do
+      if s ~= 1 or cap ~= "" then
+	 table.insert(t,cap)
+      end
+      last_end = e+1
+      s, e, cap = str:find(fpat, last_end)
+   end
+   if last_end <= str:len() then
+      cap = str:sub(last_end)
+      table.insert(t, cap)
+   end
+   return t
+end
+
+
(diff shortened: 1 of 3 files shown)


lua/keymap/smartSelection.lua
```diff
@@ -0,0 +1,102 @@
...
+   end
+   if last_end <= str:len() then
+      cap = str:sub(last_end)
+      table.insert(t, cap)
+   end
+   return t
+end
+
+
+-- does string start with value - from internet
```
Sheeo: These generic helper functions should either be already available (I'll look for these in the moho files), or we should make them available for other lua modules to use.

Sheeo:
As it turns out, GPG were really smart with their category system.

`ParseEntityCategory(str)` returns a `Core.EntityCategory`, which contains methods overrididing `__mul` and `__sub`, so this is possible with categories:

```
EntityCategoryFilterDown(ParseEntityCategory("AIR HIGHALTAIR ANTIAIR")-ParseEntityCategory("BOMBER")
```

But not this:

```
EntityCategoryFilterDown(ParseEntityCategory("(AIR HIGHALTAIR ANTIAIR)-BOMBER"))
```


I cannot, however, find any documentation about this and nowhere does it look like they use a different symbol for set difference in `EntityCategory` strings.

It does however, mean that we can use the built in set difference and multiplication methods for this implementation, and we should rework it to use these.

lua/keymap/smartSelection.lua
anihilnine: Agree, but I had limited time and wanted to get it in this patch.

anihilnine:
Yes fair enough. I don't have time right now for that though.

Sheeo:
Indeed it's not too important, I investigated mainly for my own curiosity.

lua/keymap/smartSelection.lua
ChrisKitching: There are some useful functions to be found in lua/system/utis.lua.

ResinSmoker:
I've worked with ParseEntityCategory quite a bit and can post ya a few examples on its usage if needed.

Sheeo:
@ResinSmoker if you have examples of doing category difference that would be awesome.

I've tried "-", "/" and "\" but to no avail.

ResinSmoker:
If I can get access to my laptop I'll fire you off a few example of cat parcing. Keep in mind that tonight is my wife's night and as such I am lucky to be able to reply via the iPad.  : )

Kevin

ResinSmoker:
I have a few moments so ill relate what I can minus the example in the hope I can set ya on the right path...

First off what your attempting isn't possible at all as that function isn't developed enough. Not sure why GPG didn't add this but it's what we have currently...

Ok here is what I normally do.

I make a table of the CAT's I'll either looking for or looking to omit. Next I make a blank variable and have it = something like ALLUNITS then ( + / - ) loop in the entires from your table. Next perform the parse via your entity categories contains function.

Sorry that I can't be more specific but this is something that's always been a pain in the arse from day one.

Kevin

Sheeo:
Indeed. My wish is to improve it so future modding can do this painlessly.

anihilnine:
Didnt I just kind of do that?

ResinSmoker:
Check out the following example script: http://pastebin.com/79yxC2eW

Look for the FilterTargets function

Resin

Sheeo:
@anihilnine sorry, I wasn't clear enough. Indeed you did!

If at all possible I'd like us to simply fix `ParseEntityCategory` however, removing the need for any other modification.

anihilnine:
I don't know how to override the default behavior. It's not like the original implemenation is in lua anywhere - it's baked in c++ somewhere. Perhaps we still can override it but are you really sure you want to? Not sure how performance friendly my one is ... at the moment it doesn't matter if it's just used every now in then to select some units but perhaps not so great if the sim starts using it.

ResinSmoker:
What would be the odds of GPG releasing the original C++ to us unbaked so we can correct it?

Sheeo:
@ResinSmoker Close to zero unless somebody contacts them and throws a moneybag.

@anihilnine let's get this into 3641, if you would contact me about the amendments I've talked about so we can get it done that'd be great.

ResinSmoker:
In a way by improving FAF, we are promoting the sale of Forged Alliance. Technically they're receiving / making money from the communities efforts. Hence I see no reason why this wouldn't be in their collective  best interests.

Sheeo:
I guess somebody just needs to contact them.

ResinSmoker:
I was wondering if you found my filter technique helpful...?

ResinSmoker:
Any ideas about how we can contact them?

Sheeo:
Your technique for filtering is what I was mentioning [here](https://github.com/FAForever/fa/pull/301#issuecomment-68624851), and yeah -- what I'd like for set subtraction to be implemented wholly inside `ParseEntityCategory`, relieving scripts from needing to do this using the EntityCategory class, in turn making the code for hotkeys quite readily readable.

I don't have any ideas, nope. I think @visionik is the guy with the ideas in this area :)

anihilnine:
Resin's way of doing this is nicer than mine. At the moment I use 'ConExecute("Ui_SelectByCategory " .. somestring)' to actually get all of the units. That's bad because I do it via the console and instead we could use 'ParseEntityCategory'. However, then we would have to do extra parsing for the five modifiers like +inview. The way I do it via console handles this for us.

anihilnine:
You would have to pull out those modifiers and feed them in here
UISelectionByCategory(expression, addToCurSel, inViewFrustum, nearestToMouse, mustBeIdle)

anihilnine:
woah. UISelectionByCategory actually takes a string like "AIR MOBILE", not a EntityCategory. This means if we want to use EntityCategory then there is no out-of-the-box way to support modifiers like +inview, which is very important. It's too hard - we will end up writing code comparing mouse position to all units to work out which one is the nearest etc. Therefore I feel we shouldn't use EntityCategory at all.

Sheeo:
Well that's a bit depressing, since `UISelectionByCategory` likely just uses an `EntityCategory` under the hood :/

PR #306, opened 2015-01-04
Switch in-game skin to the real factional skin on game start

ChrisKitching:
This stops "random" being a skin after the game has started, causing you to get the proper factional UI for the faction you actually ended up being.

This also obviates the need to import a ton more resources to fix the whole default/uef skin problem.


Sheeo:
:+1:

PR #310, opened 2015-01-04
Remove logspamful system messages

ChrisKitching:
Since the slots now correctly display "Connecting to X...", we can lose some of the logspam.

This largely mitigates problems related to the "hide system messages" checkbox (which has for now been removed).


Sheeo:
So, regarding the mysterious thing that keeps putting system messages into the chatbox, is this not it?: https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/lobby.lua#L4349

ChrisKitching:
Most likely. Well spotted.

PR #327, opened 2015-01-05
Fix transport-related hard-crash

aeoncleanse:
This is a fix based entirely on guesswork from the following error:

warning: Error running lua script: Unit:GetCargo only valid for transport units
        stack traceback:
               [C]: in function `GetCargo'
               ...rogramdata\faforever\gamedata\lua.nx2\lua\shield.lua(522): in function `RevokeTransportProtection'
               ...rogramdata\faforever\gamedata\lua.nx2\lua\shield.lua(437): in function <...rogramdata\faforever\gamedata\lua.nx2\lua\shield.lua:411>

I honestly can't think of or find a single reason for a non-transport unit to make it to this call, so my only idea so far is that the unit had died when the call was made...

Any input appreciated!

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -506,7 +506,7 @@ Shield = Class(moho.shield_methods,Entity) {
     },
 
     ProtectTransportedUnits = function(self)
-        if EntityCategoryContains(categories.TRANSPORTATION, self.Owner) then
+        if EntityCategoryContains(categories.TRANSPORTATION, self.Owner) and self.Owner:IsDead() != true then
             self.Owner:SetCanTakeDamage(false)        
             local Cargo = self.Owner:GetCargo()
             for _, v in Cargo do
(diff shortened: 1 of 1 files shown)


Sheeo:
What transport related hard crash is this?

aeoncleanse:
I did have game.log, no longer. A guy was talking about it in aeolus and I noticed because it's my own code. This was the only error in game log, and nothing of note happened in the replay. The player had just crashed a czar into a T3 base with shields, killed the player, and the error popped up a few moments after the 'player dead' message popped up, crashing the game.

I don't understand how this could have happened.

Sheeo:
Looking at this, we need to do some review of this sim code.

Unless we get proof that this is causing an actual crash, I'm reluctant to merge

aeoncleanse:
Agreed, I'm not confident in this one either

Sheeo:
Closing this because it's inconclusive.

PR #352, opened 2015-01-08
Fixing Rate of Fire buffs of weapons

### LINKED ISSUE #350: Salvation Adjacency bugged?
http://forums.faforever.com/forums/viewtopic.php?f=3&t=9162&p=90558#p90558

Crotalus:
http://forums.faforever.com/forums/viewtopic.php?f=3&t=9162

This partly fixes the bug with artillery weapons. Currently rate of fire buffs isn't respected all across the weapons code, making the progress bar bug for instance.

There's also confusement around the energy drain buffs, it's currently lowering the total energy needed but the drain / second is unaffected. This will make the UI report same drain but it will be a shorter timeframe (which isn't obvious).

Not sure what the best way is to solve this, maybe reduce the energy drain / s with the same % amount. Another solution would be to tie the energy drain to the rate of fire, the faster it fires the more it drains / second, thus making that variable in blueprint unnecessary.

Fixes #350


Crotalus:
I see I've stuffed some formatting changes into the actual fix commit, I'll try split them apart.

Sheeo:
Indeed that's not very obvious. On the same note, these buffs are undocumented excpept for our wiki.

Would be nice if we could somehow describe these in the unit blueprint database.

Crotalus:
9c36a5796cdb333c53b36c2d05dbf018b776278e

PR #354, opened 2015-01-09
Feature/ally overflow

Crotalus:
Show team overflow in economy / score

(Quality is now only visible if actually set in the game)


Xinnony:
:+1:

aeoncleanse:
This feature needs some improvement before merge. The big one is that it displays an odd string sometimes.

Sheeo:
Is this not what Crotalus is adressing with "(Quality is now visible if actually set in the game)"?

What I'd like to see is for all of it to be optional.

Xinnony:
Is already a option, no ?

Sheeo:
In that case: :+1:

Crotalus:
Yeah the ally overflow stuff is an option. Quality indicator should be own UI option I think, they don't have anything in common really except being in the same pull request

Sheeo:
Needs rebasing, let's consider getting this optionally into 3641?

PR #355, opened 2015-01-09
Cleanup of lua/ files

Crotalus:


ChrisKitching:
Looks great. Another thing of this sort that might be useful to add is conversion of all comments to use -- instead of #, as is standard lua syntax. Makes things look nicer in some editors (as that, plus your other changes, means things are actually standard Lua syntax at last).

Crotalus:
The only thing left now to standard lua syntax is getting rid of "continue"

Sheeo:
This looks good. I'm concerned about having to rebase all the other PR's after the fact, though :)

Crotalus:
Yes, I can rebase this instead and merge it at some time when there's less PRs

Sheeo:
Gotcha, let's do that.

PR #357, opened 2015-01-10
Pause ENGINEERSTATION units on tranfer

### LINKED ISSUE #353: disable hive assist after it was given to another player
Issue came from here:
http://forums.faforever.com/forums/viewtopic.php?f=42&t=9135&start=20

In short - when hives are given over to another player they continue assisting whatever was assisted before the transfer, making them a perfect tool to screw an allies eco/spam/build.
These should behave like engies (and drones?) - when given over, they should be paused/disabled so that the assist is voluntary.

I'm a strong supporter of the idea - nothing as frustrating as the described situation - happened to me yesterday.

aeoncleanse:
fixes #353 (Untested - Need Devserver)

### DIFF
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -164,6 +164,9 @@ function TransferUnitsOwnership(units, ToArmyIndex)
                 unit:DisableShield()
             end
         end
+        if EntityCategoryContains(categories.ENGINEERSTATION, unit) then
+            unit:SetPaused(true)
+        end
         unit:OnAfterTransferingOwnership(owner)
     end
     return newUnits


Crotalus:
Maybe should have special check for transfers at deaths, you don't want stations to be paused in that case.

aeoncleanse:
Actually I'd prefer it if they did pause, else suddenly your eco might be drained to where you don't want it. It's the same issue. Best to give the player maximum control I think.

Sheeo:
I'm thinking we can get this into 3641 -- please contact me about testing when you have the time

Sheeo:
Right, merging with develop and we'll test on dev pre 3641 release, which is coming up (No due date as of yet)

PR #359, opened 2015-01-10
Making UI prefetch more aggressive

Crotalus:
Prefetch is currently only used for UI textures, these should be loaded as fast as possible to prevent stutter. Currently the prefetcher pauses for 3 seconds every time there is disk load (which is constantly due to replay / log saving).

This patch makes the asset load much faster into memory.

The prefetcher never release memory anyway so this won't change total memory usage, it just makes sure the textures end up faster into memory.

### DIFF
--- a/lua/ui/game/gamemain.lua
+++ b/lua/ui/game/gamemain.lua
@@ -222,7 +222,9 @@ function CreateUI(isReplay)
     end
 
     Prefetcher:Update(prefetchTable)
-
+    -- UI assets should be loaded fast into memory to prevent stutter
+    ConExecute('res_AfterPrefetchDelay 100')
+    ConExecute('res_PrefetcherActivityDelay 1')
 
     ##below added for FAF
     import("/modules/displayrings.lua").Init()	##added for acu and engineer build radius ui mod


ChrisKitching:
This seems like it might lead to lots of stuttering very early on while everything is loading?

Crotalus:
Yes maybe, best thing would be if it could be preloaded in lobby. Not sure if thats possible. Currently prefetcher is initialized in UserInit.lua

Sheeo:
What are the default values for `res_AfterPrefetchDelay` and `res_PrefetcherActivityDelay` (I assume 3s for the latter)?

Crotalus:
250ms / 3s

It doesn't accept values < 1 for the res_PrefetcherActivityDelay which is too bad. 1 second pause is forever when it comes to disk load.

It could probably burst read it all into memory within a very short timespan.

Seems like __blueprints isn't loaded at the time of the lobby so prefetching it there isn't possible.

Sheeo:
How well has this been tested? What are the specific characteristics of the increase?

Crotalus:
SSD users shouldn't see any difference, non-SSD will have the UI textures read faster at start of game, hopefully reducing later needs for disk reads.

I've relaxed the options now, 1 second delay at disk activity and 100 ms delay.

The other option we have is to try remove Prefetching all together, this will free up some RAM and let the OS handle everything instead. Could help if users are running the game on low memory and have swapping. I can test how much impact it has on overall memory footprint.

Then we should warm up the disk cache ourself in the pre-lobby so there's a chance of it being in the filesystem cache.

Issue #284, opened 2015-01-03
Torpedo defense blocks allied torpedoes


Sheeo:
[Forum link](http://forums.faforever.com/forums/viewtopic.php?f=3&t=9095)

Is this intentional behavior?

aeoncleanse:
This can be closed, the PR fixes it and should be merged into 3641

Sheeo:
Please reference it directly, using a link or the issue number.

PR #362, opened 2015-01-10
Stop depth charges and flares drawing in allied entities

aeoncleanse:
Addresses #284


Sheeo:
Good job figuring this out. I do think this needs discussion; so let's spin off a thread on the forums.

On another note, we should be hooking `defaultantiprojectile.lua` instead.

aeoncleanse:
Closed after rebase to retarget at release/3641

PR #366, opened 2015-01-12
Ping range change for colors in connectivity.lua + volatile tag for nukers, stealth tag for HARMS

Varaxis:
To reduce confusion and lag accusations based on in-game (F11 window) yellow colored pings and the assumption that yellow is bad, the range for yellow ping was changed from 200-400ms to 400-499, red to 500+, and green for under 400ms.

### DIFF
--- a/lua/ui/game/connectivity.lua
+++ b/lua/ui/game/connectivity.lua
@@ -3,7 +3,7 @@
 --* Author: Ted Snook
 --* Summary: Connectivity Dialog
 --*
---* Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright Â© 2005 Gas Powered Games, Inc.  All rights reserved.
 --*****************************************************************************
 
 local UIUtil = import('/lua/ui/uiutil.lua')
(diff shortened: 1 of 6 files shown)


aeoncleanse:
:+1:

Varaxis:
To make players aware that strategic missile launchers have death weapons (10k dmg, 6 aoe) and should be careful about adjacency to other units, it now says that they are volatile in their tooltip.

To make players aware that HARMS, Cyb t3 torp launchers, actually have personal stealth, personally not knowing this until I sandboxed why I couldn't see them with sonar and radar in game, they are now tagged to say so in their tooltip.

ChrisKitching:
Why the closure? This PR looks pretty okay to me. Someone will be along to merge it soon enough ;)

Sheeo:
Yeah @Varaxis what happened here? :)

Varaxis:
I've been convinced that gaming communities like this are toxic and decided to cut any sort of connection that might keep me attached. I was pondering how to make this place more welcoming to newbies, but I see that the biggest fault is something I can't address, which is the people themselves, with me losing hope after seeing the mods themselves troll and become hostile to others. I've spoken to high rated players who have gone casual, effectively quitting, and now see things from their perspective. The negativity is contagious, and the amount of negativity that erupted last night was enough to send me on my way along this path. I don't want to experience petty frustration like that voluntarily.

Sheeo:
I would like some more information on what happened there, @Varaxis. Would you please message me about it?

Regards.

Issue #199, opened 2014-12-16
Allow rebuilding of UEF satelite when it dies


Sheeo:
Rather than the current solution, where the satelite base dies.

quark036:
Close this?

ChrisKitching:
Issue still exists.

aeoncleanse:
1c67c36320faf01a61176027dbb4c1cf2d2b1313

PR #367, opened 2015-01-12
Satellite Rebuild

aeoncleanse:
addresses #199 

There are still some quirks to work on, and I need some help...

1: If I can, I'd like to reset the animation pose when it rebuilds, so it can use the launch animation again. I've had to lock it for now. Standard SetRate(-tv) does not work for some reason.
2: Factory shows up as an idle factory in top-right idle factories UI, but with no icon - FIXED
3: Defaults to the build tech level queue of the last builder, rather than defaulting right to the Experimentals tab when you click on it
4: Some odd UI bugs where the build UI will be out of alignment.

I've given it 300 buildrate (10 T3 Engies worth) and the new Sat costs the same as building a whole new novax.


Sheeo:
Also needs rebasing. Would be cool if this could be done for 3641.

PR #370, opened 2015-01-14
Fix Torpedo Bombers by introducing a new bomb-drop aim mechanic and changing projectile behaviour

aeoncleanse:
Balance impact, needs discussion. This changes balance from current with the objective of allowing the balance team to give each one deliberate balance nudges.


Sheeo:
Let's get this refined with a rebased, updated PR.

PR #339, opened 2015-01-06
Popup dialog, Ranked Label, fix UnitRestricts button, change default UnitCap option

Xinnony:
- Add Popup quick dialog (UIUtil)
- Add new Layout and the Layout file (LayoutHelpers)
- Add Ranked Label
- Fix and Optimize LobbyPreset UI, Lobby Options UI and Changelog UI (https://github.com/FAForever/fa/issues/272)
- Fix the "Unit Restricts" button (https://github.com/FAForever/fa/issues/323 https://github.com/FAForever/fa/issues/313)
- Separate Mod and Mod UI in Lobby Preset
- Change Unit Cap default (for Ranked settings)
- Changes "Score" and "Team Spawn" default options
- Set Ranked button set more ranked options
- Typo
- Convert all Tab to Space
- Change version to 2.6.1

Sheeo : Sorry is the last time i make large PR, next time i make a PR one by one.
Ps : I make always PNG file, but a devlopper can convert to DDS if he wants.

PR #373, opened 2015-01-15
Allow users to once again view the unit restrictions.

ChrisKitching:
Simple fix to get the old behaviour back.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -1768,7 +1768,9 @@ local function UpdateGame()
         local notReady = not playerOptions.Ready
 
         UIUtil.setEnabled(GUI.becomeObserver, notReady)
-        UIUtil.setEnabled(GUI.restrictedUnitsOrPresetsBtn, isHost and notReady)
+        -- This button is enabled for all non-host players to view the configuration, and for the
+        -- host to select presets (rather confusingly, one object represents both potential buttons)
+        UIUtil.setEnabled(GUI.restrictedUnitsOrPresetsBtn, not isHost or notReady)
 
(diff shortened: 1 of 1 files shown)


ChrisKitching:
Shame this missed the hotfix: it's a bit of a gamebreak :/

Xinnony:
Already added in : https://github.com/FAForever/fa/pull/339

ChrisKitching:
That's right, but 339 is a bit of a tangled mess at this point, so I had a go at extracting the more immediately-mergeable parts of it.
In this case, since the work in 339 doesn't produce the same behaviour as was present before the regression, this warranted a fresh commit, not just a cherry-pick.
We can discuss your proposed new behaviour over in 339, or elsewhere. This PR provides a direct "un-regress" option with no additional baggage or complication.

Apologies if my comments on your PR don't make it clear how I got to this point.

Xinnony:
My PR is widely clearer than yours.

Exemple here :
https://github.com/ChrisKitching/fa/commit/49f0d460c197cdd6e4e1eff150771300ea8b5b51#diff-1898ae6dfa782099d74a2bc0b76f078aL4772
line 4772 > 4783

It was much clearer before your change.

Sheeo:
@Xinnony pull requests like these are preferred, since I can fire up FA and test this change individually, then merge it straight away and feel confident we do not have regressions -- I too, was working on pulling out parts of #339 for individual review before my exams hit.

I think with the comment in place, this is a good addition. I do think we should attempt to avoid the somewhat obtuse expression to determine visibility. Perhaps we could have two different buttons?

Sheeo:
@ChrisKitching please follow the conventions specified [here](https://github.com/FAForever/fa#contributing)

PR #374, opened 2015-01-15
Cleanup the mechanism for setting default options

ChrisKitching:
This switches specifying the set of options deemed acceptable for a game to be "ranked" to a neater, more declarative solution.
Some miscellaneous cleanup of CreateUI was required to get the upvalue count down: Sorry for the slightly offtopic commits.
Some cherry-picking from the tangled mess that is PR #339 occurred.

Noteworthy changes:
- Game options are stored in the profile *by value* now, not by key. Yes, this will break existing saved data, no, I don't care - it simplifies things a lot and nobody will be bothered about resetting their options once.
- A set of values for each game option can now be specified as being acceptable for a game to be considered "ranked".

It's unclear where, if anywhere, the determination of rankness of a game is actually *used*. @Sheeo: perhaps you can clarify this one loose end?

### DIFF
--- a/lua/ui/dialogs/mapselect.lua
+++ b/lua/ui/dialogs/mapselect.lua
@@ -730,7 +730,7 @@ function SetupOptionsPanel(parent, singlePlayer, curOptions)
                 if data.data.default then realDefValue = data.data.default end
                 line.combo:AddItems(itemArray, defValue, realDefValue, true) -- For all (true for enable (default) label)
                 line.combo.OnClick = function(self, index, text)
-                    changedOptions[data.data.key] = {value = data.data.values[index].key, pref = data.data.pref, index = index}
+                    changedOptions[data.data.key] = {value = data.data.values[index].key, index = index}
                     if line.combo.EnableColor then
                         line.combo._text:SetColor('DBBADB')
                     end
--- a/lua/ui/dialogs/mapselect.lua
+++ b/lua/ui/dialogs/mapselect.lua
@@ -3,7 +3,7 @@
 --* Author: Chris Blackwell
 --* Summary: Dialog to facilitate map selection
 --*
---* Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright ï¿½ 2005 Gas Powered Games, Inc.  All rights reserved.
 --*****************************************************************************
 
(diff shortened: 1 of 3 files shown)


Xinnony:
Already added in : https://github.com/FAForever/fa/pull/339

Sheeo:
The determination of rankedness is done wholly serverside, so unless we do a callback of sorts to ask the server for potential rankedness, this feature can never be correct. I'm not really sure what to do with this, since it's a really controversial and touchy subject.

I'll need to bring that up specifically with the council, and we'll need to take it into account with future changes to rating.

I think we need to omit the 'ranked' label for 2.6.1, and keep it as a 'bug-fix' release only.

ChrisKitching:
> The determination of rankedness is done wholly serverside, so unless we do a callback of sorts to ask the server for potential rankedness, this feature can never be correct. I'm not really sure what to do with this, since it's a really controversial and touchy subject.

It sounds, then, like what we'll want is a table somewhere in a file to its self specifying the constraints, rather like the literal one from the PR:

```Lua
RANKED_OPTIONS = {
        Victory = {'demoralization'},
        CheatsEnabled = {'false'},
        CivilianAlliance = {'enemy'},
        GameSpeed = {'normal'},
        FogOfWar = {'explored'},
        UnitCap = {'1000'},
        PrebuiltUnits = {'Off'},
        NoRushOption = {'Off'},
        TeamSpawn = {'fixed'},
        TeamLock = {'locked'},
    }
```

A sane approach would seem to be to shove this into a file to itself which we `doscript` into the Lua code, and which the server code parses to build some sensible datastructure it uses to do its side of the checking.

I'll rephrase this PR as one which does cleanup and simplifies the machinery behind determining rankedness, without actually showing the textbox. We can then trivially add that when we finish deciding which constraints we actually want to enforce.

ChrisKitching:
Good to go: All that was required was the omission of one commit.

Sheeo:
One comment here: Options do not show up in the game options list unless they have been changed at least once.

Is this because of changes to the structure in `game.prefs`, and how will it affect launching the game if they have not been changed?

I think users would be concerned by the fact that the game suddenly forgot their settings, so we need to deal with this somehow.

Spinning off an issue for this #381

PR #379, opened 2015-01-17
Show resources in map previews

ChrisKitching:
This commit introduces a new class to represent map previews, with the capability to show resource points as well.
This allows us to remove a lot of duplicated logic for handling the large map, as well as adding resource previewing to the map previews in the map select screen and main lobby essentially for free.
To make the performance of the rendering of the resource icons acceptable, a mechanism for object-pooling sets of identical bitmaps is introduced (Avoiding the expensive operation of creating and destroying many Bitmaps whenever the map view tries to update, which would otherwise introduce considerable "jank" to the map selection screen).

This also makes the in-lobby map preview square.


ChrisKitching:
(Note the dependency on #374, which one of myself or GitHub failed to express correctly :P )

ChrisKitching:
Shiny screeshots:
https://www.dropbox.com/s/ugh0r9u1ig1w3p7/Screenshot%202015-01-17%2015.33.56.png?dl=0
https://www.dropbox.com/s/hxzjn89f38lq4oo/Screenshot%202015-01-17%2016.27.27.png?dl=0

Sheeo:
Looks nice! I would personally like an option to toggle the resources view, since I think it looks a bit cramped on certain maps.

Thoughts?

ChrisKitching:
> Looks nice! I would personally like an option to toggle the resources view, since I think it looks a bit cramped on certain maps.

We can of course add a checkbox to the lobby options screen to toggle this.
An alternative proposal might be to simply make the icons smaller, like this:
https://www.dropbox.com/s/akvqi7t3tbv2gvl/Screenshot%202015-01-18%2020.40.29.png?dl=0

> if you would rebase a7b9bbd forward on top of develop that would make my life easier, since it does not rebase cleanly.

Of course, though it rebased cleanly for me.

lua/ui/lobby/lobby.lua
```diff
@@ -1145,8 +1146,7 @@ local function AssignRandomStartSpots(gameInfo)
 
                 if autoTeams == 'lvsr' then
                     local midLine = GUI.mapView.Left() + (GUI.mapView.Width() / 2)
-                    if(not GUI.markers[i].marker) then return end
```
Sheeo: This refactor really brings out the code smell around here.

On this note, why was this conditional here before? Is this an error case we are now handling earlier?

lua/ui/lobby/lobby.lua
```diff
@@ -1711,6 +1710,12 @@ end
 
 local function UpdateGame()
     LOGX('>> UpdateGame', 'UpdateGame')
+    -- This allows us to assume the existence of UI elements throughout.
+    if not GUI.uiCreated then
+        WARN("UpdateGame() pointlessly called before UI creation!")
```
Sheeo: If possible I think we should investigate what's left in the `debug` module to see if we can extract information about the caller here.

Would be a nice generic thing to have.

lua/ui/lobby/lobby.lua
ChrisKitching: So far as I can tell, this was just unnecessarily defensive programming. An error case might have existed here when it was possible to call UpdateGame before the UI had been created, but now I've stopped that from being a thing we can just ignore it.

lua/ui/lobby/lobby.lua
ChrisKitching: We could, or we could just call MAKE_SNAFUCATED().

Sheeo: Per our conventions I suggest simply `MakeSnafucated()`.

ChrisKitching:
Amended to use debug.traceback (unfortunately I couldn't justify make_snafucated, given the existence of this :P).

Two new commits added to eliminate the silly calls it was warning about.

Thoughts about the "smaller icons" vs. "add a checkbox" thing?

Sheeo:
Good work :+1:

Please keep our commit message guide lines in mind.

Issue #322, opened 2015-01-05
Kick player and Close the Host bug


Xinnony:
Kick player not write a message on the kicked player.

Just see 'Connecting to the host expired'.

Xinnony:
Exemple if you host a game, and you kick me.
I not see "You have been kicked by the host" dialog.
And need wait timeout connection for see "Connecting to the host expired"

And if the Host close the game hosted, the player not get a message for know "The host close the game" and need wait the timeout too.

OCGDFS:
maybe its related to that: if im observer without getting a spot, and im waiting until the game is hosted i only see lost connection to player a,b,c... in chat window, but im still in the lobby

ChrisKitching:
The code for evicting players delegates the actual work, as it always has, to lobbyComm.EjectPeer, which is natively-implemented.
This limits our options somewhat. Probably the best approach is to implement a sort of "elective eviction" protocol. When an eviction occurs, the host sends a message to the target, and the target client understands that as an instruction to exit. We can then have the host call EjectPeer in the usual way shortly afterward to stop people from being able to modify their clients to make themselves impossible to kick.

A similar approach, invoked from the exit handler, can be employed to handle hosts abandoning lobbies. I've noticed in the past that I get the "Unable to connect to the host" message when what's really happened is "Host abandoned lobby".

PR #380, opened 2015-01-18
Cleanup some debugging code

ChrisKitching:
We have this thing called "moholog".


Sheeo:
I would prefer it if we don't induce logspam for everyone -- although we do indeed need this for debugging connections.

I would prefer we strip all of these, and introduce a wrapper/decorator for `lobbyComm` for debugging purposes. We'll need this for tracking down #322 and friends.

PR #392, opened 2015-01-21
Economy overlay cleanup

Crotalus:
* Cleaned up the economy overlay code
The code was messy and non-DRY. I've just replicated the old behaviour for now, should probably be refined further, seems like the current warnings etc are a bit weird. 

* Show reclaim / second in the reclaim overlay
Useful especially when watching replays to get a grip how much someone is reclaiming


Crotalus:
e70de3d18589acdee319a1453f10be536917e331

PR #394, opened 2015-01-21
Prevent default options from being improperly set on blank profiles

ChrisKitching:
Currently, if a host hosts a game using a profile that doesn't have sensible values for default options set on it, the options would be set to values equal to the index in the corresponding lobbyOptions value array where the default value lives.
This would prevent the game from starting, as the resulting values weren't valid settings. This would also prevent options from showing up in the box summarising game options.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -3816,33 +3816,23 @@ function RefreshOptionDisplayData(scenarioInfo)
             tooltip = { text = optData.label, body = optData.help }
         }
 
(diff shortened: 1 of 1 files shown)


Sheeo:
Looking good. Please rebase your commit and [follow conventions](https://github.com/FAForever/fa#contributing), though.

ChrisKitching:
Ta-Da!

Sheeo:
:+1:

ChrisKitching:
75d84f68625bce2d480b6e0759068ba7179e9291

Issue #201, opened 2014-12-16
Get rid of all GPG-copyrighted code, use hooks/rewrites instead


Sheeo:
Will keep this issue updated with status.

CookieNoob:
https://github.com/FAForever/fa/search?p=1&q=Copyright&type=&utf8=%E2%9C%93

currently the approach seems to be to upload the original file and then make modifications to the functions... while this is nice for developing stuff, the amount of copyrighted code is increasing a lot. Do we still want to get rid of this code or did this get dropped now? @aeoncleanse

Sheeo:
@CookieNoob our policy isn't changing on this. That being said; it's not a top priority.

MrRowey:
#3930

Garanas:
Safe to say: the policy has changed as per the board meeting of end of May.

PR #396, opened 2015-01-22
Weapon enhancement

aeoncleanse:
A bit of cleanup, moves overcharge from unit scripts into defaultunits for centralisation. No functional changes.


Sheeo:
4fb45e1f1eca397f7a182f04ef49bd8ec56259a5 introduces gpg-files, I'd prefer it if we stick to the purpose of #201.

aeoncleanse:
I recognise your concern, but A: It's almost inevitable these files are going to have to be added at some point, given how broad each one is, especially if RK's is ever integrated, and B: This cuts a lot of duplication out of the ACU script files, which are a ruddy mess so that's a really good thing.

Sheeo:
Issue is simply that we cannot do this legally.

Just import the file, strip the implementation blocks and destructively hook what you need to hook, rewriting required code as you go along. Following suit with my example from #201 this should be fairly straight forward.

aeoncleanse:
So you want me to hook it? OK!

Sheeo:
I would prefer it if we could simply chug it all in and provide a full base game distribution, but, we need to follow suit of projects like http://openttd.org and do it right :)

aeoncleanse:
Or we can get Visionik to buy ALL THE THINGS and do it 'Right' that way :D

aeoncleanse:
This is now ready

Crotalus:
:+1: 

d793f213c4b9e34303d9b5294a9154c78b9215c8
f5ddc3d13be8d76d0e124c87e6982acbbe9df7ff
3bffc029adbad2f4ec760b8d0127f160df4ecf55
8f2a1ee83511151197ce499b8784c1ee0049feb4
4564199035c0cc280890536a591ff8bdabfcfeac
b2b0a1789d5121fa6a7c213cf2c47ead1c0a4693
d5ab7c426fe8d72bbd48d87ee60c947f24f3f9a6

PR #397, opened 2015-01-22
Bring back vanishing checkboxes

ChrisKitching:
Some GPG code we're not currently modifying was failing to create checkboxes because it was relying on the old naming convention.
The old convention named checkboxes "radio", which seems irritating. Here, we neatly dodge this problem by entirely ignoring the parameter GPG's calls to CreateCheckboxStd gives us. They always want to create a checkbox with the default textures, so let's rename our fancy checkbox-creating method to something else and turn the one GPG's untouched code uses into a single-argument default-checkboxes-only one.

### DIFF
--- a/lua/ui/controls/radiobutton.lua
+++ b/lua/ui/controls/radiobutton.lua
@@ -43,7 +43,7 @@ RadioButton = Class(Group) {
             end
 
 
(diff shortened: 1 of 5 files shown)


Sheeo:
:+1:

PR #403, opened 2015-01-22
Eliminate "dead zone" between combobox elements.

ChrisKitching:
There used to be a region between elements in combobox that, when clicked, would close the combobox without selecting an option.
This is generally not what it desired, and it's been bugging me, so let's clean it up.


Sheeo:
:+1:

ChrisKitching:
6b719d4

PR #406, opened 2015-01-24
Feature/Target-priority

Crotalus:
With this change, when the user is manually issuing attack orders the blueprint-id of the
targetted units is added to each weapon's priority list (in addition to the default ones).

The effect of this is that units will have a "memory" of what to shoot,
for instance when you choose to attack ACU with colossus and then move, it will continue to attack the ACU before attacking anything else.

There is unlimited use for this, you can for instance order attack on t1
tanks with ACU, when you start to move the ACU it will prioritize those kind of units before shooting at something else. Attack T3 power generators with a T3 Arty and it will then continue to shoot at T3 Pgens until you give it another order.

It will take into account if you queue up attack targets, effectively making you
able to create priority queues for groups of units.

Personally I think this takes the game to a whole new level, the units are suddenly not completely stupid and follow the commander's order.


ChrisKitching:
Previously I had noticed if I gave an attack order on a particular unit, and then gave a move order, it'd just forget what I'd told it to shoot and resume its usual target-selecting heuristic while it's walking.
... You fixed that? That's awesome.

I'm not sure the thing about prioritising a whole class of units just because you once ordered it to shoot at one of those. You can well imagine there would be pathological cases for that. Wouldn't the best semantics be that, if you issue an attack order on a particular enemy, the unit will prioritise shooting at that until another attack order is given or until that target is destroyed.
Otherwise, your unit becomes differently stupid when it proceeds to prioritise shooting t1 tanks ten minutes later just because you once ordered it to exterminate a t1 tank that was causing a pathfinding problem.

aeoncleanse:
I'm with chris, surely it would be better to do this on an entityID basis rather than a unit blueprint id?

Crotalus:
Yeah I first tried to get SetTargetPriorities() to work with single entities but couldn't get it working, not sure if ParseEntityCategory() supports if, maybe needs another format than just the entity ID. I also tried to mess with setting targets manually with weapons but then you have to check each weapon seperately according to range / ability to shoot target.

On the other hand, imagine for instance when you fight navy and select your 5 battleships and want them to target enemy battleships (or shield boats). Then you just attack a battleship and after that it will shoot at those first.

At the moment you need to babysit every unit in game to do stuff like that. Another example is to target mexes with cruisers. 

So with this change I would guess the total babysitting of units would be vastly reduced. 

It would be possible to add a switch to the UI to use this behavior or not, then the players themselves can decide if they want it or not.

aeoncleanse:
It IS a good change, you're right, but I think two things need to happen. First, it needs to be on a toggle somehow, or perhaps a secondary type of attack command, perhaps a keyboard modifier. Second, we're going to have to be very, VERY precise when telling people how this works, and how and when to use it, or people will attempt to use it in the wrong situations and get annoyed with us.

Sheeo:
If we're going for this, I think we should be getting in manual override for targetting as well (i.e. target painting).

aeoncleanse:
Agreed. Crotalus, did you check out Domino's target painter? It may contain the solution to targeting individuals.

lua/sim/Weapon.lua
```diff
@@ -0,0 +1,529 @@
...
+                        end
+                    end
+                end
+            end
+        else
+            error('*ERROR: Trying to setup a turreted weapon but there are yaw bones, pitch bones or muzzle bones missing from the blueprint.', 2)
+        end
+
+
+        local numbersexist = true
```
Sheeo: I'm a numbersexist and I find this offensive.

Sheeo:
If I'm not misunderstanding completely, this thing relies on Entities having their blueprint id as a category attached? Is this done in code somewhere?

Also, if the cagegories for an entity are not tied unmutably to the blueprint ID of an entity, we could just add the entity id to the list, allowing us to filter by entity id using EntityCategory expressions.

Crotalus:
I've tried to find where the blueprint-id is added to a unit's categories list but can't find it, maybe it's done in engine. I've also not found a way to manually add categories runtime in lua, otherwise we could indeed add ID_273 etc as a category.

As to the target painter mod. If we for a moment ignore that it's using DMS_Lite (which in itself is a huge change to FAF):

It's using one thread per unit to constantly scan the surroundings in 128 distance for targets. And then loops through every enemy unit to sort out matching targets. In additation to this it fires off events every time a weapon enters ReadyFireState and fires etc. So it's trying to replicate the whole engine targeting system but in lua. I wouldn't dare to guess how many bugs / weird behavior this will cause. 

And it's also having 7-8 commands to paint stuff, I've never really understood where any of that is actually useful in the game. It seems more like a mod which is cool to play around with in UI

aeoncleanse:
OK forget that then :p We just need to be clear about what this does, that it sets a TYPE of unit as high priority, not an individual.

Drimocus:
Sounds cool, I'm an idiot with no experience in this whatsoever, but it would be nice to:
- have a UI thingy that indicates 'this unit is using additional target priorities' 
- have a reset button/keybind that resets (selected or maybe all) units to the default target priorities

I have no idea if this is possible, but it would be nice. Either way I think it's already better than the fixed priorities we have atm. (target priorities are only used when there's no manually set target queue right?)

Crotalus:
Yes, we should have some kind of priority queue which you can manipulate in UI (with blueprint icons) and also able to bind hotkeys. Maybe also able to save priority templates that you can use globally? So that the units get a different default priority.

I'm thinking maybe a hard "Stop" should reset priorities for now? The current "Stop" behavior is kinda hard, resetting loading nuke silos etc.

And yeah, if you target something manually this overrides the priorities.

Sheeo:
Also need to keep #201 in mind here.

Crotalus:
Now schooking sim/Weapon.lua instead

Sheeo:
:+1:

Sheeo:
Would be nice to get an updated PR targeting develop.

PR #407, opened 2015-01-24
Introduce general popupbox class

ChrisKitching:
This introduces a (relatively low-level) class for generating popup boxes.
You generate a UI Group containing the stuff you want in the popup, construct a Popup object with it, and use Hide/Show to make it appear/disappear.

The Popup class handles:
- Positioning the popup centrally on the screen, above other UI elements
- Darkening rest of the screen
- Causing the popup to close when the escape key is pressed, or when the "shadow" is clicked.
- Drawing a dynamically-sized nine-patched background behind the popup.

This eliminates the need to maintain images for every popup background, allowing us to dynamically generate popups of arbitrary size with two lines of code.

IceDreamer also got all excited and redid some graphics meaning the combination yields excessive coolness:
https://www.dropbox.com/s/5hrhqvxob15ox3s/Screenshot%202015-01-24%2016.33.58.png?dl=0
https://www.dropbox.com/s/5i3d96pldyns39a/Screenshot%202015-01-24%2016.33.14.png?dl=0
https://www.dropbox.com/s/rbk3s1gyhvmoxml/Screenshot%202015-01-24%2016.34.06.png?dl=0


Sheeo:
Looking good.

lua/ui/lobby/ModsManager.lua
```diff
@@ -51,46 +52,46 @@ function HostModStatus(availableMods)
...
 end
 
+local modsDialog
 function NEW_MODS_GUI(parent, IsHost, modstatus, availableMods)
-    GUI = parent
+    -- Evil hack because something, somewher in this tangled mess of evil is Destroy()ing the
+    -- dialog between reuses and I can't be arsed to figure out what it is.
+    -- The contents of this file needs torching and rewriting.
+    if modsDialog then
+        modsDialog:Destroy()
```
Sheeo: We should figure out the callsite. `debug.traceback` to the rescue?

lua/ui/lobby/lobby.lua
```diff
@@ -5012,106 +5006,81 @@ function Country_GetTooltipValue(CountryResult, slot)
...
-            GUI.RuleLabel:DeleteAllItems()
-            GUI.RuleLabel:AddItem(wrapped[1] or '')
-            GUI.RuleLabel:SetColors("B9BFB9")
-            GUI.RuleLabel:AddItem(wrapped[2] or '')
-            RuleTitle_SendMSG()
-        end
+        ruleDialog:Hide()
     end
+
+    -- Ok button
```
Sheeo: If you don't mind me saying, these comments seem a bit redundant.

ChrisKitching: They are: I was just preserving them (the original was declared on line 5091).

Stabby stabby, then?

Sheeo: Unless they serve as a useful visual seperator -- I haven't actually looked at the code in an editor recently, so I wouldn't know.

Any opinions on this?

lua/ui/lobby/ModsManager.lua
ChrisKitching: Ah yes, OnDestroy() is a thing, so this should be straightforward.

ModManager really is pretty unapproachable at the moment, and since what it's supposed to do is relatively simple I'm inclined to advocate just binning it and starting over.

Sheeo: I'll take your word for this.

lua/ui/lobby/lobby.lua
ChrisKitching: I do sort of like some of them, to be honest.

Sheeo: Let's keep them then :)

lua/ui/lobby/ModsManager.lua
ChrisKitching: It turns out that this is the right behaviour anyway (at least, until we revamp the setting dialog).
It's being destroyed because the lobby calls Destroy() on the map select dialog when you close it (and because the popup is a child thereof, it gets destroyed, too).

Considering the implications of not destroying the game options dialog when it's closed it perhaps a bit out of scope. We can address this in later work, methinks.

ChrisKitching:
I missed a spot in the presets code. Decided to add a simple input dialog class to simplify a few things further.

Xinnony:
Nice PNG files ...
yet foutage of mouth.

Sheeo:
:+1:

PR #412, opened 2015-01-26
Prune flag database

ChrisKitching:
Remove flag entries for which we have no corresponding image, and flag images for which we have no corresponding entry.


aeoncleanse:
I think what we should be doing is making sure all the countries recognised by the server have a flag available, which they don't. The flag graphics are in the file system, but they have no code reference. I was going to do this when I became incredibly bored one day...

ChrisKitching:
Okay, now it's consistent with the flag database we're using:
https://github.com/FAForever/server/blob/develop/geoip.py#L2

ChrisKitching:
IceDreamer, here is a list of country codes that don't have flag images:

01
a1
a2
ap
bv
cc
ck
cx
fx
gf
gi
gp
hm
io
kp
la
mf
mq
nc
nu
o1
pf
pm
re
sj
tf
tk
tl
vi
wf
yt

Here is a place that converts these codes into real country names (but you need to use the code as the filename):
http://dev.maxmind.com/geoip/legacy/codes/iso3166/

Place the images in textures/ui/common/countries when you're done.

Sheeo:
Good work.

Sheeo:
dbc1f5ecb26a717163ff71b0bffd7e7f3215f218

PR #413, opened 2015-01-26
Create flags for some of the missing country codes

aeoncleanse:
From #412 - This isn't all of what needs doing, there are still some codes with no flags. I'll find the rest soon.


ChrisKitching:
You're completely mad.

ChrisKitching:
97b45ce

PR #414, opened 2015-01-26
Migrate more things to use the Popup class

ChrisKitching:
I missed a few spots last time round, notably UIUtils.QuickDialog. No more does the "You were kicked" dialog clash horribly with the rest of the UI.
I also omitted to delete a number of the old background resources.

Lastly, the large map preview was migrated, just because it looks awesome. Good suggestion, IceDreamer:
https://www.dropbox.com/s/ivf5gzz81tw2e6r/Screenshot%202015-01-26%2004.50.52.png?dl=0


Sheeo:
Looks dandy :+1:

Before merge, be sure to test the ingame uses of QuickDialog; exit-confirmation is one.

ChrisKitching:
59f07c2

Sheeo:
Exit confirmation with save option now looks like this: exit-confirmation

Not sure where else there are QuickDialogs with three buttons, but we need this tested.

ChrisKitching:
Looks like QuickDialog uses totally different layout code for two vs. three buttons.

*sigh*

I'll get right on it.

ChrisKitching:
See #426 .

PR #417, opened 2015-01-26
Cybran buildbeam

Crotalus:
The cybran engineers have an almost invisble build-beam to the flying
drone. This change use the ACU build beam instead so you easily see what a
specific engineer is building / idling

Suggestion thread:
http://forums.faforever.com/forums/viewtopic.php?f=42&t=9300


Crotalus:
487ba6cd14fe418b49428321ff02bd504deeaeea

PR #419, opened 2015-01-26
New factional lobby backgrounds

aeoncleanse:
Please note, in order to compress to DDS properly, the 'wide' stretchers are now 12px wide rather than 10.
There may need to be some adjustment in the code to account for this, there's a just-about perceptible
vertical line where I think the transparency is being doubled because of this.


ChrisKitching:
dfc192d

PR #424, opened 2015-01-28
Hint the targeting AI where to hit cybran sonar

Crotalus:
This helps Seraphim Destroyer to hit the cybran sonar correctly.

@aeoncleanse: 
You have a hitbox patch for 3642 right? Maybe this solves some of the trouble without having to raise the actual units after completion. What do you think?

### DIFF
--- a/units/URB3102/URB3102_unit.bp
+++ b/units/URB3102/URB3102_unit.bp
@@ -1,4 +1,9 @@
 UnitBlueprint {
+    AI = {
+        TargetBones = {
+            'Deck04'
+        },
+    },
     Audio = {
         Destroyed = Sound {
             Bank = 'URLDestroy',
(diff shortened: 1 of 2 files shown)


aeoncleanse:
This won't work. The bone in question is below the water, this is the first thing I tried. Because it's technically underwater, weapons which have AboveWaterTargets = true will not register it as a legitimate target bone. This is the reason raising the structure was necessary, or, to do it more correctly, we need to open the model itself and add a brand-new bone.

All this is generic for the ones I did end up raising. Where possible, I used existing, above-water bones. I think the Cybran Sonar may have been one of the good cases where the bone fix works, I know the UEF certainly isn't.

There's no need for this PR.

Crotalus:
Yeah, cybran sonar had some surface bones, that's why it worked I guess. Closing this

PR #425, opened 2015-01-28
Intel refresh at give / upgrade / destroy

### LINKED ISSUE #98: T1/T2 Intel and Targeting bug
If you scout a T1 Mex, then leave, then it upgrades to T2, it can no longer be targeted by Cruisers (And possibly by anything). Is this more to do with the intel buggery we're investigating in the future?

Confirmed in the replay below, this is the gate to a cesspool of awful coding and very bad gameplay. This is a huge issue.

content.faforever.com/faf/vault/replay_vault/replay.php?id=2739652

He provided this replay too, showing more related intel issues.

content.faforever.com/faf/vault/replay_vault/replay.php?id=2739683

Crotalus:
There's a bug where intel blips for destroyed / upgraded / given structures
don't get upgraded for enemy brains. This is problematic because you
cannot target given / upgraded stuff.

This is a workaround for this. When a structure dies it checks the enemy
brains if someone has a blip for it, if so it gets a brief vision on the
unit to get the updated intel.

Fixes #98

### DIFF
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -112,6 +112,7 @@ function TransferUnitsOwnership(units, ToArmyIndex)
         end
 
         -- changing owner
+        unit:RefreshIntel()
         unit = ChangeUnitArmy(unit,ToArmyIndex)
         if not unit then
             continue
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -383,6 +383,33 @@ StructureUnit = Class(Unit) {
 
     end,
 
+    -- Refresh intel on a destroyed / upgraded unit by setting vision on the actual blip.
+    -- The expired blip will actually be destroyed right after when the intel system notices it's no longer there
+    RefreshIntel = function(self)
+        local army = self:GetArmy()
+        for i, brain in ArmyBrains do
+            if army ~= i and not IsAlly(i, army) then
+                local blip = self:GetBlip(i)
+
+                if blip then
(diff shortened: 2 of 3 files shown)


Crotalus:
Of course, the best thing would be to not even show that something has upgraded but unfortunately this isn't supported

Sheeo:
Please keep #201 in mind here.

Varaxis:
Interesting. This bug seems to only affect units made out of freshly made factories, like bombers straight out of a new factory, but doesn't affect the acu, or other units that have been on the field for a while, depending on when the building changed due to upgrade or ownership change. I've confirmed that it exists in not only on FAF, but also the steam version.

Sheeo:
Needs to use the same method for intel retrieval as used in the laser bug @ChrisKitching may want to take a look.

Crotalus:
Can't use same method unfortunately (initializing intel on the changed unit), partly due to the unit can / is destroyed and partly because I think the vision for the unit will be affected then when you DisableIntel('Vision') afterwards.

So we need to create a temporary entity which gains the vision at the affected position. I schooked VizMarker now at least so we don't have to import whole gpg file

Crotalus:
Got rid of the whole vizmarker thing by putting the intel on the soon to be expired blip instead, seems to be working ok

Crotalus:
@Sheeo @Zock @ppipp This should probably be included in balance / beta branch so we can test it

ad4e59e1a2f5764c903222a8da9e05a6c37d3ac1

PR #428, opened 2015-01-28
Fix UI corruption when watching replays

### LINKED ISSUE #416: UI corruption when watching replays
The UI sometimes goes all splurked when watching replayed.
Reported here:
http://forums.faforever.com/forums/viewtopic.php?f=3&t=9309

And earlier (but without the useful "only when watching replays" caveat) here:
http://forums.faforever.com/forums/viewtopic.php?f=3&t=9236

Could it be that these people are attempting to load the now-defunct "random" skin?

ChrisKitching:
The "random" skin shall not be loaded in-game. This is enforced at game start (when the player who selected "random" is switched to the factional configuration that their randomness resolves to), and during skin rotation, but not when starting a replay, where it uses the skin for your last-used faction, which might be "random".

... So this ought to fix #416. Probably.


Sheeo:
This is pattogo's replaysync service, not what's used for viewing live replays/replays.

I believe we need to look at `autolobby.lua`

PR #429, opened 2015-01-29
Delete unused UIUtil functions

ChrisKitching:
An exact rehash of 351eb17c64e0ecdd0f105d2b0c4f394f06e3054c.

These functions are unused in our code and GPG's code. The old commit was backed out due to concerns about breaking mods.

I have since downloaded the entire mod database and grepped it for the names of each of these functions. No results were found.

Since these functions are both unused and represent obsolete, suboptimal ways of doing what they do (such as making an ItemList into a textbox, instead of just using an actual textbox), I am confident it is now safe to remove them.


Sheeo:
:+1:

PR #431, opened 2015-01-29
Get the lobby closer to release appearance

aeoncleanse:
This replaces all the buttons, positions everything correctly, and begins work on the presets section, at least visually. It contains several merged PRs, and I am certain there's a better way to do this to avoid merge commits, but I'm not sure how yet. If anyone wants to take that responsibility off my hands, Sheeo, Chris, you have push access.

One issue: The exit button in the large map preview doesn't seem to be mouse-over detecting very well. Not sure why, it works fine in the presets screen.


ChrisKitching:
On the whole, this is highly excellent.

You seem to have broken the preset dialog's layout a little: the details list appears behind the profiles list.

I also think your "Disabled" buttons are a bit too faint: they're almost invisible on my screen.

It might be a good idea to vertically-align the observer button. This might look better if the button becomes taller.

You should change your `LayoutHelpers.DepthOverParent` call for the close button of the large map preview to use "2" instead of 1: that'll fix your mouse-over problems.

Don't worry too much about the merge commits: you built atop multiple outstanding PRs, so a merge commit is unavoidable for now:they'll get rebased away eventually.

aeoncleanse:
Commented code is for the help dialogue box, not sure how to make it happen so I paused half-way through to get this out for review.

Sheeo:
I'd prefer some screenshots when dealing with design pull requests.

The CPU benchmark test button is seemingly broken on hover (No texture error in log, though), and the 'ranked settings' button always has the green hover effect applied.

I don't see why we needed to move the CPU benchmark button into a seperate row -- I think it looks a bit awkward.

I think the inconsistent left/right margins on the bottuns are a bit off, and there should be more than there is now at minimum, since they feel cramped otherwise.

The "Lobby Options" button could do with a rename to simply "Options", perhaps lobby presets should simply be called "presets" too. This leads me in the direction that "Game Options" is a bit awkwardly positioned and named.

In the future, we should consider splitting the "game options" dialog, moving the map-selection portion into a dialog spawned when clicking a button near the map preview, and the game options section to a dialog spawned when clicking a button near the game options list, perhaps inside it?

aeoncleanse:
http://s356.photobucket.com/user/aeoncleanse/media/Untitled-2_zpsmi0upopx.png.html

Looks like a couple of Aeon texture exports went wrong, easy fix.

Left/right margins? Youwhat?

Renaming sure.

This is just a prettyfication. Eventually remember I want to re-design the entire lobby to a new layout.

Sheeo:
I have to admit I prefer the button groups, since the functions represented by the buttons are indeed related, and it takes up less space.

Two rows of buttons gives the look and feel of a BeatPad, which I don't think is fitting for the lobby.

The left/right margins for "Lobby Options", "Game Options" and "Lobby Presets" are a lot more narrow than "Back" and "Launch the Game", I think they feel cramped. Renaming is one way to mitigate the issue :)

aeoncleanse:
We're going to have to agree to disagree on the button bar then, because IMO the bar was an ugly, broken, pixellated bit of crap.

As for renaming, yeah that sounds like a good plan :)

aeoncleanse:
Now looks like... http://s356.photobucket.com/user/aeoncleanse/media/Untitled-1_zpsfjpilvbs.png.html

aeoncleanse:
And now like... http://s356.photobucket.com/user/aeoncleanse/media/Untitled-1_zpspgazenub.png.html?sort=3&o=0

PR #432, opened 2015-01-29
Set factional background during CreateUI

### LINKED ISSUE #430: Lobby Background doesn't stick
It's defaulting to the normal background every time I start the game, rather than sticking with my selected 'Factions'. This is observed in my new PR's branch, which also merges 3 prior PRs.

ChrisKitching:
Fixes #430


ChrisKitching:
c710d41

PR #433, opened 2015-01-29
Lobby speedup

Crotalus:
A feeble attempt at reducing some lobby lag 

* Only send game rules to the joining player
* Send benchmark results in one batch
* Support for sending multiple game options 
* Support for sending multiple player options
* Cache result of CPU benchmark

Some more info in the commit messages

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -1740,12 +1740,16 @@ local function UpdateGame()
         -- TODO: Since these stats are all constants, we should figure out the right place to do this
         -- job once.
         if not playerOptions.MEAN then
-            SetPlayerOption(localPlayerSlot, 'MEAN', argv.playerMean, true)
-            SetPlayerOption(localPlayerSlot, 'DEV', argv.playerDeviation, true)
-            SetPlayerOption(localPlayerSlot, 'COUNTRY', argv.PrefLanguage, true)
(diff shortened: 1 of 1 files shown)


ChrisKitching:
These are, in general, good ideas.

The reasons I didn't do these already are basically:
- Since these involve the wire protocol, they're mildly irritating to test.
- Because of that, there's scope for breakage if you get this wrong.
- Optimising network messages isn't that useful for reducing UI lag. Cutting down the calls to UpdateGame (and eventually killing it entirely) will make a difference (you're doing some of that here, too, which is encouraging).

In short: This looks generally great, but needs very careful and comprehensive testing, and don't get too excited about the perf wins you'll see.

Crotalus:
Good input! I've updated the PR now. 

Yeah, some of these changes helps with reducing # calls to UpdateGame() so hopefully some improvement.

Sheeo:
We'll need some thorough testing of this with the server, making sure it doesn't act up.

ChrisKitching:
@Crotalus We've decided to go nuts and land the various bits of "make the lobby comm protocol not stupid" work that've been lying around.
If you could rebase this on current develop (which I just set fire to), you're welcome to merge it right away.
Don't forget to throw away the very last commit: I merged that earlier on.

Crotalus:
Removed the rule stuff also, saw that you use GameOption for that now

PR #436, opened 2015-01-29
Conservative ModManager cleanup

ChrisKitching:
Some very conservative cleanup of the nightmare that is ModManager.lua.


ChrisKitching:
27352e0

PR #440, opened 2015-01-30
Some logging on death of commanders

Crotalus:
The OnDefeat / FAFWin stuff is very messy, adding some extra logging to
see if it could be the cause to desynced replays


Crotalus:
I'm closing this, not very useful + chris seems to be refactoring gameresult etc in develop

PR #448, opened 2015-01-31
Remove auto-rand-map function

ChrisKitching:
This feature was strange:
- The map would be randomised when the game is launched.
- It was still possible to select a particular map from the map selection screen in the usual way, but this selection would be overridden by the start-time randomisation.
- The map preview would show the map selected in the map select screen, not indicating that the real map is random.
- The randomly-chosen map would make no account of the number of players needed, causing excess players to end up as observers.
- Some of the interactions between this and auto-teams and "random"(auto-balanced) spawns mechanisms were undefined.
- It was a tangled unmaintainable mess.

We might want to reimplement this in a less insane way later. For now, let's cull it.


aeoncleanse:
We DO want to re-implement it, I use this setting in a lot of my games. It's nice to be able to go into a friendly game with your mates with no idea which map you are, which faction you are, or what teams you're on.

PR #449, opened 2015-01-31
Simplify random map logic

ChrisKitching:
The random map button in the lobby used to take about 3 seconds to do its job, as it would load the "game options" menu and click the button...

Here, we make that less insane. A random map is selected that has enough slots to hold the players currently in the game, in a sensible amount of time.

The random map button in the game options menu is simplified to simply select a random element from the current map list.

The logic for filtering maps is refactored to use a functional filter pattern, for noticeable performance win.

### DIFF
--- a/lua/EnhancedLobby.lua
+++ b/lua/EnhancedLobby.lua
@@ -1,8 +1,8 @@
 function GetAIList()
-	#Table of AI Names to return
+	--Table of AI Names to return
 	local aitypes = {}
 	
-	#Defualt GPG AIs
+	--Defualt GPG AIs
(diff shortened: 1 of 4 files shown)


aeoncleanse:
Ultimately, I propose we need three random map functions.

1: The one in the main lobby. This should randomly select any map with enough slots for the number of players, exactly as Chris has just made it do.

2: The one in the options screen. This should randomly select any map from the filtered list, exactly as Chris just made it do.

3: The Random Map setting, which doesn't pick a map until the game launches. It should use the same pick logic as 1, but just not activate until launch.

PR #457, opened 2015-02-01
Positioning adjustments in lobby, add logo + text

aeoncleanse:


ChrisKitching:
Screenshots are necessary to review layout work. Please save your reviewers some time and provide your own in future.

1) https://www.dropbox.com/s/98f618hlmqpwjxb/Screenshot%202015-02-01%2019.56.30.png?dl=0
2) https://www.dropbox.com/s/zz3r6mgejwkgg8k/Screenshot%202015-02-01%2019.57.09.png?dl=0
3) https://www.dropbox.com/s/lcb8yo9twideo59/Screenshot%202015-02-01%2019.57.31.png?dl=0
4) https://www.dropbox.com/s/1h4uci1zelbfdpc/Screenshot%202015-02-01%2019.57.42.png?dl=0
5) https://www.dropbox.com/s/dqtln0kn2q4eb9x/Screenshot%202015-02-01%2019.58.18.png?dl=0

aeoncleanse:
I even uploaded some ready, forgot to post the links. Thanks Chris for doing that!

ChrisKitching:
1) The scrollbar positioning changes seen here are excellent, as is the movement of the two checkboxes.

I'm concerned by the FAF logo:I think this should be much smaller, and don't think it's a particularly good idea to put the version numbers in such a prominent position in the UI (99% of the users don't care about the version number, so put them somewhere out of the way).
One approach might be to shrink the FAF logo to be the same height as the sum of the heights of the two textboxes, and move the whole shebang into one of the corners.

You should definitely split the logo-stuff into a different commit from the layout tweaks.

3) We probably want larger writing, space between the textarea and the edge comparable to the space between the button and the edge, and a larger box to compensate for all of the above.
.. An "AlertDialog" class that is shared by this and the changelog wouldn't go amiss, either...

4) It's nice that the buttons are now disabled (but that should be in a different commit to the popup work).
Isn't the rightmost scrollbar too close to the edge? Don't we want a consistent spacing between content and the edge of the dialog on both sides?
Should the font be changed to match the title font in the new settings screen? (or the font there changed to match *this* font?

5) It quickly becomes apparent that this dialog is much too small.
We have plenty of space: we could readily make the dialog tall enough to see the entire settings list, and wide enough that sort-of-descriptive names are practical.

aeoncleanse:
FAF logo position was proposed by Sheeo. Having it there, lined up on the left with the other elements is the only place it can go really. Moving the versions there was my idea because there's so much empty space, and it sort of makes sense. Can't have it in the top-left corner because that will look silly at higher resolutions. I have a commit ready which shrinks the logo, but I'm not sure where to move it to...

Can you explain 3 more thoroughly? And yeah, that class would be good considering how much I just nicked :)

Scroll-bar is 2px from the edge, same as the other scrollers. An argument could be made in this case for 0 or 4 px, to line up with the X button. I was going to go through and change fonts as the very last thing before we push this update, because it's easy to change them, and I think we need more input deciding which to use.

Just figured out how to make the font bigger for that popup box, so I'm doing that now.

ChrisKitching:
> FAF logo position was proposed by Sheeo.

@Sheeo: Opine, please.

> last thing before we push this update

Don't delay!

> Can you explain 3 more thoroughly?

Of course. Do you want information on the how, or the what?

aeoncleanse:
Delay is from a desire to have all the stuff like the options screen etc merged and elements positioned so we can see the effect font has on everything.

The what please. I'm just not sure what 3 was talking about. I'm making the font bigger on the help popup as we speak.

ChrisKitching:
> Can't have it in the top-left corner because that will look silly at higher resolutions.

How so?

Thinking about it, a good idea might be to put in the bottom right, where the version number used to be.
Yes, it'll be tiny and obscure for large resolutions... that's kinda the point. As it is it's somewhat jarring.

ChrisKitching:
Popup's looking better, but there's still a strangely large amount of space below the OK button compared to the space between the text and the edge.

Was it not obvious that the numbers in my earlier message are referring to the numbered screenshots in the previous message? That's why there's no two and they seem to imply you have context.

aeoncleanse:
If I matched it to, say, 5 px in from the metal and the edge, when larger resolutions stretch the screen it will be hovering in the middle of nothingness, aligned with nothing.

You say jarring... I think it looks pretty, professional, and like it belongs.

The button is centred. If you prefer, I can shrink the box such that the 12px gap around the text is matched below and above the button?

And no, I didn't cotton on to that ^^

aeoncleanse:
OK so here we have presets and help box...

http://s356.photobucket.com/user/aeoncleanse/media/Untitled-1_zpsr7uwp5z7.png.html?sort=3&o=0

http://s356.photobucket.com/user/aeoncleanse/media/Untitled-1_zps55y10n4l.png.html?sort=3&o=0

ChrisKitching:
Those both look excellent now.

I think the last remaining complaint is the logo. Let's see what Sheeo has to say on that, and then I'll look at the code.

Sheeo:
I do think with its current size, the logo becomes too prominent overall. Is the game title/type gone now, or?

In my opinion we should put the top-left or top-right, with smaller version numbers right nearby. I'm not too worried about big resolutions, running 2560x1440 myself and this happens often in normal applications -- tiny logoes in the corner. It doesn't look or feel silly by now.

aeoncleanse:
Sheeo is that with the big one from OP, or the little one as it is now?

Sheeo:
Why was this closed?

aeoncleanse:
Because the commits went in

ChrisKitching:
Slightly borked the merge, forgot to post the hash here. Sorry.

PR #459, opened 2015-02-02
Don't handle mouse events for disabled buttons

ChrisKitching:
This prevents disabled buttons from losing their "disabled" appearance under certain conditions.


Sheeo:
:+1:

PR #464, opened 2015-02-04
Restoration Field now affects allied units

Crotalus:
Fixed so restoration field now affects allied units as well. 

Also did some cleanup / refactoring


ppipp:
Tested it, seems to work fine.

Sheeo:
Appears to work, but during testing I found this issue: https://github.com/FAForever/fa/issues/804

Merged with https://github.com/FAForever/fa/commit/b39bb64765ab474c078910846deda8cedace52fb

PR #465, opened 2015-02-04
Call UpdateGame less

ChrisKitching:
A variety of very trivial steps toward using UpdateGame less.

Future work should exploit the WatchedValue mechanism to further reduce the amount of work being done here (but this is a reasonably good first step that'll make the next stage simpler)


Sheeo:
:+1:

PR #469, opened 2015-02-04
Exterminate skinning.lua

ChrisKitching:
The only remaining use for this utter insanity was to try and draw a nine-patch background around the hotstats screen.

This use didn't work anyway, and fixing it isn't straightforward because of how amusingly mangled the layouts are in hotstats. So, this PR doesn't produce any visible changes, it just deletes a small nightmare from the codebase.
Ultimately, we might want to use the NinePatch class to build the feature ZeP was trying to get here, but with four lines not 822.


Sheeo:
How does this look now?

ChrisKitching:
Unchanged.

Sheeo:
:+1:

ChrisKitching:
To clarify: I haven't replaced a stupidly-implemented thing with a more sensibly-implemented equivalent, I've thrown away a stupid broken thing. :P

PR #473, opened 2015-02-05
Migrate unit manager to new UI look-n-feel

ChrisKitching:
This marks the passing of GPG's original theming from the lobby.
\o/

- A small amount of insanity is removed from ModManager.lua.
- ModManager now uses a different highlight graphic.
- The unit manager now looks awesome, with faction-specific skinning:
https://www.dropbox.com/s/ujs95vr294w06qq/Screenshot%202015-02-05%2001.07.07.png?dl=0

- This restrictedUnitsDlg.lua is about .5 the size of GPG's one.

Good work, @IceDreamer, getting these assets produced!


Sheeo:
What is this skin? The random one?

It looks very nice. I don't know about the radiobuttons though, I think they could use a bit more polish.

It's certainly better than what was there before already.

aeoncleanse:
I've got some thoughts to fiddle around with the buttons a bit, they need to fade out for the lasers for a start, and I'll nudge things around a little at some point. Chris and I also need to talk about the two 9-patch classes, popup and the border surround one, I think the system can be simplified and a whole bunch of visual artefacts killed in one go.

Sheeo:
I don't mean there are visual artefacts, I mean I think they look dull (The radiobuttons).

aeoncleanse:
Yup I'll address that too. The artefacts are actually present in all popups and pixel-level edits have proven unable to answer the problem.

aeoncleanse:
Chris, new assets are up on the repo replacing all the panels, if you could just switch them all in it should look great.

ChrisKitching:
Revised image:
https://www.dropbox.com/s/4d2nm2krbk2abdq/Screenshot%202015-02-07%2007.40.21.png?dl=0

Sheeo:
Great work on the borders!

If we could put a gradient of sorts on the circles I think that would suit it well -- not if it's too much work though. I think we're already well ahead of the previous release in terms of UI polish.

aeoncleanse:
Yup, wanted to see how these looked. I think I've got space here to nudge the left column to the right now, out of the way of the lasers.

PR #475, opened 2015-02-07
Remove owner-check on text pings

Crotalus:
All can now remove text pings (otherwise they would just move it offscreen
anyway)


Crotalus:
d3c8d9bd61317f1e668ca1ce2b2a6dd147dd6a85

PR #476, opened 2015-02-07
New option: Show reclaim value of wreckage / props

Crotalus:
If this is enabled, mass values of props / wreckage on the map are shown if Ctrl+Shift is
pressed simultaneously.

The value will be the initial value of the reclaim and partial reclaim
won't change it.

Especially useful when playing new maps without knowing where the reclaim
is. Reclaimable rock/tree with mass value below 20 isn't shown


ChrisKitching:
Does this cause any extra work to take place while the buttons are *not* held?

We don't want to cause performance regressions with our new toys!

Crotalus:
When buttons not held it will hide buttons if shown, otherwise nothing. It will have to check for the buttons held down every beat but that's the same as scumanager.

Hmm, an optimization can be to shut off beat function completely if the option is disabled.

ChrisKitching:
Is it possible to have a function be called when a button is pressed or released? A button-press event listener? This sounds like the sort of thing that definitely should exist, I've just not gone hunting for it.

If so, you can add the beat function when the button is pressed, and remove it when the button is released, right? (since you care about two buttons, you'll really want to have one function which queries the state of both buttons and register it to handle events for both buttons)

In any case, the polling approach is probably not *too* expensive, but it'd be nice if we can shave off as much per-tick overhead as we can.

Crotalus:
Yeah, I will look into hooking into WorldView:HandleEvent(), maybe that's triggered for keypressed also.

Another option would be to make it a pure toggle with regular keybindning, then it wouldn't need to be listening / polling at all

Crotalus:
Seems like WorldView is not catching keypress events, only mouse stuff.

Crotalus:
Ok, found out a way. Hooking into OnCommandGraphShow() which triggers when a user press Shift.  When this happens a beat function is added which checks for 'Ctrl' keypress.

Sir-Prize:
Will it be possible to re-bind Ctrl+Shift? I suspect it won't be a problem for most people, possibly even me, but I do have a lot of hotkeys that use Ctrl+Shift+[something] so I will get the reclaim pulsed a lot when I am playing, particularly in combat. Probably not a big deal.

If the hotkey for this can be rebound - can we bind it to Alt and not overwrite the use of alt as attack move?

Awesome feature, looking forward to it, and reclaim heatmaps for previews etc would also be seriously great.

Crotalus:
Yeah, it would be possible to add support for key-mapping but with maybe with a toggling function in that case, otherwise we would have to make a 1 tick watch-thread that keeps track of what keys are pressed
(atm it's hooking into waypoints function to know when shift is down)

Not that I think a watch-thread would be that expensive anyway, scumanager uses one already so we could combine everything like that in a single thread

anihilnine:
Once this is done we will also be able to make reclaim painter as a ui mod

Sheeo:
It's slated for 3641 according to milestone, I do think it's a cool feature :+1:

Crotalus:
#595

PR #481, opened 2015-02-07
Make buttons less cluttered

aeoncleanse:
Mod Manager > Mods, Unit Manager > Units


Xinnony:
Better : Units > Restrictions
Because you can restrict :
- Units
- Upgrade
- Templates
- ...

Sheeo:
Can we have some screenshots?

Also; wasn't the unit manager previously named unit restrictions or the like?

I do like Xinnonys idea of simply calling it "Restrictions".

aeoncleanse:
I checked, Restrictions and Unit Manager take up the same space, so still cluttered, though I think Restrictions is a more accurate description of the function... What do we want to go with?

Xinnony:
"Restricts" ?

aeoncleanse:
Changed to "Restrictions and merged into incoming PR"

PR #484, opened 2015-02-09
Add support for sixteen slots

ChrisKitching:
Because why not, eh?

https://www.dropbox.com/s/gzg4iylen0tj6cb/Screenshot%202015-02-09%2001.00.09.png?dl=0

It looks surprisingly okay in the UI, I didn't have to shrink any of the stuff below the slots (the slots are just a bit closer together, and I moved the "auto kick" checkbox down a little bit.
We could instead easily have it adapt this "compact" layout only when a map supporting >12 players is selected. We also presumably require someone to *make* such maps. :P

I know we ultimately want to move to 16, but it appears we can do fourteen right now, hilariously easily. Getting sixteen in seems like it'll require some layout wizardry (but honestly seems plausible if we jiggle things about a bit. Does the "auto kick" feature actually need a checkbox? Surely if disabled it is literally an "I enjoy having desyncs" button (it toggles the feature which automatically kicks players who lack sim mods the host has activated).
Having killed that checkbox, we could then move the "enable observers in-game" checkbox to become the sixth button in that cluster of five, and then we've freed up enough space for one more slot, meaning we only have to scrounge 21 pixels more height by "jiggling" to get the last one in: this seems quite doable... but maybe not right now.

@Sheeo: You're fixing a ton of connection retardedness this cycle. Now everyone isn't being jammed through the proxy for no reason we might find 7v7 to be an entirely plausible thing...


Xinnony:
Kill "Auto kick" checkbox :+1: 
And move the "Observers in-game" to a "Lobby Option" is the better i thinks.

ChrisKitching:
Good plan. Let's do that.

https://www.dropbox.com/s/6e1a657om92b092/Screenshot%202015-02-12%2020.43.21.png?dl=0

Xinnony:
Move **Hide default Options** on **Options UI** ?

And for **Observers in-game** i suggest to move in **lobbyOptions.lua**, not in **Options UI** (my bad sentence).

ChrisKitching:
> Move Hide default Options on Options UI ?

Could do, though that doesn't help us get more space for slots so I'm not too worried about it.
It also seems like something people might want to access relatively often, so it's possibly a good idea to leave it where it is. Thoughts, @aeoncleanse / @Sheeo ?


> And for Observers in-game i suggest to move in lobbyOptions.lua, not in Options UI

Assuming you're referring to the lobby options panel when you say "options UI", that's already what I did:

https://www.dropbox.com/s/gv0xcjjrh4m1m1c/Screenshot%202015-02-12%2023.21.34.png?dl=0


Anyhoo, after talking to Sheeo we've decided to postpone this feature until next cycle: it requires more work on the Python client than we currently have time to do.

Xinnony:
:+1:

aeoncleanse:
What about keeping the moving around of things in this cycle, but sticking at the current actual slot limit. The space won't be noticed too much.

ChrisKitching:
> What about keeping the moving around of things in this cycle, but sticking at the current actual slot limit.

That happened.

aeoncleanse:
If you can get rid of the default options checkbox on the main screen, I can push the active options display to a full 10 panels in my final reshuffle, which would be extremely good for a number of things :)

PR #486, opened 2015-02-12
Call UpdateGame even less

ChrisKitching:
Moving closer to being able to start using incremental updates properly. Just a wee bit too much coupling remaining yet for that to be straightforward.


Sheeo:
Is this meant for 3641?

ChrisKitching:
That was the intent, but if you want to get all "strictly bugfixes" about this then several of these can be pruned (as they are instead performance improvements).
It would perhaps make sense to include high-value perf wins, such as not doing UpdateGame whenever any player option is changed.

Sheeo:
I reckon since we have a bunch of untested stuff in develop already, we might aswell get this in too.

PR #493, opened 2015-02-18
Attackdecals cleanup

Crotalus:
Cleaned up attack decals code to be more generic and also added support for damage radius for tactical missiles.

We should probably redo the damage radius texture: textures/ui/common/game/AreaTargetDecal/weapon_icon_small.dds

Currently the colors are off (white / yellow). Would be better if it was transparent with red target reticle or maybe a more gradient "splash damage" like texture.


Crotalus:
I've updated the PR now according to @ChrisKitching input. Haven't touched GPG's way of handling the different command modes yet though.

Crotalus:
I should add that currently it checks the attacker's weapons every OnUpdateCursor() / frame update as long as it is in attack command mode. This is the same behavior as original and not optimal.

A better way would be to grab the decal data when a command mode actually is set, probably OnCommandIssued() or similar, and then only update location of the decals in the OnUpdateCursor() function.

Crotalus:
Only updating the actual attack decals on command mode change now instead of in a frame function

aeoncleanse:
This sounds cool, but you're sure it doesn't mess up the scaling nuke reticle?

Crotalus:
Yes, scaling of nuke should still work, it supports multiple decals

PR #494, opened 2015-02-19
Add JSON parser to CreateLobby (Yes. Really.)

ChrisKitching:
This can form the basis of a (moderately hilarious) IPC channel from FAF to the ingame lobby, using JSON strings. This means we can finally send messages from python-land into the lobby, which lets us do various shiny things (initially, manage connection state more interestingly).

On a related note, does anyone know if this SyncReplayServer nonsesne is actually used? As far as I can tell the binary is never launched with the /syncreplay option that would be necessary to invoke it, so we might as well cull it.
If it is used, it might be nice to refactor it to use the new, extensible messenger protocol.


Sheeo:
:+1:

@fafafaf Do you have an opinion on a change of the syncreplay message to fit in line with this?

fafafaf:
AFAIK you cant pass a string to **SetHasSupcom**, it only works with an **int**.
I suggest to use **CreateLobby** instead.

@ChrisKitching its launched from other program, thats why you cant find anything.

ChrisKitching:
> AFAIK you cant pass a string to HasSupcom, it only works with an int.

I think you might be giving Lua's type system rather too much credit.
Certainly, this works when you invoke HasSupcom with a string. What remains to be seen is if the native entry point performs type checking. If it doesn't, we're all set.

> its launched from other program, thats why you cant find anything.

Since we're shipping code to support the syncreplay stuff, it'd be great if we could see the code. Is it on GitHub somewhere? It'd seem to make sense to add it to our cluster of repos so we have a complete picture (and so people stop breaking the support your code depends upon), and I could probably contribute usefully.

fafafaf:
> I think you might be giving Lua's type system rather too much credit.

It could not even touch the lua layer.  At first, its handled by the networking core in the .exe.
I actually did the test again:

```WARNING: GPGNET: Error processing gpg.net command HasSupcom: incorrect argument type, expected int```

with **CreateLobby** i was able to send up to **4061 chars** to FA. (it failed with longer strings)
Code:
```python
 .. self.clientSend(Pack(CreateLobby = [0,0,"x" * 4061,0,0])) ..
```
FA:
```lua
 .. LOG("received: ",string.len(desiredPlayerName)," chars") ..
```
LOG:
``` INFO: received: \0004061\000 chars ```
> Is it on GitHub somewhere?

My fa related stuff is in this repo: https://bitbucket.org/fafafaf/

> .. I could probably contribute usefully

The plan was that FAF will include the syncing support in the _official_ client, so my program wont be needed at all.

ChrisKitching:
> It could not even touch the lua layer. At first, its handled by the networking core in the .exe.

So it *does* typecheck in native code. Delightful.

> with CreateLobby i was able to send up to 4061 chars to FA. (it failed with longer strings)

We could in principle use the integer parameters to CreateLobby to support multi-packet JSON objects, but let's not. Someone might actually start using them.

Sheeo:
Unless `DisconnectFromPeer` uses peer uid as an integer, I suggest we go with this.

ChrisKitching:
Merged with release.

fafafaf:
FYI: it only works in lobby state

ChrisKitching:
> FYI: it only works in lobby state

That is the idea.

PR #497, opened 2015-02-23
Always wait 1 tick before calling Unit:Destroy()

Crotalus:
Seems like the engine wants the unit to live at least 1 tick to be able to
grab the pose in TryCopyPose(). Maybe it sometime garbage collects the
unit before it actually can copy it.

This is mitigated by waiting at least 1 tick before destroying the unit.


Crotalus:
9c9baafc6cb67e6b1e21810c7c102cd949a1dce9

PR #502, opened 2015-03-07
Make it possible to pause all mobile units

Crotalus:
Pausing a unit will make it hold at next order in command queue.
The current order will be carried out though. It will also accept new orders in the queue so a good way to prepare orders without the units executing them immediately.

This is useful for many things, pausing transports to make sure all leave at the same time, bombers to sync attacks etc.

Suggestion was made here:
http://forums.faforever.com/forums/viewtopic.php?f=42&t=9547

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1355,6 +1355,16 @@ MobileUnit = Class(Unit) {
         Unit.OnKilled(self, instigator, type, overkillRatio)
     end,
 
+    OnPaused = function(self)
+        self:SetBlockCommandQueue(true)
+        Unit.OnPaused(self)
+    end,
+
+    OnUnpaused = function(self)
+        self:SetBlockCommandQueue(false)
+        Unit.OnUnpaused(self)
+    end,
+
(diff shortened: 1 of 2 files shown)


lua/defaultunits.lua
```diff
@@ -389,18 +389,18 @@ StructureUnit = Class(Unit) {
         -- Does the unit have any adjacency buffs to use?
         local adjBuffs = self:GetBlueprint().Adjacency
         if not adjBuffs then return end
-
```
Sheeo: Can you split the whitespace fixes into a seperate commit?

Crotalus: Sure, done! :)

Sheeo: Thanks :)

Sheeo:
Looks good :) :+1:

ChrisKitching:
Merged as ac0734fc2e08403c86915f225f44aa8c6f2c0c9e
Cherry-picked, because there is something deeply, deeply wrong with this branch.

PR #503, opened 2015-03-07
Disable assist while factory is upgrading

Crotalus:
Assisting with an upgrading factory results in cancelling the current
upgrade which probably isn't what the user wants.


Crotalus:
Putting this into release/3641

Issue #501, opened 2015-03-01
Shield disruptor not working on personal shields


Sheeo:
http://forums.faforever.com/forums/viewtopic.php?f=3&t=9518&p=95067#p95045

aeoncleanse:
How did we overlook this one? 0_0

ResinSmoker:
Any lua errors from this?

Sheeo:
It fails silently, #504 fixes it.

We should look at the performance implications of doing this for all units at some point, however. It would be more ideal to use the `State` class to avoid having to do this for all units in all states.

Crotalus:
2706b373a6497c156d7088980752d34a45c3ada2

PR #504, opened 2015-03-08
Fix Shield Disruptor VS Personal Shields

aeoncleanse:


Sheeo:
@ResinSmoker do you have any input on this before we go for it?

ResinSmoker:
I'd suggest making sure that the instigator is valid as often this can be
lost or become corrupted.

Resin

Sheeo:
Fixes #501

Crotalus:
2706b373a6497c156d7088980752d34a45c3ada2
3a358a7bad2b20c978a78cf63f9ea0aad4229ec8

PR #505, opened 2015-03-08
Various bugfixes

Crotalus:
Fixed some errors in game.log


ChrisKitching:
In general, you seem to be adding defensive null checks to naively prevent lua runtime errors.

Often, it's better to solve this sort of problem by preventing nils from ever reaching the problematic place, otherwise you just end up with defensive nil checks *everywhere*.
This might not be possible in al cases, but, generally, defensive nil checks hide other bugs. The fact that a nil is being passed to a function that can't meaningfully process them is possibly a symptom of a larger problem.

... But not always :P

Sheeo:
Indeed I'm skeptical about the Cybran massfab unit--I don't remember seeing these rotator errors in previously.

DarkmatterVale:
@Sheeo,

If you change that back to 10, does that fix the problem?

Sheeo:
I don't know -- I have yet to be able to reproduce crashes arising from it consistently.

There was another destruction thread value changed with c72643cfeb1e997a8d819f4eeb6f650067320079 -- for seaunits.

It's worth investigating :)

Sheeo:
@Crotalus if this PR is directed at 3641, it should target the release/3641 branch.

PR #506, opened 2015-03-08
SafeQuit() refactor

Crotalus:
Don't call both ExitGame() and ExitApplication(). The former is called for
non-online stuff.

Also, the function was in "c-boundary" so WaitSeconds() always failed,
thus risking SessionEndGame() not to complete.

Maybe this helps with desynced / corrupted replays, needs to be tested
ofc.


Sheeo:
:+1:

Could this target release/3641 instead though?

Crotalus:
Hmm, is it possible to retarget or need new PR?

lua/ui/dialogs/eschandler.lua
```diff
@@ -12,19 +12,25 @@ local Utils = import('/lua/system/utils.lua')
...
+        return
+    end
+
+    if SessionIsActive() then
+        ForkThread(function ()
+            ConExecute('ren_oblivion true')
+            ConExecute('ren_ui false')
+            SessionEndGame()
+            WaitSeconds(1)
+            ExitApplication()
```
Sheeo: This may indeed be the way to go.

Sheeo:
Yes, we would need a new PR.

But let's just merge it into release/3641 and close the PR afterwards. I give my :+1:

PR #510, opened 2015-03-12
Refresh UI-mods in lobby

Crotalus:
Another take at making UI-mods refresh correctly in lobby so you don't
have to rejoin to make them active / inactive

~~Needed to force a session refresh by calling PrefetchSession() with an
incremental number due to internal caching~~


Crotalus:
Damn, this creates alot of ugly errors in log... any other suggestion how to make this  work? We need to make sure sessionInit.lua is called before launch to update __active_mods

Crotalus:
sessionInit.lua is setting __active_mods but in which scope is this file? I cannot access '__active_mods' from lobby UI-side.

Crotalus:
Hmm, maybe disable the prefetch all-together so SessionInit only is called at the actual game start, I will try that

Crotalus:
Yeah that worked

ChrisKitching:
I was never clear on the semantics of PrefetchSession, but it seems to get called a lot (such as each time a map is clicked in the map list).

In general, it's probably a good idea to defer as much work as possible until launch-time.

PR #511, opened 2015-03-13
Various bugfixes

Crotalus:
#505 pointing to release/3641 now

Got the same cybran slider error in the latest replay I watched btw


Crotalus:
Closing this until we find a better way to prevent the errors

PR #512, opened 2015-03-13
Refresh UI-mods in lobby

Crotalus:
#510 rebased to release/3641

Another take at making UI-mods refresh correctly in lobby so you don't
have to rejoin to make them active / inactive

PrefetchSession() seems to be a null operation if called on a map already
prefetched. This results in UI-mods being non-changeable unless you also
change the map at the same time.

By not calling PrefetchSession() at all we prevent this and
sessionInit.lua is called at game launch instead

### DIFF
--- a/lua/MODS.LUA
+++ b/lua/MODS.LUA
@@ -213,7 +213,6 @@ function SetSelectedMods(s)
         end
 
         SetPreference('active_mods', s)
-        UpdateUIMods()
     end
 end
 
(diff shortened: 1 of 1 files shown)


Sheeo:
Interesting. So with this we no longer call `PrefetchSession` at all?

Crotalus:
Yeah, then sessionInit.lua is called at launch instead of at prefetch call. I haven't noticed any performance difference but I'm running on SSD atm. Theoretically initial lobby startup could get faster but load time worse.

Crotalus:
Closing this, prefetch session is no longer called in lobby.lua

ChrisKitching:
@Crotalus ... For now.

PR #514, opened 2015-03-13
Update unitview.lua

Sheeo:
Change removes a divide-by-zero error which can be triggered by any non-standard shield type, allows for custom shield types to show properly (4DC Drone shield) while removing redundant / repeated script sequences


lua/ui/game/unitview.lua
```diff
@@ -448,25 +448,29 @@ function UpdateWindow(info)
...
+	if info.shieldRatio > 0 then
+	   local unitBp = info.userUnit:GetBlueprint()
+	   local ShieldMaxHealth, ShieldRegenRate
+	   if unitBp.Defense.Shield.ShieldMaxHealth > 0 then
+	      ShieldMaxHealth = unitBp.Defense.Shield.ShieldMaxHealth
+	      ShieldRegenRate = unitBp.Defense.Shield.ShieldRegenRate or 0
+	   elseif unitBp.Enhancements[getEnh.GetEnhancements(info.entityId).Back].ShieldMaxHealth > 0 then
+	      ShieldMaxHealth = unitBp.Enhancements[getEnh.GetEnhancements(info.entityId).Back].ShieldMaxHealth
+	      ShieldRegenRate = unitBp.Enhancements[getEnh.GetEnhancements(info.entityId).Back].ShieldRegenRate or 0
+	   elseif GetShieldParams(info.userUnit, 'MyShield') then
```
Sheeo: Is this a 4DC function? It is not defined in this file or anywhere in the FA lua api.

lua/ui/game/unitview.lua
```diff
@@ -448,25 +448,29 @@ function UpdateWindow(info)
...
+	   end
+	   if ShieldMaxHealth > 0 then
+	      controls.shieldText:Show()
+	      if ShieldRegenRate > 0 then
+	         controls.shieldText:SetText(string.format("%d / %d +%d/s", math.floor(ShieldMaxHealth*info.shieldRatio), ShieldMaxHealth, ShieldRegenRate))
+	      else
+	         controls.shieldText:SetText(string.format("%d / %d", math.floor(ShieldMaxHealth*info.shieldRatio), ShieldMaxHealth))
+	      end
+	   end
+	end
```
Sheeo: Please uses spaces for indentation, it makes it easier to do code review.

lua/ui/game/unitview.lua
ResinSmoker: Yes this is a 4DC function from DMS... now as 99.9999999% of shields created are the "default" we're safe using this. Otherwise I'm forced to /hook most all of unitview within 4DC to correct whats an issue from an FAF change.

Sheeo: We cannot have a reference to non FAF code in FAF code, that is confusing at best and fatal at worst.

I do see your issue however, this function is massive, and hooking the whole beast is rather unconstructive.

If we could split up the logic in this function a bit, perhaps you would be able to hook where needed, instead of having to resort to overriding it all?

ResinSmoker: The 4DC portion could be omitted as this would allow the shields to
function normally but they'd not display their health via the UI like
normal shields.

Sheeo: Okay then, can you update your patch to address this?

ResinSmoker: There is another solution as well but i need a clue getting the sim data
from a unit. If i can gain access to the SIM "self" for that unit i could
come up with another check that is never going to fail. Do you have any
ideas?

Resin

Sheeo: You cannot access sim objects directly from the UI state, for very good reasons.

You can have the sim pass data to the UI state through the units sync table, however.

ResinSmoker: Do we have access to MyShield and it components ?

ResinSmoker: Is it possible to use the entity ID available in the UI to get the SIM entity?

Sheeo: No. Pass sim data through the sync table to get access to it from the UI state.

ResinSmoker: I have another way which will work but will only become active if 4DC is loaded. In a nutshell I will be altering blueprint.lua via 4DC to add a Boolean handle to each unit during the BP first loadin. Thus when 4DC runs it will modify the Bp's of all unit to hold this handle. This handle acts like a key allowing the extra bit in FAF to do its magic during the requested UI event. This also prevents other scripts from triggering my function without 4DC being loaded.

If you don't understand that's okay it will make sense after I get things sorted on my end. I'll shoot you a copy soon as I can.

Sheeo: This should not be necessary; and it doesn't really solve the problem -- even if you have a value to represent whether 4DC is loaded or not (In every single unit), there would still be a reference to non-FAF code within FAF, which is precisely what we should not have.

I will reiterate the right way to go:

> If we could split up the logic in this function a bit, perhaps you would be able to hook where needed, instead of having to resort to overriding it all?

ResinSmoker: Such a hook would only be valid so long as Unitview.lua in FAF doesn't change at some point. Another thing I'd all but have to copy 90% of unitview.lua to effect the hook.

Now I can offer a null solution for FAF and then have the functional bits on my end.

Sheeo: That goes for every hook you will do against FAF, as we don't have a stable, supported public API defined.

No I'm precisely suggesting that we _split up the logic in this function_ so you don't have to destructively hook the entire `UpdateWindow` function, but instead could hook into a more suitable function.

ResinSmoker:
Sheeo I'm not able to edit this for resubmission. I'm currently signed in and have access via my iPhone's Gith app.

ResinSmoker:
Try assigning it to me as a task. It may be available for me then.

Sheeo:
Do you mean to update the code?

You need to push to your `ResinSmoker-unitview-patch-1` branch on your own repository, it will update this pull request.

ResinSmoker:
All I need to do to make this work is remove lines 460 thru 463. That will work for FAF and allow me to hook in on my end via 4DC.

Sheeo:
Perhaps this [git cheat sheet](https://training.github.com/kit/downloads/github-git-cheat-sheet.pdf) may be helpful?

Sheeo:
If all else fails, go to [this page](https://github.com/ResinSmoker/fa/settings/collaboration) and make me a collaborator on your github repository.

ResinSmoker:
Was able to add you as a collaborator but at the Sphar been able to edit or push anything.

ResinSmoker:
Fuck I hate voice to txt... Anyway what I meant was "so far I've been unable to push or edit anything"

Sheeo:
I changed the code to be less repetitive and removed the reference to 4DC code.

ResinSmoker:
Looks good and thank you

Sheeo:
Actually, looking at the code again, how did you trigger a divide by zero error?

ResinSmoker:
A custom shield has a shield ratio, this was triggering a shield UI event will a shield size of zero.

Sheeo:
Can I see the log file or have steps to reproduce?

I am asking because I realize this code does not do division. There is no such thing as a shield UI 'event', either.

ResinSmoker:
Use 4DC and DMS in a skirmish. Play seraphim and use the shield drone on any unit. (Drone is T2 air) during any mouse over of the newly shielded unit, the UI will spam the log so long as the mouse remains over the unit. Now if shield disabled log does not spam. All log entries will show FAF Unitview as the source of the error. This is because the shield drone can add shields to units without any shield entries it their blueprints as per its design / intent.

Sheeo:
I can see why an error would occur, but it would not be division by zero.

ResinSmoker:
The script attempts to divide NIL which amounts to the same result, failure.

Kevin

Sheeo:
It would attempt to multiply `shieldRatio` by `nil`.

ResinSmoker:
Well you know what was implied that's all that matters. Keep mind that I'm bouncing back-and-forth between my kids my home life and modding which if I'm lucky I can spend a few minutes on the day.

ResinSmoker:
Just tried out your latest change... so far so good. No errors and I should be able to hook'n discreetly on my end.

Sheeo:
Sounds good :)

Sheeo:
This can actually become part of the lobby hotfixes, since it's UI code.

Changed milestone accordingly.

ResinSmoker:
Awesome and thank you

Crotalus:
This should be rebased to `release/lobby-2.7`, right?

ChrisKitching:
If it really is to be shipped as a hotfix, push it to that branch, if it's been looked at by everyone who cares. I think sheeo means to publish hotfixes tonight, so if you push now it'll get tested with the other stuff and probably go out.

ResinSmoker:
Sweet

Sheeo:
I'll be able to push this tomorrow, sorry for the delay @ResinSmoker.

ResinSmoker:
I'm guessing the hotfix was delayed?

Resin

Sheeo:
Yes I'm terribly sorry, I've been a bit swamped lately.

ResinSmoker:
Im a bit disappointed but i do understand... Thats life right :P

ResinSmoker:
Any luck with this?

Btw anything i can help with?

aeoncleanse:
Yeah Resin, can you still open SCM files for modeling? Specifically, we need bones adding to a couple of units.

ResinSmoker:
Part of why I was looking at using Sketchup is that my copy of 3D Max is no longer viable. If you know of a way to bypass the restrictions then I'm all ears. Currently most sites that have codes / cracks are all Trojan virus sites.

Kevin

aeoncleanse:
I suppose I'm not supposed to endorse PirateBay on here... But it's the place to get 3D Max! Trouble will be finding the supcom plugin

ResinSmoker:
I have a valid copy of 3D Max /w plugin its just my reg code for it no longer works.

Resin

aeoncleanse:
If it's valid, how did the code run out? I'd contact the people who make 3DMax if I were you!

ResinSmoker:
Problem is that i can no longer prove anything as the email address associated with it has long since be cancled.

Crotalus:
Cherry-picked this change into `release/3641` and `develop`

228b3c8774a06b980bbd17b770fe46a1375e6f4b
7fe857f80606261e1245f8fb46d0b663f632a7be

Sheeo:
Ah thanks Crotalus!

PR #515, opened 2015-03-15
Fix garbage collection of wrecks

Crotalus:
Seems like the first attempt to fix pose errors didn't work.

More testing of the error reveals that it's due to some form of
garbage collection of wrecks. When there's been enough of unit deaths on
the field it starts to remove random wrecks.

This results in wrecks sometimes getting removed in the same tick as they are
created. This causes the error with TryCopyPose() when it tries to set a
pose of a deleted wreck.

It's unnecessary to copy pose of units that didn't animate at death so
this fix at least reduce the amount of spam in log

I've not found out a way to raise the bar of the garbage collect. ~~Maybe we
should implement our own GC of wrecks and delete the oldest ones instead,
this would prevent newly created wrecks to insta-disappear~~ (see below)

### DIFF
--- a/lua/ScenarioFramework.lua
+++ b/lua/ScenarioFramework.lua
@@ -20,6 +20,42 @@ local Utilities = import('/lua/Utilities.lua') # enabled so we can hide strat ic
 PingGroups = import('/lua/SimPingGroup.lua')
 Objectives = import('/lua/SimObjectives.lua')
 
+local MAX_WRECKS = 2000
+local NumWrecks = 0
+
+function GCWreckID(wreck)
+    NumWrecks = NumWrecks + 1
+    wreck.CreatedTick = GetGameTick()
+
+    if NumWrecks > MAX_WRECKS then
+        GCWrecks()
+    end
+end
+
+function GCWrecks()
+    local x, y = GetMapSize()
+    local reclaimables = GetReclaimablesInRect(0, 0, x, y)
+    local wrecks = {}
+    local current_tick = GetGameTick()
+
+    for _, r in reclaimables do
+        -- exclude wrecks created last 2 minutes
+        if r.IsWreckage and current_tick > r.CreatedTick + 1200 then
+            table.insert(wrecks, r)
+        end
+    end
+
+    local HEADROOM = 100
+    NumWrecks = table.getsize(wrecks)
+    if NumWrecks > MAX_WRECKS then
+        table.sort(wrecks, function(a, b) return a.MaxMassReclaim > b.MaxMassReclaim end)
+        for i=MAX_WRECKS-HEADROOM, NumWrecks do -- GC <headroom> extra wreck to reduce GC calls
+            wrecks[i]:Destroy()
+        end
+        NumWrecks = MAX_WRECKS - HEADROOM
+    end
+end
+
 # Cause the game to exit immediately
 function ExitGame()
     Sync.RequestingExit = true
@@ -1936,7 +1972,7 @@ end
 function GenerateOffMapAreas()
 	local playablearea = {}
 	local OffMapAreas = {}
-	
+
 	if  ScenarioInfo.MapData.PlayableRect then
 		playablearea = ScenarioInfo.MapData.PlayableRect
 	else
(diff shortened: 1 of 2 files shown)


Crotalus:
Testing now, 1000 wrecks is maximum on the field, after that it starts to randomly remove wrecks.

Crotalus:
Seems like the engine reuse destroyed entity IDs, therefore new wrecks can get IDs lower than existing wrecks, resulting in them getting deleted first in the GC. 

Overriding default GC for wrecks to prevent newly created wrecks to get insta-deleted, this should fix the pose errors.

aeoncleanse:
Ideally, isn't this the kind of thing we want to try to poke the engine code into accepting a higher number of entities?

Sheeo:
I'm not sure about this at all, we need serious performance regression testing in large games to see if doing this work lua-side is efficient enough.

ResinSmoker:
I created a mod a while back that assigned a fixed lifetime to wrecks based on the amount of overkill...

This could be used to snuf out wrecks before this becomes an issue but in a controlled way.

Sheeo:
So I was looking at this code the other day.

With your cleanup of `DefaultUnits.lua`, you centralized the `CreateWreckageProp` function into `unit.lua` so it's used for all units.

Before that commit, `MobileUnit` and `CommandUnit` both had their own copy, one which didn't call `prop:AddBoundedProp()` on the wreckage.

Looking at the code itself, `prop:AddBoundedProp()` is called after all mutation on `prop` has been done. Perhaps we should just move the call to the bottom of the function?

ResinSmoker:
Well CreateWreckageProp() should be part of Unit.lua IMO as it's used by all "unit" classes, not just specific ones. Now as for prop:AddBoundedProp(), it appears to already be part of the default FA Unit.lua as well as FAF's version. My version of CreateWreckageProp() via my corpse_decay mod, handles all of the wreckage based functions via it's own Utility file to keep the organization of Unit.lua manageable and easier to navigate / search for any Lua errors that may crop up.

AddBoundedProp() appears to be an engine-side function as I can't find a reference to it in either Lua or Moho. Beyond that I've kept the call in on my copy despite not having a 100% clear idea of what this function call is really doing. Now if i were to guess, it adds the newly created prop to a table so the engine can auto delete corpses when they become too many, Removing AddBoundedProp should prevent the  auto cleanup from taking place as technically with my mod running its not needed. My question about AddBoundedProp() is why is the mass value being passed into this table unless it's being used in someway to determine what the engine removes first when corpses are being vaporized.

Sheeo:
Sure, I didn't mean to say we shouldn't have it centralized, I only pointed out that `AddBoundedProp` was _not_ called for all wrecks before this patch.

Sort of obviously `AddBoundedProp` adds the wreck to the garbage collection that this PR is about.

Crotalus:
Nice find, should we try to exclude AddBoundedProp() on Mobile / CommandUnit then? If this could be a source of trouble.

If I remember correctly I tried to comment out the call completely to see if wrecks were GCed anyway after a while and I think that was the case. I can say with 100% though so I guess I'll test again

Sheeo:
I don't know if we should try to exclude it -- and I seem to have remembered the categories wrong.

For good measure [here is the commit](https://github.com/FAForever/fa/commit/4ae122baad60a6836030a34ffde8cdc79b3e05c2#diff-2098c157610f97ed5d0d41fd36dc17fbL2158) at a `CreateWreckageProp` function that does not call `AddBoundedProp`. (HoverLandUnit)

Sheeo:
How is your investigation on this looking?

Crotalus:
Removing AddBoundedProp() results in wreck still in the game (no GC whatsover)

Putting the call right before returning the prop doesn't change anything, I'll see if it's possible to add it later (1 tick after TryCopyPose has been called)

Crotalus:
Ok so:

* We should never call TryCopyPose() if not needed - for non-death animated units, this will reduce / remove the error in the logs (mostly happens with ASF).

AddBoundedProp() testing:
* Original behavior (FA) was to always call AddBoundedProp() after CreateProp(), FAF seems to have changed this (unintentionally?) so for example hover wrecks never recycle.
* Moving AddBoundedProp() in CreateWreckageProp() didn't change anything at all
* Without AddBoundedProp() the pose errors never appears but  the wrecks stay in game indefinetly, how big of a performance hit this will result in I don't know. 
* Delaying AddBoundedProp() 1 tick caused the sim to get instable, sometime crash and sometime hanged

It's a bit strange, I cannot find any documentation of AddBoundedProp() in the LUADOC, I wonder if this was patched into the game in a hurry. 

It would interesting to see what happens if we disable AddBoundedProp() completely and see what happens to the stability of the game. To prevent excessive wrecks we can always run a custom GC function once per 5 minutes or so. The good thing with a custom function would be that it removes the oldest (or cheapest?) wreck instead of now the most recent ones.

ResinSmoker:
Removing wrecks via a script was my original suggestion. However why limit the selection criteria to the oldest or cheapest when the units overkill is clearly available. This way units that have excessive amounts of overkill, will have a reduced amount of mass available for reclaim reduced and thus their lifetime in game should be shorter.

Sheeo:
@ResinSmoker the overkill ratio is used already when creating the wreck, to calculate the mass of it.

I'm unclear as to what you're suggesting, over sorting by the mass value?

ResinSmoker:
Base the wrecks Decay-timer on it tech level. Example, tech-1 3min, tech-2 6min, tech-3 9min....

This tech timer can then be further reduced / variable based on the level of overkill. Conversely the mass available to reclaim should also be adjusted with the overkill.

Crotalus:
Tested garbage collection with a max limit of 2000 wrecks:

`info: Garbage collect wrecks took 0.0080337524414063 seconds`

So neglible impact on performance, currently the headroom so it will only run GC every 100 wrecks after it reaches the limit of 2000.

Crotalus:
fa253e2385b6736c57a59e3f648201b1ec1f8466

Could probably be improved with auto-merged labels and option to change keys to active it but at least a start.

Sheeo:
@Crotalus please come on slack and discuss this. I would like this reverted.

ChrisKitching:
Backed out for the time being.

In general, we could plausibly continue to use AddBoundedProp if we ensure we always check that a wreckage has been deleted immediately before using it. Finding all such callsites is awkward but not impossible.

We should investigate the possibility of not doing garbage collection... at all. The game seems capable of handling a great many props (have you *seen* how many trees there are?!), so this may well not end badly.

> info: Garbage collect wrecks took 0.0080337524414063 seconds

8ms in one simulation timestep is not negligible. There's plenty of other stuff that needs to get done in that simulation step, it may well cause lag spikes you can "feel".

If we are to do our own GC, this implementation is highly nonoptimal. For a wreck limit `n`, your collector's sorting step takes time proportional to `n*lg(n)`. You could readily design a collector that incrementally removes the oldest wreck at the time of allocating a new one (if we have too many), in constant time, without the overhead of the fairly large allocation implied by a call to GetReclaimablesInRect.

aeoncleanse:
Agreed, 8ms is definitely something some players, such as myself, will be able to feel if it's added in. Some won't, some will.

ResinSmoker:
Less then a 10th of a second in a large match shouldn't be noticeable unless the SIM speed is crawling. I'd suggest making the collection rate variable so this can be avoided, while giving the best overall performance / appearance. It's possible for the UI to get the current SIM setting / RL-rate but not for the SIM as I understand it. This should be corrected to allow for effects that rescale or don't display at all, based on the games performance.

Resin

ChrisKitching:
To clarify: My point wasn't "Humans can detect 8 ms", but that "there is a lot of stuff to do in a simtick, 8ms is a nontrivial amount of the time available in the simtick, and the algorithm here is O(nlog(n)), if not worse, so it won't scale pleasantly."

PR #517, opened 2015-03-17
Air contrails tweaks

Crotalus:
Reduce the impact of massive groups of ASF by adding LOD 128 to the exhaust clouds / top speed beams and not render the exhaust if user is on low fidelity.

Made exhaust beam a bit more visible also and added missing smoke (typo) in UEF ASF blueprint


Sheeo:
While I'm all for this, I don't think it should target the release branch. We're a bit in over our heads currently with all the stuff we need to do testing on for 3641.

aeoncleanse:
I agree with Sheeo in general, but this is a pretty isolated change which is easy to test and I can't see breaking anything else. Assuming, that is, Crotalus sandboxed it himself before PR ;)

PR #518, opened 2015-03-17
Air contrails

Crotalus:
#517 moved to develop

Reduce the impact of massive groups of ASF by adding LOD 128 to the exhaust clouds / top speed beams and not render the exhaust if user is on low fidelity.

Made exhaust beam a bit more visible also and added missing smoke (typo) in UEF ASF blueprint


Sheeo:
:+1:

Crotalus:
Merged this with `release/3641` due to only being cosmetic changes and seems to working fine.

Issue #519, opened 2015-03-18
Enemy cannot reclaim upgrading mex that is paused


Cheyans:
Basically what's in the title
Replay: 3219275

Crotalus:
Neutral engineer didn't have any energy so capture is stalling for that reason

Crotalus:
Reopened, seems that it fails to reclaim in Patrol / Attack mode.

Crotalus:
For reference, seems to be a very old bug, 3610 has the same problem

aeoncleanse:
So I think this is fixed in 3641, close issue?

Sheeo:
If this is fixed in 3641, we don't have a pointer to the commit -- can anyone dig that up?

aeoncleanse:
PR 572, not merged yet, my bad. It is, however, ready-looking.

Crotalus:
Yeah, not merged yet due to SetCollisionShape('None') known to crash sim (with sinking air wrecks). Maybe it's to use in this case but I guess it can wait until next patch, we have lots of potential bugs already in 3641

aeoncleanse:
My plan was to get things moving again. To that end, I want to merge a whole bunch of PRs, then look at the list of things which are meant to be fixed and test them all. In the process, any errors can be logged and we can hunt them down and fix the whole lot.

How does no collision shape crash with sinking air wrecks?

Crotalus:
When no collision shape was set on dying air units that was sinking it caused the sim to crash randomly, probably because they sank through seabottom into infinity or something. Probably had to do with the wreck being attached to the sinking projectile.

So my **guess** would be that this change would be safe but you never know :)

aeoncleanse:
That's what I thought, which is why I want to merge and test. Then in the slight chance it's not worked, we can flag and fix it :)

Sheeo:
@aeoncleanse merging into the release branch isn't a precondition for testing.

aeoncleanse:
Which is why I checked it out and did some testing. I cannot make it crash, at least not so far. Trouble is, with something like this, there may not BE a way to replicate it every time, if it ever happens.

PR #520, opened 2015-03-18
Disable collision shape on workitem

Crotalus:
Fix for #519 

Seems like area commands like patrol / move attack can target the workitem (self.DisallowCollisions doesn't stop this).

Disabling the collision shape of the workitem takes care of this (and would probably deprecate the need for SetCanTakeDamage and DisallowCollisions).

The only question now is if SetCollisionShape('None') is crashing the party or not...

The bug isn't critical so we could postpone for 3642


Sheeo:
As heuristically proven by [this diff](https://github.com/FAForever/fa/compare/3639...3640) `SetCollisionShape('None')` should be considered harmful in certain cases.

Before we introduce this we should be sure of exactly which cases these are, and address all of the occurrences at the same time.

So I'd say to definitely postpone for 3642.

Crotalus:
-> #572

PR #524, opened 2015-04-01
Remove initializations with nil

JeroenDeDauw:
IDE told me those are evilll

### DIFF
--- a/lua/ui/game/economy.lua
+++ b/lua/ui/game/economy.lua
@@ -3,7 +3,7 @@
 --* Author: Chris Blackwell
 --* Summary: Economy bar UI
 --*
---* Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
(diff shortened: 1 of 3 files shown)


lua/ui/game/economy.lua
```diff
@@ -3,7 +3,7 @@
 --* Author: Chris Blackwell
 --* Summary: Economy bar UI
 --*
---* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
```
JeroenDeDauw: huh

Sheeo:
They aren't evil -- just explicit, since `local a` will assign `nil` to `a` by default.

JeroenDeDauw:
IDE lists this as perf issue, which is kinda odd

perf

Sheeo:
Well it will likely result in an extra initialization, so it could be a performance issue in a tight loop.

PR #554, opened 2015-04-11
Added personal stealth unit desc to T2 torp launchers - aeon and sera

shalkya:
adding stealth in unitdb

### DIFF
--- a/units/UAB2205/UAB2205_unit.bp
+++ b/units/UAB2205/UAB2205_unit.bp
@@ -66,6 +66,7 @@ UnitBlueprint {
     Display = {
         Abilities = {
             '<LOC ability_sonar>Sonar',
+            <LOC ability_personalstealth>Personal Stealth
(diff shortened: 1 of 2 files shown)


Sheeo:
Thanks for this, there is the slight problem that it is not valid lua code :) I'll take care of this when merging.

Crotalus:
Interesting that Travis doesn't bork, we don't syntax-check bp-files I guess?

Sheeo:
@Crotalus I fixed this here 371b226986a451dcda3f3691203f8698181be7c9

Crotalus:
Ahhh, I looked at the already fixed script then :)

shalkya:
yes sorry, i didn't know how it worked before. But i'm trying to change a mod, so i understand lua a little bit better now :)

Crotalus:
6c7fafcba09f7e81521bfe33ebdadbffb1fdc56e

PR #555, opened 2015-04-11
Added personal stealth unit desc to T2 torp launcehrs - uef and cybran

Exotic-Retard:
t2 torp launchers have personal stealth - updated unit description to match this

hope i dont break anything :|

### DIFF
--- a/units/UEB2205/UEB2205_unit.bp
+++ b/units/UEB2205/UEB2205_unit.bp
@@ -66,6 +66,7 @@ UnitBlueprint {
     Display = {
         Abilities = {
             '<LOC ability_sonar>Sonar',
+            '<LOC ability_personalstealth>Personal Stealth',
             '<LOC ability_torpedo>Torpedoes',
         },
         IdleEffects = {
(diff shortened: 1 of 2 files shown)


aeoncleanse:
I'd rather remove the stealth! 
1 - How long have they had it?
2 - Why do they have it and the others don't?
3 - Why do they have it for free (No E drain)?
4 - Why can't it be toggled?

Basically... WTF!!!!

Sheeo:
They've had it since the first commit in this repo. Don't know if it was present in the base game, but it's likely.

Sheeo:
Either way, bugfixing is the motivation for 3641, so this PR is good.

shalkya:
we don't know since when they had it, but i don't remember of someone that changed it.
All T2 torpedo launcher does have it (exotic retard and I wanted to make the change, so we split up the work)
yes no E drain, and no toggle. I think we should remove that, they are already under build.

Crotalus:
6c7fafcba09f7e81521bfe33ebdadbffb1fdc56e

PR #557, opened 2015-04-12
Show the correct factional loading screen for automatches

ChrisKitching:
This should make automatches show the right factional loading screens.

Needs devserver testing. If found to work we could probably bundle it with one of the many outgoing hotfixes. :P

### DIFF
--- a/lua/ui/lobby/autolobby.lua
+++ b/lua/ui/lobby/autolobby.lua
@@ -276,6 +276,13 @@ local function InitLobbyComm(protocol, localPort, desiredPlayerName, localPlayer
 
     lobbyComm.GameLaunched = function(self)
         GpgNetSend('GameState', 'Launching')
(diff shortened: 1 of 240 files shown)


lua/ui/lobby/autolobby.lua
```diff
@@ -4,7 +4,7 @@
 --* Summary: Autolaunching games from GPGNet.  This is intentionally designed
 --* to have no user options as GPGNet is setting them for the player.
 --*
---* Copyright � 2006 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright � 2006 Gas Powered Games, Inc.  All rights reserved.
```
Sheeo: Let's use the unicode symbol, as with 790160ad470ceee8304951a78880b2410aaad516

Sheeo:
Did this end up working?

PR #560, opened 2015-04-13
Refactor rebuild

### LINKED ISSUE #550: several shield sACU can't start a building at the same time
here is the replay : content.faforever.com/faf/vault/replay_vault/replay.php?id=3105472

it's at the very beginning.

you can find logs here : http://forums.faforever.com/forums/viewtopic.php?f=3&t=9446

Crotalus:
Removed alot of buggy code that checked rebuilding wrecks, which caused rambo-SCU to refuse stuff when grouped.

The new code takes into account the progress of the reclaim in the wreck
so if you've reclaimed 50% of the value the rebuild progress will start at
25%

Fixes #550


Sheeo:
Very nice :+1:

Crotalus:
3991341ca2931e8f5fafb3ebceaa7ee53875d581

PR #565, opened 2015-04-16
Make the economy display less insane

ChrisKitching:


lua/ui/game/economy.lua
```diff
@@ -3,7 +3,7 @@
 --* Author: Chris Blackwell
 --* Summary: Economy bar UI
 --*
---* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
```
Sheeo: Didn't you rebase on top of a commit to fix this earlier?

ChrisKitching: Ah yes, but in a PR that seems to never have merged.

Sheeo:
Yeah I reassgined https://github.com/FAForever/fa/pull/557 back to you after commenting :)

Crotalus:
Some input:

* Did some cleanup earlier with this in `develop`:
https://github.com/FAForever/fa/blob/develop/lua/ui/game/economy.lua

Don't know the status of `develop` compared to `release/3641`  / `release/lobby-2.7` though

* You seem to have removed the option to turn off smart economy indicator options which people will complain about

Sheeo:
I discussed the removal of the smart economy option on slack with Chris -- I don't know if it's something that people will complain about, surely most would like to be notified when they are about to overflow power or stall mass?

ChrisKitching:
> Did some cleanup earlier with this in develop:

I discovered this slightly after starting work on this (I rather foolishly started on the wrong branch: I'll rebase this onto develop in a bit).
After noticing your commit, I stole large chunks of it :p

The main way in which our cleanup work differs is that yours pushes the preference checks deeper into the beat function, but mine pulls these invariant conditions out of the beat function into setup code. The result is my meta-function that configures the beat function to be a specialised version that avoids checking things that don't change during the simulation.
Several of your approaches for simplifying the actual update logic were stolen from your commit.

> This also has some code with averaging the rate of reclaim.

You're referring to this, I assume:
https://github.com/FAForever/fa/commit/e70de3d18589acdee319a1453f10be536917e331#diff-b8566dcca39ef9061ce7cda1976a12d3R296

I hadn't noticed that until now. I'm certainly not a fan of averaging the overall deltas (there was some code for that in the "smart" mod that I killed), but averaging the reclaim might make it more legible (as that value is liable to jitter *lots*)

The downside with this sort of damping is it gives the impression that the reclaim you're getting is rather higher than it really is. It becomes the case that the resource delta displayed is no longer the sum of income plus the reclaim delta.

> Don't know the status of develop compared to release/3641 / release/lobby-2.7 though

This PR should've been against develop. Whoops.

> You seem to have removed the option to turn off smart economy indicator options

This is true, but I changed the way they behave to make them less rage-inducing.

The smart economy option appeared to do basically two things:

- Damping of the delta display.
- Change the colouring behaviour for the delta text.

The former was discarded after talking to the others. The logic would make the displayed delta in each timestep be 0.95 * displayed_value_last_time + 0.05 * actual_delta_this_timestep.

The intention is presumably to stop the number changing very frequently so it's easier to read.

You will notice that this logic *doesn't* change how frequently the number actually gets changed, it just makes it report incorrect deltas to you, which is liable to be potentially confusing.


The colouring behaviour from the original mod was fairly annoying. I've modified it to be less so. It now differs from the stock behaviour in only two ways:
- When mass is >90% full and has +ve delta, it flashes grey-white.
- When energy is <10% full and has -ve delta, it flashes grey-white.

The original behaviour would cause strange (and irritating) behaviour, such as turning the numbers red when storage went below 50% full (which seems silly, as you surely *want* that for mass most of the time...)

This now seems an entirely sensible default (unlike the original behaviour).

Crotalus:
Should probably be rebased + merged into 3641

Sheeo:
How are we looking on this?

lua/ui/help/tooltips.lua
```diff
@@ -2382,11 +2382,5 @@ Tooltips = {
...
         title = 'Zoom Pop Distance',
         description = 'Adjusts distance to which Zoom Pop zooms to.',
-    },
-    options_gui_smart_economy_indicators = {
-        title = 'Smart Economy Indicators',
-        description = 'Smart Economy indicators. Rate display works differently now. Efficiency is "infinite" when consumption is zero. Values greater than 100% are possible again. Mass rate begins to flash bright white if stored mass is > 80% and rate is climbing. Energy rate begins to flash bright white if stored energy is < 20% and rate is sinking. Mass and energy bars are vertically exchanged (mass is heavier, thus it is below energy - TA style). Mass bar, as well as income and store numbers, are gray - TA style.',
-    },
-
-
+    }
```
aeoncleanse: Missing comma?

lua/ui/game/economy.lua
```diff
@@ -276,281 +283,202 @@ function CommonLogic()
...
-                controls.expense:SetText(string.format("-%d", math.ceil(requestedAvg)))
+
+        -- Finally, glue all the bits together into a a resource-update function.
+        return function()
+            local econData = GetEconomyTotals()
+            local simFrequency = GetSimTicksPerSecond()
+
+            if showReclaim then
+                -- Show, if enabled, the reclaim values.
+                reclaimed:SetText(math.ceil(econData.reclaimed[tableID]))
```
aeoncleanse: tableID nonexistent error :/

ChrisKitching: I suspect bitrot. I'll poke it later.

Sheeo:
Review status: 0 of 4 files reviewed, 2 unresolved discussions, some commit checks failed.

---

<sup>**[lua/ui/help/tooltips.lua, line 2385 \[r1\]](https://reviewable.io:443/reviews/faforever/fa/565#-JqHVxux0EisByE-m19N-r1-2385)** ([raw file](https://github.com/faforever/fa/blob/b99ed9e0a4b784d79e69c301a41fce4965dd9761/lua/ui/help/tooltips.lua#L2385)):</sup>
Ommitted comma, not missing :)

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/faforever/fa/565)

aeoncleanse:
Incoming and outgoing Energy/s is not shown for some reason, and if it's meant to flash/go red, it doesn't.

aeoncleanse:
Additional thought, because E is maxed so much more often and so much more usefully, instead of having it glow/flash as it approaches full, you should tie the glow into fullness, have it flash red when there is >X energy overflowing, and have the number flash when >Y. Say, 100 for X and 5000 for Y?

Sheeo:
This needs rebasing

ChrisKitching:
Needs rebase, fixed the reclaim display problem uninsanely. Could use a test, too (I'm without my Windows box for the weekend)

Sheeo:
What is 7e345e5?

ChrisKitching:
Someone fucked up some newlines in develop and autocrlf doesn't like it. It'll go away when I rebase, just wanted to give IceDreamer et. al. the chance to test it out.

ChrisKitching:
Merged with fixes.

PR #568, opened 2015-04-17
Flash intel for enemy civilians at start

### LINKED ISSUE #567: Pulse 100% vision at the start of ladder matches or otherwise show civilians
As per the title, people lose games on ladder (and probably on team games too) partly due to not knowing where the civilians are. it's incredibly frustrating for new players to find out that, no, you can't just go through the middle of Syrtis Major, or due south on Balvery Mountains, or pretty much anywhere on Sleipnir with an early engineer or important group of units. What options are there to do this and how easy is it?

Crotalus:
Fixes #567

### DIFF
--- a/lua/sim/ScenarioUtilities.lua
+++ b/lua/sim/ScenarioUtilities.lua
@@ -1,0 +1,1531 @@
+----[                                                                             ]--
+----[  File     : ScenarioUtilities.lua                                           ]--
+----[  Author(s): Ivan Rumsey                                                     ]--
+----[                                                                             ]--
+----[  Summary  : Utility functions for use with scenario save file.              ]--
+----[             Created from examples provided by Jeff Petkau.                  ]--
+----[                                                                             ]--
+----[  Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.             ]--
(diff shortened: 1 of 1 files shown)


Sheeo:
I marked the milesone as for the balance patch -- but I guess we should discuss inclusion into 3641?

Crotalus:
Yes, I wonder how it plays with special mods like survival etc. Some of them may rely on civlians structure being hidden from view. I'm not into those maps myself so maybe someone else has more input on this.

Sir-Prize:
I would suggest that it is not immediately integrated in to the "FAF" mod - just the "ladder 1v1" mod. If it was possible to do this as a setting, not a mod, that would be ideal. Essentially adding another two options under "Civilians" in the lobby menu as "Enemy (Scouted)" and "Neutral (Scouted)" possibly. Anyway, just food for thought, I would have thought that's more difficult to do without source code. What you have planned in this PR is fine IMO.

Sheeo:
Needs rebase.

@Sir-Prize unfortunately, the ladder1v1 mod isn't a seperate mod we can merge this into by its own. A lobby option (That is enabled by default) seems a good idea.

PR #570, opened 2015-04-21
Refactor cybran buildeffects

Crotalus:
* Show red build beams for cybran factories
* Build bots now can start next project without need of respawning
* Scale the number of build bots with the build power of the builder


Crotalus:
0022372aef2437510bfce4a236ae2239f2bac582

PR #571, opened 2015-04-21
Set TargetPriority for weapons when loading BPs

Crotalus:
Instead of setting the TargetPriority manually in each blueprint we instead set the values when loading the blueprint. By looking at each weapon spec we can decide what's a reasonable priority should be for that specific weapon.

This will make it much easier to modify target priorities in the future and keep them synced between all units.

The goal for this iteration is not to change the actual target priority of units, just to move it into this new system


aeoncleanse:
Seems cool, but we need it so that we can use the blueprint to override this for specific units. If it already does, cool!

Crotalus:
Yep, blueprint values takes precedence over the default ones

Sheeo:
I like simplifying the blueprints, but I have three concerns:

 - It's not obvious for someone looking at the blueprints to realize that there are defaults defined elsewhere
 - With mods that hook/override the blueprint loading, this may break
 - It's no longer possible to introspect this information solely by parsing blueprints, you now need to execute this code too, making it harder for tools like unit databases to show meaningful data

These concerns go for the GSR code too.

Sheeo:
Closing this in favour of getting FAForever/faftools#4 finished

PR #573, opened 2015-04-23
Feature/netlag

Crotalus:
Rebased #66 

Setting net_Lag to slightly above the highest ping among the players in a
game. This will reduce command lag, especially for 1v1 games.

Refactored the code to use average ping instead of a snapshot. It's also
taking variable pings into account by adding a headroom of std-dev*2.5.

Logging (although reduced) is still activated, this should be turned off
when everything seem to be working as intended


ChrisKitching:
We probably don't want this.

Changing the netlag parameter while the simulation is running is potentially extremely dangerous. We could definitely entertain patches that set it to something sensible just before the simulation starts based on some measurements taken in the lobby.

When you issue a command, this is introduced into the command stream at a point in the future determined by the netlag parameter.
If the netlag parameter is adjusted downwards, a window is created in which my next command can be scheduled for *before* my previous one. If this leads to a nonsensical command stream, the game desynchronises.

Auto-adjust netlag: Considered harmful.

PR #581, opened 2015-04-27
Make content of storages count

Crotalus:
When your storages are killed / captured / given away the actual content
in the storage unit goes with it.

The actual wreck of mass-storages should maybe contain some fraction of the stored mass to make it even more fun.

Suggestion thread here:
http://forums.faforever.com/forums/viewtopic.php?f=42&t=9860

### DIFF
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -65,7 +65,7 @@ function TransferUnitsOwnership(units, ToArmyIndex)
     table.sort(units, function (a, b) return a:GetBlueprint().Economy.BuildCostMass > b:GetBlueprint().Economy.BuildCostMass end)
 
     local newUnits = {}
-    for k,v in units do
+    for k, v in units do
         local owner = v:GetArmy()
         -- Only allow units not attached to be given. This is because units will give all of it's children over
         -- aswell, so we only want the top level units to be given.
@@ -165,7 +165,9 @@ function TransferUnitsOwnership(units, ToArmyIndex)
         if EntityCategoryContains(categories.ENGINEERSTATION, unit) then
             unit:SetPaused(true)
         end
+        v:HandleStorage(ToArmyIndex)
     end
+
     return newUnits
 end
 
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1169,6 +1169,7 @@ Unit = Class(moho.unit_methods) {
             self.DisallowCollisions = true
         end
 
+        self:HandleStorage()
         self:DoUnitCallbacks( 'OnKilled' )
 
         if self.UnitBeingTeleported and not self.UnitBeingTeleported.Dead then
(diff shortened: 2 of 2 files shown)


Zockyzock:
how is "actual content" distributed? If i have 1000 mass in storage, and 4 storages (2000 capacity), does each storage have 250 mass, or have the first 2 killed storages 0 mass, and the last 2 killed each 500?

quark036:
Pretty sure it's the second one. There was a forum thread about changing it so that people would have to think more about capping a forward mex, or be more strategic about where they put storage

Zockyzock:
Isn't it already the case then, if you have 1000 mass in storage and lose 1 of your 2 storages, you will also lose the 500 excess mass that you can't store anymore?

aeoncleanse:
Yes you will.  This change makes it so if you have 5k storage,  and 2k in storage,  and lose a storage,  it will take you down to 1.5k

Zockyzock:
I think it would be better if the 2k mass get distributed over the 5k storage (10 storages), = 200 mass per storage, and you lose only this 200 mass.

Crotalus:
Yeah it assumes the mass is distrubuted instantly through all storages so you will lose your <storage ratio> * <max capacity of storage> if it dies.

I guess you could simulate some delay of the storages getting filled also, not that hard. If we show the actual mass number stored when you hover a mass storage it will get more clear and logical.

aeoncleanse:
Don't see any need for a delay. Looks like the current code will do nicely to me.

Zockyzock:
The next patch is supposed to only fix things, not add anything new. Should go into the one after though.

Varaxis:
If you zoom into storage, you may notice that they actually grow/shrink in size depending on your storage levels. With multiple, they aren't animated as if they're filled 1 at a time like quark suggests, instead growing the same amount. Figure with whatever tech is behind them, that transfer resources immediately on loss to remaining ones or to allies if full, can also function to avoid losing contents if captured, but when's the last time consistency and lore made a game interesting depth (sarcasm)?

jasonnorthrup:
Is there code for adding mass to wrecks?

quark036:
When the wreck is created, you can set what percent of the units mass it is, could probably set that higher than 1 if you want

jasonnorthrup:
But can it vary from one wreck to another based on mass shored? If so, then reclaiming an opponent's MS is preferable to capturing it.  If not, then more mass is obtained by capture.  Before the patch it would only make a difference if you decreased total storage capacity below total stored mass.  From the opposite perspective, it makes some sense now to put some storage in secure areas, rather than solely around mexes.  If mass can be reliably stored in wrecks, it probably  even makes sense to self destruct them to save on unit cap, then reclaim as needed to liquidate mass.

ChrisKitching:
You can set a wreck value to whatever you want. look at the Wreckage class.
So. Yes. Clearly, if you do this, you need to make blowing it up and reclaiming it equivalent to just reclaiming it (w.r.t contents. I suppose..)

PR #582, opened 2015-04-28
Bugfix/sinking engineers

Crotalus:
* Non-naval wrecks didn't have the no-collision flag set
* Destroy build effects when engineer is sinking


lua/sim/Unit.lua
```diff
@@ -1660,6 +1662,27 @@ Unit = Class(moho.unit_methods) {
         self:Destroy()
     end,
 
+    DestroyAllBuildEffects = function(self)
```
Sheeo: Since all that TrashBags do is iteratively call :Destroy(), they work with trashbags as contents too.

This code would be much simpler if we utilized that idea.

PR #585, opened 2015-04-28
Adding immunate to own nuke ability for sera sml

shalkya:
more information here : http://forums.faforever.com/forums/viewtopic.php?f=2&t=9856
tried my best to make it understandable, but i'm not sure what i should have wrote


units/XSB2305/XSB2305_unit.bp
```diff
@@ -85,6 +85,7 @@ UnitBlueprint {
     Display = {
         Abilities = {
             '<LOC ability_manuallaunch>Manual Launch',
+            '<LOC ability_immunateownmissile>Immunate own missile',
```
ChrisKitching: You didn't add "ability_immunateownmissile" to the strings files, so this won't work in localisation.
"immunate" is also not a word. :P

shalkya:
Yes i'm sorry, i wanted to do it by myself. But i know nothing about coding. (my english is bad too :s) 
you can add "ability_immuniteownmissile" to the strings files, if you want. Or let me do it, but i'll need to learn a little about all that before. (and it will take some time, since i'm in exam next week)

ChrisKitching:
Not to worry: there's no hurry, and I don't know if we want to do this, or if we just want to make it *not* be immune any more. Is this really a bug?

In any case, you will find the localisation files at LOC/language-id/strings_db.lua. The english ones are in LOC/en/strings... for example.
Each entry there has a key and a value. When you do "<LOC somekey>", the "LOC" lua function will replace it with the value with the name "somekey" from the language file for the language the user is set up to use.

Feel free to have a dig and figure this out. If you do, there's actually a lot of strings around which are just hardcoded, so if you wanted to be super helpful I could point you at them and you can fix all of them (or as many as you can be bothered with :) ).

You should put any new string into all the language files, even if you don't know that language (just put it in there as english instead). We tend to get stuff translated in batches, and at least that way *all* we have to do is translate the things in the language files, instead of going hunting for all the uses of each string.

Good luck! :)

PR #588, opened 2015-05-07
Refactor/unitview

Crotalus:
Much cleaner if UI-side uses the regen value set by sim-side


Crotalus:
Refactored unitinfo update mechanics. Instead of always running unit info update every frame we start a thread if needed and updated every tick instead

Sheeo:
This conflicts with current release branch, would you rebase it?

Crotalus:
:white_check_mark:

aeoncleanse:
Tested this on top of release/3641 and it's working. Enable the show reclaim option in the options menu under interface, then ctrl+shift to show reclaim for larger things.

Crotalus:
3590ea83438eb88b7519aa027111b6317aaa5c50
40785a54b932d1010f870dc7367cd1e7e0ef7a37
07fa8b84b7c6ae4ce8284c0cc42dcecb2c9ec3de

PR #592, opened 2015-05-10
Double Absolver's hover elevation

aeoncleanse:
Reduces the frequency of shots from the very low-mounted gun simply
hitting the ground. A big problem thanks to such high range.


Crotalus:
Merged

4394d09ee48da07f6de9ccbe43416bd8a610925c

PR #595, opened 2015-05-10
New option: Show reclaim value of wreckage / props

Crotalus:
PR #476 -> `release/3641`

If this is enabled, value of props / wreckage are shown if Ctrl+Shift is
pressed simultaneously.

The value will be the initial value of the reclaim and partial reclaim
won't change it.

Especially useful when playing new maps without knowing where the reclaim
is.

Reclaim with mass value below 20 isn't shown


Crotalus:
fa253e2385b6736c57a59e3f648201b1ec1f8466

Could probably be included with auto-merging of labels close to eachother and way to change hotkeys

PR #596, opened 2015-05-10
SCU Billy explosion test

aeoncleanse:
I don't think this looks very good. Additionally, Billy should be buffed to be useful IMO, in which case it should keep the current boom.


aeoncleanse:
Closing this, people can still find it if they want to check out the crappy explosion

PR #599, opened 2015-05-16
Update strings_db.lua

shalkya:
adding ability_immuniteownmissile="immunite to own missile" to the strings files


shalkya:
i'm sorry i don't know how to add other commits in 1 pull request... :s

aeoncleanse:
I'm not worries by the lack of github knowledge here, god knows I found it tough enough to begin with. However, let's try to keep FAF as professional as it can be. If you're not a native English speaker, don't be scared to contact one of us before putting something like this forward, we don't bite!

Could you correct these four PRs by changing:

ability_immuniteownmissile="immunite to own missile"

to

ability_noselfdamage="No self damage"

I also question why we need this. Why should the Sera nuke launchers be immune to their nukes? Are the others? What's the logic?

ChrisKitching:
Surely the bug is that it's immune to its own nuke. We should fix that (and any related problems).

shalkya:
changing it by "no self damage" isn't very wise. 
since most of the unit can't damage themself (talk to ZLO about it if you want). 
Here it's specific about the nuke and the missile. 
i have already made a topic about it in forum : http://forums.faforever.com/forums/viewtopic.php?f=2&t=9856

aeoncleanse:
But this label would only be used on units which need it specifically mentioned, it doesn't auto-apply.

shalkya:
i don't understand what you are trying to say.

aeoncleanse:
I'm saying

1 - You've done this because the Sera nukes are immune to themselves. Do we even want this?
2 - If we do, then we don't need a nuke-specific description. We need a description which is generic, broad, and can be applied to all those things which need it. This is just a small tooltip, and will only appear on the units we want it to appear on, so it can be kept broad so we can use it for other units in the future.
3 - No matter what the description says, the description name and the description itself should be professional, descriptive, and correct English.

shalkya:
1- well since it always been like that, and nearly nobody care, i think the best way is to informe people of it. This doesn't change the game. It was a known feature for experimental nuke, less known for the normal nuke. 
2- As i said, nearly all unit in the game can't "self damage". the only known (to damage themself) so far are : sml and missiles reflected by loyalists. So i think this is important to not the difference entre the sera sml and the other.
3- Yes that's for sure, i try to do my best.

aeoncleanse:
If they have always been like this, and the other three are not, it's a bug. It's never been a 'Feature' of the Yolona Oss that it doesn't kill itself.

shalkya:
you are the one that is saying it's a bug. There is no proof it was not intended. I don't know why we should consider it as a bug. At least there is other bug that are way more easy to determine as bug (for exemple : regen on aeon sACU break when the sACU get vet...)
Here i bring a first step : making the feature known for everyone. Then we can decide if it's a bug or not. But i feel like it will take some time to get this sorted out.

aeoncleanse:
The proof is in firstly logic. Nukes are one of the few things which logically and for gameplay reasons there is NO REASON for them to not kill themselves. You don't accidentally fire at yourself, you cannot accidentally hit yourself.

Secondly, if three nukes do hurt themselves, the three from the original game, and the two new ones don't, the new ones are bugged. It's not a feature, it's not documented, it's a bug. Course I could be wrong, and all nuke launchers are immune, but someone needs to test it.

shalkya:
i don't see why this shouldn't be a feature of the experimental nuke. And why it shouldn't be feature of the sera faction. (we all know they have less unit, maybe it was on purpose that they have more special feature) (this discution is getting retarded)
as ZLO said "It kinda goes as personal opinion to have that feature or remove it"
icedreamer, you trying to argue, and you haven't tested it yet... yes there is only sera faction that can't kill their own sml with their nuke. You could have at least posted your point of view, in the topic earlier.

ChrisKitching:
The only sort-of-plausible edgecase in which this might be usable is when a missile being launched hits a passing aircraft. Discuss this issue here:
https://github.com/FAForever/fa/issues/605

Ideally without ripping each other's hair out any more.

PR #604, opened 2015-05-17
Detect seabottom impact of sinking navy

### LINKED ISSUE #603: Navy units fall through seabottom
`release/3641`

Naval units have no collision detection (not attached to a projectile like hover / air units), so the sea-bottom collision -> create wreck code will never be run.

Crotalus:
Couldn't get the methods in #603 working for the sinking naval units, seems like the animation doesn't trigger any impact at all (I guess the actual phsyical unit isn't moving)

Restored the seabed check for naval units until we can come up with a better way (doubled the poll time to reduce cpu-load). Can keep this PR open until closer to release if we find a better way. At least we got rid of the seabed watcher for non-naval wrecks which should be an improvement anyhow

Fixes #603

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1369,6 +1369,7 @@ Unit = Class(moho.unit_methods) {
                 sinkAnim:SetRate(rate)
                 self.Trash:Add(sinkAnim)
                 WaitFor(sinkAnim)
+                self.StopSink = true
             end
         end
     end,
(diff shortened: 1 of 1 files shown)


ChrisKitching:
This seems an oddly difficult thing to do.

The Czar has some useful-looking logic driving its death-damage business, but the OnAnimTerrainCollision callback invoked by the native CollisionDetector object fires on impact with the water's surface, and not again on seabed contact: not too useful.

Clearly, the position of the unit *is* changing during the animation, otherwise the SeabedWatcher wouldn't work: it's just polling the position. As a last resort, we could use the known water-depth and projectile properties to just calculate the time at which the wreck will reach the bottom and just do an appropriately-sized WaitSeconds() in DeathThread.

The logic used to create footprints looks like the most promising lead at the moment. It seems that the native CollisionDetector invokes OnAnimCollision when a registered bone hits a terrain surface, excluding the water's surface, if you don't call EnableTerrainCheck(true) on it. This is used to generate walking-unit footprints on the seabed, and might be usable for our purposes, though for reasons unclear it doesn't seem to want to co-operate.

Crotalus:
Yeah, the weird thing is that GetPosition() returns a value as if the unit is still on the surface while GetPosition(bone) gets the actual depth value.

I think it'll be hard to come up with suitable WaitSeconds-values unless we know how fast the animation  makes the unit sink (they quite different sink times)

PR #606, opened 2015-05-18
Add tick fraction to AddBoundedProp sort-value

Crotalus:
Ok guys, a new attempt of fixing the GC wreck issue (#515) :) 

AddBoundedProp takes a (mass) value to be able to GC the least valuable
wreck first. By adding a fraction of GameTick to this value we get wrecks
sorted on a mass+time basis instead.

In effect GC will see wrecks as having a  "decay-value" with 0.5 mass every second. 

Would still be a corner case of the least valueable unit wrecking risk to instadisappear unless there is a wreck with lower value on field.

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -25,6 +25,8 @@ local BuffFieldBlueprints = import('/lua/sim/BuffField.lua').BuffFieldBlueprints
 
 local RECLAIMLABEL_MIN_MASS = import('/lua/sim/prop.lua').RECLAIMLABEL_MIN_MASS
 
+-- Used to give AddBoundedProp a way to delete oldest wreck first
+
 SyncMeta = {
     __index = function(t,key)
         local id = rawget(t,'id')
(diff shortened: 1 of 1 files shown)


ChrisKitching:
But this doesn't solve the problem with GetBoundedProp.

GetBoundedProp is a magical native-code machine that garbage collects props for you. It does this by destroying the underlying native objects. There is no sane way it can reach into lua and kill off all your references to the killed object (that would be... highly nontrivial), so you may keep references to exterminated objects.
This means, whenever you try and use a prop that is subject to the whims of GetBoundedProp, it might have been Destroy()'d already, and hence not have an actual native-code object backing it. This causes explosions.

Our options are, therefore, as follows:
- Hope the above educated guess is correct, ensure every time we touch wreckage with check IsDestroyed() first.
- Don't do any wreck collection at all (this will probably be surprisingly okay...)
- Implement a constant-time low-overhead garbage collector. A set of mass-tiered dequeues constructed via pointers sellotaped to the existing table nodes seems like a good way to go.

Sheeo:
I propose we don't do GC'ing at all for 3641 and take this up for consideration in a follow up patch after we've seen the effects on real world usage.

Considering historically that not all wrecks were under the native GC, this shouldn't be too bad.

Crotalus:
Yup, the only thing this really solves is to delete oldest wreck of same unit type first... for instance when you have hundreds of asf on the seabottom it will remove the oldest first when there's another asf-battle.

Yeah, it would be interesting to disable AddBoundedProp completely and leave wrecks un-GCed. Maybe it reduces some of the unexplained sim-crashes + getting rid of the pose-errors in log.

PR #629, opened 2015-05-28
Bugfix/capture

### LINKED ISSUE #617: Gifting units cancels ongoing captures
When gifting a unit that is being captured, the capture is cancelled. Original report [here](http://forums.faforever.com/forums/viewtopic.php?f=3&p=100268#p100268)

Crotalus:
Disable give on units being captured

Also fixed a bug with progress bars of capturing engineers not being in
sync if they not began at same time.

Fixes #617

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,8 @@ Exploits
 
 - Instant crash exploit partially fixed
 
+- No longer possible to give away units in progress of being captured
+
 
 Bugs
 ----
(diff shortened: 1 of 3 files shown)


aeoncleanse:
The capture sync is not working. Yes, the bars sync up, but an engie who joined the party later continues on with a full bar once the first stops. The unit is only capped after all the engies finish, which only happens when the final one to join has gone through the whole motion. Huge waste of energy. It's also laggy for some reason, the bars take a juddery second or so to turn on.

Crotalus:
Good catch, it didn't allow itself to give the unit until all had capture it, try now

aeoncleanse:
This is now working when I check the branch out directly, so I'm giving it the :+1:

PR #639, opened 2015-05-30
Search attackers for the largest damage radius

### LINKED ISSUE #627: Crash when selecting multiple sized nukes
In release/3641, Chris found that when you select multiple nukes with different damage radiuses, the game crashes thanks to the reticle scaling. I think this will also apply to T3 Arty, TMLs etc thanks to the way Crotalus expanded the idea to all of those.

aeoncleanse:
Fixes #627

### DIFF
--- a/lua/ui/controls/worldview.lua
+++ b/lua/ui/controls/worldview.lua
@@ -38,21 +38,38 @@ end
 -- Gets the weapons of attackers, nil if not same units
 local function GetAttackerWeapons(forceReticle)
     local attackers = GetSelectedUnits()
-    local bp = nil
+    local returnWeapons = nil
 
     if attackers then
+        local doneIds = {}
+        local totalMaxRadius = 0
         for i, u in attackers do
             if forceReticle or EntityCategoryContains(categories.SHOWATTACKRETICLE, u) then
-                if not bp then
-                    bp = u:GetBlueprint()
-                elseif bp.BlueprintId ~= u:GetBlueprint().BlueprintId then
-                    return nil
+                local bp = u:GetBlueprint()
+                local Id = bp.BlueprintId
+                if not doneIds.Id then
+                    table.insert(doneIds, Id)
+
+                    local Weapons = bp.Weapon
+                    local MaxRadius = 0
+                    for k, v in Weapons do
+                        if v.DamageRadius then
+                            MaxRadius = math.max(v.DamageRadius, MaxRadius)
+                        end
+                        if v.NukeOuterRingRadius then
+                            MaxRadius = math.max(v.NukeOuterRingRadius, MaxRadius)
+                        end
+                    end
+                    if MaxRadius > totalMaxRadius then
+                        totalMaxRadius = MaxRadius
+                        returnWeapons = Weapons
+                    end
                 end
             end
         end
     end
-
-    return bp.Weapon
+    
+    return returnWeapons
 end
 
 local function NukeDecalFunc()


lua/ui/controls/worldview.lua
```diff
@@ -38,21 +38,38 @@ end
...
         for i, u in attackers do
             if forceReticle or EntityCategoryContains(categories.SHOWATTACKRETICLE, u) then
-                if not bp then
-                    bp = u:GetBlueprint()
-                elseif bp.BlueprintId ~= u:GetBlueprint().BlueprintId then
-                    return nil
+                local bp = u:GetBlueprint()
+                local Id = bp.BlueprintId
+                if not doneIds.Id then
+                    table.insert(doneIds, Id)
```
Crotalus: This should be `doneIds[Id] = true`

lua/ui/controls/worldview.lua
```diff
@@ -38,21 +38,38 @@ end
...
+        local totalMaxRadius = 0
         for i, u in attackers do
             if forceReticle or EntityCategoryContains(categories.SHOWATTACKRETICLE, u) then
-                if not bp then
-                    bp = u:GetBlueprint()
-                elseif bp.BlueprintId ~= u:GetBlueprint().BlueprintId then
-                    return nil
+                local bp = u:GetBlueprint()
+                local Id = bp.BlueprintId
+                if not doneIds.Id then
```
Crotalus: if not doneIds[Id]

lua/ui/controls/worldview.lua
```diff
@@ -38,21 +38,38 @@ end
...
+
+                    local Weapons = bp.Weapon
+                    local MaxRadius = 0
+                    for k, v in Weapons do
+                        if v.DamageRadius then
+                            MaxRadius = math.max(v.DamageRadius, MaxRadius)
+                        end
+                        if v.NukeOuterRingRadius then
+                            MaxRadius = math.max(v.NukeOuterRingRadius, MaxRadius)
+                        end
```
Crotalus: Could replace the if statements with this:
`MaxRadius = math.max(v.NukeOtherRingRadius or v.DamageRadius or 0, MaxRadius)`

lua/ui/controls/worldview.lua
aeoncleanse: Surely it would be == true, and does checking the existence of a table value work as bool?

lua/ui/controls/worldview.lua
aeoncleanse: Would this take account of such things as a weapon which has both of the values, which might happen in a mod?

Crotalus: True, if a unit has both a nuke weapon and a normal weapon with decal and the nuke weapon is less range my suggestion above will return wrong value.

This would probably be better:
`MaxRadius = math.max(v.NukeOtherRingRadius or 0, math.max(v.DamageRadius or 0, MaxRadius)`

But now it's starting to look ugly :)

lua/ui/controls/worldview.lua
Crotalus: Yeah but your check `if not doneIds.Id` checks the doneIds-table for a key named 'Id' (same as if you would write `doneIds['Id']`, not the desired behavior you want. 

By using the blueprint id as key and setting it to true you can check the table if the key exists with the suggested `doneIds[Id]`

If you just insert the blueprint ID's into the table you need to use table.find(doneIds, Id)` but this is more expensive.

ChrisKitching: IceDreamer, you're perhaps thinking of tables as lists, when their semantics are really closer to that of hashmaps.

A table maps "keys" to "values". Both keys and values can be any Lua value.
When we do someTable[x], we are accessing the value in the table that is associated with the key equal to the current value of x. Not to be confused with someTable["x"], which looks for something keyed by the string, "x".

Assignment works in a similar way, just using the table expression as an lvalue.

Table.insert finds the next integer key that is free in a table and associates the value with it.
An implementation detail of tables is that, when integer keys are present, they may be backed by an array instead of a map. A map may also be present for some of the keys.

So, when you want a list, you're really just using integer keys in a table. The {a,b,c} notation for table literals is a handy shorthand for this,but is equivalent to {1:a, 2:b, 3:c}.

lua/ui/controls/worldview.lua
ChrisKitching: You never need to handle more than one decal type at a time, though, because you only ever have one command mode.

So, use a predicate to filter the set of selected weapons and then find the maximum radii you care about from that. (I've also implemented this. Not clear why ice is doing work he knows I'm in the process of doing: there's plenty to go around, after all)

Issue #647, opened 2015-06-07
Refactor inheritance chain of construction units


Sheeo:
Spinoff from [this discussion](https://github.com/FAForever/fa/commit/a6948c9#commitcomment-11556014).

We need the inheritance chain cleaned up for construction units. Exmples of bad inheritance:
- CommandUnit doesn't inherit from ConstructionUnit
- StructureConstructionUnit doesn't inherit from ConstructionUnit

Moreover, many classes that derive directly from their base are defined in `defaultunits.lua`, making it difficult to hook a base class without also redefining all derived classes. We should look into moving these definitions into seperate files, making it easier to mod.

MrRowey:
@Garanas would this come under the Optimisation that you have been working on

Garanas:
This would be code style. It is very relevant, but also not worth the effort for now.

Garanas:
This can break mods that expect the current hierarchy, closing as not planned.

PR #649, opened 2015-06-08
Split up defaultunits.lua

ChrisKitching:
Sheeo mentioned over in #647 how nice it would be if the classes were split into separate files, so I did it.

This enables the "factory fixes" rubbish to go away (merged into the classes), and makes things really surprisingly neat.

Similar work could happen to aeonunits/cybranunits/seraphimunits/terranunits.lua.


ChrisKitching:
(I spawned one of every unit to verify that this didn't end the world. Any missing imports will be pretty rapidly picked up as we test, methinks. Would be nice to see this shipped for the improved modding API (we've already broken a bunch of mods for 3641: better to do it all at once methinks and make the fixing maximally pleasant, than to keep on doing it a bit at a time? ))

Sheeo:
Looks good.

It will break all mods though; I'm not sure the other changes we did in 3641 will break quite as many.

A migratory solution could be to keep references to the classes in defaultunits.lua, and keep using that file from the units — but import the bases from there, still giving the advantageous file structure but somewhat preserving compatibility.

aeoncleanse:
Sounds like a good solution Sheeo

Crotalus:
Yeah, keep references in defaultunits.lua and emit a warning about them being deprecated and due to be removed in a future release.

Sheeo:
Please rebase to target develop :)

PR #650, opened 2015-06-08
Fix some more problematic collision instances

aeoncleanse:
This is a refined, focused version of my work from last year. It contains the most essential pure hitbox/targetbone adjustments, as well as a hack to fix the fact that 90% of ships cannot target some Sonar/Torpedo Launchers. This hack should be replaced by the addition of new bones at a model level, as soon as we find someone with the know-how to pull it off.

Fine tuning complete

Some of the import paths will need changing alongside @ChrisKitching 's defaultunits.lua split


aeoncleanse:
Ready to roll, the Torp/Sonar boxes etc, which are the bigger changes, have been fine-tuned.

aeoncleanse:
@Sheeo - Absolute essentials are Megalith, Cooper, Stealth Boat, Salem, and the hovertanks. Ythotha should probably also go in because it can do the whole kiting lasers things like SCUs.

ChrisKitching:
This should probably be rebased considering the model changes.

aeoncleanse:
It has been for many of them. I'll take another look this evening.

Sheeo:
Needs rebase targeting develop

PR #653, opened 2015-06-09
Tell ACU shield not to go through unused OnDamage

aeoncleanse:
Fixes personal shield ACUs not playing the damage warning. Damage is passed from unit.lua's OnDamage() directly to ApplyDamage for personal shields. shield.lua's OnDamage is now unused in most cases.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1905,14 +1905,11 @@ ACUUnit = Class(CommandUnit) {
         local aiBrain = self:GetAIBrain()
 
         -- Mutate the OnDamage function for this one very special shield.
-        local oldOnDamage = self.MyShield.OnDamage
-        local newOnDamage = function(shield, instigator, amount, vector, dmgType)
-            oldOnDamage(shield, instigator, amount, vector, dmgType)
-
+        local oldApplyDamage = self.MyShield.ApplyDamage
+        self.MyShield.ApplyDamage = function(...)
+            oldApplyDamage(unpack(arg))
             aiBrain:OnPlayCommanderUnderAttackVO()
         end
-
-        self.MyShield.OnDamage = newOnDamage
     end,
 
     DoTakeDamage = function(self, instigator, amount, vector, damageType)


lua/defaultunits.lua
```diff
@@ -1905,14 +1905,14 @@ ACUUnit = Class(CommandUnit) {
...
         -- Mutate the OnDamage function for this one very special shield.
-        local oldOnDamage = self.MyShield.OnDamage
-        local newOnDamage = function(shield, instigator, amount, vector, dmgType)
-            oldOnDamage(shield, instigator, amount, vector, dmgType)
-
+        local oldApplyDamage = self.MyShield.ApplyDamage
+        local newApplyDamage = function(shield, instigator, amount, vector, dmgType, doOverspill)
+            oldApplyDamage(shield, instigator, amount, vector, dmgType, doOverspill)
+            
             aiBrain:OnPlayCommanderUnderAttackVO()
```
Sheeo: This is what plays the warning, it gets called in both versions. Surely this change doesn't affect that?

I'm unclear as to why you want to delegate to a somewhat private function here.

aeoncleanse:
Chris built this so that, for ACUs, when they make the shield, the audio command is added to that unit's shield script. So far so obvious.

The problem arises where personal shield ACUs did not play the warning because of the way personal shields work now. For a bubble, incoming damage hits the shield, which calls shield.lua OnDamage, which triggers the audio. For personal shields, incoming damage hits the unit base, calling unit.lua OnDamage, which then passes the damage directly into shield's ApplyDamage. 

Since unit.lua doesn't direct to OnDamage, the warning didn't play. I chose to do this instead of changing unit.lua to pass things to OnDamage because I assumed that shield.lua OnDamage is effectively empty for a good reason. Didn't we make it empty and put everything in ApplyDamage ourselves when we refactored shields?? What was the reasoning then?

Sheeo:
Protip: Reply in the line note so the discussion is tied to the code in question :)

Gotcha, mutating `ApplyDamage` is likely the right thing to do.

On a style note, I would prefer:

    self.MyShield.ApplyDamage = function(...)
      oldApplyDmg(unpack(arg))
    end

PR #657, opened 2015-06-09
Stop Engineering station drones posting an error when taking off/landing

### LINKED ISSUE #656: Engineering Station drones cause error release/3641
WARNING: Error running OnTransportDetach script in Entity xeb0204 at 206a4f08: e:\github\fa\lua\terranunits.lua(534): attempt to call field `OnTransportDetach' (a nil value)
         stack traceback:
         	e:\github\fa\lua\terranunits.lua(534): in function <e:\github\fa\lua\terranunits.lua:532>

On exit

WARNING: Error running OnTransportAttach script in Entity xeb0204 at 2069a008: e:\github\fa\lua\terranunits.lua(529): attempt to call field `OnTransportAttach' (a nil value)
         stack traceback:
         	e:\github\fa\lua\terranunits.lua(529): in function <e:\github\fa\lua\terranunits.lua:527>

On landing

aeoncleanse:
Fixes #656

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1612,10 +1612,7 @@ AirTransport = Class(AirUnit) {
             unit:DisableShield()
             unit:DisableDefaultToggleCaps()
         end
-        if not EntityCategoryContains(categories.PODSTAGINGPLATFORM, self) then
-            self:RequestRefreshUI()
-        end
-        --Added by brute51
+        self:RequestRefreshUI()
         unit:OnAttachedToTransport(self, attachBone)
     end,
 
@@ -1624,9 +1621,7 @@ AirTransport = Class(AirUnit) {
         self:MarkWeaponsOnTransport(unit, false)
         unit:EnableShield()
         unit:EnableDefaultToggleCaps()
-        if not EntityCategoryContains(categories.PODSTAGINGPLATFORM, self) then
-            self:RequestRefreshUI()
-        end
+        self:RequestRefreshUI()
         unit:TransportAnimation(-1)
         unit:OnDetachedToTransport(self)
     end,
--- a/lua/terranunits.lua
+++ b/lua/terranunits.lua
@@ -526,12 +526,14 @@ TPodTowerUnit = Class(TStructureUnit) {
     
     OnTransportAttach = function(self, bone, attachee)
         attachee:SetDoNotTarget(true)
-        TStructureUnit.OnTransportAttach(self, bone, attachee)
+        self:PlayUnitSound('Load')
+        attachee:OnAttachedToTransport(self, attachBone)
     end,
     
     OnTransportDetach = function(self, bone, attachee)
         attachee:SetDoNotTarget(false)
-        TStructureUnit.OnTransportDetach(self, bone, attachee)
+        self:PlayUnitSound('Load')
+        attachee:OnDetachedToTransport(self, attachBone)
     end,
     
     FinishedBeingBuilt = State {


lua/terranunits.lua
```diff
@@ -526,12 +528,12 @@ TPodTowerUnit = Class(TStructureUnit) {
     
     OnTransportAttach = function(self, bone, attachee)
         attachee:SetDoNotTarget(true)
-        TStructureUnit.OnTransportAttach(self, bone, attachee)
+        AirTransport.OnTransportAttach(self, bone, attachee)
```
ChrisKitching: You're delegating to a class you're not inheriting from.

Don't!

aeoncleanse: Proper way to do it?

I can't just inherit, that does a bunch of things I don't want. Is there really a problem with nicking a foreign function in a unique edge-case like this?

Sheeo: It violates common OOP design guidelines for sure, and stating in a comment that we inherit it when we don't is pretty misleading.

What stuff from the AirTransport class does the kennel drone need here?

I don't see anything from the `AirTransport` definitions that we need to do here -- it contains an exception seemingly for this unit though?

```
if not EntityCategoryContains(categories.PODSTAGINGPLATFORM, self) then
    self:RequestRefreshUI()
end```

Comment (aeoncleanse) [inline lua/terranunits.lua]: It contains the attach/detach functions because Chris moved them out of unit.lua. Figured i wouldn't put the detach one in yet because I was bound to have got something a bit warped

Comment (Sheeo) [inline lua/terranunits.lua]: Yes but what does the kennel drone station need to do here that's done in AirTransport? Seemingly nothing?

Comment (aeoncleanse) [inline lua/terranunits.lua]: The kennel is a transport... It needs to attach/detach the drones...

Comment (Sheeo) [inline lua/terranunits.lua]: None of that stuff is done in `AirTransport.OnTransportAttach`:

```
        self:PlayUnitSound('Load')
        self:MarkWeaponsOnTransport(unit, true)
        if unit:ShieldIsOn() then
            unit:DisableShield()
            unit:DisableDefaultToggleCaps()
        end
        if not EntityCategoryContains(categories.PODSTAGINGPLATFORM, self) then
            self:RequestRefreshUI()
        end
        --Added by brute51
        unit:OnAttachedToTransport(self, attachBone)
```

This is a _callback_ called by the engine when the drones have been attached/detached, so presumably it needs to play a sound, that's it. It does not:
 - Need to mark drone weapons as being on transport
 - Need to disable any shielding
 - Call the extraneous callback

Comment (aeoncleanse) [inline lua/terranunits.lua]: In this case, the solution is to forgo the callback entirely, and just play the sound directly and set donottarget?

Comment (Sheeo) [inline lua/terranunits.lua]: Yes unless a base implementation further up the inheritance chain happens to define this behavior still, in which case just remove the override.

Comment (ChrisKitching) [inline lua/terranunits.lua]: And don't forget to remove the kennel-specific code sheeo pointed out (as well as any other such code) from AirTransport.

File: lua/defaultunits.lua [hunk 2]
```diff
@@ -1612,10 +1612,6 @@ AirTransport = Class(AirUnit) {
             unit:DisableShield()
             unit:DisableDefaultToggleCaps()
         end
-        if not EntityCategoryContains(categories.PODSTAGINGPLATFORM, self) then
-            self:RequestRefreshUI()
```
Sheeo: Now this doesn't get called at all.

What we should do here is just to remove the conditional check for the `PODSTAGINGPLATFORM` category, and always Request UI refresh.

aeoncleanse: In the pod right? Yeah I think I just missed the refresh

aeoncleanse: Right, this should be better now

Sheeo: The diff is the same

aeoncleanse: On defaultunits yeah, I think it should be? The stuff relating to pods has been removed from defaultunits, and the stuff pods care about is now in terranunits where it belongs.

Sheeo: I'll make it more clear:

Before this change, `AirTransport.OnTransportDetach()` calls `RequestRefreshUI()` _unless_ the 'transport' has the `PODSTAGINGPLATFORM` entity category.

Now it never calls `RequestRefreshUI()`, and the replacement function in `TPodTowerUnit` _always_ calls it (It should never).

ChrisKitching:
Merged.

PR #659, opened 2015-06-13
Support factories templates for AI

speed2CZ:
Added templates for all support factories. AI in coop missions is now able to rebuild support factories.


aeoncleanse:
:+1:

PR #663, opened 2015-06-14
Allow Airborne AA to hit water landed transports

### LINKED ISSUE #662: Interceptors can't fire at transports landed on water
Is this intended behaviour?

If so then should interceptors be able to fire at transports landed on land?

aeoncleanse:
Fixes #662 

I see no logical reason they should be able to target landed gunships, bombers, transports which are on land, but not transports landed on water. I think this is simply an oversight by GPG, who forgot that Transports could land on water (They have to in order to be able to pick things up from water, I think), and so restricted AA to only Land/Air (The only two layers anything but transports can find themselves)

This also has a commit which changes the targeting restrictions of the Cybran T2 Transport EMP weapon, for reasons outlined in the commit.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -32,6 +32,7 @@ Physics
 - Yolonna Oss now deals damage to itself
 - Flares and depth-chages no longer draw in friendly projectiles
 - Fix shots sometimes going right through Walled PD
+- Airborne AA weapons can now hit Transports which have landed on water
 
 Other
 - T3 Torpedo Bomber can use Attack-Move properly again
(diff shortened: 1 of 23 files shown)


Sheeo:
Please retarget this for 3642.

aeoncleanse:
There is no 3642 yet?

Also, this is *hardly* a balance change dude... I mean come on, who's really going to complain about this? It's flat out an oversight.

Sheeo:
"Balance" doesn't mean "will people complain about this".

More that this change is wholly motivated by gameplay and only directly affects it, there are no inherent/obvious "bugs" here.

It could very well be intentional and you could very well argue that we should do the reverse -- disallow hitting landed transports generally.

It doesn't fit into a 'bugfix' patch for this reason.

HardlySoftly:
This is definitely balance change imo

aeoncleanse:
I would very much argue that innocent behavioural oversight by the developers qualifies firmly as a bug. And no, I can't prove it, but I don't believe that anyone here believes it was anything other than that, not really.

In this case though, I don't think it's worth saying much more than: Sheeo - Onto what branch should I target this???

Sheeo:
`develop` targets  3642.

PR #666, opened 2015-06-14
Allow Airborne AA to hit water landed transports

### LINKED ISSUE #662: Interceptors can't fire at transports landed on water

aeoncleanse:
Fixes #662

I see no logical reason they should be able to target landed gunships, bombers, transports which are on land, but not transports landed on water. I think this is simply an oversight by GPG, who forgot that Transports could land on water (They have to in order to be able to pick things up from water, I think), and so restricted AA to only Land/Air (The only two layers anything but transports can find themselves)

This also has a commit which changes the targeting restrictions of the Cybran T2 Transport EMP weapon, for reasons outlined in the commit.

### DIFF
--- a/mod_info.lua
+++ b/mod_info.lua
@@ -3,7 +3,7 @@
 -- Documentation for the extended FAF mod_info.lua format can be found here:
 -- https://github.com/FAForever/fa/wiki/mod_info.lua-documentation
 name = "Forged Alliance Forever"
-version = 3641
+version = 3642
 copyright = "Forged Alliance Forever Community"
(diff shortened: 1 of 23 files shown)


Sheeo:
This development history is interesting.

aeoncleanse:
Agreed... wtf?!

Sheeo:
Seems like your local develop doesn't point at the same one on this repo.

Crotalus:
baf14cf2fe0b9f8e7a9613693b4ed2ca5aebe2f8 .. 329cb53cd668c7d844ab75e3357a4c0438fa23bb

PR #670, opened 2015-06-19
Fix to make legacy checkbox code work

### LINKED ISSUE #669: Team overflow togglebutton UI issue
[Original bug report](http://forums.faforever.com/forums/viewtopic.php?f=3&p=101598#p101598).

Potentially a regression with changes in maui. Needs verification.

Crotalus:
This should be seen as a quickfix to getting legacy checkbox code to work
(diplomacy / debug window etc).

We should probably decide on a naming convention for the UI-textures, at
the moment the modded FAF ones differ from original.

Fixes #669


ChrisKitching:
> We should probably decide on a naming convention for the UI-textures, at
> the moment the modded FAF ones differ from original.

I did: I selected the least insane of the four-or-so GPG was using and started enforcing it with an iron fist. :P

PR #674, opened 2015-06-19
Cleanup seraphim builders

Crotalus:
Seraphim builders had their own OnStartBuild, probably to override the tarmac behavior. The problem is that Unit.OnStartBuild has important checks for rebuild bonus etc. This wasn't activated on seraphim engineers.

Now that tarmacs are spawned on the actual built unit instead this method can safely be removed


Crotalus:
ACU / SCU had the same problem, fixed it also

ChrisKitching:
Merged with 3dde69cb427774890e5e1c81a7c5158c490f23ae.

PR #675, opened 2015-06-19
Fixed regression with callback refactor

Crotalus:
Regression caused some errors when running with AI-players, making them idle at start.

* Fixed bug with wrong argument order to AddUnitCallback (GPG has
 non-logical order to begin with)

* AddOnUnitBuiltCallback / DoOnUnitBuiltCallback needs special handling
 due to the extra `category` argument


Crotalus:
a2d2503633896026130cb80eb01856fde31225c4

PR #677, opened 2015-06-19
Kills dead vet code

aeoncleanse:
These functions have been completely overridden by FAF's vet system and are now defunct that I can see.


ChrisKitching:
Not too useful to add a hook just to kill a fairly small function: I've merged the other commit that's useful from here.

PR #679, opened 2015-06-21
Bugfix/disappearing wreck

### LINKED ISSUE #623: You can instantly destroy reclaim of buildings from across the map.
Only works with same faction, since you need to be able to build the building which made the wreck.

Replay 3474855.
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=3474855

You *used* to be able to use this to rebuild buildings at 1/2. It was made so that this didn't work, but as you can see, effects still remain.

This also was part of a bug that let you rebuild a line of buildings from 1/2 from dragging from a wreck - that was also fixed.

To clarify, no reclaim mass is received.

Crotalus:
This is not the most beautiful solution to #623 but I can't come up with a better either. Input welcome!

So:

If a builder is about to rebuild something (decided by a call to GetRebuildBonus) we
check in SetRebuildProgress if the actual bonus never was triggered. This
means that there's a risk for a disappearing wreck.

We then search for the potential wrecks on the map, iff one of these wrecks
are getting destroyed right after (same tick) we just recreate it

This scan for wrecks will happen very, very seldom, thus should not impact the
overall performance of the sim

There is a minuscule risk that it recreates a wreck that disappeared for some other reason; garbage collected or that it was reclaimed the same tick - but shouldn't be a problem. The clone function sets the mass value of the wreck to the current reclaim progress so it would be worthless anyway. 

Fixes #623

### DIFF
--- a/lua/sim/Prop.lua
+++ b/lua/sim/Prop.lua
@@ -13,7 +13,7 @@
 local Entity = import('/lua/sim/Entity.lua').Entity
 local EffectUtil = import('/lua/EffectUtilities.lua')
 
-RECLAIMLABEL_MIN_MASS = 20
+local RECLAIMLABEL_MIN_MASS = import('/lua/sim/ScenarioUtilities.lua').RECLAIMLABEL_MIN_MASS
 
 Prop = Class(moho.prop_methods, Entity) {
 
--- a/lua/sim/Prop.lua
+++ b/lua/sim/Prop.lua
@@ -97,9 +97,27 @@ Prop = Class(moho.prop_methods, Entity) {
         Entity.Destroy(self)
     end,
 
+    CloneWreck = function(self)
+        local propid = self:GetBlueprint().BlueprintId
+        local pos = self.CachePosition
+        local ori = self:GetOrientation()
+        local unitbp = __blueprints[self.AssociatedBP]
+        local max = self.MaxMassReclaim / (unitbp.Economy.BuildCostMass * (unitbp.Wreckage.MassMult or 0))
+        local reclaimed = 1 - (self.MassReclaim / self.MaxMassReclaim) * self:GetFractionComplete()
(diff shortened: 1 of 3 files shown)


ChrisKitching:
Until a couple of hours ago I was working on a very similar approach. I found that the wheels fell off because the wreck we wanted to clone had already been destroyed by the time OnStartBuild gets called, causing all its fields to have been nulled out.

Crotalus:
Yeah, noticed that GetPosition() returns 0, 0, 0 but self.CachePosition worked just as well. 

It seems to be working, haven't tested too thoroughly so the wheels may still roll off I guess

ChrisKitching:
But yes, at a high level this approach is probably the neatest, as other ways seem to involve large datastructures tracking wrecks, and as you say we really only run the hideously expensive thing in cases where the bug would've been reproduced before.

Unfortunately, this does lead to some odditites: engineers reclaiming a wreck that gets cloned will stop, and if you move the build template onto another wreck of the right sort it'll still delete the original.

ChrisKitching:
> Yeah, noticed that GetPosition() returns 0, 0, 0 but self.CachePosition worked just as well.

I just found the reason mine wasn't working: I was calling wreck.Clone instead of wreck:Clone, so of course all the fields were nil.
Now it all makes sense: we expect the C object to be deleted, but we can reconstruct what we need from the stuff in lua-land. (GetPosition(), for example, is a native-code function that doesn't work usefully after the underlying object has been destroyed).

I'll combine the best parts of both approaches and land it. I hadn't noticed CreatePropHPR: that seems usefull...

ChrisKitching:
A few bits of this ended up in b5ab7b4d7527e4041276b6ddcae60880202618e3. Seems we worked in parallel somewhat.
Would be interesting to hear your thoughts on a few things (notably the necessity of tagging wrecks with the tick number)

PR #700, opened 2015-06-27
Allow Satellite to be rebuilt

aeoncleanse:
Need help getting parts of this to work, particularly some UI mess top-right and getting the animation to reverse.


aeoncleanse:
Build UI is messed up for some reason, visually and it doesn't start on the tech 4 tab. Animations aren't right for reversing. Game hangs when building a new sat is completed for some reason.

aeoncleanse:
Right, this is now working apart from the UI being a bit messed up. The new approach is a bit... ugly... But it works far, far better and more smoothly than anything else I tried, and it's such a niche case... Remember, Satellite rebuild is only possible when you've blocked a Nuke with the Satellite.

Sheeo:
This needs to be rebased to target `develop`, thanks!

PR #706, opened 2015-06-28
Restrict number of hive buildbots #704

### LINKED ISSUE #704: 3641 Test: please clamp the number of floating cybran engineer bots
Due to the number of bots scaling with the build speed, if you have a mod that increases the build speed, the number of bots quickly add up and rape your FPS. There are several mods that increase build speed, and I don't think any of them are going to be fixed by their authors any time soon, if ever. This is such a simple fix that I think you guy could easily fix it on your end before it even becomes a problem.

duk3luk3:
Hive buildbot number scales with buildpower. This creates huge spam and
performance problems if the build power is modded above the standard
values. This commit clamps the number to 10. Fixed #704

### DIFF
--- a/lua/EffectUtilities.lua
+++ b/lua/EffectUtilities.lua
@@ -416,6 +416,11 @@ end
 
 function SpawnBuildBots( builder, unitBeingBuilt, BuildEffectsBag)
     local numBots = math.ceil((10+builder:GetBuildRate()) / 15)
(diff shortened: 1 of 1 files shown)


Sheeo:
Thanks for the contribution :) I took the liberty of simplifying and adding a comment for this, hope you don't mind. :+1:

PR #707, opened 2015-06-29
ACU/SCU tac missile upgrade rebalance

Varaxis:
Altered the trajectory of missiles to not fly as high and the build time of missiles changed to 840, from 300 for ACU and 1800 with SCU. Currently takes 7 seconds to make a missile with T2 suite unassisted--can get many missiles off before victim can move an engineer into a good position to start TMD, and since the missiles fly so high, they were difficult for Aeon to defend against.

Change forces the players to have multiple engineers assisting to get similar or greater effect, affecting the viability of being dropped alone in the sea to get a better angle around TMD.

Missiles still have 3 HP, with no change to the trajectory of missiles when fired at targets near the generously low minimum range.

### DIFF
--- a/projectiles/SIFLaanseTacticalMissileCDR/SIFLaanseTacticalMissileCDR_proj.bp
+++ b/projectiles/SIFLaanseTacticalMissileCDR/SIFLaanseTacticalMissileCDR_proj.bp
@@ -1,0 +1,65 @@
+ProjectileBlueprint {
+    Audio = {
+        Impact = Sound {
+            Bank = 'Impacts',
+            Cue = 'XSS_Missile_Impact',
+            LodCutoff = 'Weapon_LodCutoff',
+        },
+        ImpactTerrain = Sound {
+            Bank = 'Impacts',
+            Cue = 'XSS_Missile_Impact',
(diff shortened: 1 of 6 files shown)


aeoncleanse:
IMO this is un-needed. By and large, the community has requested a very simple set of changes.

1 - Make the Seraphim's minimum effective range the same as the UEF's (Simple, done by increasing MuzzleVelocity from 5 to 10)
2 - Reducing Seraphim's AOE to 2, same as UEF's
3 - Changing trajectory to ensure that the missile can be properly countered

By and large, the UEF's missile upgrade is fair, aside from against UEF and Aeon TMD, where it flies over.

Sheeo:
Hi @Varaxis, thanks for taking the time to look into this.

Until we have a balance team we can't really merge pull requests such as this. This is happening shortly, however, so it may be worth leaving this PR open for balance review soon.

Sheeo:
Perhaps @Zockyzock or @ppipp would take a look at this now?

ppipp:
I will test these changes. They seem in line with things we discussed, with a few others.

Zockyzock:
taken care of in 3644, closing this

PR #714, opened 2015-06-30
Make sure originalBuilder is reset

### LINKED ISSUE #697: Engineers assist loading SMD at cost of building
Original forum report [here](http://forums.faforever.com/forums/viewtopic.php?f=3&t=9578#p95893)

Crotalus:
Fixes #697

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1782,7 +1782,9 @@ Unit = Class(moho.unit_methods) {
             end
         end
 
-        self.originalBuilder = builder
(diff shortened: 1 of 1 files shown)


ChrisKitching:
While this appears to work, I find the fact that this works _deeply_ troubling.

Instead of (well, perhaps *as well as*), it might be worth a more holistic unfucking of the consumption system. That shit was pretty tangled when GPG finished it, and `n` years of FAF's loving embrace has definitely not improved it ;)

Issue #705, opened 2015-06-28
3641 Test: if you build a lot of things that take a short time to finish, the cybran engineer bots go EVERYWHERE


Divran:
![untitled](https://cloud.githubusercontent.com/assets/1485728/8397690/a29aec5c-1dd4-11e5-89f1-3bf6477b8cbb.png)

duk3luk3:
Looks like a feature to me ;-)

Divran:
inb4 the enemy finds your cloaked and stealthed base due to the cybran bots flying across the map

ChrisKitching:
The proper fix here may be to just make them turn more sharply

aeoncleanse:
Reduce speed, increase turnrate.

ChrisKitching:
Surely we want the speed to be the same but the acceleration and turning rate to be higher. This problem is caused, I think, by them overshooting the target and having to loop round and come back (like a bomber might behave)

aeoncleanse:
Test dropping speed and upping turnrate first,  simple and it should do the trick

ChrisKitching:
Tweaking the speeds doesn't really fix the problem. The weird behavious is caused by the drones effectively being repeatedly given move orders and having them cancelled in successive ticks.

By adding a check to dampen this behaviour, the drones behave less absurdly from 9332703.

ChrisKitching:
... Gonna call that "done". This is definitely an edgcease anyway: builds that complete in one simtick aren't really a thing.

Divran:
They are when I play (even without mods)

PR #716, opened 2015-06-30
Slow down Cybran drones

aeoncleanse:
#705


ChrisKitching:
But there's more to it than that: If the construction completes within one simulation timestep the drones get confused anyway because they never have a chance to act upon their orders.

This may well be why GPG made them always dock between jobs.

aeoncleanse:
Closed in light of Chris' alternate fix

Issue #726, opened 2015-07-04
3641 Test: Possible regression with units sinking.


bamboofats:
WIth the patch as of 04/07/15, several naval units do not sink after being destroyed. The model just get stuck there, consequently do not get turned into wreakage as they never hit the bottom.
If the unit is surfaced, its stuck at the top, if dived, is stuck in the dive depth.
I tried with all sea units, and as many floating land units as possible, but the list includes

UEF T3 SONAR  (the other cybran aeon sinks ok)
All 4 factions T1 submarines
Submarine hunters: Cybran Aeon Sera
All 3 strategic T3 subs.
Basically all navals that can dive except sera destroyer and experimentals.

Reproducible?

ChrisKitching:
Reproduced. Probably a regression from when we reintroduced the seabedwatcher logic.

ChrisKitching:
Fixed in 35f5612

PR #728, opened 2015-07-04
Fix sinking regressions

aeoncleanse:
#726


aeoncleanse:
Greeee, stupid develop!

PR #729, opened 2015-07-04
Fix sinking regressions

aeoncleanse:
This causes naval units with no animation to sink using the sinker
mechanism. Also causes units WITH an animation to have their wreck
spawned in the right place.

#726


ChrisKitching:
Merged with amendements: 35f5612

PR #732, opened 2015-07-07
MEXs and T1 PD on top

CodingSquirrel:
Increase the icon priority of MEXs and T1 PD so that they appear on top of other buildings. The intention is to make sure MEX show above mass storage and T1 PD show above walls.


CodingSquirrel:
Maybe I'm just missing the proper interaction here, but it seems to me that StrategicIconSortPriority is all but completely ignored for the most part. Even when the StrategicIconSortPriority was significantly lower for the PD or MEX than the units surrounding it it would still show up below the other icons. However, for both of them using the priority of 64 made them jump above the surrounding units.

quark036:
I think it has to do with the last category of each unit: sortstrategic, sortconstruction, or sorteconomy. I'm not at home right now, but I'll do some testing when I get back and see if that affects it.

CodingSquirrel:
Just looking at the PD and wall sections, they both have the SORTDEFENSE category. We don't have the mass storage in the repo (at work as well and can't check the vanilla files) but I'd imagine the MEX and storage have the same category.

quark036:
Ah yeah, good point, I was thinking about them vs other buildings, not them vs eachother.

CodingSquirrel:
Added radar as well

ChrisKitching:
Merged at 246d22291dc4cd4437562367b6be0f8a99e0bd70

PR #733, opened 2015-07-07
Fix for Seraphim and UEF ACU attack circle

CodingSquirrel:
Fixed an issue where UEF and Seraphim ACU would incorrectly show a
radius circle for regular ground attacks and wouldn't show a circle for
tac missiles.

### DIFF
--- a/lua/ui/controls/worldview.lua
+++ b/lua/ui/controls/worldview.lua
@@ -49,7 +49,7 @@ local function GetSelectedWeaponsWithReticules(filterFunc)
             local bp = u:GetBlueprint()
             for k, v in bp.Weapon do
                 if filterFunc(v) then
-                    weapons[bp.BlueprintId] = v
+                    weapons[bp.BlueprintId .. v.Label] = v
                 end
             end
         end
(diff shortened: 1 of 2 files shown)


lua/ui/controls/worldview.lua
```diff
@@ -49,7 +49,7 @@ local function GetSelectedWeaponsWithReticules(filterFunc)
             local bp = u:GetBlueprint()
             for k, v in bp.Weapon do
                 if filterFunc(v) then
-                    weapons[bp.BlueprintId] = v
+                    weapons[v.Label] = v
```
CodingSquirrel: I only tested quickly with the ACU. Looking at the code and the history a little closer it looks like this should support multiple units as long as they're firing the same "type"? If so then perhaps "weapons[bp.BlueprintId .. v.Label]" would be more appropriate here.

CodingSquirrel:
Tested this time with a variety of units, including static t2 arty, tml, acu, acu tac missiles, and cruisers. Looks like it's all working properly with the minor change.

CodingSquirrel:
Merged the two changes into one.

ChrisKitching:
Useful parts merged at d18e64d823368eda6fcafe798c524ebc3778d7c2 (and successor).

As for the missile decals, they're a bit weird. We don't (I think?) support the existence of units with multiple  manual-fire missile weapons, so keeping more than one per unit isn't useful. We don't however, want to consider the nuclear missile when we shouldn't be doing (which was seemingly the cause of the funky problems there).

An early return is also possible here.

PR #737, opened 2015-07-09
Feature/targeting memory

Crotalus:
#406 rebased to `develop`

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -1,5 +1,5 @@
----- 
----- 
+----
+----
 ---- This module contains the Sim-side lua functions that can be invoked
 ---- from the user side.  These need to validate all arguments against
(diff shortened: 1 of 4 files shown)


Sheeo:
This should be interesting for @Zockyzock and @ppipp

Crotalus:
@ppipp Try now

Zockyzock:
Great change. But i don't like adding new UI Elements, making the game more complicated/overloaded. I think it has to work without. 

Is it possible to check if all units from your manual priority (e.g. mex with cruisers) in range are destroyed, and when that is the case, switch back to normal priority?

ppipp:
Now it seems to work well. In my opinion, this behaviour should have been in the stock game, so even if it's a very big nerf to ACU, I think it should be included because it's much more userfriendly and intuitive.

PR #160, opened 2014-12-08
Refactoring deathweapons

Crotalus:
Colossus

* An additional death animation
* Death weapon now using the generic codepath in sim/Unit.lua

On a side note, the way death weapons work at the moment is silly, if
you're staying 4.1 from the unit 0 damage, 4 - 7500 damage.

I will send in a gradient damage patch to make damage variable depending
on distance from centrum

PR #739, opened 2015-07-09
Refactoring deathweapons

Crotalus:
Rebased #160 

**Colossus**
   * An additional death animation
   * Death weapon now using the generic codepath in sim/Unit.lua

**Gradient death damage**
   * More damage near impact of the experimental

**Support for damage at impact bones**
   * The crush damage is dealt when the bones of a experimental hits something. The damage specified in the blueprint are the maximum damage dealt. It's divided among the different crush bones

Forum thread here:
http://forums.faforever.com/forums/viewtopic.php?f=2&t=9317


Zockyzock:
really cool, should be able to put this in 3645

PR #741, opened 2015-07-10
Fix UnitView display issues

CodingSquirrel:
Fix the unit detail display showing up uninitialized at startup.
Fix the unit detail display overlapping with building info display when rolling over the items in the construction panel.


lua/ui/game/gamemain.lua
```diff
@@ -540,7 +540,6 @@ function HideGameUI(state)
             statusClusterGroup:Show()
             import('/lua/ui/game/worldview.lua').Contract()
             import('/lua/ui/game/borders.lua').HideBorder(false)
-            import('/lua/ui/game/unitview.lua').Expand()
```
CodingSquirrel: We don't need to expand the unitview panel if nothing is selected yet. It just brings up the empty panel for no reason.

CodingSquirrel:
I don't know where or why, but at some point (in the commit history, not in game time) the unit view window stopped fading in and out. It just instantly pops up and disappears. That should also probably be fixed.

ChrisKitching:
Tracking 3641 for somewhat serious UI brokenness.

ChrisKitching:
Excellent work.

This UI is still somewhat gnarly:
- Mousing-over an unselected unit no longer causes the unitview to appear (it does so in stock, however)
- When you cease having your mouse on a construction menu item, the construction detail tooltip stuff fades out, but the unitview appears instantly. We should remove the fade animation from these construction detail tooltips so there's no transient overlap (which looks slightly annoying)
- The "abilities" box just looks retarded in some situations, such as for this interceptor:
https://www.dropbox.com/s/sx7592yrutyxoce/Screenshot%202015-07-10%2023.58.38.png?dl=0

ChrisKitching:
Merged at 3be26ae1a56cd4f6aeab513a2d15bc58f73e3de7

PR #747, opened 2015-07-12
Keybind reset and unassign buttons refresh UI

CodingSquirrel:
The UI will now update with all of the new values when using the reset all button and the unassign key button.


ChrisKitching:
Merged at 63d8176466fd30b771e670f6f3f11182c39d5917

PR #748, opened 2015-07-12
Fixed bug with build effects

Crotalus:
Paused commanders still had the build effects going.

~~Did some refactoring so the build effect stuff is handled in the base Unit class instead.~~

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1749,12 +1749,11 @@ ConstructionUnit = Class(MobileUnit) {
     end,
 
     OnStartBuild = function(self, unitBeingBuilt, order )
-
-    if unitBeingBuilt.WorkItem.Slot and unitBeingBuilt.WorkProgress == 0 then
-        return
-    else
-        MobileUnit.OnStartBuild(self,unitBeingBuilt, order)
-    end
+        if unitBeingBuilt.WorkItem.Slot and unitBeingBuilt.WorkProgress == 0 then
+            return
+        else
+            MobileUnit.OnStartBuild(self,unitBeingBuilt, order)
+        end
         -- Fix up info on the unit id from the blueprint and see if it matches the 'UpgradeTo' field in the BP.
         self.UnitBeingBuilt = unitBeingBuilt
         self.UnitBuildOrder = order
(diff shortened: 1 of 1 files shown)


Crotalus:
An excercise for the future could be to merge self.UnitBeingBuilt and self.BuildingUnit also, could as well check if self.UnitBeingBuilt is nil

ChrisKitching:
We should not be moving specialised logic into Unit.

We really need to introduce classes for "things that build other things", but that's not something we should do on release day.
Instead, for now, let's do the minimum required to fix this bug, duplicated as necessary, and cull it later on when we make the inheritance chains for construction units vaguely sensible.

Crotalus:
There, changed the patch into a quickfix instead

ChrisKitching:
Oh,that was quick, thanks :D

If you'd like to lend a hand on the overall unfucking of all this crap, take a look at:
https://github.com/FAForever/fa/issues/647

(It's more of a problem statement than anything terribly helpful: I've not really made much of a start).
There's a bunch of duplicated logic between the various things that build other things, and tons of stuff lurking in Unit. Really, we want to introduce base classes for stuff-that-builds-stuff and pull all the specialised crap out of Unit. Unit is already way too big (but recently we've made some progress pulling specialised stuff out of it, such as all the air transport rubbish).

lua/defaultunits.lua
```diff
@@ -1749,12 +1749,11 @@ ConstructionUnit = Class(MobileUnit) {
...
-    if unitBeingBuilt.WorkItem.Slot and unitBeingBuilt.WorkProgress == 0 then
-        return
-    else
-        MobileUnit.OnStartBuild(self,unitBeingBuilt, order)
-    end
+        if unitBeingBuilt.WorkItem.Slot and unitBeingBuilt.WorkProgress == 0 then
+            return
+        else
+            MobileUnit.OnStartBuild(self,unitBeingBuilt, order)
+        end
```
ChrisKitching: Am I going blind, or is this actually a no-op chunk?

Crotalus: Yeah, indenting

PR #750, opened 2015-07-12
Fixing chat regression

CodingSquirrel:
Apparently the redundant skin setting is required to stop the chat from
breaking.


Sheeo:
Good catch.

PR #759, opened 2015-07-13
Satellite rebuilding

aeoncleanse:
Only remaining problem is the UI not working properly with it.


Uveso:
@aeoncleanse 

Change in `construction.lua` line 405:

```
-        defaultTabOrder = { t3 = 1, t2 = 2, t1 = 3 }
+        defaultTabOrder = { t4 = 1, t3 = 2, t2 = 3, t1 = 4 }
```
and the UI will handle the Factory thats is only able to build experimental's.

The UI only checks techlevel 1-3 to activate the build tab. If your factory only can build tech4 (experimental) the experimental tab will not set to active.

Thats why you see the icons from the unit that you selected before the SatCenter.

Uveso:
btw, if you CTRL+K the satellite it calls `OnKilled` inside XEA0002 and
```
        self.IsDying = true
        self.Parent.Satellite = nil
```
is set correctly.

But if the satellite is destroyed by a nuke, `OnKilled` will not be called. Only `OnDestroy` is called (not present in XEA0002).
This will lead to not set the upper veriables and rebuilding the sat is blocked.

And if you use Black ops with this;
```
        BuildableCategory = {
            'SATELLITE',
        },
```
This allowes the XEB2402 to build the Artemis satellite.
Better writing this:
```
        BuildableCategory = {
            'xea0002',
        },
```

PR #761, opened 2015-07-13
Fix engies consuming resources repairing dead buildings

aeoncleanse:


changelog.md
```diff
@@ -63,10 +63,11 @@ Other
 - Current command mode no longer reset if an engineer dies in the current selection
 - Fixed bug with chat messages having wrong recipients in replay
 - Seraphim and Aeon T2 shields can now pause while upgrading to T3
-- Fixed RateOfFire buffs not being applied correctly
+- Fixed RateOfFire buffs not being applied correctl
 - SCUs will no longer rebuild buildings they lack to ability to build
 - Upgraded buildings rebuilt by assisting SCUs will now cost the correct amount
 - FAF no longer blows up if the game hasn't first been launched to create a profile
+- Engineers no longer drain resources to repair a unit which is in the process of blowing up
```
Sheeo: 3641 is released already, we can't amend it like this.

Sheeo:
I don't think that this is worth the extra complexity for what it gives -- this happens in very rare cases and it's never for too long. The consumption update system needs an overhaul overall -- and I think we should aim for this instead. Closing this for now.

PR #765, opened 2015-07-14
Fixing issues with Unit Info hover

CodingSquirrel:
Fixes a regression where the unit info hover doesn't come up unless you have the Single Unit Selected Info option turned on. Mostly reverts the code which caused the issue, and puts the check back on the OnFrame function.

Also restored the fade in and fade out for the info, since that basically came free with the reversion of the code. That commit can be left out of the merge though if you guys disagree with adding it back. One argument I would make for it (aside from the aesthetically pleasing animation) is that it helps the window freak out a bit less when scrolling quickly across build icons that have spaces between them. Currently this results in either an instant switch back and forth to the selected unit, or the pane instantly disappearing and reappearing very quickly (depending on your setting).


lua/ui/controls/worldview.lua
```diff
@@ -165,14 +163,6 @@ WorldView = Class(moho.UIWorldView, Control) {
             self:ResetDecals()
         end
 
-        if options.gui_enhanced_unitview ~= 0 then
```
Sheeo: Doesn't this completely remove the use of the option, then?

We should remove it from the options panel, if that's the case.

CodingSquirrel: Not removed. Inverted and moved to line 532 of unitview.lua.

lua/ui/game/unitview.lua
```diff
@@ -511,46 +526,16 @@ function CreateUI()
...
-
-        currentInfo = nil
-        controls.bg:Hide()
-    end
 end
 
 function OnSelection(units)
+    if options.gui_enhanced_unitview == 0 then
+        return
+    end
```
CodingSquirrel: Don't need to save selection if we're not going to use it because of the option. If the option is turned off then selected Unit is always nil.

CodingSquirrel:
I think I just found a bug with shielded units.

CodingSquirrel:
That section of code should like like this change, but for some reason some old code found its way back in. bbd80749343608228e7de3085bf8ba27b2df3802

CodingSquirrel:
Umm...nevermind on that shield issue? I'm actually a bit confused. In the repo there's this section here (https://github.com/FAForever/fa/blob/develop/lua/ui/game/unitview.lua#L433) that is causing issues with the shield locally, but when I submitted to remove that section nothing actually showed up in the PR.

CodingSquirrel:
@Sheeo @ChrisKitching Are we waiting on anything to push this out?

Sheeo:
Yes squirrel, we're waiting for me to be at a terminal. Unfortunately right now that's how it is — in the near future we will have automated deployment that will allow a broader range of people to push updates. Apologies for the inconvenience, I'll do this as soon as I can tomorrow.

ChrisKitching:
(I'll probably get a chance to test and merge this and the mod stuff tonight, so it's okayish)

CodingSquirrel:
No problem, just wanted to know if there was anything else I needed to do, or if we were waiting on something else to go with it or something.

ChrisKitching:
Merged

PR #770, opened 2015-07-14
Adjust torp bomber torps to not bottom out

aeoncleanse:


aeoncleanse:
Closed because another approach was picked

PR #771, opened 2015-07-16
Hoverslow

aeoncleanse:


lua/defaultunits.lua
```diff
@@ -1856,6 +1856,20 @@ AircraftCarrier = Class(SeaUnit, BaseTransport) {
 HoverLandUnit = Class(MobileUnit) {
 }
 
+SlowHoverLandUnit = Class(HoverLandUnit) {
+    OnLayerChange = function(self, new, old)
+        HoverLandUnit.OnLayerChange(self, new, old)
+        -- Slow these units down when they transition from land to water
+        -- The mult is applied twice thanks to an engine bug, so careful when adjusting it
+        -- Newspeed = oldspeed * mult * mult
+        if old == 'Land' and new == 'Water' then
```
Sheeo: Should these cases not simply be `if new == 'Water' then slow() else normal() end`?

There could be other layerchange transitions.

aeoncleanse: Hmmm. i did it like this in the mod because I was just modding it into unit.lua so I was worried about things like planes landing etc. Here though... The only units inheriting are the ones we want to, so if that's limit, I suppose we could. It also fixes the side effect of the Tempest-build hovertanks not slowing down properly.

ChrisKitching: > There could be other layerchange transitions.

Such as when you drop a hover unit off an air transport into the sea.

lua/defaultunits.lua
```diff
@@ -1856,6 +1856,22 @@ AircraftCarrier = Class(SeaUnit, BaseTransport) {
 HoverLandUnit = Class(MobileUnit) {
 }
 
+SlowHoverLandUnit = Class(HoverLandUnit) {
```
Sheeo: With the blueprint change in place, this code should just be moved into `HoverLandUnit`.

Make sure that the inheritance chain up to doesn't override `OnLayerChange` (It shouldn't), but if so, make sure it calls the base class.

ChrisKitching: You should probably always call the base class anyway, so we can add intermediate methods without obscure madness.

aeoncleanse: If I put it into HoverLandUnit, all hovering units will check for that bp when only 2 have it. Is it not better to keep it confined for now, then move it across later if/when Zock decides he wants to apply speed modification to a larger number of units?

PR #772, opened 2015-07-16
Clean up comments and tabs

aeoncleanse:
Cl

### DIFF
--- a/lua/armordefinition.lua
+++ b/lua/armordefinition.lua
@@ -1,81 +1,71 @@
-﻿#****************************************************************************
-#**
-#**  File     :  /lua/armordefinition.lua
-#**  Author(s):
-#**
-#**  Summary  :
-#**
-#**  Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
-#****************************************************************************
-#
-# Armor Type Definitions
-#
+﻿-----------------------------------------------------------------
+-- File     :  /lua/armordefinition.lua
+-- Author(s):
+-- Summary  :
+-- Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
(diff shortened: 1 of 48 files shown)


Sheeo:
Cl?

aeoncleanse:
I have no clue. And WHY is travis failing. Grrrr

lua/armordefinition.lua
```diff
@@ -1,81 +1,67 @@
...
+-- File     :  /lua/armordefinition.lua
+-- Author(s):
+-- Summary  :
+-- Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
+-----------------------------------------------------------------
 
+-- Armor Type Definitions
+-- Laid out as follows:
+-- 'Armor Name'
+-- 'Damage Type - Multiplier'
```
Sheeo: This comment should be something like:

Armor type definitions are of the form:

    {
        {
          'Name',
          'DmgType Multiplier'+
        }*
    }

But honestly I prefer just having them inline, it's more readable.

Sheeo:
This kind of work really shouldn't be targeted at the balance branch @aeoncleanse, it's unrelated to balance.

I know some of the files aren't in the develop branch, but could you split off a PR targeted at develop to clean up those, and then we'll clean up these files when they make it to develop?

aeoncleanse:
I targeted at balance because this is exclusively a cleanup of files affected by the balance branch so far. All this does is clean up the files balance has changed so far, and no others.

ChrisKitching:
These sorts of commits are not useful: they clutter `git blame` and don't add any actual value.

Do this as you go along to code you touch. You can also do this to GPG-files in the commit you add them with.

PR #774, opened 2015-07-16
Stop harb shooting while reclaiming

aeoncleanse:


ppipp:
I tested it, seems to work well enough.

PR #796, opened 2015-07-20
Make structure wrecks/blips invisible to enemies

Sheeo:
GPG left a comment about buggy prop intel, but I'm unable to reproduce the bug.

This kills any intel blips when a structure unit is destroyed, encouraging scouting.

This is an alternative solutin to Crotalus' in #425. A problem remains with the mass reclaim labels being visible even for invisible wrecks.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -15,6 +15,7 @@ local Util = import('utilities.lua')
 local EffectUtil = import('EffectUtilities.lua')
 local EffectTemplate = import('/lua/EffectTemplates.lua')
 local ScenarioUtils = import('/lua/sim/ScenarioUtilities.lua')
+local ScenarioFramework = import('/lua/ScenarioFramework.lua')
 local EffectUtil = import('EffectUtilities.lua')
 local Entity = import('/lua/sim/Entity.lua').Entity
 local Buff = import('/lua/sim/Buff.lua')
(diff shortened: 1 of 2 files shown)


Sheeo:
@Crotalus @Zockyzock @ppipp need your opinions on this.

I don't think we should put something this controversial into 3643, and we definitely need to nail it so the mass labels don't appear.

We need to test for the engine bug and verify that it actually exists.

Crotalus:
A problem is that scouted wreck aren't visible on the map afterwards, resulting in map props not visible at start of game. I guess they could be excluded from the 'Intel' stuff.

I don't think there is an easy way to check if a prop is currently visible for someone (to solve the mass label stuff).

ChrisKitching:
Yeah, this is about where I ground to a halt on this work myself: persistence of intel with wreckage isn't simple. Unlike units, props don't have an event listener for OnSpottedBy (which we could use to make them persist), so I'm not sure how to proceed there.

If we can solve the "once a wreck is spotted it persists" problem, we can fairly straightforwardly handle the "default map wrecks need to start visible" thing: add a twiddle to the routine that spawns the wrecks that tacks on either a VizMarker, or does InitIntel/EnableIntel/DisableIntel for each wreck (similar to the approach I used to make laser source units visible)

Sheeo:
We could investigate how much load making wreckage props inherit from unit instead would cost. Of course disabling as much AI and other unit like routines as possible.

Issue #780, opened 2015-07-17
Sera T2 Transport allows active shields in hooks


Varaxis:
Sera T2 Transport allows all shields to be active, both bubbles and personal shields. No other transport allows this, based on testing in sandbox.

duk3luk3:
Should be an easy fix, I'm taking a look at it.

duk3luk3:
Investigation progress: Sera T2 Transport is not part of the mod, so its script gets loaded from base files, so it doesn't inherit from airtransport base class. I suspect this is where the shield disable stuff happens.

duk3luk3:
Indeed the BaseTransport class in https://github.com/FAForever/fa/blob/develop/lua/defaultunits.lua#L1646 has shield disable logic for transported units.

So what probably needs to be done to fix this is to redo the Sera T2 Transport (XSA0104) to properly inherit from AirTransport base class.

Who is familar with the reclassing stuff? @Sheeo? @aeoncleanse?

Sheeo:
@ChrisKitching is the one who did the original refactor -- it saves sim execution time.

Indeed that's a easy fix.

ChrisKitching:
> So what probably needs to be done to fix this is to redo the Sera T2 Transport (XSA0104) to properly inherit from AirTransport base class.

Yup. Do that. It's almost certainly a trivial matter of deleting most of the script :P

> it saves sim execution time.

And, more usefully, gives us abstractions that make sense (AirTransport itself inherits from BaseTransport (a base class of "things that hold other things" which encompasses things such as aircraft carriers, the CZAR, and kennels, all of which previously had variously inconsistent copies of identical code in them :/))

duk3luk3:
I won't have the time to fix and test this for a while.

PR #802, opened 2015-07-21
Fix Sera T2 Transport - shields of loaded units were not disabled

anihilnine:
Changed unit script to inherit defaultunits AirTransport class as per:
https://github.com/FAForever/fa/issues/780


ChrisKitching:
Please add the new GPG script file in one commit, and then do your modifications to it in another commit.
This makes it much easier to determine which things we've changed relative to GPG's original code.

Looks like you did the right thing here, anyway :)

anihilnine:
that makes very much sense

Sheeo:
I did this, thanks for the contribution @anihilnine :)

Issue #806, opened 2015-07-21
sera destroyers don't surface after building


Zockyzock:

quark036:
this looks similar to #611, which was fixed with eb3a1f3

Sheeo:
Seems like it's a regression from that very commit:

    -	OnStopBeingBuilt = function(self, builder, layer)
    -		SSubUnit.OnStopBeingBuilt(self, builder, layer)
    -		IssueDive({self})
    -	end,

This code would make the subunit surface again.

ChrisKitching:
> This code would make the subunit surface again.

And reintroduce the very bug that commit fixes. They would once again toggle surfacedness when changing ownership.

That is not the solution.

ChrisKitching:
We probably instead want to just make the unit be built in the surface layer so we don't need to faff about issuing dive orders when it's built. We should similarly make aircraft be built in the air layer so we can throw away the code that babysits that particular thing.

ChrisKitching:
(Or possibly build it submerged and stay submerged. One or the other)

quark036:
Is there no detection to check if it is surfaced?

Sheeo:
So, testing this -- it turns out building the destroyer on land looks rather bad. It gets build in midair and floats downward until it hits the water layer.

I think we should go for the other solution.

ChrisKitching:
... wait, what?

Dont build it on land, build it in the water surface layer.

bamboofats:
![untitled-1](https://cloud.githubusercontent.com/assets/13047748/8880377/c15bea9c-326a-11e5-8646-7de87299c577.jpg)

This fix appears to bug out the destroyer textures. Now fresh out of the factory(left) still has the construction texture. Expected(right)

ChrisKitching:
The  recentCrotalus PR fixes this. Needs deployment. @sheeo

Sheeo:
Bamboofats please describe how you got the rendering to the right. Thats not tje expected rendering of the destroyer if you ask me...

Sheeo:
@cktitching please link the PR if it is relevant.

ChrisKitching:
https://github.com/FAForever/fa/commit/a46e5ca4b9ae4f47fa2f1182b61ee07364a4cd69

bamboofats:
![untitle3](https://cloud.githubusercontent.com/assets/13047748/8887774/2db89c6e-32c3-11e5-9260-9ef8b66622f6.jpg)
[...]

Issue #611, opened 2015-05-19
Seraphim destroyer submerges when gifted


Sheeo:
[Original bug report](http://forums.faforever.com/forums/viewtopic.php?f=3&t=9965#p100231)

shalkya:
i think it's something related with the fact that sera destro is the only unit build underwater, and that automatically submerge after been built. 
gifting the unit would be as if the unit as been built once again (i don't know if that's clear for you though)

PR #807, opened 2015-07-22
Seraphim Destroyer no longer submerged when gifted or built

quark036:
Checks if currently submerged before issuing dive command, addresses #806 and #611

### DIFF
--- a/units/XSS0201/XSS0201_script.lua
+++ b/units/XSS0201/XSS0201_script.lua
@@ -48,6 +48,11 @@ XSS0201 = Class(SSubUnit) {
         end
 
         SSubUnit.OnKilled(self, instigator, type, overkillRatio)
+		
+		OnStopBeingBuilt = function(self, builder, layer)
+			SSubUnit.OnStopBeingBuilt(self, builder, layer)
+			if self:GetIsSubmerged == -1 then
+				self:IssueDive
     end,
     
     OnMotionVertEventChange = function( self, new, old )


Sheeo:
As Chris noted [here](https://github.com/FAForever/fa/issues/806#issuecomment-123809350), this solution will not do it.

We instead need to make the destroyer get built on land.

quark036:
Ok, sorry about that. In the luadoc, issuedive and getissubmerged didn't have parentheses afterwards, whereas others did, so I assumed they were more like properties and didn't need them.

Sheeo:
Yes the luadoc isn't terribly consistent about these things, unfortunately.

The only real way to verify stuff works is to launch the game :) Have you gotten a working environment for running the game with changes?

quark036:
I do, but not with me now

Probably shouldn't have done a pull request without testing, wasn't really paying attention

units/XSS0201/XSS0201_script.lua
```diff
@@ -48,6 +48,11 @@ XSS0201 = Class(SSubUnit) {
         end
 
         SSubUnit.OnKilled(self, instigator, type, overkillRatio)
+		
+		OnStopBeingBuilt = function(self, builder, layer)
+			SSubUnit.OnStopBeingBuilt(self, builder, layer)
+			if self:GetIsSubmerged == -1 then
+				self:IssueDive
```
ChrisKitching: Missing "end": syntax error.

units/XSS0201/XSS0201_script.lua
```diff
@@ -48,6 +48,11 @@ XSS0201 = Class(SSubUnit) {
         end
 
         SSubUnit.OnKilled(self, instigator, type, overkillRatio)
+		
+		OnStopBeingBuilt = function(self, builder, layer)
+			SSubUnit.OnStopBeingBuilt(self, builder, layer)
+			if self:GetIsSubmerged == -1 then
```
ChrisKitching: This is equivalent to `false`: You're comparing a function to -1, not invoking it.

units/XSS0201/XSS0201_script.lua
ChrisKitching: This is a nop: not calling the function.

quark036:
Yes, I get it, sheeo pointed that all out

ChrisKitching:
Whoops. :P

quark036:
Serves me right for not paying attention

aeoncleanse:
Use the fact that (I think) builders assign themselves as a property to a build unit. So a unit knows what built it. A unit transferred, the new one is simply spawned, and will have that value nil. Make it so the unit, OnStopBeingBuilt, only issues the surface command if it has a builder assigned.

ChrisKitching:
Sounds a lot harder than just building it on the surface to start with.

aeoncleanse:
Of course, see if you can do that first :)

quark036:
I feel like my method would work if I didn't screw it up, but yeah, building it on the surface would be better.

PR #810, opened 2015-07-24
Use self.UnitBeingBuilt instead of getter

### LINKED ISSUE #809: GetUnitBeingBuilt() doesn't exist for mobile construction units
Error in log:

```
warning: Error running OnPaused script in Entity uel0208 at 1afb5f08: ...data\faforever\gamedata\lua.nx2\lua\defaultunits.lua(1746): attempt to call method `GetUnitBeingBuilt' (a nil value)
         stack traceback:
         	...data\faforever\gamedata\lua.nx2\lua\defaultunits.lua(1746): in function <...data\faforever\gamedata\lua.nx2\lua\defaultunits.lua:1741>
```

Crotalus:
Seems like GetUnitBeingBuilt() function isn't defined for mobile construction units.
Until we a better inheritance chain, let's just use self.UnitBeingBuilt

Fixes #809

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -532,10 +532,6 @@ FactoryUnit = Class(StructureUnit) {
         self.BuildingUnit = false
     end,
 
-    GetUnitBeingBuilt = function(self)
-        return self.UnitBeingBuilt
-    end,
-
     OnDestroy = function(self)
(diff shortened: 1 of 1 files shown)


ChrisKitching:
The getter is pointless anyway.

PR #816, opened 2015-07-25
Chat fixes

CodingSquirrel:
Multi-line messages no longer time out faster the more lines it's on.
New messages no longer necromance the previous line.
Line wrapping is much less stupid.

### DIFF
--- a/lua/maui/text.lua
+++ b/lua/maui/text.lua
@@ -210,29 +210,38 @@ function WrapText(text, lineWidth, advanceFunction)
             local wordWidth = advanceFunction(word)
 
             -- if the word is longer than the max width, break it up in to multiple lines
-            if wordWidth > lineWidthFunc(table.getsize(result)) then
-                -- start long word on its own line (only if there is already a result)
+            if wordWidth > lineWidthFunc(curLine) then
                 if result[curLine] then
                     curLine = curLine + 1
                 end
 
-                -- determine how many lines are needed to display this word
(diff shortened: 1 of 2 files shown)


quark036:
Necromance?

CodingSquirrel:
You ever notice when your chat feed is empty and someone says something in chat, all of a sudden the previous line that was said however long ago shows up again?

quark036:
Ah

ChrisKitching:
Isn't that a sort of nice feature in some ways? Instant context.

CodingSquirrel:
Except when it's not related and it looks like something old just happened. I always get confused for a moment when I see something like "RAS Completed" pop up even though it finished a few minutes ago. Plus there's the chat window for that.

ChrisKitching:
That's true. Makes sense, then. I'll review the code when I'm not drunk, but it sounds pretty sane.

CodingSquirrel:
Thanks. I imagine you can't do any testing from there, but I'd appreciate if anyone could try breaking the word wrap.

ChrisKitching:
Wouldn't a central word wrap function that takes a line length and some text and returns a list of lines be useful?
I suspect specialised WrapText functions are sprinkled everywhere.

CodingSquirrel:
That's what the function in text.lua is for. It takes a line of text, a function to calculate line length, and for calculating pixel length, and returns a list of lines.

Issue #811, opened 2015-07-24
improvements to ingame lobby


Zockyzock:
- [x] Add a search function for maps
- [ ] Don't dismiss map filters when choosing a map/closing the window
- [ ] Make it possible to delete maps from folder by hitting delete when map is selected

I guess this is the wrong repo, but i don't know where to put it, sorry.

duk3luk3:
It's the right repo, the mapselect dialog lives in lua/ui/dialogs. I've started working on it.

duk3luk3:
The name filter was pretty easy once I figured out how to make the layout UI work.

>  -Don't dismiss map filters when choosing a map/closing the window

This is going to be annoying because due to the way dialogs work in FAF the dialog is recreated from scratch every time it is opened, so the filters need to be saved/restored in the lobby code that handles the button that launches the dialog.

>  -Make it possible to delete maps from folder by hitting delete when map is selected

I have no idea how to even approach that.

ChrisKitching:
Just pickle the activated filter state into the profile and load it when the dialog is opened.

That's also how dialogs worked pre-faf. Keeping the dialog content alive after closing it means you need to implement careful life cycle management of dialog to avoid leaking invisible UI elements. It's just not worth it.

duk3luk3:
#818 Now also contains filter caching.

The code isn't something to write home about but I have to say I am kinda proud of what I've achieved.

duk3luk3:
@ChrisKitching / @Sheeo review and test pls?

Garanas:
I am closing this issue: the map filters are kept between games (likely through preference file) and we can not delete files via the game as the IO functionality that we need for that are disabled.

PR #818, opened 2015-07-26
Add name filter to lobby mapselect dialog

duk3luk3:
Add name filter ui and logic to lobby mapselect dialog
Add localization for new title element
~~Cache filters selected in mapselect dialog to profile prefs~~
Reselect map when filters change

Advances #811


duk3luk3:
Hey, for some reason I didn't see your comments. I replied to the inline comments.

PR #823, opened 2015-07-26
Fixing Jammer Crystals

speed2CZ:
It was set to stop jamming only when killed, now it should work properly
even when reclaimed.
(Learning with Github part2: Lets hope I wont fuck up anything)


units/XSC9002/XSC9002_Script.lua
```diff
@@ -0,0 +1,90 @@
...
+    end,
+
+    OnKilled = function(self, instigator, type, excessDamageRatio)
+        if self.airChildUnit then self.airChildUnit:Destroy() end
+        if self.landChildUnit then self.landChildUnit:Destroy() end
+
+        SStructureUnit.OnKilled(self, instigator, type, excessDamageRatio)
+    end,
+
+    OnReclaimed = function(self, instigator, type)
```
CodingSquirrel: OnDestroy should handle both of these cases.

speed2CZ: well, it didn't

units/XSC9002/XSC9002_Script.lua
```diff
@@ -0,0 +1,84 @@
...
+            WaitSeconds(Random(7,13))
+
+            self.airChildUnit:Destroy()
+            self.airChildUnit = nil
+
+            #WaitSeconds(Random(2,7))
+        end
+    end,
+
+    OnDestroy = function(self, instigator, type, excessDamageRatio)
```
CodingSquirrel: OnDestroy only takes self as input. Same for where it's called on line 76

ChrisKitching:
This should be one commit.

You can force push the branch you have in your PR to update the content here: no need to open new PRs if you bugger up the code (I do this *all the time* :P)

speed2CZ:
I will need some advice how to do that, so far I work with trial/error method

PR #826, opened 2015-07-27
Set GRR to 3 as default

Crotalus:
This helps with units running after a faster moving unit while on attack
move (or assist ground) while getting shot by other units. It will
instead give up and find a new target along its ordered path.

As a test, put a destroyer on attack move into a group of frigates, after the destroyer locks onto the nearest frigate, move that unit out of the destroyer's reach. Without this patch it will go after like a moron, pointing the gun at the out of range frigate and only fire torpedos at the other ones.

### DIFF
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -488,6 +488,9 @@ function PreModBlueprints(all_bps)
             if br then
                 if not bp.AI then bp.AI = {} end
                 bp.AI.GuardScanRadius = br
+                if not bp.AI.GuardReturnRadius then
+                    bp.AI.GuardReturnRadius = 3
+                end
             end
         end
 


Sheeo:
I think we should test out this change in the beta.

duk3luk3:
@Sheeo @ChrisKitching Looks ready to merge?

ChrisKitching:
I think the view is we want blueprints to be more literal, with much less of this sort of load-time rewriting.

This makes unit databases and blueprint compilation simpler.

duk3luk3:
That's a good idea and you should make an issue for it but until we have figured out a new way to handle blueprints (Sheeo was thinking about a blueprint compiler), I think we should merge this anyway if it's worth trying it out.

ChrisKitching:
That still conflicts with the "we want blueprints to be more literal and not break unit databases" thing.

duk3luk3:
So you want a freeze on all blueprint-related changes until we've figured out a way to get rid of Blueprint.lua?

ChrisKitching:
That's not even almost what I said.

We want the properties of blueprints to be explicitly defined in blueprints (such that it is sufficient to look only at the blueprint files to determine all the properties of a unit, without the need to apply complex interpretation logic (such as applying default values)).

duk3luk3:
It's madness to set GRR to 3 in all blueprints by hand...

Sheeo:
@ppipp Could you cherry pick this into the beta and test it out? :)

Crotalus:
`release/3652`

PR #827, opened 2015-07-27
Normalize Notha aircrash damage

quark036:
It was 25, all other t2 fb are 200

Credit to KrogothFTW and Korbah for pointing it out.


Sheeo:
I'd like to point out that t1 and t2 transports have highly irregular crash damage as well.

@Zockyzock @ppipp want to make a decision on this? :)

quark036:
Want me to put those into this pr too?

quark036:
Also, t1 transports all have 100 crash damage

quark036:
For t2, cybran and aeon have no crash weapon, and uef and sera both do 25 dmg, which seems odd that its lower than t1 transports

quark036:
What seems sensible is make all the t2 transports do 200 crash damage, like the t2 fbs

quark036:
Oops I messed up with github, making a new pr

PR #834, opened 2015-07-28
Fixed bug with consumption and paused builders

### LINKED ISSUE #829: Engineer problem perhaps even worse
Replay #3694494.  Mods enabled; tested without, same problems.

Unless used very carefully, "Pause" now means that construction will be paused, but resources will still be expended.

Honestly, why not go back to 3641 engineers until this can really be fixed?

Crotalus:
A long standing (pre 3640) bug which caused paused engineer to still drain
resources but not contributing to the actual process of the building.

This was also a problem with paused commanders and starting upgrades. They
then drained resources without adding progress.

Replay 3694494 shows the bug

Fixes #829

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2132,16 +2132,20 @@ Unit = Class(moho.unit_methods) {
             self:CheckAssistersFocus()
         end
 
-
-
         local bp = self:GetBlueprint()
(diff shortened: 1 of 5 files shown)


Crotalus:
Fixed another long-standing bug with engineers not starting to assist paused, upgrading commanders

ChrisKitching:
I think we stole all the bits of this we care about, maybe let's do a new PR for the rest of it (plus it bitrotted)

PR #835, opened 2015-07-28
Normalize Notha deathweapon damage

quark036:
It was 25, other t2 fbs are 200


quark036:
From the other pr:

T2 transports also have odd crash damage, with cybran and aeon having 0, and uef and sera having 25. It would make sense to make them all 200 like the fighter bombers.

T1 transports all have 100 crash damage

quark036:
So, @Zockyzock , @ppipp , what do you want me to do? Could do nothing, just the notha deathweapon, or normalize all the t2 transports too.

quark036:
Dammit Github. I screwed up again

PR #838, opened 2015-07-29
Replace GetUnitBeingBuilt() with UnitBeingBuilt

Crotalus:
There's also calls by original scripts not hooked in FAF repository, what should be done about those?

Either we hook / replace in those or we simply remove / reduce the warning in GetUnitBeingBuilt(), maybe reduce it to one notification per game (log spamming causes performance degradation)


Sheeo:
Woops, didn't see this before I made cc671ff611e43e2b81e0ab5204effcb4fa54db32. Looks like we made the same changes?

Sheeo:
To accomodate the logspam, my quick solution is 3c0a6957a0b50c3c456fe33adce07ab28921f5d3.

However, we really need to keep https://github.com/FAForever/fa/issues/201 in mind and not introduce more GPG code.

We're growing in scale and we really need to comfortably say that we're not shipping copyrighted material.

PR #853, opened 2015-08-11
Attempt at solving beam not switching off

bamboofats:
Isnt the best way to do this. Uses the WatchForHoldFire to cut the beam if hold fire, or if stopped being in an "attack order" state. But i cant figure out how to differentiate the state between idle, and automatic attack (no command given). So i let the function cut the beam (during auto attack without command) and restart it every 1 sec. This makes the beam start sound play every 1 second though, so I mute it if the time between beam stop and restart is 1 tick.
This does appear to fix the beams for various attack scenarios.

### DIFF
--- a/lua/sim/defaultweapons.lua
+++ b/lua/sim/defaultweapons.lua
@@ -951,7 +951,7 @@ DefaultBeamWeapon = Class(DefaultProjectileWeapon) {
 
     OnCreate = function(self)
         self.Beams = {}
-
+        self.BeamCheck = true
         -- Ensure that the weapon blueprint is set up properly for beams
         local bp = self:GetBlueprint()
         if not bp.BeamCollisionDelay then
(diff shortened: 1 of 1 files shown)


bamboofats:
https://github.com/FAForever/fa/commit/37821a67d4a729a81cdadc51aae6a0b92ed4ef95
https://github.com/FAForever/fa/commit/6e2cddfe8c50578db6be4abc6260db012080ab98

Sheeo:
Thanks for the contribution, @bamboofats :)

ChrisKitching:
Yes, sorry for being a bit brisk earlier, thanks for letting me steal chunks from your PR: I wasted a few hours looking for a better way yesterday. That line of yours I nicked reduced it to the easier problem of fixing the sound effect bug, which let us include it in today's release.

Thanks!

bamboofats:
Not a problem at all!

Issue #843, opened 2015-08-03
spread attack not working as intended


shalkya:
http://forums.faforever.com/forums/viewtopic.php?f=3&t=10485&p=106485#p106485

this make spread attack un-reliable. better to fake a pause, and manual target opponent unit. Since first strike/path can change a game. for exemple, you can kill 100% of your opponent's cruisers with your torpedo bomber (if they were all targeted well, by the good number of torpedo) and spread attack make your torpedo bomber only attack 60% of his cruiser on the first path, making ton of overkill. you lose all your torpedo bomber because of that, and the game in the end.

CodingSquirrel:
So I was playing around with this a bit and while having a stable split attack is nice, random still probably gives the best results. The naive approach I tried (first unit kills 1,2,3, and 4, second unit kills 2,3,4, and 1, etc.), if you have something like 5 t1 queued up first and then 5 t3 queued after, once the t1 is dead the first 6 units are killing the first t3, while the last 4 are killing the other 4 t3. I'm having a hard time coming up with a suitable algorithm that solves that and other similar issues.

Perhaps a hybrid, where the first attack is stable but the rest are random?

Sheeo:
https://github.com/FAForever/fa/pull/851

quark036:
So is this good?

CodingSquirrel:
Pull request was moved to #858 and is awaiting review.

PR #858, opened 2015-08-13
Improving Spread Attack

CodingSquirrel:
First Attacks are uniformly spread across targets. All further attacks
are random.

### DIFF
--- a/lua/spreadattack.lua
+++ b/lua/spreadattack.lua
@@ -95,7 +95,7 @@ function SpreadAttack()
   end
 
   -- Switch the orders for each unit.
-  for _,unit in ipairs(curSelection) do
+  for index,unit in ipairs(curSelection) do
     local unitorders = ShadowOrders[unit:GetEntityId()]
 
     -- Only mix orders if this unit has any orders to mix.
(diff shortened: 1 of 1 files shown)


lua/spreadattack.lua
```diff
@@ -132,8 +132,25 @@ function SpreadAttack()
       if beginAttack == nil or endAttack == beginAttack then
         break
       end
+      
+      -- Rearrange the first few attack orders (equal to the number of targets) so that the targets are uniformly attacked on the first pass.
+      -- For example, 3 units attacking 8 units (? denotes random target):
+      -- Unit 1: 1, 4, 7, ?, ?, ?, ?, ?
+      -- Unit 2: 2, 5, 8, ?, ?, ?, ?, ?
+      -- Unit 3: 3, 6, ?, ?, ?, ?, ?, ?
```
Sheeo: Instead of random targets on the second pass, why don't we just do round robin? This ensures that attacks are distributed uniformly even if you have three times as many units as targets.

Sheeo:
While we're touching this, I'd like to see `spreadattack.lua` split up into sim / ui parts, and have the respective parts moved into `/lua/sim` and `/lua/ui` where they belong.

The sim state should never import UI code, and the UI state should never import sim code. This way we can keep our sanity and minimize confusion.

lua/spreadattack.lua
CodingSquirrel: Explained my reasoning in #843 and the accompanying thread as well.

Crotalus:
`release/3652`

Issue #860, opened 2015-08-15
Megalith does not gain speed boost when being built underwater


bamboofats:
A megalith constructed underwater does not gain the water speed multiplier until it exits and reenters the water.

According to the file XRL0403_script.lua, the speedmult is only applied upon layer change, but not when the unitstopsbeingbuild.

Tested before and after the fix in a sandbox and it works.

On a side note, the self:DisableUnitIntel('Sonar') stuff recently introduced for the monkeylord and megalith could apply to the unitstopsbeingbuild part too, as the sonars might still be on when the units are completed on land at first.

aeoncleanse:
Nice one. Go. Fix.  :p

duk3luk3:
Did you make a PR?

bamboofats:
I didnt! You can go ahead and make one if you got the time. Thanks!

duk3luk3:
It sounded like you made a fix?

bamboofats:
Ah, i made a fix through my own mod. Now I just made a PR for the main game.

duk3luk3:
:+1:

duk3luk3:
Review pls.

PR #862, opened 2015-08-16
Megalith underwater speed boost

bamboofats:
https://github.com/FAForever/fa/issues/860

### DIFF
--- a/units/XRL0403/XRL0403_script.lua
+++ b/units/XRL0403/XRL0403_script.lua
@@ -74,6 +74,7 @@ XRL0403 = Class(CWalkingLandUnit) {
         CWalkingLandUnit.OnStopBeingBuilt(self,builder,layer)
         local layer = self:GetCurrentLayer()
         -- If created with F2 on land, then play the transform anim.
+        local LandSpeedMult = self:GetBlueprint().Physics.WaterSpeedMultiplier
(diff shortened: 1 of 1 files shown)


ChrisKitching:
Why not just call OnLayerChange and delete all the duplicated logic?

ChrisKitching:
(Also, verify that units _don't_ get a layer change event when they're built, I had assumed they did)

bamboofats:
Im not too sure about the other ways, pardon me. But if you got a better approach do use it! 
I tested one megalith built on land, then sent underwater,   2nd one built underwater, 3rd one spawned umderwater.
Then set them on a race across the ocean. Only the one built on land moves faster than the other 2.

But spawning a cybran destroyer on land automatically triggers the layer change to land. For the megalith it doesnt work properly. (since vanilla, I checked)

aeoncleanse:
Instead of duplicating logic,  just call layerchange from stopbeingbuilt?  No,  it doesnt call on build by default. Works for cybran because you aren't building it on land,  you're spawning it

duk3luk3:
Triggering layer change from stopbeingbuilt sounds like the right[tm] fix, I hope I can have a look at this soon.

Sheeo:
Looks like #872 supercedes this.

PR #870, opened 2015-08-19
Bail out of EnableUnitIntel if no DisableIntels

### LINKED ISSUE #869: Unit.lua EnableUnitIntel function logic error
Working on #860 I came across this.
It appears that when a unit is started building, `OnLayerChange` is called before the unit's object is fully initialized, in particular before the `SetupIntel` function runs, so for units where a layer change triggers intel changes, the `IntelDisables` field doesn't exist yet.

    WARNING: Error running OnLayerChange script in Entity url0402 at 1c3b9b08: e:\git\fa\lua\sim\unit.lua(2248): attempt to perform arithmetic on field `?' (a nil value)
             stack traceback:
             	e:\git\fa\lua\sim\unit.lua(2248): in function `EnableUnitIntel'
             	e:\git\fa\units\url0402\url0402_script.lua(67): in function <e:\git\fa\units\url0402    \url0402_script.lua:63>

The fix should be relatively simple (bail out of Enable/Disable intel function if `IntelDisables` field doesn't exist yet) and I will PR it when I am done with #860.

(Possibly there is a deeper problem here that requires a more comprehensive fix but unless I run into it directly I will just apply the simple fix)

(It would be really nice though if we could find out where `OnLayerChange` when starting build is being called from, but I guess it's the engine)

(Oh also the "simple fix" of bailing out already exists in `UnitDisableIntel` so that strengthens my case to just add it to `UnitEnableIntel` as well)

duk3luk3:
The EnableUnitIntel field is sometimes called before intel is initialized
(SetupIntel function).
Bail out and print a warning instead of just letting the function crash.
We add the warning to to DisableUnitIntel function where the
bail-out was already present as well for completeness.
Fixes #869

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2212,7 +2212,10 @@ Unit = Class(moho.unit_methods) {
 
     DisableUnitIntel = function(self, intel)
         local intDisabled = false
-        if not self.IntelDisables then return end
+        if not self.IntelDisables then
+            WARN('DisableUnitIntel called before IntelDisables field initialized.')
+            return
+        end
         if intel then
             self.IntelDisables[intel] = self.IntelDisables[intel] + 1
             if self.IntelDisables[intel] == 1 then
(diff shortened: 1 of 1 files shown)


duk3luk3:
This commit is now in the new PR I'm going to open for the layerchange fixes.

PR #872, opened 2015-08-19
Hotfix/layer change

### LINKED ISSUE #860: Megalith does not gain speed boost when being built underwater

duk3luk3:
To make amphibious units behave properly regardless of whether built on
land or underwater, the MobileUnit base class now has a OnStopBeingBuilt
handler that calls OnLayerChange.
Additionally we clean up the layer logic for Cybran Destroyer, Monkeylord
and Megalith, the main units affected by this.
Fixes #860 and #869

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1157,24 +1157,24 @@ RadarJammerUnit = Class(StructureUnit) {
     OnStartBuild = function(self, unitbuilding, order)
         StructureUnit.OnStartBuild(self, unitbuilding, order)
         self:SetMaintenanceConsumptionInactive()
-        self:DisableIntel('Jammer')
-        self:DisableIntel('RadarStealthField')
+        self:DisableUnitIntel('Construction', 'Jammer')
+        self:DisableUnitIntel('Construction', 'RadarStealthField')
     end,
 
     -- If we abort the upgrade, re-enable the intel
     OnStopBuild = function(self, unitBeingBuilt)
         StructureUnit.OnStopBuild(self, unitBeingBuilt)
         self:SetMaintenanceConsumptionActive()
-        self:EnableIntel('Jammer')
-        self:EnableIntel('RadarStealthField')
+        self:EnableUnitIntel('Construction', 'Jammer')
+        self:EnableUnitIntel('Construction', 'RadarStealthField')
     end,
 
     -- If we abort the upgrade, re-enable the intel
     OnFailedToBuild = function(self)
         StructureUnit.OnStopBuild(self)
         self:SetMaintenanceConsumptionActive()
-        self:EnableIntel('Jammer')
-        self:EnableIntel('RadarStealthField')
+        self:EnableUnitIntel('Construction', 'Jammer')
+        self:EnableUnitIntel('Construction', 'RadarStealthField')
     end,
 
     OnStopBeingBuilt = function(self,builder,layer)
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1428,6 +1428,14 @@ MobileUnit = Class(Unit) {
     CreateCaptureEffects = function( self, target )
         EffectUtil.PlayCaptureEffects( self, target, self:GetBlueprint().General.BuildBones.BuildEffectBones or {0,}, self.CaptureEffectsBag )
     end,
+
+    -- Units with layer change effects (amphibious units like Megalith) need
+    -- those changes applied when build ends, so we need to trigger the
+    -- layer change event
+    OnStopBeingBuilt = function(self,builder,layer)
+       Unit.OnStopBeingBuilt(self,builder,layer)
+       self:OnLayerChange(layer, 'None')
+    end,
 }
 
 --------------------------------------------------------------
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -23,6 +23,7 @@ local Buff = import('/lua/sim/buff.lua')
 local AIUtils = import('/lua/ai/aiutilities.lua')
 local BuffFieldBlueprints = import('/lua/sim/BuffField.lua').BuffFieldBlueprints
 local Wreckage = import('/lua/wreckage.lua')
+local Set = import('/lua/system/setutils.lua')
 
 -- TODO: We should really introduce a veterancy module at some point.
 -- XP gained for killing various unit types.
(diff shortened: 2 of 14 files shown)


duk3luk3:
Things to test:
- Does this affect any other units with special OnStopBeingBuilt logic
- Does this affect any other amphibious units
- Does this affect any other units with special OnLayerChange logic

duk3luk3:
I have updated this PR.

I am not completely happy with it, because it is still possible to fuck up the intel state, and the OnLayerChange event fired by the engine would fuck it up if I hadn't introduced a rather silly way of ignoring it.
The right way to do it would be to just make the intel state unfuckuppable, which is exactly what I did in my original PR for this with minimal effort, but @Sheeo didn't like that.

lua/defaultunits.lua
```diff
@@ -1428,6 +1428,14 @@ MobileUnit = Class(Unit) {
     CreateCaptureEffects = function( self, target )
         EffectUtil.PlayCaptureEffects( self, target, self:GetBlueprint().General.BuildBones.BuildEffectBones or {0,}, self.CaptureEffectsBag )
     end,
+
+    -- Units with layer change effects (amphibious units like Megalith) need
+    -- those changes applied when build ends, so we need to trigger the
+    -- layer change event
+    OnStopBeingBuilt = function(self,builder,layer)
+       Unit.OnStopBeingBuilt(self,builder,layer)
+       self:OnLayerChange(layer, 'JustBuilt')
```
Sheeo: I'm not fond of this.

'JustBuilt' isn't a member of the valid set of game layers. Isn't there a better way we can do this?

lua/sim/Unit.lua
```diff
@@ -2186,75 +2205,80 @@ Unit = Class(moho.unit_methods) {
...
+    -- (enhancement) that is not present, the intel requiring energy and energy being stalled, etc.
+    -- The intel is turned on using the EnableIntel engine call if all disablers are removed.
+    -- As an optimisation, EnableIntel and DisableIntel are only called when going from one disabler
+    -- present to zero, and when going from zero disablers to one.
 
-    DisableUnitIntel = function(self, intel)
+    DisableUnitIntel = function(self, disabler, intel)
         local intDisabled = false
-        if not self.IntelDisables then return end
+        assert(self.IntelDisables, 'DisableUnitIntel called before IntelDisables field initialized.')
```
Sheeo: Since IntelDisables is initialized in OnPreCreate, this assertion is redundant

lua/sim/Unit.lua
```diff
@@ -2186,75 +2205,80 @@ Unit = Class(moho.unit_methods) {
...
+        assert(disabler and disabler ~= "", 'Cannot track Intel disablers without disabler name')
```
Sheeo: In line with our discussion on slack, this precondition checking isn't necessary, since we control all the callsites for DisableUnitIntel.

lua/sim/Unit.lua
```diff
@@ -113,6 +114,20 @@ Unit = Class(moho.unit_methods) {
             self.Trash = TrashBag()
         end
 
+        self.IntelDisables = {
```
Sheeo: I think this could be even cleaner using ckitching's [setutils](https://github.com/FAForever/fa/blob/develop/lua/system/setutils.lua).

Instead of having tables and manually juggling the `n` value, for each intel type, keep a set of currently active disablers.

This simplifies the datastructure to:

    {
        Radar = {'Energy'=true, 'IntelToggle'=true},
        ...
    }

For eg a unit currently having radar disabled by both the toggle button and energy stall.

It also enables us to simplify all the callsites that disable unit intel to remove the disabler from all sets by name, instead of several function calls with almost identical parameters.

I imagine something like:

    function DisableUnitIntel(disabler)
       for inteltype, disablers in self.IntelDisablers do
           self.IntelDisablers[inteltype] = setutils.Subtract(disablers, {disabler})
       end
    end

duk3luk3: Yes. I had no idea we had setutils. This can be done so much better with a proper set class. ermahgerd.

lua/defaultunits.lua
ChrisKitching: Can't you just call it with "old" and "new" both being the layer it is currently on?

duk3luk3:
Updated PR to use setutils.

duk3luk3:
It's pretty now.

duk3luk3:
Fixed a disabler category screwup that Sheeo noticed.

lua/sim/Unit.lua
```diff
@@ -486,30 +501,30 @@ Unit = Class(moho.unit_methods) {
         elseif bit == 2 then --Jamming toggle
             self:StopUnitAmbientSound( 'ActiveLoop' )
             self:SetMaintenanceConsumptionInactive()
-            self:DisableUnitIntel('Jammer')
```
Sheeo: Are you sure you want to be using the same 'Toggle' string for each different toggle? 

The entire point of using this set based datastructure is that we track what activity disabled something, instead of just using a semaphore type.

With this, setting bit 2 (Clearing jamming), followed by setting bit 3 (clearing all intel) and then clearing bit 2 leads to jamming being active, even though the global toggle is still set.

You sure this is intended?

duk3luk3: The behaviour is not changed compared to what it was before. I don't want to change the behaviour in this PR.

Sheeo: Yes, it has changed. Previously jamming would be inactive in the situation described before.

duk3luk3: I'll check this as soon as I can.

duk3luk3: You're right, so off to redo all the OnScriptBitClear/Set stuff. How do we go about this?

* Are there are any units that need stealth toggled from the intel toggle?
* Why do we have this in the Unit class anyway? Shouldn't this be in a subclass? Do we have to start using multiple inheritance by defining a StealthUnit class with the appropriate OnScriptBitClear/Set functions that can be used as a mixin?

ChrisKitching: > Why do we have this in the Unit class anyway? Shouldn't this be in a subclass? Do we have to start using multiple inheritance by defining a StealthUnit class with the appropriate OnScriptBitClear/Set functions that can be used as a mixin?


Yes.


There's tons of stuff in Unit that shouldn't be there, quite a lot of which has already been moved out (notably the stuff for transports and the stuff for ACUs). Still tons to do. Go nuts.

duk3luk3:
Updated to add bit indication to every `Toggle` disabler.

Sheeo:
Merged with 81feb57104b5e75a61fc182153d721786ed9aa6a

PR #880, opened 2015-08-23
Fix #874

### LINKED ISSUE #874: We broke the balance display in lobby
Whoever did that is bad and should feel bad.

duk3luk3:
A parameter got dropped while redoing the lobby localization.

This PR fixes that and makes StreamText more robust against future occurrences of this.


ChrisKitching:
Merged fd42f23

Good spot, sorry for the delay, too many trains!

Issue #878, opened 2015-08-23
Seraphim Nuke and self damage


bamboofats:
In 3641, the Yolona Oss self damage was fixed ref https://github.com/FAForever/fa/issues/605
https://github.com/FAForever/fa/commit/81ff6c5ec21ad29d91e930ba190c80df7f30264e

However, the basic seraphim nuke launcher still does not damage itself. The changes in the commit should be applied to this nuke controller as well
effects/Entities/InainoEffectController01/InainoEffectController01_script.lua

Sheeo:
Please do provide a pull request :)

bamboofats:
This should be simple enough!

PR #897, opened 2015-08-30
Make Seraphim T3 Nuke damage itself

bamboofats:
https://github.com/FAForever/fa/issues/878


ChrisKitching:
Well, that was a hilarious thing for GPG to miss.

Varaxis:
I thought it was Seraphim faction diversity, since their T4 nuke also did the same before you guys nixed that.

ChrisKitching:
It's pretty unhelpful faction diversity, because it doesn't protect anything except the launcher. Your base is flattened, but at least the nuke launcher is okay!

Varaxis:
Bases are pretty big, and exp assaults that can't be stopped in any other manner makes it worth it. ACU still may have ras and eng suite, so a nuke and half a base still useful than the enemy wiping out the base and moving on to another. Easier to aim the nuke if they're right there trying to kill you.

Issue #756, opened 2015-07-13
When ACU dies, check if the damage that killed it was friendly


Zockyzock:
If yes, report game ID and timestamp to the moderator team or create a list visible by them. 

If the amount of false-reports is too high, exclude damage from ACU explosions and/or other sources that lead to non-abusive friendly fire kills.

aeoncleanse:
We could have such a detection pop up a dialog along the lines of... "You have been killed by friendly fire.  The deliberate killing of team-mates is against FAF rules.  If you feel your death was deliberate or unsportsmanlike,  press the report button below." and provide a report button which notifies mods,  and an exit button.

VoRGoR:
I'm all for this.

Anything in-game that can raise awareness of the fact that there is moderation and rules also gets a plus from me. Added to that an easy way (click) of actually doing it, removes the problem of people being too lazy.

Zockyzock:
To be effective, it also needs to check ALL friendly fire damage, on mex, pgs, t4 in construction..

aeoncleanse:
This issue has pretty much been addressed with the team-kill detection and reporting system I think.

PR #900, opened 2015-09-01
Teamkill warn

quark036:
This is the first half of the answer to #756 
It checks if the acu was teamkilled, creates a dialog to ask the player if they want to report it, and then uses GPGNetSend to send the information. However, I was not able to figure out how relayserver.py takes that information and deals with it, so I don't have the part about sending the info to a mod. If someone could help me figure that out, or add me to Slack so I can talk to you about it, that would be great


ChrisKitching:
You need to rebase this onto current develop and force push the branch. Please eliminate the merge commits while you're at it.

quark036:
I think I did it right in #903

Issue #893, opened 2015-08-28
Atlantis AA friendly fire intended?


Varaxis:
The AoE on Atlantis anti-air is indiscriminate, damaging everything it hits, included allied aircraft and even the Atlantis's owner's air force.

Other AA units checked for friendly fire, all coming up negative:
Cruisers, inc Sera's AoE
Naval carriers
SAMs, inc Sera T3 AA Turret
Static flak and hover flak

Easy way to test is with czars set to hold fire (copy-paste during pause to get them in position): 
![Example of Atlantis friendly fire testing](http://imgur.com/jO40CUi.jpg)

duk3luk3:
Is this intended? @Zockyzock @ppipp

Cheyans:
AA behavior should be consistent across the board

Zockyzock:
Yes, if they should deal ff or not is a more difficult question, but making it consistent to begin with is good.

PR #901, opened 2015-09-02
Remove friendly fire from atlantis AA

quark036:
Easy fix for #893


units/UES0401/UES0401_unit.bp
```diff
@@ -459,6 +459,7 @@ UnitBlueprint {
             CannotAttackGround = true,
             CollideFriendly = false,
             Damage = 280,
+			DamageFriendly = false,
```
Sheeo: This line is indented with tabs.

Sheeo:
Apologies for the late reply. Please rebase the tabs away and I think we're good to go :)

quark036:
I think I deleted my repo shortly after I created this, so I remade it in #907

Issue #545, opened 2015-04-10
Lobby chat doesnt allow selection


Exotic-Retard:
Please allow selecting text from the game lobby chat.

not sure if this applies to in game chat as well

Sheeo:
I believe the OP wants copy-paste support along with this

ChrisKitching:
This is, in theory, not quite impossible.

Crotalus:
Send string via "GPGNet" to FAF Lobby which manipulates windows clipboard? :)

Sheeo:
It's bat-shit-crazy, but I like it.

ChrisKitching:
That's not the insane part.

You see, gpg doesn't provide a multi line text control, nor does it seem to provide a selectable single line control.

As a result, we have to arrange a set of single line text controls to form a multiline control. Each row needs to have multiple textboxes, one coloured to indicate selection, and to move text between the boxes in response to mouse events to show selection.

It isn't *technically* impossible, but we're not doing it soon. :P

MrRowey:
@4z0t nice one for you if you could figure it out ;)

4z0t:
I can make a huge crutch that uses edit to display text, but I will try it anyway xd

Garanas:
I'd rather you spent your time on something else, but it is your time 😃 .

4z0t:
Studying for example 😌

PR #903, opened 2015-09-02
Detect teamkills and ask if they should be reported

quark036:
Here's the cleaned up PR of #900 

This is the first half of the answer to #756 
It checks if the acu was teamkilled, creates a dialog to ask the player if they want to report it, and then uses GPGNetSend to send the information. However, I was not able to figure out how relayserver.py takes that information and deals with it, so I don't have the part about sending the info to a mod. If someone could help me figure that out, or add me to Slack so I can talk to you about it, that would be great


Sheeo:
Your indentation settings appear to use tabs, this is inconsistent with the rest of the files. Please use 4 spaces for indentation :)

lua/ui/dialogs/teamkill.lua
```diff
@@ -0,0 +1,73 @@
...
+
+    dialog.OnClosed = function(self)
+        dialog = false
+        if reportToMod:IsChecked() then
+			local armiesInfo = GetArmiesTable()
+		local focusArmy = armiesInfo.focusArmy
+			local victimName = armiesInfo.armiesTable[teamkillTable[3]].nickname
+			local killerName = armiesInfo.armiesTable[teamkillTable[2]].nickname
+			--this all won't work yet, need to figure out how to get player name etc.
+			WARN("Was teamkilled: " .. victimName)
```
Sheeo: I would prefer this logging to happen when the detection is made, inside the sim. This way teamkills always appear in the game log.

lua/defaultunits.lua
```diff
@@ -2007,6 +2007,16 @@ ACUUnit = Class(CommandUnit) {
             if instigatorBrain and not instigatorBrain:IsDefeated() then
                 instigatorBrain:AddArmyStat("FAFWin", 1)
             end
+			--if we are teamkilled
+			if IsAlly(self:GetArmy(), instigator:GetArmy()) then
+				WARN('Teamkill detected')
+				if not Sync.Teamkill then 
+					Sync.Teamkill = {} 
+				end
+				table.insert(Sync.Teamkill, GetGameTimeSeconds())
```
Sheeo: Since tables are used for representing arrays in lua anyway; it would be more readable to use a table of the form:

    { time = GetGameTimeSeconds(), instigator = instigator.GetArmy(), victim = self:GetArmy() }

Making the code that reads the values more readable.

lua/ui/dialogs/teamkill.lua
```diff
@@ -0,0 +1,73 @@
...
+			WARN("At time: " .. killTime)
+			WARN("Killed by: " .. killerName)
+			GpgNetSend('Teamkill Warning', string.format(" %s was teamkilled at time %d by %s", victimName, teamkillTable[1], killerName ))
```
Sheeo: The format of this sould just be `GpgNetSend('Teamkill', time, instigator, victim)`.

I'll handle the serverside integration over at https://github.com/FAForever/server/issues/95.

quark036: Ok, thanks for the help. If I force push to that branch it will update my
pr right?

Sheeo: Yes, that's right :)

lua/ui/dialogs/teamkill.lua
quark036: I can't get the nicknames from inside the sim because armiesInfo is UI side, but I can report the army number, and then Ill keep the reporting in teamkill with the nickname

quark036:
Ok, I changed all those things, and my testing showed it working fine

quark036:
Fixed it again :/

lua/ui/dialogs/teamkill.lua
```diff
@@ -0,0 +1,71 @@
...
+    local title = UIUtil.CreateText(dialogContent, "<LOC teamkill_0001>Teamkill Detected", 14, UIUtil.titleFont)
+    LayoutHelpers.AtTopIn(title, dialogContent, 5)
+    LayoutHelpers.AtHorizontalCenterIn(title, dialogContent)
+
+    local infoText0 = TextArea(dialogContent, 590, 20)
+    infoText0:SetText(LOC("<LOC teamkill_0002>You have been killed by friendly fire. The deliberate killing"))
+    LayoutHelpers.Below(infoText0, title)
+    LayoutHelpers.AtLeftIn(infoText0, dialogContent, 5)
+	
+	local infoText1 = TextArea(dialogContent, 590, 20)
```
Sheeo: I'm really not fond of pointing this out again, but it seems there's another tab character here.

I'm also not clear on why you're making this many textareas -- I haven't worked much with the UI system myself, perhaps @ChrisKitching or @CodingSquirrel could poke in with an opinion here?

quark036: I sincerely apologize for wasting your time with the tab characters. I went back through and checked every line I changed to make sure they were all good this time.

I have to make 3 text areas because there isn't a multiline text control, or at least not that I saw, so I preferred 3 text areas to a really wide box.

As to the localization, in the strings_db, it says not to edit it but to use the LocStrings.xls to edit the strings, but I couldn't find that anywhere. Do you have that, or should I edit the strings_db files, and if so, should I just use google translate?

Thanks for your time

quark036: Ah, I found it: in #545 ckitching said there is no multiline text control

ChrisKitching: > Ah, I found it: in #545 ckitching said there is no multiline text control

There is now, I wrote one, it's called TextArea (not to be confused with GPG's TextField, the builtin one).
https://github.com/FAForever/fa/blob/develop/lua/ui/controls/textarea.lua

ChrisKitching: (It's also horrifying and uses ItemList, but it's still many times better than using a big pile of TextFields)

ChrisKitching: And stop apologising so much already :P

quark036: Ok, I'll change it to textarea when I get home. What should I do with regards to localization?

ChrisKitching: Well, localise all your strings, use the LOC function as usual, put them into the bottom of every strings_db file in english and we'll get them translated in bulk later on.

quark036: Ok. Is 'using the LOC function as usual' what I already did with <LOC teamkill_002>?

Sheeo: @quark036 no problem at all with the tab characters :) My remark was more that I feel naggy when pointing it out again ;)

Regarding localization, it should be a single LOC string for each logical piece of text.

More than just referencing a string with `<LOC teamkill_002>` in it, what Chris is saying is that you need to add the strings to each strings_db.lua file in the LOC directory.

I'll do this if you're not clear on how to get it done, when merging.

quark036:
Ok, I used one textarea to do the multiline stuff, and I added the strings to the localization strings_db's. 
What's wrong with it next :D

ChrisKitching:
Looks pretty great code-wise. A screenshot of what it looks like would be nice (so we can whinge about your layouts ;)

It might be better to use two buttons instead of a checkbox?
You also probably want to override the Popup class' default behaviour for background clicking and pressing escape (both of which call Close() by default, though you probably want to suppress this, so the user can't accidentally dismiss the dialog without making a choice).

quark036:
image

Here is a screenshot.

ChrisKitching:
> Here is a screenshot. (with ninja edit after I didn't like it and changed it a bit)

Pretty good.

Going to really nitpick now:

Other dialogs in FAF have the content somewhat further from the border. Take a look at one of them and steal the right padding value.

The label on the checkbox strays too close to the OK button.

I really think two buttons (on opposite corners of the box) labelled "Punish" and "Forgive" would make it clearer.

quark036:
A) I stole the buffers from the desync dialogue
B) Ok fine I changed the buffers and the buttons

quark036:
Updated screenshot:

image

aeoncleanse:
Buttons and text should be 13 px in from the laser line, standard. Otherwise, looks good. Centre the text, possible?

quark036:
Nothing else is 13, looking in the other dialogues

quark036:
Also, you'd have to ask ckitching, he wrote text area, I don't see a way to center the text

CodingSquirrel:
I do actually think the checkbox might be a better way to go. Just having the two buttons like that it's very easy for someone to click the wrong one on accident or simply because they're not paying attention. Checkbox would likely cut down on false positives.

quark036:
Dammit

quark036:
Ok, you guys debate which you like, and I will change it at the end

ChrisKitching:
> I do actually think the checkbox might be a better way to go. Just having the two buttons like that it's very easy for someone to click the wrong one on accident or simply because they're not paying attention. Checkbox would likely cut down on false positives.

Having just one button is bad for the same reason, but in that case there's only one click target. People will just quickly click the dialog away without thinking about it. Combined with a default-on checkbox, it seems dangerous.
Plus, it looks cooler now, right? More "balanced"?

I don't think you can center text in TextFields. It's looking much better now, anyway.

I think the last thing you have left to do is squash together your commits (as we don't want the commits produced during code review to be part of the global history), and then we're all good :D

Apologies for being so nit-picky, but it's kinda my job ;)
I'll give sheeo another chance to whinge, and then we'll merge this.

quark036:
How do I squash together the commits?

CodingSquirrel:
Then default the checkbox to off. :D

quark036:
It currently defaults to off

ChrisKitching:
> How do I squash together the commits?

`git rebase -i` is your friend. 

Here's a really good tutorial on history rewriting with git, which is what you want here:
https://www.atlassian.com/git/tutorials/rewriting-history

I think we should probably keep the big shiny buttons, something something nice UI. :P As long as we make sure they're far enough apart it's a non-problem, I think.
Another idea to possibly think about is to disable the "punish" button for 5 seconds after the dialog comes up (display a countdown on it), so people *can't* insta-click.

... That might be a stupid idea.

quark036:
Thanks for the link, that was great. I feel like disabling the buttons will annoy some people and cause others to report it as a bug :/

I edited the history (mwahahaha I am become Big Brother), does it look good?

ChrisKitching:
Looks fine. Good work!

Sheeo:
@ChrisKitching we're not fully done here.

We need to ensure the dialog isn't shown when watching replays, and that the GpgNetSend function isn't called as well (Since that will error otherwise).

PR #904, opened 2015-09-04
Fix restofield buff not stacking

### LINKED ISSUE #899: Seraphim acu restoration field BUG
Seraphim ACU restoration field has a bug that it lowers hp regeneration of units in range which have more HP regen then 200. For example SACU upgraded with nano regen should have 265 hp regen stock, and when in range of acu restoration field it should be like ~450 hp/s regen, but in reality it is only less then 200 hp/s - the ACU effect overrides the nano effect. The situation is worse as now the restoration field affect allies.

This might ring the bell:
BuffType = 'COMMANDERAURA', Stacks = 'REPLACE',

quark036:
Change the Seraphim ACU restoration field buff from STACKS = 'REPLACE' to STACKS = 'ALWAYS'
Fixes #899


duk3luk3:
Going to put on the reviewer hat for this one, do you have any idea why it was set to REPLACE before?

quark036:
So that it wouldn't stack, probably :)

ChrisKitching:
> Going to put on the reviewer hat for this one, do you have any idea why it was set to REPLACE before?

The way the restoration field is implemented is a bit weird. It basically says "At regular intervals (I think 5s), find every unit in range and apply this buff to them".
I suspect this was done to prevent problems with it being applied twice and providing a double buff. If you get it to sync up tick-for-tick it should be okay, but please do test. You need to verify the expiry time of this buff matches the repeat period of the regen field, or you'll have periods of double-buffing.

quark036:
The duration of the buff is 5, and the thread fires the buff every 5 seconds, so I think it should be ok. (see lines 115 and 137) in xsl0001_script.lua

ChrisKitching:
> The duration of the buff is 5, and the thread fires the buff every 5 seconds, so I think it should be ok. (see lines 115 and 137) in xsl0001_script.lua

So there's the problem: if the times don't sync up with the simulation timesteps you'll get the occaisional 10th of a second (one tick) where it's not working.

You should make both use a number of ticks instead of a number of seconds, I think, to avoid this rounding error.

quark036:
In summary of the slack discussion: changing stacks = replace to stacks = always does seem to give it the correct regen rate, but that rate is not what is displayed, ticks vs seconds doesn't matter because waitseconds is aliased to ticks, it'd probably be too expensive to fork a thread for every unit in the area to check if their buff should be terminated, the way the regen percent and regen amount buffs stack is really weird and should probably be changed, the way buffs work is really weird and should probably be changed, testing this seems to yield wildly different results every time I try, and this is a really weird issue overall

duk3luk3:
> the way the regen percent and regen amount buffs stack is really weird and should probably be changed

I'm pretty sure they don't stack at all. Here is how I would fix it so they *do* stack: https://github.com/duk3luk3/fa/commit/d349fbe18e1b019a2e0d13621480983c34a4039e

(Oh yeah, and `BuffCalculate` ignores the buffName parameter entirely, oh joy)

Crotalus:
Big problem here is the RegenCeiling of 15 resp. 200, the field will actually debuff units having more regen than this :/

quark036:
Well, that was what I thought at the beginning, which could be fixed by changing STACKS = REPLACE to STACKS = ALWAYS. However, it turns out everything is insane.

quark036:
You need the regen ceiling so that units with already high regen don't get insanely high. You just need it to stack with rather than replace the default regen

quark036:
Also, #923 is linked to this, with everything being insane

aeoncleanse:
This won't fix the problem. It will simply allow two seraphim players to have their regen aura stack with each others units, if that's even allowed.

aeoncleanse:
Which, as it turns out, it is! @Zockyzock @ppipp  - Do we want to allow allied players to double-stack the regen aura?

aeoncleanse:
This is not on the table at the moment. Two ACU Regen fields of different levels stack properly (They don't) in 3652, and don't go infinite.

PR #905, opened 2015-09-05
Fix for the omni radius buff definition.

The-Balthazar:
Omni buff used to check the units base radar radius rather than its omni radius, leading to the omni suddenly becoming the same as a units radar as altered by the buffs.


ChrisKitching:
Paging @Zockyzock

aeoncleanse:
:+1:

Crotalus:
28416fe5f0c11a7d82cebb836f9c29348c06bf95

Issue #404, opened 2015-01-22
UEF TMD stops firing


Sheeo:
Replay ID 3013192

aeoncleanse:
For reference, this is the only change to the weapon in the mod.

UnitBlueprint {
    Merge = true,
	BlueprintId = "ueb4201",						#UEF TMD
    Weapon = {
        {
            TargetCheckInterval = 0.1,
            TrackingRadius = 1.1,
        },
    },
}

UEF TMD is the only one which is projectile-based, and the reason it is the worst is because it seriously overkills incoming threats, and can also miss. These changes cut the overkill and no-fire down a lot by forcing each one to target check every tick, meaning no more stupid pauses. The TrackingRadius reduction means that packed TMD will not overkill so much because they all locked onto one projectile when it was further out, then don't target check in time.

Neither of these can possibly be causing this issue.

aeoncleanse:
This is absolutely dead-end, there is NOTHING in the log at all, not even with full debugging turned on.

Sheeo:
Why do you expect something to be in the log?

aeoncleanse:
Because the log is my friend? ;) I can't even begin to imagine what might be causing this though, it's almost as if the TMD has some kind of ammo...

ChrisKitching:
> Because the log is my friend? ;) I can't even begin to imagine what might be causing this though, it's almost as if the TMD has some kind of ammo...

It's debugging time!

Since it's a pain in the backside to reproduce this bug, my suggestion is that you add excessively verbose loggging to the code controlling the TMD, reproduce the bug, and see what's different.
This should, at least, point you in the right direction.

aeoncleanse:
That's the thing, the code controlling the TMD is almost completely default. There's no modification other than some spinup FX which just... can't... be causing this. Not without throwing a tantrum in the log...

How do I get a replay to launch my git branch?

Sheeo:
Sorry, I never got back to you on this.

Launching replays using modified code requires a bit of hacking at the moment -- it's something I'd like to be able to do easily with the client update.

DeimosEvotec:
So, anybody still interrested in this?
I can reproduce this bug with nearly all units that shoot tacs in this replay id 3798200

aeoncleanse:
Interested, sure, just nobody is any more clued up on wtf could be causing this

Crotalus:
[...]

Issue #913, opened 2015-09-13
Weapons locked on out of range target


Crotalus:
If a weapon locks on a target and this target gets out of fire range (but still in tracking range) the weapon will be deadlocked until the target is outside tracking range.

For instance, had an Aeon T2 PD happily locking a unit right outside fire range while other units walked right up to it.

Weapon.AlwaysRecheckTarget doesn't work either, it's only triggered after each shot of the weapon.

aeoncleanse:
Really not sure there is anything we can do about this :/

PR #914, opened 2015-09-13
Prevent weapons from getting stuck at targets

Crotalus:
Seems like there isn't a way to prevent #913 from happening by just changing values in the blueprint.

This fix will prevent weapons from getting stuck indefinitely at a target out of range.

Normally the weapon goes over to RackSalvoFiringState immediately and the scan for a new target won't happen.


CodingSquirrel:
Does this unconditionally reset the target every 5 seconds? If so couldn't that produce unwanted target swaps even if it's still in range?

Crotalus:
As soon as the weapon can fire the reset won't happen.

But you're right, we need to explore the potential problems so we don't fuck up general targetting / micro, preferably fix #404 at the same time

Crotalus:
`release/3652`

PR #915, opened 2015-09-13
Don't try to flash player's icon if score panel is disabled

speed2CZ:
Fixes https://github.com/FAForever/fa-coop/issues/21
In coop, score panel is disabled and it was causing warnings in logs.
by Washy


Crotalus:
:+1: 

a1c955b838437dba359a8874e3ad817a02a885a5

PR #918, opened 2015-09-13
Disable resources sharing for AI in Coop

speed2CZ:
Fixes https://github.com/FAForever/fa-coop/issues/13


speed2CZ:
After some thinking I decided to be evil, keep resources sharing for enemy AI and handle this in the map script for friendly AI

PR #924, opened 2015-09-19
Bugfix/gunship in transport

### LINKED ISSUE #920: Invincibility Bug
Replay 3872051 created to show this.

Do the following:

(1)  Make a UEF T2 Gunship.
(2)  Load the T1/T2 unit you want invincible onto it.
(3)  Right click (not dock) the gunship into a CZAR or Atlantis.
(4)  Kill the CZAR/Atlantis

This can be done with up to transport capability of the CZAR/Atlantis.  Mobile shields function for the CZAR/Atlantis as well, though not perfectly (due to size, I think).

The unit(s) carried by the gunship(s) will be invincible and invisible after the CZAR/Atlantis dies  They have hitboxes, make sounds, etc., but they do not physically appear on the map.  They have all of their usual abilities but cannot be killed.

Crotalus:
Fixes #920

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1686,6 +1686,11 @@ BaseTransport = Class() {
     DetachCargo = function(self)
         local units = self:GetCargo()
         for k, v in units do
+            if EntityCategoryContains(categories.TRANSPORTATION, v) then
+                for k, u in self:GetCargo() do
+                    u:Kill()
+                end
+            end
             v:DetachFrom()
         end
     end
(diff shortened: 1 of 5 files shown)


lua/defaultunits.lua
```diff
@@ -1686,6 +1686,11 @@ BaseTransport = Class() {
     DetachCargo = function(self)
         local units = self:GetCargo()
         for k, v in units do
+            if EntityCategoryContains(categories.TRANSPORTATION, v) then
```
Sheeo: Won't this mean units can't 'accidentally' survive crash drops anymore?

Sheeo:
Wouldn't it be better to make sure that you can't dock a loaded transport into another transport?

Crotalus:
Good idea, if I only could find out a way to hook into the actual loading process. The only trigger I've found working so far is OnAddToStorage which is called after the units is loaded.

I guess we could order a specific unload order immediately at that time with the drawback that other load orders running simultaneously at the carrier will get cancelled.

Maybe call DetachFrom() alternative Kill() on the unit in the gunship cargo when the gunship is added to storage,

Crotalus:
`release/3652`

PR #925, opened 2015-09-19
Bugfix/unit underwater sounds

Crotalus:
There seems to be a sound bug with underwater units, they never play(non-ambient) sounds while in water.

By refactoring ambient sound entities to include all kind of unit sounds we bypass this bug


Crotalus:
dfb363dcdc6685bdf4daf8f94178a7da2b06ae8f .. 74bf33595c16cbbef4ac72b2db5eda8126aa2a11

PR #927, opened 2015-09-20
Feature/auto overcharge

### LINKED ISSUE #789: Automatic overcharge toggle
For the balance patch we would like to have an automatic overcharge toggle for sera SCUs, as well as for ACUs (for consistency). The idea is that you can right click on the overcharge icon to make overcharging automatic: it will overcharge as soon as you have enough energy. It's similar to how you can right click on the ACU TML icon to build TMLs automatically. If possible this would be turned ON by default for SCUs, and OFF by default for ACUs.

Crotalus:
Initial work with auto overcharge.

Right click on overcharge icon to enable

Known quirks with the UI, not possible to disable / enable auto overcharge while the button is disabled (between shots and / or low energy). 

Fixes #789


duk3luk3:
> Known quirks with the UI, not possible to disable / enable auto overcharge while the button is disabled (between shots and / or low energy). 

Why not disable the button only when energy storage is too low for OC to work at all, and queue up an OC command otherwise that fires when there is enough energy?

Sheeo:
Looks good overall. @Zockyzock and @ppipp should merge this into the balance branch.

Crotalus:
`release/3652`

irswat:
awesome

PR #929, opened 2015-09-20
Flash intel for enemy civilians at start

### LINKED ISSUE #567: Pulse 100% vision at the start of ladder matches or otherwise show civilians

Crotalus:
Rebased #568

Fixes #567


Crotalus:
Added lobby option to be able to control the reveal behavior. Also darkened the color for enemy civilians so you can differ them from neutral ones.

Crotalus:
`release/3652`

PR #930, opened 2015-09-20
Cleanup/aibrain

Crotalus:
General cleanup of aibrain.lua, moved score stuff into own file and also pulled out the AttackManager. 

Optimally all AI stuff should be made more modular so it's easier to extend / modify it, currently we have Sorian functions all over there place.

As a sidenote - maybe we should expand the score data so we get juicer information available to replay mods etc. There's a discussion about such a mod here:

http://forums.faforever.com/forums/viewtopic.php?f=41&t=10686


ChrisKitching:
Looks generally okay. I haven't dug too deep into AI stuff, but remarked a little on the things I've actually read before.

Crotalus:
6eb6976aa9025d770d461b5dce5e64213496fc60 .. 16fc950ac73e86a837254dae55abc346b9926879

PR #931, opened 2015-09-21
Feature/persist ferry

Crotalus:
Irritating when ferry beacons disappear when the last transport of a ferry order dies. By letting an invisible ferry helper unit be part of the ferry order we go around this problem.

This concept could also be used with persistent build orders so you can have them surviving even if the engineers die.

Still rough in the edges but needs discussion before wasting anymore time with it.

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -19,12 +19,37 @@ function DoCallback(name, data, units)
     end
 end
 
+function SecureUnits(units)
+    local secure = {}
+
+    for _, u in units do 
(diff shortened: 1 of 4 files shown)


duk3luk3:
Anything to remove tedious actions from the game. I can't test this right now, how did you implementing deleting a "phantom" ferry route?

Crotalus:
Right click a start of the route should delete it, at least in theory. Haven't tested it that much yet.

Still need to destroy the actual phantom unit after the route is removed though

duk3luk3:
Sounds good! :+1:

Crotalus:
`release/3652`

PR #932, opened 2015-09-21
Feature/transport lock

Crotalus:
With this feature, a locked unit won't get unloaded when ordering a transport to unload all units. Especially useful for mobile stealth or when you want to drop all except a few.

Right click a unit in a transport to lock it in place. Unlock by right clicking again or specifically select it and unload. Lock is ignored if you give unload order with only locked units in transport.

Need a texture background for locked units, maybe a red one?

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -19,12 +19,39 @@ function DoCallback(name, data, units)
     end
 end
 
+function SecureUnits(units)
+    local secure = {}
+    if units and type(units) ~= 'table' then
+        units = {units}
+    end
+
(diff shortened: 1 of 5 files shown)


duk3luk3:
> Need a texture background for locked units, maybe a red one?

How about just a in a corner and shading the icon? Otherwise I really like this idea. :+1:

Sheeo:
We already have a corollary feature with the ability to selectively unload units. Perhaps this should be made more obvious as well? I suspect it's a feature that's hard to discover :)

Crotalus:
`release/3652`

PR #933, opened 2015-09-22
Let the ahwassa bomb deal damage on its way down

Crotalus:
@Zockyzock 
This should be seen as a suggestion for an alternative fix to the latest Ahwassa balance changes.

With this change it won't be possible to block off the bomb with ASF, the projectile will fly straight through and deal some damage on the asf flying into it. The radius is small though so no death of the whole cloud.


Zockyzock:
Sounds nice

ChrisKitching:
Seems a pretty good idea.

PR #934, opened 2015-09-22
Add UI support for multiple upgrade at the same time

Crotalus:
Useful for cybran shields, radar, hives etc.

Maybe need to add a tech-level icon for those that are near identical (mexes for instance).

### DIFF
--- a/lua/system/setutils.lua
+++ b/lua/system/setutils.lua
@@ -2,21 +2,6 @@
 --
 -- Utility functions for working with this type of set.
 
--- Transforms a table into a set
-function Set(t)
-    local set = {}
-    local ins = {}
-
-    for k, v in t do
-        if not ins[v] then
-            table.insert(set, v)
-            ins[v] = true
-        end
-    end
-
-    return set
-end
-
 --- Returns a ∪ b as a new set
 function Union(a, b)
     local u = {}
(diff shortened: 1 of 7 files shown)


lua/system/setutils.lua
```diff
@@ -2,6 +2,21 @@
 --
 -- Utility functions for working with this type of set.
 
+-- Transforms a table into a set
+function Set(t)
```
Sheeo: I would personally prefer it if we don't allow sets which map values to anything but `true`. This is confusing at best and would lead to inconsistent behavior at worst if such sets are used with the other functions in setutils.

lua/system/utils.lua
```diff
@@ -427,6 +427,10 @@ function table.filter(t, filterFunc)
     return newTable
 end
 
+function table.unique(t)
+    return import('/lua/system/setutils.lua').Set(t)
```
Sheeo: I'd suggest to implement table.unique directly here instead.

lua/system/setutils.lua
ChrisKitching: You could, however, imagine a "fromList" function that uses the elements of a list as the keyset for a table of mappings to true.
... Didn't I write one of those? It may never have got pushed.

Crotalus:
33efb270bb9ebdcc5b386871209f231a815c7917 .. d18efefd7167c6774637408314f8272e359fdc57

duk3luk3:
Why does this not work for factories, but still changes the factory upgrade icons to have the arrow in them? It makes for very confusing UI, see #1057

PR #935, opened 2015-09-22
Hotfix for firebeetles

aeoncleanse:
Further simplified the way beetles work and walked through it from the start.

### DIFF
--- a/lua/selfdestruct.lua
+++ b/lua/selfdestruct.lua
@@ -1,8 +1,10 @@
 function ToggleSelfDestruct(data)
-    -- supress self destruct in tutorial missions as they screw up the mission end
+
+    -- Suppress self destruct in tutorial missions as they screw up the mission end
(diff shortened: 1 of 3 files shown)


ChrisKitching:
In what sense is this a "hotfix"? What problem is it fixing? Do you even know what this word means? :P

aeoncleanse:
Hehe. It's a hotfix because when pip and I were talking, he mentioned beetles and hotfixing in the same sentence. The problem it's fixing is stupid complexity allowing it to blow at the wrong times in some situations, which is why we had to revert.

Whatever :8ball:

duk3luk3:
Why is this closed? Was it merged?

aeoncleanse:
Crotalus reworked it

irswat:
fire beetles were working correctly before the last update if I recall correctly. With the faf beta that is up now the firebeetles do not explode at all. The detonate button pauses them and attack they just surround the unit and sit there.

PR #936, opened 2015-09-23
merging the selection sort mod

Bra1nwash3d:
http://forums.faforever.com/forums/viewtopic.php?f=41&t=10859

### DIFF
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -1708,8 +1708,10 @@ function FormatData(unitData, type)
         end
         CreateExtraControls('construction')
         SetSecondaryDisplay('buildQueue')
+
     elseif type == 'selection' then
         local function SortFunc(unit1, unit2)
+            LOG('x')
             if unit1.id >= unit2.id then
                 return false
             else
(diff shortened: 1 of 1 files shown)


lua/ui/game/construction.lua
```diff
@@ -1708,8 +1708,10 @@ function FormatData(unitData, type)
         end
         CreateExtraControls('construction')
         SetSecondaryDisplay('buildQueue')
+
     elseif type == 'selection' then
         local function SortFunc(unit1, unit2)
+            LOG('x')
```
CodingSquirrel: Leftover debug logging?

Bra1nwash3d:
yea i left that in, but noticed that the sortfunc never gets called because the table uses the unitId as index, whereas the table.sort function only sorts tables with indexes incrementally starting at 1
this function is ignored anyway

lua/ui/game/construction.lua
Bra1nwash3d: yea i left that in, but noticed that the sortfunc never gets called because the table uses the unitId as index, whereas the table.sort function only sorts tables with indexes incrementally starting at 1
this function is ignored anyway

quark036:
Washy just uploaded another version, I don't know if what you merged had the updated code
http://forums.faforever.com/forums/viewtopic.php?f=41&t=10859

PR #937, opened 2015-09-23
Increase size of some units to reduce hitting the floor thanks to low-mounted weapons

aeoncleanse:
Adjusts Rhino, Othuum, and both Snipers. Adjustments minor, so after a couple of minutes you won't even notice the difference.

Othuum and Aeon sniper required hitbox changes. Aeon one is very small, just to keep the bones contained. Othuum's hitbox is stupidly small anyway, this makes it fit the new model nicely.


Crotalus:
ee0c1a7f1d7ef7b92beed79681138c36de5ab63b .. db7ed35a2f3ffe11d3b14c9b64aa2ff73a9b04ff

PR #942, opened 2015-09-24
Add support for auto-rotating structure against enemy

Crotalus:
When building a defense structure its rotation will be towards the closest unit, if no unit is found it will turn towards the mid as default


quark036:
We could rotate it again if it killed its target and doesn't have a new one

Crotalus:
`release/3652`

PR #943, opened 2015-09-27
Repair wreckage / unit -> rebuild it

Crotalus:
* Order repair on a wreckage to rebuild it.

* A structure which is repaired and dies will automatically, if possible, be rebuilt by the repairers. A normal right-click order (assist) on a damaged structure won't cause rebuild though.

~~Planning to add support for repair ground also, which would mean to repair / rebuild anything within range (GuardScanRadius) of the ordered units.~~

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -19,7 +19,24 @@ function DoCallback(name, data, units)
     end
 end
 
+function SecureUnits(units)
+    local secure = {}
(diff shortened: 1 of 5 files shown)


Crotalus:
35c59ac1722ab9fe086b02fa0f810ba2cd20b8e6 .. bedda4301884adc5642e8caaab27390a0f0fbc7d

PR #945, opened 2015-10-01
Refactor/modules

Crotalus:
Get rid of /modules/ and /lua/spreadattack.lua


Sheeo:
:+1:

Sheeo:
Keymap ought to be in `/lua/ui` however.

CodingSquirrel:
Maybe I'll revive #858 and add my changes on top of this. And while I'm at it possibly fix the ShadowOrders to take into consideration if you change or remove an order before you hit splitattack.

Issue #262, opened 2014-12-29
Make the stop button cancel all units in the build queue except the one currently in production


julesjacobs:
Double clicking on the stop button will cancel the entire queue including the unit currently in production. Usually that is a mistake, so it should not be the single click behavior :)

PR #951, opened 2015-10-03
Feature/soft stop

Crotalus:
New keymap 'soft_stop' which can be mapped by the user to whatever key wanted. 

With soft_stop, all factory orders except the current will get deleted. If you want to delete all orders press soft_stop again or use normal stop command to clear the queue instantly

Also added special stop-handling for silo units to prevent accidental cancellation when building missiles in SML, strat subs etc. Not sure if this also should be counted as soft_stop or always default. Shift-Stop will cancel all in this case.


Crotalus:
Seems like #262 contains some discussions about this. Also:
http://forums.faforever.com/forums/viewtopic.php?f=42&t=9122

Crotalus:
Maybe should scrap the option and instead make it as an bindable, that way people can choose whatever behavior they want themselves

Crotalus:
Added soft stop support to silo units to prevent accidental stop of missile building

CodingSquirrel:
Based on the code and the fact that the phim battleship has the category SILO it probably does, but does the soft stop also work for it and other nuke ships as well? Nothing worse than having a battleship building a nuke and pressing stop to make it stop moving.

Crotalus:
Yeah should work for all units which has the SILO category defined

Crotalus:
`release/3652`

PR #952, opened 2015-10-03
Added Russian Localization and updated lobby strings

Exotic-Retard:
GLORIOUS russian localization is here! i also added localization tags to the lobby strings and put them in all the other loc files for easy editing.

for some reason the russian loc file has ~1k more strings than the faf ones, dunno maybe its just that cool or sth : D

i haven't completely translated the russian one, cos some things i don't know what they mean in english, and the other loc files are left completely the same, just with the new strings in them.

hopeing this is done right : |


Sheeo:
Travis reports syntax errors in these files:

```luac: ./loc/FR/strings_db.lua:7566: unfinished string near '"'
luac: ./loc/US/strings_db.lua:6753: unfinished string near '"'
luac: ./loc/IT/strings_db.lua:7568: unfinished string near '"'
luac: ./loc/ES/strings_db.lua:7566: unfinished string near '"'```

File: loc/ES/strings_db.lua [hunk 1]
```diff
@@ -7562,3 +7562,187 @@ teamkill_0002 = "You have been killed by friendly fire. The deliberate killing o
...
+lobui_0730="Score is disabled"
+lobui_0740="Share Conditions"
+lobui_0741="Kill all the units you shared to your allies and send back the units your allies shared with you when you die"
+lobui_0742="Full Share"
+lobui_0743="You can give units to your allies and they will not be destroyed when you die"
+lobui_0744="Share Until Death"
+lobui_0745="All the units you gave to your allies will be destroyed when you die"
+aisettings_0001="AIx Cheat Multiplier"
+aisettings_0002="Set the cheat multiplier for the cheating AIs."
+aisettings_0003="Cheat multiplier of 1.0"
```
Sheeo: You should use a single string for this, for instance like in:

`uimod_0027 = "%d game mods and %d UI mods activated"`

We then need to tweak the code that uses this string and substitute in the cheat multiplier literal.

Sheeo:
@Exotic-Retard if you fix the syntax errors I'll make the changes to the strings so they aren't duplicated for each number

Crotalus:
We should probably try to get this in `release/3652`?

Crotalus:
Merged here:
5e4052e7f6046131b5a4563190c393f0b211932b..babb91467bc23a512e2d394ebf1fa2ee9295a734

Added support for dynamic lobby options strings so key can be in the text

PR #953, opened 2015-10-04
Economy Numbers Improvement

aeoncleanse:
A few improvements. Rounds numbers to nearest integer to cut down on
display flickering. Performs fewer math functions. Rearranged the file a
bit for readability.

The big improvement: Attempts to prevent reclaim income from appearing
in the generated income column.

### DIFF
--- a/lua/ui/game/economy.lua
+++ b/lua/ui/game/economy.lua
@@ -353,63 +353,90 @@ function ConfigureBeatFunction()
 
         -- The quantity of the appropriate resource that had been reclaimed at the end of the last
         -- tick (captured into the returned closure).
-        local lastReclaim = 0
+        local lastReclaimTotal = 0
+        local lastReclaimRate = 0
 
         -- Finally, glue all the bits together into a a resource-update function.
         return function()
             local econData = GetEconomyTotals()
             local simFrequency = GetSimTicksPerSecond()
-
+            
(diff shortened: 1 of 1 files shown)


CodingSquirrel:
I did a few minutes of poking around but couldn't find where the income numbers that the score uses are populated. Do you know if this will affect that? Hopefully we can stop that from showing reclaim numbers too.

aeoncleanse:
I poked around too, same result. The Score display was the next thing I was gonna make sure of. As it is, I've no idea if it affects it or not. What I will say it that it cannot affect it negatively.

Varaxis:
Income doesn't affect score display. It's only spending, acu death/kills, nuclear events, that seem to affect it.

aeoncleanse:
Sorry for the misunderstanding Varaxis, should have been clearer: We're talking about the mass counter for each player which shows up in the score area when in neutral. Casters keep an eye on it to see how the players are doing, and reclaim really messes around with it.

Varaxis:
Ah. No, it's my misunderstanding. Noticed that it was tackled in the eco window, sort of. I see what you mean. Now you got me interested. Noticed someone did a mod that has played around with that window, called ally overflow.

Crotalus:
e01ebc73ef941d59bc8393032af28be8be747dc7

PR #954, opened 2015-10-05
Yield smarter in WaitSeconds on UI side

Crotalus:
Instead of always yielding 1 frame at a time, decide the amount of frames to yield by using the average fps number. With lots  of UI co-routines running this should reduce the context switching substantially

### DIFF
--- a/lua/ui/game/gamemain.lua
+++ b/lua/ui/game/gamemain.lua
@@ -35,18 +35,6 @@ local sendChat = import('/lua/ui/game/chat.lua').ReceiveChatFromSim
 local oldData = {}
 local lastObserving
 
-function UpdateAverageFPS()
-    local now
-    local wait_frames
-
-    while true do
-        now = CurrentTime()
-        wait_frames = AvgFPS * 10
-        WaitFrames(wait_frames)
-        AvgFPS = math.max(10, math.min(200, math.ceil(wait_frames / (CurrentTime() - now))))
-    end
-end
-
 -- Hotbuild stuff
 modifiersKeys = {}
 -- Adding modifiers shorcuts on the fly.
@@ -249,8 +237,6 @@ function CreateUI(isReplay)
     if options.gui_render_enemy_lifebars == 1 or options.gui_render_custom_names == 0 then
         import('/modules/console_commands.lua').Init()
     end
-
-    ForkThread(UpdateAverageFPS)
 end
 
 local provider = false
--- a/lua/ui/game/gamemain.lua
+++ b/lua/ui/game/gamemain.lua
@@ -35,6 +35,18 @@ local sendChat = import('/lua/ui/game/chat.lua').ReceiveChatFromSim
 local oldData = {}
 local lastObserving
 
+function UpdateAverageFPS()
+    local now
+    local wait_frames
+
+    while true do
+        now = CurrentTime()
+        wait_frames = AvgFPS * 10
+        WaitFrames(wait_frames)
+        AvgFPS = math.max(10, math.min(200, math.ceil(wait_frames / (CurrentTime() - now))))
+    end
+end
+
 -- Hotbuild stuff
 modifiersKeys = {}
 -- Adding modifiers shorcuts on the fly.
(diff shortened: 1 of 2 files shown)


duk3luk3:
Aren't there places in the code that rely on WaitSeconds waiting for a precise number of frames? Unless I'm confusing it with a different function, the semantics of `WaitSeconds(n)` are currently "Wait exactly n*10 frames", I remember this being used for sim-relevant timings in some code I touched.

Is `CurrentTime()` based on clock time or frames? If it's based on clock time, don't use it - we want `WaitSeconds` to be frame-based; if it's based on frames, well, just use frames.

Why can't Waitseconds just be

    function WaitSeconds(n)
      WaitFrames(n*10)
    end
?

Crotalus:
WaitSeconds on sim-side = 10*WaitTicks(1)
WaitSeconds on ui-side = Wait 1 frame, check if we've waited enough - tied to the fps

So WaitSeconds on UI side isn't tied to gameticks at all, if you want something running every gametick you should use AddBeatFunction().

The problem with current UI-WaitSeconds() is that a single WaitSeconds(60) will wake up ~6000 times  (@ 100 fps) to check if 60 seconds has elapsed. 

I can see a problem though with this solution not taking into account sudden frame drop which would skew the elapsed wait time some. Maybe should make it pessimistic and only wait some fraction of the total estimated frames.

duk3luk3:
If it's just for UI update intervals, i.e. totally non-critical to simulation and user experience except that something in the UI might not update for 5 seconds instead of 1 during sudden graphics fps drop, then this is totally fine.

> I can see a problem though with this solution not taking into account sudden frame drop which would skew the elapsed wait time some. Maybe should make it pessimistic and only wait some fraction of the total estimated frames.

```
function WaitSeconds(n, pessimistic)
    pessimistic = pessimistic and 0.1 or 1
    local goal = CurrentTime() + n
    local wait_frames
    repeat
        wait_frames = math.ceil(math.max(1, AvgFPS*0.1, n * AvgFPS * pessimistic))
        WaitFrames(wait_frames)
        n = goal - CurrentTime()
    until n < 0.01
end
```

Yeah that would probably be useful for things that users are likely to notice UI freezes on.

ChrisKitching:
I would suggest just not bothering with average FPS at all: that extra thread to track average FPS is a bit of an expense.
UI-side WaitSeconds is never time-critical, it's just used for pretty UI things. plus or minus 2-ish seconds is basically okay. Just take an instantaneous FPS calculation in WaitSeconds, use it to work out how many frames you need to wait, and do that?

Crotalus:
Yeah sounds reasonable, like that?

Crotalus:
By having it outside the function we get a "free" FPS estimate the next time WaitSeconds is called, otherwise we would have to recalculate it again.

Crotalus:
`release/3652`

TAPAKAH4Life:
The stated problem this is supposed to solve is that at 100 fps the original Waitseconds() will wait x2 as long as at 50 fps. A strange problem to attempt to solve.
The new function calculates its value based on FPS (which varies dramatically during gameplay). Do any of you know with certainty that Waitseconds() is not used to calculate a parameter which will be calculated once and used throughout gameplay, without the possiblility to be updated by a freshly updated Waitseconds(). 
Any statements to assauge my concerns are appreciated.
function WaitSeconds(n)
   local start = CurrentTime()
   local elapsed_frames = 0
   local elapsed_time = 0
   local wait_frames    repeat
       wait_frames = math.ceil(math.max(1, AvgFPS*0.1, n * AvgFPS))
       WaitFrames(wait_frames)
       elapsed_frames = elapsed_frames + wait_frames
       elapsed_time = CurrentTime() - start
   until elapsed_time >= n    if elapsed_time >= 3 then
       AvgFPS = math.max(10, math.min(200, math.ceil(elapsed_frames / elapsed_time)))
   end
end

Crotalus:
Read the original post again, with this patch a call to WaitSeconds(1) will wake up / context switch to the coroutine ~1-2 time(s) instead of 100 times (with someone running at 100fps). This function is on the UI side and all uses of these are visible in the lua code and used for regular updates of the UI. The other way the UI updates is through OnBeat()-functions which is run every game tick.

TAPAKAH4Life:
Thank you for the clarification

Issue #949, opened 2015-10-02
Cybran drones leave annoying wreck


aeoncleanse:
Spawned from #940.

It's not a bug, but it's poor design. Making the drones fall from the sky, crash, and burn, was a great visual idea. An unintended consequence was leaving a wreck. Thought not to be a problem, the wreck is 0 mass but gets in the way of building things, so it takes up time while the engineers have to reclaim it for 0 value. 

#941 Addresses this whole thing. It keeps the crash and burn, just transmits overkill damage through the code to ensure that a wreck is not spawned. It's clean, keeps the visuals, and tidies up the annoyance of having these wrecks all over the shop. I request it be re-opened and merged.

speed2CZ:
Either the wreck should have some value so its worth the time you need to spend reclaiming or no wreck at all.
First option is a problem that the value would have to be something worth and that cant be put there, so option 2 is way to go. Make the game less annoying.

aeoncleanse:
Exactly, the wreck has 0 value. We're talking about the wreck from the tiny drones every Cybran Engie spawns when they build stuff. Have them with a value and you're generating free mass! I know some Cybran players would love this, but... hehehehe

Crotalus:
Yes, not good that the wrecks blocks when building new structures, especially when they always lie in the vicinity of a current base.

I guess it would be possible to make the wrecks non-collidable so units would ignore them when building. Another option would be to let them expire very quickly

duk3luk3:
:+1: to keeping the visuals but removing the wreck.

PR #956, opened 2015-10-09
Prevent Cybran drones leaving wrecks

aeoncleanse:
#949 

There's some support there, and aeolus has been positive too. Let's get this in. They still die, crash, explode, all the visual glitter, they just don't leave a wreck.

### DIFF
--- a/units/URA0001/URA0001_script.lua
+++ b/units/URA0001/URA0001_script.lua
@@ -42,6 +42,12 @@ URA0001 = Class(CAirUnit) {
     OnKilled = function(self)
         self:StopBuildingEffects()
     end,
+    
+    -- Don't make wreckage
+    CreateWreckage = function (self, overkillRatio)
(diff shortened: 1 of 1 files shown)


Crotalus:
Hmmm, maybe we should add a check to general wreckage code never to spawn a wreck that has mass value of 0?

Sheeo:
I'm sure that would make a lot of players happy :)

Though some maps are built with wrecks of 0 value. Might be worth looking into making this either not reclaimable or remove it.

aeoncleanse:
Or just do what I've done... Use the existing overkill code to prevent a pointless and unique annoyance.

ChrisKitching:
You could just make them leave nonzero-valued wrecks.

Subtract a bit from the engineer wreck, add it to the drone wrecks, job done.

Crotalus:
`release/3652`

PR #958, opened 2015-10-12
Cybran beams

Crotalus:
With the changes of hitboxes and intel for beams we can restore the old beam effect of the PD and t2 tank, for nostalgic reason maybe?


aeoncleanse:
We could...  I kinda prefer the pew-pew-pew...  :/

speed2CZ:
Yes, those were superior to what we have now

Zockyzock:
Would like to have them back, but i think there was another reason for them, @ppipp knows more about it

Crotalus:
`release/3652`

PR #960, opened 2015-10-13
Refactor/unit

Crotalus:
Some refactoring of Unit to the specific unit type in defaultunits.lua instead. 

Later we should also probably split up defaultunits.lua into seperate files and only keep references in defaultunits.lua for BC.

Added some footprints to Megalith / Spider as well


ChrisKitching:
> Later we should also probably split up defaultunits.lua into seperate files and only keep references in defaultunits.lua for BC.

I have an old, outdated PR that did some of that lying around somewhere. You might want to steal partr of it. (never got merged because it was right before 3641 shipped and then other things ate all my time)

Crotalus:
Did some other of the commented changes as well, we probably need to refactor the whole movement effects system someday. All the bags etc seem a bit messy for what it does.

ChrisKitching:
It's awful, yes.

Sheeo:
Looks like this hasn't progressed for a while.

PR #962, opened 2015-10-18
Cleanup/victory

Crotalus:
* Removed unused victory callback hack
* Got rid of FAFWin / FAFLoss
* Calculate score by using `Enemies_Commanders_Destroyed` instead of FAFWin hack


Crotalus:
`release/3652`

PR #963, opened 2015-10-18
Firebeetles bugfix

Crotalus:
Tuned @aeoncleanse fixes for fire beetles, should work for insta-ctrl-k now


Crotalus:
`release/3652`

PR #964, opened 2015-10-18
Tune reclaim values of tree split

Crotalus:
Last time tree split was tuned it was nerfed too much, they didn't take into account the duration to reclaim each of the single trees (engineer on patrol has a min limit of 1 sec per tree or so).

This is obvious if you watch replays of high level play on Setons for instance. Land scout / labs are sent out to split the opponents forests. The tree split exists only for effect-reasons (to get trees to fall when you walk through the forest) so we should try to keep it as balanced as possible.

With this change we ignore the single tree blueprint values, instead calculate per-tree value by using the tree group's mass-energy / (n_trees_after_split * 0.95).

The 5% bonus is to compensate for the slower reclaim and probably too low anyway but it's a start, this way we won't overdo the buff and can see how gameplay adapts.

### DIFF
--- a/lua/sim/Prop.lua
+++ b/lua/sim/Prop.lua
@@ -201,27 +201,32 @@ Prop = Class(moho.prop_methods, Entity) {
     -- If not given, it defaults to one directory up from this prop's blueprint location.
     --
     SplitOnBonesByName = function(self, dirprefix)
+        local bp = self:GetBlueprint()
+
         if not dirprefix then
             -- default dirprefix to parent dir of our own blueprint
-            dirprefix = self:GetBlueprint().BlueprintId
-
             -- trim ".../groups/blah_prop.bp" to just ".../"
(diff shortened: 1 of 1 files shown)


Crotalus:
Another way to balance this would be to increase reclaim time of group trees instead, the downside with that is that it'll fuck up current build order more than the above change.

Zockyzock:
You could also argue that tree groups are there for map makers and that the "BO-whoring" that is amplified by tree groups is bad for the game. As solution you could always break tree groups when they start getting reclaimed.

Then, sending a unit to kill enemy tree groups doesn't have to be bad, it's not uncounterable after all and just another strategy. Instead of raiding mex you raid tree groups, and have to defend against this raids as you have to defend mex. 

Not saying either is definitely good, but we have to think about where we want to go with tree reclaim in general.

PR #967, opened 2015-10-21
Refactor offmap protection

Crotalus:
Less threading and planes now continue on their original order after returning to the playable area

### DIFF
--- a/lua/ScenarioFramework.lua
+++ b/lua/ScenarioFramework.lua
@@ -1983,182 +1983,76 @@ function GenerateOffMapAreas()
 	
 end
 
-function AntiOffMapMainThread()
-	WaitTicks(10)
-	GenerateOffMapAreas()
-	local OffMapAreas = {}
-	local UnitsThatAreOffMap = {}
-	-- WARN(repr(ScenarioInfo.ArmySetup))
-	
-	while ScenarioInfo.OffMapPreventionThreadAllowed == true do
-		OffMapAreas = ScenarioInfo.OffMapAreas
-		NewUnitsThatAreOffMap = {}
-		
-		for index,OffMapArea in OffMapAreas do
-			local UnitsThatAreInOffMapRect = GetUnitsInRect(OffMapArea)
-				if UnitsThatAreInOffMapRect then
-					for index, UnitThatIsOffMap in  UnitsThatAreInOffMapRect do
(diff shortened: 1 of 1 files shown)


Sheeo:
I never realized the previous offmapping code was this bad. Are we even sure we want all this overhead to prevent offmapping? I do believe the game itself has measures for it.

duk3luk3:
> planes now continue on their original order after returning to the playable area

I always thought losing orders was an intentional punishment for offmapping...

aeoncleanse:
It is, can we have that bit back please. And no Sheeo, the game just let it happen. It was an incredibly powerful tactic/exploit back in the day.

CodingSquirrel:
Can there at least be some buffer zone right at the edge if it's kept? I know sometimes I've tried to skirt the edge of the map with bombers, but they go a little bit too far and completely lose their orders.

duk3luk3:
I don't see a way to tell "trying to offmap" from "trying to *just* not offmap".

CodingSquirrel:
My point was that even if you are giving them orders to fly within the map sometimes they go just barely outside and then just stop. Hence a buffer zone. I'm not saying you need to tell the difference, I'm saying that having them lose orders just because they touch the edge isn't too great. So giving a little bit of extra space before that applies, to account for the fact that sometimes they fly a little wide might not be a bad thing.

Gap is a good example of a map where I've had this happen many times. Bases basically go straight to the edge of the map and so having planes fly along the edge is sometimes beneficial over having them fly straight into the middle of the base. Too bad if they go too close they just stop what they're doing and decide to have a nap.

irswat:
perhaps a conditional check on angle of approach to the axis coordinate for each edge might be useful. for example greater than or equal to 5 or 10 degrees.

irswat:
for groups of squadrons to could use a median or  mean angle of approach.

irswat:
another possible option of course is create an active edge function that triggers when a plane flying at a certain speed moves within a point or two of the edge of the map so that it repels the plane slightly like a magnet.

aeoncleanse:
If that last comment is possible, we should just do that. By far the best solution.

duk3luk3:
Massive offmapping abuse with Ahwassa: https://www.youtube.com/watch?v=Ank8MoS4SP0 / Replay ID 4302695

PR #971, opened 2015-10-23
Update teamkill detection

quark036:
Update teamkill detection to have everyone send a report that a teamkill happened. Everyone sends a 'TeamkillHappened' message, but it probably only needs to be investigated if you get a 'TeamkillReport', which only gets sent if the victim wants to report the teamkill.

(also removed some duplicated warns from defaultunit)

### DIFF
--- a/lua/UserSync.lua
+++ b/lua/UserSync.lua
@@ -69,6 +69,10 @@ function OnSync()
     end
 	
     if Sync.Teamkill then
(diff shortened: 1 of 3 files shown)


Crotalus:
:+1:  In balancetesting

172f1dd24937e4e84034afc43fa2341d917a75db

Issue #966, opened 2015-10-19
Warning in log when selection civilian units/structures.


speed2CZ:
```
WARNING: Invalid icon for unit uec1101
WARNING: stack traceback:
WARNING:         ...orever\gamedata\lua.nx2\lua\ui\game\construction.lua(434): in function `GetBackgroundTextures'
WARNING:         ...orever\gamedata\lua.nx2\lua\ui\game\construction.lua(782): in function <...orever\gamedata\lua.nx2\lua\ui\game\construction.lua:643>
WARNING:         ...orever\gamedata\lua.nx2\lua\ui\game\construction.lua(858): in function `SetControlToType'
WARNING:         ...nce\gamedata\lua.scd\lua\ui\controls\specialgrid.lua(109): in function `CalcVisible'
WARNING:         ...nce\gamedata\lua.scd\lua\ui\controls\specialgrid.lua(167): in function `Refresh'
WARNING:         ...orever\gamedata\lua.nx2\lua\ui\game\construction.lua(296): in function `OnCheck'
WARNING:         ...alliance\gamedata\mohodata.scd\lua\maui\checkbox.lua(45): in function `SetCheck'
WARNING:         ...orever\gamedata\lua.nx2\lua\ui\game\construction.lua(2094): in function <...orever\gamedata\lua.nx2\lua\ui\game\construction.lua:1967>
WARNING:         ...orever\gamedata\lua.nx2\lua\ui\game\construction.lua(2965): in function `OnSelection'
WARNING:         ...\faforever\gamedata\lua.nx2\lua\ui\game\gamemain.lua(419): in function <...\faforever\gamedata\lua.nx2\lua\ui\game\gamemain.lua:409>
WARNING:         ...\faforever\gamedata\lua.nx2\lua\ui\game\gamemain.lua(852): in function <...\faforever\gamedata\lua.nx2\lua\ui\game\gamemain.lua:850>
WARNING:         ...\faforever\gamedata\lua.nx2\lua\ui\game\gamemain.lua(870): in function <...\faforever\gamedata\lua.nx2\lua\ui\game\gamemain.lua:868>
```
If you open cheat menu and select Civilian filter. Spawn unit and select it. This is example of one of the UEF civilian buildings.

Started few patches ago.

duk3luk3:
This is also part of #895 and will probably not be fixed because cheat menu is not visible to most users.

speed2CZ:
[...]

Issue #895, opened 2015-08-29
Create some dummy textures or something to get rid of warnings in createunit (cheat) dialog


duk3luk3:
https://gist.github.com/duk3luk3/28f8a16af16a789b2ade

ChrisKitching:
It's in the cheat dialogue, which is only really used for debugging.

Those textures are for campaign units and such that don't have defined icons. Adding icons for them would just be unnecessary bloat. We shouldn't add a check to "UIFile" to handle this case, because that function is _ridiculously_ hot and this would probably have negative perf implications.

One sensible approach might be to simply not load the blueprints for units in the "OPERATION" category.

A nice way to go about that would be to modify the unit restriction system to simply not load blueprints for restricted units. You could then make the restricted unit set always contain the "OPERATION" category, and this problem goes away (by deleting all the problematic units).
This would also slightly improve game load time, which would be nice.

duk3luk3:
Yes, that sounds like a good idea.

ChrisKitching:
If you can do that, you can also remove my extra check from Unit's OnStopBeingBuilt that causes incorrectly created units to explode.

(There is presently nothing stopping UI mods from issuing build orders to make restricted units, so we need to sim-side enforce it somehow. Not loading the blueprint is a nice way of doing it, but I couldn't figure out how quickly enough). This'd be a very welcome change.

duk3luk3:
`LoadBluePrints` in called from `ruleInit.lua` (in mohodata.scd) which is called by the engine it appears.  
Like `simInit.lua` explains in the top comment, blueprints are loaded before ScenarioInfo is processed, which gets us unit restrictions.

I don't see a way to turn this around.

ChrisKitching:
This is exactly the problem I wasn't able to solve quickly enough. With extra magic this can probably be solved (probably by delaying blueprint loading, or retroactively deleting some blueprints)

duk3luk3:
As @speed2CZ notes in referenced issues, this problem also impacts mapmakers. We should probably think about what the easiest achievable not-a-total-hack fix is rather than what the Right[TM] solution is.

Also revisiting @ChrisKitching's comment
> We shouldn't add a check to "UIFile" to handle this case, because that function is ridiculously hot and this would probably have negative perf implications

The following appears to me: You know what has *terrible* perf implications?

```        
WARN(debug.traceback(nil, "Warning: Unable to find file: " .. skins[currentSkin()].texturesPath .. filespec))
```

[...]

PR #972, opened 2015-10-23
Hotfix/uifile logspam

duk3luk3:
This merge request fixes not only one (#895), no, it fixes TWO issues (#966)! And if you fast-forward merge it **RIGHT NOW**, I'm throwing in a fix that actually makes the UIFileCache work **FOR FREE**!

### DIFF
--- a/lua/ui/uiutil.lua
+++ b/lua/ui/uiutil.lua
@@ -92,21 +92,6 @@ currentLayout = false
 changeLayoutFunction = false    -- set this function to get called with the new layout name when layout changes
 
 local UIFileCache = {}
-local FileCache =  {}
-
-local oldDiskGetFileInfo = DiskGetFileInfo
-function DiskGetFileInfo(file)
-    if FileCache[file] == nil then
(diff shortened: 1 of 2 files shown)


lua/ui/uiutil.lua
```diff
@@ -327,9 +368,12 @@ end
 
 --* given a path and name relative to the skin path, returns the full path based on the current skin
 function UIFile(filespec)
+    if UIFileBlacklist[filespec] then return filespec end
+
     local skins = import('/lua/skins/skins.lua').skins
     local useSkin = currentSkin()
     local currentPath = skins[useSkin].texturesPath
```
Crotalus: What happens here if useSkin = nil? Wouldn't that trigger an exception?

Crotalus:
Nevermind, lua is happy with indexing nil... 

:+1: 

231583497fda33ed4d84d0f8d5b7f6461558c128
28be3adf2abbaf149ae784bf1df6ee844764ef3a
7266c01323c94f9edc28916857d1f7aaf0a727a8

PR #976, opened 2015-10-24
Fix #891

### LINKED ISSUE #891: Cybran sACU is able to emp units without the upgrade
Certain units, e.g t2. Read above.

### LINKED ISSUE #973: Disabledbf needs to be a set
https://github.com/FAForever/fa/blob/develop/lua/sim/weapon.lua#L448

Many inefficient, such wrong, wow.

duk3luk3:
Due to faulty table manipulation, disabled buffs wouldn't get cleared from
the damageTable.

- Remove unneeded GetDamageTable override in DefaultProjectileWeapon
- Shadow unneeded Weapon schook with empty file
- Add the stuff those two did into the main Weapon class
- Turn Weapon.BuffDisablers into a proper set (Fixes #973)
- Change weirdness-based addition and subsequent incorrect removal of
  buffs to efficiently not adding disabled buffs in the first place

### DIFF
--- a/lua/sim/defaultweapons.lua
+++ b/lua/sim/defaultweapons.lua
@@ -12,13 +12,6 @@ local CalculateBallisticAcceleration = import('/lua/sim/CalcBallisticAcceleratio
 -- Most weapons derive from this class, including beam weapons later in this file
 DefaultProjectileWeapon = Class(Weapon) {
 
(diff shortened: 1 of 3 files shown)


Crotalus:
:+1: 

2c083a13c0877969f308c02359ef00374be62a06

PR #978, opened 2015-10-25
Make mapselect not spaz out after changing mods

duk3luk3:
When the map list changed, e.g. when switching from normal FAF to Coop,
resulting in the previously selected map not being available anymore, the
missing check for scenarioInfo being set caused the UI to bug out.

- Check for scenarioInfo being set


duk3luk3:
This affects mainly devs and mapmakers but should be fixed regardless.

Issue #923, opened 2015-09-16
Another Seraphim ACU Restoration problem


SwayGin-FAF:
I was looking into this due to hepko's issue, and it has some other problems.

When leaving the field, the max HP and regen rate still appear in the unit's info in bottom left.

The regen rate doesn't actually take effect once a unit takes damage; it's just a typo.  However, upon taking damage, one of two things occur:
(1) If the damage is less than the HP buff (i.e. 10% of max life of buffed unit) then HP of buffed unit goes to normal max HP (i.e. 100% from 110%) on the first shot.
(2) If the damage is more than the HP buff, then HP buff is fully applied to reduce the damage.

E.g., an othuum has 6700 HP, so 670 HP buff for total of 7370 HP.  A percival shot does 1600 HP damage, so the othuum's HP after one shot is 7370 - 1600 = 5770.  On the other hand, a LAB does 7 damage, but after one shot the HP buff is removed, i.e. that one shot reduces the othuum HP to 6700;  the first shot effectively does 670 damage.  Similarly, an aeon T1 sub that does 75 HP damage normally does 4900 damage to a buffed Seraphim T3 battleship on it's first shot.

I don't think the HP buff is designed to be maintained outside the ACU range (if so, a real advantage to having ACU's assist factories where units get buffed as they slowly roll off), but whatever the case it seems inconsistently applied, as it relatively increases the value of low damage/high rate of fire units as a counter.  Splash damage also completely removed the HP buff.  

For example, mixing in a couple T1 artillery, LABS or T2 gunships to a Percy vs. Othuum fight helps prevent Percies from wasting 670 of their 1600 HP damage on the buff.  Similarly, mixing some T1 subs into a battleship fight is greatly beneficial due to this effect; the first aeon t1 sub shot does 75 damage normally, but effectively 4900 damage to a Seraphim T3 battleship.

As no ASF HP is outside the interval [1473,1800] (UEF is exactly 1800), the number of "rounds" (1 round = 2 shots for UEF/Cybran) to destroy another ASF required will always be 5, while T1 inties are likely too slow to provide a significant relative benefit when mixed with ASFs in a similar fashion to the above.  For lower tier air fights, the 2100 mass cost of the upgrade is prohibitively expensive.  So, placing an ACU near an air factory (somewhat easy to do, as usually far from the front unlike land factories) isn't an issue.

So, for land and (especially) naval fights, this effect is quite significant, whatever it's intention outside of the ACU restoration range.

[...]

PR #980, opened 2015-10-25
Fix HP Regen Buff Caluculation

### LINKED ISSUE #899: Seraphim acu restoration field BUG

aeoncleanse:
Merges the two subcategories of regen to stop them interfering, and removes the floor/ceil mechanic (Only used by Seraphim Regen Aura) because it's going to be a constant pain in the butt, interfering with other legit buffs left right and center.

This *should* fix #899 and #923 (The latter when combined with @duk3luk3 's #976 )

Hopefully that stupid upgrade will never bother us again...


ppipp:
What do you mean by removing the "stupid ceiling mechanic"? It's not stupid at all, it prevents the ability to be completely op with high HP units and have it significant for ones with low hp.

aeoncleanse:
It's stupid because it didn't work. Moreover, the way buffs are done, it *can* never work. Particularly the 15/s on the first one, as soon as you come into range of anything with legit higher regen than that (A large number of things!) poof! It's gone.

The field with the 200 cap has a mult of 0.0111111. 200 / 0.1111111 = 18000, which means only things with 18000 HP + are going to be affected anyway, which is just SCUs, which happen to have a built in regen upgrade which goes higher than that. The whole situation is poor game design. 

What we want is to have a ceiling which affects only the bit of regen applied by the field, but the way the buffs are calculated, I don't see a way to sensibly do that. I'm looking though.

aeoncleanse:
Made it work after all

aeoncleanse:
Retargeting

Issue #981, opened 2015-10-26
Free unit exploit


aeoncleanse:
[e]Exotic_Retard: 
get mega bruild some egg
[e]Exotic_Retard: 
ctrl k egg while its hatching
23:56
[e]Exotic_Retard: 
reclaim egg, recieve free brick
[e]Exotic_Retard: 
apparently you could use sacu rebuild feature to also rebuild the egg, but in some recent patch that doesnt work anymore
23:57

aeoncleanse:
Fix merged in 3652

PR #982, opened 2015-10-27
Fix free egg unit exploit

aeoncleanse:
Addresses #981

Will get someone who knows *exactly* how to pull this off to confirm my test that it's fixed before giving a thumbs up.


Crotalus:
What about the reclaim case? If you have enough buildpower you can reclaim the egg during it's "unfold" phase and get the wreck that way.

aeoncleanse:
I thought about that too. Any ideas on how to prevent?

aeoncleanse:
Retargeting

PR #987, opened 2015-10-28
Fix free egg unit exploit

aeoncleanse:
Tested working. Addresses #981. We made need to hit this issue again to address a slight variation where you reclaim it while unfolding without killing it.


Crotalus:
:+1:
560c1021e29c3b8a52e6d18f295f4814f270b804

PR #988, opened 2015-10-28
Increase transport hover height

aeoncleanse:


Sheeo:
Not a balance patch and we're in feature freeze.

aeoncleanse:
Oops, meant this to hit develop. mah bad!

Issue #892, opened 2015-08-28
Shields get overkill damage from nukes (or something)


Exotic-Retard:
replay id: 3809087

when the shield is activated, and the nuke is not in range of the actual shield building at all, but is in range of the shield, then the building also dies. a nuke fired in the same location with the shield deactivated, then the nuke doesn't harm it. affects allied and enemy shields the same way it seems.

as a bonus(!), this also contains cyb engineering drones leaving wrecks when engy stations are destroyed, giving 0 reclaim. maybe consider removing it? (its kinda a cool feature though)

hope this helps

aeoncleanse:
Long since fixed

Issue #836, opened 2015-07-28
personal shield doesn't transfert damage to hp


shalkya:
it's well explained here : http://forums.faforever.com/forums/viewtopic.php?f=3&t=10417

in few word : you can land a nuke or a T4 bomb, anything, on a 1% remaining shield, shield will disappear, but no damage will be done to hp.

ChrisKitching:
Pretty sure this is intended. We may wish to special case nukes

shalkya:
intended ? wtf ?

ChrisKitching:
Fairly sure this is the shield behaviour in vanilla, too (with the possible exception of nukes, as we fixed the area damage bug)

ppipp:
This is certainly not intended. All personal shields should have (or behave as if they have) the parameter             PassOverkillDamage = true, and all personal shields should transfer the extra damage to the unit except maybe in certain rare circumstances, the exact same way static shields don't block nukes.

Sheeo:
Yes indeed, the only shields which shouldn't have this parameter are normal bubble shields, where the building itself has low hp.

Nukes that collide with the shield, t4 bombs etc. may need to be specialcased, or we could make them not collide with the shields?

duk3luk3:
This should be tagged <balance>.

shalkya:
i don't think so, this bug was bring since the last patch.

Crotalus:
Indeed, it's a regression

ChrisKitching:
60cd3b890ded077edceaeee5b49c798fe64d8e6c

ChrisKitching:
(All personal shields except those on command units have the general PassOverkillDamage flag, for commanders we preserve the original behaviour except for nukes)

shalkya:
please make it working for every thing, not only nuke. I don't see the point of fixing it only for nuke.
ahwassa bomb, and czar crash, can still deal nearly 0 damage, depending of the health left on the shield.

Sheeo:
@Zockyzock @ppipp what are your opinions on this?

Zockyzock:
Needs to work for everything, as he said and it used to work.

shalkya:
Can you make it work for everything please ?

ppipp:
Yes, it should have been done for last patch. It should work the same for everything, not just nukes. Since there will be a hotfix soon for firebeetles, this should be done too.

duk3luk3:
Related: #892

aeoncleanse:
[...]

PR #991, opened 2015-10-29
Fix shield structures dying from nukes out of range, and make sure personal shields pass overkill

aeoncleanse:
Addresses #836 and #892. Attempts to really nail down how shields should be working with the following statements:

There is no situation in which we want a personal shield mesh to NOT pass overkill to the unit
The harbinger is the only instance of a sphere shield we do not want to be able to get underneath. We therefore DO want this to pass overkill
There is no situation in which we want a bubble shield we can get underneath to pass overkill to the unit
There is no situation in which we want a bubble shield to NOT spill damage to surrounding shields

Reverts Chris's change because it was completely unnecessary, and caused #892.
Since the overkill code is in the subclasses now, there's no need for the bp value PassOverkillDamage.


aeoncleanse:
Retargeting

PR #993, opened 2015-10-29
Protect continental when it has the shield up

### LINKED ISSUE #992: Regression - Continental takes body damage from flak with shield up again

aeoncleanse:
This stops it dying to the engine's AOE bug. Fixes #992


aeoncleanse:
Retargeting

PR #994, opened 2015-10-30
Fix for wrecks not giving any resources

speed2CZ:


Crotalus:
54528d32a1cce71b1b9bc4eab4a59887aca4656b

PR #999, opened 2015-11-01
Fix #883 - Return early from DoDamage if no damage

### LINKED ISSUE #883: ERROR: THERE IS NO INSTIGATOR FOR DAMAGE ON THIS COLLISIONBEAM
info: *ERROR: THERE IS NO INSTIGATOR FOR DAMAGE ON THIS COLLISIONBEAM = \000{
info:   CollideFriendly=false,
info:   DamageAmount=0,
info:   DamageFriendly=false,
info:   DamageRadius=0,
info:   DamageType="Normal"
info: }

https://dpaste.de/dQeN

duk3luk3:
The DoDamage function caused logspam due to a logic error that would
conclude a fault condition whenever there was no damage to be done.
We fix this by returning early.

### DIFF
--- a/lua/sim/CollisionBeam.lua
+++ b/lua/sim/CollisionBeam.lua
@@ -86,6 +86,8 @@ CollisionBeam = Class(moho.CollisionBeamEntity) {
 
     DoDamage = function(self, instigator, damageData, targetEntity)
         local damage = damageData.DamageAmount or 0
+        if damage <= 0 then return end
+        
         local dmgmod = 1
         if self.Weapon.DamageModifiers then
             for k, v in self.Weapon.DamageModifiers do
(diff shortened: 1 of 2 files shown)


aeoncleanse:
This may not be a good plan. I think there are some units, particularly in mods, which do 0 damage on some weapons. Dummy weapons are a good example. I think it would be more useful to find out why the hell DoDamage is being called at all for collisionbeam. If anything, escape early with no instigator.

duk3luk3:
The only thing DoDamage does is... do damage. If there is no damage to be done, there is no reason to do anything.

aeoncleanse:
not true. For example, an AOE stun would probably work with 0 damage called on all impacted units, going through unit OnDamage to actually do the stun.

duk3luk3:
Please look at the diff of my commit. It does not change anything except spuriously emitting the error that there is no instigator when the problem is in fact that the damage is zero.

And if the damage is zero, the function already doesn't do anything. My change doesn't make it do less.

aeoncleanse:
Ah, line 96. Ignore me :D

Crotalus:
473d32e234217241569f10ee4d4c426daa91e8a5
532f4701abaab930b9355b9f8cb6e7d2e215f88a

PR #1000, opened 2015-11-01
Fix coop missions stuck on Operation Complete

duk3luk3:
Heavy-Duty refactoring broke coop in as of yet unknown ways. Here's a
hotfix that avoids attempts to load the _operation.lua file when the
ScenarioInfo.campaignInfo table is missing and thus there is no opKey.


Sheeo:
You should get an avatar for filing PR/issue #1000

duk3luk3:
I agree.

Crotalus:
bf70fc28e3c77ed5c33a1bc03054d43a23146db4

PR #1006, opened 2015-11-03
Hotbuild support factories

madformuse:
I've done quite a lot of testing and everything seems in order.


speed2CZ:
I think this should be done in some more complicated way, then just adding all factories on one key. This way you won't speed up anything if you have to press one key so many times to get the building you want.

Sheeo:
You'll click at most 3 times, and the most important factories (t3) show up first. I can't think of a complicated solution (This isn't normally requested :8ball:) that would do better?

madformuse:
I was wondering if starting with higher tech factories would be better but I decided to fit in with the existing convention.

As someone who actually gets past the t1 phase what factories do you make the most in late game?

Actually scrap that... I fiorgot, you only make sparkies :p

madformuse:
Actually Sheeo all 3 land would show up first. I wanted all high tech first but radar / pd etc cycle through all the tech levels for a unit before moving on.

I would 've happy to change them all but I hope it wouldn't screw anyone over changing all the orders.

Sheeo:
Well maybe cycling t3land, t3air, t3naval, t2land ... is better then?

madformuse:
Do you think we need to make it consistent with the other groups? Or more to the point change them so they're consistent?

madformuse:
Come to think of it showing all T3 factories first would not work without a re-think because it organises them in sub groups of Land, Sea and Air which each contain all the techs and the "builders" group lists all three sub-groups. So the order is dictated by the organisation.

madformuse:
Sorting on tech after the list is compiled could work and would work for all groups. Do we really want this?

Zockyzock:
cycle t3 land, t3 air, t3 navy, then t2, then t1

Crotalus:
40b9706d4ce356aab9320680357874bccfe1465b .. af710072af26fa8f690b3336dfc13fd1f6a86a66

PR #1007, opened 2015-11-04
Calculate resource income for score window display without reclaim influence

aeoncleanse:
All FA's casters have been asking for this since SupCom BETA. At. Freaking. Last.


lua/score.lua
```diff
@@ -133,6 +134,23 @@ function ScoreThread()
...
+            ArmyScore[index].resources.massin.rate = brain:GetArmyStat("Economy_Income_Mass", 0.0).Value - massReclaimRate
+            ArmyScore[index].general.lastReclaimedMass = reclaimedMass
+            
+            local reclaimedEnergy = brain:GetArmyStat("Economy_Reclaimed_Energy", 0.0)
+            local energyReclaimRate = reclaimedEnergy - lastReclaimedEnergy
+            ArmyScore[index].resources.energyin.rate = brain:GetArmyStat("Economy_Income_Energy", 0.0).Value - energyReclaimRate
+            ArmyScore[index].general.lastReclaimedEnergy = reclaimedEnergy
+        end
+        WaitSeconds(0.1)
+    end
```
Sheeo: This will cause the remaining part of this function to be dead code. Are you _absolutely_ sure this is intended?

aeoncleanse: No. I'll have to funtion this out then

lua/score.lua
```diff
@@ -195,6 +195,25 @@ function ScoreThread()
     end
 end
 
+function ScoreDisplayResourcesThread()
+    -- For certain stats, we need to do this every tick. We can't for all because it is quite heavy CPU
+    -- We don't need to sync every tick though, just make sure the number is right
```
Sheeo: Why can't this just happen in the upper score gathering loop?

aeoncleanse: I'm not entirely sure what you mean? The trouble with this was that most of the score params are fetched from engine and stored every 3 seconds, which is fine for score calculation. However, for these two, we need to fetch, calculate, and store every tick to ensure the displayed number is correct, else we run into some problems trying to average the reclaim influence.

aeoncleanse:
retargeting

PR #1010, opened 2015-11-08
Allow modded unit icons to be found

aeoncleanse:
So this UI cache has never worked... Which was dumb. Quite a noticeable speed increase when navigating the lobby and build menus actually.

This also allows unit icons for custom mod units to be shown by default, and in a far simpler manner than BlackOps' Icon Support mod. uiutils really didn't like having __active_mods there until I put in the sim argument to tell it it was being called from a place we knew was only happening post-sim-initiation. If there's a better check for sim state from lua, I'd like to know, but this works for now.


Crotalus:
As to the UI cache not working, it was fixed in these commits (release/3652):

ec5f987332723ff6eae22300a9ce4e7b508c4296
029a3fab6f43efb4be886414eb46428e8fb23485

I haven't checked out blackops custom icon mod, the normal VFS shadowing for mods doesn't take care of this? As to the name "sim" it's a bit misleading because the function is still called in UI-state?

aeoncleanse:
Oh sweet, and we even solved it basically the same way lol. I'll rebase onto 3652 and plop the two together. 

I agree it's misleading, I'd love a better way. Something like IsGameLaunched would be handy lol. I'll rename it. But no, the normal shadowing simply doesn't, and has never, worked for mod icons. The BO mod I have taken a thorough look at and it does a bunch more things than this, but my GOD ALMIGHTY is it overly complex... Wowzer!

aeoncleanse:
Closing to retarget

PR #1012, opened 2015-11-08
Bugfix not reclaiming

### LINKED ISSUE #1004: Assisting engineers fail to reclaim
Replay number 4040083 as an example.

To reproduce:
  1) Set engineers to assist another engineer (main engineer)
  2) Set main engineer to reclaim some unit (e.g a factory)
  3) Tell main engineer to stop

The assisting engineers will continue to reclaim the unit forever, the unit's hp will reach 0 but it will not be destroyed.

Only tested with t1 engies and ACU as main engineer, and the unit as a t1 factory.

aeoncleanse:
Fixes #1004


aeoncleanse:
Freaking target...

PR #1016, opened 2015-11-09
Calculate resource income for score without reclaim influence

aeoncleanse:


Crotalus:
a6182b9debc1272f080689480313faa6c03ecaf5

PR #1018, opened 2015-11-09
Protect continental when it has the shield up

aeoncleanse:
This stops it dying to the engine's AOE bug.


Crotalus:
426ae57e7955acf9c3fbc6db348ea338d56abd4a

PR #1027, opened 2015-11-17
Fix log spam when switching army view in campaign

speed2CZ:
Army unit count isn't updated every tick and switching army view in campaign caused few seconds log spam until it was updated. This sets it to 0 until it's updated instead of nil. 

by Washy


speed2CZ:
@Crotalus please add ^^

Crotalus:
Modified the commit some but should be same effect in the end

4c7350f04330d04b1cb636af50909fdf8351ff38

PR #1030, opened 2015-11-20
Personal stealth units in transport

aeoncleanse:
Stops transports being obvious as such from their radar signature. A loaded transport now appears as a normal plane, a single signature.


Sheeo:
Why should this be in 3652?

aeoncleanse:
I just targeted it at the branch I forked it from. Zock says not this time, but it's nice to have ready in case it's wanted.

PR #1035, opened 2015-11-25
Fix a couple of things missed in the nuke refactor

aeoncleanse:
Slow nukes need 'nuke' damagetype
Missed a re-path in the exp nuke script
Need the effectEntity to be saved to self


aeoncleanse:
Stupid develop default yada yada yada...

Issue #1028, opened 2015-11-18
Ctrl-k countdown doesnt get cancelled.


Exotic-Retard:
so you press ctrl k on something, and what used to happen before is if you give it any order it cancels the countdown
now even the stop command doesn't seem to do it.
i want this functionality back. so that any order cancels the countdown. including move orders.
replay: 
4091917
this affected me in a couple of games now so yeah : /
now im being civilized about this here but i do have a few choice ways of describing this issue, and i would really appreciate it if this was fixed. preferably for next patch.

aeoncleanse:
So I've had a look. Anyone got any clue where the code for this is/was? I can't find it.

Crotalus:
lua/selfdestruct.lua contains the self-destruct stuff.

aeoncleanse:
Actually got around to testing this ingame... It works fine! It cancels the boom, at least in 3652.

PR #1038, opened 2015-11-26
Improve SelfDestruct and Beetle code

aeoncleanse:
Checking out #1028 I found some improvements to be made. Should be nonfunctional. Beetle tested fully for all three triggers and normal death.


Crotalus:
Merged
d000dea23f65a98ef5bcd14f85f7e8c1a706ce9e

PR #1042, opened 2015-11-28
Show resources from player perspective

aeoncleanse:
Just a small improvement to the score window top right for casters. The fix I made before successfully stops reclaim showing up, but didn't work when watching from a player's perspective.


Crotalus:
Hmmm there was another bug with watching replays, score wasn't show in overlay. I've fixed that in this commit:

563ed9c67e043d278dd6fe42435db1a4983d2c41

Most likely this will fix this issue as well, unless we want replay observers really to get "what it looked like for a player".

aeoncleanse:
Crotalus was right, his commit did indeed fix the issue this was addressing. Closing.

Issue #1047, opened 2015-12-07
Restricted ACU upgrade can be finished by starting/stopping allowed upgrade and then starting restricted upgrade 


HUSSAR-mtrela:
1. Start the game with restricted "Teleport" in unit manager
2. Start upgrading any allowed enhancement , e.g. Tech 2 Engineering Suite
3. Stop upgrading enhancement in middle of progress or close to finish
4. Start upgrading "Teleport"  enhancement and it has progress from the stopped enhancement!!!

Result: 
The "Teleport"  enhancement finishes even it was restricted.

Root Cause: 

EnhanceTask.lua throws nil error when setting unit:SetWorkProgress(current) on line 68
Construction.lua does not disable/hide restricted enhancements because EnhanceCommon.GetRestricted() function always returns empty table.

Once restricted enhancement  is finished it cannot be removed by starting other enhancements in the same tab.

Proposed fix: 
- Construction.lua should hide/disable restricted enhancements by checking RestrictedCategories in SessionGetScenarioInfo().Options against restrictedUnits table 
- EnhanceTask.lua should also check for RestrictedCategories  becasue Construction.lua can be overridden in UI mod

HUSSAR-mtrela:
Please assign it to me because I have a fix for this issue
Also add labels: UI, Bug, Critical

Sheeo:
This must be a sim problem.

PR #1050, opened 2015-12-08
Fixed restricted commander upgrade issue #1047

HUSSAR-mtrela:
this is a fix for high sim bug

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1704,7 +1704,7 @@ Unit = Class(moho.unit_methods) {
             if self:IsValidBone(v) then
                 self:ShowBone(v, children)
             else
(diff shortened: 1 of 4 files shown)


Crotalus:
Nice work with the PR, it's pointing to `develop` instead of `release/3652` though. I'll rebase and merge your changes.

I'll add some additional commits also to fix the fact that EnhCommon.GetRestricted() is never set/used. The restriction table should be synced from simside so we are able to change the restrictions on the fly in a game.

Crotalus:
Merged here:
267313f45ac56853071a5684e3a2a8914954eeda..e78ecdfa42ccc777d08226f0645ae5ab1c087158

Sheeo:
On another note: In the future please use consistent indentation settings -- we use spaces in this project.

HUSSAR-mtrela:
@Sheeo I'm sorry for the indentation issue and I usually use spaces but I forgot to set this setting in notepad++ after I upgraded to Win8. I will make sure my future changes also use spaces to indent code.

PR #1051, opened 2015-12-09
Reintroduce Russian Translation

aeoncleanse:
So I couldn't recreate the crash sheeo mentioned to me as the reason for reverting the translation changes. This branch has the reverts rebased away to re-enable it, as well as a couple of new, translated tooltips and a bit of blacklisting to cut out some logspam.

This branch launches the game OK my end, with all sorts of options combos. The tooltips are all working in english, so the fancy use of %s seems to be working fine. 

Is it possible this was a deployment issue?


aeoncleanse:
@HUSSAR-mtrela has also tested this and found no crash.

aeoncleanse:
OK so we have mystery crashing from sheeo, though I can't make it happen, and the code is apparently a mess. Need to get rid of numbers as strings, and apparently restructure it a bit, though frankly I like the way this has been done. @Crotalus any ideas?

PR #1053, opened 2015-12-10
Have beams kill missiles

### LINKED ISSUE #628: Beams collide with missiles. Should they?
Softly brought this to my attention. Confirmed for ML and Sera PD

aeoncleanse:
Asked Zock what to be done, wants consistent behaviour. Quick aeolus polls had most saying impact should happen and missile should die to the beam.

Fixes #628

### DIFF
--- a/lua/sim/CollisionBeam.lua
+++ b/lua/sim/CollisionBeam.lua
@@ -101,6 +101,10 @@ CollisionBeam = Class(moho.CollisionBeamEntity) {
             if radius and radius > 0 then
                 if not damageData.DoTTime or damageData.DoTTime <= 0 then
                     DamageArea(instigator, self:GetPosition(1), radius, damage, damageData.DamageType or 'Normal', damageData.DamageFriendly or false)
+                    -- Do non-AOE damage to the target if a missile to make beams kill missiles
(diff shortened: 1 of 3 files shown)


lua/sim/CollisionBeam.lua
```diff
@@ -101,6 +101,10 @@ CollisionBeam = Class(moho.CollisionBeamEntity) {
             if radius and radius > 0 then
                 if not damageData.DoTTime or damageData.DoTTime <= 0 then
                     DamageArea(instigator, self:GetPosition(1), radius, damage, damageData.DamageType or 'Normal', damageData.DamageFriendly or false)
+                    -- Do non-AOE damage to the target if a missile to make beams kill missiles
```
Sheeo: This comment is more confusing than it is beneficial. What do you mean?

aeoncleanse: This whole thing happened in the first place because the engine DamageArea function disregards all projectile entities in the radius. That's why the Cybran TMD, which uses a beam, has no radius, so it uses DoDamage instead, which is allowed to do the job.

Sheeo: I'm not questioning the code, merely the comment.

Let's put something like "Damage missiles in the area directly, because they aren't caught by DamageArea."

PR #1054, opened 2015-12-10
Tooltip for overcharge

aeoncleanse:
So people know you can right-click for autofire


aeoncleanse:
The inability to retarget a PR is really pissing me off...

PR #1067, opened 2015-12-20
Translation ITA

Lyvnah:
[translation ita.txt](https://github.com/FAForever/fa/files/67975/translation.ita.txt)


aeoncleanse:
Turned into #1073

PR #1068, opened 2015-12-21
Re-add the translation with the bugs fixed

aeoncleanse:
Turns out @Crotalus simply missed a handful of obscure locations which needed adapting to the new layout. This now works in my tests. A friend of mine is working on re-building the system to be less stupid, but this will do for the time being.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1620,20 +1620,6 @@ AirUnit = Class(MobileUnit) {
         self:AddPingPong()
     end,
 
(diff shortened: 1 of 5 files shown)


Crotalus:
No translation related commits here?

Sheeo:
Indeed. Reopen when these are found from the bitrot.

aeoncleanse:
Found the bitrotted stuff. I've pushed it to release/3652 so this can remain closed.

PR #1072, opened 2015-12-24
Have shields check for 1 Energy not 0

aeoncleanse:
Fixes shields sometimes not turning off properly on low energy. Was most apt with personal shields, as their lower consumption meant sometimes their share of E was <1 itself, possibly leading to not turning off on power cycle (Theory).


Sheeo:
Did we really need a pull request for this?

PR #1073, opened 2015-12-24
Italian Translation Update

aeoncleanse:
Thanks to Lyvnah. Probably needs adapting to the dynamic strings update when that's finally remerged.


aeoncleanse:
Updated with dynamic strings. Will merge in when Travis finishes.

PR #1077, opened 2016-01-05
Hotfix/1043

### LINKED ISSUE #1043: Possible to build factories in factories due to hotbuild
This commit makes it possible to order factories in factories from hotbuild UI:

2d82debb30c50997f6236ccd50fd8d2ce5284e48

Build a t2 naval yard, then order a t1 naval factory to build support factory. I think the problem is that the command is sent from hotbuild as BuildFactory instead of Upgrade.

The real problem is implementation of engymod though, the `'BUILTBYTIER1FACTORY' on support factories makes it possible to order build instead of upgrade.  That category shouldn't be needed but it's there to make it appear in the build UI. Should be changed when construction.lua is refactored

duk3luk3:
Introduces a new build restriction set that stop factories from building
(not upgrading) factories.

This is a workaround to fix #1043 until engymod can be fixed properly.

While we're at it, I also removed a useless CheckBuildRestriction override.

Please check that it doesn't mess with anything.

To test:

[x] There are no units with the `Factory` **primary** category built from `FactoryUnit` factories
[ ] Removal of the CheckBuildRestriction function from `ConstructionUnit` doesn't screw something up
[x] Aircraft carriers still work as expected, tested with Cybran Carrier
[x] Megalith eggs still work as expected

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2036,15 +2036,6 @@ ConstructionUnit = Class(MobileUnit) {
             self.BuildingOpenAnimManip:SetRate(-(self:GetBlueprint().Display.AnimationBuildRate or 1))
         end
     end,
-
-
-    CheckBuildRestriction = function(self, target_bp)
-        if self:CanBuild(target_bp.BlueprintId) then
-            return true
-        else
-            return false
-        end
-    end,
 }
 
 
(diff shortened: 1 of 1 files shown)


aeoncleanse:
Confirmed the first, can't imagine any problem with the second. Merging.

aeoncleanse:
57322217791cf2b85bf3b4000996b939e6531549

PR #1078, opened 2016-01-06
Increase TMD range by 1 to stop ground-fire abuse of TMLs

aeoncleanse:


Sheeo:
@Zockyzock

Zockyzock:
Needs confirmation if its enough

Zockyzock:
It's not enough. I have to find out what's the best way of solving the issue without buffing TMD too much. If this is holding the patch back, ignore it, and it will be in the next balance patch.

Zockyzock:
fire rate from 0.5 to 0.56
range from 0.3 to 0.31

PR #1079, opened 2016-01-07
Fix #1066 by whitelisting selected map

### LINKED ISSUE #1066: ingame map select filters conflict with faf map select
im sure everyone has had this before: host some game, click options, and then the map changes to some retarded voodoo tournament map or something.

been here for a long time but i think ive tracked it down:
its a combination of a few things:
1. the map filters in the game lobby
2. the map select in the host game dialog
3. the auto select of maps in the options of the ingame lobby

in short; when you select a map in the faf game dialogue (which is actually very nice and much faster than in game, and works with keyboard and text search)
and the filters in the game lobby are such that the map is filtered out,
it defaults to the first map in the available list
and it automatically selects that map
and if you press ok without even mousing over the map list then it sets that map.

solution?

make it reset the map filters to all on every game host (they are written to game.prefs afaik). this removes the issue while keeping all other functionality.

duk3luk3:
This is a tentative fix. It simply makes the selected map ignore filters.

If nobody has a better idea I suggest we adopt this.

One alternative I briefly entertained was always checking and relaxing the filters so the selected map would never be excluded, but that would be pretty messy.

### DIFF
--- a/lua/ui/dialogs/mapselect.lua
+++ b/lua/ui/dialogs/mapselect.lua
@@ -964,25 +964,27 @@ function PopulateMapList()
     local count = 1
     for i,sceninfo in scenarios do
         local passedFiltering = true
-        -- Subject this map to every activated filter.
-        for k, filter in currentFilters do
-            if not filter(sceninfo) then
-                passedFiltering = false
-                break
+        
+        -- If this is the currently selected map, mark it for reselection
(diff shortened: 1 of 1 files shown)


aeoncleanse:
Merged 9d04b36be06506b583dddebf909ddfa2f38811ce

PR #1085, opened 2016-01-11
Redo engymod restriction system so it works

duk3luk3:
This fixes a slew of issues with engymod.

*WIP*

Test with seraphim land factory, to test with all other support facs script must be updated to inherit from SupportFactoryUnit like done for seraphim land factory. I will do that as soon as someone has reviewed and tested this, because I don't want to waste time editing all the files until I know it's going to stick.


lua/defaultunits.lua
```diff
@@ -626,13 +652,10 @@ StructureUnit = Class(Unit) {
...
 ---------------------------------------------------------------
+
+
 FactoryUnit = Class(StructureUnit) {
     OnCreate = function(self)
-        -- Engymod addition: If a normal factory is created, we should check for research stations
-        if EntityCategoryContains(categories.FACTORY, self) then
-           self:updateBuildRestrictions()
-        end
-
```
Crotalus: Make sure HQ-status is updated when a factory is transferred between players, I think OnCreate() / OnDestroy() is called in these instances (due to units being recreated at transfer), not so sure that OnStopBeingBuilt() is though.

lua/system/class.lua
```diff
@@ -2,6 +2,20 @@
...
+              StructureUnit
+                  |
+              FactoryUnit
+               /       \
+SeraphimFactoryUnit    SupportFactoryUnit
+                \     /
+                ZSB9501
+
+In the unforked inheritance path, Unit, StructureUnit and FactoryUnit can all normally define and override a function. That same function may then be overriden in *either* SeraphimFactoryUnit or SupportFactoryUnit.
+
```
Crotalus: An alternative would be to change class.lua to support multi-inheritance and let the "last inherited" class method override the others, like the current hooking works. Could be confusing I guess.

duk3luk3: I'd say that as long as we don't **have** to touch class.lua, we shouldn't.

lua/defaultunits.lua
duk3luk3: Pretty sure it is called, I'll test that though.

duk3luk3:
New version. Now has all support facs inherit from SupportFactoryUnit.

Test that restrictions work by commenting out stuff in OnSelection in construction.lua so it doesn't filter out units that aren't allowed to be built/upgraded.

duk3luk3:
So I tested sharing with an AI.

I verified that OnDestroy() and OnStopBeingBuilt() gets called properly.

But it doesn't work.

Steps to reproduce:

1. Copy lua/ui/game/diplomacy.lua from GPG and remove the data.human check from line 174
2. Start a skirmish with an AI ally
3. Ctrl-K AI CMDR
4. Build factory and upgrade it to T2 HQ
5. Verify everything works as expected with your army
6. Spawn engy for AI player and build T1 fac
7. Transfer T2 HQ from you to AI
8. AI T1 Fac does not show upgrade

If I remove the filtering from construction.lua like said before

> Test that restrictions work by commenting out stuff in OnSelection in construction.lua so it doesn't filter out units that aren't allowed to be built/upgraded.

then I can order an upgrade and it builds properly.

Someone please take a look at this.

aeoncleanse:
Could be an error with AI brains which, I think, have many differences from human ones. I suggest we put this into beta once 3652 ships to test with two players, or else use the dev server.

lua/aibrain.lua
```diff
@@ -47,6 +47,75 @@ local Points = {
...
+        if not self.HQFacs[faction][layer] then
+            self.HQFacs[faction][layer] = {}
+        end
+
+        self.HQFacs[faction][layer][tech] = (self.HQFacs[faction][layer][tech] or 0) + 1
+        -- Sync
+        Sync.HQFacs = table.deepcopy(self.HQFacs)
+        
+        -- Check back on Sync table
+        WARN('Checking back Sync table from Sim')
```
micheljung: Why would we log this WARN?

lua/defaultunits.lua
```diff
@@ -291,40 +291,66 @@ StructureUnit = Class(Unit) {
...
-        local performUpgrade = false
-
-        if targetBp.General.UpgradesFrom == builderBp.BlueprintId then
-            performUpgrade = true
-        elseif targetBp.General.UpgradesFrom == builderBp.General.UpgradesTo then
-            performUpgrade = true
-        elseif targetBp.General.UpgradesFromBase ~= "none" then
-            -- try testing against the base
-            if targetBp.General.UpgradesFromBase == builderBp.BlueprintId then
+        if order == 'Upgrade' then
```
micheljung: Invert condition to reduce nesting (early return);

```
if order ~= 'Upgrade' then
    Unit.OnStartBuild(self,unitBeingBuilt, order)
    self.UnitBeingBuilt = unitBeingBuilt
    return
end

// Do the rest
```

lua/aibrain.lua
```diff
@@ -47,6 +47,75 @@ local Points = {
...
+        local faction = bp.General.FactionName
+        local layer = bp.General.Icon -- this is hacky but it works
+        local tech = bp.General.TechLevel
+
+        -- sanity check
+        if not self.HQFacs[faction] then
+            WARN('Hit unitialized HQFacs in RemoveHQFac, unit id ' .. bp.BlueprintId)
+        end
+        if not self.HQFacs[faction][layer] then
+            WARN('Hit unitialized HQFacs in RemoveHQFac, unit id ' .. bp.BlueprintId)
```
micheljung: This code is not only duplicated (same log 3 times) and would lead to having the same log three times, but also fails;

if `self.HQFacs[faction]` is unset, `self.HQFacs[faction][layer]` will throw an error.

Proper way of doing it: 
```
if not self.HQFacs[faction]
       or not self.HQFacs[faction][layer]
       or not not self.HQFacs[faction][layer][tech] then
   WARN('Hit unitialized HQFacs in RemoveHQFac, unit id ' .. bp.BlueprintId)
end
```

lua/defaultunits.lua
```diff
@@ -1938,7 +1986,45 @@ ConstructionUnit = Class(MobileUnit) {
...
+
+    -- Engymod check: Needs to allow building of support facs iff there is an HQ
+    -- Almost identical to SupportFactoryUnit.CheckBuildRestriction but needs to restrict only support fac build, not all units
+    CheckBuildRestriction = function(self, target_bp)
+        -- Check normal build restrictions first
+        if not MobileUnit.CheckBuildRestriction(self, target_bp) then
+            return false
+        end
+        -- Now check if we are building a support factory
+        if EntityCategoryContains(categories.SUPPORTFACTORY, target_bp.BlueprintId) then
```
micheljung: Invert condition to reduce nesting

lua/sim/Unit.lua
```diff
@@ -3199,8 +3062,10 @@ Unit = Class(moho.unit_methods) {
 
     CheckBuildRestriction = function(self, target_bp)
         if self:CanBuild(target_bp.BlueprintId) then
+            WARN('Unit CheckBuildRestriction succeeded from '.. debug.traceback())
             return true
         else
```
micheljung: `else` is not needed since there's a `return` in the `if`;

```if self:CanBuild(target_bp.BlueprintId) then
    WARN('Unit CheckBuildRestriction succeeded from '.. debug.traceback())
    return true
end

WARN('Unit CheckBuildRestriction failed from '.. debug.traceback())
return false
```

W/o the log statements (like it was before) it could just be written as:

`return self::CanBuild(target_bp.BlueprintId)`

lua/sim/Unit.lua
```diff
@@ -3199,8 +3062,10 @@ Unit = Class(moho.unit_methods) {
...
+            WARN('Unit CheckBuildRestriction failed from '.. debug.traceback())
             return false
         end
```
micheljung: `else` is not needed since there's a `return` in the `if`;

```if self:CanBuild(target_bp.BlueprintId) then
    WARN('Unit CheckBuildRestriction succeeded from '.. debug.traceback())
    return true
end

WARN('Unit CheckBuildRestriction failed from '.. debug.traceback())
return false
```

W/o the log statements (like it was before) it could just be written as:

`return self::CanBuild(target_bp.BlueprintId)`

lua/defaultunits.lua
```diff
@@ -291,40 +291,66 @@ StructureUnit = Class(Unit) {
...
+            elseif targetBp.General.UpgradesFrom == builderBp.General.UpgradesTo then
                 performUpgrade = true
+            elseif targetBp.General.UpgradesFromBase ~= "none" then
+                -- try testing against the base
+                if targetBp.General.UpgradesFromBase == builderBp.BlueprintId then
+                    performUpgrade = true
+                elseif targetBp.General.UpgradesFromBase == builderBp.General.UpgradesFromBase then
+                    performUpgrade = true
+                end
             end
```
micheljung: This block can be written as:

```
local performUpgrade =
    targetBp.General.UpgradesFrom == builderBp.BlueprintId
    or targetBp.General.UpgradesFrom == builderBp.General.UpgradesTo
    or targetBp.General.UpgradesFromBase ~= "none"
        and (targetBp.General.UpgradesFromBase == builderBp.BlueprintId
        or targetBp.General.UpgradesFromBase == builderBp.General.UpgradesFromBase)
```

However readability is more important than slick code

duk3luk3: Unnecessary if/then/else blocks do not improve readability so I changed it to your version :-)

lua/defaultunits.lua
duk3luk3: In this case I believe inverting the logic would make the code less readable.

duk3luk3:
Rebased and fixed some stupid shit.

There's (at least) one problem left: You can't queue units on a factory that is upgrading. That is a UI regression.

Someone please figure out how to check whether a factory is currently upgrading from `construction.lua`.

HUSSAR-mtrela:
@duk3luk3 are you using CheckBuildRestriction() to check if a unit is restricted? If so that is not working as expected when units are restricted in scenario scripts.

PR #1112, opened 2016-01-23
Disable functions in AI bases in Coop

speed2CZ:
This will make engies, sacu, acu not form a new platoons that would work in the base after base was set as inactive. Also stops scouting.


aeoncleanse:
Merged as 793d112e164125697b31fb886a405ab253b0f598

Issue #1111, opened 2016-01-23
Fix the Soft Stop feature


aeoncleanse:
@Crotalus @Sheeo @duk3luk3 this is one of the last few things remaining for 3652 to ship.

Crotalus:
Fix soft-stopping what? Factories? Silos? UI issue with pressing Stop icon? Something with mapping the function?

aeoncleanse:
Read in the #bugreports channel on slack. 

"report about beta bug :

[1:54] 
soft stop : doesn't work with commander and engie, act like a normal stop (for move, construction, attack, reclaim....... order)

[1:55] 
soft stop : cancel in factory all order if only 1 type of unit is queud up.

keyser [2:15 PM] 
stop stop : doesn't work on scu, hive, cybran shield, radar, sonar, mexx. It act like normal stop

[2:16] 
*soft stop

[2:17] 
soft stop : doesn't work on facto if you queue up T2 and T3 upgrade. if you press soft stop twice it cancel the whole queue. if there is some unit queue up after the T3 upgrade, the soft stop will remove them, but T3 upgrade won't be removed.

[2:21] 
normal stop : isn't working on nuke, tml, smd, nuke sub (only cancel the target selection, but not the missile being build)(it is working the soft stop)

[2:21] 
(btw we can't pause the sub nuke in normal game)

keyser [2:26 PM] 
acu/scu upgrade : you can't delete them (right click), you can only stop them with normal stop or soft stop (soft stop act like normal stop)"

Crotalus:
He must have dreamt about general units being able to soft-stop their orders, only been done for factories / silos.

aeoncleanse:
I've tested this just now on Factories, Shields, Engineering Stations, SML/SMD and TMLs. Appears to be working fine. Normal stop is also working as expected. Soft stop with one unit type in queue in Factories is working as expected.

PR #1114, opened 2016-01-27
Tweak soft stop

Crotalus:
* Always soft-stopping silo units could be confusing for players expecting default stop behavior for all units.

* Enable soft-stop for all buildings which show a queue

* Disable stop_last override behavior

Maybe solves #1111


Sheeo:
I completely missed this pr, sorry.

PR #1118, opened 2016-02-11
Readded TabelCat( ... )

siiikooo0743:
Readded TabelCat( ... ) for backward compatibility reasons. Mods use it and unnecessarily crash with FAF.


Sheeo:
Actually, it's better we just leave out this file entirely.

It's exactly the same in the base game distribution and there are no changes other than TableCat missing (It was renamed table.concatenate).

Thanks.

PR #1120, opened 2016-02-14
Make Auto-Overcharge more robust

### LINKED ISSUE #1109: Make Auto-OC toggleable when the button is inactive. Fix the tooltip popup.
This is the last remaining obstacle to 3652 AFAIK.

Crotalus:
Disabled buttons by default don't have hit-testing enabling so this quick
fix override this bevhavior for overcharge

Partly fixes #1109, still needs to change tooltip


aeoncleanse:
Just so I'm up to speed, normal behaviour is for control.Disable to call DisableHitTest to turn of click detect, and with this not called, it will behave when disabled as though it were enabled? I ask because we want ONLY right-click behaviour to function, not all clicks.

Crotalus:
Seems like there is defensive checks in checkbox like:
`if self._controlState != "disabled" then`

but on the other hand this is not try for click events. In this case I guess a leftclick will result in it going into `EnterOverchargeMode` but then return early because of not enough energy, i.e. nothing happens. 

Best thing would be to inform player about the energy needed when clicking on the overcharge icon, at the moment new players must be confused.

aeoncleanse:
Testing this revealed serious shortcomings. Yes, it can now be toggled when the button is off. However:

1: If toggled while off, you become totally unable to fire AT ALL until (I assume) the queued OC is fired once you get enough storage.
2: The greyed out button pulses as it would if it were enabled. This is misleading and annoying. It should just have the ring placed around the icon, no pulsing.

aeoncleanse:
3: Auto-OC works during upgrades. It shouldn't, I think, as you can't fire OC while upgrading.
4: If you toggle while it's auto-upgrading, the toggle breaks and the ACU gets confused. It fires OCs when it perhaps shouldn't.

Crotalus:
1, 2, 3, 4 hopefully fixed with latest commits, there is still corner cases I would guess. For instance auto-OC can fire while acu is about to get loaded into a transport, something which isn't possible with normal orders. Maybe can change the state check from 'Enhancing' to explicit 'Attacking' / 'Moving'  (if those aren't true when about to load into transport)

aeoncleanse:
No tooltip when mouse over and no E storage.

aeoncleanse:
1 - Tooltip while button is disabled is still broken, displaying incorrect text
2 - ACU out of range of units, just. Toggle it on, toggle it off. Walk in range. OC fires.

aeoncleanse:
Actually, the broken tooltip is on the auto-OC mode, charged or not, not whether you have energy to fire.

Crotalus:
:bell: (I don't think github send notification on new push)

aeoncleanse:
:+1:

aeoncleanse:
Merged as 2c6e79c7b92b612e36d52de23a9e36bcb3959af9

PR #1123, opened 2016-02-16
Base Manager, build before assist with engie

speed2CZ:
If base was destroyed and engieneer got added to base manager (from different base as an expansion engineer), it tried to assist on factories witch lead to error. Now it will build first.


aeoncleanse:
Merged as d36aa398ef3a3f88bb6ef1fc7d484d3d5e8b637a

PR #1124, opened 2016-02-17
Refactor reclaim labels

Crotalus:
* Cluster mass labels close to eachother when zoomed out to get less clutter.
* Don't reposition labels if camera zoom hasn't changed

Optimal would be another cluster ring texture - maybe something with dark green border / filled light green (with alpha for transparency) and a mass icon. Using the assist ring gfx at the moment.

Maybe "color" the wreckages to show exactly what cluster ring a certain wreckage belongs to in case several are overlapping eachother but may be overkill.

Another feature could be to be able to tune the radius / distance variables in interface options


Crotalus:
Made a new ring gfx for the mass clusters, 80 kb image instead of current 1 mb one

3ab0a8ab3cc726e10b786b0b5199fa029b45c859

PR #1128, opened 2016-02-22
Fix SetFormation function

speed2CZ:
It has to be in PlatoonData. See
https://github.com/FAForever/fa/blob/develop/lua/AI/AttackManager.lua#L236


aeoncleanse:
Merged as ef62afb7ad0e9eb3d7e5bc0355fb00a1fa9c5528

PR #1129, opened 2016-02-22
Add Carrier class into generating naval

speed2CZ:


aeoncleanse:
Merged as d9dcb0359c1c2ff529277440786afa1d3d221e53

PR #1132, opened 2016-02-23
UEF, Cybran T2 air fighter template for AI

speed2CZ:
Platoon templates for campaign AI that will allow it to build UEF and Cybran T2 air fighters. Also removed build condition for T3 torp bombers so AI can build it even if player has no navy.


aeoncleanse:
Merged as cd6189476abbbd5d788c41ab2e52b1de7085c517

PR #1133, opened 2016-02-23
Add NukeSubs into generating naval

speed2CZ:
Support for Nuke subs, disabled by default since they are not wanted
everytime.


aeoncleanse:
Merged as bf9be787bff2528dfe7e8b8c5d934da60c021167

Issue #1138, opened 2016-02-29
Build range circle


speed2CZ:
Could we make this build range circles join into one when selected more units as others range circles do? And add it to the Strategic Overlay menu to toggle it there?
http://puu.sh/npRtD/18e7c7a1fc.png

Crotalus:
Current build range circles are an ugly hack which uses a decal with texture to show the build range of an engineer. 

No idea why that solution was chosen, I'll take a look and see if we can get the build range into the overlay system instead

Crotalus:
Take a look at the PR and see if it works as expected

Crotalus:
The miscellaneous range circle was used for engineer stations / air staging earlier, should probably be it's own category in the filter window, otherwise messing up combine-mode for military.

Crotalus:
Resolved for now until someone finds the hidden flaw of the new implementation :gem:

speed2CZ:
Actually I think I found one, its the same range as for refuelling stations. Those should be seperate

Crotalus:
Yeah, unfortunately we can only reuse the existing ones - not create new classes, some alternatives are (order by subjective preference):

* Remove scan radius indicator for refueling station which isn't very useful (already very high radius, maybe increase it to "infinite" anyway?)
* Use the old "fake circle texture" on refueling station instead when selecting it / mouse over it
* Hack: Let refueling station use another radius type, some intel one maybe?. This would be a hack by setting intel on the building but always have it turned off (if a military one is used we create a dummy weapon to render circle).
* Hack2: Turn the circles on / off depending on what's selected or not,

speed2CZ:
I guess removing it would be the best idea with "infinite" range.

Crotalus:
The dock button already seems to be infinite range, I guess it's the suck-in-patrolling-planes that use the radius

Crotalus:
So, If you click "dock" with the planes they always go to the nearest refueling place, this is not the case when patrolling planes are sucked in by a guarding air staging unit.

Currently the default air-staging guard radius is 300 which means they already cover quite a bit of a map, so if you have several staging units changes are it won't refuel at the nearest one possible. This will get even worse by setting guard radius even higher.

I'm starting to think the logical thing is to actually LOWER StagingPlatformScanRadius by a lot, thus forcing players to actually have the patrol route near / over a refueling place, that way they will have control of the mechanics.

Crotalus:
I move this to a new issue

PR #1139, opened 2016-02-29
Refactor/buildrange

Crotalus:
Use overlay system instead of ugly hack 

#1138


speed2CZ:
This works great :+1:

Crotalus:
09580a6b96c14839c1c092b879667ab783d16172

PR #1140, opened 2016-02-29
Fade out wreckage when reclaiming it

Crotalus:
Shade the wreckages so they fade out (instead of popping out of existance) as they are being reclaimed, main part of fading occurs at the end of the reclaim.


speed2CZ:
:+1:

Crotalus:
c9d24e4862ca7db489aaf7cc0578547d61bd8c10

Issue #1127, opened 2016-02-20
Allow bind keys for Show Reclaim


speed2CZ:
This feature is nice but I'd really like if I could set it on different keys than Ctrl+Shift

shalkya:
+1

Crotalus:
Try out the PR

Crotalus:
2c7ca39 made this possible with UI-mod

PR #1142, opened 2016-03-02
Toggle reclaim labels

Crotalus:
#1127 

Also makes the reclaim labels UI-moddable


Crotalus:
Reclaim labels UI-moddable now instead

Issue #1144, opened 2016-03-03
Auto-refuel of patrolling planes


Crotalus:
So, If you click "dock" with the planes they always go to the nearest refueling place, this is not the case when patrolling planes are sucked in by a guarding air staging unit.

Currently the default air-staging guard radius is 300 which means they already cover quite a bit of a map, so if you have several staging units chances are it won't refuel at the nearest one possible - so a plane could go 300 km and refuel instead of the carrier 5 km away. 

I'm starting to think the logical thing is to lower StagingPlatformScanRadius by a lot, thus forcing players to actually have the patrol route near / over a refueling place, that way they will have control of the mechanics. With such a change we have to make sure AI doesn't rely on "patrol-refuel" patrolling planes no matter how far they are from staging

Uveso:
fixed, can be closed.

micheljung:
Where has it been fixed? How has it been fixed? Did someone review? So many questions..

Uveso:
1:
It was fixed inside:
Blueprints.lua
UAA0310_unit.bp
UAB5202_unit.bp
UAS0303_unit.bp
UEB5202_unit.bp
UEL0401_unit.bp
UES0401_unit.bp
URB5202_unit.bp
URS0202_unit.bp
URS0303_unit.bp
XEB0204_unit.bp
XRB0104_unit.bp
XRB0204_unit.bp
XRB0304_unit.bp
XRB3301_unit.bp
XSB5202_unit.bp
XSS0303_unit.bp
(commit: Crotalus:airstaging-fixes)

2:
Default StagingPlatformRadius was reduced from 300 to 50, 
OverlayMisc was removed from airstaging units and 
the radar overlay is now used for perimeter range.

3:
Yes. i did. All changes to the blueprints are not complex and should work. Also the change inside blueprints.lua is only marginally and easy to check while only reading it. This will work.

Exotic-Retard:
i dont really get why overlaymisc was removed. surely that would help to explain why your planes sometimes do and sometimes dont refuel. how it is currently sounds like it could cause a lot of confusion. having a short range only makes this more needed not less imo :/

(also a semi-separate issue is that overlaymisc is used to show buildranges, and so its colour is changed from purple to ugly dark yellow. i would rather change the colour back to purple and then use it properly as an overlay for.... miscellaneous ranges : D )

Crotalus:
[...]

PR #1145, opened 2016-03-03
Airstaging fixes

Crotalus:
Refactor airstaging radius based on #1144


Crotalus:
More logical this way when patrolling planes actually use stuff in their patrol path

PR #1147, opened 2016-03-04
Simpler method for nuke damage

aeoncleanse:
No need for the extra file and flexibility when this function is unlikely ever to be used for non-nukes due to expense and power.


Crotalus:
I guess it depends on what readahead / imagination you're using when guessing about need of flexible code.

The hilarious thing in this case is that first you want someone else to do the "heavy-lifting"  (which in this case was trivial)  due to lack of time and then coming in like a dark knight deciding that flexibility needs to be reduced and spend unnecessary brain cycles on a change like this.

:laughing:

aeoncleanse:
Hahaha hey, I had uni work :)

PR #1156, opened 2016-03-17
No lag in singleplayer

Crotalus:
Removes command lag for singleplayer games (vs AI / sandbox etc).

One objection could be that uber-sandbox nerds complain their orders go through too fast compared to a real network game. Against AI that isn't an issue, could be alleviated by checking win conditions or some option in gameplay.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -93,6 +93,7 @@ Patch 3652 (pending)
 - Tweaks to hive build effects to reduce performance impact
 - Cleanup of enhancement sync code to make it faster and more secure
 - Entities now re-use repeated sound FX instead of creating new ones every time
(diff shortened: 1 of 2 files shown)


aeoncleanse:
That objection seems dumb to me. @Sheeo are there any possible unknown side effects of this???

Sheeo:
No, there aren't. Frankly I'm amazed this wasn't already the case.

I wouldn't want the changelog to say this though:

    100% command responsiveness in singleplayer mode

Because commands are "100% responsive" even without .5s added lag?

aeoncleanse:
Merged as cfe4dab93011cdb428d2dba864e67287e80e50eb

PR #1157, opened 2016-03-17
Reduce load on render thread

Crotalus:
Current clamp is set at max 100 fps, users running with vsync off but with an
adapter @ 60 hz could get a waste up to 40 fps. With this change we try to find out the current adapter's hz setting so we don't render unecessarily.

A test with 1000 patrolling UEF planes where I had 100 fps @ 60 hz  (render thread maxxed out @ 100% CPU) resulted in 30% less cpu usage for that thread. This was on a 4-core / 8 ht machine so someone running 2 cores or less would probably see even more.

This will also take care of crazy users who manually set it to 0 to "max out fps" while not understanding that they will just stall cpu while rendering stuff which isn't shown anyway due to adapter limit.

TA should be happy.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -93,6 +93,7 @@ Patch 3652 (pending)
 - Tweaks to hive build effects to reduce performance impact
 - Cleanup of enhancement sync code to make it faster and more secure
 - Entities now re-use repeated sound FX instead of creating new ones every time
+- Reduce load on render thread
 
 **Other**
 - Added game-side support for future achievement system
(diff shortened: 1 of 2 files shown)


aeoncleanse:
Well... I was sceptical, but this should work. Guess I'll have to test it to make sure people who DO run at 100FPS such as myself aren't accidentally hit.

Crotalus:
Current clamp is 100 fps so for you there should be no difference, question is how the 120 / 144 hz guys currently are experiencing things. With the current 100 clamp something like 60 or 72 hz could give less tearing.

aeoncleanse:
Many of those with higher refresh rates will also be running async, so no worries. I still need to check mine just to make sure this code actually works :p

aeoncleanse:
Merged as 1abbf1f3c7ffc60fe0b9b0d0944d9f17ad541d4f

PR #1160, opened 2016-03-20
Refactor labels for performance + key for toggling

Crotalus:
* Get UI draw calls out of UserSync (AddReclaim / RemoveReclaim) to minimize sync impact
* Labels don't redraw their position when not needed (camera unchanged)
* When over 1000 labels on map they will hide temporarily while zooming in / out to prevent major frame-drop.
* Possible to bind a key for toggling the labels.


aeoncleanse:
This will wait for 3653 minipatch

PR #1162, opened 2016-03-21
Fix bug with intel effects at stall

### LINKED ISSUE #1155: 3652 - strange static stealth stuff
note: i could not replicate this in 3650. only the beta patch.
when you start energy stalling, stuff switches on/off repeatedly.
stealth is one of those things, and thats fine. so when you are e stalling it jumps up and down, and turns on /off

the issue is when you manually disable stealth generators (stationary) they continue behaving as if they are enabled! they dont drain e but they do start/stop their animations and effects. 
this is OBVIOUS with the aeon stealth gen, but true for all factions.

and then you stop energy stalling - then all stealth enables, but still being manually disabled
they all look like they are enabled, but dont drain e and dont provide stealth (you can still tell by the toggle and the strategic icon being crossed out)

whats perhaps more strange is that they toggle/trigger/something the OnIntelEnabled/OnIntelDisabled bits inside the unit scripts - which is visually annoying but if you have functionality in there then its functionally annoying too.

heres a picture of my disabled but unpacked aeon stealth generator: http://puu.sh/nE3mn/ff58aa9977.jpg

no sim mods were running when i was testing this (i know what you're all thinking!)

Crotalus:
Preferably tested by someone else also before merging it

Fixes #1155

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2320,7 +2320,7 @@ Unit = Class(moho.unit_methods) {
             return intEnabled
         end
     
-        local intEnabled = false
+        local intEnabled
         
         -- We need this guard because the engine emits an early OnLayerChange event that would screw us up here.
         -- The NotInitialized disabler is removed in OnStopBeingBuilt, when the Unit's intel engine state is properly initialized.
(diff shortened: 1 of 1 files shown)


aeoncleanse:
I think I've come up with a much cleaner way. Problem was the intelwatchthread still being alive and calling this even when the unit was toggled off.

PR #1163, opened 2016-03-21
UnitsAroundPoint -> GetUnitsInRect

Crotalus:
Ok, now it got obvious why `GetUnitsAroundPoint()` is located on the brain object. 
It takes brain-stuff into account. In this case (without any arg) it will only return visible units for the specific brain. Function is probably meant to be used when an AI is grabbing stuff to make decisions / react.. `GetUnitsInRect` just grabs units regardless of brain intel / allegiances 

Very non-obvious from the superb LUADOC ;)


aeoncleanse:
Merged b33d902a738852c3b7dd3d5089ea1af1dfb85241

Issue #1186, opened 2016-03-31
in game mod manager showing mods as requiring dependencies.


Exotic-Retard:
http://puu.sh/o0X8c/afb2a4c5af.jpg

see picture for explanation - in lower right corner we see i have 11 ui mods activated, in the mod manager i have 9 since it thinks that those 2 are disabled because they lack dependencies (which i have). the mods also run in game. so this is a purely display bug it seems.

hope this helps

HUSSAR-mtrela:
I will try to reproduce this issue. Those disabled mods might depend on a newer version of the common mod than v1.0 that you have. Can you check if those disabled mods have UID in "requires" table (mod_info.lua) and it is matching UID of the common mod?

Exotic-Retard:
sorry for not replying earlier. the UIDs match for both mods. 
requires = {"zcbf6277-24e3-437a-b968-Common-v1"}
is in both and the uid of the common mod tools that i have is: 
uid = "zcbf6277-24e3-437a-b968-Common-v1"

since those mods are working in game, i kinda expected them to be set up properly.

hope this helps

aeoncleanse:
@HUSSAR-mtrela  -  We should also be using the right buttons in the Mod Manager. It should be consistent with the buttons used throughout the rest of the lobby.

HUSSAR-mtrela:
thanks @Exotic-Retard for checking uid 
@aeoncleanse Well, they are not just click buttons but rather custom toggle buttons because standard toggle buttons are too big since their are sized by their background image. I will see if I can change the background texture so it look like more like in other buttons.

aeoncleanse:
The standard buttons are fine, trust me, and they work fine for toggles too. They can be easily arranged relative to the background group, or anything else for that matter :) Alternatively, use a simple checkbox as elsewhere, but I won't be shipping with these buttons, they're far too out-of-place. Just check how similar functions have been done elsewhere in the lobby. Pretty sure we had toggles for this too.

HUSSAR-mtrela:
![mods-filters](https://cloud.githubusercontent.com/assets/11665978/14415325/c545afc0-ff6b-11e5-9796-d7ab9cec5fab.png)

I updated styling of filter buttons and now they match other buttons

aeoncleanse:
Nicely done :)

HUSSAR-mtrela:
[...]

Issue #990, opened 2015-10-28
New Upgrade system ignores unit restrictions


speed2CZ:
Noticed in coop that you are allowed to upgate mexes/radars etc even when those units are restricted

Crotalus:
And the unit actually upgrades? If so it the UI is responsible for enforcing restrictions = has to change

speed2CZ:
yes they do upgrade

duk3luk3:
And they don't explode after upgrading?

speed2CZ:
no

HUSSAR-mtrela:
i will try to fix this issue and I think it should be fixed in sim/unit rather than in UI otherwise any UI mod would exploit this issue.

HUSSAR-mtrela:
@aeoncleanse @speed2CZ 
so there are  two things that need to be fixed for this issue:
1. ScenarioFramework.AddRestriction() funtion should pass restricted categories to game.lua and update the restrictedCategorySet variable and override UnitRestricted() funtion. 
2. Somewhere in Unit.Lua, we need to check if a unit has restricted upgrade:
- OnStartBeingBuilt
local isRestriced = Game.UnitRestricted(self:GetUnitId(), self)
- or OnStartBuild 
local upgradeID = self:GetBlueprint().General.UpgradesTo
local upgradeBP = __blueprints[upgradeID]
local isRestriced = Game.UnitRestricted(upgradeID, upgradeBP)

not sure how to stop building of the restricted unit such that it does not start again?
self:OnStopBuild() or self:Kill()

So far, I got part number 1 working and I need a bit help with part number 2.

aeoncleanse:
Could simply return end in onstartbuild? If not, self:Destroy().

HUSSAR-mtrela:
@speed2CZ @aeoncleanse 
I got a fix for this issue and I will commit it tomorrow night because I need to cleanup and remove a lot of my debug changes. Should I log a warning message if someone is trying to cheat by building restricted unit? I mean this warning might be useful for someone setting restriction in a scenario and then wondering why units are blowing up when running this scenario.

aeoncleanse:
Go ahead and merge the stuff in the coop mod @speed2CZ . With that, this this then addressed as of @HUSSAR-mtrela 's latest?

aeoncleanse:
a3e78b63929ea98f07f02f8d990adf586dee1a8b tested all working. Changes to maps required, but simple.

HUSSAR-mtrela:
[...]

Issue #798, opened 2015-07-20
Don't allow construction of drones when engineering station restriction is on


Sheeo:
https://github.com/FAForever/fa/tree/feature/no_drones starts work on this.

TODO:
- [x] Modify UI to disallow construction of drones when the restriction is enabled
- [x] Add localizations

HUSSAR-mtrela:
Sheeo, I would like to work on this issue if you dont mind,

Sheeo:
Of course, by all means :) :+1:

aeoncleanse:
7feef8fa305d073b0f77bfb3da18e7b5fa4f060d

Sheeo:
TODO list not completed on this issue.

aeoncleanse:
Which UI is in need of modification??

HUSSAR-mtrela:
I think that he means UI for unit construction

HUSSAR-mtrela:
closing because this issue was fixed with PR #1187 and a few hotfixes that followed it.

PR #1187, opened 2016-04-02
Fixing scenario/UI restrictions

HUSSAR-mtrela:
### Fixes for issues:
- [x] #990 New Upgrade system ignores unit restrictions 
- [x] #798 Don't allow construction of drones when engineering station restriction is on
- [x] Setting unit restrictions in scenario script does not apply to specific army
- [x] #1186 In-game mod manager showing mods as requiring dependencies
- [x] #1199 Units Manager is using an old icon for Seraphim faction

### Other changes:
- improved performance of checking for restricted units
- added validation to functions that adds and removes restrictions


aeoncleanse:
How will this interact with @duk3luk3 's changes? Is it meant to stack on top of them?

lua/game.lua
```diff
@@ -102,27 +102,61 @@ function GetConstructEconomyModel(builder, targetData, upgradeBaseData)
...
+--- NOTE do not call this function to set build restriction on an unit (e.g. factories/carriers)
+function AddRestriction(categories)
+    
+    if type(categories) ~= 'userdata' then
+        WARN('Game.AddRestriction() called with invalid argument type: ' .. type(categories))
     end
+    -- reset units cache every time new restriction is added
+    evaluatedUnits = {} 
+    --LOG('Game.Restrictions adding ' .. tostring(categorySet)) 
+
```
aeoncleanse: Don't need the commented logging code

lua/game.lua
```diff
@@ -102,27 +102,61 @@ function GetConstructEconomyModel(builder, targetData, upgradeBaseData)
...
+    end
+    if not unit then
+        WARN('Game.UnitRestricted(Unit) called with nil argument.')
+        return false
+    end
+    local unitID = unit:GetUnitId() 
+    if unitID == nil then
+        WARN('Game.UnitRestricted(Unit) called without unit class argument.')
+        return false
+    end
```
aeoncleanse: Do we want to be returning false (affirmative in this case) here instead of crashing out? These error conditions should never be allowed right?

lua/sim/Unit.lua
```diff
@@ -1931,10 +1931,13 @@ Unit = Class(moho.unit_methods) {
...
-        -- If someone thinks they're being clever by using a UI mod to violate unit restrictions,
-        -- thoroughly ruin their day.
-        if Game.UnitRestricted(self:GetUnitId(), self) then
-            self:Kill()
+        -- prevent UI mods from violating game/scenario restrictions 
+        local id = self:GetUnitId()
+        local bp = self:GetBlueprint()
+        if Game.UnitRestricted(self) then
+            WARN('Unit.OnStopBeingBuilt() Cannot create restricted unit: ' .. (bp.Description or id) )
+            if self ~= nil then self:Kill() end
```
aeoncleanse: Kill() will leave reclaim. Best to use Destroy() I think.

lua/sim/Unit.lua
```diff
@@ -2179,6 +2182,18 @@ Unit = Class(moho.unit_methods) {
...
     OnStartBuild = function(self, built, order)
+        
+        -- prevent UI mods from violating game/scenario restrictions 
+        local id = built:GetUnitId()
+        local bp = built:GetBlueprint() 
+        if Game.UnitRestricted(built) then
+            WARN('Unit:OnStartBuild() Cannot build restricted unit: ' .. (bp.Description or id) )
+            self:OnFailedToBuild() -- don't use: self:OnStopBuild()
+            built:OnFailedToBeBuilt()
+            built:Kill()
```
aeoncleanse: Here too, need to check if this leaves reclaim

aeoncleanse:
OK so this really doesn't work. It gives a repeated destruction loop on build start, and the UI is not blocked. 

```    ScenarioFramework.AddRestriction(Player, categories.TECH2)
    ScenarioFramework.AddRestriction(Player, categories.TECH3)```

That's the code used. Curiously, this bit: ScenarioFramework.RestrictEnhancements({'ResourceAllocation', 'Teleporter'}) works perfectly.

**Comment by aeoncleanse** [comment]
Right, so I had forgotten that this is the sim-only half, and apart from the death loop, it does at least prevent the build. We now need to fix the broken UI, which is ONLY broken for units which use the chain-upgrade feature.

File: lua/game.lua (hunk 2 above)
Comment (HUSSAR-mtrela) [inline lua/game.lua]: i can change it to an error

File: lua/sim/Unit.lua (hunk 3 above)
Comment (HUSSAR-mtrela) [inline lua/sim/Unit.lua]: I will try it using Destroy() instead

File: lua/sim/Unit.lua (hunk 4 above)
Comment (HUSSAR-mtrela) [inline lua/sim/Unit.lua]: Indeed, we should not leave any reclaim

File: lua/game.lua (hunk 2 above)
Comment (aeoncleanse) [inline lua/game.lua]: No I mean currently when the code fails, it looks to the game like it successfully returned a restricted unit. We don't want that. We want it to know it failed, and stop the sim, no?

**Comment by HUSSAR-mtrela** [comment]
Restriction of enhancements is different but much simpler than unit restrictions. I will look into if I can fix the issue with the death loop and fix the chain-upgrade feature in UI.
Yes, my changes are meant to stack on top of changes from @duk3luk3 but he might have to modify his code for classes that inherit from Unit class and use Game.UnitRestricted(self) I will review his changes in detail once I finish fixing above issue.

**Comment by HUSSAR-mtrela** [comment]
@aeoncleanse FYI, I found a fix for the issue with chain-upgrade in UI and I will commit my changes later today. It require a few minor changes to the restriction system.

**Comment by aeoncleanse** [comment]
Nice one, I look forward to seeing your solution :)

**Comment by HUSSAR-mtrela** [comment]
@aeoncleanse I finished fixing the issue with chain-upgrade and I improved performance of handling of game/scenario restrictions. Now you can get a table with ids of restricted units, like you suggested in our last skype session. Also, I added logic for checking for units restricted by a unit lower in the upgrade chain, for example T2 mex restriction will also restrict/prevent building of T3 mex 
Since It was possible (until now) to circumvent our restriction logic on UI side because we were checking restriction on T2 mex but not units that are higher in the chain-upgrade. My fixes now works even with units that have multiple levels of upgrades, e.g. Cybran shields.

**Comment by speed2CZ** [comment]
There are several issues with this: 
It won't allow the script to create restricted unit, for example in my mission I have intro where player get's moved on T2 transport but he has restriction to just T1 units. So it won't create the transport.

Other issue is that it doesn't restrict unit's to army which is specified, but to all of them. AI was also unable to to build anything I've restricted for players.

**Comment by aeoncleanse** [comment]
Good points both, but it's pretty important we stamp out PVP cheating before worrying about AI. Can this not be worked around by temporarily un-restricting the T2 Transport, and making sure missions never allow the AI to have something the player doesn't?

**Comment by speed2CZ** [comment]
That is totally not possible.

**Comment by speed2CZ** [comment]
Player and AI have almost always have different unit restrictions. And giving players unitss he's not able to build as reinforcement or so is also very common.

**Comment by aeoncleanse** [comment]
Is it the UI stuff or the sim-side killing stuff which is the trouble? If it's the sim we can test for skirmish vs PVP and only use that for PVP?

**Comment by HUSSAR-mtrela** [comment]
Alright, I was not aware of these kind of requirements for the restriction system. So one should be able to do the following in a scenario script:
- [x] Set restrictions for all players, and/or
- [ ] Set restrictions per a player/AI 

I can correct it such that restrictions set per a player are respected and I should be able to add it tonight.
@speed2CZ as to the issue with T2 transport, your restriction should look like the following to restrict all tech 2 units except T2 transport:
`ScenarioFramework.AddRestriction(army, categories.TECH2 - (categories.AIR * categories.TRANSPORT))` 

In order to give a player reinforcement units that are restricted. you should remove restriction just before giving reinforcement and then add them back in:
`ScenarioFramework.RemoveRestriction(army, categories.TECH2)`
`-- add here code for giving reinforcement units`
`ScenarioFramework.AddRestriction(army, categories.TECH2)`

This way reinforcement units will not be destroyed by the Sim when transferring them and the player will still not be able to build them except of using reinforcement units until they die to enemy attacks.

**Comment by speed2CZ** [comment]
Set restrictions per a player/AI 
Do this part rather to set restriction just to that army, it's not just Player/AI have different restrictions, it's different armies have different restrictions.

The part with unrestricting the units could be done in the functions that is creating units/giving to player I guess? Thing is I dont know how to do that. so if you would want to look into that. I know it's setting to ignore unit cap when doing so. So adding reming restrictions for the units it's spawning.

Giving units from different army: https://github.com/FAForever/fa/blob/release/3652/lua/ScenarioFramework.lua#L218
Creating units https://github.com/FAForever/fa/blob/release/3652/lua/sim/ScenarioUtilities.lua#L803
There are more functions for creating units like CreateArmyUnit, CreateArmyGroup, CreateArmyGroupAsPlatoon and dunno in which it would need to be set.

**Comment by HUSSAR-mtrela** [comment]
Thanks @speed2CZ for the info about Giving units to different armies. Yes, you are right about setting unit restrictions per an army instead of player/AI and I will fix it soon,

I think that modifying these files will just make it more complex code and potentially introduce more issues. We should keep current behavior that prevents map scripts from giving restricted units to armies. And it should be up to map creators to toggle restrictions if they are trying to transfer restricted units:
`ScenarioFramework.RemoveRestriction(army, categories.TECH2)`
`-- add here code for giving reinforcement units`
`ScenarioFramework.AddRestriction(army, categories.TECH2)`
@aeoncleanse Do you agree with above or should I introduce complexity in code that transfers units?

**Comment by aeoncleanse** [comment]
I absolutely think that it should be dealt with in the map script. Having a robust PVP system is much more important than saving a few lines of code in the campaign maps. Since the campaigns bypass the UI-level restrictions anyway and it's the sim stuff, it means it's only in the way for spawning and/or transferring (Which involves spawning) restricted units. Therefore the best solution to me is that we put a function into ScenarioUtilities.lua or wherever which turns off ALL restrictions. Then you spawn your stuff. Then you re-restrict everything. In theory this should be as simple as emptying and re-filling a table, and if we have the system revised to be per-army, there's no risk of exploits emerging either.

**Comment by speed2CZ** [comment]
```
WARNING: ScenarioFramework.AddRestriction() called with invalid army argument "1" instead of army name, e.g. ARMY_1 
```
Code for all missions looks pretty much like this:
```
ScenarioInfo.Player = 1
local Player = ScenarioInfo.Player

ScenarioFramework.AddRestriction(Player, categories.TECH2)
```
same for other armies

**Comment by HUSSAR-mtrela** [comment]
@speed2CZ it looks like skirmish scenarios use army name and mission scenarios use army index to set units restrictions for a given army. For example:
`ScenarioFramework.AddRestriction(1, categories.TECH2) -- in mission`
`ScenarioFramework.AddRestriction("ARMY_1", categories.TECH2) -- in skirmish `

So I changed logic to support both methods of setting restrictions and now you should not get warning messages about invalid army argument.

**Comment by speed2CZ** [comment]
Could you now make the code as IceDreamer suggested to unrestrict thing when spawning units and then restrict it again?

Also I've tried this code on random mission (UEF M1) and got this error at the start:
```
WARNING: BeginSession() failed: attempt to get as UserData a boolean value
         stack traceback:
         	[C]: in function `EntityCategoryFilterDown'
         	c:\users\multimedia\documents\github\fa\lua\game.lua(261): in function `ResolveRestrictions'
         	c:\users\multimedia\documents\github\fa\lua\game.lua(181): in function `RemoveRestriction'
         	...imedia\documents\github\fa\lua\scenarioframework.lua(1033): in function `RemoveRestriction'
         	...imedia\documents\github\fa\lua\scenarioframework.lua(2249): in function `RemoveRestrictionForAllHumans'
         	...alliance\maps\scca_coop_e01\scca_coop_e01_script.lua(88): in function `OnStart'
         	...users\multimedia\documents\github\fa\lua\siminit.lua(169): in function <...users\multimedia\documents\github\fa\lua\siminit.lua:160>
         	...users\multimedia\documents\github\fa\lua\siminit.lua(277): in function <...users\multimedia\documents\github\fa\lua\siminit.lua:273>
         	...users\multimedia\documents\github\fa\lua\siminit.lua(367): in function `BeginSession'
```
I guess removing restriction is a problem here?
The code from the mission:
https://github.com/FAForever/faf-coop-maps/blob/master/SCCA_Coop_E01/SCCA_Coop_E01_script.lua#L88

And the function it uses:
https://github.com/FAForever/fa-coop/blob/master/mods/coop/hook/lua/ScenarioFramework.lua#L19

**Comment by HUSSAR-mtrela** [comment]
Yes, I can make this change and fix issue with removing restriction on multiple human players. Thanks for links to the code files which helped me quickly track root cause of this issue.

**Comment by HUSSAR-mtrela** [comment]
Now, you should be able to transfer restricted units using this code:
` ScenarioFramework.IgnoreRestrictions(true)`
` ScenarioFramework.GiveUnitToArmy(unit, armyIndex)`
` ScenarioFramework.IgnoreRestrictions(false)`

**Comment by speed2CZ** [comment]
Shouldn't this be put into those fucntions that are giving/creating units? They already ignore unit cap, when used you always want to units get spawned/given.

**Comment by aeoncleanse** [comment]
@speed2CZ - No. While we have control over some of the unit creation tools, we also want this system to work for the engine creation commands. Yes it's going to mean a lot of changes in coop, but it's not as bad as it could be. Most stuff is spawned in one stage at start of mission or mission stage, or at odd, specific events. So just turn restrictions off on map load, let everything spawn, then restrict again.

**Comment by aeoncleanse** [comment]
To save space with when you want to just do one bit at a time, also could write a new pair of functions: CreateArmyGroupIgnore, GiveUnitToArmyIgnore, which is literally just disable, old command, enable, but in one line.

**Comment by aeoncleanse** [comment]
Merged in up to cb0be5230063021124bb9aa538c5611c18a156ae as d2510e124c363e910570164b761665a59493fb4a 14/04/16

**Comment by HUSSAR-mtrela** [comment]
@aeoncleanse should I create a new PR for my last 3 commits or can you merge them to release branch?
- https://github.com/FAForever/fa/pull/1187/commits/6a1e8d7f59fcc0f266614b2221e6bbcad07e19b1
- https://github.com/FAForever/fa/pull/1187/commits/e8d7c1a03244f070918020303080dd05ffeff7c4
- https://github.com/FAForever/fa/pull/1187/commits/fe24a3fc1fe8fe23df1ce8246e168447aa9706c5

**Comment by aeoncleanse** [comment]
I'll merge them in once I review them

File: lua/ui/lobby/ModsManager.lua [hunk 5]
```diff
@@ -536,29 +550,24 @@ function RefreshModsList()
...
+                    mod.tags['DISABLED'] = true
+                else 
+                    mod.sort = 'GAME'
+                    mod.tags['GAME'] = true
+                    if mods.activated[uid] then
+                        mods.sim.active[uid] = mod
+                    else
+                        mods.sim.inactive[uid] = mod
+                    end
+                end
```
HUSSAR-mtrela: above change simplifies if statements and correctly tags mods with missing dependencies

lua/ui/lobby/ModsManager.lua
```diff
@@ -721,23 +728,28 @@ end
...
-           mods.activated[b.modInfo.uid] then 
+                   mods.activated[b.modInfo.uid] then 
+           return false 
+        elseif a.modInfo.sort == 'UI' and 
+               b.modInfo.sort == 'GAME' then 
+           return true 
+        elseif a.modInfo.sort == 'GAME' and 
+               b.modInfo.sort == 'UI' then 
            return false 
         else
```
HUSSAR-mtrela: these additional elseif statements will ensure that UI mods are listed before Game mods. In next version I will split all mods to two scroll groups (UI and Game) so it will be easier and faster to find mods.

lua/ui/lobby/ModsManager.lua
```diff
@@ -241,24 +241,24 @@ function CreateDialog(parent, availableMods, saveBehaviour)
...
     
     position = position + 110
     local filterDisabledMods = CreateModsFilter(dialogContent, modsTags.DISABLED)
     Tooltip.AddControlTooltip(filterDisabledMods, { 
         text = 'Filter Disabled Mods', 
         body = 'Toggle visibility of all disabled mods in above list of mods.' } )
-    LayoutHelpers.AtRightIn(filterDisabledMods, dialogContent, position)
-    LayoutHelpers.AtBottomIn(filterDisabledMods, dialogContent, 10)
+    LayoutHelpers.AtLeftIn(filterDisabledMods, dialogContent, position)
+    LayoutHelpers.AtBottomIn(filterDisabledMods, dialogContent, 15)
```
HUSSAR-mtrela: I re-positioned mod filters to the right because other dialog messages (e.g. game/lobby options) have OK/cancel buttons on the right side and additional buttons on the left.

speed2CZ:
the code for removing restrictions that is in the mission scripts not longer causes warninigs but it doesnt work, units are still restricted

aeoncleanse:
a3e78b63929ea98f07f02f8d990adf586dee1a8b

PR #1188, opened 2016-04-04
Reduce energy drain with adjacency

### LINKED ISSUE #1185: Ravager Adjacency
When reloading Ravager uses -50E/s but if you add T1 pgen next to it it raises to -55E/s and then with more pgen the value is decreasing but from -55
0 pgens -50
1 pgen  -55
2 pgens -54 etc

Some other random data: 1 T3 pgen -51E/s, 1 T2 Pgen -52 
so right now no adjacency best adjacency :D

Crotalus:
Change energy drain as well with adjacency so we get the UI to show the actual
energy cost reduction for a weapon with adjacency.

Fixes #1185


Sheeo:
Looks like we'll want this commit in develop?

Issue #1183, opened 2016-03-29
Code Cleansing


HUSSAR-mtrela:
Go through all LUA code files and clean them up:
- replace "##" with "--"
- replace "!=" with "~="
- replace tabs with spaces
- remove multiple blank lines with single blank line
- fix typos in comments
- fix indent of source code 
- did I miss something?

Best time to do these changes, is just after 3652 patch

HUSSAR-mtrela:
@speed2CZ please add labels: 3653, nice to have, code style,

aeoncleanse:
We operate a policy of 'Fix it when you use it' to avoid polluting the blame and history.

Sheeo:
Indeed, I did not actually read the issue when adding labels.

We want to avoid want huge commits that only do this.

yousless:
Hi,
Since I don't know about lua programming maybe I can help you with this. I have some questions though..

Looking at the files I noticed that some times ## were used like this:
##################################################
# Comment
##################################################
Should I replace them with hyphens here also.

How many spaces should I replace a tab with?

I will look on how to spell check comments later..

For indenting I am not sure how good emacs lua-mode is..

aeoncleanse:
If you're affecting that file, yes. I try personally to remove comment walls wherever not totally necessary, and shrink them everywhere else. I also make sure to have -- Comment, with both the space and the capital letter, but I'm just pedantic :)

HUSSAR-mtrela:
closing this issue since we agreed that clean up should happen on a file basis or when needed.

PR #1194, opened 2016-04-10
Code cleansing

yousless:
As per issue  Code Cleansing #1183

I cleaned up the code a bit by doing the following:
Replaced ## with --
Replaced Tab with four spaces
Replaced != with ~=


HUSSAR-mtrela:
@yousless, you should have done this after we release 3652 patch. Because now your changes will conflict with all pending pull requests for 3652 patch. Also, why did you add new script files? e.g. AANDepthCharge03_script.lua

yousless:
@HUSSAR-mtrela, I will work on it after the patch then. Concerning the scripts like AANDepthCharge03_script.lua, it is not really a new a script (the original is under ./projectiles/AANDepthCharge03/). Apparently something went wrong while I am copying recursively.

aeoncleanse:
This wouldn't be merged anyway. As was clearly stated in the issue, we clean things up _as we come to them_, not in one big go.

yousless:
Sure. Never mind I was trying to help and sorry for the inconvenience caused here..

HUSSAR-mtrela:
@yousless you should join our FA Forever channel on slack so that you can coordinate your work with us. I'm sure we can find something for you to do even if you don't have programming skills

yousless:
@HUSSAR-mtrela I don't know how to join. I have some programming skills, but no lua :/
I am more into hardware VHDL and C/C++ and I can do some scripting ..

PR #1195, opened 2016-04-12
Hold shift to delay giving away units until orders are complete

Crotalus:
Hold down shift while giving away units to put the actual transfer on the units order queue. 

Will give the units away when current orders has been executed.
(often found a need to do this, moving transports somewhere and then giving them away etc).

The player that gets the units gets a notification ping / sound

(overall we should probably add support for more directed pings, irritating to hear pinging when it has nothing to do with your army / part of the map, just need to find out a good way to handle it UI/control wise so it is easy to decide the behavior)


Crotalus:
Overall we should probably add support for more directed pings, irritating to hear pinging when it has nothing to do with your army, just need to find out a good way to handle it UI/control wise so it is easy to control the behavior.

PR #1200, opened 2016-04-15
Rotate ACU towards mid

Crotalus:
To make start position more equal, otherwise north has a slight advantage when maps is about walking immediately to mid


HUSSAR-mtrela:
we should also apply this change to all T3 artillery structures when they start being built.

speed2CZ:
thats already done in 3652 to all defence structures

Nojoke185:
seems the AI ACU's are rotated to the nearest 90 degrees?
(based on offline test on desert arena with 7 AI)
each AI was rotated to the nearest 90 degree angle it seemed
maybe this change is only applied to human players?


Also tried the ultimate test - gap of 64 ACU's! and some of the player ACU's were not pointed to mid (they were the opposite)
this map may however have some custom stuff happening?


final observation - on launching the game am experiencing a long delay before the launch happens (20-30 seconds)

Issue #1202, opened 2016-04-15
UnitsManager shows incorrect mass production for some units in tooltips


HUSSAR-mtrela:
Reported by @PerfectWay on forum thread:
http://forums.faforever.com/viewtopic.php?f=45&t=11336&sid=41029a21b41b033930ec7cf76e9ac216&p=125309#p125309

I noticed a few issues related to the economic characteristics of the units:
- T2 Mass Fabs show "Mass production: 0", but it should be 1
- ACUs (all factions) show "Mass production: 0", but it should be 1
- SACUs (except Seraphim) show "Mass production: 0", but it should be 1

aeoncleanse:
It's probably this, no?

```local function init(value) 
    return value > 1 and value or 0 
end ```

Why is this here again?

HUSSAR-mtrela:
I'm working on it but I dont think this code is the root cause of this issue. This code ensure that stats don't throw error when a value is nil.

aeoncleanse:
It also ensures that values ranging 0-1 are returned 0, even if == 1. Should it not be return value => 1 ?

HUSSAR-mtrela:
yes, that should be => 1

aeoncleanse:
eda30c36089c41aef50cd13c2df02c71fa70e5e8

HUSSAR-mtrela:
@aeoncleanse please push PR  #1206 because it has a hotfix for this issue

aeoncleanse:
It's done. A close comment with a hash means the fix is merged :)

Issue #1171, opened 2016-03-25
faf beta bug


Lyvnah:
in game lobby, when you swap two players, one of them disappears for a a while and then he show up in observers

aeoncleanse:
Regression? @Crotalus @speed2CZ You seen this at all, or know if it's in normal game too?

speed2CZ:
As a host if you switch someone to your slot, you get moved into obs and the slot bugs out and the person you moved gets invisible and can't switch to obs and back to refresh or anything

aeoncleanse:
So this only happens when swapping a player to your own slot, and does it happen in pre-beta or is it a regression?

speed2CZ:
```
WARNING: Error running OnClick script in CScriptObject at 1150c100: ...lua\ui\lobby\data\watchedvalue\watchedvaluetable.lua(38): attempt to call field `?' (a nil value)
         stack traceback:
         	...lua\ui\lobby\data\watchedvalue\watchedvaluetable.lua(38): in function <...lua\ui\lobby\data\watchedvalue\watchedvaluetable.lua:37>
         	...ta\faforever\gamedata\lua.nx4\lua\ui\lobby\lobby.lua(751): in function `teamSelectionEnabled'
         	...ta\faforever\gamedata\lua.nx4\lua\ui\lobby\lobby.lua(757): in function `SetSlotInfo'
         	...ta\faforever\gamedata\lua.nx4\lua\ui\lobby\lobby.lua(5096): in function `MovePlayerToEmptySlot'
         	...ta\faforever\gamedata\lua.nx4\lua\ui\lobby\lobby.lua(5152): in function `SwapPlayers'
         	...ta\faforever\gamedata\lua.nx4\lua\ui\lobby\lobby.lua(388): in function <...ta\faforever\gamedata\lua.nx4\lua\ui\lobby\lobby.lua:354>
         	...ta\faforever\gamedata\lua.nx4\lua\ui\lobby\lobby.lua(2092): in function `OnClick'
         	...faforever\gamedata\lua.nx4\lua\ui\controls\combo.lua(265): in function <...faforever\gamedata\lua.nx4\lua\ui\controls\combo.lua:262>

```
new one

Crotalus:
Tried to replicate with local LAN game but cannot reproduce in 3652 code at least.

aeoncleanse:
Bug confirmed non-LAN

Crotalus:
[...]

PR #1206, opened 2016-04-19
Hotfixes for GameLobby

HUSSAR-mtrela:
- [x] Corrected if statement that "fixed" issue #1202 
- [x] Adding logs for debugging issue #1171


HUSSAR-mtrela:
@aeoncleanse actually we should have used ">=" instead of "=>" in the init() function :smile: 
Also, these logs might not be enough but we should start with it.

aeoncleanse:
8e8fff77851279a65064be1c9545eb0b7495be33

PR #1207, opened 2016-04-20
Simplify the table manipulation

aeoncleanse:
@speed2CZ need you to test this. In theory, this should finalise all adjustments to restrictions.

### DIFF
--- a/lua/game.lua
+++ b/lua/game.lua
@@ -182,6 +182,7 @@ function IsRestricted(unitId, army)
     if bps.Ignored then 
         return false 
     end
+
     if restrictions.Global[unitId] then
        return true
     end
@@ -215,7 +216,7 @@ local function GetUnitsUpgradable()
     local units = {}  
     for id, bp in __blueprints or {} do 
         -- check for valid/upgradeable blueprints 
-        if bp ~= nil and bp.General and (string.len(id) > 4) and
+        if bp and bp.General and (string.len(id) > 4) and
          ((bp.General.UpgradesFrom and 
            bp.General.UpgradesFrom ~= 'none') or 
           (bp.General.UpgradesTo and 
@@ -237,7 +238,28 @@ function ResolveRestrictions(toggle, cats, army)
     -- initialize blueprints info only once
     if table.getsize(bps.ids) == 0 or
         table.getsize(bps.upgradeable) == 0 then
-        bps.ids = table.keys(__blueprints)
+
+        local used = {
+            BUILTBYTIER1FACTORY = true,
+            BUILTBYTIER2FACTORY = true,
+            BUILTBYTIER3FACTORY = true,
+            BUILTBYTIER3COMMANDER = true,
+            TRANSPORTBUILTBYTIER1FACTORY = true,
+            TRANSPORTBUILTBYTIER2FACTORY = true,
+            TRANSPORTBUILTBYTIER3FACTORY = true,
+        }
+        
+        -- Limit it to unit blueprints inside certain categories
+        for id, bp in __blueprints do
+            if bp and bp.BlueprintId and bp.Categories then
+                for index, cat in bp.Categories do
+                    if used[cat] then
+                        table.insert(bps.ids, bp.BlueprintId)
+                        break
+                    end
+                end
+            end
+        end
         bps.upgradeable = GetUnitsUpgradable()
     end
     
@@ -248,6 +270,7 @@ function ResolveRestrictions(toggle, cats, army)
            restrictions.Global[id] = toggle
         end
     end
+
     if army then
         -- find ids of units restricted for each army
         if not restrictions.PerArmy[army] then
--- a/lua/game.lua
+++ b/lua/game.lua
@@ -103,17 +103,12 @@ local restrictions =  {
     PerArmy = {}, -- set in ScenarioName_Script.lua
 }
 -- stores info about blueprints
-local bps = { 
-    -- table with identifiers of restricted units  
-    restrictions = {
-        Global = {},  -- generated base on restrictions.Global table
-        PerArmy = {}, -- generated base on restrictions.PerArmy table
-        Ignored = false -- flag whether or not to ignore all restrictions
-    },
+local bps = {
     -- table with blueprints that can be upgraded, e.g. T2 shields
-    upgradeable = {}, 
+    upgradeable = {},
     -- table with identifiers of all blueprints, e.g. xab1401 - Aeon Paragon
-    ids = {},  
+    ids = {},
+    Ignored = false,
 }
 
 -- function for converting categories to string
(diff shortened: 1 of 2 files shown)


lua/game.lua
```diff
@@ -253,52 +236,48 @@ local function GetUnitsUpgradable()
...
+        if not restrictions.PerArmy[army] then
+           restrictions.PerArmy[army] = {}
         end
-        for key, category in categories do
-            if type(category) == 'userdata' then 
-                local ids = EntityCategoryFilterDown(category, bps.ids)
-                for _, id in ids do
-                   bps.restrictions.PerArmy[index][id] = true
-                end
+        for key, category in cats do
```
HUSSAR-mtrela: This for loop is throwing error message because `cats` is userdata not a table so you cannot loop over it, instead you should remove this for loop and leave its content:
`--for key, category in cats do`
`local ids = EntityCategoryFilterDown(cats, bps.ids)`
`            for _, id in ids do`
`               restrictions.PerArmy[army][id] = toggle`
`            end`
`        --end`

HUSSAR-mtrela: This for loop is throwing error message because `cats` is userdata not a table so you cannot loop over it, instead you should remove this for loop and leave its content:
`--for key, category in cats do`
`local ids = EntityCategoryFilterDown(cats, bps.ids)`
`            for _, id in ids do`
`               restrictions.PerArmy[army][id] = toggle`
`            end`
`        --end`

lua/game.lua
```diff
@@ -141,84 +136,72 @@ end
...
 end
+
 -- checks whether or not a given blueprint ID is restricted by
 -- * global restrictions (set in UnitsManager) or by
 -- * army restrictions (set in Scenario Script)
 -- e.g. IsRestricted('xab1401', 1) -> checks if Aeon Paragon is restricted for army with index 1
 -- note that global restrictions take precedence over restrictions set on specific armies
 function IsRestricted(unitId, army)
-    if bps.restrictions.Ignored then 
+    if bps.Ignored then
```
HUSSAR-mtrela: for readability purpose, I suggest changing condition in this if statement to: `if restrictions.Ignored then`

lua/game.lua
```diff
@@ -141,84 +136,72 @@ end
...
-    ResolveRestrictions()
+
+    ResolveRestrictions(false, cats, army)
 end
+
 -- toggles whether or not to ignore all restrictions
 -- note this function is useful when trying to transfer restricted units between armies
 function IgnoreRestrictions(isIgnored)
-    bps.restrictions.Ignored = isIgnored
+    bps.Ignored = isIgnored
```
HUSSAR-mtrela: for readability purpose, I suggest changing condition in this statement to: `restrictions.Ignored = isIgnored`

lua/game.lua
```diff
@@ -141,84 +136,72 @@ end
...
 function GetRestrictions()
-    return bps.restrictions
+    return restrictions
 end
 -- sets a table with ids of restricted units { Global = {}, PerArmy = {} }
 function SetRestrictions(blueprintIDs)
-    bps.restrictions = blueprintIDs
+    WARN('SetRestrictions')
+    debug.traceback()
+    table.print(blueprintIDs)
```
HUSSAR-mtrela: above 3 lines should be either removed or commented out because there are left from debugging

HUSSAR-mtrela:
@aeoncleanse please see my comments in your commit https://github.com/FAForever/fa/pull/1207/commits/6dddc615bd04f9a223a5e846cc620000d25f79b5

@speed2CZ I have tested this PR with FA Campaign Mission 2 - Coop (Dawn) and adding/removing restrictions works as expected. However, this PR should also be tested with the "Cybran Mission 2 - Artifact" map because it has a few triggers that remove restrictions using somewhat complex categories. However, I was not able to start this mission because I got the following error message which is not related to the restriction system but rather something is missing in this mission:
```
WARNING: Error running lua script: ...alliance\maps\scca_coop_r02\scca_coop_r02_script.lua(338): attempt to loop over field `HumanPlayers' (a nil value)
         stack traceback:
         	...alliance\maps\scca_coop_r02\scca_coop_r02_script.lua(338): in function `callbackFunction'
         	a:\github\faf\fa\lua\scenarioframework.lua(1606): in function <a:\github\faf\fa\lua\scenarioframework.lua:1600>
```

speed2CZ:
I applied hussar's fix and from a quick test it seems to be working so far. Testing cybran mission 1
I've noticed one issue. There's ALLUNITS restriction and then slowly unrestricting units. Thing is it also restricts those cybran build drones that are part of the build effect. Those should be ignored by restrictions imo.
```
WARNING: Unit.OnStopBeingBuilt() Army 1 cannot create restricted unit: <LOC ura0001_desc>Build Bot Effect
```

speed2CZ:
Then there's a problem if you capture restricted unit it just disappears

lua/game.lua
aeoncleanse: Nice.

lua/game.lua
aeoncleanse: Can't do that. Both get and set restrictions read and overwrite the restrictions table as a whole. Ignored has to stay where it is.

lua/game.lua
aeoncleanse: Thanks, missed em.

aeoncleanse:
IMO we need to limit the bp harvest from __blueprints. @HUSSAR-mtrela is there a way to do that, perhaps limiting it to things which have BuiltBy categories?

aeoncleanse:
That's interesting speed2. Is it any unit, or which have you tried it on?

speed2CZ:
Any restricted unit, I've tried it in cybran mission 1 at the begining you have an objective to capture uef T2 radar, and then in next objective I've tried to capture uef T1 mexes.

HUSSAR-mtrela:
@aeoncleanse yes it is possible to limit the bp harvest from __blueprints and that's what I'm doing in the UnitsManager (using the CategoriesSkipped defined in UnitsAnalzer). However this table is local so you will need to either make it global or add function in UnitsAnalzer for returning this table. Alternatively you could, copy this table to Game.lua and add additional categories that the restriction system should be ignoring.

HUSSAR-mtrela:
If you want I can limit the bp harvest from __blueprints tonight (in 8 hours). Btw, the "for loop" worked for you @aeoncleanse because we changed the part for adding restrictions in a map script from using multiple categories to using a table with categories. When you remove this "for loop" then we will not have to change code that adds restrictions in map scripts.

aeoncleanse:
I know. The loop was an error on my part, hence corrected. I'll take a look at limiting the bp search.

aeoncleanse:
410607930ef0b9c45a004dbde6a5ccde344cf577

PR #1209, opened 2016-04-21
Hotfix for unit restrictions

### LINKED ISSUE #1208: Restriction system does not restrict any of SACUs and some other units
The latest change (https://github.com/FAForever/fa/commit/410607930ef0b9c45a004dbde6a5ccde344cf577#diff-8f15a3675c02cf9bbf80d174fcbedb6e) in Game.lua limits restrictions to a half or all units. Using bp.Categories is not fully reliable way to find all valid/build-able units in the __blueprints table. Also, AFIK there is no set of categories that could be used to find these units unless one would go through all blueprints. Even if we find a set of categories then we would need to maintain it and there might not be a good way to enforcing this set of categories onto modded units. 

 Compare attached list of blueprints:
- [bps.now.txt](https://github.com/FAForever/fa/files/229187/bps.now.txt) 333 units
- [bps.correct.txt](https://github.com/FAForever/fa/files/229186/bps.correct.txt) 622 units

Also, the Game.GetUnitsUpgradable() does not returns only upgradable units but rather all units.

HUSSAR-mtrela:
- fixes  #1208 limitation of units in restriction system  
- fixes  #1208 function that finds upgradable units  
- corrects table functions in utils.lua

### DIFF
--- a/lua/game.lua
+++ b/lua/game.lua
@@ -211,20 +211,26 @@ local function SortUnits(bp1, bp2)
         return true
     end
 end
+-- check for valid unit blueprints (not projectiles/effects)
+local function IsValidUnit(bp, id)
+    if bp and (string.len(id) > 4) and not string.find(id, '/') then
+        return true
+    end
+    return false
+end
 -- gets blueprints that can be upgraded, e.g. MEX, Shield, Radar structures
 local function GetUnitsUpgradable()
     local units = {}  
     for id, bp in __blueprints or {} do 
         -- check for valid/upgradeable blueprints 
-        if bp and bp.General and (string.len(id) > 4) and
-         ((bp.General.UpgradesFrom and 
+        if bp and bp.General and IsValidUnit(bp, id) and 
+         ((bp.General.UpgradesFrom ~= '' and 
            bp.General.UpgradesFrom ~= 'none') or 
-          (bp.General.UpgradesTo and 
+          (bp.General.UpgradesTo ~= '' and 
            bp.General.UpgradesTo ~= 'none')) then
-
             local unit = table.deepcopy(bp)
             unit.id = id -- save id for a reference
-            table.insert(units, unit) 
+            table.insert(units, unit)  
         end
     end
     -- ensure units are sorted in increasing order of upgrades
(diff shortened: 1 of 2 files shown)


HUSSAR-mtrela:
@aeoncleanse @speed2CZ can you verify this PR ?

lua/game.lua
```diff
@@ -211,55 +211,50 @@ local function SortUnits(bp1, bp2)
...
+-- gets ids of valid units
+local function GetUnitsIds()
+    local units = {}
+    for id, bp in __blueprints or {} do
+        if IsValidUnit(bp, id) then
+            table.insert(units, id)
+        end
+    end
+    return units
+end
```
aeoncleanse: Ah yeah that's a much nicer way of doing this. Nice one :)

lua/game.lua
```diff
@@ -211,55 +211,50 @@ local function SortUnits(bp1, bp2)
         return true
     end
 end
+-- check for valid unit blueprints (not projectiles/effects)
+local function IsValidUnit(bp, id)
+    if bp and (string.len(id) > 4) and not string.find(id, '/') then
+        return true
+    end
+    return false
+end
```
aeoncleanse: Nice :)

lua/game.lua
```diff
@@ -211,55 +211,50 @@ local function SortUnits(bp1, bp2)
...
 -- gets blueprints that can be upgraded, e.g. MEX, Shield, Radar structures
 local function GetUnitsUpgradable()
     local units = {}  
     for id, bp in __blueprints or {} do 
         -- check for valid/upgradeable blueprints 
-        if bp and bp.General and (string.len(id) > 4) and
-         ((bp.General.UpgradesFrom and 
+        if bp and bp.General and IsValidUnit(bp, id) and 
+         ((bp.General.UpgradesFrom ~= '' and
```
aeoncleanse: Pretty sure this does nothing.

aeoncleanse:
4a13f5ccb3a142a2a1bc4b97b32edf2b7ddceea0

lua/game.lua
HUSSAR-mtrela: actually it filters 120 (upgradable) units out of all 622 units.

aeoncleanse: No, I mean, I'm pretty sure that there is no bp with an UpgradesFrom entry that is blank. Therefore, "if bp.General.UpgradesFrom" and "if bp.General.UpgradesFrom ~= ' ' " should be the same.

PR #1215, opened 2016-04-22
Hide start position / faction if random

Crotalus:
There are UI mods floating around which shows start position of a player by looking at the scenario files for ArmyID ->  position. Another easy way is to press F11 if you memorized the positions (Setons - 1, 2 for front .. 7, 8 air)

<Link href="http://forums.faforever.com/viewtopic.php?f=41&t=11577" target="_blank">http://forums.faforever.com/viewtopic.php?f=41&t=11577</Link>

With random / balanced spawns this should be considered an exploit. By shuffling the team positions sim side we prevent user side to pin down the start positions.

Another improvement could be a game option "Incognito" which completely hide the names of the enemies (show in replay / if players choose to reveal themselves during game). Random colors then also ofc. Currently high ranked players are ganked (and vice versa with weak players). Would be better if such strategies were a result of scouting an opponent's gameplay, not simply by looking at the commander name.

Finally, random faction players should probably be unknown faction until scouted.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -76,6 +76,7 @@ Patch 3652 (pending)
 - Allowed Aeon Aircraft Carrier to build Bombers and Gunships, same as the others.
 - Fixed restriction system on Sim side such that it cannot be compromised by UI mods 
 - Fixed restriction system that prevented removing restrictions on already restricted units in scenario scripts
+- It is no longer possible to find out a player's starting position by using the connection window (F11)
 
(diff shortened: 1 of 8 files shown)


Crotalus:
Added basic support for hiding random faction until scouted. It doesn't take alliances etc in account so any enemy scouts = all see what faction someone is. 

It is possible to track the reveal per team and sync faction manually to the affected players

PR #1225, opened 2016-04-26
Minor fix in restriction presets

### LINKED ISSUE #1224: T3 Preset Restriction does not restrict quantum gateway unit
In Units Manager

If you restrict with the T3 tech preset, you can still build a quantum gateway (with a T3 ACU) but you cannot build any SCU's from it which makes it a bit pointless.
Maybe the quantum gateway unit should be restricted also for that scenario?

HUSSAR-mtrela:
Fixed #1224 by including gateway structures to `No T3 restriction preset` and T3 factories which cannot be upgraded to T4 so they are useless when `No T3 restriction preset `is enabled


HUSSAR-mtrela:
@aeoncleanse this PR has a fix for an minor issue so can you push it to the release when you get a chance?

Issue #1245, opened 2016-05-03
Free intel on upgraded structures.


CodingSquirrel:
I'm basing this off of watching my own replay from my opponent's perspective.

After gaining initial intel on my structures my opponent was able to see every time I upgraded a structure. Not just the darkened out icon, full view as if they had scouted. They saw mexes and factories. Not only that but they were able to see the upgraded factories working, all of the animations running, the unit in production, and see the first second or so as the unit came off the pad.

I wasn't able to see much going the other way, since my opponent didn't upgrade much. Except he got a t3 mex (all t1 stuff but a t3 mex, go figure) and I was able to watch it animate and spin when I had no vision on it.

Even if it's fixed so that the animations and such don't play, I think it's worth having a community discussion whether we want this new "full intel" on upgraded structures or the old way that they went dark. I personally don't like the new way because at least the old way had the chance that the building was destroyed instead, and you couldn't tell if it was still t2 or t3 if you didn't scout again. Now you can see it go from t1 all the way to t3 without ever scouting past the first minute.

aeoncleanse:
The new intel was meant to be an instant flash if I remember, not visible because it should happen inside a single tick... @Crotalus I think this is your handiwork, care to elaborate?

JJ173:
I did some testing and it does more than just animate, apparently it gives brief intel of nearby units and structures. I can also see what's being built by upgraded factories. I used an Easy AI to test the problem. Scouted at the start of the game as you would a normal game. When the AI completed an upgrade, the intel became available, I could see units moving around the upgraded Mex's as well as being able to see what is being built by an upgraded factory. Bare in mind I only scouted the once when all of those structures were at Tech 1. 

https://i.gyazo.com/6b425858d45c8f78068ef15bfb34e26e.png

https://i.gyazo.com/8734fe8a7bbb3de40f12bef678b9b7d1.png

Crotalus:
Hmm, yeah, the problem with the "dark" intel blips is that they are completely untargetable, if the engine was clever they would reference to the upgraded building when right-clicked.

We could flash with 'Radar' vision instead, that would reset blip but make the actual uit invisible from   "vision" recon. This will result in the factory having icon but being invisible until scouted, like battleships only on sonar.

aeoncleanse:
Well, @Crotalus my problem is why is the intel blip so big? I mean, it's giving pretty serious vision for a noticeable length of time, why is it not just for one tick for a tiny, 1 radius area???

[...]

PR #1248, opened 2016-05-03
Just move away / delete old blip

Crotalus:
Alternative to flashing vision at the location of dead blips, we just move the dead blip away and blip it outside of map.

Downside with this is that "zoomed-in-vision" of upgraded building is unknown until
scouted but strategic icon still show what it is.

Issue #1245


Crotalus:
Hmm yeah, this didn't work either, unless unit was scouted during the upgrade the blip "underneath" would never get shown

PR #1253, opened 2016-05-07
Intel fix

### LINKED ISSUE #1245: Free intel on upgraded structures.

Crotalus:
Make sure vision for upgraded blips isn't permanent …

If a blip for some reason survives over 5 ticks we manually delete it. Old blips of upgraded units sometime survive for some unknown reason, maybe the blip structure are reused by the engine or similar.

Only one clear thread is spawned regardless of how many blips are cleared so should be effective.

#1245

### DIFF
--- a/lua/aibrain.lua
+++ b/lua/aibrain.lua
@@ -398,9 +398,6 @@ AIBrain = Class(moho.aibrain_methods) {
     --   val: true or false
     -- calls callback function with blip it saw.
 
-
-
-
     OnIntelChange = function(self, blip, reconType, val)
         if self.IntelTriggerList then
             for k, v in self.IntelTriggerList do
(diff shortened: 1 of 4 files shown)


Crotalus:
Ok, found the cause of permasight so this isn't needed

PR #1255, opened 2016-05-07
Delayed flash

Crotalus:
Need to keep track references of blips because the actual builder unit will be destroyed before we're refreshing...

Fixes permavision

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -126,6 +126,10 @@ StructureUnit = Class(Unit) {
     end,
 
     OnStopBeingBuilt = function(self,builder,layer)
+        if self.IsUpgrade and builder then
+            builder:RefreshIntel(true)
+        end
+
         Unit.OnStopBeingBuilt(self,builder,layer)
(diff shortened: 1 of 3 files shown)


aeoncleanse:
There are more changes needed. We want this to be activated _only_ if coming from an upgrade process. To that end, it should be called at OnStopBuild inside UpgradingState in defaultunits.lua for structureunit class. Oh, and on transfer from player to player too, of course.

aeoncleanse:
Why still call inn unit.lua instead of using states?

Crotalus:
There was some other upgrade code in Unit concerning upgrades which probably also should be in StructureUnit. Moved refresh stuff to StructureUnit but left the rest for now, could break something.

aeoncleanse:
Dude, seriously... PUT IT IN UPGRADINGSTATE... We do not need to be seeing this for ANYTHING other than an upgrading building!!!!

Crotalus:
:clap:

aeoncleanse:
Theeere we go :+1:

Crotalus:
Need another approach

Issue #1262, opened 2016-05-12
"Shield dome toggle" for mixed active/disabled shields


bartvh87:
Select multiple shield buildings (any faction / tech level) including both active and disabled shields, then use the [Shield dome toggle] button in the UI, or hotkey such as [Toggle all abilities]. All shields will be disabled.

This is undesirable because a second toggle action requires the shields to slowly recharge, while e.g. a fire base may have been suddenly exposed. If toggle actions attempted to activate every shield, this would be avoided, and disabling shields would still be practical by using toggle twice.

Cheers,
Zappa

duk3luk3:
Please check out #1267, does that work?

aeoncleanse:
ca1736e9a0b9ab86ec1826884cd703ff07fc8adb

PR #1267, opened 2016-05-13
Shield toggle always turns on mixed shields (#1262)

duk3luk3:
Enhances ScriptButtonBehavior function with a special case for mixed
shields so they are always turned on instead of off. This allows to
avoid double clicking shield button when user wants to turn on all
shields resulting in all shields turning off and then recharging.


duk3luk3:
Now tested and working!

aeoncleanse:
:+1: Tested working with bubbles and personal

aeoncleanse:
ca1736e9a0b9ab86ec1826884cd703ff07fc8adb

PR #1274, opened 2016-05-15
Change ballistic calculations

Crotalus:
Untangled code + fixing y-velocity bug mentioned here:

http://forums.faforever.com/viewtopic.php?f=2&t=12404

### DIFF
--- a/lua/sim/CalcBallisticAcceleration.lua
+++ b/lua/sim/CalcBallisticAcceleration.lua
@@ -1,8 +1,8 @@
 local XZDist = import('/lua/utilities.lua').XZDistanceTwoVectors
 
--- This table stores numbers of bombs left in a cluster bomb run
--- format : bombs_left[entityId] = <n_left>
-local bombs_left = {}
+-- This table stores last acceleration and numbers of bombs left in a cluster bomb run
+-- format : bomb_data[entityId] = {n_left=<n_left>, acc=<last_acc>}
+local bomb_data = {}
 
 CalculateBallisticAcceleration = function(weapon, projectile)
     local bp = weapon:GetBlueprint()
@@ -13,8 +13,8 @@ CalculateBallisticAcceleration = function(weapon, projectile)
     if not launcher then return acc end
     local id = launcher:GetEntityId()
 
-    if MuzzleSalvoSize > 1 and bombs_left[id] == nil then
-        bombs_left[id] = MuzzleSalvoSize
+    if MuzzleSalvoSize > 1 and bomb_data[id] == nil then
+        bomb_data[id] = {acc=4.75, n_left=MuzzleSalvoSize}
     end
 
     -- Get projectile position and velocity
@@ -22,12 +22,26 @@ CalculateBallisticAcceleration = function(weapon, projectile)
     local proj = {pos=projectile:GetPosition(), vel=VMult(Vector(launcher:GetVelocity()), 10)}
     local entity = launcher:GetTargetEntity()
     local target
+
     if entity and IsUnit(entity) then
         -- target is a entity
         target = {pos=entity:GetPosition(), vel=VMult(Vector(entity:GetVelocity()), 10)}
     else
         -- target is something else i.e. attack ground
         target = {pos=weapon:GetCurrentTargetPos(), vel=Vector(0, 0, 0)}
+    end
+
+    if not target.pos then -- target no longer alive
+        if bomb_data[id] then
+            -- use same acceleration as last bomb
+            acc = bomb_data[id].acc
+            bomb_data[id].n_left = bomb_data[id].n_left - 1
+            if bomb_data[id].n_left < 1 then
+                bomb_data[id] = nil
+            end
+        end
+
+        return acc
     end
 
     -- calculate flat(exclude y-axis) distance and velocity between projectile and target
@@ -39,15 +53,18 @@ CalculateBallisticAcceleration = function(weapon, projectile)
         dist.pos = dist.pos * math.clamp(1 - bp.DropBombShort, 0, 1)
     end
 
-    if bombs_left[id] ~= nil then -- bomber will drop several bombs
+    if bomb_data[id] ~= nil then -- bomber will drop several bombs
         -- calculate space between bombs, this is multiplied by 0.5
         -- to get the bombs overlapping a bit
         local len = MuzzleSalvoDelay * dist.vel * 0.5
-        local current_bomb = MuzzleSalvoSize - bombs_left[id]
+        local current_bomb = MuzzleSalvoSize - bomb_data[id].n_left
 
         -- calculate the position for this particular bomb
         dist.pos = dist.pos - (len * (MuzzleSalvoSize - 1)) / 2 + len * current_bomb
-        bombs_left[id] = bombs_left[id] > 1 and bombs_left[id] - 1 or nil
+        bomb_data[id].n_left = bomb_data[id].n_left - 1
+        if bomb_data[id].n_left < 1 then
+            bomb_data[id] = nil
+        end
     end
 
     -- how many seconds until the bomb hits the target in xz-space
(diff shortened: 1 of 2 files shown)


aeoncleanse:
Did you include some code for the deliberate short drop? It could well come in handy for Torp Bombers in future.

Crotalus:
Didn't find any units using that value so assumed it was some legacy code. I'll re-add it

aeoncleanse:
Thanks. Sorry, it's not legacy. More... Future-legacy? EQ might also use it, dunno.

Crotalus:
DropBombShort 0.5 will drop it 50% of distance short. Could add more flags to weapons in the future I guess so we can control bomb behavior (straight down, non-linear bomb carpets etc)

aeoncleanse:
That's excellent, exactly as intended. At some point I'm going to have to slowly walk through the unfamiliar math functions you're using, but this looks good. I've suggested Zock include it in balance mod for testing and release with that.

aeoncleanse:
Could you add a new commit, adding detailed step-by-step documentation of what this code actually does, so future modders who aren't the best at ballistic math can follow it through?

lua/sim/CalcBallisticAcceleration.lua
```diff
@@ -1,146 +1,55 @@
...
--- I personally think that 0 is better, but if you want to try, feel free to play with this...
-local alpha = 0
-
--- This table stores the acceleration previously calculated.
--- This serves for bombers carrying several bombs: acceleration is calculated
--- for first bomb, then stored in this table and reused for the next bombs
--- format : bomber_table[entityId] = {acc, remaining_bombs}
-local bomber_table = {}
+function ScaleV(v, scale)
+    return Vector(v.x * scale, v.y * scale, v.z * scale)
```
Sheeo: Lua core defines `VMult`, you sure you don't want to use this?

Sheeo:
> Code refactoring is the process of restructuring existing computer code—changing the factoring—without changing its external behavior. ([Source](https://en.wikipedia.org/wiki/Code_refactoring))

Please stop labelling pr's as "refactors" when they are not :) Other than that, looks good from video, haven't tested myself.

Crotalus:
Yup fail title as usual :) 

>If done extremely well, code refactoring may also resolve hidden, dormant, or undiscovered computer bugs or vulnerabilities in the system by simplifying the underlying logic and eliminating unnecessary levels of complexity. If done poorly it may fail the requirement that external functionality not be changed, and/or introduce new bugs.

This particular change shouldn't affect ballistic behavior except that y-velocity bug which by definition wouldn't have been catched by the imaginary unit tests in the first place so if I just untangled the code without knowing about the bug I guess it would have been ok to call it a refactor ;)

With no unit tests in current FAF lua code we probably shouldn't ever use "refactor" wording when doing non-trivial code change due to the lack of testing if it really holds up to that promise.

Nice catch with `VMult`, I'll use that instead and add some comments about the calculations at the same time.

Crotalus:
btw, the notion to manipulate gravity is kinda weird, would probably be better to give the bomb an initial velocity instead and always use the same gravity acceleration. But the original behavior worked this way so I guess it should be left alone for now.

Sheeo:
👍 Nice amendment. And yes, refactoring is rather impossible without being able to verify somewhat that the external behavior remained unchanged.

lua/sim/CalcBallisticAcceleration.lua
```diff
@@ -21,34 +17,52 @@ CalculateBallisticAcceleration = function(weapon, projectile)
...
     if bp.DropBombShort then
+        -- deliberately drop bomb short by % ratio, could be useful for torpedo bombers
         dist.pos = dist.pos * math.clamp(1 - bp.DropBombShort, 0, 1)
     end
 
-    if bombs_left[id] ~= nil then
+    if bombs_left[id] ~= nil then -- bomber will drop several bombs
+        -- calculate the total length of the carpet bombing, this is multiplied by 0.5
+        -- to get the bombs overlapping a bit
         local len = MuzzleSalvoDelay * dist.vel * 0.5
```
aeoncleanse: So this is the length of space between two bomblets, not the length of the whole spread?

lua/sim/CalcBallisticAcceleration.lua
```diff
@@ -21,34 +17,52 @@ CalculateBallisticAcceleration = function(weapon, projectile)
...
+        -- calculate the position for this particular bomb
         dist.pos = dist.pos + (len * MuzzleSalvoSize)/2 - len*bombs_left[id]
```
aeoncleanse: This line confuses me. If we are trying to move the target of the first bomblet closer in XZ space to the launcher, and position subsequent bombs at the appropriate spacing after that, surely ```(len * MuzzleSalvoSize)/2``` should be ```(len * (MuzzleSalvoSize - 1))/2``` given that the total length of a salvo's spread over, say, 5 bombs will have only 4 inter-bomb gaps?

Therefore is it not simpler to do ```target = target - (half total distance between bomb 1 and n) + (Length of space between each bomblet = bomb number starting at 0) ?

lua/sim/CalcBallisticAcceleration.lua
```diff
@@ -21,34 +17,52 @@ CalculateBallisticAcceleration = function(weapon, projectile)
...
         bombs_left[id] = bombs_left[id] > 1 and bombs_left[id] - 1 or nil
```
aeoncleanse: Is this functional shorthand for 
```if bombs > 1 then
    bombs = bombs - 1
else
    bombs = nil

**Comment by duk3luk3** [comment]
> bomb using some way to increase drag - parachute?

There are real bombs that do that so it's not unrealistic ;)

**Comment by aeoncleanse** [comment]
@Crotalus can I get some response to my questions before I merge this?

File: lua/sim/CalcBallisticAcceleration.lua (hunk 4 above)
Comment (Crotalus) [inline lua/sim/CalcBallisticAcceleration.lua]: Yup

File: lua/sim/CalcBallisticAcceleration.lua (hunk 2 above)
Comment (Crotalus) [inline lua/sim/CalcBallisticAcceleration.lua]: Yeah, updated the comment to reflect this

File: lua/sim/CalcBallisticAcceleration.lua (hunk 3 above)
Comment (Crotalus) [inline lua/sim/CalcBallisticAcceleration.lua]: Yeah, seems like the right thing to do, check out latest commit which should do this.

Only problem I can see is if the "bomb spacing" is larger than the bomb damage radius + target size for cases where total bomb count is even, that would result in the actual target could be missed between two bombs. At the moment we don't have any units with such big bomb spacing though.

Comment (Crotalus) [inline lua/sim/CalcBallisticAcceleration.lua]: Hmm, maybe should offset bombing aim with len / 2 when dropping uneven bomb count to always get a guarantee of the double overlap damage on the stuff you targeted instead of one bomb dropping straight on.

Comment (aeoncleanse) [inline lua/sim/CalcBallisticAcceleration.lua]: IMO the case where the spacing is too large such that it might miss target entirely needs to be dealt with in unit blueprint. AFAIK, if the bombs are set to form a single strip (Sensible MuzzleSalvoDelay and DamageRadius combo) then there's no situation this should be a problem.

IMO double vs single hit also falls in that category. What people have always wanted and will now expect is for the target spot to be exactly half-way down the impact zone.

**Comment by aeoncleanse** [comment]
```WARNING: Error running lua script: attempt to use as number a nil value
         stack traceback:
         	[C]: in function `VDist2'
         	d:\github\fa\lua\utilities.lua(46): in function <d:\github\fa\lua\utilities.lua:45>
         	d:\github\fa\lua\sim\calcballisticacceleration.lua(34): in function <d:\github\fa\lua\sim\calcballisticacceleration.lua:7>
         	d:\github\fa\lua\sim\defaultweapons.lua(133): in function `CheckBallisticAcceleration'
         	d:\github\fa\lua\sim\defaultweapons.lua(124): in function `CreateProjectileAtMuzzle'
         	d:\github\fa\lua\sim\defaultweapons.lua(658): in function <d:\github\fa\lua\sim\defaultweapons.lua:593>
```

This is apparently possible. Pretty sure it happened when the target died half-way through, as I was hitting a bunch of T1 with multiple Janus.

Crotalus:
Hmm yeah that probably happens often, I added commit that stores the last bombs acceleration in case the target is gone for some reason, in that case it uses the previous bomb's acceleration.

aeoncleanse:
That appears to have done it! Can you rebase onto develop ready for merging?

Crotalus:
Rebased it on latest develop

aeoncleanse:
02826fd718ef12ebcc351adb2fd421b5d416e949

PR #1275, opened 2016-05-15
Introduce new spawn options

Crotalus:
Rebased #1217 

Spawn options are now: Fixed, Random, Optimal balance and Flexible balance.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -1188,9 +1188,10 @@ function PossiblyAnnounceGameFull()
     GpgNetSend("GameFull")
 end
 
---- Assign the "random" (really autobalanced) start positions.
 local function AssignRandomStartSpots()
-    if gameInfo.GameOptions['TeamSpawn'] ~= 'random' then
+    local teamSpawn = gameInfo.GameOptions['TeamSpawn']
+
+    if teamSpawn == 'fixed' then
         return
     end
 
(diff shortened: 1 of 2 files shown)


Sheeo:
Crotalus; this is not a refactor.

Crotalus:
Flexible is the current autobalance, it has 5% slack on the quality setup, if it finds an optimal balance of 95% it will pick be allowed to pick a setup of > 90% to make it a bit more random.
(otherwise rehosting a game with exactly same players will always end up with the same team setup)

lua/ui/lobby/lobby.lua
```diff
@@ -1284,68 +1308,56 @@ local function AssignRandomStartSpots()
...
-        r = autobalance_random(ratingTable, teams)
-        q = autobalance_quality(r)
+        s = autobalance_random(ratingTable, teams)
+        q = autobalance_quality(s)
 
-        if q > 0.85 and q > best.quality*0.95 then
-            table.insert(results, {quality=q, result=r})
-
-            if(table.getsize(results) > 4) then break end
+        if q > setups[1].quality * frac then
```
Crotalus: Here is the quality check, for flexible it is 0.95 and for optimal 1

aeoncleanse:
This appears to work ingame and the code is good to my skills. @duk3luk3 can I get you to give this a once over with a more experienced eye?

aeoncleanse:
cf44b1c0ffca7d648c472628585d8878ab1126a1

Nojoke185:
when testing this had the following observation...

when spawn is set to random, optimal balance or flexible balance the new map preview switching does not function
i.e, there is no player id on the map, rating etc
clicking between players does not initiate player switching

unless this is expected behaviour?

image

aeoncleanse:
I saw it, but I think it's expected because with any of the non-fixed modes, we don't know which player should end up where yet, right? It balances teams  automatically. Or did I misunderstand this completely?

PR #1226, opened 2016-04-27
Assist t2/t3 mex with engineers to build mass storages

Crotalus:
To help building mass storages we could let an assist command on a non-capped T2/T3 mex result in a build mass storages command (mentioned here: #731)

This PR should be seen as a proof of concept, needs a bit of tuning (check for existing storages, if a unit really can build them to prevent invalid orders etc). Also needs to be an UI option so it can be turned off.

PR #1277, opened 2016-05-16
Assist t2/t3 mex with engineers -> build MS

Crotalus:
Rebase of #1226

Assist mex with engineer(s) to order them build mass storages, needs to be turned on in options.

* Checks if mex is already capped
* If mex currently is upgrading, you need to shift-right click twice on the mex to get the engineers to also build mass storages. This to prevent unwanted build order when you only want to assist upgrade
* Won't trigger on non-upgrading T1 mex

All above behavior is in UI so people can mod it to their liking.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -468,6 +468,14 @@ StructureUnit = Class(Unit) {
     end,
 
     OnKilled = function(self, instigator, type, overKillRatio)
+        local scus = EntityCategoryFilterDown(categories.SUBCOMMANDER, self:GetGuards())
+        if scus[1] then
+            for _, u in scus do
+                u:SetFocusEntity(self)
+                self.Repairers[u:GetEntityId()] = u
+            end
+        end
+
         Unit.OnKilled(self, instigator, type, overKillRatio)
     end,
 
@@ -478,7 +486,8 @@ StructureUnit = Class(Unit) {
                 self.Repairers[id] = nil
             else
                 local focus = u:GetFocusUnit()
-                if focus == self and u:IsUnitState('Repairing') and not u:GetGuardedUnit() then -- not assist
+                if focus == self and ((u:IsUnitState('Repairing') and not u:GetGuardedUnit()) or
+                                      EntityCategoryContains(categories.SUBCOMMANDER, u)) then
                     table.insert(units, u)
                 end
             end
--- a/lua/ui/game/commandmode.lua
+++ b/lua/ui/game/commandmode.lua
@@ -11,6 +11,8 @@ local Construction = import('/lua/ui/game/construction.lua')
 local UIMain = import('/lua/ui/uimain.lua')
 local Orders = import('/lua/ui/game/orders.lua')
 local commandMeshResources = import('/lua/ui/game/commandmeshes.lua').commandMeshResources
+local Prefs = import('/lua/user/prefs.lua')
+
 
 --[[
  THESE TABLES ARE NOT ACTUALLY USED IN SCRIPT. Just here for reference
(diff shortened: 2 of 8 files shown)


Nojoke185:
assisting a T3 mex in my testing didnt build any surrounding mex storage
T2 worked ok

Crotalus:
Ah, try with latest commit

Nojoke185:
that fixed it!

aeoncleanse:
:+1:

JaggedAppliance:
Scus don't cap mexes like everything else, other than that it seems to work correctly from my testing.

Nojoke185:
yes that was my observation also - see Crotalus's response as to why SCU's work as they do in the original PR 1226 linked above
would be great to get this in game soon

aeoncleanse:
@Crotalus, need you to clean up SCU behaviour before I merge this.

Crotalus:
SCUs are working now as well by removing REBUILDER category from scus and instead let them use the repair -> rebuild mechanics we introduced earlier.

As a future improvement we should make rebuild work for an area also (assist ground). This wasn't working with the current REBUILDER category either so that would be an improvement.

lua/defaultunits.lua
```diff
@@ -468,6 +468,14 @@ StructureUnit = Class(Unit) {
     end,
 
     OnKilled = function(self, instigator, type, overKillRatio)
+        local scus = EntityCategoryFilterDown(categories.SUBCOMMANDER, self:GetGuards())
```
aeoncleanse: Why do we need this for SCUs? I thought all engies could now 'repair' death building wrecks, so shouldn't this list be being populated for them somewhere too?

lua/defaultunits.lua
```diff
@@ -478,7 +486,8 @@ StructureUnit = Class(Unit) {
                 self.Repairers[id] = nil
             else
                 local focus = u:GetFocusUnit()
-                if focus == self and u:IsUnitState('Repairing') and not u:GetGuardedUnit() then -- not assist
+                if focus == self and ((u:IsUnitState('Repairing') and not u:GetGuardedUnit()) or
```
aeoncleanse: Again, why do we need an exception here checking SCUs? Do we want them to rebuild a dead building if they're using a command other than repair??

lua/defaultunits.lua
Crotalus: Engineers gets populated into self.Repairers when OnStartRepair is called. There is no similar thing for OnStartAssist

lua/defaultunits.lua
Crotalus: SCUs assisting something will automatically rebuild it if it dies, this is what REBUILDER category does, handled by the engine.

Nojoke185:
Checked the latest commit for this and all seems good.

T1 through T3 engineer will cap a T2 or T3 mex
Same for T2-T3 ACU and SCU

T1 mex (not upgrading) if assisted will not cap
Upgrading Mex is assisted before capping (2nd right click caps)

SCU rebuild mechanics seems to work ok also
Checked also that kennel, hive and shoulder drones do not build mass storage - they do not
Menu options works - when turned off capping mechanic does not work

lua/ui/game/commandmode.lua
```diff
@@ -144,6 +146,36 @@ function AddCommandFeedbackByType(pos, type)
...
+
+function AssistMex(command)
+    local units = EntityCategoryFilterDown(categories.ENGINEER, command.Units)
+    if not units[1] then return end
+    local mex = GetUnitById(command.Target.EntityId)
+    if not mex or IsDestroyed(mex) then return end
+
+    local eco = mex:GetEconData()
+    local bp = mex:GetBlueprint()
+    local is_capped = eco.massProduced == bp.Economy.ProductionPerSecondMass * 1.5
```
aeoncleanse: What about checking for > blueprint value, to allow this to re-cap partially capped mex?

Exotic-Retard: what about adjacent/missplaced/terrain things?
i guess we could include adjacency with sth like  
production*1.5-(0.125*adjacentthings) == massproduced

aeoncleanse:
e113a9bf198cceb74fde47019f2da6164c88538f

Nojoke185:
As an observation, noticed today that you can assist an underwater mex to add storage (although the storage does not actually get built)

The engineer will start to move to the mex and the planned mass storage units will be drawn
once the engineer arrives the mass storage build queue is cleared
Maybe there needs to be a check made to see if it is valid to add storage to the mex (i.e. check if the mex is on land 1st?) otherwise it is a wasted journey for the engi

PR #1278, opened 2016-05-16
Fix bug with torpedos countering tml acu

### LINKED ISSUE #1164: Torpedo killing sub nuke missile
Seems like collisions between torpedos and missiles are allowed which resulted in a sub nuke missile being destroyed 1 tick after it was lauched.

On one hand it's cool that such thing can happen but in this case it was non-obvious that it actually happened (people are used with nuke missiles disappearing mysteriously but this was beta).

Crotalus:
Rebase of #1220

Torpedos that hit something underwater at the same time as they fire a
missiles makes it disappear

Fixes #1164

### DIFF
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -158,38 +158,30 @@ Projectile = Class(moho.projectile_methods, Entity) {
     OnCollisionCheck = function(self, other)
         -- If we return false the thing hitting us has no idea that it came into contact with us.
         -- By default, anything hitting us should know about it so we return true.
-        if (EntityCategoryContains(categories.TORPEDO, self) and EntityCategoryContains(categories.TORPEDO, other)) or
-            (EntityCategoryContains(categories.TORPEDO, self) and EntityCategoryContains(categories.DIRECTFIRE, other)) or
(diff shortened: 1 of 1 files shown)


lua/sim/Projectile.lua
```diff
@@ -158,38 +158,30 @@ Projectile = Class(moho.projectile_methods, Entity) {
...
-		local bp = other:GetBlueprint()
-		if bp.DoNotCollideList then
-			for k, v in pairs(bp.DoNotCollideList) do
-				if EntityCategoryContains(ParseEntityCategory(v), self) then
-					return false
-				end
-			end
-		end
+        local dnc
+        for _, p in {{self, other}, {other, self}} do
```
aeoncleanse: This code struck me as clever... :)

aeoncleanse:
Restarted travis

aeoncleanse:
eff63596cbac1eddef92737f53b3f0208ab11b94

Exotic-Retard:
that fix is great and all but, looks like you forgot depth charges, which still do collide since they dont have the torpedo category. in fact they barely have any categories.
`        'UEF',
        'PROJECTILE',
        'ANTINAVY',`

PR #1282, opened 2016-05-16
Support for toggling reclaim labels + performance improvements

Crotalus:
* Get UI draw calls out of UserSync (AddReclaim / RemoveReclaim) to minimize sync impact
* Labels don't redraw their position when not needed (camera unchanged) or not on screen
* When over 1000 labels on map they will hide temporarily while zooming in / out to prevent major frame-drop.
* Possible to bind a key for toggling the labels (Alt+R as default)


Nojoke185:
noticed while scrolling sideways around the map with these enabled that the scrolling sometimes gets stuck
seems intermittent - works ok sometimes
tried this on Daroza's sanctuary as it has loads of reclaim

HUSSAR-mtrela:
@Crotalus I wonder if we could cluster multiple reclaim labels and just show one label when map is zoomed out and they are close to each other or if they are overlapping.

HUSSAR-mtrela:
also the clustered label could show total reclaim in a given cluster of reclaim

Crotalus:
@HUSSAR-mtrela See one of my earlier PR's for a clustering solution, it was rejected by subjective reasons though so made the reclaim labels UI-moddable instead so people can choose themselves how they look.

This code should be somewhat compatible with the label code in current release:
https://github.com/FAForever/fa/commit/3ab0a8ab3cc726e10b786b0b5199fa029b45c859

JaggedAppliance:
nojoke is right about the scrolling, it gets interrupted often while the labels are activated. You have to zoom in or out, or toggle the labels to get it to work again. Alt + R was not bound for me either. 

Performance is definitely improved.

aeoncleanse:
@Crotalus - Need to solve the Alt-R being bound, and the scroll sticking.

Crotalus:
Bound the toggle to Alt-R as default now.

Mid-mouse scrolling stuck issue was interesting, will have a look.

JaggedAppliance:
The scrolling being interrupted is not actually to do with this PR, i just tested on a normal game and activating the labels and trying the scrolling showed the same problem. (Yeah should've tried that earlier)

Crotalus:
Yeah, noticed it was a problem with the current version of the labels as well. I guess it wasn't as obvious with ctrl-shift due to not mid-click scrolling as much.

Try the latest version in this PR now, I hopefully fixed that bug. Also rebased to latest `develop`

JaggedAppliance:
Yep, it's fixed now.

Nojoke185:
checked this again with the changes and did not see any issues
performance seemed good
key binding the menu option worked

aeoncleanse:
Fabulous work everyone. I'll review the changes myself, then we can look at getting them into beta.

aeoncleanse:
4ef88b6d8b22aba295c092b4b99c63548a3affd0

PR #1283, opened 2016-05-16
Let the Ahwassa projectile burn straight through ASF without being blocked

Crotalus:
Rebase of #933 

With this change it won't be possible to block off the bomb with ASF, the projectile will fly straight through and deal some damage on the asf flying into it. The radius is small though so no death of the whole cloud.

### DIFF
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -156,10 +156,15 @@ Projectile = Class(moho.projectile_methods, Entity) {
     end,
 
     OnCollisionCheck = function(self, other)
+        local same_army = self:GetArmy() == other:GetArmy()
+
+        if not EntityCategoryContains(categories.PROJECTILE, other) then -- called from Unit:OnCollisionCheck
+            return not same_army or self:GetCollideFriendly() == true
+        end
+
         -- If we return false the thing hitting us has no idea that it came into contact with us.
         -- By default, anything hitting us should know about it so we return true.
-
-        if self:GetArmy() == other:GetArmy() then return false end
+        if same_army then return false end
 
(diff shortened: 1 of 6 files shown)


Zockyzock:
how much damage & how big is the radius?

Crotalus:
Very tniy collision radius atm, 2
https://github.com/FAForever/fa/pull/1283/commits/28e99a2e49ad1007e4854f8c08a800d4bc208334#diff-119b734e31c5dd1a26a6b7f4a9cae14aR16

Same damage as a normal bomb, 11k, i.e. kills asf which flies into it.

Both values should probably be moved out to blueprint so they are easily modified.

Zockyzock:
Yes, that would be great.

Zockyzock:
Can you move the values to the blueprint and then make a pull request to the balance branch?

aeoncleanse:
@Crotalus - We'd like this with bp values, retargeted at balance/3660 please

lua/sim/Projectile.lua
```diff
@@ -156,10 +156,15 @@ Projectile = Class(moho.projectile_methods, Entity) {
     end,
 
     OnCollisionCheck = function(self, other)
+        local same_army = self:GetArmy() == other:GetArmy()
```
aeoncleanse: Errrm... Why are we doing things in Projectile.lua? Or Unit.lua for that matter? Everything we need to do this change can be done in the Ahwassa projectile...

aeoncleanse: Got it. This is universal collision logic to allow projectiles to properly check collisions for friendly colliding allowed.

aeoncleanse:
Recreated as https://github.com/FAForever/fa/pull/1545 with some fixes.

PR #1285, opened 2016-05-17
Kick observers before checking connection issues

Crotalus:
Otherwise the host need to manually kick observers with connection issues before able to launch the game.


loc/US/strings_db.lua
```diff
@@ -4066,6 +4066,7 @@ lob_0004="Observers not allowed."
 lob_0005="Kicked by host."
 lob_0006="No clients allowed in limbo at launch"
 lob_0007="Host abandoned lobby"
+lob_0008="Game was launched."
```
aeoncleanse: I'd like this added to the other LOCs please, and translated to as many as you can. Exotic_Retard can do Russian.

lua/ui/lobby/lobby.lua
```diff
@@ -5417,6 +5409,13 @@ function InitHostUtils()
             end
 
             return -1
+        end,
+
+        KickObservers = function(reason)
+            for k,observer in gameInfo.Observers:pairs() do
+                lobbyComm:EjectPeer(observer.OwnerID, reason or "KickedByHost")
+            end
+            gameInfo.Observers = WatchedValueArray(LobbyComm.maxPlayerSlots)
```
ChrisKitching: Care should perhaps be taken about emptying a `WatchedValueArray` by reinitialising it. This will discard listeners, which may in general produce undesired behaviour. We probably want a clear() function on that class that empties it without removing listeners.

lua/ui/lobby/lobby.lua
```diff
@@ -1580,30 +1576,22 @@ local function TryLaunch(skipNoObserversCheck)
             return
         end
 
-        if anyOtherObservers then
-            if skipNoObserversCheck then
-                -- we send the observer list before kicking the players, in case they are not registered as observer
-                -- and won't disconnect correctly before the game launch.
```
ChrisKitching: Either this comment is wrong, or you've introduced a subtle bug by no longer sending the observer list when kicking the observers.

Crotalus: `sendObserversList` was removed with e3a5cd7c53d498b4db2a748b7f20b360544f6788 so current version probably pukes around here.

aeoncleanse:
@Crotalus has there been any progress on this since the feedback in June?

Crotalus:
Added translation of the message now.

The other point about WatchedValueArray is part of the earlier code (only moved it around) so should be working. 

Haven't tested the actual function though in lobby so needs to go through beta-phase.

loc/US/strings_db.lua
Exotic-Retard: lob_0008="Игра была запущена."
enjoy.

aeoncleanse:
New PR #1809 rebased

PR #1287, opened 2016-05-18
Intel refresh act 3 - spawn separate vision entity

Crotalus:
Spawn a vision entity which is destroyed when the new unit is actually
seen by someone

Tested this with 4v4 AI and didn't manage to get permavision or non-refreshed blips.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -360,6 +360,7 @@ StructureUnit = Class(Unit) {
                 NotifyUpgrade(self, unitBuilding)
                 self:StopUpgradeEffects(unitBuilding)
                 self:PlayUnitSound('UpgradeEnd')
+                self:RefreshIntel(unitBuilding)
                 self:Destroy()
             end
         end,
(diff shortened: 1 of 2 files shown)


aeoncleanse:
This won't be going into release/3654. I've assigned it to 3656

aeoncleanse:
The bug this fixes is that a building which has been scouted and is under radar cover which then upgrades cannot be targeted directly. To test, run that situation. Ensure the upgraded building can be targeted. Ensure that there's no permavision revealing units being built in the enemy factory. Ensure that the blip has refreshed properly to the new upgraded building's.

Nojoke185:
Scouted enemy base - included factory/mexes
Radar cover created over enemy base (no further vision scouting performed)

MEX:
When one mex upgraded, was able to see it fade out (unit model was see through somewhat)
Did not observe this for any other T2 mex upgrades
After upgrade the mex icon went to T2 icon and model also updated - no vision scouting was performed to see this change

FACTORY:
Same for Factory - after upgrade the icon went to T2 icon and model updated - no scouting was performed to see this

ATTACK ORDER:
Both before and after I was able to target the buildings

Allowed mex and factory to go to T3. Worked same for those updates also. Also able to order an attack

No permavision seen for units built in enemy factory
You can see the radar blips as they roll off but no vision of them

Expected behaviour seemed to work as described

aeoncleanse:
Nice one @Nojoke185 . There's this new "Review" feature on here, could you just hit "Review Changes" and approve? No need to write anything further.

Nojoke185:
sorry the last comment I had was - it is expected behaviour that the upgraded icon and model will display to the user without further scouting (vision)?
or is this upgrade change picked up and provided by the radar?

I should try the same test without radar cover I guess!...

aeoncleanse:
It is... desired... It's the lesser of three evils if you will :)

aeoncleanse:
f7ac782a6825d53ebac51494a946e42ac1a65ef4

PR #1293, opened 2016-05-19
Report victim / instigator id with teamkill

### LINKED ISSUE #1291: GPGNet TeamkillReport: include player's ids
After a quick talk with sheeo we would like to have the GPGNetSend for Teamkill enriched with the teamkiller's and the victim's player id (id from the server).
Nicknames shall be kept for debug reasons when looking into logs.

`GpgNetSend('TeamkillReport', killTime, victimId, victimName, killerId, killerName)`
[original call here (line 59)](https://github.com/FAForever/fa/blob/develop/lua/ui/dialogs/teamkill.lua)

This makes server handling far easier.

Unfortunately I have no idea how to catch the player-ids.

Crotalus:
Fixes #1291

### DIFF
--- a/lua/UserSync.lua
+++ b/lua/UserSync.lua
@@ -66,14 +66,22 @@ function OnSync()
             end
         end
     end
-	
-    if Sync.Teamkill then
-        local armiesInfo = GetArmiesTable()
-        local victimName = armiesInfo.armiesTable[Sync.Teamkill.victim].nickname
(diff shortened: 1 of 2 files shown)


Sheeo:
👍 looks good.

aeoncleanse:
28edc09a9f47921503ca965de98a34e45235f9aa

PR #1295, opened 2016-05-21
Add EnableDiskWatch-support for UI so you can hot reload lua files

Crotalus:
Have a "global" reference table with the UI controls so they can be found if an UI lua file is reloaded.

### DIFF
--- a/lua/ui/controls.lua
+++ b/lua/ui/controls.lua
@@ -1,0 +1,10 @@
+local controls = {}
+
+function Get(key)
+    if not key then
+        key = debug.getinfo(2, "S").source
+    end
+
+    controls[key] = controls[key] or {}
+    return controls[key]
+end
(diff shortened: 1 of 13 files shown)


Sheeo:
Does this actually do the trick and allow you to hot reload all UI code?

Crotalus:
`touch lua/ui/game/*.lua` works without getting spam in the log about destroyed elements.

It doesn't force a redraw though so already instanced UI elements can have function pointers to "old version" of code requiring reselect / redraw to actually see the change.

If we restructure the UI code some more we probably could get that to be done automatically as well.

Exotic-Retard:
wait; if you can mess with enablediskwatch, how far can you mess with it?

currently it does not work on hooked files properly. when you reload a hooked file for something like a unit, it tries to reload all the textures *from that directory* which are usually not there.

making it use a fallback directory (base game dir) or something would be the best thing ever and essentially fix it for hooked mods.

also: currently you can reload dds textures from ui mods with it, and it works. dunno if that is useful info though

aeoncleanse:
c059c8d3198411628fa1752b64c96a482e835f8e

PR #1299, opened 2016-05-23
Added player rating label on top map preview

Justify87:
**LblPlayerRatingMapview()** Checks if scenario is selected, checks if there are any exisiting labels and deletes them, then it creates a rating label on top of map preview for every player currently in the lobby (including AI)

**DestroyLblPlayerRatingMapPreview()** In case we need to destroy the labels without executing the whole function

**LblPlayerRatingMapview()** gets called in **SetSlotInfo()**, **ClearSlotInfo()** and **DoSlotBehaviour()** since these functions handel the events we need to look at if we want to add or remove the labels. They also already check if random spawn is selected in the options.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -1006,6 +1006,7 @@ local function GetRandomFactionIndex()
     return randomfaction
 end
 
+
 local function AssignRandomFactions()
     local randomFactionID = table.getn(FactionData.Factions) + 1
     for index, player in gameInfo.PlayerOptions do
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -360,8 +360,7 @@ local function DoSlotBehavior(slot, key, name)
     elseif key == 'occupy' then
         if IsPlayer(localPlayerID) then
             if lobbyComm:IsHost() then
-                local currentSlot = FindSlotForID(localPlayerID)
-                HostUtils.MovePlayerToEmptySlot(currentSlot, slot)
+                HostUtils.MovePlayerToEmptySlot(FindSlotForID(localPlayerID), slot)
             else
                 lobbyComm:SendData(hostID, {Type = 'MovePlayer', CurrentSlot = FindSlotForID(localPlayerID),
                                    RequestedSlot = slot})
@@ -415,7 +414,7 @@ local function DoSlotBehavior(slot, key, name)
             HostUtils.AddAI(name, key, slot)
         end
     end
-
+    LblPlayerRatingMapview()
 end --\\ End DoSlotBehavior()
 
 local function IsModAvailable(modId)
@@ -723,7 +722,6 @@ function SetSlotInfo(slotNum, playerInfo)
     local slot = GUI.slots[slotNum]
     local isHost = lobbyComm:IsHost()
     local isLocallyOwned = IsLocallyOwned(slotNum)
-    local ratingLabel = GUI.mapView.ratingLabel[slotNum]
 
     -- Set enabledness of controls according to host privelage etc.
     -- Yeah, we set it twice. No, it's not brilliant. Blurgh.
@@ -817,15 +815,6 @@ function SetSlotInfo(slotNum, playerInfo)
     slot.ratingText:SetText(playerInfo.PL)
     slot.ratingText:SetColor(playerInfo.RC)
 
-    -- Showing the rating labels on the map preview
-    ratingLabel:Show()
-    if slotState == 'ai' then
-        ratingLabel:SetText(playerInfo.PlayerName)
-    else
-        ratingLabel:SetText(playerInfo.PL)
-    end
-
-
     slot.numGamesText:Show()
     slot.numGamesText:SetText(playerInfo.NG)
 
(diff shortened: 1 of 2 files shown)


lua/ui/lobby/lobby.lua
```diff
@@ -5432,3 +5435,132 @@ function InitHostUtils()
...
+
+        -- we need to do this because of windows resizing
+        marker_left     = GUI.mapView.startPositions[slots].Left()-((GUI.Width()-GUI.panel.Width())/2) -9
+        marker_top      = GUI.mapView.startPositions[slots].Top()-((GUI.Height()-GUI.panel.Height())/2) -11
+
+        local lblDepth  = 4
+        local ratingLabelTextSize = 10
+
+        ---Create labels, it's dumb, but it works
+        ---lets hope there are never more than 12 players on a map
```
Sheeo: We've been working to support arbitrary numbers of players in other areas of the codebase. Please do use a loop structure instead so we don't have yet another place to rework :)

lua/ui/lobby/lobby.lua
```diff
@@ -5432,3 +5435,132 @@ function InitHostUtils()
...
+        if slots == 1 then
```
Sheeo: You shouldn't be assigning these to the global `GUI` table, there should be a better place to store your references so they don't clutter a global.

lua/ui/lobby/lobby.lua
```diff
@@ -5432,3 +5435,132 @@ function InitHostUtils()
...
+        GUI.mapViewPlayerRating_1 = UIUtil.CreateText(GUI.panel, rating, ratingLabelTextSize, 'Arial Gras', true)
+        LayoutHelpers.AtLeftTopIn(GUI.mapViewPlayerRating_1, GUI.panel, marker_left, marker_top)
+        LayoutHelpers.DepthOverParent(GUI.mapViewPlayerRating_1, GUI.mapPanel, lblDepth)
+        end
+        if slots == 2 then
```
Sheeo: Looks like you want something like:

    ratingLabels[slot] = UIUtil.CreateText(...)
    LayoutHelpers.AtLeftTopIn(ratingLabels[slot], ...)
    ..depthOverParent(ratingLabels[slot])

Instead of this monstrosity that only works for 12 players anyway ;)

lua/ui/lobby/lobby.lua
aeoncleanse: What I think sheeo means is that we can use string concat here to cut down on the amount of code. Because this is already looping over the slots series, you can just do:

```GUI.mapViewPlayerRating_ .. slots = UIUtil.CreateText(GUI.panel, rating, ratingLabelTextSize, 'Arial Gras', true)```

Or something similar, and each time the code loops, it will use the new slots number to create that text.

File: lua/ui/lobby/lobby.lua [hunk 4]
```diff
@@ -5432,3 +5435,132 @@ function InitHostUtils()
         end
     }
 end
+
+
+function DestroyLblPlayerRatingMapPreview()
+    if GUI.mapViewPlayerRating_1 then GUI.mapViewPlayerRating_1:Destroy() end
```
aeoncleanse: Likewise here as well. If we put the mapViewPlayerRating stuff inside a table of their own (So you'd have GUI.mapRatings.mapViewPlayerRating_1, you can loop over mapRatings to kill them all.

lua/ui/lobby/lobby.lua
```diff
@@ -5432,3 +5435,132 @@ function InitHostUtils()
...
+            local PeerRating = playerInfo.PL
+            rating = PeerRating
+        elseif player.Human and (player.OwnerID == localPlayerID) then --check if local player
+            local myPlayerData = GetLocalPlayerData()
+            rating = myPlayerData.PL
+        else --what is left should be AI, so just insert name instead of rank
+            rating = player.PlayerName
+        end
+
+        if not slots then
```
aeoncleanse: This can be run above peer checking to avoid unnecessary code execution I think

Sheeo:
Hey @Justify87 thanks for your first pull request! :)

I left some specific comments inline in the diffs. Generally though, we've been trying to clean up `lobby.lua` so it doesn't repeat itself as much and doesn't rely so much on manual UI placement.

The `mapView` object that you are adding labels to is actually represented by a class; see: https://github.com/FAForever/fa/blob/develop/lua/ui/controls/resmappreview.lua#L14.

So the ideal solution here is to extend this class with optional support for rendering player ratings on top.

aeoncleanse:
Indeed, welcome. I agree with all sheeo's put, and will be on hand if you get confused about what needs to be done. Just drop me a PM on the forums or something :)

Justify87:
Thanks for your suggestions, @Sheeo and @aeoncleanse. I actually never worked with classes before. I tried to loop through the labels, but could not find a way to make it work, so I did it without a loop.

I took a look at that class mentioned by Sheeo and added the labels to it pretty easily, since the class allowed me to loop through an array. I then rewrote my initial function in the lobby.lua and tried to set the text of the labels according to the rating of the players. Unfortuantly it did not work. I will post the details on my forum thread.

Justify87:
Sorry for all the commits, but they did not appear in my branch or in my pull request even 45min after I pushed them, so I got confused a reverted and pushed again a few times.

Sheeo:
Github has been having some issues today, this would be why. They're still [having issues](https://status.github.com/), so you may still experience some issues. I'll wait with further review until it's back to normal -- in the meantime, please do rebase away your 'revert' commits.

Justify87:
I think I failed horribly at rebasing.

Justify87:
Should I just close it? The checks aren't finishing.

PR #1302, opened 2016-05-24
Fix Hydrocarbon Power Plant's icon background

Uveso:
The icons from all 4 Hydrocarbon Power Plant's are showing 'land' as iconbackground, but they should show 'amph' (half Water/Land). Like the Mass Extractor the Hydrocarbon can be build on layer Land and Seabed!

Affected Units: xsb1102 urb1102 ueb1102 uab1102
Changed General.Icon = 'land' to General.Icon = 'amph'


aeoncleanse:
:+1:

PR #1308, opened 2016-05-24
Justify87's map preview slot switch

aeoncleanse:
He closed his PR #1299 , but I have this copy with a bit of cleanup done.

PR #1309, opened 2016-05-24
Feature/player swap map preview

Justify87:
Click on a not-empty slot in the map preview and click on a another not
empty slot to swap those players.
lobbymagic

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -5197,8 +5197,7 @@ function InitHostUtils()
 
             -- Move the observer into the slot the first player came from.
             HostUtils.ConvertObserverToPlayer(FindObserverSlotForID(toOpts.OwnerID), moveFrom, true)
-            WARN(fromOpts.PlayerName)
-            WARN(toOpts.PlayerName)
+
             -- %s has switched with %s
             SendSystemMessage("lobui_0417", fromOpts.PlayerName, toOpts.PlayerName)
         end,
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -65,6 +65,8 @@ local Warning_MAP = false
 local LrgMap = false
 
 local HostUtils
+local mapPreviewSlotSwapFrom = 0
+local mapPreviewSlotSwap = false
 
 local teamIcons = {
     '/lobby/team_icons/team_no_icon.dds',
@@ -3357,6 +3359,11 @@ function ConfigureMapListeners(mapCtrl, scenario)
                         else
                             lobbyComm:SendData(hostID, {Type = 'MovePlayer', CurrentSlot = FindSlotForID(localPlayerID), RequestedSlot = slot})
                         end
+                        -- if first click is a not empty slot and second click is a empty slot: reset vars
+                        if mapPreviewSlotSwap == true then
+                            mapPreviewSlotSwap = false
+                            mapPreviewSlotSwapFrom = 0
+                        end
                     elseif IsObserver(localPlayerID) then
                         if lobbyComm:IsHost() then
                             local requestedFaction = GetSanitisedLastFaction()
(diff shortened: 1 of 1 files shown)


aeoncleanse:
ce3282eda088c8fa1f3517cb003c179aaa0dcc71

ChrisKitching:
@aeoncleanse You should not be performing nontrivial code review yourself. This patch needs more work:
https://www.dropbox.com/s/mm8x7jb3ha0li4g/Screenshot%202016-05-31%2019.15.40.png?dl=0

This will be reverted, let's go round again when we have tested and reviewed this properly.

aeoncleanse:
I considered it trivial because it's a simple change and one I have tested on four maps and seen fully functioning. Indeed, I cannot reproduce your screenshot.

Nojoke185:
tried this out and found the following:

Load multi player lobby
load any 4 player map (map/no of slots are probably irrelevant)
Occupy one slot with your army (by default this is already selected on entry to the lobby)
fill the other 3 slots with different AI types (easy, rush, adaptive) using the drop down list boxes on main menu part
on small map preview (top right)...
click on your player
next click on any opponent AI army slot
you are moved but the AI army slot (your previous slot) is now empty rather than swapped
Not sure if this is intentional behavoir or not

repeat with other 2 AI and you will eventually clear them all out

aeoncleanse:
I believed it was. This works by hijacking the normal slot logic, you know, when you use the drop-downs to switch players around as host? When you do this with an AI, the 'Occupy' action is unavailable, only able to change or remove the AI.

Now you mention it though, perhaps 'Occupy' should work in both cases?

aeoncleanse:
And then, having said that, the map switch begins working with AI perfectly without even rebooting the game. WTF is going on here?

Justify87:
It is either a Show()/Hide() issue or the slot numbers in the newly added loop to refresh the large map preview doesnt deal with the real slot numbers.

Justify87:
An AI vanishes when being swaped with a player, this is the default behaviour.

Justify87:
RefreshLargeMap() has to be called in ConfigureMapListeners() mouse handler marker.OnClick in order to refresh the map when a marker is clicked. Since refreshing the map is bound to the button showing the large map preview only.

This is not part of the player swap functionality of this pull request tho. It belongs to #1308

aeoncleanse:
I found a better way, just finalising it now.

Justify87:
@aeoncleanse , @Nojoke185 

The problem with swapping two AI or a human and an AI is, that the AI gets the HostID as its OwnerID. The HostID in this case is equal to the LocalPlayerID. So an AI has the ID of the local player (which is always the host, afaik). And this ID is then used to determine which observer in the observer list should be moved into a game slot. And all of this has to do with player swapping because player swapping is not a direct/clean swap, the observer list is used as a temporary "bucket" for that:

Player 1 becomes Obs
Player 2 is moved to Player 1 slot
Obs is moved to Player 2 slot

Now if there is the following scenario:
Host is observer (1). All game slots are occupied by AI or Players.
Host now wants to swap two AI. The execution looks like this:

Ai 1 becomes Obs 2
Ai 2 is moved to Ai 1 slot
Obs 1 is moved to Ai 2 slot

So Obs 2 (which is Ai 1) remains in the observer list. Until you click the observer button, the AI is then moved to a player slot. Because based on the OwnerID, the system thinks it moves me (the host) from observer to players (the top most observer with the right ID in the list is moved first).
Instead of setting the OwnerID of the AI to the HostID, I have set it to 999999 since -1 is already in use, I think. This solved the problem. I'm just testing it a bit more and will clean up the code. After that I will create a new pull request.

aeoncleanse:
Undergoing active testing in branch: ice

aeoncleanse:
cb886ca9ec81ef622b522897cc3ca6a306f39ced

Issue #1237, opened 2016-05-02
Slots closed in lobby not visible as such to newly joining players


aeoncleanse:
http://forums.faforever.com/viewtopic.php?f=3&t=12353

Not a 3652 regression.

Crotalus:
For reference, this line is the culprit resetting ClosedSlots for someone just joining:

https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/lobby.lua#L1491

When someone joins their game state is reset with UpdateAvailableSlots() and ClosedSlots is reset before it even can be rendered correctly in UI.

aeoncleanse:
Moving this back into the limelight. @Justify87  - What happened with this? Did your fix work?

Uveso:
I think we have a fix for it now: #1567

aeoncleanse:
1d98cf18ef15be8ee3c22d5fa16eb573704e11cd

PR #1311, opened 2016-05-24
Fixed closed slot not showing for new player

Justify87:


aeoncleanse:
@Justify87 can you explain the issue this solves a bit more?

Justify87:
@aeoncleanse, #1237

PR #1316, opened 2016-05-26
Enhancement: Enable Tooltip on queuestack, unitstack and attachedunit

Uveso:
This will enable Tooltips to UnitIcons at:
- build cue (ACU, SACU, Engineer, Factories, mobile factories)
- selection panel (singel Unit, multiple units, UnitUpgrades)
- attached units on transporter

If a Unit has a construction panel, click  on "Selection and Storage" Tab to see the Unit Icon and Tooltip. (ACU, SACU etc.)

### DIFF
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -1051,9 +1051,8 @@ function OnRolloverHandler(button, state)
                 button.dragMarker:Destroy()
                 button.dragMarker = false
             end
-            UnitViewDetail.Hide()
         end
-    else
+    end
         if state == 'enter' then
             if item.type == 'item' or item.type == 'queuestack' or item.type == 'unitstack' or item.type == 'attachedunit' then
                 UnitViewDetail.Show(__blueprints[item.id], sortedOptions.selection[1], item.id)
@@ -1063,7 +1062,7 @@ function OnRolloverHandler(button, state)
         else
             UnitViewDetail.Hide()
         end
-    end
+    
 end
 
 function OnClickHandler(button, modifiers)
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -1055,7 +1055,7 @@ function OnRolloverHandler(button, state)
         end
     else
         if state == 'enter' then
-            if item.type == 'item' then
+            if item.type == 'item' or item.type == 'queuestack' or item.type == 'unitstack' or item.type == 'attachedunit' then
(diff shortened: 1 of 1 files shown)


aeoncleanse:
Works in Factory construction panel, but not in unit queue? Same with QGate. Can it be done? Everything else good.

Uveso:
This should work at least in Qgate. Don't know, what exactly Unit Queue is, but should work there also :)

If you can see the tooltip on a landfactory, you should see it on the Qgate.
"item.type=item" is for the upper icon line inside the construction menu.
The 2nd line is "=queuestack".
On both, factory and Qgate.

If you select units on the field, they will be shown as "item.type=unitstack". There is now also a tooltip.
And "=attachedunit" enables the tooltip on Transporters (Attached units)

So the only button with no tooltip should be the icon for build-templates. 

Tested it right now, while writing this. FAF-Develop, no mods.

aeoncleanse:
This doesn't work with the build queue on a QGate or Facs...

aeoncleanse:
image

Uveso:
Well, if you mean with "queue" the 2nd line (queuestack), then go to options and disable "Draggable Build Queue".

But one thing on your screenshoot is strange.
The Unit Information window "QGW R-32 Tech 3 Quantum gate" should be hidden , if you don't hover the mouse over the Quantumgate !?!

If i move the Mouse over the quantumgate (does not matter if its selected)  i see the same InfoWindow, but if i move the mouse away from the gate, i dont see it !?!

If i select the Gate and hover the mouse over any Subcommander (1st and 2nd line),
i see as InfoWindow the subcommder, with Info about the mass/energy cost, and i see the Unit/Weapon Tooltip:

sacutooltip

aeoncleanse:
It was indeed the draggable build queue option. Is there a way you can make this work even with that on, or are the two fundamentally incompatible?

Uveso:
There is a way :)

aeoncleanse:
8430f873af48227d7e3f2c846161468b1d0f1180

simonrad:
Thanks so much for fixing this!!  Any rough idea when it will be available to play?

Uveso:
Here you can see the FAF DevBlog:
http://faforeverdevblog.tumblr.com/

For update plans look at:
2016-07-11 Developer Conference

This does not tell you when this update will be released, but it gives you a rough idea how udates are planed in general.
(Game patches will be released, every two months)

PR #1297, opened 2016-05-22
Bugfix for Template Icons and names

Uveso:
Fixed a crash, when a icon is aleady saved as "false" inside build_templates
Added search for modded unit icons while saving build_templates to Game.prefs
Removed prefix in template names.
Changed Warning "Warning: Unable to find file..dds" to Debug text.

For detailed description see issue #1228

Issue #1228, opened 2016-04-28
Saving templates dont save the correct IconName with modded units.


Uveso:
This Error only appears with modded units.

Saving build-templates don't use a correct uniticon, if the first unit inside the templatearray is not an original (unmodded) unit.

Example:
First build a modded unit like 'brnt1hpd' (total mayhem UEF T1 Powerdefence)
Second build a powerenerator like 'ueb1101' (UEF T1 Pgen)
Now select both buildings and klick "save template".

Inside game.prefs, this will be saved:
```
                    templateData = {
                        4,
                        2,
                        {
                            'brnt1hpd',
                            84,
                            0,
                            0
                        },
                        {
                            'ueb1101',
                            447,
                            2,
                            0
                        }
                    },
                    name = '<LOC brnt1hpd_desc>brnt1hpd',
                    icon = false
```
As u can see, it is saving "icon=false". This causes some problems inside construction.lua

Some lines from construction lua starting at line 786: (last beta build)
```LUA
        elseif type == 'templates' then
            control:DisableOverride()
            SetIconTextures(control, control.Data.template.icon)
            control:SetNewTextures(GetBackgroundTextures(control.Data.template.icon))
            control.Height:Set(48)
            control.Width:Set(48)
            if control.Data.template.icon then
                control.Icon:SetTexture(UIUtil.UIFile('/icons/units/' .. control.Data.template.icon .. '_icon.dds', true))
            else
                control.Icon:SetTexture('/textures/ui/common/icons/units/default_icon.dds')
            end
            control.Icon.Height:Set(48)
            control.Icon.Width:Set(48)
```
SetIconTextures(control, control.Data.template.icon) will fail, because "control.Data.template.icon" is "false" and not a string/UnitID/IconName

To solve the problem we need to change the function GetInitialIcon(template) inside build_templates.lua.
This is the original function:
```LUA
function GetInitialIcon(template)
    for _, entry in template do
        if type(entry) != 'table' then continue end
        if DiskGetFileInfo('/textures/ui/common/icons/units/'..entry[1]..'_icon.dds') then
            return entry[1]
        else
            return false
        end
    end
end
```
First of all the function is only looping over the first element of the template array. (after the first table is found)
Then, if the icon from the first unit inside the array is not found, it is returning "false"
Even if there are more units inside the Template, they will not checked for valid names.

This is a corrected version of GetInitialIcon:
(need to import a function from uiutil.lua, if the first Unit is not an original unit.)
```LUA
[...]

PR #1321, opened 2016-05-27
Bugfix for Template Icons and Names II

Uveso:
(Reworked PR from #1297)

Fixed a crash, when a icon is aleady saved as "false" inside build_templates
Added search for modded unit icons while saving build_templates to Game.prefs
Removed prefix in template names.
Changed Warning "Warning: Unable to find file..dds" to Debug text.

For detailed description see issue #1228

### DIFF
--- a/lua/ui/game/build_templates.lua
+++ b/lua/ui/game/build_templates.lua
@@ -36,7 +36,7 @@ end
 function GetInitialName(template)
     for _, entry in template do
         if type(entry) != 'table' then continue end
-        return __blueprints[entry[1]].Description
+        return (string.gsub(__blueprints[entry[1]].Description, '^<[^>]*>', '')) -- removes <LOC xyz_desc> from name
     end
 end
 
@@ -44,11 +44,15 @@ function GetInitialIcon(template)
     for _, entry in template do
         if type(entry) != 'table' then continue end
         if DiskGetFileInfo('/textures/ui/common/icons/units/'..entry[1]..'_icon.dds') then
-            return entry[1]
+            return entry[1] -- Original unit found
         else
-            return false
+            local UIUtil = import('/lua/ui/uiutil.lua')
+            if UIUtil.UIFile('/icons/units/' .. entry[1] .. '_icon.dds', true) then
+                return entry[1] -- Modded unit found.
+            end
         end
     end
+    return 'default' -- If we don't find a valid IconName; return string 'default'
 end
 
 function AddTemplate(newTemplate)
--- a/lua/ui/game/build_templates.lua
+++ b/lua/ui/game/build_templates.lua
@@ -1,0 +1,114 @@
+#*****************************************************************************
+#* File: lua/modules/ui/game/build_templates.lua
+#* Author: Ted Snook
+#* Summary: Build Templates UI
+#*
+#* Copyright © 2007 Gas Powered Games, Inc.  All rights reserved.
+#*****************************************************************************
+
+local Prefs = import('/lua/user/prefs.lua')
+local templates = Prefs.GetFromCurrentProfile('build_templates') or {}
+local UIUtil = import('/lua/ui/uiutil.lua')
+
+function CreateBuildTemplate()
+    GenerateBuildTemplateFromSelection()
+    local template = GetActiveBuildTemplate()
+    ClearBuildTemplates()
+    if table.getsize(template) > 0 then
+        AddTemplate(template)
+    end
+end
+
+function Init()
+    import('/lua/ui/game/gamemain.lua').RegisterChatFunc(ReceiveTemplate, 'Template')
+end
+
+function ReceiveTemplate(sender, msg)
+    if Prefs.GetOption('accept_build_templates') == 'yes' then
+        local tab = import('/lua/ui/game/construction.lua').GetTabByID('templates')
+        if tab then
+            import('/lua/ui/game/announcement.lua').CreateAnnouncement(LOC('<LOC template_0000>Build Template Received'), tab, LOCF('<LOC template_0001>From %s', sender))
+        end
+        AddTemplate(msg.data)
+    end
+end
+
+function GetInitialName(template)
+    for _, entry in template do
+        if type(entry) != 'table' then continue end
+        return __blueprints[entry[1]].Description
+    end
+end
+
+function GetInitialIcon(template)
+    for _, entry in template do
+        if type(entry) != 'table' then continue end
+        if DiskGetFileInfo('/textures/ui/common/icons/units/'..entry[1]..'_icon.dds') then
+            return entry[1]
+        else
+            return false
+        end
+    end
+end
+
+function AddTemplate(newTemplate)
+    table.insert(templates, {templateData = newTemplate, name = GetInitialName(newTemplate), icon = GetInitialIcon(newTemplate)})
+    Prefs.SetToCurrentProfile('build_templates', templates)
+end
+
+function GetTemplates()
+    return Prefs.GetFromCurrentProfile('build_templates')
+end
+
+function ResetTemplates()
+    Prefs.SetToCurrentProfile('build_templates', false)
+end
+
+function RemoveTemplate(templateID)
+    table.remove(templates, templateID)
+    Prefs.SetToCurrentProfile('build_templates', templates)
+end
+
+function RenameTemplate(templateID, name)
+    templates[templateID].name = name
+    Prefs.SetToCurrentProfile('build_templates', templates)
+end
+
+function SetTemplateIcon(templateID, iconPath)
+    templates[templateID].icon = iconPath
+    Prefs.SetToCurrentProfile('build_templates', templates)
+end
+
+function SendTemplate(templateID, armyIndex)
+    armyIndex = armyIndex
+    if table.getn(templates[templateID].templateData) > 22 then
+        UIUtil.QuickDialog(GetFrame(0), "<LOC build_templates_0000>You may only send build templates with 20 or less buildings.", 
+            "<LOC _Ok>", nil, nil, nil, nil, nil,
+            true,  {worldCover = true, enterButton = 1, escapeButton = 1})
+    else
+        SessionSendChatMessage(armyIndex, {Template = true, data = templates[templateID].templateData})
+    end
+end
+
+function SetTemplateKey(templateID, key)
+    local used = false
+    for i, template in templates do
+        if i == templateID then continue end
+        if template.key and template.key == key then
+            used = true
+            break
+        end
+    end
+    if used then
+        return false
+    else
+        templates[templateID].key = key
+        Prefs.SetToCurrentProfile('build_templates', templates)
+        return true
+    end
+end
+
+function ClearTemplateKey(templateID)
+    templates[templateID].key = nil
+    Prefs.SetToCurrentProfile('build_templates', templates)
+end
--- a/lua/ui/uiutil.lua
+++ b/lua/ui/uiutil.lua
@@ -422,7 +422,7 @@ function UIFile(filespec, checkMods)
         end
         
         if not found then
-            WARN(debug.traceback(nil, "Warning: Unable to find file: " .. origPath .. filespec))
+            SPEW('[uiutil.lua, function UIFile()] - Unable to find file:'.. origPath .. filespec)
             found = filespec
         end
 


lua/ui/game/build_templates.lua
```diff
@@ -0,0 +1,118 @@
...
+        AddTemplate(template)
+    end
+end
+
+function Init()
+    import('/lua/ui/game/gamemain.lua').RegisterChatFunc(ReceiveTemplate, 'Template')
+end
+
+function ReceiveTemplate(sender, msg)
+    if Prefs.GetOption('accept_build_templates') == 'yes' then
```
micheljung: Invert condition to avoid nesting:
```
if Prefs.GetOption('accept_build_templates') ~= 'yes' then
    return
end
```

lua/ui/game/build_templates.lua
```diff
@@ -0,0 +1,118 @@
...
+        if tab then
+            import('/lua/ui/game/announcement.lua').CreateAnnouncement(LOC('<LOC template_0000>Build Template Received'), tab, LOCF('<LOC template_0001>From %s', sender))
+        end
+        AddTemplate(msg.data)
+    end
+end
+
+function GetInitialName(template)
+    for _, entry in template do
+        if type(entry) != 'table' then continue end
```
micheljung: Use lua's `~=` instead of `!=`

lua/ui/game/build_templates.lua
```diff
@@ -0,0 +1,118 @@
...
+        return (string.gsub(__blueprints[entry[1]].Description, '^<[^>]*>', '')) -- removes <LOC xyz_desc> from name
+    end
+end
+
+function GetInitialIcon(template)
+    for _, entry in template do
+        if type(entry) != 'table' then continue end
+        if DiskGetFileInfo('/textures/ui/common/icons/units/'..entry[1]..'_icon.dds') then
+            return entry[1] -- Original unit found
+        else
```
micheljung: Since you returned in the `if`, there doesn't need to be an `else` block

lua/ui/game/build_templates.lua
```diff
@@ -0,0 +1,118 @@
...
+    for i, template in templates do
+        if i == templateID then continue end
+        if template.key and template.key == key then
+            used = true
+            break
+        end
+    end
+    if used then
+        return false
+    else
```
micheljung: same here; `else` not needed

Uveso:
@micheljung 
I am not sure what you mean.
All your comments are related to the original programmscripts. All those "unnecessary nesting" and "!=" where already there. I don't change it.

Or do you want me to optimize build_templates.lua in general like whitespace cleaning etc. ?

micheljung:
You asked for a review. The diff shows that all this code was added. I reviewed the diff :-)

aeoncleanse:
Did you check the Commits, @micheljung ? GPG code was added in the first one.

micheljung:
The single commits are irrelevant, the sum of them is what counts (because that's what will change in the end) and that I find in "Files changed"

aeoncleanse:
Except for in the case of when files have been added! Then, one must look at the _true_ diff, not the overall.

micheljung:
I reviewed *all* the code. I don't know what else you want me to do

Uveso:
Well, then let me optimize `build_templates.lua` before it get merged.

aeoncleanse:
Well, I suppose that you didn't find anything wrong with any of the stuff Uveso changes is a good thing? ^^ :)

micheljung:
In the one string replacement, one comment, one log change, no :-)
But I just found another little thing in the last piece of code where he actually did something ;D

lua/ui/game/build_templates.lua
ChrisKitching: (Please do this sort of thing nearly all of the time: deeply nested if/else blocks get very confusing, and you usually end up with a stack of "else return" junk at the bottom)

lua/ui/game/build_templates.lua
ChrisKitching: I'm slightly surprised this didn't fail CI: don't we have a CI pass that checks if the code is valid Lua? (Something that's useful to keep tool integrations working nicely, as well as to avoid silly mistakes)

lua/ui/game/build_templates.lua
```diff
@@ -0,0 +1,114 @@
+#*****************************************************************************
+#* File: lua/modules/ui/game/build_templates.lua
+#* Author: Ted Snook
+#* Summary: Build Templates UI
+#*
+#* Copyright � 2007 Gas Powered Games, Inc.  All rights reserved.
+#*****************************************************************************
```
ChrisKitching: Note: This commit is technically illegal.

Uveso:
This is the optimized scipt. i can't commit it, because of an error in github.
But if this is ok i can make a new and propper commit.

```LUA
----------------------------------------------------------------------------
-- File: lua/modules/ui/game/build_templates.lua
-- Author: Ted Snook
-- Summary: Build Templates UI
--
-- Copyright © 2007 Gas Powered Games, Inc.  All rights reserved.
----------------------------------------------------------------------------

local templates = Prefs.GetFromCurrentProfile('build_templates') or {}
local UIUtil = import('/lua/ui/uiutil.lua')

function CreateBuildTemplate()
    GenerateBuildTemplateFromSelection()
    local template = GetActiveBuildTemplate()
    ClearBuildTemplates()
    if table.getsize(template) > 0 then
        AddTemplate(template)
    end
end

function Init()
    import('/lua/ui/game/gamemain.lua').RegisterChatFunc(ReceiveTemplate, 'Template')
end

function ReceiveTemplate(sender, msg)
    if Prefs.GetOption('accept_build_templates') ~= 'yes' then return end
    local tab = import('/lua/ui/game/construction.lua').GetTabByID('templates')
    if tab then
        import('/lua/ui/game/announcement.lua').CreateAnnouncement(LOC('<LOC template_0000>Build Template Received'), tab, LOCF('<LOC template_0001>From %s', sender))
    end
    AddTemplate(msg.data)
end

function GetInitialName(template)
    for _, entry in template do
        if type(entry) ~= 'table' then continue end
        return (string.gsub(__blueprints[entry[1]].Description, '^<[^>]*>', '')) -- removes <LOC xyz_desc> from name
    end
end

function GetInitialIcon(template)
    for _, entry in template do
        if type(entry) ~= 'table' then continue end
        if DiskGetFileInfo('/textures/ui/common/icons/units/'..entry[1]..'_icon.dds') then
            return entry[1] -- Original unit found
        end
        if UIUtil.UIFile('/icons/units/' .. entry[1] .. '_icon.dds', true) then
           return entry[1] -- Modded unit found.
        end
    end
    return 'default' -- If we don't find a valid IconName; return string 'default'
end

function AddTemplate(newTemplate)
    table.insert(templates, {templateData = newTemplate, name = GetInitialName(newTemplate), icon = GetInitialIcon(newTemplate)})
    Prefs.SetToCurrentProfile('build_templates', templates)
end

function GetTemplates()
    return Prefs.GetFromCurrentProfile('build_templates')
end

function RemoveTemplate(templateID)
    table.remove(templates, templateID)
    Prefs.SetToCurrentProfile('build_templates', templates)
end

function RenameTemplate(templateID, name)
    templates[templateID].name = name
    Prefs.SetToCurrentProfile('build_templates', templates)
end

function SetTemplateIcon(templateID, iconPath)
    templates[templateID].icon = iconPath
    Prefs.SetToCurrentProfile('build_templates', templates)
end

function SendTemplate(templateID, armyIndex)
    armyIndex = armyIndex
    if table.getn(templates[templateID].templateData) > 22 then
        UIUtil.QuickDialog(GetFrame(0), "<LOC build_templates_0000>You may only send build templates with 20 or less buildings.",
            "<LOC _Ok>", nil, nil, nil, nil, nil, true,  {worldCover = true, enterButton = 1, escapeButton = 1})
        return
    end
    SessionSendChatMessage(armyIndex, {Template = true, data = templates[templateID].templateData})
end

function SetTemplateKey(templateID, key)
    for i, template in templates do
        if i ~= templateID and template.key == key then
            return false
        end
    end
    templates[templateID].key = key
    Prefs.SetToCurrentProfile('build_templates', templates)
    return true
end

function ClearTemplateKey(templateID)
    templates[templateID].key = nil
    Prefs.SetToCurrentProfile('build_templates', templates)
end
```

aeoncleanse:
Just add the changes as a new commit on top...

lua/ui/game/build_templates.lua
```diff
@@ -36,19 +36,23 @@ end
 function GetInitialName(template)
     for _, entry in template do
         if type(entry) != 'table' then continue end
-        return __blueprints[entry[1]].Description
+        return (string.gsub(__blueprints[entry[1]].Description, '^<[^>]*>', '')) -- removes <LOC xyz_desc> from name
```
ChrisKitching: Don't do this! Use the LOC function.

In case you don't know, the `<LOC xyz>` tag specifies the ID of this string in the localisation DB. Everything else in the string is optional descriptive information (usually the english version of the string to make things more readable when seen in code).
The correct way to display such strings in the UI is to call the global `LOC()` function on them. This replaces the string with the value in the localisation database (the files in the `loc` directory: https://github.com/FAForever/fa/tree/develop/loc ) for the language the user has configured.

If no `LOC` tag exists in the string, `LOC()` is a no-op.

There is also `LOCF` which allows pattern substitution in localisation strings.

What you have done here uses the optional (effectively meaningless) descriptive string as the name. The practical effect will be that names will default to english even for non-english players.

lua/ui/game/construction.lua
```diff
@@ -484,15 +484,19 @@ end
 function GetBackgroundTextures(unitID)
     local bp = __blueprints[unitID]
     local validIcons = { land = true, air = true, sea = true, amph = true }
-    if not validIcons[bp.General.Icon] then
-        if bp.General.Icon then WARN(debug.traceback(nil, "Invalid icon" .. bp.General.Icon .. " for unit " .. tostring(unitID))) end
-        bp.General.Icon = "land"
+    local icon = "land"
+    if unitID and unitID ~= 'default' then
```
ChrisKitching: The first half of this conditional is redundant: `unitID ~= 'default'` implies that `unitID` is truthy, since `'default'` is truthy. Comparison to `nil` is permitted, so this change is completely safe.

lua/ui/game/construction.lua
```diff
@@ -484,15 +484,19 @@ end
...
+        if not validIcons[bp.General.Icon] then
+            if bp.General.Icon then WARN(debug.traceback(nil, "Invalid icon" .. bp.General.Icon .. " for unit " .. tostring(unitID))) end
+            bp.General.Icon = "land"
```
ChrisKitching: This mutates the blueprint!

ChrisKitching:
> i can't commit it, because of an error in github.

Just force-push your branch.



> Just add the changes as a new commit on top...

Do not do that.

ChrisKitching:
> I reviewed all the code

Reviewing GPG code isn't _super_ useful to do upfront. Unfucking that is probably a separate task, perhaps one for a different PR (since their code is fairly uniformly terrible)

Uveso:
The error while commiting was caused by changing the Fileformat from ANSI to UTF-8...

ChrisKitching:
> [Edit] btw, why we cant use !=. Its hard to type ~= after 32 years of typing != :)

The main reason is that != is a GPG-specific extension to Lua, so tools (such as IDE support) for Lua don't parse != correctly.

ChrisKitching:
I should also add: use of # instead of -- for comments is another GPG Lua extension we should try and avoid, for the very same reason. Standard Lua parsers can't accept it, so it breaks nice IDE support and so on.

Uveso:
Ah i see, then the 2nd case ('!') only exist inside the SupCom internal lua interpreter.
```
      case '~': {
        next(ls);
        if (check_next1(ls, '=')) return TK_NE;
        else return '~';
      }
      case '!': {
        next(ls);
        if (check_next1(ls, '=')) return TK_NE;
        else return '!';
      }
```
Thanks for explaining this !

Sheeo:
@Uveso where do you have that source from?

Uveso:
@Sheeo 
this source is not from SupCom.exe itself, if you mean that.

I used this to patch a normal lua interpreter to see/understand why i can or can't use "!=" in lua.

Uveso:
Last small update/optimization comitted.
From my side this is ready to test/merge.

lua/ui/game/build_templates.lua
```diff
@@ -0,0 +1,99 @@
...
+    end
+    templates[templateID].key = key
+    Prefs.SetToCurrentProfile('build_templates', templates)
+    return true
+end
+
+function ClearTemplateKey(templateID)
+    templates[templateID].key = nil
+    Prefs.SetToCurrentProfile('build_templates', templates)
+end
```
aeoncleanse: Newline please? :)

PR #1323, opened 2016-05-28
ACU kills now trigger a GPGNet-command to enforce ratings server-side

Brutus5000:
In preparation for https://github.com/FAForever/server/issues/109

The trigger is sent for every ACU kill even though 1 time would be enough, but this can be handled server side.

Suggestions are welcome.

### DIFF
--- a/lua/UserSync.lua
+++ b/lua/UserSync.lua
@@ -77,4 +77,8 @@ function OnSync()
             import('/lua/ui/dialogs/teamkill.lua').CreateDialog(Sync.Teamkill)
         end
     end
+	
+    if Sync.EnforceRating then
+        GpgNetSend('EnforceRating')
+    end
 end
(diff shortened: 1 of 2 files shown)


aeoncleanse:
This will currently send for _any_ death reason. Don't we only want this to trigger when an enemy is the cause of death?

Brutus5000:
Woops. Made some cleanups and must have moved it. It's supposed to be inside the next if clause. I will fix it.

aeoncleanse:
Do we want to report if it is a team-kill?

Brutus5000:
I think we do. Otherwise you could "last-minute-kill" your partner before your opponents can kill him and then leave the game and prevent rating.

aeoncleanse:
Merged 44349849ce6340182c5d6d9dd0413179b53dd0c9

PR #1325, opened 2016-05-29
This will enable the game to load merged blueprints from ingamedirectory

Uveso:
I found a way to merge unitblueprints like the Hydrocarbon's from inside the gamedirectory.
I added a smal loop inside Blueprints.lua to load merge blueprints from the new "unitshook" folder.

### DIFF
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -88,10 +88,13 @@ end
 local function StoreBlueprint(group, bp)
     local id = bp.BlueprintId
     local t = original_blueprints[group]
-
     if t[id] and bp.Merge then
+        local child = bp.Merge
         bp.Merge = nil
         bp.Source = nil
+        if type(child) == 'string' and t[child] then
+            t[id] = table.merged(t[id], t[child])
+        end
         t[id] = table.merged(t[id], bp)
     else
         t[id] = bp
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -594,13 +594,12 @@ function LoadBlueprints(pattern, directories, mods, skipGameFiles, skipExtractio
                 safecall("Blueprints Loading org file "..file, doscript, file)
             end
         end
-    end
-
-    -- Loading ingame Merge-Blueprints from /unitshook/
-    for k,file in DiskFindFiles('/unitshook', '*_unit.bp') do
-        SPEW('Merging Blueprint file: '..file)
-        BlueprintLoaderUpdateProgress()
-        safecall("Blueprints Loading org file "..file, doscript, file)
+        -- Loading ingame Merge-Blueprints from /unitshook/
+        for k,file in DiskFindFiles('/unitshook', '*_unit.bp') do
+            SPEW('Merging Blueprint file: '..file)
+            BlueprintLoaderUpdateProgress()
+            safecall("Blueprints Loading org file "..file, doscript, file)
+        end
     end
 
     local stats = {}
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -595,6 +595,14 @@ function LoadBlueprints(pattern, directories, mods, skipGameFiles, skipExtractio
             end
         end
     end
+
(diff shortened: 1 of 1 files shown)


Uveso:
Adding multi id and category search to merge blueprint function.

originalmerge with merge = true and a single blueprint ID:
```LUA
UnitBlueprint {
    Merge = true,
    BlueprintId = "uab1102",
    General = {
        UnitName = 'Test Dummy',
    },
}
```

Also accepting multi unit ID's
```LUA
UnitBlueprint {
    Merge = true,
    BlueprintId = {
        uab1102=true,
        ueb1102=true,
        urb1102=true,
        xsb1102=true,
    },
    General = {
        UnitName = 'Test Dummy',
    },
}
```

And also accepted Category search: Merge = array, BlueprintID will be ignored
```LUA
UnitBlueprint {
    Merge = {
        'TECH1',
        'MASSEXTRACTION',
        'STRUCTURE',
    },
    BlueprintId = "",
    General = {
        UnitName = 'MassDummy',
    },
}
```

aeoncleanse:
I cannot think of any reason I'd use the final one :/

Uveso:
I cant tell you that. I also prefer the original style with single merge.
If i want to modify more Blueprints with the same value i would use ModBlueprint() or similar.

Its just an attempt to see, if this will fit Crotalus need with those inheritance.

I could use it, if its in the game, but i don't really need it.

Crotalus:
A usage for "category filter merge" could be if you want to set some specific weapon targeting for all units within a certain category, on the other hand we risk a clear overview of where stuff is actually set (like the GSR stuff dynamically set).

With inheritance I more meant something like this:

abc/abc1103.bp:
```
UnitBlueprint {
<all common stuff between all mass extractors>
}
```

uab1103/uab1103.bp:
```
UnitBlueprint {
    Merge='abc1103'
    Categories = {
        'AEON',
    },
    Display = {
        -- aeon specific stuff here
    }
}

**Comment by Uveso** [comment]
ok i changed a little bit.

We need 2 blueprints:

Parent_unit.bp
```LUA
UnitBlueprint {
    Merge = "child",
    BlueprintId = "uab1102",
    General = {
        Icon = 'air',
    },
}
```
child_unit.bp
```LUA
UnitBlueprint {
    General = {
        UnitName = 'NamedFromChild',
    },
}
```
Merge = "child" will lead to merge everything from "child_unit.bp" to "unit uab1102", then
"Parent_unit.bp" will be merged as normal to "unit uab1102".

Now you can use the child as base and you can still overwrite things from child with the parent blueprint.

aeoncleanse:
This parent child thing makes no sense to me. Why do we want all this complexity?

Crotalus:
Tested the inheritance here:
https://github.com/FAForever/fa/compare/develop...Crotalus:bp-inheritance

Base ASF file:
https://github.com/FAForever/fa/blob/4eb824fa7c9fcdda0a6548df24c9a89b7ff6442c/units/base/asf_unit.bp

https://github.com/FAForever/fa/blob/4eb824fa7c9fcdda0a6548df24c9a89b7ff6442c/units/URA0303/URA0303_unit.bp

Seems to work fine, we could probably get rid of even more boilerplate from each separate race-bp, audio sources etc should be able to be autogenerated, maybe mesh stuff as well depending on how consistent the mesh paths are.

(For weapons we maybe should provide some new `DPS` or similar so you can specify that in a parent template and then get the correct value per shot calculated automatically depending on a child units RateOffire)

Crotalus:
You don't see any sense in inheritance? Hmm, all base structures for instance use same blueprints value except the things that differ visually.

Crotalus:
Yeah maybe you're right for once, it's probably pointless.

Uveso:
Then back to the roots :)
I restored the StoreBlueprint function.
Now the PR only contains the ability to merge from UnitsHook folder.

Does it makes sense to make a new PR with only 1 Commit, then this one ?

aeoncleanse:
Yeah probably rebase it into sensible commit history.

Uveso:
closed and uploaded a clean PR here:  #1339

PR #1326, opened 2016-05-29
Transfering Hydracarbon blueprints to new unitshook folder

Uveso:
Deleted the full blueprints and insert the merge blueprints for Hydros to the new unitshook folder

This is for testing the new merge function from #1325


aeoncleanse:
This may possibly be actual genius. I'm not sure, my geniusometer broke...

Uveso:
hehe thanks :)

you can create directories to sort all those merge blueprints.
Like /unitshook/General.Icon/ for the hydrofixes or  /unitshook/WeaponBalance/ and so on.

You can also have multiple merge-blueprint of the same unit in different folders. 
They will all be merged.

Lets test this a week or so. Maybe Hussar can have a look at his UnitManger and check if i messed something up there.

If this works stable i will cut all unitblueprints to smal pieces and move every change to a seperate merge blueprint.

This way we will have a good overview over all blueprint changes and i can sort out all the useless stuff.

aeoncleanse:
Don't do the latter. We need to keep FAF mod friendly. For now, we'd just use this for new stuff to avoid importing further GPG code. We can then look at backtracking to un-shadow some files, but we should keep to fa/unitshook/unitID/.bp

Uveso:
Well, first of all i will not make a single PR with 455 Unit changes :dancer: 
If i do this i will make a PR for a single unit and we can discuss this.

What do you exactly mean with mod friendly ?
A mod can't see the merge process. For a mod this is a original untuched blueprint.

We are merging the blueprint right after loading the original.
It's in terms of programming the same like we modify the original blueprint without a merge.

And i was also thinking userfriendly; For us programmer.
For an example i want to change the lodcutoff of all units. I would like do do this in a seperate folder then editing all existing merges in unitshook root dir.
And if i messed this all up i can simply delete the lodcutoff subdir and all changes are redone.

Or maybe this can convince you:
Lets think about the 1st April.
A merge-subdir with changes to all unit-weapons and make them to mini Nukes.
I dont want to be the one on 2nd April who can delete all those merges, if we don't do this in a seperate folder *smile*

Well, first lets see if this works stable and play around with this. Then we can discuss, how to use it.
And i promise i don't cut all blueprints to smal pieces without asking hehe

aeoncleanse:
Mod friendly means for the modders. It should look the same layout as a mod does to avoid confusing people. It's more important to see what's changed _for a unit_ than to see which units have a bunch of sweeping changes.

For silly events, course we can do sweeping. It just shouldn't be the norm :)

Crotalus:
Another improvement could be to be able to be explicit on what blueprint you want to merge, i.e:
```Merge = 't1mex_base'```

and then let all mexes "inherit" from a base blueprint, that way we would get rid of duplicate code and always be sure of them having same HP etc. In short, inheritance for blueprints.

**Comment by Uveso** [comment]
@iceddreamer:
Ah yes i see. This will let still space for a directory structure like this:
```
/unitshook/UAB1102/UAB1102_unit.bp
/unitshook/UAB1102_Icons/UAB1102_unit.bp
/unitshook/UAB1102_Patch/UAB1102_unit.bp
/unitshook/UAB1102_DontKickmePlease/UAB1102_unit.bp
```
Or do you really mean we should go the classic way ? - "/unitshook/UAB1102/UAB1102_unit.bp" only ?

@Crotalus:
i have not the slightest idea how to do this.
This is the loading code:
```LUA
        for k,file in DiskFindFiles('/unitshook', '*_unit.bp') do
            BlueprintLoaderUpdateProgress()
            safecall("Blueprints Loading org file "..file, doscript, file)
        end 
```
BlueprintLoaderUpdateProgress is located inside ForgedAlliance.exe at 00A15EF0 and 00B654 5C. So this is out of range here. And this already stops me from going on.
And the function safecall is located in utils.lua and uses "pcall(fn, unpack(arg))". I dont know how to get any data from the merge blueprint here.

Crotalus:
When you do that safecall what happens is that it will run the actual blueprint file and it contains 
```
UnitBlueprint {
}
```

which is shorthand for
```
UnitBlueprint({
})
```

so what happens is that UnitBlueprint() is called in system/Blueprints.lua, which in turn calls StoreBlueprint(), that's where the actual merge magic happens

aeoncleanse:
Better would be unithook/UID/patch.bp, icons.bp, moarstuffplox.bp

Uveso:
@Crotalus 
That was the missing link!!! Thanks alot Crotalus.
I did not know that LUA can execute functions this way. Nice:)

After a quick trace i can see the merge data as lua array inside StoreBlueprint.
And there is already a "if bp.Merge then". 
Well i know the entrypoint for the fix know.

Good test could be to find a way for merging the icon = amph for all carbons with only one mergefile.
Thanks, i have work to do :)

...Well actually i am going to sleep now :) But i will have a look at this tomorrow.
And btw, thanks for all the great help!

Uveso:
@aeoncleanse 
Yes, this is good. Let's do it this way.

@Crotalus 
Thanks a lot! This was the missing link.
I will take a look at this tomorrow. Now i have to sleep :)

Uveso:
Just a last idea before sleeping:
original:
```LUA
UnitBlueprint {
    Merge = true,
    BlueprintId = "uab1102",
    General = {
        UnitName = 'test dummy',
    },
}
```
new with Blueprintid Array for 4 units with 1 merge:
```LUA
UnitBlueprint {
    Merge = true,
    BlueprintId = {
        "uab1102" = true,
        "ueb1102" = true,
        "urb1102" = true,
        "xsb1102" = true,
    },
    General = {
        UnitName = 'test dummy',
    },
}
```

Uveso:
ahhhrg i have to write it down.
```LUA
local function StoreBlueprint(group, bp)
    local function Store(group, bp, targetid)
        local t = original_blueprints[group]
        if t[targetid] and bp.Merge then
            bp.Merge = nil
            bp.Source = nil
            t[targetid] = table.merged(t[targetid], bp)
        else
            t[targetid] = bp
        end
    end
    if type(bp.BlueprintId) == 'string' then
        local targetid = bp.BlueprintId
        Store(group, bp, targetid)
    elseif type(bp.BlueprintId) == 'table' then
        local Source = bp.Source
        local Merge = bp.Merge
        for targetid in bp.BlueprintId do
            bp.Merge = Merge
            bp.Source = Source
            bp.BlueprintId = targetid
            Store(group, bp, targetid)
        end
    end
end
```

This will do the trick with the multi id array.

Uveso:
i updated #1325 
Now merging accepts Multiple Unit ID's and Category Search.

Uveso:
Depends on #1339 ! :)

aeoncleanse:
Closed due to non-acceptance of #1339

PR #1332, opened 2016-05-31
Make ratings greyer when rating deviation is higher

ChrisKitching:


ChrisKitching:
It ends up roughly like this for deviation >=150:  https://www.dropbox.com/s/ndaepfjifq4wr2q/Screenshot%202016-05-31%2020.32.28.png?dl=0

Linearly transitioning to white as it goes down to 100.

aeoncleanse:
Yeah, this works. I like it. :+1:

Issue #1333, opened 2016-05-31
Submerge/Surface button


shalkya:
http://forums.faforever.com/viewtopic.php?f=42&t=12523&p=127771#p127771

labels : 
nice to have
ui (?)
enhancement
low

Uveso:
Well, pressing the dive button sends this:
`
IssueCommand(GetUnitCommandFromCommandCap(RULEUCC_Dive))
`
This command is send to all selected units.
But thisis a toggle command. You cant tell all units to dive or surfacing, only to toogle their actual state.
(You cant change this like the shields in #1262 )

maybe possible to deselect all units who are already dived, issue a command to all others to dive, and the reselect all units.

But deselect units and reselecting them is more a hack then a good solution.
Maybe someone else has a better idea.

Crotalus:
IssueUnitCommand takes a list of units instead of selection so maybe that would work? Otherwise we could invent a new simcallback which takes care of the dive / surface logic depending on unit state.

Another problem with the dive button is the apparent "lag" before the unit change state from surfaced / submerged which is extra problematic with the toggling

Uveso:
Crotalus! Thanks again!

IssueUnitCommand did the trick !

now it was easy.

PR #1335, opened 2016-06-01
Fix Submerge/Surface Button

Uveso:
More information:
#1333

SUB's dive-toogle button has now the same behavior then Shield toogle

If submerged and surfaced SUB's are selected, the Dive button will first submerge all SUB's.
If you have other units inside the selction, they will be ignored. Only SUB's are processed.

### DIFF
--- a/lua/ui/game/orders.lua
+++ b/lua/ui/game/orders.lua
@@ -155,7 +155,7 @@ end
 local function CreateOrderButtonGrid()
     controls.orderButtonGrid = Grid(controls.bg, GameCommon.iconWidth, GameCommon.iconHeight)
     controls.orderButtonGrid:SetName("Orders Grid")
-	controls.orderButtonGrid:DeleteAll()
+    controls.orderButtonGrid:DeleteAll()
 end
 
 -- local logic data
@@ -248,9 +248,9 @@ end
 -- used by things that build weapons, etc
 local function BuildOrderBehavior(self, modifiers)
     if modifiers.Left then
-	    IssueCommand(GetUnitCommandFromCommandCap(self._order))
+        IssueCommand(GetUnitCommandFromCommandCap(self._order))
     elseif modifiers.Right then
-	    self:ToggleCheck()
+        self:ToggleCheck()
         if self:IsChecked() then
             self._curHelpText = self._data.helpText .. "_auto"
             self.autoBuildEffect = CreateAutoBuildEffect(self)
@@ -261,7 +261,7 @@ local function BuildOrderBehavior(self, modifiers)
         if controls.mouseoverDisplay.text then
             controls.mouseoverDisplay.text:SetText(self._curHelpText)
         end
-	    SetAutoMode(currentSelection, self:IsChecked())
+        SetAutoMode(currentSelection, self:IsChecked())
     end
 end
 
@@ -279,7 +279,7 @@ end
 -- used by subs that can dive/surface
 local function DiveOrderBehavior(self, modifiers)
     if modifiers.Left then
-	    IssueCommand(GetUnitCommandFromCommandCap(self._order))
+        IssueCommand(GetUnitCommandFromCommandCap(self._order))
         self:ToggleCheck()
     elseif modifiers.Right then
         if self._isAutoMode then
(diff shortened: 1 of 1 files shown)


lua/ui/game/orders.lua
```diff
@@ -279,7 +279,31 @@ end
 -- used by subs that can dive/surface
 local function DiveOrderBehavior(self, modifiers)
     if modifiers.Left then
-	    IssueCommand(GetUnitCommandFromCommandCap(self._order))
+        local unitList = GetSelectedUnits()
+        local submergedSUB = false
+        local surfacedSUB = false
+        -- Searching the unitlist for SUB's and memorizing submerged and surfaced state
```
Crotalus: Maybe do something like this to remove the need for the second loop:
```
local unitList = EntityCategoryFilterDown(categories.SUBMERSIBLE, GetSelectedUnits())
local surfaced, submerged = {}, {}

for _, v in unitList do
    if GetIsSubmerged({v})
        table.insert(submerged, v)
    else
        table.insert(surfaced, v)
    end
end

if submerged[1] and surfaced[1] then
    IssueUnitCommand(submerged, GetUnitCommandFromCommandCap(self._order))
else
    IssueCommand(GetUnitCommandFromCommandCap(self._order))
end
```
(should be seen as pseudo-code, didn't test it)

Uveso:
Hey Crotalus :)
I know what you mean. Filling an array wit all surfaced subs, and send a command with the array instead of a loop and single commands.

But i did it intentionally without an array.
1st reason: The game has one bottleneck. Memory flow-rate. So i try to use not to much memory like arrays etc.

And the 2nd reason for an 2nd loop was some calculation...
The 2nd loop will ony executed at ~25% of all keypresses.
Because, if u pressed it first, all SUBs are in the same state after then, and will not trigger the 2nd loop again.

So we save at 75% rate the building of an unneded Array.

To optimise the calculation speed, we could bail out with "break" from the first loop after the first
surfacedSUB = true AND submergedSUB = true.
But this will need an additional if-then loop. (Memory!)
And the whole loop is only executed on a keypress. So we don't need the best performance here.

What do you think ?

Crotalus:
Yeah it's hard to know exactly what is a win or not, especially when we don't have any clue about what's going on behind the scene.  I guess it could be tested, spawn 500 subs and time the implementations :)

If I would guess I think the actual `IssueUnitCommand` would be the most impactful here in a network game at least. Try spamming stop-key in a network game and you grind it all down. Every call to that function will require additional stuff written to replay stream, sent over network etc.

But as you say, the code is executed very seldom so I guess it's not a big deal anyway. .

Uveso:
Well, impact to the network is a good point. I did not tested it.
But i can change the 2nd loop to create first the unitarray and then send it with 1 Command.
This way the array will only be build if really needed, and we save the network :)

Uveso:
Oh Wow. The networkimpact is more then huge.

Old Loop 499 single commands:
DiveOrderBehavior() Runtime Loop 1: (0.000 seconds) seconds.
DiveOrderBehavior() Runtime Loop 2: (0.067 seconds) seconds.

The first peek represent the 499 single commands. The 2nd peak is the single command to all units.
timingoldloop

And the new Loop with single command with SUB-array:
DiveOrderBehavior() Runtime Loop 1: (0.000 seconds) seconds.
DiveOrderBehavior() Runtime Loop 2: (0.002 seconds) seconds.

Well as expected now, both peeks are equal:
timingnewloop

Well let's do it with an Array in this case :D

Crotalus:
Nice testing! With 500 commands in a row the command cost is quite noticable indeed.

`ren_showframetimes` is sometimes interesting also to see how much time is spent in sim / render etc, `ren_frametimeseconds` to scale it timewise.

aeoncleanse:
Given the above investigation, is this ready to roll Uveso?

Uveso:
Yes, in my eyes this is ready to roll.

Just to remind:
The dive button will now submerge all selected SUBs, if they are in differen states. (surfaced/submerged)

aeoncleanse:
Merged bbc824b904c24d5b8db2cb8d0d17fc02ce9e6936

Issue #1336, opened 2016-06-01
dummy weapon on T2 aeon transport for GGs


shalkya:
http://forums.faforever.com/viewtopic.php?f=42&t=11956

Uveso:
please close.

PR #1337, opened 2016-06-02
Adding a 0 Damage Weapon to Guide Weaponfire of LABs

Uveso:
More Info: #1336

I added a Weapon with 0 damage to the Aeon Transporter to guide the weaponfire from loaded units.

Dont know if there is a better way, but at least it works :)

### DIFF
--- a/units/UAA0104/UAA0104_script.lua
+++ b/units/UAA0104/UAA0104_script.lua
@@ -25,6 +25,7 @@ UAA0104 = Class(AirTransport) {
         SonicPulseBattery2 = Class(AAASonicPulseBatteryWeapon) {},
         SonicPulseBattery3 = Class(AAASonicPulseBatteryWeapon) {},
         SonicPulseBattery4 = Class(AAASonicPulseBatteryWeapon) {},
+        GuidanceSystem = Class(AAASonicPulseBatteryWeapon) {},
     },
 
     -- Override air destruction effects so we can do something custom here
--- a/units/UAA0104/UAA0104_unit.bp
+++ b/units/UAA0104/UAA0104_unit.bp
@@ -546,6 +546,36 @@ UnitBlueprint {
             Turreted = true,
             WeaponCategory = 'Anti Air',
         },
+        {
+            Damage = 0,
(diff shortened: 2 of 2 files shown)


aeoncleanse:
Why bother with this? Nobody complains about the T1 GG. I say we do it to all four T1 Transports as well, or to none.

Uveso:
They maybe don't complain, because LAB's can fire when loaded on all T1 transports.
The AEON T2 transport was the only transport where those units don't fire.

aeoncleanse:
The units should fire just fine. The complaint is that, unlike the other T2 transports, you can't use the transport's weapon as a target guide to force them to fire _at a specific target_.

Uveso:
Yes you are right, with T1 transporter and T2 Aeon Transport you have to micro the transporter over the target but the units can fire. (Also on the T2 Aeon.)

The dummyweapon only change the behavior of the transporter to stay/cycle over the target.

aeoncleanse:
@Zockyzock This one's your territory. The dummy allows LABs to track a specific target by attacking that target with the transports. IMO either everything (Including T1 Trans) gets this, or we leave it the way it is. Your call.

Uveso:
@aeoncleanse 
Just in case we need to add this to T1 Transports, i dont have an idea how to make it.
Just adding a dummy weapon don't work here.
It works on the T2 Aeon, because it has already a working weapon (only air), but the dummy weapon gives the ability to "aim" at groud targets.
If you have no weapon or only a dummy weapon, it will not get any target at all.

This will lead to ugly solutions like a aktive weapon with a smal Turret angle, so the turret can't aim at anything and can't fire then.
In this case, the weapon will fire sometimes. If you are flying uphill and the transporter guns has the right angle to the target, it will fire 1-2 shots.

aeoncleanse:
It's not too tough. We'll wait on Zock's answer before worrying about it.

Zockyzock:
There is no need for t1 or Aeon t2 transports to target fire.

aeoncleanse:
Closed: Zock verdict

PR #1339, opened 2016-06-03
Merging Blueprints from UnitsHook folder

Uveso:
New PR with clean history.
More Info: #1325


aeoncleanse:
Sorry to be a bother, but does this give the ability to do a list of bps in one merge file? That's the one neat trick from the other PR I'd like carried over in addition to simple merges. Also, look for any .bp, not just *_unit.bp please :)

Uveso:
Blueprints can now merged in normal style like:

/unitshook/UAB1102/UAB1102_unit.bp
```LUA
UnitBlueprint {
    Merge = true,
    BlueprintId = "uab1102",
    General = {
        Icon = 'amph',
    },
}
```
And you can use a ID Array instead of a string:

/unitshook/HydroIcons.bp
```LUA
UnitBlueprint {
    Merge = true,
    BlueprintId = {
        "uab1102",
        "ueb1102",
        "urb1102",
        "xsb1102",
    },
    General = {
        Icon = 'amph',
    },
}
```
Log will show in this case:
```
DEBUG: Merging Blueprint : /unitshook/hydroicons.bp
DEBUG: Merging to Blueprint : uab1102
DEBUG: Merging to Blueprint : ueb1102
DEBUG: Merging to Blueprint : urb1102
DEBUG: Merging to Blueprint : xsb1102
```

aeoncleanse:
Do we need the Hydro changes here? I think not. Can you move that multi-ID merge to your other PR branch with the Hydro changes.

@ChrisKitching - To my eye this works and does what I'd like it to. It also seems to play fine with the unit manager. Is there anything you can see that's glaringly wrong?

Uveso:
@aeoncleanse,
This blueprint was only for debugging. I forgot to delete it. :/

aeoncleanse:
Right, bad news on this. I've had a discussion with Chris, and he thinks that the value added by this is outweighed by the increased complexity of tracking down what each bp will finally look like in-game, so we should simply continue to shadow all the bp files that we use.

Sheeo:
We're not done with that discussion @aeoncleanse.

aeoncleanse:
News to me!  I'll leave you and chris to fight it out :)

Uveso:
@Sheeo 
I really need those merges / hooks. :)
Since i am reworking some Units mods, i like to make some changes also to the original units.
There are many changes like Lod-cutoffs  that need to be changed on every unit.
If i am allowed to do my changes, i have to shadow every single unit.
Maybe hooks are better in this case.

aeoncleanse:
For a mod, you can mod the blueprint loading in the mod. If for main game, why do all LOD's need changing???

Uveso:
Because the LOD setting on almost any unit is not reflecting the real proportions and altitude.

Also there is no need to display a unit, if the unit is smaler then the smales Strategic icon. You can't se the unit anyway.
And Aircrafts are higher then land units, so we should see them a bit longer if we zoom out.

I already changed the most lods in some Unit mods. and it looks far better and feels better then before.
And because we don't display smal units under the strategig icon, the game will run faster.

aeoncleanse:
right...

Uveso:
At least we know how to do it, if we will need it in the future :)

PR #1342, opened 2016-06-05
Add Self Destruct Confirmation

### LINKED ISSUE #1341: Implement Self Destruct confirmation
Discussion here: http://forums.faforever.com/viewtopic.php?f=42&t=12537

Mod that does this is called Suicide Confirmation V3

Would be nice to have this optional. Enable/disable in options.

speed2CZ:
Fixes #1341 
Asks before self destruction selected units. By default asks only when ACU is selected. Can be turned on for all units or turned off completelly.

### DIFF
--- a/etc/faf/blacklist.lua
+++ b/etc/faf/blacklist.lua
@@ -94,6 +94,9 @@ Blacklist = {
     ['1749EEN4-DE86-5CC2-39AC-35692BDE76XF'] = UPGRADE,
     ['EEFFA8C6-96D9-11E4-9DA1-460D1D5D46B0'] = UPGRADE,
 
+    -- Suicide Confirmation
+    ['BB236B78-CB23-4E78-8CD7-BA55AE987545'] = INTEGRATED,
+
     -- ==================================================================
     -- INTEGRATED mods due to added preset restrictions in Units Manager:
     -- ==================================================================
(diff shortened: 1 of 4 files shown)


speed2CZ:
It seems to work only for normal self destruct (the one with 5sec delay), need help with making it work for instant one as well. Didnt manage to get it working.

Nojoke185:
few comments:

1)
set confirm self destruct to all units (on)
the prompt shows when you control k a unit
the prompt also shows when you cancel the control k on a unit
would it be better to only show for the control k and not the cancel of the same?

2)
having set the confirm self destruct to all units (on)
change this to ACU only
select a non ACU unit
control K
the prompt shows
according to the settings this should not happen now
same issue seen when setting confirm self destruct to off

speed2CZ:
Did you restart the game when changing the settings? I know I had issue with more settings applying after a restart so when I was testing it I did restart it.

Nojoke185:
Hi

No I did not restart game between setting changes (On/ACU only/Off)
If a restart is required then 2nd comment is not a problem then

Thanks

Exotic-Retard:
this need for a restart probably needs to be written down somewhere so other people dont think its an issue like that

aeoncleanse:
Well if it's done like that it just won't go in. We need that toggle to work in-game.

aeoncleanse:
Sorry speed, this one's remaining a UI mod.

PR #1344, opened 2016-06-06
Make AI players swappable in the lobby

Justify87:
Makes AI players swappable in the lobby. Changed the OwnerID of the AI to 999999, so the system recognizes the difference between an AI and the host during swapping, because we are swapping players/AI with the use of the observer list as a temporary "bucket".

This solves the issue with #1309 where AI slots would get cleared if a human swaped with an AI.


aeoncleanse:
I'm merging this into the 'ice' feature branch for testing, because it ties closely in to what that branch is about (Getting the labels all showing and map-based player switching working).

aeoncleanse:
cb886ca9ec81ef622b522897cc3ca6a306f39ced

Issue #1346, opened 2016-06-07
Option are greyed out and unavailable for the host, if host is an observer and opens&closes the options dialog


Justify87:
1. host a game
2. make yourself an observer
3. open the options dailog
4. close the dialog
5. ability to change the game options is not available anymore for the host

Justify87:
GetPlayersNotReady() doesn't return any players, if all slots are empty. So the system thinks the game is ready to launch. I added an if clause that checks if host is an observer.

aeoncleanse:
88f470e732bcf68443434e4e4ae02e2f5f0c43fc

PR #1350, opened 2016-06-07
Fixes game options disappearing when host is observer

Justify87:
Related issue #1346

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -5368,13 +5368,16 @@ function InitHostUtils()
             -- Is at least one person not ready?
             local playerNotReady = GetPlayersNotReady() ~= false
 
-            UIUtil.setEnabled(GUI.gameoptionsButton, playerNotReady)
-            UIUtil.setEnabled(GUI.defaultOptions, playerNotReady)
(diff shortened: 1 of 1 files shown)


Justify87:
Different solution:
Host should be allowed to be observer and launch the game. Someone might
host a game just to cast it as an observer.

GetPlayerCount() only counts the players/AI that occupy a game slot

aeoncleanse:
This doesn't seem to work. Four AIs in slots, host in observer, change map, and the options button greys out.

Justify87:
Strange. When I tested it, It worked. I will take a look.

aeoncleanse:
Take a look at ice branch. I just posted a working solution.

lua/ui/lobby/lobby.lua
```diff
@@ -5368,13 +5368,16 @@ function InitHostUtils()
             -- Is at least one person not ready?
             local playerNotReady = GetPlayersNotReady() ~= false
```
ChrisKitching: Best approach is probably to change GetPlayersNotReady() to return true when there are no players.

That'd be a consistent way to solve this whole class of problems.,

Justify87:
@aeoncleanse 

Options get unavailable as soon as I (the host) become an observer, with no AI or Players present in any slot. Closed slots are ignored too.

PR #1351, opened 2016-06-08
Turn off deflate compression

Crotalus:
Some players crash due to deflate compression algorithm trying to unpack corrupt packets.

Some reports:
http://forums.faforever.com/viewtopic.php?f=3&t=12552
https://forums.faforever.com/viewtopic.php?f=3&t=9811
https://forums.faforever.com/viewtopic.php?f=3&t=10542

If we need compression we can do that in the FAF client instead, multicore ftw ;) (although I guess FA probably runs this in own thread)

### DIFF
--- a/lua/userInit.lua
+++ b/lua/userInit.lua
@@ -13,6 +13,8 @@ doscript '/lua/globalInit.lua'
 local AvgFPS = 10
 WaitFrames = coroutine.yield
 
+ConExecute('net_CompressionMethod 0')
+
 function WaitSeconds(n)
     local start = CurrentTime()
     local elapsed_frames = 0


aeoncleanse:
Is there any way to test this other than simply rolling it out?

Crotalus:
You can check it's off with `net_CompressionMethod` in console while in a game. If I remember correctly there also is some message in log about deflate compression used in log when joining a lobby, that message should be changed. To see that it actually isn't compressing packages you would have to use WireShark or similar network sniffer.

Testing that it plays nicely in a network game is another matter, I've tested with 3 clients in a LAN which seems to work fine.

ChrisKitching:
> Some players crash due to deflate compression algorithm trying to unpack corrupt packets.

Turning off compression will only help this problem is the corruption is caused by the compression. Do we have any evidence that this is the case?

If the corruption is caused by another, random source (such as UDP transport corruption), disabling compression will tend to increase the incidence of corrupt packets, because the amount of data being sent will be larger (and hence there will be more data exposed to potential corruption).

aeoncleanse:
Yeah this is what i meant, not testing if the change is functional, bit if its a solution at all. I guess we could get aome people who have the problem a lot to test it out on beta?

Crotalus:
Yeah we can only guess about the reasons for deflate() to crash the game.

IIRC UDP's header checksum  doesn't protect the payload. When we now route all packets through the client we maybe could add a CRC to the payload as well? That way we can drop (or track / log) malformed packets before they hit the game, crashing it. Wouldn't be a 100% proof system though.

Another thing is packets that arrive out of order, not sure how the game handles those, maybe related to the `net_ackdelay` value?

micheljung:
How could we even disable it? I tried this once but gave up

Crotalus:
You need to set `net_CompressionMethod 0` before creating the lobby. I've confirmed getting it to work by looking at the UDP packets.

Sheeo:
I missed this discussion by partaking in private with varying parties.

Just so we're clear; UDP guarantees that your packets arrive as-sent by means of CRC checking, so we're definitely dealing with a game compression/uncompression bug here (Which part we can't say, because of the following). In addition; the error message emitted by the game is highly suspicious:

    info: Error decompressing data from ...: CDeflateOutputFilter::BufWrite(): call to deflate() failed.

In that you will not be _writing_ compressed data to a buffer that you're receiving data from.


While a wrongful error message cannot prove the code is also wrong, it's suspicious at the very least.

Crotalus:
Ok, I've tested what happens if you simulate a corrupt UDP packet arriving... not pretty unfortunately.  With deflate on the game crashes with the error message above. If you turn deflate off it hangs the sim instead. Seems like GPG trusted UDP to always deliver sane packets.

ChrisKitching:
> Seems like GPG trusted UDP to always deliver sane packets.

What with CRC, isn't it supposed to always return sane packets, or no packets at all?

Crotalus:
The UDP handler won't see that the packet is corrupt, the checksum validates the corrupt payload. Not something that happens often but still possible. Exactly how often I have no idea, The naive assumption would be that it lucks 1 per 65535 corrupted packets.

Now also tried in a real FAF game which was over anyway, 4 people crashed instantly so the packets get over the net without problem.

micheljung:
@Crotalus yes, but how do you set `net_CompressionMethod 0`?

Uveso:
https://github.com/FAForever/fa/pull/1351

While i dont know much about the zlib and how the comression works, i see the forum post's in a statistical way.

If i read the first post, i see:
info: Error decompressing data from lmlicenses.wip4.adobe.com:54801: CDeflateOutputFilter::BufWrite(): call to deflate() failed.
(lmlicenses.wip4.adobe.com:54801 means at least its not from the FAF serVer or another FAF player)

on the second post:
info: Error decompressing data from 79-117-163-203.rdsnet.ro:6112: CDeflateOutputFilter::BufWrite(): call to deflate() failed.
info: Error decompressing data from c27-253-56-230.thoms4.vic.optusnet.com.au:6112: CDeflateOutputFilter::BufWrite(): call to deflate() failed.
info: Error decompressing data from ip179-100-211-87.adsl2.static.versatel.nl:6112: CDeflateOutputFilter::BufWrite(): call to deflate() failed.

3rd Post:
info: Error decompressing data from 70-59-73-228.mpls.qwest.net:6112: CDeflateOutputFilter::BufWrite(): call to deflate() failed.


And 19.June 2014:
Source: http://forums.faforever.com/viewtopic.php?f=26&t=7855
info: Error decompressing data from 92.38.128.139:6112: CDeflateOutputFilter::BufWrite(): call to deflate() failed.


A Post from 18.May.2009
Source: http://forums.demigodthegame.com/352065
Uhyre: I have an error with the CDeflateOutputFilter::BufWrite(): call to deflate() failed string with Supreme Commander: Forged Alliance. Almost every multiplayer game crashes after several minutes leaving this in error log.

Another post from 14.May 2009:
Source: http://forums.stardock.com/352018
info: Error decompressing data from secure.disc-soft.com:2614: CDeflateOutputFilter::BufWrite(): call to deflate() failed.
(like the adobe one, secure.disc-soft.com:2614 is not a FAF server or a FAF Player)

I found several Errors like this in 2015 and 2016. So this error/issue was there every time.

But there are only a few Player who get this error.
If this where a FAF related problem, way more player would have this problem, and it would not explan the adobe and Secure.Disc-Soft errors.

In my eyes the problem is not the zlip compression inside the game.
It a problem with the internetconnection. Mostly Wlan or not using a RJ45 cable in general.

But in this case its only my opinion (statistical view) not knowledge.

ChrisKitching:
> The UDP handler won't see that the packet is corrupt, the checksum validates the corrupt payload. Not something that happens often but still possible.

Right, you can corrupt the payload and the checksum in complementary ways so the checksum is still "correct", but now you're screwed either way, aren't you? Now you have a corrupt message. It doesn't matter if it's compressed or not, it's still going to break everything. Why would disabling compression change that?

aeoncleanse:
Yeah this looks like a general setup  issue and nothing we can prevent. I'd love to know how packets from all those random-ish sources seem to be making it to the engine of this game though, that seems very odd. Or are they not, but are crashing an OS service the game depends upon?

Crotalus:
@micheljung `ConExecute('net_CompressionMethod 0')`

@ChrisKitching Yup, game crashes both with / without deflate so probably not a problem with the compression function itself.

Adding an extra CRC checksum to the payload data maybe could reduce the risk even more, the other option would be to actually check that the packets are sane - which would require parsing of the payload but not even that would cover all cases... at least if someone has a malicious intent.

So disabling deflate is pointless for now, closing.

PR #1360, opened 2016-06-16
Add terrain avoidance to tempest

Exotic-Retard:
it moves up when over terrain so its torpedo tubes dont clip into it and stop firing. when moving into deeper water it moves down a little bit, but not too much.

This is on a slider. this means the actual position of the unit doesn't change, so its ok for small movement but making it move way deeper would need more work.


Exotic-Retard:
i wanted to pull from a different branch so closing this one,

PR #1365, opened 2016-06-20
Improve campaign base manager expansions

speed2CZ:
One more value IncomingEngineers was needed in ExpansionBaseData of base
manager. It tracks engineers that are sent into expansion base.
If expansion base needs engineers is determined by number of engineers
working in expansion base manager + engineers that are on the way to
expansion.

### DIFF
--- a/lua/AI/OpAI/BaseManager.lua
+++ b/lua/AI/OpAI/BaseManager.lua
@@ -1,11 +1,11 @@
-#****************************************************************************
-#**
-#**  File     :  /lua/ai/OpAI/BaseManager.lua
-#**
-#**  Summary  : Base manager for operations
-#**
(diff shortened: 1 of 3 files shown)


aeoncleanse:
Nice one @speed2CZ . Could you please re-write this a little to hook the lua rather than shadowing it, since you are only touching a small part of the file that you imported? Thanks.

speed2CZ:
No, I dont know how, and it's just the beginning. There are more things to fix and improve with campaign AI.

aeoncleanse:
OK I'll do it. I await future AI PRs :p

speed2CZ:
so @aeoncleanse what's it gonna be?

aeoncleanse:
2a99c9cf7b0afc6baffe1ab1a218c104e503e980

PR #1367, opened 2016-06-23
Add terrain avoidance to tempest (proper branch)

Exotic-Retard:
(this one is from a nicer branch for me, but otherwise identical to #1360)

it moves up when over terrain so its torpedo tubes dont clip into it and stop firing. when moving into deeper water it moves down a little bit, but not too much.

This is on a slider. this means the actual position of the unit doesn't change, so its ok for small movement but making it move way deeper would need more work.

### DIFF
--- a/units/UAS0401/UAS0401_script.lua
+++ b/units/UAS0401/UAS0401_script.lua
@@ -39,14 +39,12 @@ UAS0401 = Class(ASeaUnit) {
         ChangeState(self, self.IdleState)
         
         if not self.SinkSlider then --setup the slider and get blueprint values
-            WARN('creating slider')
             self.SinkSlider = CreateSlider(self, 0, 0, 0, 0, 5, true) --create sink controller to overlay ontop of original collision detection
             self.Trash:Add(self.SinkSlider)
         end
         
-        local bp = self:GetBlueprint()
-        self.elevation = bp.Physics.Elevation or 0 -- backup value is -15
-        --self.SinkSlider:SetGoal(difference)
+        --local bp = self:GetBlueprint() -- saved for later, sets depth to elevation in bp
+        --self.elevation = bp.Physics.Elevation -- saved for later, sets depth to elevation in bp
         self.WatchDepth = false
         
     end,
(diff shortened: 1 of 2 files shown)


aeoncleanse:
This PR doesn't seem to include the new animation?

aeoncleanse:
Or am I getting this confused with the HARMS stuff like the nooblet I am? :D

aeoncleanse:
@Zockyzock Do we classify this as a fix? This code allows them to go below ground-fire range.

Zockyzock:
@aeoncleanse "them" is tempests, or all subs? The possibility to ground fire on underwater units need to be consistent. (until some potential changes happen to ground fire in far future)

If this change makes tempest not groundfireable, but subs still are, its maybe something for the future, but not yet.

aeoncleanse:
This affects tempest and atlantis only

HUSSAR-mtrela:
does this PR fixes issue with multiple tempest units stop moving when they colliding with each other?

Exotic-Retard:
i guess some clarification is needed - this elevates the model above the seabed when it gets too shallow. it doesnt actually make them go deeper, just adds the ability to do so.

by elevating them above the seafloor it means they clip through the ground less, and thats a good thing because their torpedo tubes are not stopped from firing (as much)

this is a bug fix. not a balance change. 

It could be applied to all submerged units if needed, however since they are smaller that would actually make a difference. these two are so big they have the most issues and just so you know groundfiring is extremely inconsistent anyway due to the way the game sets elevation. (its absuuuuuuurd)

aeoncleanse:
So submerging in the first place works, but if you go into a shallow area and out again, it won't re-dive properly. What's up?

Exotic-Retard:
thats engine stuff if you mean actual unit position. was always present. very annoying but sadly true. i tried fixing that before making this fix, and failed miserably.

if you mean the slider gets stuck somehow then i have no idea at all.

aeoncleanse:
Well, if the unit comes _up_ for beaches or whatever, but then gets stuck at a higher-than-normal elevation unless you toggle the submerge button off and on, this entire PR just seems pointless... Am I wrong?

Exotic-Retard:
no you are wrong. this pr is purely a bugfix for the torpedo tubes getting stuck in the terrain. thats all it does, via animations. if the unit gets to shallow water thats pretty much guaranteed to happen, regardless of whether it will go back down or not. the slightly going down partial fix there, its kinda a side effect and not the purpose of this.

aeoncleanse:
You're using sliders. These change the physical location of the unit mesh... There is no animation change here..

Exotic-Retard:
sliders are animators but simple ones, in any case the point is the mesh is changed but not the unit position (hitbox, meshextents, lifebars, ect)

PR #1368, opened 2016-06-23
Make sure wrecks are destroyed after rebuild

### LINKED ISSUE #1366: Queued SCU starts at 50% after rebuild of Quantum Gate
Not sure if this is true for presets only.
I was playing as cybran

1. Build a quantum gate
2. Build RAS SCU preset
3. Quantum gate got destroyed during build of SCU
4. Rebuild quantum gate on top of wreck
5. Queue RAS SCUs
6. Build will start at a discounted price of 50% mass and energy. ^^

There is actually a wreck of the SCU standing in the quantum gate. I guess this is the reason it gets the 50% bonus. All following SCUs will also start at 50% as long as the wreck remains in the quantum gate.

before build

after build

Replay ID: 4863652
(Water of isis)

Crotalus:
Factories which contain a wreck of a partial built unit got rebuild bonus without wreck being destroyed.

Fixes #1366

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2128,6 +2128,12 @@ Unit = Class(moho.unit_methods) {
                 local progress = p:GetFractionComplete() * 0.5
                 -- set health according to how much is left of the wreck
                 unit:SetHealth(self, unit:GetMaxHealth() * progress)
+
+                -- Clear up wreck after rebuild bonus applied if engine won't
+                if not unit.EngineIsDeletingWreck then
+                    p:Destroy()
+                end
+
                 return
             end
         end


aeoncleanse:
Errrm... I guess this is one way, but why are half-built units in factories being left as wrecks _at all_? Surely it's easier and more... reliable to simply ensure that FactoryUnit destruction destroys anything it's building? Or that Factories not be allowed to use the rebuild bonus feature?

Crotalus:
65412a8b1a9224fe70b970da946d6fb1d65fa355 changed the behavior from destroy -> leave wrecks in factories.

Justify87:
So does this change mean, that the wreck disappears after the first unit is build with 50% bonus? Or does it get removed before the factory/quantum gate starts building again?

Crotalus:
The build progress will get updated according to the wreck value (same as other rebuild stuff) and then the wreck disappears. In this case it's kinda unloreish that a half-gated in unit even leaves a wreck.

As mentioned earlier we also could remove wreck at the start of building a unit, although then people would need to reclaim the wreck manually to not lose mass. Option 2 would be to leave the wreck untouched and set progress to 0% instead. Option 3 would be to revert the wreckage of half-built units.

aeoncleanse:
Given that this was added as a feature then, I give it a :+1:

aeoncleanse:
Hang on. From where is getrebuildbonus called, and in what circumstance would it not be, therefore requiring this change's code to be activated?

Justify87:
@Crotalus, I guess you mean that since the quantum gate isn't really a factory, it shouldn't get treated as such. Units don't get produced, they gate in. Mhm. Difficult. If somone would build a cloak preset, that would be a huge mass and energy investment wasted right there.

But then you have to question the fact, why assisting a quantum gate with build capacity, would shorten the time to gate through? Since you can't assist the hypothetical quantum gate the SACU is sent from.

aeoncleanse:
Errrm. This doesn't actually... Work...

aeoncleanse:
Merged bdca949bee921bd0c14f06747250583ed3672011

PR #1372, opened 2016-06-27
Fix desync issues with coop

Crotalus:
Never use GetFocusArmy() with anything related to sim state, this will desync game.

### DIFF
--- a/lua/ScenarioPlatoonAI.lua
+++ b/lua/ScenarioPlatoonAI.lua
@@ -651,7 +651,7 @@ end
 
 --####################################################################################################################
 --### SplitPatrolThread
---###     - Gives random patrol chain from the list to each unit of a platoon 
+--###     - Gives random patrol chain from the list to each unit of a platoon
 --###
 --### PlatoonData
 --###     - PatrolChains - List of chains to choose from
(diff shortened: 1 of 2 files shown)


speed2CZ:
Wouldn't it be better to actually decide what army you want to attack? Since purpose of this was to snipe player's stuff. Not sure what GetCurrentEnemy() does but it could choose some other ai instead of player I assume?

Also is it used somewhere else than in campaign?

Crotalus:
Yeah if it only is in campaign the old code is desync safe, not in coop though.

Maybe should make sure it grabs a human enemy at least. What GetCurrentEnemy() is depends on the AI how it picks its enemies.

speed2CZ:
I think it is only in campaign, would make sense if it desync games otherwise.

So I'd suggest to pass it what army it should look for the same way it gets what category to look for. Via PlatoonData.

speed2CZ:
And if enemy army isnt specified, look for human players. To make sure we dont break it everywhere where its used right now. 
What do you think?

aeoncleanse:
@Crotalus Any progress here?

Crotalus:
Ran a test with this code against the current easily reproducable desyncs in coop (Prothyon mission 2 for instance). It didn't help against those.

I noticed lua/SimObjectives.lua uses lots of GetFocusArmy() which probably will split the code paths, not sure if this is used in coop?

Crotalus:
With `sallyshears` on mission 2 didn't desync so that verifies problems with the SimObjective, I changed GetFocusArmy() to GetPlayerArmy() instead which will return the primary player id in the game. This assumes that all players are allies in coop, sharing the same intel.

speed2CZ:
While you're at it, can you also look at file lua\cinematics.lua (it was not added to faf yet). It also contains GetFocusArmy() several times, so I'd assume it could cause problems.

aeoncleanse:
344a4395fd995874a8666e75ead58f47f2dc703d

PR #1375, opened 2016-06-30
Bugfix EXCEPTION_ACCESS_VIOLATION (...) at address 0x006cbc63

Uveso:
Related Forum Post:
http://forums.faforever.com/viewtopic.php?f=31&t=11289
http://forums.faforever.com/viewtopic.php?f=3&t=11430
http://forums.faforever.com/viewtopic.php?f=3&t=12582
http://forums.faforever.com/viewtopic.php?f=3&t=12570

To reproduce the bug:
Select 4 turtle AIx (cybran) on any 10x10 Map and let them fight.
With AI Cheat and Buildfactor of 1.5 the game will freeze/crash after 18-25 minutes with the Error EXCEPTION_ACCESS_VIOLATION (...) at address 0x006cbc63

This was caused by using the wrong plattons for unit "xrb0104" (Engineering Station).

The short story: A building can't move :)

I excluded ENGINEERSTATION from EngineerBuildAI and TransferAI platoons.
Now xrb0104 is only using a platton with ManagerEngineerAssistAI plan.

### DIFF
--- a/lua/AI/PlatoonTemplates/EngineerPlatoonTemplates.lua
+++ b/lua/AI/PlatoonTemplates/EngineerPlatoonTemplates.lua
@@ -77,7 +77,7 @@ PlatoonTemplate {
     Name = 'T2EngineerBuilder',
     Plan = 'EngineerBuildAI',
     GlobalSquads = {
-        { categories.ENGINEER * categories.TECH2, 1, 1, 'support', 'None' }
+        { categories.ENGINEER * categories.TECH2 - categories.ENGINEERSTATION, 1, 1, 'support', 'None' }
     },
 }
 
@@ -85,7 +85,7 @@ PlatoonTemplate {
     Name = 'T2EngineerTransfer',
     Plan = 'TransferAI',
     GlobalSquads = {
-        { categories.ENGINEER * categories.TECH2, 1, 1, 'support', 'none' },
+        { categories.ENGINEER * categories.TECH2 - categories.ENGINEERSTATION, 1, 1, 'support', 'none' },
     },
 }
 
@@ -101,7 +101,7 @@ PlatoonTemplate {
     Name = 'CybranT2EngineerBuilder',
     Plan = 'EngineerBuildAI',
     GlobalSquads = {
-        { categories.CYBRAN * categories.ENGINEER * categories.TECH2, 1, 1, 'support', 'None' }
+        { categories.CYBRAN * categories.ENGINEER * categories.TECH2 - categories.ENGINEERSTATION, 1, 1, 'support', 'None' }
     },
 }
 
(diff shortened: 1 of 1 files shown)


aeoncleanse:
Brilliant!

Sheeo:
Nice find.

Crotalus:
Very nice! Do you know exactly what part of the EngineerBuildAI / TransferAI that causes the crash? 

That way we maybe could overload the culprit engine function and prevent it from getting faulty parameters in case something else calls it in this way.

Uveso:
@Crotalus 
I wished it where so easy.
The game don't crash on a specific command or function. At least it don't crash while executing EngineerBuildAI or TransferAI.

I also checked almost every command to find out what exactly is causing this. But i don't found it.

I am almost sure it has something to do with movement, because if i made the Engineerstation moveable, the game don't crash.

If we want the exact reason we could disassemble the SupCom.exe with IDA and see what at adress 6cbc63 is causing the error. But this isnt fun at all and will not be done in a few hours.

aeoncleanse:
5c4592653dffd3c4250418162e436e114865764c

Stuntaneous:
The forums seemingly won't send a goddamn activation email for an account there so here is my response. 

Issue persists as of a day ago. Steam version of FA. It occurred early, maybe like 15 minutes into the match, it was 1v1 against a non-cheating normal (not air, water, etc) Sorian AI. I was UEF, they were Cybran. I think the speed was set to adjustable but not changed in-game. The map was.. I forget the name, but.. green with water surrounding one major island and two small islands in the top left and bottom right corners. The land on the main island composing of two semicircular areas with two crossing points. 

http://pastebin.com/7AE6E99Z

Uveso:
@ Stuntaneous
Your EXCEPTION_ACCESS_VIOLATION Is at address 0x0133d267.
This PR / bugfix is related to the error at address 0x006cbc63

And welcome to FAForever ;)

Btw, even if we have patched this here inside the development version you still can't play it on FAForever. It needs some testing before it will be merged to a public beta or release version.

Stuntaneous:
Ello. Thought it may be relevant. If not, I'm not sure where to put it, or if I even can anywhere on the forums as the activation email remains elusive. I've got another for your collection.. same map, same AI, etc, the AI was Aeon though and this took longer at something like 40 minutes. 

http://pastebin.com/6RLUFqvu

Uveso:
@Stuntaneous 
Every EXCEPTION_ACCESS_VIOLATION is relevant. But first we should get you an forum account.

Maybe you don't received a confirmation email because your emailprovider is blocked. Most of the temporary email providers (gmx, gmail etc.) are blocked. Also Yahoo is blocking traffic from FAF. So try again with another email.
In the meantime we can continue here:
#1381

Stuntaneous:
Ah, it's a gmail address. I don't remember the last time I had an issue with it being blocked, it isn't exactly hotmail or the days of free email back when everyone used hotmail! I can use another address but I'm reluctant as I'm more likely to hold onto the gmail account over time.

PR #1374, opened 2016-06-27
Fix not being able to launch AI. Fixes #1370

aeoncleanse:

Issue #1370, opened 2016-06-25
Can't start Game with AI and AI faction is not selectable.


Uveso:
Skirmish and multiplayer:

AI don't has a puldown menu to select faction.
![nopulldown](https://cloud.githubusercontent.com/assets/17804547/16358720/28e8ee7a-3b1c-11e6-921b-a55fb886e4ae.jpg)

And if any Ai is selected, on buttonpress "gamestart" this error occurs, and the game don't start:

```LUA
WARNING: Error running OnRelease script in <deleted object>: ...03\gamedata\mohodata.scd\lua\system\localization.lua(129): bad argument #2 to `format' (string expected, got nil)
         stack traceback:
         	[C]: in function `format'
         	...03\gamedata\mohodata.scd\lua\system\localization.lua(129): in function `LOCF'
         	...programdata\faforever\repo\fa\lua\ui\lobby\lobby.lua(3251): in function `EveryoneHasEstablishedConnections'
         	...programdata\faforever\repo\fa\lua\ui\lobby\lobby.lua(1578): in function <...programdata\faforever\repo\fa\lua\ui\lobby\lobby.lua:1532>
         	...programdata\faforever\repo\fa\lua\ui\lobby\lobby.lua(2779): in function `OnClick'
         	c:\programdata\faforever\repo\fa\lua\maui\button.lua(85): in function <c:\programdata\faforever\repo\fa\lua\maui\button.lua:78>
```

(Also tried english language. Same effect)
[EDIT] Also tried normal AI, not only Sorian's.

aeoncleanse:
Yes, this seems to have broken is cb886ca9ec81ef622b522897cc3ca6a306f39ced. Working on a fix...

aeoncleanse:
OK so I got it to launch with an exception on line 3246 to let it ignore id == 999999, but I still don't have the dropdown, and I can't figure out why it doesn't work.

Uveso:
If you like, i can take a look at it.

Uveso:
Well,

The Owner ID of an Ai is not longer the Host; its 999999 now.

-OwnerID = hostID,
+OwnerID = 999999,

This will lead into a false "owned" status if we Check  an AI Slot:

```LUA
function IsLocallyOwned(slot)
    return (gameInfo.PlayerOptions[slot].OwnerID == localPlayerID)
end
```
We have to change the function like this:

```LUA
function IsLocallyOwned(slot)
[...]

PR #1376, opened 2016-06-30
Fix/swapplayers

aeoncleanse:
This stops the player swap mechanism using a stupid "Empty>move to observers>move>move from observers" model, and allows AI switching properly.

PR #1378, opened 2016-07-03
BugFix Recognise new AI OwnerID 999999 as locally owned

Uveso:
Smal patch for #1374 issue(#1370)


ChrisKitching:
Doesn't #1376 influence this work rather significantly?

Uveso:
@ChrisKitching 
Oh yes, after merging 1376 this pr makes no sense. :)
closed !

PR #1379, opened 2016-07-03
Bugfix Play nuke VO for observers

Uveso:
the old IF ELSE statement was playing the nuke warning for non observers x-times (for every player sync one warning). This was making the sound verry loud.

New If Else only play the nuke warning once, also if you are an observer the sound will only played once (for player index 1 only)


aeoncleanse:
Good catch mate

PR #1383, opened 2016-07-06
Adds various enhancements to the large map preview

Justify87:
- large map preview size is now set to 95% of window height to make use of emtpy screen space on modern systems
- large map preview now gets properly destroyed when closing
- added various labels to the large map preview (player name, rating,
number of games, faction)
- added checkbox to hide labels on the large map preview
- added DepthOverParent to the team number label


aeoncleanse:
Large map shouldn't be destroyed on close, it should be hidden. This is to avoid the performance hit of making it again and again and again.

Justify87:
I tested this ony my VM many times, which has basicly no performance at all, and did not notice any performance issues between hiding and destroying the large preview.

aeoncleanse:
@ChrisKitching ? - Though I think he'll say "Every little helps"

ChrisKitching:
I'll look at this properly later, but some things to know:

Creation of Bitmaps is surprisingly expensive. I wrote a `TexturePool` class to cache the mass, hydrocarbon, and reclaim icons. Search the file for "TexturePool" to see how this is used. It works by keeping a bunch of bitmaps off-screen and hidden, unhiding them and providing them to the caller on request (or creatng a new one if it's run out of stored ones).

Depending on how you "destroyed" the map preview, if you didn't destroy the TexturePools you would have avoided most of the performance hit.

If your icons are per-user, there are so few of them it is not useful (or wise) to bother with a pool. If (as I think you are) you are only destroying your new icons, this definitely is no big deal from a performance perspective.The pool is helpful because of the very large number of icons that have to be created all at once (for all the mass points and bits of wreckage), which was creating a 1-ish second freeze when opening the preview when I was testing it at the time.

"It runs sanely in my stupid VM" is probably a pretty good way to check if something is stupid. :P

lua/ui/lobby/lobby.lua
```diff
@@ -4162,14 +4160,15 @@ function CreateBigPreview(parent)
 
     -- The LrgMap shouldn't be destroyed due to issues related to texture pooling. Evil hack ensues.
     local onTryMapClose = function()
-        LrgMap:Hide()
-        LrgMap.isHidden = true
+        LrgMap:Close()
+        LrgMap = nil
+        return
```
ChrisKitching: This is a memory leak. Don't do this.

This will cause the Lua objects to be garbage collected without causing Destroy to be called on the bitmaps held in the texture pools.

I don't understand why you want to destroy the preview instead of hiding it anyway. The only cost to keeping it around is memory (which we free up anyway when we delete the map preview when the game starts), and it permits it to be re-shown maximally quickly and without there being any fuss about object lifecycles and so on.

lua/ui/lobby/lobby.lua
```diff
@@ -4180,6 +4179,16 @@ function CreateBigPreview(parent)
     -- Keep the close button on top of the border (which is itself on top of the map preview)
     LayoutHelpers.DepthOverParent(closeBtn, mapPreview, 2)
 
+    local chkBoxLabels =UIUtil.CreateCheckbox(dialogContent, '/CHECKBOX/', "Show/Hide labels")
+    LayoutHelpers.AtLeftTopIn(chkBoxLabels, dialogContent, 10, 1)
+    LayoutHelpers.DepthOverParent(chkBoxLabels, dialogContent, 2)
+    chkBoxLabels.OnCheck = function(self, checked)
+        if checked == true then
```
ChrisKitching: `if checked then`

lua/ui/lobby/lobby.lua
```diff
@@ -4180,6 +4179,16 @@ function CreateBigPreview(parent)
...
+            mapPreview:HideLabels(MapUtil.LoadScenario(gameInfo.GameOptions.ScenarioFile))
```
ChrisKitching: If memory serves, LoadScenario doesn't employ caching...

lua/ui/controls/resmappreview.lua
```diff
@@ -244,21 +331,114 @@ ResourceMapPreview = Class(Group) {
...
+            if playerOptions[slot] then
+                self.slotInfoBG[slot]:Show()
+                self.slotInfoBGHead[slot]:Show()
+                self.gamesPlayedLabel[slot]:Show()
+                self.playerNameLabel[slot]:Show()
+                self.factionBitmap[slot]:Show()
+                self.ratingLabel[slot]:Show()
+            end
+        end
+
```
ChrisKitching: As well as `Hide()` and `Show()` I added a `UIUtil.SetVisible()`, which allows you to set hiddenness based on a boolean. Example use:
https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/lobby.lua#L926

Please use this instead of duplicating everything twice. If unclear, you'd have one function which takes a boolean and passes it to a bunch of calls to `setVisible` with each of the UI elements, and then you'd make `ShowLabels` and `HideLabels` just call that function with true and false respectively.

ChrisKitching: It's also worth knowing about `Group` (search in lobby.lua for examples of it). A `Group` is a UI element that doesn't alter what things look like, but allows you to apply layout and visibility functions to elements in aggregate. If you put a bunch of elements in a `Group` and `Hide()` it, all of the elements are hidden at once.

I think this will allow you to even further simplify this code.

lua/ui/controls/resmappreview.lua
```diff
@@ -244,21 +331,114 @@ ResourceMapPreview = Class(Group) {
...
+            end
+        else -- if no player is present, hide all labels. Since the rating label is shown on the little preview, it doesn't get really hidden
+            rating = ""
+            if self.showAdvancedPlayerStats == true then
+                self.slotInfoBG[slot]:Hide()
+                self.slotInfoBGHead[slot]:Hide()
+                self.gamesPlayedLabel[slot]:Hide()
+                self.playerNameLabel[slot]:Hide()
+                self.factionBitmap[slot]:Hide()
+            end
```
ChrisKitching: Put these in a Group, then this and the similar block above can be replaced with one call.

lua/ui/controls/resmappreview.lua
```diff
@@ -62,10 +75,36 @@ ResourceMapPreview = Class(Group) {
...
+            v:Destroy()
+        end
+
+        for k, v in pairs(self.slotInfoBG) do
+            v:Destroy()
+        end
+
+        for k, v in pairs(self.slotInfoBGHead) do
+            v:Destroy()
+        end
```
ChrisKitching: If you call `Destroy` on a `Group`, it recursively `Destroy`s all child elements.
That'd allow you to replace these five loops with one.

See `/lua/maui/group.lua` and my other comment (further down) about `Group`.

lua/ui/controls/resmappreview.lua
```diff
@@ -32,12 +37,20 @@ ResourceMapPreview = Class(Group) {
         self.hydromarkers = {}
         self.wreckagemarkers = {}
         self.startPositions = {}
+
         self.ratingLabel = {}
+        self.gamesPlayedLabel = {}
+        self.playerNameLabel = {}
+        self.factionBitmap = {}
+        self.slotInfoBG = {}
+        self.slotInfoBGHead = {}
```
ChrisKitching: I feel like what you really want here is a class to represent "The label we stick next to a player slot". This could extend `Group` and handle its own layout.

This would be a powerful abstraction: all the code here then can stop treating all the component parts separately, and just interact with it in a nice way. Rather like how the ComboBox control works (it encapsulates a bitmap, a label, and a bunch of logic to fiddle about with them).

lua/ui/controls/resmappreview.lua
```diff
@@ -199,6 +238,19 @@ ResourceMapPreview = Class(Group) {
...
+
+
+        -- get faction icons
+        local factionBmps = {}
+        local factionTooltips = {}
+
+        for index, tbl in FactionData.Factions do
+            factionBmps[index] = tbl.SmallIcon
+            factionTooltips[index] = tbl.TooltipID
+        end
```
ChrisKitching: There doesn't seem to be any point in copying these things to locals, just reference them directly? Make a local for `FactionData.Factions` if you want to make the lookups shorter, but there's no need to construct factionBmps or factionTooltips in this way.

lua/ui/controls/resmappreview.lua
```diff
@@ -32,12 +37,20 @@ ResourceMapPreview = Class(Group) {
...
+        self.factionBitmap = {}
+        self.slotInfoBG = {}
+        self.slotInfoBGHead = {}
 
         self.mapPreview = MapPreview(self)
         self.mapPreview.Width:Set(size)
         self.mapPreview.Height:Set(size)
         LayoutHelpers.AtLeftTopIn(self.mapPreview, self)
+
+        self.showAdvancedPlayerStats = showAdvancedPlayerStats or false
```
ChrisKitching: Since this should be a boolean, either `or false` is redundant or your code is flawed (and permits this to be a non-boolean).

lua/ui/controls/resmappreview.lua
```diff
@@ -209,16 +261,51 @@ ResourceMapPreview = Class(Group) {
...
+                self.gamesPlayedLabel[slot] = UIUtil.CreateText(self.mapPreview, '', 10, 'Arial Gras', true)
+                self.playerNameLabel[slot] = UIUtil.CreateText(self.mapPreview, '', 12, 'Arial Gras', true)
+                LayoutHelpers.AtLeftTopIn(self.playerNameLabel[slot], self.slotInfoBG[slot], 20, 1)
+                LayoutHelpers.AtLeftTopIn(self.gamesPlayedLabel[slot], self.slotInfoBG[slot], 2, 20)
+                LayoutHelpers.AtLeftTopIn(self.ratingLabel[slot], self.slotInfoBG[slot], 2, 35)
+
+                local bmp = Bitmap(self, UIUtil.UIFile(factionBmps[table.getn(factionBmps)]))
+                self.factionBitmap[slot] = bmp
+                LayoutHelpers.AtLeftTopIn(self.factionBitmap[slot], self.slotInfoBG[slot], 2, 2)
+                LayoutHelpers.DepthOverParent(self.factionBitmap[slot],self, 13)
```
ChrisKitching: All this crap would be way nicer to have in that custom Group class I mentioned earlier.

ACUButton is another example of the creation of an aggregated UI component like this (and one that's used in the large map preview)

lua/ui/controls/resmappreview.lua
```diff
@@ -209,16 +261,51 @@ ResourceMapPreview = Class(Group) {
...
 
-            startPositions[slot] = marker
+            if self.showAdvancedPlayerStats == true then
+                self.slotInfoBG[slot] = Bitmap(self)
+                self.slotInfoBG[slot]:SetSolidColor('FF212123')
+                self.slotInfoBG[slot].Width:Set(75)
+                self.slotInfoBG[slot].Height:Set(50)
+                self.slotInfoBG[slot]:SetAlpha(0.75)
+                LayoutHelpers.AtLeftTopIn(self.slotInfoBG[slot], marker, -45, -20)
+                LayoutHelpers.DepthOverParent(self.slotInfoBG[slot],self, 1)
```
ChrisKitching: You'd only need at most one of these DepthOverParent calls if you moved this stuff into an aggregate UI component class (one for the aggregate instance itself, instead of having to fiddle about with each element separately).

lua/ui/controls/resmappreview.lua
```diff
@@ -209,16 +261,51 @@ ResourceMapPreview = Class(Group) {
...
+
+                self.slotInfoBGHead[slot] = Bitmap(self)
+                self.slotInfoBGHead[slot]:SetSolidColor('FF212123')
+                self.slotInfoBGHead[slot].Width:Set(75)
+                self.slotInfoBGHead[slot].Height:Set(19)
+                self.slotInfoBGHead[slot]:SetAlpha(0.95)
```
ChrisKitching: You might like to use a local instead of writing `self.slotInfoBGHead[slot]` a billion times.

Justify87:
@ChrisKitching , thanks for your input. :-) I will redo all the code and make this label a class of its own using the group class.

I switched from hiding to detroying the large preview, because the comment above the line stating 

> "The LrgMap shouldn't be destroyed due to issues related to texture pooling. **Evil hack ensues**."

made me think, that hiding isn't the proper way of doing this. I guess my assumption was wrong.

lua/ui/controls/resmappreview.lua
Justify87: I did this because it should default to false, if not specified. Like the other one above (buttonDisabled)

PR #1386, opened 2016-07-12
Rotate acu towards mid

Crotalus:
To make start position more equal, otherwise north has a slight advantage when maps is about walking immediately to mid

### DIFF
--- a/lua/sim/ScenarioUtilities.lua
+++ b/lua/sim/ScenarioUtilities.lua
@@ -338,6 +338,7 @@ function CreateInitialArmyGroup(strArmy, createCommander)
                 cdrUnit:HideBone(0, true)
                 ForkThread(CommanderWarpDelay, cdrUnit, 3)
             end
+            cdrUnit:RotateTowardsMid()
         end
     end
 
--- a/lua/sim/ScenarioUtilities.lua
+++ b/lua/sim/ScenarioUtilities.lua
@@ -322,7 +322,6 @@ function CreateSubGroup(tblNode, strArmy, strGroup, ...)
     return platoonList, tblResult, treeResult
 end
 
-
 ----[  CreateInitialArmyGroup                                                     ]--
 ----[                                                                             ]--
 function CreateInitialArmyGroup(strArmy, createCommander)
@@ -338,7 +337,17 @@ function CreateInitialArmyGroup(strArmy, createCommander)
                 cdrUnit:HideBone(0, true)
                 ForkThread(CommanderWarpDelay, cdrUnit, 3)
             end
-            cdrUnit:RotateTowardsMid()
+
+            local rotateOpt = ScenarioInfo.Options['RotateACU']
+            if not rotateOpt or rotateOpt == 'On' then
+                cdrUnit:RotateTowardsMid()
+            elseif rotateOpt == 'Marker' then
+                local marker = GetMarker(strArmy) or {}
+                if marker['orientation'] then
+                    local o = EulerToQuaternion(unpack(marker['orientation']))
+                    cdrUnit:SetOrientation(o, true)
+                end
+            end
         end
     end
 
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1817,6 +1817,17 @@ Unit = Class(moho.unit_methods) {
         self:SetRotation(angle + current_yaw)
     end,
 
+    RotateTowards = function(self, tpos)
+        local pos = self:GetPosition()
+        local rad = math.atan2(tpos[1]-pos[1], tpos[3]-pos[3])
+        self:SetRotation(rad * (180 / math.pi))
+    end,
+
+    RotateTowardsMid = function(self)
+        local x, y = GetMapSize()
+        self:RotateTowards({x/2, 0, y/2})
+    end,
+
     OnStartBeingBuilt = function(self, builder, layer)
         self:StartBeingBuiltEffects(builder, layer)
         local aiBrain = self:GetAIBrain()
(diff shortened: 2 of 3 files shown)


ChrisKitching:
Isn't the rotation something that can be set by the map authors for each start position?

If so, this is probably not something we should override ourselves.

speed2CZ:
No, map authors only place markers with starting positions.

Sheeo:
They're denoted in the save file using army markers which _do_ have an orientation:

```lua
['ARMY_4'] = {
    ['color'] = STRING( 'ff800080' ),
    ['type'] = STRING( 'Blank Marker' ),
    ['prop'] = STRING( '/env/common/props/markers/M_Blank_prop.bp' ),
    ['orientation'] = VECTOR3( 0, -0, 0 ),
    ['position'] = VECTOR3( 330.5, 26, 50.5 ),
},
```

If this orientation isn't respected when placing the commander, we should just make it so?

aeoncleanse:
I agree. This should be done via the map scripts.

speed2CZ:
With that you mean updating every map so we have this cool feature? Idk if thats the best way to do it. Also its not super obvious to rotate the markers in the map editor. That might take ages before people learn to use it.

Crotalus:
No one will ever update old maps so the correct solution is probably to assume 0, 0, 0 as default rotation in which we rotate towards mid, otherwise use the one designated by the map creator.

Crotalus:
Seems like the orientation set in map wasn't respected at all so fixed that + lobby option to rotate the ACU towards mid.

speed2CZ:
Only if you wanted to break your map and it would require editing save.lua with some text editor. Map editor support only rotation around Y axis.

Other thing is I cant imagine game where you would not want ACU rotated. I think its an improvement that should be always enabled

speed2CZ:
I should also mention that you can't type the rotation in the editor to have it same for both players like you can do with units, but only rotate manualy with mouse which is not perfect.

ChrisKitching:
> I should also mention that you can't type the rotation in the editor to have it same for both players like you can do with units, but only rotate manualy with mouse which is not perfect.

Editor bugs should not influence code in not-the-editor.

speed2CZ:
Just that not many ppl will do it if it can't be done easily in the map editor.

Sheeo:
I like this solution.

Zockyzock:
I don't think we want to start inflating lobby options morr, if it's a good idea it can be implemented and be always enabled, but if there are significant downsides the small improvement doesn't warrant a new option.

JaggedAppliance:
Rotation works correctly, only problem is you can start walking before your acu has actually gated in :D

aeoncleanse:
@Crotalus What is the status on this? I think we want it defaulted, not an optional thing.

Crotalus:
Updated the functionality to always be on as default now without a visible lobby option. If a map / mod wants to disable the rotation it's possible to manually specify ScenarioOptions.RotateACU == 'Marker' to make it respect map marker rotation or 'Off' (in reality anything else than 'On' / 'Marker') to disable it.

Also rebased onto latest `develop`

lua/sim/ScenarioUtilities.lua
```diff
@@ -338,7 +337,17 @@ function CreateInitialArmyGroup(strArmy, createCommander)
...
             end
-            cdrUnit:RotateTowardsMid()
+
+            local rotateOpt = ScenarioInfo.Options['RotateACU']
+            if not rotateOpt or rotateOpt == 'On' then
+                cdrUnit:RotateTowardsMid()
+            elseif rotateOpt == 'Marker' then
+                local marker = GetMarker(strArmy) or {}
+                if marker['orientation'] then
+                    local o = EulerToQuaternion(unpack(marker['orientation']))
```
aeoncleanse: What does this function do?

lua/sim/Unit.lua
```diff
@@ -1817,6 +1817,17 @@ Unit = Class(moho.unit_methods) {
         self:SetRotation(angle + current_yaw)
     end,
 
+    RotateTowards = function(self, tpos)
```
aeoncleanse: This function is unused. Do we need it?

lua/sim/ScenarioUtilities.lua
```diff
@@ -338,6 +337,17 @@ function CreateInitialArmyGroup(strArmy, createCommander)
                 cdrUnit:HideBone(0, true)
                 ForkThread(CommanderWarpDelay, cdrUnit, 3)
             end
+
+            local rotateOpt = ScenarioInfo.Options['RotateACU']
+            if not rotateOpt or rotateOpt == 'On' then
+                cdrUnit:RotateTowardsMid()
+            elseif rotateOpt == 'Marker' then
+                local marker = GetMarker(strArmy) or {}
+                if marker['orientation'] then
```
aeoncleanse: To be clear, map-makers would need to set this value in which file, and where in that file?

Crotalus: In the corresponding map files, maps/<mapname>/<mapname>_save.lua. There's a 'orientation' key in the same data structure as the marker position for each start position. Not sure the current map editor supports changing it there (speed2 remark earlier in this PR).

lua/sim/ScenarioUtilities.lua
Crotalus: Converts the marker's euler-orientation to the corresponding quaternion values which SetOrientation likes.

https://en.wikipedia.org/wiki/Conversion_between_quaternions_and_Euler_angles

lua/sim/Unit.lua
Crotalus: `RotateTowards()` is used by `RotateTowardsMid()`. If you meant `Rotate()` I left that in if we in the future wants to rotate something relative to the current orientation.

aeoncleanse:
d16cdcba69fb297c0bf4ebf4f33a417b0054d160

Issue #1392, opened 2016-07-25
UEF gunships can't be given attack order when landed on water


JaggedAppliance:
As the title says, uef gunships that are landed on water can't be given an order to attack units, you have to give a move order so they take off then give the attack order.

Uveso:
After testing i know, you mean the Tech2 Gunship "Stinger", not the Tech2 Gunship.

This is easy to explain. Gunships can't land on water :)

The stinger is not only a gunship, its also a transporter!

The Transporter ability makes the Stinger land on water.
But he can't attack from water, because all weapons need the layer Air to aim and fire. But while landed, the Layer is Water. And the stinger dont has weapons, that can fire while in water :)

You can test this by building any transporter. While in air, the transporter can aim at enemies. But if landed it cant aim. like the Stinger.

Uveso:
I made a fix, ready for testing: #1393

Uveso:
hehe, no problem. It was not meant as a reproach :)

Uveso:
Merged into develop, i guess this can be closed.

Exotic-Retard:
somewhat skirting round the topic - how would one go about making non transporting gunships land on water?
or perhaps the opposite - making transports not land on water, is that possible?

aeoncleanse:
In short: No. Landing on water seems tied to the Transport category.

PR #1393, opened 2016-07-25
Fix UEF Gunship Stinger don't accept attack orders while on water

Uveso:
See issue #1392

This PR simply adds the Layer `Water` to `FireTargetLayerCapsTable`.
So the stinger will accept attackorders while floating on water.


Exotic-Retard:
its pretty interesting that only the uef t2 gunship can land on water, is there a reason for this? perhaps its linked to the transport clamp its got?

aeoncleanse:
Yes, it is :)

Nojoke185:
checked this one out and it seems to work fine
a landed T2 UEF gunship (on water) - when given an attack order goes and attacks the designated target

aeoncleanse:
ca884ee1446d1dde46c035f71177eb7d2175c7d2

PR #1400, opened 2016-07-27
Bugfix for "All Race Template"

Uveso:
The All Race Template can't handle FAF Factories and is not able to work propper with unit mods like the featured mod BlackOps.

Also there is a bug, where the Template icon is not updated, when a new icon is set for the Template and all units from the template where created from another faction.

There where also some optimizing issues.
For detailed info see:
https://github.com/aeoncleanse/Mods/pull/145


aeoncleanse:
e26710ba15c95235673248951007f355fd193469

PR #1402, opened 2016-07-27
Adds build-mode for SACU's build with presets

slinkingant:
Extracted the original buildmodedata.lua and added all unitnames for the
sacu-presets to allow build-mode to function as a regular sacu.

Added description for each sacu as mentioned as is visible in-game,
order is the same as in the quantum gateway.


aeoncleanse:
4d730070a5dac3c5f1bb6e691390d64525612ff3

PR #1404, opened 2016-07-28
Bugfix for option "share conditions"

Uveso:
This was posted in the forum:
http://forums.faforever.com/viewtopic.php?f=3&t=12813&p=131448#p131448

The share conditions are not working propper.

First of all, if played with an sorian AI, the AI will always share their units after dead to human players. Doesn't matter what option was chosen.
I created a new function to handle the sorian PlatonHandle without transfering all units.

The function "Share after dead" was also transfering units to the allies, but should only set the new ownership to the unit.
I used the script now for a new added option: "Transfer Units after dead".

And i created a new script for "Share after dead" with only transfering the ownership from already shared units.

Rounded up with Option and Localisations stuff.

GameOptions are now:
Share After Death (was former FullShare)
`You can give units to your allies and they will not be destroyed when you die`

Share Until Death (works like before)
`Kill all the units you shared to your allies and send back the units your allies shared with you when you die`

Transfer After Death (new)
`All your units will be transfered to your ally with the highest score if you die`


PlaYsiA:
good work!

Uveso:
This need some russian translation:

```
lobui_0741="Destroy or share units when you die"
lobui_0742="Share After Death"
lobui_0745="Destroy all the units you shared to your allies and send back the units your allies shared with you when you die"
lobui_0738="Transfer After Death"
lobui_0739="All your units will be transfered to your ally with the highest score when you die"
```

Exotic-Retard:
"Share After Death" and "Transfer After Death" are too similar names, and would get confused.

i would stick to the familiar "Full Share" so we know its unchanged and dont get confused.

speed2CZ:
Yes, please keep it "Full Share" everyone knows what it means and is used to that for years. We can't change that anymore

aeoncleanse:
I've got lots more feedback from the community here too. Basically, the new "Share After Death" option has been described as "Cancer" by many. It's too punishing, basically just kills you for not knowing the mode. It's "Transfer in the second before you die, or you just screwed your team".

So at a base level, everyone seems to want two modes:
- Full Share: Anything transferred during the game stays where it is when you die. Everything you control on death gets transferred to the player on your team with the highest score.
- Share Until Death: All units you built are killed when you die.

I also had requests for three fun new modes.
- Become Civilian: Units built by you are transferred to a Civilian, non-aggressive AI when you die.
- Become Enemy AI: Units built by you are transferred to a Civilian, aggressive AI when you die.
- Gain Control: Units controlled by you are transferred to the control of your killer. Units built by you and transferred to another player before death stay where they are. Perfect for FFA games with a twist.

If we could get those in, that'd be lovely. I'm curious to know if the first two new modes could possibly be made to work on maps which don't naturally have any Civilian AIs...

Uveso:
Looks like its really confusing, because its not
Share `After` Death & Transfer After Death
it's:
Share `Until` Death & Transfer After Death

So we better go back to full Share :)

(Is the meaning of "share until" and "transfer after" really so similar ? In my language its fundamental different)


Gain Control should be easy to implement. Maybe also the civilian options.
I will test it.

Uveso:
Option Full Share is back and Gain Control is added.

Gain Control:
All units will be transfered to the enemy with the highest score. if he reached his unitcap, all remaining units will be transfered to the (enemy)player with the 2nd highes score.
Borrowed unts will stay where they are and UnitOwnership will be corrected to the new Owner.

aeoncleanse:
I'm sorry my friend, but that's not _quite_ as requested by people. There was no desire for a simple "Give it to the enemy team". The suggestion was "Give to the player who got the kill on me", or put another way, "Kill someone to get their stuff". If that player is over the unit cap, remaining units need to be killed off.

Any luck with the trickier "Create Civilian" modes?

Uveso:
hehe thats what i am working on, check the civilian stuff and finding out who is killing who.

Uveso:
Gain Control now transfers the units to the enemy attacker.
Quite fun even in AI vs AI games where i tested this :)

Uveso:
Update. new ShareOption:
`"Civilian Desertion"`
`"After your dead all units will desert to civilians, given that civilian armies exist on the map, otherwise they will die."`

Dont know if i should name this like this. just a work title :)

aeoncleanse:
Can we get two versions, one where they become neutral, one where they become aggressive?

lua/aibrain.lua
```diff
@@ -531,56 +527,164 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                        end
+                    end
+                end
+            elseif shareOption == "TransferAfterDeath" then
                 -- this part determines who the allies are
                 for index, brain in ArmyBrains do
                     brain.index = index
                     brain.score = CalculateBrainScore(brain)
-                    if IsAlly(selfIndex, brain:GetArmyIndex()) and selfIndex ~= brain:GetArmyIndex() and not brain:IsDefeated() then
+                    if IsAlly(selfIndex, index) and selfIndex ~= brain:GetArmyIndex() and not brain:IsDefeated() and not ArmyIsCivilian(index) then
```
aeoncleanse: Why are we using index here and brain:getarmyindex everywhere else?

lua/aibrain.lua
```diff
@@ -531,56 +527,164 @@ AIBrain = Class(moho.aibrain_methods) {
...
+            local enemies = {}
+            local civilians = {}
             local selfIndex = self:GetArmyIndex()
             WaitSeconds(10)
             -- this part determiens the share condition
             local shareOption = ScenarioInfo.Options.Share or "no"
-            -- "no" means full share
-            if shareOption == "no" then
+            local victoryOption = ScenarioInfo.Options.Victory or "no"
+            if shareOption == "ShareAfterDeath" then
```
aeoncleanse: I think this was the option people want to go away.

lua/aibrain.lua
```diff
@@ -531,56 +527,164 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                -- transfer ownership of borrowed units
+                for index, brain in ArmyBrains do
+                    if IsAlly(selfIndex, brain:GetArmyIndex()) and selfIndex ~= brain:GetArmyIndex() and not brain:IsDefeated() then
+                        local units = brain:GetListOfUnits(categories.ALLUNITS - categories.WALL, false)
+                        if units and table.getn(units) > 0 then
+                            for _,unit in units do
+                                if unit.oldowner == selfIndex then
```
aeoncleanse: We now have multiple modes that want to have pretransferred units survive. Function this out?

lua/aibrain.lua
```diff
@@ -531,56 +527,164 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                        for _,unit in units do
+                            RemovePlatoonHandleFromUnit(unit)
+                        end
+                    end
+                    if victoryOption == 'demoralization' then
+                        KillerIndex = ArmyBrains[selfIndex].unitStats.LastKilled.COM
+                        TransferUnitsOwnership(units, KillerIndex)
+                    else
+                        KillerIndex = ArmyBrains[selfIndex].unitStats.LastKilled.unit
+                        TransferUnitsOwnership(units, KillerIndex)
```
aeoncleanse: So we transfer to killer, good...

lua/aibrain.lua
```diff
@@ -531,56 +527,164 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                -- this part determines who else is enemy
+                for index, brain in ArmyBrains do
+                    brain.index = index
+                    brain.score = CalculateBrainScore(brain)
+                    if IsEnemy(selfIndex, index) and index ~= KillerIndex and not brain:IsDefeated() and not ArmyIsCivilian(index) then
+                        table.insert(enemies, brain)
+                    end
+                end
+                -- This part determines which enemy has the highest score and transfers ownership of all units which are not already transfered to the killer
+                if table.getn(enemies) > 0 then
```
aeoncleanse: Then we transfer again? The way this mode will be used, I think players want leftover units to be killed. Besides, unit cap should IMO be transferred alongside the units, so the delta unitcap of the recipient is unchanged

lua/sim/Unit.lua
```diff
@@ -1233,7 +1233,10 @@ Unit = Class(moho.unit_methods) {
 
         --Notify instigator of kill
         if instigator and IsUnit(instigator) then
+            ArmyBrains[self:GetArmy()]:SetUnitStat('LastKilled', "unit", instigator:GetArmy())
```
aeoncleanse: Ouch. This is a super hot function. Ia there no other way to determine which brain killed another? Surely this information is already recorded somewhere in sim.

aeoncleanse:
Good point made in aeolus, these modes need to be unrated

aeoncleanse:
Need to be sure to consider draws and quits. Seems players would like transfer to killer AND transfer to enemy team as two modes

lua/aibrain.lua
Uveso: Its was a test. I was not completly sure that ArmyBrains array index is the same then ArmyIndex :)

lua/aibrain.lua
Uveso: Hard to say, because originaly the `Share after dead` function was transfering units after dead to other players. Now `Share after dead` is working like the old tooltip said. Only set the new ownersip of the unit and don't transfer new units.

lua/aibrain.lua
Uveso: This way i have an better overview while programming.
When everything is working like it should i will optimize this.

lua/aibrain.lua
Uveso: Good point, i will kill the leftover units.

lua/sim/Unit.lua
Uveso: well its not so hot. Thinks like this will be executed every time the COM gets damage.
```
    DoTakeDamage = function(self, instigator, amount, vector, damageType)
        WalkingLandUnit.DoTakeDamage(self, instigator, amount, vector, damageType)
        local aiBrain = self:GetAIBrain()
        if aiBrain then
        end

        if self:GetHealth() < ArmyBrains[self:GetArmy()]:GetUnitStat(self:GetUnitId(), "lowest_health") then
            ArmyBrains[self:GetArmy()]:SetUnitStat(self:GetUnitId(), "lowest_health", self:GetHealth())
        end
    end,
```
Lowest health is pumped every damage tick to the array. I guess send this once if a unit dies is ok.
And i dont find another place where this is saved.

Uveso:
`Q: Can we get two versions, one where they become neutral, one where they become aggressive?`
It has already a own Option:
Civilians = neutral, enemy or none

`Q:Seems players would like transfer to killer AND transfer to enemy team as two modes`
Will add option `"Transfer to Enemy Team"`

lua/aibrain.lua
Uveso: we should not change the unitcap here because unitcap share has his own option and we would break this here.

aeoncleanse:
Why closed?

Uveso:
good question, maybe i hit the wrong button :)

Well new update:
Changed the `brain:getarmyindex`
made Subfunctions
transfer to killer don't send leftover units to allies now.
New option `Defectors`
`All your units will be transfered to your enemy with the highest score when you die.`

Btw, if 2 playes get a draw, all units will be destroyed.
If no share condition meets, the rest of all units will be always destroyed.

lua/sim/Unit.lua
Uveso: btw, this could also be used for archivements like who killed the most players/COMs in game etc. :)

PR #1405, opened 2016-07-28
Prevent dropped units from accelerating in air

Crotalus:
By queuing up move / attack orders after a load transport order you could get units to accelerate in air, jumping towards their next target. Common use is ACU sniping with fire beetles.


Zockyzock:
But you can still queue up the orders? Then this is a great fix. :) 

Would like to put it into current balance beta if you confirm.

Crotalus:
Yes order queue is untouched, the units are flagged immobile from start of drop until they hit the ground.

aeoncleanse:
0543861407c36c24013fafcd7a842b8830cbef24

To be tested in beta

PR #1409, opened 2016-07-30
Fix Cinematics in Coop

speed2CZ:
`GetFocusArmy()` on line 123 https://github.com/FAForever/fa/pull/1409/files#diff-68df115e91dfbe20a9fa220e6fb8b635R123 needs to be replaced
@Crotalus can you help me with this one?
I've tried to use the same trick as you https://github.com/FAForever/fa/commit/f684f89e4f7860a208c6f9fe8436d7fee379d111 and it made the replays work, but camera is not following the units.

### DIFF
--- a/lua/cinematics.lua
+++ b/lua/cinematics.lua
@@ -121,9 +121,11 @@ end
 -- Zoom is measured in LOD units. It's the value of the width of the view frustum at the focus point
 function CameraTrackEntities(units, zoom, seconds)
     local army = GetFocusArmy()
-    for i, v in units do
-        if army ~= v:GetArmy() then
-            units[i] = v:GetBlip(army)
+    if army ~= -1 then
+        for i, v in units do
+            if army ~= v:GetArmy() then
+                units[i] = v:GetBlip(army)
+            end
         end
     end
 
--- a/lua/cinematics.lua
+++ b/lua/cinematics.lua
@@ -1,0 +1,205 @@
+#****************************************************************************
+#**
+#**  File     :  /lua/cinematics.lua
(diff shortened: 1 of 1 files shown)


Crotalus:
Try and see if you can get away doing like this:

```
function CameraTrackEntities(units, zoom, seconds)
local army = GetFocusArmy()
    if army ~= -1 then
        for i, v in units do
            if army ~= v:GetArmy() then
                units[i] = v:GetBlip(army)
            end
        end
    end
    ...
```

i.e. put a check that FocusArmy isn't -1 before doing that blip filtering, GetBlip grabs the radar blip of a unit from a certain armybrain's perspective. I guess it's there to be able to camera track blips

speed2CZ:
Ok this does work, I just thought about this: "Never use GetFocusArmy() with anything related to sim state" isn't this the case as well?

Crotalus:
Yeah that's the basic rule but in this case it seems like the functions doesn't affect the actual sim state, only touching Camera stuff. 

If we start getting desync in 2+ player games or replays after with this fix though we can come up with another solution.

aeoncleanse:
I'd like some solid testing of multiplayer for desyncs please. Is it possible to test this as a mod, or will your require beta server testing?

speed2CZ:
I think ive tested it with two pc via LAN (dont remember now) so I'll do one more test tonight to be sure

speed2CZ:
@aeoncleanse tested, seems to be working fine :)

aeoncleanse:
d1a613bdaca374baa35851c0d10cf15da66120ef

PR #1410, opened 2016-07-30
Eqaeontmd

JaggedAppliance:
This is for aeon tmd, other commits aren't relevant as you can see but i'm not sure how to clean this up properly. This adds additional collision spheres so aeon tmd can catch projectiles that are high up when before those projectiles would just fly over unmolested.


aeoncleanse:
Merged

PR #1420, opened 2016-08-06
Optimizing Blueprints Loading 

HUSSAR-mtrela:
- [x] optimize table.hash function (+30% performance)
- [x] optimize bluprints.lua to save bp.CategoriesHash instead of discarding it (+32% performance)
- [x] optimize sim lua files to use bp.CategoriesHash instead of `for loops` to look up bp.Categories
- [x] optimize units analyzer to use bp.CategoriesHash instead of looking up bp.Categories
- [ ] optimize lobby.lua to pre-load bps in smaller chunks instead of one big task
  
  **Note this PR is in progress and it will be ready when all above task are complete**

### DIFF
--- a/lua/system/utils.lua
+++ b/lua/system/utils.lua
@@ -237,24 +237,35 @@ function sort_down_by(field)
     end
 end
 
---- table.keys(t, [comp]) -- Return a list of the keys of t, sorted.
+--- table.keys(t, [comp], [skip]) returns a list of the keys of t, sorted.
 --- [comp] is an optional comparison function, defaulting to less-than.
-function table.keys(t, comp)
+--- [skip] is an optional boolean that specifies if sorting is needed, defaulting to true
+function table.keys(t, comp, skip)
     local r = {}
     if not t then return r end -- prevents looping over nil table
+    local n = 1
     for k,v in t do
-        table.insert(r,k)
+        r[n] = k -- faster than table.insert(r,k)
+        n = n + 1
     end
-    table.sort(r, comp)
+    if not skip then table.sort(r, comp) end
     return r
+end
+
+--- table.keysfast(t) returns a list of the keys of t, in unspecified order.
+--- this function is about 3 times faster than table.keys(t, [comp], [skip])
+function table.keysfast(t)
+    return table.keys(t, nil, true)
 end
 
 --- table.values(t) Return a list of the values of t, in unspecified order.
 function table.values(t)
     local r = {}
     if not t then return r end -- prevents looping over nil table
+    local n = 1
     for k,v in t do
-        table.insert(r,v)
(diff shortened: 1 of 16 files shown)


aeoncleanse:
@ChrisKitching - Just inform you this is happening, probably best you run an eye over it at some point.

HUSSAR-mtrela:
@aeoncleanse I think  @ChrisKitching should review this  PR when it is complete. So far I have minor improvements and UI thread blocking should be resolved when I complete the last task.

lua/system/utils.lua
```diff
@@ -316,43 +327,76 @@ end
 --- table.reverse {'one','two','three'} => {'three', 'two', 'one'}
 function table.reverse(t)
     if not t then return {} end -- prevents looping over nil table
-    local reversed = {}
+    local r = {}
```
Sheeo: I don't see why substituting `reversed` for `r` was required here? This is certainly less descriptive.

Not that I generally mind succinct algorithms, but just shortening variable names for the sake of it seems redundant.

lua/system/utils.lua
```diff
@@ -339,11 +339,14 @@ function table.indexize(t)
...
 
 --- Converts a table to a new table with values as keys and values equal to true
+--- it is useful for quickly looking up values in tables instead of looping over it 
+--- table.hash { [1] = 'A',  [2] = 'B',  [3] = 'C' } => 
+--             { [A] = true, [B] = true, [C] = true } 
 function table.hash(t)
     if not t then return {} end -- prevents looping over nil table
     local lookup = {}
     for k, v in t do
-        lookup[tostring(v)] = true
```
Sheeo: This also means that you'll get peculiar errors when you're trying to hashify a table containing values of type `number | string`. Consider this:

```lua
local t = {[1] = 'A', [2] = 1, [3] = 3}
table.hash(t)
=> { 'A' => true, [3] = true }
```

Your 30% performance figure isn't backed by enough data (How many iterations did you run?).

lua/system/utils.lua
HUSSAR-mtrela: I found it confusing or difficult to understand when names of variables have similar names of functions. Also, these change makes names of variables consistent with rests of variables in this file.

lua/system/utils.lua
HUSSAR-mtrela: @Sheeo you brought valid point here and I will see if there is other way of creating a key for hash table that would not require using tostring() function.

I calculated my performance figures with 10 iterations on tables with 15000+ items. Also @Uveso notice that table.hash(t) was noticeably slower when it was using tostring() function.

HUSSAR-mtrela:
Closing this PR because it is obsoleted by PR#1766

PR #1423, opened 2016-08-09
Try to not create needless classes

aeoncleanse:
I've tried to condense this code down so that we don't need 3 classes. A middle-road would be to inherit the new classes properly and only override oncreate, but this is surely the most concise?

However, this code is NOT WORKING. I don't get it, but it only creates the one sphere. The offset clearly isn't propagating correctly. @Crotalus @duk3luk3 @Uveso Any insights?


Uveso:
Maybe a dump idea, but we could give the launcher 2 additional weapons with different projectile speed. So they will spread while flying.

aeoncleanse:
It's been tried. The idea of this is to catch missiles at the very uppermost range of the dome. Such missiles, they clip the top, even an accelerated projectile wouldn't reach them in time. I'm more interested in why this change doesn't work. It _should_ be a nonfunctional code cleaning :/

Uveso:
Well, i recreated this inside the develop version, and it works. 
3spheres

PR #1435, opened 2016-08-19
Automatically create Unit Display Abilities

Uveso:
This adds the ability to autocreate Unit-Display-Abilities to construction.lua
Also added Ingame Options:

`Auto create Display Abilities`
`Off`: Use original abilities from Unitblueprint
`Medium`: Use original abilities and auto-create if no abilities where found
`Full`: Always auto-create abilities.
`Advanced`: Merge auto-create and unitblueprint abilities.

(Advanced will also include and display unknown Abilities from mods like Black Ops etc.)

Dependencies:
This PR needs CategoryHash: #1420


aeoncleanse:
@ChrisKitching This looks like another one that could use your practised eye when you get some time.

Uveso:
New abilities are now also shown inside the small ability-only window:

abilitywindow

Issue #1294, opened 2016-05-20
AI sometimes doesn't die


aeoncleanse:
http://forums.faforever.com/viewtopic.php?f=3&t=12470

Uveso:
Verified.
I wrote a simplified explanation in the forum.
A not ending game is mostly a aftereffect from a "Game object has been destroyed" Error, if you are using the "Assassination"  victory setting.

If this occours you can see a dead unit on the Map that don't has a wreck.
Console Commands like "Destroyall" or "Killall" dont work here to delete the last unit.

Uveso:
I'ts merged into develop. Do we close this then ?

Crotalus:
Yeah, good work!

PR #1438, opened 2016-08-20
Bugfix Game don't end in Assassination or Supremacy mode

Uveso:
Issue #1294
Replaced the function GetCurrentUnits() with a new one. Now units are
counted accurate and the game will end in all gamemodes like expected.

### DIFF
--- a/lua/victory.lua
+++ b/lua/victory.lua
@@ -7,6 +7,17 @@ local victoryCategories = {
     eradication=categories.ALLUNITS - categories.WALL,
 }
 
+function CountCurrentUnits(brain,categoryCheck)
+    local ListOfUnits = brain:GetListOfUnits(categoryCheck, false)
+    local count = 0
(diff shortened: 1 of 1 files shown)


lua/victory.lua
```diff
@@ -7,6 +7,17 @@ local victoryCategories = {
     eradication=categories.ALLUNITS - categories.WALL,
 }
 
+function CountCurrentUnits(brain,categoryCheck)
+    local ListOfUnits = brain:GetListOfUnits(categoryCheck, false)
+    local count = 0
+    for index,unit in ListOfUnits do
+        if unit.CanBeKilled == true and unit.Dead == false then
```
HUSSAR-mtrela: i suggest using this more readable if statement:
if unit.CanBeKilled and not unit.Dead then

aeoncleanse: Absolutely

PR #1440, opened 2016-08-22
Fix builder hook message (warning)

Uveso:
After hooking builders or platoons the game will post a warning about overwriting.
This looks like an error but is a normal hook and should only post a LOG message.

Example before:
`
WARNING: Duplicate PlatoonTemplate detected - overriding/appending old: \000ACUUpgrades - Gun improvements
`

and after the patch:
`
INFO: Hooked PlatoonTemplate: \000ACUUpgrades - Gun improvements
`


aeoncleanse:
9077aa208fde5a487052c852b7e1862679d3da31

Issue #1434, opened 2016-08-18
Wrong Tooltip shown for T2 land support factory


Uveso:
from:
http://forums.faforever.com/viewtopic.php?f=2&t=12944

UEF, Cybran and Aeon TECH1 and TECH2 support factories are showing the tooltip from Naval factory

PR #1447, opened 2016-08-27
BugFix Tooltip T2/T3 support factory

Uveso:
This will fix the issue #1434

Unitdescriptions from supportfactories had wrong indexes in unitdescription.lua


aeoncleanse:
13ae8c4b91a7aab705274cafee6c16dde6436e5c

Issue #1418, opened 2016-08-02
Sometime GameOptions.Quality is saved as -1#INF in Game.Prefs which make it corrupt


JaggedAppliance:
When it loaded up it was in the minimum res, 1024. It was fullscreen even though i use windowed mode. It didn't have any of my ui mods active and I couldn't even use the ctrl+shift command to show reclaim.

[logs.zip](https://github.com/FAForever/fa/files/397944/logs.zip)

JaggedAppliance:
I hosted a lobby and it had the "what's new in lobby" window. The window said it was patch 3636.

Crotalus:
Seems like your Game.prefs got corrupted somehow and renamed to `Seems like your Game.prefs got corrupted somehow and renamed to `C:\Users\Admin\AppData\Local\Gas Powered Games\Supreme Commander Forged Alliance\Game.prefs.bad`. Could be interesting to see exactly what caused it, line 547 in that file, something with BuildMult.

```
Warning: Loading "C:\Users\Admin\AppData\Local\Gas Powered Games\Supreme Commander Forged Alliance\Game.prefs" failed: ...ata\Local\Gas Powered Games\Supreme Commander Forged Alliance\Game.prefs(561): `}' expected (to close `{' at line 547) near `BuildMult'
warning: Unable to load your preferences (corrupt or out-of-date file.)  Current file will be renamed: [C:\Users\Admin\AppData\Local\Gas Powered Games\Supreme Commander Forged Alliance\Game.prefs.bad]
warning: Can't open lua file "C:\ProgramData\FAForever\bin\..\Installed.Prefs"
warning: unable to load Installed.Prefs; using empty initial prefs
```

JaggedAppliance:
[logs.zip](https://github.com/FAForever/fa/files/397968/logs.zip)
Here's the 2 gameprefs

Crotalus:
Yikes, GameOptions table at line 547 contains among other things:
```
...
FogOfWar = 'explored',
Quality = -1.#IND,
BuildMult = '2.0',
...
```

Quality is undefined and that syntax is not correct lua `inf` or `nil` would have been better at least.

If you want your old settings just nuke that Quality line completely (line 560).

Lets keep this issue open to investigate why Quality was even saved in this state.

JaggedAppliance:
Great, thanks for the help crotalus.

speed2CZ:
I'll close this, another investigation can be done in #1468 as icedreamer asks

PR #1449, opened 2016-08-28
Bug fix GameOptions.Quality is saved as -1#INF

Uveso:
Issue:
Sometime GameOptions.Quality is saved as -1#INF in Game.Prefs which make it corrupt #1418

gameOptions.Quality is set inside Lobby.lua:
lobby.lua line 1263
```LUA
gameInfo.GameOptions['Quality'] = data.quality
```
lobby.lua line 1936
```LUA
gameInfo.GameOptions.Quality = quality
```
The quality value is calculated inside the function computeQuality
```LUA
local quality = Trueskill.computeQuality(teams)
```
So we are getting the wrong value "-1.#IND" from computeQuality inside trueskill.lua.

To fix this i changed the last line from function computeQuality(team) to make sure we are returning a number.
changed from:
```LUA
    return round((result * 100), 2)
```
to:
```LUA
    return tonumber(round((result * 100), 2))
```


aeoncleanse:
How would the old version ever have not been a number? What is result, a number as a string?

Uveso:
Result should always be a number.
But "-1#INF" (negative infinity) is a error return from language c/c++. Because LUA dont know this expression it is converted to a string.

aeoncleanse:
I understand the error, but where is the C error coming from?

```return round((result * 100), 2)``` 
Surely this can only return a non-number value if 'result' is going in as a non-number. Why is that ever the case?

**Comment by micheljung** [comment]
It could also happen if `result` is a float that overflows when multiplied by 100, which can be avoided by round by `round(result, 4) * 100`

**Comment by Uveso** [comment]
Oh yes, rounding before multiplying is a good idea.

I don't know why this error happens. Maybe if a player leave the lobby while this function is executed ?
I tryed to debug this with some random numbers, but it never failed with negative infinity.

**Comment by micheljung** [comment]
Maybe you could call this function in a loop with random numbers between -200 to 2000 until the `-1#INF` happens, then we can figure out why.

Another guess I had is that a division by zero happens, but:
```
> print(1/0)
1.#INF
```
Still, it might be a hint.

**Comment by Uveso** [comment]
Because i dont know how the function really works i used already a bruteforce methode. Feeding the function with random numbers. But i got never a negativ infinity. Thats the problem :)
Maybe just the debugline was printing a wrong expression, and i got those errors who knows :)

My guess is, maybe a global variable is changed from outside the function and crashed it then. Nothing we can easily reproduce, in case i was a leaving player etc.

Thats why the tonumber thing was my best choice. More a failsave then a fix :)

**Comment by aeoncleanse** [comment]
And we're certain feeding a corrupt C value into tonumber will result in a... number... ?

**Comment by duk3luk3** [comment]
Yes, this is almost certainly due to a division by zero, and calling tonumber() won't help at all, because negative infinity (which becomes -1#INF when converted to string / printed) is a completely valid number as far as Lua is concerned.

Use these functions to check for infinity / nan:

```
local function isINF(value)
  return value == math.huge or value == -math.huge
end

local function isNAN(value)
  return value ~= value
end
```

Better yet, check the math to make sure you never divide by zero.

Uveso:
if you try to do `x = 1/0` you will get in lua: `1.#INF`
`X = 0/0` will produce `-1.#IND`

So both expressions are lua-numbers.
But we want to filter `-1#INF`.
Does tonumber()  recognize this as number, even if this is a c/c++ expression ?

duk3luk3:
Like I said lua uses standard IEEE754 doubles, which have +/- Infinity, NaN, and subnormal numbers.

We don't have to worry about subnormal numbers (nobody does, anyway) but yes, tonumber() will just do nothing at all to something that is already a number.

Also note that 1.#INF / 1.#IND is just the result of using printf(), Lua itself has no constants for infinity / nan except math.huge (which is only guaranteed to be "the largest number", but works out to being equal to +infinity when using ieee754 doubles). Which explains why it blows up when trying to parse the game prefs.

Uveso:
math.huge is not working inside trueskill.lua.

I also want to changed back the order with multiply and rounding.
it is generating different results:
`round(0.123456789, 4) * 100 == 14.285715103149`
and if we use the old order:
`round((0.123456789 * 100), 2) == 12.375`

To make sure that Gamequality is between 0 and 100.
What if i ask for `quality > 0 and quality < 100,` will this filter `-1#INF. `out ?

duk3luk3:
> math.huge is not working inside trueskill.lua.

It's quite possible it doesn't exist in GPG's lua version, but it's identical to the result of ```1/0``` anyway.

>  What if i ask for ```quality > 0 and quality < 100```, will this filter ```-1#INF.``` out ?

yes, if quality should always be in range ```0..100```, this is what I would do:

```if not quality >= 0 and quality < 100 then quality = 0 end```

Or whatever default value other than ```0``` is best.

Uveso:
Thank's Lukas, i guess i got it now right :)

asdfghjklqwertyu:
I don't want to waste peoples time with redundant/known information, but while you seem to have found a solution to fix the issues with the function in the truskill lua, I don't see anything found here or elsewhere for the cause.

This seems to be an issue related to the Phantom-X mod.

In Phantom matches everyone starts allied, but can break at any time and "teams" are not known until assignment.

If you watch a replay for any phantom match, it will show "Q:-1.#J%" as the quality rating during the match. I have checked with others and everyone sees the same during matches, and while watching Phantom-X replays and I and others saw normal quality ratings during test FFAs, Xtreme Wars and other modded matches. Only Phantom matches seem to produce the corrupting value, and it seems to take as little as playing a single Phantom match to re-corrupt the prefs file.

Perhaps just disabling quality ratings for Phantom matches is the easy solution?

Happy to help in any other way if I can.

duk3luk3:
```if quality <= 0 or quality > 100 then return 0 end```

This is bad because it won't catch Nan. There was a reason I wrote the logic the other way around ;)

PR #1450, opened 2016-08-28
Bugfix Enhancement Tooltips sometimes not shown.

Uveso:
Make sure, enhancement icons are enabled, even if they are greyed out.
We need this to have an active button handler to show the tooltip.

### DIFF
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -779,7 +779,7 @@ function CommonLogic()
             control.LowFuel:SetNeedsFrameUpdate(false)
             control.BuildKey = nil
             if control.Data.Disabled then
-                --control:Disable()
+                control:Enable()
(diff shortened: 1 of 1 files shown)


aeoncleanse:
Could you give me instructions on how to test this please, and in what situation this is necessary. A before and after if you will :) Thanks.

aeoncleanse:
OK, this was in response to a bug I actually found. On low resolutions, inside the ACU enhancement menu, if one uses the arrow buttons to scroll across them, the tool-tip on mouse-over stops working. It only happens if you use the increment scroll arrows, not if you use the "Jump to end" one.

I think a better solution to this issue will be finding out why the arrow keys disable the tooltip in the first place, when the jump button doesn't.

aeoncleanse:
6664e7ebb8566810f5197a455c734f500cc60fbe

Issue #1403, opened 2016-07-27
Swapping player in the lobby also swaps their designated team with them, thus overriding the auto team setting


Justify87:
It doesn't matter how you swap them (via map preview or "move player to slot x"), the team will always move with the player to the new slot. Could be a result of the change in the player swapping function: #1376

aeoncleanse:
Please check non-beta and see if move player to slot does this. I suspect it will.

Justify87:
Just tested it. Works in non-beta.

aeoncleanse:
Works meaning it's broken or not broken?

Justify87:
It is not broken in non-beta

Justify87:
In the old swapping function, temporary moving the player ot the observer list seems to have reset the players team.

aeoncleanse:
Oh. Gonna have to deal with that then.

PR #1451, opened 2016-08-29
Fix/team number moving with player (#1403)

Justify87:
Swapping players also moves their team with them. This fixes the issue.


aeoncleanse:
282da9545915c4102179b6312fe22481d1106244

PR #1464, opened 2016-09-02
Rebalance aeon t1 maa

JaggedAppliance:
Allow aeon t1maa to be one shot by non-aeon bombers, and give it some range as compensation.
Note: this is for beta


aeoncleanse:
19cf89034f16039a490be1591c7981cfdaa1add6

PR #1473, opened 2016-09-07
Feature/unit shield collision

Exotic-Retard:
well, now air units collide with shields, with the help of a companion projectile. that bits easy, it just follows the aircraft and thats it. on impact, the shields absorb some crash damage and then the air unit bounces off the shield. this is where it gets interesting due to how the air units and projectiles interact with the ground, but anyway it works. as far as i can tell.


lua/defaultunits.lua
```diff
@@ -1718,7 +1718,12 @@ AirUnit = Class(MobileUnit) {
 
         for i, numWeapons in bp.Weapon do
             if(bp.Weapon[i].Label == 'DeathImpact') then
-                DamageArea(self, self:GetPosition(), bp.Weapon[i].DamageRadius, bp.Weapon[i].Damage, bp.Weapon[i].DamageType, bp.Weapon[i].DamageFriendly)
+                local Damage = (bp.Weapon[i].Damage - self.ShieldDamageAbsorbed)
+                --WARN(Damage .. ' crash damage dealt to ground') -- since our shield collision subtracts damage, this may be nice to know how much is left
+                if Damage > 0 then
+                    DamageArea(self, self:GetPosition(), bp.Weapon[i].DamageRadius, (bp.Weapon[i].Damage - self.ShieldDamageAbsorbed), bp.Weapon[i].DamageType, bp.Weapon[i].DamageFriendly)
```
Sheeo: You already computed `(bp.Weapon[i].Damage - self.ShieldDamageAbsorbed)` right above and saved it into `Damage`. You should reuse the value now instead of computing it again :)

lua/defaultunits.lua
```diff
@@ -1754,6 +1759,34 @@ AirUnit = Class(MobileUnit) {
...
             self:DisableShield()
+            
+            self.ShieldDamageAbsorbed = 0 --we use this to work out how much damage to subtract from the deathimpact.
+            
+            local this = self --waiting for our projectile to collide with a shield or the ground, and then modifying deathweapon
+            self:EnableShieldCollision(
+                function()
+                    local bp = self:GetBlueprint()
+                    local i = 1
+                    local numWeapons = table.getn(bp.Weapon)
```
Sheeo: `numWeapons` appears to be unused after this point, except for being overwritten in the loop by the weapon value?

Exotic-Retard:
ok i fixed the damage part, and i went through and changed local variables to lowercase first (at least some of them xdddd)

numWeapons is used simply to get the number of weapons in the bp so we can cycle through them. this is because the unit is dead (or something) and so GetWeaponByLabel doesnt work, so we look at the bp instead and find the first one that contains the label we want, then use that as the crash damage, and stuff

so yeah i guess its not used after that line, but im not super sure what you mean/want me to do about it

Sheeo:
> numWeapons is used simply to get the number of weapons in the bp so we can cycle through them.

No, it isn't. It's simply read and then discarded. The `for` loop:

```lua
for i, numWeapons in bp.Weapon do ...
```

Sets `numWeapons` to be `bp.Weapons[i]` for each iteration.

Exotic-Retard:
hum, i just thought that numWeapons (a value in the unit blueprint) is the total number of weapons in the blueprint, and so we cycle from 1 to numWeapons (normally 2 or w.e). i.e. going through each weapon in the blueprint :| and bp.Weapons is the table containing all the weapons in the blueprint... or what? in any case thats what gpg did so i just copied them there xD

Sheeo:
If the loop had been like:

```lua
for i=1, numWeapons do
    .... bp.Weapons[i]
end
```

That would've been the case. But this is a [generic for](https://www.lua.org/pil/4.3.5.html) that iterates the contents of the table using the `ipairs()` function (which gpg has hidden from you), so instead you provide two references to be used for the iteration key and the iteration value.

Exotic-Retard:
i understood barely anything, but what should i do about it? if anything?

aeoncleanse:
It means you don't need to declare numweapons above, as it's created and overwritten as a local in the for loop.

Exotic-Retard:
¯\_(ツ)_/¯ well, whatever, it works i guess xD

anything else?

JaggedAppliance:
I ctrlk'd some strats, they bounced off the shield as they should then got stuck on the actual shield gen for a long time before getting destroyed.

aeoncleanse:
Supplanted by #1581

PR #1487, opened 2016-09-13
Kill timer thread if called from manual result

speed2CZ:
If you ended this objective by calling manual result when the timer was
still running, it didn't stop the timer.


aeoncleanse:
Where is this needed? Coop? How does one test it?

speed2CZ:
Yeah coop, do you know anything else that would be using objectives? :P
It can be tested in my new mission, I'll handle it, ot find some slave.

aeoncleanse:
K, let me know when this has been verified working

speed2CZ:
@aeoncleanse seems to be working

aeoncleanse:
7616c73c1e917dfe6fd9a4368a9854eb826d7b0c

PR #1490, opened 2016-09-13
HotFix No-Rush option unavailable for german players

Uveso:
With the german language file, shiped with the original game, the option `No Rush` is not displayed properly.
This is how it looks with english language settings:
rushe
And this is how it looks in german:
rushd
This is caused by reusing of the location tag `lobui_0320` `("<LOC lobui_0320>%s",)`

This tag is defined inside the german language file with: 
`lobui_0320="5"`

To fix this i added a complete updated german localization file.

Also all FAF related strings are translated if they had a LOC tag.


aeoncleanse:
125c2e851fb970ae84bfea95198de0a8ded0df30

PR #1496, opened 2016-09-14
Fix localisationsstring tooltipui0694 for "Rehost"

Uveso:
Lobby Preset button shows wrong tooltip Headline "Rehost":

rehost2

tooltipui0694 is already occupied with `tooltipui0694 = "Lobby Presets".`
Moved rehost now to `tooltipui0696` and preset works again:

rehost1


aeoncleanse:
74c9b8a09a23c58a278eead43b3cd5df55251f83

PR #1499, opened 2016-09-15
Fix Option "Cheat/Build multiplier"

Uveso:
Cheat and build multiplier showing 1.0-1.9 two times, and showing "1" as default (should be 2):

cheatmult


aeoncleanse:
d2c710672198e7141728bf557856bab11e2cc1d8

PR #1500, opened 2016-09-15
Fix wrong loc tag for Random Map button

Uveso:
Fixed localisation for Random Map button.


aeoncleanse:
572efc1d26a59d06aadc1ed069211c27b40585ac

PR #1503, opened 2016-09-16
Fix translation Mod manager

Uveso:
Added LOC() function calls for Button text inside modmanager window.
Added <LOC> tags to button text.
Added german translation.


lua/ui/lobby/ModsManager.lua
```diff
@@ -292,7 +292,7 @@ function CreateModsFilter(parent, tag)
             states.active,
             states.highlight,
             states.disabled,
-            tag.name,
+            LOC(tag.name),
```
aeoncleanse: Why is this LOC different in format than the others?

Uveso: well if you mean why i don't put the LOC call directly inside the modsTags array ?
I have no idea :) Fix is on the way

aeoncleanse:
c7f71ecad662b94ff12b70cc88fc08ee64a64d2e

PR #1506, opened 2016-09-16
Fix missing tooltip text for "Cycle Preview for Hotbuild" and translation

Uveso:
Added options_hotbuild_cycle_preview to tooltips.lua
Added english tooltiptext and german localisation

Btw, how does this work? 
As i understand this while reading the code, you can press the same hotbuildkey and change something. But what are you changing?
Different units / buildings under the same key ?


speed2CZ:
Example, on W are binded factories. If you press it once it select landd fqctory, twice air factory, etc. The preview has all available factories showing, just icons and highlights the one that is currently selected.

aeoncleanse:
287276db8ef4db691da8f54ba477e77583666695

PR #1507, opened 2016-09-16
Automatically create Unit Display Abilities II

Uveso:
rebased PR #1435 

Ingame Options:

Auto create Display Abilities
Off: Use original abilities from Unitblueprint
Medium: Use original abilities and auto-create if no abilities where found
Full: Always auto-create abilities.
Advanced: Merge auto-create and unitblueprint abilities.

This PR has no dependencies and can be tested/merged as standalone


aeoncleanse:
I'm still unsure about why we need this one. It's better almost all of the time to hard-code things like this, so we do as little as possible in Blueprint loading. The primary reason from my point of view is that every thing you have adjusted in the Blueprint loading phase is a thing that's different in-game compared with the hard code which a modder or developer might be looking at. I believe @ChrisKitching  had other reasons too.

So in short, how do these modes differ? Which abilities are missing?

Uveso:
Well, this PR is not changing the blueprint anymore. Only the first version did.
Now the abillity array is created in the time between moving the muse over an icon and displaying the Tooltip.

If we use this, its possible to delete the display ability array from every blueprint. its optional, you can leave it or delete it, doesent matter for the function itself.
But deleting the abilities from blueprint will make them faster on loading. (that was the Plan)

The benefit is, if someone change someting inside the bluprint like removing radar from bombers, we dont need to adjust the display abilities. (We had a PR only for the bomber display ability because balace team changed the radar but not the ability array.)

There are also many abilities missing in modded units, and this way we get those abilities.

Also if you have an modded unit without a unitdescription and without an abilitiy array the Tooltip will not be shown, and advanced unit stats and weaponstats are not displayed.

Also its a kind of debugging. For example; if you see an cybran torpedolauncher with the "cloak" ability, you know there is something wrong inside the unitblueprint (like it is actual)

(btw, i killed the PR while trying to rebase it. will make a new one shortly....)

lua/ui/game/unitviewDetail.lua
```diff
@@ -8,6 +8,7 @@ local ItemList = import('/lua/maui/itemlist.lua').ItemList
 local Prefs = import('/lua/user/prefs.lua')
 local options = Prefs.GetFromCurrentProfile('options')
 local UnitDescriptions = import('/lua/ui/help/unitdescription.lua').Description
+local LocalisationUS = {} doscript('/loc/' .. 'us' .. '/strings_db.lua', LocalisationUS)
```
aeoncleanse: What does this do, and why on earth do we need it?

lua/ui/game/unitviewDetail.lua
```diff
@@ -450,3 +459,348 @@ function SetupUnitViewLayout(parent)
...
+        return true
+    end
+    return false
+end
+function CreateDisplayAbilities(bp)
+    -- Without this we have to reload the game to get the new option if the user changed something.
+    options = Prefs.GetFromCurrentProfile('options')
+    
+    -- Check if we have CategoriesHash
+    if not bp.CategoriesHash then
```
aeoncleanse: So this process will only happen once per unit type on mouseover right?

lua/ui/game/unitviewDetail.lua
```diff
@@ -450,3 +459,348 @@ function SetupUnitViewLayout(parent)
...
+        return bp.CategoriesHash.OVERLAYSONAR and bp.Intel.SonarRadius > 0
+    end,
+    ["ability_omni"] = function (bp)
+        return bp.CategoriesHash.OVERLAYOMNI and bp.Intel.OmniRadius > 0
+    end,
+    ["ability_hover"] = function (bp)
+        return bp.CategoriesHash.MOBILE and bp.Physics.MotionType == 'RULEUMT_Hover'
+    end,
+    ["ability_amphibious"] = function (bp)
+        local BitArray = DezimalToBinary(bp.Physics.BuildOnLayerCaps)
```
aeoncleanse: Dezimal? You what?

lua/ui/game/unitviewDetail.lua
Uveso: This is used here:
`newAbilities[count] = '<LOC '..ability..'>'..LocalisationUS[ability]`

We need the english original text here as fallback text in case we are using a language that don't has a translation for this.

lua/ui/game/unitviewDetail.lua
Uveso: Yes this will only happen once. And its only temporary until PR #1420 is merged.
I added this, so we can test this without the PR from Hussar.

lua/ui/game/unitviewDetail.lua
Uveso: Oh yes this is a nasty one.

BuildOnLayerCaps from the blueprint
 ```
       BuildOnLayerCaps = {
            LAYER_Air = false,
            LAYER_Land = true,
            LAYER_Orbit = false,
            LAYER_Seabed = false,
            LAYER_Sub = false,
            LAYER_Water = false,
        },
```
are stored inside `bp.Physics.BuildOnLayerCaps` as a single integer number like `548691` (decimal). oh lol Dezimal is german :)

We need to decode this into a binary and check the single bits for the layers.

PR #1508, opened 2016-09-16
Fix Tooltip "hotbuild cycle time"

Uveso:
Added Tooltip for "hotbuild cycle time" and translation strings
Added german translation


aeoncleanse:
c265998d1fc0daf775cc0aeb04547fccdfa087e7

Issue #1481, opened 2016-09-10
Option "Share Unit Cap at Death" not working correctly


Uveso:
Share unitcap is taking neutral armies into account while splitting the unitcap after player/Ai dead.

Played a map with 5 player and 2 neutral armies with a unitcap of 500 for all players.

After one player died, 500 units should be evenly distributed to the 4 players.
Makes 625units for the remaining 4 players.

But the game only give 583 units.
83 = 500 units are divided through 6.
The 4 players and the 2 civilian armies.

Uveso:
I planed to fix this next week, so you can assign this to me.

Uveso:
PR ist up

aeoncleanse:
We'll keep this open until merged in.

aeoncleanse:
5e91a56402deb9f00f823b729b50314dffe80763

PR #1510, opened 2016-09-17
Bugfix "Share Unit Cap at Death"

Uveso:
issue #1481

Fixed this by removing civilian armies from alive list.


Uveso:
Played with this last Weekend.
2 Player and several AI's. ~16 hours playtime.
Works perfect :)

micheljung:
*AIs

Uveso:
I validated this with a debugline and checked if the index of the `Brain` is indeed the index position inside the global `ArmyBrains` array.

I used brain.index because using a variable is faster then executing a function. And because we are already lopping over the ArmyBrains array its easy to get the index from there.

aeoncleanse:
5e91a56402deb9f00f823b729b50314dffe80763

PR #1513, opened 2016-09-17
add option to lay drop multiple lines of bombs at once

Exotic-Retard:
you can configure spread and number of lines from the blueprint, so
thats cool.

we spawn child projectiles and add velocity perpendicular to the motion
of the aircraft so they spread out.

however there appears to be a slight problem with the actual code that
does the spreading - the direction and magnitude of the displacement
vector appears to be correct, but for some reason it doesnt actually
land there. i suspect mystery lateral acceleration breaking all my
maths. or me being retarded. ah well.

obviously, this code is courtesy of the equilibrium team; and IceCreamer owes me one for this i believe ))))))


aeoncleanse:
Fraid it's not _quite_ the effect we're after :( As a follow-up, could you investigate the following instead: Have the bomb drop from the central bone (Exhaust in this case), then spawn two other bombs at a set distance from it, on each side, perpendicular to the direction of motion, but with the same velocity and acceleration as the original. This should allow parallel streams with no complex lateral acceleration and much less additional code and calculations.

Exotic-Retard:
ok then im not sure what we are after exactly. i was under the impression you wanted to drop multiple lines of exactly parallel bombs. is that correct? indeed the code currently does not achieve this (i have an idea where to fix this) but if thats not what we want then i require more explanation

Exotic-Retard:
also the way i did is is nicer than what you describe - my bombs are seen to be falling out of the same bomb and then fall away, rather than for large spreads fall from outside the plane. i dont add any lateral acceleration and my maths is mostly 2d ballistics and a for function which was very satisfying to write :)

aeoncleanse:
Your bombs spread out from a point. There must be lateral acceleration. 

But anyway, what we want is for the bombs to drop from multiple locations across the bomber, not from a single bomb door. Really at the end of the day I want to know why in the name of god the Rack system simply is not working properly.

PR #1524, opened 2016-09-20
Fix Option Tooltip "Single Unit Selected Rings"

Uveso:
Added LOC tags inside tooltips.lua (options_gui_enhanced_unitrings)
Added location strings to language files.
Added german translation


speed2CZ:
Wouldnt it be easier to do all these in one run? :P

aeoncleanse:
I've asked him to :p

Uveso:
Sure its easier, but the i don't get practice in typing Git shell commands :dancer:

PR #1527, opened 2016-09-20
Feature/rebalance reclaim

Crotalus:
- Remove factory move-attack feature / glitch / hack (reduce range 26 -> build-range 10)
- Engineers no longer split trees, which in effect will make patrol reclaiming the new move attack
- Increase reclaim time, currently by a factory of 5 - tuneable in lua/prop.lua
- Patrol (and move attack) reclaim speed is 75% of manual reclaim speed, in case we want to premier manual reclaim.
- Set the total value of the splitted trees to the sum of the tree group + 5% (https://github.com/FAForever/fa/pull/964)

All values should be seen as random set, just to get testing and a sense if the code works as intended.


aeoncleanse:
@Crotalus Please rebase this onto deploy/fafdevelop

aeoncleanse:
Parts of this PR have been cherry-picked in here: 2f90e6b089304536d0be0bcb4aba939c14f6e8c8

PR #1530, opened 2016-09-20
Bugfix/make map name filter in game option menu more obv

### LINKED ISSUE #1526: Map / Name-Filter-Text-Box is invisible
Inside options the textbox for the Map name Filter is not visible:

namefilter1

After 10 years playing i found this out:

namefilter2

Would be nice, if we could make the backgroundcolor of the textfield different from the normal background.

Justify87:
Cursor color changed to a whitish red, background enabled and set to a greenish grey
Fixes #1526 

namefilter


aeoncleanse:
72c09f9e0ca6dd08d2587e645c8de519bf4456b7

PR #1531, opened 2016-09-21
Bugfix for option "share conditions" II

Uveso:
Rebased PR from #1404


lua/aibrain.lua
```diff
@@ -531,57 +527,124 @@ AIBrain = Class(moho.aibrain_methods) {
...
+            end
+            local function TransferUnitsToKiller()
+                local KillerIndex = 0
+                local units = self:GetListOfUnits(categories.ALLUNITS - categories.WALL - categories.COMMAND, false)
+                if units and table.getn(units) > 0 then
+                    if SorianAI ~= nil then
+                        RemovePlatoonHandleFromUnit(units)
+                    end
+                    if victoryOption == 'demoralization' then
+                        KillerIndex = ArmyBrains[selfIndex].unitStats.LastKilled.COM or selfIndex
```
aeoncleanse: Why do we ever want to transfer ownership back to ourselves? We are always the owner no?

lua/aibrain.lua
```diff
@@ -531,57 +527,124 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                        TransferUnitsOwnership(units, KillerIndex)
+                    else
+                        KillerIndex = ArmyBrains[selfIndex].unitStats.LastKilled.unit or selfIndex
```
aeoncleanse: Same here

lua/aibrain.lua
```diff
@@ -531,57 +527,124 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                        TransferUnitsOwnership(units, KillerIndex)
+                    end
+                end
+                WaitSeconds(1)
+            end
+            local function ReturnBorrowedUnits()
                 local units = self:GetListOfUnits(categories.ALLUNITS - categories.WALL, false)
-                -- return borrowed units to their real owners
+                if SorianAI ~= nil then
```
aeoncleanse: Use 'if SorianAI then' instead? Probably clearer.

aeoncleanse: Even better, can move the check into the platoonhandle function. Start of that do 'if not SorianAI then return end'

lua/aibrain.lua
```diff
@@ -531,57 +527,124 @@ AIBrain = Class(moho.aibrain_methods) {
...
-                    TransferUnitsOwnership(units, owner)
+                for owner, unit in borrowed do
+                    TransferUnitsOwnership(unit, owner)
                 end
+                WaitSeconds(1)
             end
 
-            WaitSeconds(0.1)
+            -- get all brains
+            for index, brain in ArmyBrains do
```
aeoncleanse: Could store BrainCategories.Allies{} too?

lua/aibrain.lua
Uveso: Right, but this is transfering the units to our enemy player who killed us.

`or selfindex`
Is a saveguard for the rar case, that a commander is killed from his own units like a paragon explosion. In this rare case the command transfers the units to ourselves.
Sure we could ask, if Killerindex is true, but we dont need it this way.

lua/aibrain.lua
Uveso: done

lua/aibrain.lua
Uveso: yes we could, but since we don't have an option that is transfering units to our allies we don't need it.

PR #1532, opened 2016-09-21
Bugfix for All Race templates

Uveso:
In case we don't find an icon from the other faction return the old icon
instead of "false"


aeoncleanse:
Situation where this comes up?

Uveso:
Any unit that has not exact a unit name formated with 3 letters and 4 numbers like `URL0104`

Sadly Mods like Total mayhem uses unitnames like `BRNT3DOOMSDAY`

And the function itself is a converter function. In case it can't convert a string, it should return the original string and not false. because we are using the string even if its not converted.

PR #1533, opened 2016-09-21
Feature/share conditions

aeoncleanse:
Based on work by Uveso, I've cleaned it up a bit and clarified some of the English.


Uveso:
I viewed the code and it looks good.

PR for german language is up.  #1534

PR #1537, opened 2016-09-22
Bugfix (final) GameOptions.Quality is saved as -1#INF

Uveso:
Don't try to computeQuality with only 1 team

To reproduce the error select "Optimal balance" as Spawn option and start as single player.

After knowing this, the fix was easy.
Just verfying we have at least 2 teams before try to compute a quality.

I also removed the failsave from trueskill.lua
 (`if not result >= 0 and not result <= 100 then return 0 end`)
It was not working anyways...

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -1209,7 +1209,7 @@ end
 
 function autobalance_quality(players)
     local teams = nil
-    local quality = nil
+    local quality = 0
 
     for _, p in players do
         local i = p['player']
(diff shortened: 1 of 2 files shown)


Uveso:
Debug example for only 1 team:
```
***\000{ <metatable=table: 208651E0>
  teams={
    {
      { <metatable=table: 20865230>
        name="Uveso",
        rating={ <metatable=table: 20865208> deviation=500, mean=1500 }
      }
    }
  }
}\000***

table.getn(team)=1

result=-1.#IND

```

aeoncleanse:
25cb74b6cd32d7a512ce0aa2b3a89e451338df0f

PR #1546, opened 2016-09-24
Fix Tooltip WindowMode Button and translation

Uveso:
Added Tooltip to WindowMode button inside Lobby settings menu.
Added translation tags and german translation.

Also corrected some german translation strings.


aeoncleanse:
43e3797fe22e163b53b3177d26ddd88cb23bc215

PR #1548, opened 2016-09-24
fix uef scu aoe upgrade removal incorrect numbers

Exotic-Retard:
well i dunno who is responsible but this should be a crime imo. anyway
when you add 2 and remove -4 then add 2 again you get 0 aoe so if you
try to get the upgrade, remove, and get it again it no longer works.
great.

now that's fixed, and also NewMaxRadius was missing in here, lukily it
was called with (bp.NewMaxRadius or 25) so it wasn't so terrible, but
still

enjoy your game knowing it had balance terrorists working on it )

### DIFF
--- a/units/UEL0301/UEL0301_unit.bp
+++ b/units/UEL0301/UEL0301_unit.bp
@@ -331,7 +331,8 @@ UnitBlueprint {
             },
             Icon = 'heo',
             Name = '<LOC enhancements_0064>Remove Heavy Plasma Refractor',
-            NewDamageRadius = -4,
+            NewDamageRadius = -2,
+            NewMaxRadius = 25,
             Prerequisite = 'HighExplosiveOrdnance',
             RemoveEnhancements = {
                 'HighExplosiveOrdnance',


Exotic-Retard:
actually there is an alternative way to fix this and that is to clean up the upgrade code in the script itself:
`
        elseif enh =='HighExplosiveOrdnance' then
            local wep = self:GetWeaponByLabel('RightHeavyPlasmaCannon')
            wep:ChangeDamageRadius(bp.NewDamageRadius)
            wep:ChangeMaxRadius(bp.NewMaxRadius or 35)
        elseif enh =='HighExplosiveOrdnanceRemove' then
            local wep = self:GetWeaponByLabel('RightHeavyPlasmaCannon')
            wep:ChangeDamageRadius(self:GetBlueprint().Weapon[1].DamageRadius or 0)
            wep:ChangeMaxRadius(self:GetBlueprint().Weapon[1].MaxRadius or 25)
        end
`
this would make sure that we:
1. dont just add our damage radius but set it to our value - dunno might not be super great but i thought its safer
2. set our max radius to w.e. is in the blueprint and not what should be in the upgrade so you dont set it twice
3. set our damage radius in the same way.

aeoncleanse:
Feel free to go over all the ACUs and SCUs and do what I've done in BlackOps to make them work nicely :)

PR #1549, opened 2016-09-24
Hard code ACU GSR. Fixes #1256. Fixes #1516

### LINKED ISSUE #1256: Walking ACU when guarding something
An assisting unit with offensive abilities like ACU thinks it's a good idea walking towards an enemy unit within GuardScanRadius instead of assisting the actual structure.

http://forums.faforever.com/viewtopic.php?f=3&t=12379

### LINKED ISSUE #1516: TML assist with ACU
if you try to assist TML with acu, the acu will try to guard it, and will not actually assist the tml, but will seek nearby unit. (the acu can go pretty far and suicide on his own btw).

fix would be to make acu behave as an engineer with building.

aeoncleanse:


shalkya:
may i ask why the gsr of aeon acu is different from the other ? 
36.75 when the other are 31.5

aeoncleanse:
It has longer range on the upgraded gun.

PR #1550, opened 2016-09-25
Fix Tooltips and translation for preset menu

Uveso:
Added several tags for translation and german translation
Also deleted the file presetHelp.lua.
It was only used for holding a single textstring.


aeoncleanse:
2760ae0f8bec68cc31fba57634ab313a6ad2d9b9

Issue #793, opened 2015-07-20
Particles from certain enemy units/projectiles rendering through fog


Varaxis:
Sera T3 Heavy Arty Hovatham shots (shows as flame aura)
Aeon T3 Heavy Arty Emissary shots (shows as a distinctly defined glow shaped just like the projectile)
Aeon T3 Rapid Fire Arty Salvation shots (shows as dark smoke)
Sera Strategic nuke missile (esp the T4, very visible)
... more, which are not as big of a deal, like cyb arty projectile impacts, t4 bomber bomb, and sera scu gun. 

This type of intel proved to be enough of a strategic (dis)advantage that it could alert the enemy to react accordingly, taking guesswork out of things such as dodging a nuke, or pinpointing the location of valuable enemy units. Similar to how in the past, the Sera ACU regular shots showed through the fog when fighting civs, also showing wrecks, weapon impacts, and explosions.

duk3luk3:
Are you sure that this isn't a feature in the case of static artillery and nukes?

quark036:
If it is, it should probably be even between factions

Varaxis:
Picture it this way: a very effective defense versus arty is assisting a shield. If UEF can see the shots through the fog, they can effectively protect a wide area with this tactic following the shots and sending drones to assist shields accordingly.

duk3luk3:
Yes, this should be balanced - either all nukes and strategic artillery shots should be tracable, or none should be.

Varaxis:
There used to be a console command that let you clearly see enemy nuke missile paths, through the fog, actually drawing a red trail behind it, but it was disabled in a FAF prior patch.

bamboofats:
I dont have a solution to offer for the impact effects, Eg cybran arty ground impact glows. 
It comes from the CreateLightParticle() and they are all over the place from the projectile script files to the EffectTemplates file.

But i have for those trails visible through the fog. This should prevent tracing back the shots.
Trail issue affects
Aeon T3 static arty trail
Aeon Rapid fire art trail
Sera T3 static arty trail
Sera T3 mobile arty trail
Sera Scu regular shots trail
Sera Scu OC shots trail
Sera ACU OC shots trail
Cybran T1 arty trail
Sera T4 bomber trail (not the nukelike impact)
Sera Antinuke trail
Sera nuke trail (from static launcher or from naval nuke)
UEF nuke trail (from static launcher or from naval nuke)
Cybran nuke trail (from static launcher or from naval nuke)

[...]

PR #1552, opened 2016-09-25
Fix Rendering Weapon trails through Fog of War

Uveso:
issue #793

Thanks to @bamboofats who had the idea.
Changed EmitIfVisible to true inside emitter blueprints to hide them in the fog of war.


aeoncleanse:
b3a31c914d02e4013f5343fdb1361447a00855c9

PR #1554, opened 2016-09-26
Bugfix/Round Deviation of New Rating Tooltip

Justify87:
In the current beta, the player deviation shown in the new tooltip has about 10 decimal places.
This rounds player deviation via math.ceil() to the next bigger number.


GrotheFAF:
woundn't math.round() be more apropriate?

Justify87:
There is no math.round() in LUA, afaik. And it doesn't really matter for the purpose of this value anyway, because fluctuations are much higher than 1 point.

aeoncleanse:
5c0a494422a3014f0d484ba792a30c44dcc860ae

PR #1557, opened 2016-09-27
Disable beam-sound on dead (Monkey Lord)

Uveso:
Moved function for disable beam from DeathThread to OnKilled.
Added disable Beam sound to OnKilled


aeoncleanse:
1bfcbc8fe99d8470d8afb4eb33d4ce91bdb23096

PR #1558, opened 2016-09-27
Fix neptune beam at dead xes0307

Uveso:
In this case the beams are not disabled because the OnKilled Thread was inside the Weapon arrray and not part of the unit class. Just a wrong bracket.


aeoncleanse:
Errrrm. Ha. Hahaha. Hahahahahahahahahaha. 

This has been broken since GPG's original release of Forged Alliance in 2007...

AND NOBODY HAS EVER NOTICED

aeoncleanse:
696a68c664af815a9539cf5e221673fda7a38242

PR #1564, opened 2016-09-29
Bugfix/closed slots not visible

Justify87:
This makes people that join a game see the closed slots again
Solves #1237 unless anyone can tell me why one would do:
`gameInfo.ClosedSlots[i] = nil`
It's either true or false, but not nil
It also doesn't make any sense to put that in that if clause


aeoncleanse:
@duk3luk3 @Crotalus @Uveso  ?? Any ideas why this line is here? Is it safe to kill?

Uveso:
I am sorry, but you can't delete this.

ClosedSlots[i] is also used to view the available slots from the map.

Lets say you are choosing a 8 player map.
If you now select a 12 player map and ClosedSlots[i] = nil is deleted, you will se slot 9,10,11 and 12 as closed.

aeoncleanse:
OK so we need a proper solution...

Uveso:
Ok, we can delete the `gameInfo.ClosedSlots[i] = nil`.
But we need to enable slots in case we are changing the map and the previous map had less slots then the new one.

So we can simply add a loop to show the hidden slots, and make them free for use.

here is the full new function:
```
function UpdateAvailableSlots( numAvailStartSpots )
    if numAvailStartSpots > LobbyComm.maxPlayerSlots then
        WARN("Lobby requests " .. numAvailStartSpots .. " but there are only " .. LobbyComm.maxPlayerSlots .. " available")
    end

    -- if number of available slots has changed, update it
    if numOpenSlots == numAvailStartSpots then
        return
    end

    -- reopen slots in case the new map has more startpositions then the previous map.
    if numOpenSlots < numAvailStartSpots then
        for i = numOpenSlots + 1, numAvailStartSpots do
            gameInfo.ClosedSlots[i] = nil
            GUI.slots[i]:Show()
            ClearSlotInfo(i)
            DisableSlot(i)
        end
    end
    numOpenSlots = numAvailStartSpots

    for i = 1, numAvailStartSpots do
        if gameInfo.ClosedSlots[i] then
            GUI.slots[i]:Show()
            if not gameInfo.PlayerOptions[i] then
                ClearSlotInfo(i)
            end
            if not gameInfo.PlayerOptions[i].Ready then
                EnableSlot(i)
            end
        end
    end

    for i = numAvailStartSpots + 1, LobbyComm.maxPlayerSlots do
        if lobbyComm:IsHost() and gameInfo.PlayerOptions[i] then
            local info = gameInfo.PlayerOptions[i]
            if info.Human then
                HostUtils.ConvertPlayerToObserver(i)
            else
                HostUtils.RemoveAI(i)
            end
        end
        DisableSlot(i)
        GUI.slots[i]:Hide()
        gameInfo.ClosedSlots[i] = true
    end
end

```

Uveso:
[edit] checked this on 2 PC, Host and Client.

PR #1567, opened 2016-09-29
Bugfix/closed slots not visible

Justify87:
Makes closed slots visible again and prevents multiple closed slots from appearing after a map change

Thanks to @Uveso for providing a solution for this!


Uveso:
issue #1237 ;)

aeoncleanse:
1d98cf18ef15be8ee3c22d5fa16eb573704e11cd

PR #1568, opened 2016-09-30
Fix Tooltip "Assist Mex to Build Mass Storages"

Uveso:
Added location tags, tooltip description and german localisation.

I hope this is right:

assistmex


aeoncleanse:
5140fb72f6b84ac25dcd8f4a01ffaaf55f9dd9d2

PR #1569, opened 2016-09-30
Optimize Callbacks.CapMex - preventing string.sub

Uveso:
following the comment:
-- TODO, race specific prefix per builder, see lua/ui/construction.lua

here is a way to use the faction name to get blueprint prefixes for mex storage, instead using string.sub.

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -85,9 +85,9 @@ Callbacks.CapMex = function(data, units)
     if not mex or not EntityCategoryContains(categories.MASSEXTRACTION * categories.STRUCTURE, mex) then return end
     local pos = mex:GetPosition()
     local msid
-
+    local LetterArray = { ["Aeon"] = "ua", ["UEF"] = "ue", ["Cybran"] = "ur", ["Seraphim"] = "xs" }
     for _, u in units do
-        msid = string.sub(u:GetBlueprint().BlueprintId, 0, 2) .. 'b1106'
+        local msid = LetterArray[u:GetBlueprint().General.FactionName]..'b1106'
         IssueBuildMobile({u}, Vector(pos.x, pos.y, pos.z-2), msid, {})
         IssueBuildMobile({u}, Vector(pos.x+2, pos.y, pos.z), msid, {})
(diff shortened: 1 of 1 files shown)


Crotalus:
That comment is deprecated so should be removed.

Which solution is better performance wise in this case I don't know. Maybe we could create a generic utility function that transforms a bpid into a race specific one. It is used with templates among other things.

Uveso:
I dont know, if this is faster. 
I made a patch for the Black Ops Tech3 Mass storage and had to use the array version to also convert the Black Ops IDs. So this PR was already to 99% finished :)

to convert a blueprint id to a valid faction specific id, we can use the function from the all race template inside construction lua:
```
local function ConvertID(BPID,currentFaction)  -- currentFaction = 1,2,3,4
    local FirstLetterArray = { "%1", "x", "u", "b" }
    local SecondLetterArray = { ["Aeon"] = "a", ["UEF"] = "e", ["Cybran"] = "r", ["Seraphim"] = "s" }
    local SecondLetter = SecondLetterArray[currentFaction]
    for _, FirstLetter in FirstLetterArray do
        local NewBPID = string.gsub(BPID, "(%a)(%a)(%a)(%d+)",FirstLetter..SecondLetter.. "%3%4")
        -- =local xsb1012 = string.gsub(ueb1012, "(u)(e)(b)(1012)",x..s.. "b1012")
        if table.find(buildableUnits, NewBPID) then
            return NewBPID
        end
    end
    return BPID
end
```
At the moment the function is declared as local, but we can move it so we can include it from other scripts.

aeoncleanse:
We'll have this I think, definitely faster. Also clearer how it works.

aeoncleanse:
53623951ac2a4fcbb774942ee0856fe6ec894fd7

Crotalus:
Just to clarify, Uweso's code isn't faster, it's slower by ~50%, probably has to do with the table-creation part.

aeoncleanse:
I'm interested to know what kind of tools one might use to actually measure something like this.

Uveso:
I am using a timer function to stop the time:

```
    TimerStart()
    [...code...]
    LOG('Runtime: ('..TimerStop()..')')

```
Both functions are using the command `CurrentTime()`

PR #1570, opened 2016-09-30
Siren ground fire target priority fix

Ithilis:
While was changed toggle weapon to equilibrium style was copypast AA
target priority also to ground fire weapon. This fix it to naval
standards.
#BalanceTerroristsInvadingYourCode


Exotic-Retard:
so to translate: the groundfire weapon on the siren had aa weapon target priorities. now it has naval target priorities, however simple they may be

PR #1571, opened 2016-09-30
Add Warp In Effects for sACUs

speed2CZ:
Same effect as for ACUs, PhaseShield mesh that is added for few seconds
is just a personal sheild, so only one missing was for cybran sACU.


aeoncleanse:
Most of this code can be put into defaultunits.lua's CommandUnit class, could we do that please (Cuts down on code duplication.

speed2CZ:
Since every sACU/ACU has different mesh for the shield, should I put them into blueprint so it can be easily picked then? Or any better idea?

aeoncleanse:
That's certainly one way. Another would have been file concat on the UID.

speed2CZ:
Superior version here https://github.com/FAForever/fa/pull/1575

PR #1573, opened 2016-10-01
Fix hotstat localisation

Uveso:
Moved Text for hotstats from dialog_txt.lua to language files.
Deleted now unused dialog_txt.lua

Removed hotstats custom translation function (modcontrols_trad).
We are now using normal LOC(text) for translation

Added array for menu name translation because the Button names
are build by the index name of the menu array.

Added german translation
Added english text to all other language files for later translation

Restored the functionality to change the size of Icons, so they better fit to the diagrams.

If you don't know what icons i mean:

stats


Exotic-Retard:
ok so the two bin icons for mass/energy is how much you wasted, and that other icon looks like the acu icon so its how many acus/command units you killed?

in any case the fact you dont recognize them suggests that they could be improved. especially the bin ones

Uveso:
Restoring the icons was only by accident while fixing the tooltips / localisation stuff :)
You don't need the icons at all, there are tooltips if you hover the mouse over the bars.
And yes, the COM icon is resized to fit into the small bar.

aeoncleanse:
We don't want any commented code that isn't actually a comment. If code is meant to be dead, kill it properly. Also, a lot of stuff in the cleanup commit isn't cleaning. Things like importing new files, changing globals to locals. Why are these things needed?

Uveso:
@aeoncleanse 
`We don't want any commented code that isn't actually a comment. If code is meant to be dead, kill it roperly.
`
If you mean the commented LOG lines, they are temporary disabled to debug the new changes.
But i still need the lines to investigate the "No Data" error on the end-score-screen.
There are also some single commandlines commented. They are also needed to see the history of changes that was made and to repair the original functionality of the HotStats mod.

Sure i can delete them, but this will make debugging for the "No Data" error way harder.

`Also, a lot of stuff in the cleanup commit isn't cleaning. Things like importing new files, changing globals to locals. Why are these things needed?`

I am not aware that i imported new files or changed globals to locals.
Where do you see such things ?

aeoncleanse:
hotstats.lua in the cleanup commit - +local gamemain = import('/lua/ui/game/gamemain.lua') ?

Uveso:
corresponding to line 24:
`- local gamemain = import('/lua/ui/game/gamemain.lua')`

aeoncleanse:
a2718d4d8821f56482e69265a644f01ed5088085

PR #1575, opened 2016-10-02
Improve GateIn Effects

speed2CZ:
Move WarpIn functions from scripts into defaultunit.lua
Add effects for sACUs

Tested all un/upgraded s/ACUs with both effects.


aeoncleanse:
91814fcf4ce368132042ed5241deee584049b12a

PR #1577, opened 2016-10-04
Give T2 Artillery a MVRD value to help with closing fire

aeoncleanse:


Exotic-Retard:
other units in the game with this enabled have it at 2-5 units below max radius - so i would set it at something like 125, or at the very least 120. (lobo: 32 ->28 ; demolisher: 90 -> 85 duke: 750 -> 700)

presumably this is because we always want to keep a high firing arc so we hit terrain less, i dont see a reason why we should only start reducing at 70; travel time is pretty short even at max range

PR #1580, opened 2016-10-04
Adjust t4 sub height to clear sea floor

Exotic-Retard:
so this adds a slider which adds a cushioning effect on top of engine
side depth setting to have a larger offset above the seabed.

this doesnt affect the units "true" position however, only the position
of the bones. the ui,hitbox and meshextents remain unchanged.

so things like pathfinding are untouched.

this is the result of like 20 different adventures of first fighting fa code then fighting github and my own stupidity. enjoy.


aeoncleanse:
79b61482e483793c702ca8a4a9f68c5719703690

PR #1587, opened 2016-10-06
Coop - send right information to the server for leaderboards

speed2CZ:
After some digging in the old coop code I found that it was never
sending campaignID or difficulty. Difficulty will have to be included in
the future since it can be actually chanced now.

For now this could fix the leaderboards until they are reworked.

Server code:
https://github.com/FAForever/server/blob/08708002094aa1a9ca79065a436bdaa53287fa78/server/gameconnection.py#L337

Needs to be tested since I don't understand the server code. I guess someone with a test server acces is needed @micheljung :)

### DIFF
--- a/schook/lua/UserSync.lua
+++ b/schook/lua/UserSync.lua
@@ -131,7 +131,7 @@ OnSync = function()
 
     if Sync.OperationComplete then
         if Sync.OperationComplete.success then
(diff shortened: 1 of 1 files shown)


micheljung:
I'm very positive that this will fix it, and it won't break more than what is already broken, so green light from me ;)

aeoncleanse:
So we're getting rid of ID and difficulty reporting?

speed2CZ:
Yes, server already knows what map it is and doesn't care about the difficulty since in the time when it was written there was only one difficulty available for coop.

aeoncleanse:
63f6c120607cd6ebcf0947580e5ed7db129e247a

PR #1589, opened 2016-10-07
Fix option localisation

Uveso:
Fixed localisation tags for option "Reveal Civilians"
Add missing translation for "Maps Available"
Added localisation tags and translation for Sorian AI option text.


aeoncleanse:
0c351ed6a373d1bcc27feed6d98aaef219184255

PR #1590, opened 2016-10-07
Fix/share conditions

aeoncleanse:
Hashtag noticed that EcoManager bypassed the new modes because it auto-gives units when you die. This accounts for that by simply gifting them back to you, then carrying out the main mode action.

It also includes a change, killing walls when players die. I suspect they were excluded because the sheer count of walls might crash the game when trying to cycle through the GetListOfUnits table. I hope that by doing the walls on their own, and by forking the thread to happen while the main one waits, we will nullify this concern.


Sheeo:
This feels like a regression -- there was previously a scheme in place to make sure that gifting prior to dying didn't actually help in Share until Death conditions. What happened to this?

You shouldn't mix functional changes with regression fixes.

aeoncleanse:
Yup, the regression was only for the new modes. For SUD, code in SimUtils is called that kills everything with unit.oldowner == selfIIndex, but for the new modes I made the decision that it would be nice if previously gifted units weren't affects. I didn't consider into my decision the existence of auto-give, which would of course then make the new modes useless. This corrects that.

PR #1598, opened 2016-10-09
Add Option to select languages

Uveso:
This adds a option to change game language.

Auto search for existing / customized languages.

To select language go to `Options->interface->Language`


Uveso:
Tested this now several times.
Works perfect. Even if the tooltip says, you have to restart the game after changing the language, you don't need to.
Language changes if you load from lobby to the game, or if you return from the game to the lobby.

I won't change the tooltip text jet, just to be save that a user don't expect the language change right after klicking the button :)

Uveso:
I have a new PR for this.

PR #1602, opened 2016-10-10
Objectives - Find human brain instead of returning "Player"

speed2CZ:
Better aproach than hardcoding in an army.
Supports the future coop version which will rename Player to Player1
https://github.com/FAForever/fa-coop/pull/38
Fixes skirmish map "Black Sun Defense"

@aeoncleanse please add, tested ^^


aeoncleanse:
61d68f33cd2e3b1c3100ff91e535e36dfb6b46d2

PR #1605, opened 2016-10-12
Fix offset hack of billy nuke

Crotalus:
The old pings used army-1 for some stupid reason, billy nuke launch data never got to know that this now changed.


shalkya:
this doesn't fix normal ping (or normal nuke ping too) does it ?

aeoncleanse:
Should do. We'll find out soon.

Crotalus:
The ping entity when zoomed in actually needs armyIndex - 1 as spec.Owner, which #1606 will fix + the reclaim labels.

A potential problem could be mods that call the SpawnPing with the old Owner = index - 1 way though.

PR #1606, opened 2016-10-12
Hotfixes for 3660 (ping / reclaim labels)

Crotalus:
Ok, there is one place where armyIndex-1 needs to be set, when creating an entity with spec table the Owner field is actually armyIndex-1.  In this case it's for the zoomed in ping entity that's spawned when pinging.

I will add the reclaim label bugfix here when I manage to see what's wrong so wait to merge this PR until it's added here.

### DIFF
--- a/lua/SimPing.lua
+++ b/lua/SimPing.lua
@@ -17,6 +17,11 @@ function AnimatePingMesh(entity)
 end
 
 function SpawnPing(data)
+    if data.Owner < 1 then
+        WARN("SpawnPing: data.Owner cannot be less than 1")
+        return
+    end
+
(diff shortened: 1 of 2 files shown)


Crotalus:
Tested the new reclaim label implementation with the reclaim that had problems and it seems to display fine now.

Crotalus:
Yeah, I tested it when watching the problem replay (used an UI mod to test the new reclaim.lua) but needs some more testing.

Crotalus:
Hmm, one idea could be to use "Army" as argument to SpawnPing for the fixed index behavior and leave "Owner" for safety if an old mod calls it with index - 1. And add a warning about deprecation in that case.

PR #1611, opened 2016-10-18
Players giving objective units doesn't break the objective anymore

TheKeyBlue:
fixes FAForever/fa-coop#47 (speed2 told me to PR it here)
I created a UnitCallback that is called each time a player gives that unit to a different Player.
Each specific unit used in an objective receives a trigger for it. The 'new' unit now has to replace the old unit in the Objective: replace in the Targets list, reassign triggers, reapply marks and arrows.

### DIFF
--- a/lua/ScenarioFramework.lua
+++ b/lua/ScenarioFramework.lua
@@ -210,7 +210,7 @@ function OverrideKilled(self, instigator, type, overkillRatio)
     self:ForkThread(self.DeathThread, overkillRatio , instigator)
 end
 
-function GiveUnitToArmy( unit, newArmyIndex )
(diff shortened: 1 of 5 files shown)


aeoncleanse:
Why is the same code copy-pasted so many times?

TheKeyBlue:
I improved my commit and tested with a custom coop scenario: [give.zip](https://github.com/FAForever/fa/files/545932/give.zip)

- More refactoring
- Reapply triggers after give
- Ignore trigger after objective complete
- Adjust ScenarioFramework.GiveUnitToArmy to trigger OnGiven if wanted.
- Markings shouldn't change due to give

PR #1615, opened 2016-10-20
Optimizing storage of who killed the last unit or commander

Uveso:
Information about who killed the commander or the last unit is now
stored directly inside Armybrains without using "SetUnitStats".
While losing the ability to use the information inside hotstats history
(was planed but not implemented yet) it will be using less memory and
because we don't need to execute a function "SetunitStats" its way
faster then before.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2249,9 +2249,9 @@ ACUUnit = Class(CommandUnit) {
                     instigatorBrain:ReportScore()
                 end)
             end
-            ArmyBrains[self:GetArmy()]:SetUnitStat('LastKilled', "COM", instigator:GetArmy())
+            ArmyBrains[self:GetArmy()].CommanderKilledBy = instigator:GetArmy()
         else
-            ArmyBrains[self:GetArmy()]:SetUnitStat('LastKilled', "COM", self:GetArmy())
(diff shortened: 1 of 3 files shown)


lua/defaultunits.lua
```diff
@@ -2249,9 +2249,9 @@ ACUUnit = Class(CommandUnit) {
                     instigatorBrain:ReportScore()
                 end)
             end
-            ArmyBrains[self:GetArmy()]:SetUnitStat('LastKilled', "COM", instigator:GetArmy())
+            ArmyBrains[self:GetArmy()].CommanderKilledBy = instigator:GetArmy()
         else
-            ArmyBrains[self:GetArmy()]:SetUnitStat('LastKilled', "COM", self:GetArmy())
+            ArmyBrains[self:GetArmy()].CommanderKilledBy = self:GetArmy()
```
aeoncleanse: We might be able to forego the if statement by setting the flag to instigator:GetArmy() or self:GetArmy(). Would just have to check if instigator is nil or self:GetArmy() on self-destruct or friendly fire death. If not, the or would work. If instigator still comes through, just as your own army, we can use instigator:GetArmy() all the time!

aeoncleanse:
Merged into release/3660

PR #1619, opened 2016-10-22
Remove non-working 15, 30 NoRush options

### LINKED ISSUE #1618: Option "no rush timer" does not work
19 Oct 2016
http://forums.faforever.com/viewtopic.php?f=3&t=13331

If you set the no rush timer and start the game, you can see the timer counting down, but no no-rush-ring is displayed and you can move your units over the whole map.
This only happens when you set the no rush timer to 15 or 30.
5, 10 and 20 works.

Crotalus:
Surprisingly enough the engine only accepts, 5, 10, 20. Must be hard-coded somewhere

Fixes #1618

### DIFF
--- a/lua/ui/lobby/lobbyOptions.lua
+++ b/lua/ui/lobby/lobbyOptions.lua
@@ -326,7 +326,7 @@ globalOpts = {
                 help = "<LOC lobui_0319>Rules not enforced",
                 key = 'Off',
             },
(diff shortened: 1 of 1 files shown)


Uveso:
Yes, 
`NoRushOption....Off.5...10..20..`
are hardcoded inside supcom.exe. (starting at address 0A2F858)

PR #1620, opened 2016-10-22
Bugfix/#1613 beams ff

### LINKED ISSUE #1613: cybran tmd is targeting wrong rockets
replay ID 5299340, minute ~14 Giem front.
his tmds shot down his own rockets several times instead ov incoming enemy rockets

Crotalus:
CollisionBeam doesn't trigger OnCollisionCheck on the projectile so it didn't respect CollideFriendly. The check is now done in OnCollisionCheckWeapon

Fixes #1613

### DIFF
--- a/lua/sim/weapon.lua
+++ b/lua/sim/weapon.lua
@@ -42,6 +42,8 @@ Weapon = Class(moho.weapon_methods) {
             end
             self:ForkThread(self.AmmoThread, nuke, bp.InitialProjectileStorage)
         end
+
+        self.CollideFriendly = bp.CollideFriendly == true
     end,
 
     AmmoThread = function(self, nuke, amount)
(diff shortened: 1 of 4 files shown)


lua/sim/weapon.lua
```diff
@@ -42,6 +42,8 @@ Weapon = Class(moho.weapon_methods) {
             end
             self:ForkThread(self.AmmoThread, nuke, bp.InitialProjectileStorage)
         end
+
+        self.CollideFriendly = bp.CollideFriendly == true
```
aeoncleanse: What's the purpose  of the bool statement rather than just setting it, here and above?

Crotalus: Cached that value instead of calling CollideFriendly() to try to make the code as fast as possible, these comparisons are made alot.

Made an unscientific test here about the performance difference between value access and function call, varies between 30-50%:
http://www.tutorialspoint.com/execute_lua_online.php?PID=0Bw_CjBb95KQMT1lVejhhYVYzQ28

aeoncleanse: but since bp.CollideFriendly is either true or false, and the output of your bool is true or false, how is doing the bool calculation faster than just having a memory  pointer?

PR #1630, opened 2016-10-26
Rework the way reclaim labels detect zoom change

### LINKED ISSUE #1628: Reclaim view unreliable and slow
I often see reclaim numbers for wrecks that aren't there, or then I don't see reclaim numbers for wrecks that are there. Likewise the numbers "jump" around depending on zoom level and what part of map I look at. Before the optimizations you could rely on the numbers, now you have to zoom and fiddle around a while before they show correct values. I got used to using the feature alot but now it got degraded quite badly.

Crotalus:
Partly fixes #1628, Viba still had some problems with lingering "dead" labels though.

### DIFF
--- a/lua/ui/controls/worldview.lua
+++ b/lua/ui/controls/worldview.lua
@@ -161,6 +161,8 @@ WorldView = Class(moho.UIWorldView, Control) {
             GetCursor():Reset()
             self.LastCursor = nil
             self:ResetDecals()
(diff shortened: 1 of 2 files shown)


shalkya:
http://imgur.com/mTqElzd

aeoncleanse:
28a375bbc17c52584cd31db4a8d7d7e7ae5aafa6 undergoing testing in deploy/fafdevelop

PR #1647, opened 2016-10-28
Preserve last kick message. Fixes #1644

### LINKED ISSUE #1644: Persist kick reason message
http://forums.faforever.com/viewtopic.php?f=42&t=13378

aeoncleanse:


lua/ui/lobby/lobby.lua
```diff
@@ -417,9 +417,12 @@ local function DoSlotBehavior(slot, key, name)
 
                 SendSystemMessage("lobui_0756", gameInfo.PlayerOptions[slot].PlayerName)
                 lobbyComm:EjectPeer(gameInfo.PlayerOptions[slot].OwnerID, msg)
+
+                -- Save message for next time
+                gameInfo.LastKickMessage = str
```
Crotalus: I don't think we should store this in `gameInfo`, it will be transferred into the game / sim + stored in replays
(and maybe in presets but haven't looked into that code).

aeoncleanse: I thought this might be an issue, and hoped someone might pick up on it and tell me :) Any better place to store it?

micheljung: The preferences file

aeoncleanse: I think we definitely only want this to persist per session...

Exotic-Retard: well dunno, as long as you can clear it super easy i dont see why it shouldnt persist multiple sessions, so if you rehost or something then you can easily kick whoever you want with "missclick" kick message

micheljung: If you want to "persist" it for the current session only, just use a dedicated variable. But the definition of "persist" is that data lives longer than sessions ;)

Justify87:
At some point in the future I want to modify the dialog to include a drop down list of common kick reasons, which could be changed in the lobby settings I guess. I just don't have the time right now. Real life and stuff.

lua/ui/lobby/lobby.lua
```diff
@@ -410,11 +410,7 @@ local function DoSlotBehavior(slot, key, name)
             local kickMessage = function(self, str)
                 local msg
 
-                if str == "" or str == "Reason (optional)" then
```
aeoncleanse: I took it out purposefully, because once you've used it once, ever, you'll never see the 'Reason: Optional' again. I also think it's for the best if people feel the need to type something. If they can't be serious when kicking a player, that player simply won't join them again. Insults and abuse can be reported to moderators if it ever happens.

Justify87: Ok. The only reason I added that string, was, because giving the player the option to give kick-reasons, also adds another step of action for the user before the player gets kicked (instead of two clicks, its now three clicks and some mouse movement). The reason this is bad is, because there are some players that just join a game over and over again, despite being kicked a hundred times.

I want to face that problem with a 5-strike-perma-kick system. For the time being, I just wanted them to know that they don't have to give a reason every time they kick someone.

aeoncleanse: Well with this, the way it saves the message means that after you type your message for the first time _ever_, each time you go to kick a player, the last reason should be auto-entered. Which is of course the point of this PR :) Does that assuage your fears?

Justify87: A little bit :-D

PR #1649, opened 2016-10-29
Set ground fire as default for mobile units

Crotalus:
We could make this a gameplay option also per player basis. What's needed then is a SimCallback which provide the default firestate at start of game / whenever player changes that option.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -18,6 +18,7 @@ local ScenarioUtils = import('/lua/sim/ScenarioUtilities.lua')
 local Entity = import('/lua/sim/Entity.lua').Entity
 local Buff = import('/lua/sim/Buff.lua')
 local AdjacencyBuffs = import('/lua/sim/AdjacencyBuffs.lua')
+local FireState = import('/lua/game.lua').FireState
 
 local CreateBuildCubeThread = EffectUtil.CreateBuildCubeThread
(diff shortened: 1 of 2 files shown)


aeoncleanse:
Errr. Why do we want this?

JaggedAppliance:
Ground-fire is more useful, units act the same way as with return fire but you can more quickly and easily give ground-fire commands without having to change fire-state. Also the hotkey for changing fire-state is laggy.

speed2CZ:
Also the myth about ASFs performing better if set to groud fire

JaggedAppliance:
I guess you would be setting all factories to ground-fire rather than the actual units right?

Exotic-Retard:
theres a nice ui mod im using that already does this, just do what it does.
i think its called uiparty.
factories are nice and all but if you get units gifted to you, ect it doesnt work. it needs to be on select like in the ui mod, so you can be sure its consistent.

Crotalus:
As mentioned, with factories you would miss given units and also experimentals.

Yeah, it's possible to do with UI mods works but is not really optimal. I don't know how UI-party works but if it always set ground fire on selected units it will spam the sim / replay with SetCommandPair commands every time a user select unit(s). If it only does it if not already set it's not as bad I guess.

snufflesAU:
Would be nice to have the option to enable or disable this, in the Interface preferences. Possible?

PR #1661, opened 2016-11-04
New 'Close - spawn mex' slot option for AdaptiveMaps

### LINKED ISSUE #1643: add new version of "closed spot"
make a new version of "close spot" available while hosting Ultimate maps:
suggestion is "close spot - spawn mex" adds a civilian army (or an AI) to the spot which doesnt spawn an acu (i.e. an army is associated with the spawn but its basically an empty spot).
Reason: the script in the maps will allow for spawning of associated mex without totally cluttering the options

originally posted here:
https://github.com/FAForever/client/issues/419

TheKeyBlue:
When using AdaptiveMaps (this parameter needs to be set  true in the scenario file of the map) you can now decide which starting position mexes are available even though there is nobody starting there (or in the mirror location)
This allows for a lot more customization possibilities when using adaptive maps.

Standard maps will not be able to use the new 'Close - spawn mex' slot option.

One can find out which slots where set to 'Closed - spawn mex' with the following table:
ScenarioInfo.Options.SpawnMex

fixes #1643


CookieNoob:
this is super awesome!!!

CookieNoob:
I will try to get someone translate it to italian too and send you the translation :)

Exotic-Retard:
so what happens if we are running some non adaptive map?
does this show up and do nothing? if so would be nice to somehow hide it or something, dunno

aeoncleanse:
CookieNoob if you could checkout the code and review it properly that would be epic.

CookieNoob:
I already read the code and made some comments, they are already implemented.
@Exotic-Retard on a non-ultimate map the variable "AdaptiveMap" in the Scenario.lua file is NAN, thus the close spot function behaves like the normal function and the spots that were closed with the new function are then normal closed spots. If you swich back to an ultimate map, it will stay closed with the original function.
The option itself will not be visible when you arent hosting an ultimate map (it is already implemented like this)
The behavior of rightclicking a spawn location on the minimap in the lobby is also adjusted to cycle trough the different options "open - close (normal) - close spawn mex - open -...." and the appearance on the map is a green cross (instead of a red one).

I will re-read the code in the evening, but last time when I had a look it was pretty good and did exactly what I requested :+1:  for that KeyBlue!!

loc/IT/strings_db.lua
```diff
@@ -7815,6 +7815,8 @@ lobui_0427 = "FAF Game Lobby"
 lobui_0428 = "Kick Player"
 lobui_0429 = "Move Player to Observer"
 lobui_0430="Country not found. Please contact the development team on the forums, or ask a moderator for help"
+lobui_0431="Close - spawn mex"
```
CookieNoob: lobui_0431="Chiudi posto - crea gli estrattori di massa"

duk3luk3:
Please rebase these commits to squash related changes together (I think in this one, translations should be in one commit and everything else in another)

TheKeyBlue:
Something like this?

PR #1663, opened 2016-11-04
Improve Ping/CPU display

### LINKED ISSUE #1652: Lobby - Ping/CPU
Currently the ping bar is shown only if the ping is above 500ms, but the desciption above says `Ping/CPU`.
https://puu.sh/s0A8r/bbec25f4de.png
People especially newbies under 1000 ratings  often think that the shown CPU bar is player's ping and kick them for no reason.

Can we have this seperate, there should be enough space in the lobby for that. Or any other solution?

TheKeyBlue:
I seperated ping and CPU rating.
Ping should be fixed now and show only when a ping of >500.
I added some text to explain the color coding of the ping bar. (translations required)

fixes #1652


TheKeyBlue:
This is how it looks like btw.
pingcpu

Will rebase soon

lua/ui/lobby/lobby.lua
```diff
@@ -2080,8 +2080,8 @@ function CreateSlotsUI(makeLabel)
     local ColumnLayout = import('/lua/ui/controls/columnlayout.lua').ColumnLayout
 
     -- The dimensions of the columns used for slot UI controls.
-    local COLUMN_POSITIONS = {1, 21, 47, 91, 133, 467, 537, 607, 677, 749}
-    local COLUMN_WIDTHS = {20, 20, 45, 45, 229, 59, 59, 59, 62, 51}
+    local COLUMN_POSITIONS = {1, 21, 47, 91, 133, 395, 465, 535, 605, 677, 749} -- {1, 21, 47, 91, 133, 467, 537, 607, 677, 749}
+    local COLUMN_WIDTHS = {20, 20, 45, 45, 257, 59, 59, 59, 62, 62, 51} -- {20, 20, 45, 45, 229, 59, 59, 59, 62, 51}
```
aeoncleanse: Remove comments please

CookieNoob:
it would be nice to show the ping bar for a player also, when he has connection issues with someone else (at least to the host).
very simplified example: (Player 1 is the host)
Player 1 has 400 ping to Player 2
Player 1 has 400 ping to Player 3
Player 2 has 700 ping to Player 3
host doesn't see any ping warning even though some players have bad connections.
I am not entirely sure how it would be displayed in the best way but it would be nice to have some indication for that. Could be either a ping map (just as a table where all pings are listed) or a diagram (ofc more fancy) with lines in different colors indicating the ping or something completely different.

aeoncleanse:
338dbf434e46a9d3fa5b3910605e200f428103f8

HUSSAR-mtrela:
@CookieNoob @TheKeyBlue you could offset the player-to-host ping (the existing yellow bar) to the top a few pixels and then you could place a second bar to visualize worst/highest player-to-other ping with a number of the other player next to the the bar. This way, each player line will have two ping bars.

Alternatively, you could do some more intuitive diagram for showing ping between all players. This way, we can clearly see that player 2 has connection issues with other players: 3 and 5
faf-pingdiagram
note that in this design, I moved columns with players' colors next to the ping diagram for better visualization and lookup of players.

Exotic-Retard:
i think that while it would be more intuitive, it would also take up a lot of space, too much likely, especially since you need more rows for more players, and we can play 16 player games now.

HUSSAR-mtrela:
Well, the lobby UI does not scale up based on the window/screen size and that should be corrected (e.g. see code in UnitsManager.lua) because we will run into limitations in displaying other info in the lobby. In mean time, we can show this ping diagram in a new popup and by default just show:
image

aeoncleanse:
It doesn't scale because of the way it's implemented, using static (If scaled) bitmaps. It's stupid, but anything else would require a massive overhaul. I am very tempted to redesign it to a 1080p minimum at some point mind...

speed2CZ:
Sorry but I dont see requiring 1080p at minimum real any time soon, unless you want another outrage in aeolus like changing default keys.

Exotic-Retard:
i think some popup would be best, that mockup you showed would probably give sensory overload and takes up way too much space for the time its going to be used

PR #1665, opened 2016-11-05
Fix Empty boxes for closed spots

### LINKED ISSUE #1664: Closed spot shows only empty box
this affects players that join AFTER the spot was closed. This bug doesnt appear every time, but when it does it is there for all players that joined after these spots were closed. Opening the spots and closing them again fixes it.

capture

[forever.txt](https://github.com/FAForever/fa/files/572849/forever.txt)

[game.txt](https://github.com/FAForever/fa/files/572850/game.txt)

TheKeyBlue:
Set numOpenSlots to higher number so the initial UpdateAvailableSlots call doesn't terminate premature for 16player maps

fixes #1664


duk3luk3:
Straightforward change, need someone to test this.

duk3luk3:
Actually, this isn't straightforward. Setting

```
numOpenSlots = LobbyComm.maxPlayerSlots + 1
```

is counter-intuitive and violates the intuitive constraint that `numOpenSlots` can't be higher than `maxPlayerSlots`.

It would be cleaner to make the `UpdateAvailableSlots()` function aware of when it is being initially called for joining a lobby.

aeoncleanse:
I agree with duk3. Need to pass a new arg in to UpdateAvaiilableSlots I think. Or, check a variable turned true only after the first time.

TheKeyBlue:
I also agree that we shouldn't break constraints like that. 

Do you like how it is now ?

aeoncleanse:
0a0061091e54cf6476c02c280d88a5ba913a3003

PR #1672, opened 2016-11-08
Fix Score Screen for Coop + small improvements

speed2CZ:
Makes score screen work with coop. Displays debriefing videos + objectives (extra panel in the score screen that is normally hidden.

Feedback, rehost button for coop. Can be later modified also for skirmish if someone makes a good system out of it.


speed2CZ:
@aeoncleanse it's rebased and ready

PR #1680, opened 2016-11-12
Feature: Allow map required mods

TheKeyBlue:
We should allow map specific/required mods.
This could be very useful when creating all sorts of things: like survival, zone control (buttons instead of assisting units to buldings?),  coop and others.
It would make it easier to handle non-standard game modes that require some extra ui displays/capabilities. At the moment you would have to download the map and the mod seperately, which introduces all sorts of issues: like forgetting to enable/disable the mod, having a wrong version of the mod/map, Creators have to upload 2 things to both to the mod vault and map vault and they are both saved in different locations, etc.

We can easily add this feature because it only requires 10 changed lines in the code.
And some small adjustments to the init files. (more about this at the end)

The only thing that would be required to use this new feature is to add /mods folder inside your map folder (similar to /sounds and /movies folders) and put each mod the creator wants to use inside. 
If these mods should not be used outside this map then set 'selectable' to false in mod_info.
Lastly add a list of required mods in the scenario file.
`RequiredMods = { {uid = "THE_MOD_UID" }, }`

Now what needs changing in the init files? Not much actually.
change mount_map_dir so you get this : http://pastebin.com/nAE055MW
And make sure all `mount_` function calls happen after the functions otherwise there could be an issue with not yet initialized functions.


lua/ui/lobby/lobby.lua
```diff
@@ -1726,6 +1726,15 @@ local function TryLaunch(skipNoObserversCheck)
         gameInfo.GameOptions['Ratings'] = allRatings
         gameInfo.GameOptions['ClanTags'] = clanTags
 
+        scenarioInfo = MapUtil.LoadScenario(gameInfo.GameOptions.ScenarioFile)
+        
+        if scenarioInfo.RequiredMods then
+            for _,mod in scenarioInfo.RequiredMods do
+                gameInfo.GameMods[mod.uid] = true
```
aeoncleanse: This should check that the required mod has been loaded into the file system I think, to avoid having a nil value.

TheKeyBlue: good point.
What behaviour would you suggest, incase the mod isn't loaded?
Give a signal instantly when someone doesn't have the mod somehow?
Don't even allow you to choose the map in mapselect incase you don't have all the mods?

aeoncleanse: If you don't have them, don't let them select.

aeoncleanse:
Have put the init file change in to the APU route I'm working on

lua/ui/lobby/lobby.lua
TheKeyBlue: So when you select such a map in the options dialog, you can't confirm (press ok) your changes.
But i'm not sure what would be a good way to show why you can't play on that map.

- red map name?
- popup message whenever you select the map? (wouldn't this be annoying?)
- both?

aeoncleanse: Popup.
Should only happen rarely

TheKeyBlue:
So i added some dialog that shows which mods are required, with some description and where to find the mod.
It should be impossible to select a map for which you don't have the mods.
When you join a game without having the required mods, a message similar to not having the map will be displayed.

Some feedback required.

aeoncleanse:
Is the requiredmods lua file totally new, or a gpg file that you've included into FAF then modified? If the latter, please split the initial inclusion into its own commit for history purposes.

TheKeyBlue:
requiredmods.lua is completely new.
Its a new pop-up, so i made a new lua file for it.
It needs improvement, which is why I asked for feedback. 
Looks like this :
requiredmods
Each mod shows what its used for (visual/feature/etc) and where to find it incase you don't have it 
(mod_vault/forums/included_with_map)


map maker adds a table:

```
RequiredMods = {
    { 
        name = MODNAME, 
        description = {key = "1-word", tooltip = "longer description of what mod does"}, 
        howtoget = {key = '1-word', tooltip = "longer description of where to find mod"}, 
        uid = ... ,
    },
}
```

Example map I used when testing it: [Theta 4 Players.v0001.zip](https://github.com/FAForever/fa/files/679390/Theta.4.Players.v0001.zip)

HUSSAR-mtrela:
@TheKeyBlue I would replace UI element (TextLabel) that you use to display Location/HowToGet of a mod with non-editable TextBox such that people missing mods can copy text of the location rather than trying to remember it. Bonus point if you can make the location UI element to be a hyperlink that one can click on and open a website with location of the missing mod. Although, that might not be possible from lua/lobby.

HUSSAR-mtrela:
also, I think you should increase space for Info/Location of each mod and decrease space for mod name. You might want to see how we are listing info about mods in ModsManager and re-use some of its code.

speed2CZ:
opening URL from UI side is pretty simple, `OpenURL('url')`

speed2CZ:
this code of the score screen manages feedback after my changes https://github.com/FAForever/fa/blob/develop/lua/ui/dialogs/score.lua#L509-L521

currently it takes info only for campaign and if required info is passed, but I always thought it could gather info about either map or specific mod that was played from mod_info or whatever url link that was provided.

It just require to make some system out of that. But that is a work for someone else. (I just showed that it's possible :) )

aeoncleanse:
@TheKeyBlue I agree with these guys, can we have this linking a URL please? Even better if the URL can autodownload the mod like there's a URL for downloading replays, or somehow use the lobby-client link to send a vault download request...

lua/ui/dialogs/mapselect.lua
```diff
@@ -634,7 +653,9 @@ function CreateDialog(selectBehavior, exitBehavior, over, singlePlayer, defaultS
                 cancelButton:OnClick()
                 return true
             elseif event.KeyCode == UIUtil.VK_ENTER then
-                selectButton:OnClick()
+                if not selectButton:IsDisabled() then
+                    selectButton:OnClick()
+                end
```
ChrisKitching: How about a general solution to this?

Introduce a function on Button called Click(). Make the existing [click event listener](https://github.com/FAForever/fa/blob/develop/lua/maui/button.lua#L155) call `Click()`, and make `Click()` check `IsDisabled()` and maybe call `OnClick()`.

Otherwise, we're going to have to do this if statement every time we want to procedurally click a button, which is very not cool.

lua/ui/dialogs/mapselect.lua
```diff
@@ -669,6 +691,9 @@ function CreateDialog(selectBehavior, exitBehavior, over, singlePlayer, defaultS
         PreloadMap(row)
         local scen = scenarios[scenarioKeymap[row+1]]
         selectedScenario = scen
+        if not scen.Playable then
+            return
+        end
```
ChrisKitching: If the scenario isn't playable, why are we even showing it in the map list?

Should this be a default-on, non-disableable map list filter?

lua/ui/dialogs/mapselect.lua
```diff
@@ -970,6 +995,35 @@ function SetDescription(scen)
         description:AddItem(LOC("<LOC lobui_0758>AI Markers: No"))
     end
 
+    if scen.RequiredMods and table.getn(scen.RequiredMods) > 0 then
+        description:AddItem(LOC("Required Mods: "))
```
ChrisKitching: Call to `LOC()` without a `<LOC` tag - pretty sure this won't localise?

ChrisKitching: In case unfamiliar, why way `LOC()` works is a bit weird. You do something like:

`LOC("<LOC something>someJunk")`. This will look in the loaded [localisation file](https://github.com/FAForever/fa/tree/develop/loc) (an appropriate one is loaded based on your configured language) for a key called "something", and return that. The "someJunk" is ignored - but is usually set to the english version of the string to make the code more readable.

lua/ui/dialogs/requiredmods.lua
```diff
@@ -0,0 +1,322 @@
...
+MaxNameLength = 40
+MaxDescriptionLength = 16
+MaxHowtogetLength = 16
+
+function CreateDialog(parent, scenario)
+    dialogContent = Group(parent)
+    dialogContent.Width:Set(750)
+    dialogContent.Height:Set(350)
+
+    popup = Popup(parent, dialogContent)
```
ChrisKitching: Yay, people are still using my UI library :D

ChrisKitching:
It seems like this can, in many cases, be handled in the client (just download the mod when the user presses the host button).

Note that we actually have a channel for communicating messages from the lua back to the local client, so there's also no real reason why mod download can't be triggered by a signal when you set the map from the lobby. The local client can also [signal into the lobby](https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/lobby.lua#L563, although this feature is barely used. You can use that to signal that the download is finished, and then trigger mod re-indexing by calling the function in Mod.lua.

That way, you could do absolutely all of this behind the scenes, and it'd be awesome.

lua/ui/dialogs/mapselect.lua
TheKeyBlue: I think it's better to tell people something is wrong, instead of just hiding it. 
Since having it just hidden without knowing what is wrong can be annoying too.

TheKeyBlue:
I actually redid a lot of the requiredmods.lua.
But its not completely done so I haven't pushed it yet.

And I have no real knowledge about client at the moment. So no clue how easy anything you're suggesting actually would be.

lua/ui/dialogs/mapselect.lua
ChrisKitching: Ah, but, the "playable" flag is typically used to hide scenarios that shouldn't be shown in the multiplayer list for some reason. Ones that are "special" somehow. It's a flag you have to explicitly write in your map's config file - not something you're at all likely to do by mistake.

There's probably value in logging a warning "Skipping map X because Playable is set to false in its configuration", so map authors can see if they fucked up. Normal users aren't going to care (or understand). What they're going to see is they double click the map and nothing happens...

TheKeyBlue: Is "playable", an actual thing already for scenarios?
I remember that I just created that myself. 
I've never seen it used anywhere before.

You are right that there shouldn't be nothing happening. I'll have to check if the same dialog shows up as if you single clicked it. Which is what I intended to have happen.

aeoncleanse:
This has been open for a year now with little work or demand. There's also the possibility of using chris's alternative since client dev is now active. Closing.

PR #1682, opened 2016-11-13
Fix typo in keydescription

speed2CZ:
Better commit name than the peasant in #1681 :P


CookieNoob:
but not a fix for both typos

Issue #1653, opened 2016-10-31
aeon t3 mobile artillery has deployable display option


Exotic-Retard:
this thing here: http://puu.sh/s1CKE/fd01317bb9.jpg
bscly it shouldnt have that because it doesnt deploy.
should be some 1 line blueprint change

(you know people were blaming equilibrium for fucking up the t3 artillery, when actually not only we didnt but it was fafs fault that they were blaming us xD)

Exotic-Retard:
https://github.com/Ithilis/fa/commit/1059e34e810eeb3122cbfe4a5c304bb5b122d7a3
found the culprit

Exotic-Retard:
since no one bothered to fix this i guess i had to do it myself. ah well.

PR #1683, opened 2016-11-13
Remove Deploy ability from aeon t3 arty

Exotic-Retard:
it doesnt deploy. so it doesnt need to have this ability. its pretty
misleading. no gameplay changes, this is only a ui mistake.


Exotic-Retard:
fixes #1653

aeoncleanse:
Thanks :)

Issue #1690, opened 2016-11-19
LOC lobui_0592 "allow observers" overrides "to observers" in chat


CookieNoob:
![capture](https://cloud.githubusercontent.com/assets/21953792/20459136/54b87fb6-aeb7-11e6-9d97-f7ff0d973dc8.PNG)
https://github.com/FAForever/fa/blob/3de785a9523ba4b048e0ee7061c14dffb4abdb3f/lua/ui/lobby/lobbyOptions.lua#L223

https://github.com/FAForever/fa/blob/287276db8ef4db691da8f54ba477e77583666695/lua/ui/game/chat.lua#L812

PR #1691, opened 2016-11-19
Rename lobui_0592 to lobui_0692

CookieNoob:
fixes https://github.com/FAForever/fa/issues/1690

### DIFF
--- a/lua/ui/game/chat.lua
+++ b/lua/ui/game/chat.lua
@@ -809,7 +809,7 @@ function ReceiveChatFromSim(sender, msg)
     local towho = LOC(ToStrings[msg.to].text) or LOC(ToStrings['private'].text)
     local tokey = ToStrings[msg.to].colorkey or ToStrings['private'].colorkey
     if msg.Observer then
-        towho = LOC("<LOC lobui_0592>to observers:")
+        towho = LOC("<LOC lobui_0692>to observers:")
         tokey = "link_color"
         if armyData.faction then
             armyData.faction = table.getn(FactionsIcon) - 1


speed2CZ:
Did you just take random number, because I can't find `lobui_0692` in loc files

aeoncleanse:
Yeah it's not LOCed yet, but this works for now.

speed2CZ:
So why not to take the next free one? Instead of random one and adding it into loc files so someone can translate it later,
or removing the loc tag from it...

I feel like this is just making more work for someone else who would decide to localize some strings in lobby and then runs into this one already taken while several before this were not... and ofc this not being in loc files.

CookieNoob:
The first one that is free would be "lobui_0009" or "lobui_0620". I can change it to that if you want to. The real question is why did the guy who changed the other variable took one that was already taken imo.

Exotic-Retard:
well we should keep the numbers in as sane an order as possible, maybe you could even write some commit that fixes his locs as well?

in any case i would take the next available number. probably 620 since you presumably have more space there and its in a more chronological order i guess

aeoncleanse:
Ok. We take the next available slot, and reverse any damage done by the commit which caused all this. We make sure the new key is in all loc files. 

Soon, im gonna get uveso to write a script to sort the loc files out once and for all, splitting it by section into originals and faf.

Uveso:
lobui_0769 is the next free lobui key

btw, all keys in the german translation file are in alphabetical order.
There you can find the next free key very easy.

The english translation file ist different from all other translations. 
This is because we have english text in strings.db, unitdescription.lua and keydescription.lua.
All other languages only have the strings.db file that contains all strings also from unit and key description.lua.
There is no clean way to sort or seperate the FAF from the normal translation. There are already to many keys deleted and reused for FAF. Try to use the FAF translation with the vanilla game, and you will see what i mean :)

But I can build a ingame debug function that can simply print all translation-keys to the logwindow.
This way you can find a free key in seconds.
(I already use this inside my debugmod for translations)

PR #1693, opened 2016-11-21
ES (spanish) translation (FAF)

Uveso:
http://forums.faforever.com/viewtopic.php?f=2&t=13287&start=20#p139418

This is the ES translation for FAF from `angus000`

source:
http://pastebin.com/DunevFtV

updated source:
http://pastebin.com/vQPbn29P


Uveso:
well i guess i wrecked this with an invalid escape sequence.

PR #1694, opened 2016-11-21
Bugfix Cloaking

Uveso:
Fixed CloakFXWatcherThread to cloak the source unit and don't cloak already cloaked units.
Fixed Disabled cloak in case a unit has left a cloakfiled with its own cloak activated.
Checktime for cloak-renew is now 10 ticks and cloakfield is lasting 2 ticks longer then cloak-refresh timer to prevent unit flickering.


aeoncleanse:
We still need to exclude self, as a unit with a cloakfield but not a cloak doesn't cloak itself.

Uveso:
new PR is in progress

PR #1699, opened 2016-11-26
Display rating mean in player rating tooltip

### LINKED ISSUE #1687: Display mean in rank hover
Isn't the hover on the rank supposed to show your mean score? It shows `Your rating is rank +/- variance` instead of  `Your rating is mean +/- variance` (see screenshot).
screenshot from 2016-11-15 16-43-40

### LINKED ISSUE #1698: Round the new Trueskill mean rating shown in lobby tooltip
Probably we want it to be INT

duk3luk3:
Fixes #1698
Fixes #1687

Untested as always, maybe it needs to explicitly import utils.lua for the monkeypatching to work if it doesn't already.


aeoncleanse:
e5ff6031f52358d5b1269520b20bb8d05ed6389a

aeoncleanse:
Merging the rounding in a sec

PR #1702, opened 2016-11-27
fix bugs with "close - spawn mex"

CookieNoob:
fixes:
1) close - spawn mex is changed to close when "options" is clicked without changing the map
2) clients not showing closed spots


TheKeyBlue:
We discussed these changes and I'm positive they will fix most, if not all, of the currently known bugs.

PR #1711, opened 2016-12-03
Attempt to address 'ghost' reclaim labels: #1628

aeoncleanse:
This is a first attempt to follow the reclaim logic view through and try to identify the fault. I believe I identified one potential vulnerability: When a prop was destroyed, if it happened such that self.MaxMassReclaim had already been cleaned up, it would return nil and skip the creation of the data block to be synced, leading to the label never being called for destruction.

### DIFF
--- a/lua/UserSync.lua
+++ b/lua/UserSync.lua
@@ -57,8 +57,9 @@ function OnSync()
 		import('/modules/nukelaunchping.lua').DoNukePing(Sync.NukeLaunchData)
 	end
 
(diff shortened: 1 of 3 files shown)


duk3luk3:
Looks good to me (code review only, no testing)

TheKeyBlue:
Looks good. I was trying to figure out the issue myself and got to a similar solution for the labels with below threshold values.
Not sure if it'll fix the ghost labels issue (didn't test it yet), but its a better attempt then i could figure out so far.

PR #1713, opened 2016-12-04
Make sure to always check for reclaim labels

### LINKED ISSUE #1712: Reclaim overlay problems
I made an addition to the ultimate maps (at least some) where the reclaim value of the rocks can be adjusted at the start of the game.

Example Canis: Most rocks have a value of 38 mass.
reduction can be 20%, 40%,...

it works properly when you select 20% and 40%: the rocks show a value of 30/23 mass and the overlay disappears when you zoom after the stone is reclaimed.

it doesnt work properly when the reduction is larger: selecting 60% or 80% will make all markers show the 38 mass from the blueprint and these markers dont vanish after reclaiming the rock and zooming.

Tests were made with the latest version of develop.

[5494202-CookieNoob.zip](https://github.com/FAForever/fa/files/629431/5494202-CookieNoob.zip)
@Crotalus @aeoncleanse

duk3luk3:
...even if something screws with MaxMassReclaim

Hopefully fixes #1712 

Entirely unnecessarily complex fix, could be done much easier by just changing

`elseif self.MaxMassReclaim >= minimumLabelMass then` to `else`. The extra flag is entirely a premature optimization.

(Untested as always)

### DIFF
--- a/lua/sim/Prop.lua
+++ b/lua/sim/Prop.lua
@@ -39,6 +39,8 @@ Prop = Class(moho.prop_methods, Entity) {
             economy.ReclaimMassMax or 0,
             economy.ReclaimEnergyMax or 0
         )
+        -- flag to avoid bugs if something (like map script) screws around with MaxMassReclaim
+        self.HadReclaimLabel = false
 
         local pos = self:GetPosition()
         self.CachePosition = pos
@@ -119,13 +121,16 @@ Prop = Class(moho.prop_methods, Entity) {
         Entity.Destroy(self)
     end,
 
+
     SyncMassLabel = function(self)
         local data = {}
         local id = self:GetEntityId()
 
         if self:BeenDestroyed() then
             data.mass = 0
-        elseif self.MaxMassReclaim >= minimumLabelMass then
+        elseif self.MaxMassReclaim >= minimumLabelMass or self.HadReclaimLabel then
+            self.HadReclaimLabel = true
+
             data.mass = self.MaxMassReclaim * self.ReclaimLeft
 
             if data.mass < minimumLabelMass then -- Damaged or partially reclaimed to less than the threshold


TheKeyBlue:
I'm pretty sure you should put the `self.HadReclaimLabel = false before the `self:SetMaxReclaimValues(...)` since that will function call will set it to true.

`SetMaxReclaimValues -> 
UpdateReclaimLeft ->
SyncMassLabel ->
self.HadReclaimLabel = true`

duk3luk3:
Huh? It's set to false in the constructor of the `Prop` object. I don't know what you mean.

TheKeyBlue:
So look here https://github.com/duk3luk3/fa/blob/30a88e3d6663513718f6b4d43df7cca7ac9b0820/lua/sim/Prop.lua#L37

you have these 2 lines : 
`self:SetMaxReclaimValues( ... )`
`self.HadReclaimLabel = false`

But `self:SetMaxReclaimValues( ... )` already does `self.HadReclaimLabel = true` 
(you can see it when you follow the chain of functions from my previous post)

So you're setting it to `false` to initialize it, while it is actually already `true`

duk3luk3:
Ah I see what you mean, of course.

lua/sim/Prop.lua
```diff
@@ -119,13 +121,16 @@ Prop = Class(moho.prop_methods, Entity) {
...
     end,
 
+
     SyncMassLabel = function(self)
         local data = {}
         local id = self:GetEntityId()
 
         if self:BeenDestroyed() then
             data.mass = 0
-        elseif self.MaxMassReclaim >= minimumLabelMass then
```
CookieNoob: it is sufficient to replace this line with 
`else`

aeoncleanse:
I think we agreed this isn't the answer.

PR #1716, opened 2016-12-07
Feature/attack move dummy weps

Exotic-Retard:
so units that dont have a primary ground weapon dont stop on attack move. they probably should.
currently they all stop slightly different ranges, depending on what units they're likely to be behind, so if you attack move they all stop outside acu range.
it might make sense to adjust the ranges per faction since things like obsidians have different ranges, so the shields protect them automagically.

this affects t1 maa, t2 flak, t2 mob shields and t3 maa, but not the cybran one.

this should change balance its just a quality of life improvement, which is, of course currently implemented in equilibrium.
#BalanceTerroristsStrikeAgain
#NoHopeWithoutEquilibrium


Exotic-Retard:
this is the intended behavior if youre wondering: http://puu.sh/sHlJG/e7e0b984ba.jpg
units just stop at a certain range, configurable in the blueprint. here thats just outside range of acu

PR #1719, opened 2016-12-09
Bugfix/ghost reclaim

Crotalus:
Wrecks can actually "move", the engine reuses wreck-IDs on sim side which the earlier version didn't know about. Haven't done extensive testing yet but I think this may have been the problem if the engine reuses an ID for a label the player has seen at least once.

By always setting the position for a label when a non-delete update comes in we bypass this problem. The 
"ghost label" is then moved to wherever it belongs at the moment.

I'm not sure if the earlier zoom change is cherry-picked somewhere already so I added that as well.

### DIFF
--- a/lua/ui/game/reclaim.lua
+++ b/lua/ui/game/reclaim.lua
@@ -68,8 +68,13 @@ function CreateReclaimLabel(view, r)
 
     label:DisableHitTest(true)
     label.Update = function(self)
-        local view = self.parent.view
+        if self.isHidden then
+            LayoutHelpers.AtLeftTopIn(self, self.parent, -200, -200)
+            self:SetNeedsFrameUpdate(false)
+            return
+        end
 
+        local view = self.parent.view
         local proj = view:Project(self.position)
         LayoutHelpers.AtLeftTopIn(self, self.parent, proj.x - self.Width() / 2, proj.y - self.Height() / 2 + 1)
         self.proj = {x=proj.x, y=proj.y}
@@ -105,15 +110,16 @@ function UpdateLabels()
             if not label then
                 label = CreateReclaimLabel(view.ReclaimGroup, r)
                 view.ReclaimGroup.ReclaimLabels[id] = label
-            else
-                label:Show()
+            elseif label.isHidden then
                 label:SetNeedsFrameUpdate(true)
+                label.isHidden = false
+                label:Show()
             end
 
             n_visible = n_visible + 1
-        elseif label then
+        elseif label and not label.isHidden then
+            label.isHidden = true
             label:Hide()
-            label:SetNeedsFrameUpdate(false)
         end
 
         if label and r.updated then
--- a/lua/ui/game/reclaim.lua
+++ b/lua/ui/game/reclaim.lua
@@ -69,7 +69,7 @@ function CreateReclaimLabel(view, r)
     label:DisableHitTest(true)
     label.Update = function(self)
         local view = self.parent.view
-        local proj = view:Project(pos)
+        local proj = view:Project(self.position)
         LayoutHelpers.AtLeftTopIn(self, self.parent, proj.x - self.Width() / 2, proj.y - self.Height() / 2 + 1)
         self.proj = {x=proj.x, y=proj.y}
     end
--- a/lua/ui/game/reclaim.lua
+++ b/lua/ui/game/reclaim.lua
@@ -69,9 +69,6 @@ function CreateReclaimLabel(view, r)
     label:DisableHitTest(true)
     label.Update = function(self)
         local view = self.parent.view
-        if not view.ShowingReclaim then
-            self:SetNeedsFrameUpdate(false)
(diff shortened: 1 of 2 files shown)


aeoncleanse:
I found out about this last night while building the new approach. Does this also explain wrecks with no label?

Crotalus:
Yeah, hopefully those wrecks are the real owners of the ghost labels, remains to be seen - I only tested some games with AIs spamming units.

aeoncleanse:
Sorry

That image shows a wreck with a missing label. For the purposes of testing, limit == 1

Nojoke185:
just tried the commit, everything was working fine for 10 mins then a bomber was shot down
the wreck had no label but another label was created nearby which had no wreck
sent an engi to reclaim the label less wreck and the ghost label vanished

watched the replay and it always shows the wreck and the label attached
only saw this while playing the game live

Crotalus:
I added a commit which uses GetPosition() instead of GetCachePosition() in case of the actual lua object somehow isn't fully reset/recreated when an ID is re-used (which would be crazy).

Crotalus:
Heh, found another bug connected to the "movement after re-use" issue. When we create a label we use a position variable set in the Create function. This will still be the same value even if label.position is updated.

Added a commit which uses self.position in label.Update() instead which should take care of that.

Nojoke185:
I tried this new commit for around 45 mins on Daroza's sanctuary. I have always been able to see unlabelled wrecks and labels with no wrecks here before. After 45 mins everything was spot on, no issues seen at all. Maybe early days and needs more validation but looks an improvement so far

aeoncleanse:
I will merge this into the experimentation branch soon as I am home

aeoncleanse:
testing: 424f013dac6ceb34ccd795a4e62aaf644eff9dc5

aeoncleanse:
It's been pointed out (I noted it too) that once the labels are made, even if they're not showing, the GPU slowdown is constant. @Crotalus any ideas for how to fix this?

Crotalus:
Hmm, I thought I tested that frame updates were automatically disabled when the label or a parent was in hidden state but that doesn't seem to be the case after all.

d5a5a40 turns off the updates automatically when reclaim labels are turned on / off

aeoncleanse:
Crotalus, that commit _does_ cause the slow to go away once the button is release, but it brings back total chaos of labels when you zoom or pan or whatever while checking reclaim on and off.

aeoncleanse:
https://snag.gy/qw1QHe.jpg Example on Setons

Crotalus:
Ok, it was turning the frame updates off before they actually got hidden, try 8e7d8dc

aeoncleanse:
No, that's not it either. It no longer holds the reclaim in view, but when you de-toggle, the speed-up isn't total. It's still updating a good portion of them in the background. If you then zoom in, view only a small number, then let go, the FPS shoots back up to normal properly.

aeoncleanse:
You should use Slack if you're in the mood to actually fix this crap. Much faster comms...

Crotalus:
Something is definitely weird with Show() / Hide() / IsHidden() in combination with turning on / off frame updates. Try latest commit where I bypass IsHidden() with a manual set bool instead.

lua/ui/game/reclaim.lua
```diff
@@ -68,8 +68,13 @@ function CreateReclaimLabel(view, r)
...
     label:DisableHitTest(true)
     label.Update = function(self)
-        local view = self.parent.view
+        if self.isHidden then
+            LayoutHelpers.AtLeftTopIn(self, self.parent, -200, -200)
+            self:SetNeedsFrameUpdate(false)
+            return
+        end
 
+        local view = self.parent.view
```
aeoncleanse: Needed for?

aeoncleanse:
That one seems to have fixed the labels showing up in the wrong place due to zooming, but it still has  the issue that if you've just viewed a zoomed-out view, the majority of labels are set to NeedsFrameUpdate == true, so they still hog resources until you zoom in, trigger, and let go.

Also, even when zoomed in and static, it still loops over the whole Reclaim table. Would be nice if it didn't do that more than once if the view hasn't changed from last frame.

aeoncleanse:
```local ReclaimThread
function ShowReclaim(show)
    local view = import('/lua/ui/game/worldview.lua').viewLeft

    for id, r in Reclaim do
        local label = view.ReclaimGroup.ReclaimLabels[id]
        if label then
            label:SetNeedsFrameUpdate(show)
        end
    end

    if show then
        view.ShowingReclaim = true
        if not view.ReclaimThread then
            view.ReclaimThread = ForkThread(ShowReclaimThread)
        end
    else            
        view.ShowingReclaim = false
    end
end```

This code seems to fix it locally. Can you verify this approach, and think if there's a better way to achieve this?

aeoncleanse:
Closed - A totally reworked approach developed by @micheljung was used

PR #1720, opened 2016-12-09
Lobby: show From and To for private messages

TheKeyBlue:
After you sent a private message in the lobby, you can't see who you actually sent it too.
This is quite confusing so i added "From" and "To" to the chat messages, so you can easily see which messages you sent to whom and who sent which messages to you.

PC approved


aeoncleanse:
Can you resolve conflicts, I retargeted at deploy/fafbeta...

aeoncleanse:
f384c6aef0bc2cf08b851585f452cdf05fb32f91

PR #1730, opened 2016-12-16
Fix #1628 slow and unreliable reclaim view

### LINKED ISSUE #1628: Reclaim view unreliable and slow

micheljung:
Signed-off-by: Michel Jung <michel.jung89@gmail.com>


Crotalus:
Nice work, splitting / reloading world camera view causes it to crash though, it needs to recreate the labels / clear label table (or use weak references) to fix that.

micheljung:
Thx for the additions @aeoncleanse :) Did you fix what crotalus' pointed out?
And, if we include the issue-IDs in our commits it's easier to figure out why this commit was done :)

aeoncleanse:
Didny even see it and not sure what he means

speed2CZ:
Hotkeys Home and End and for splitting the view into two and joining it back to one. That causes an error

PR #1732, opened 2016-12-18
Sets hotbuild as default key layout + few additions

### LINKED ISSUE #1722: Hotbuild keys as default for new profiles
So as from the last dev call we've decided to truely enable hotbuild by setting keys for it's layout.

In addition to that it would be good to get rid of most of the debug keys since casual players don't need them, and move the important ones somewhere where they can't be hit by an accident (like log window).

speed2CZ:
Fixes #1722 

Sets hotbuild as default layout

Addition keys are:
Shift-Esc soft stop
F9 Network stats
Moved selec all land units from Ctrl-L to Ctrl-D (safety measure)
Ctrl-Shift-K instra self destruct
Moved log window from F9 to Ctrl-F9 as it might be still usefull when
helping someone, but less likely to be openned by accident.

PC Tokyto approved©


aeoncleanse:
Suggestion: ctrl-alt-k for insta. Harder to accidentally press and familiar to many as similar to ctrl-alt-del forr killing windows shit.

Suuggestion 2: shift-S for soft stop. S for stop. It's closer to the 95% of playeyrs who have used S to stop for 10 years now.

speed2CZ:
Just to clarify. This is not for players who played with their keys for 10 years now. It's for new player who will create a new profile in FA (ingame profile) that are not used to anything.

And everyone who uses hotbuild has Stop on Esc and it's the best place for it ;)

aeoncleanse:
But I can haz my instakill?

speed2CZ:
In this area we are not in "accidents" anymore. Only accident that would happen is that you kill it 5 sec earlier, but all ppl I talked so so far are happy with Shift :D
including PC Tokyto

PR #1733, opened 2016-12-18
Introducing LDoc (Lua Documenation) for FAF mod

silentsnooc:
`ldoc` did not run because of this comment.


silentsnooc:
**Before accepting this:**

Please note that I had to introduce a new directory in `fa/lua`. So now there is a `fa/lua/.ldoc` directory. This basically contains this:

``` 
├───.ldoc
│       ai.aibuilders.lua
│       ai.lua
│       ai.opai.lua
│       ai.platoontemplates.lua
``` 

and each file actually just contains something like this:

``` `
--- Short description of the module
-- @module modulename
```
and that's it.

**Why?**

Because ldoc requires us to define the `@module` _first_. After that we're allowed to use `@submodule`. Since the `lua/` directory is a bit messy, this seems to me as a solution to get our ldoc documentation running. 

We can still change this in the future very easily.

lua/AI/AIAddBuilderTable.lua
```diff
@@ -1,3 +1,6 @@
+--- Utility functions for the Sorian AIs
+-- @author Michael Robbins aka Sorian
```
Sheeo: Adding @author here is a bit questionable, especially considering that this is an original GPG-file not authored directly by Sorian?

Sheeo: Generally you shouldn't commit build artifacts like this to version control. It should ideally get published by proxy of jenkins during a build of the repo -- see https://github.com/FAForever/fa/blob/develop/.travis.yml for what we currently do during the CI build step

silentsnooc: There are not build artifacts involved. These changes do nothing than adding some comments to the existing Lua files. Also the `.ldoc` directory is *not* a build artifact. It's a workaround that was necessary in order to be able to group modules for the documentation.

silentsnooc:
In chat he wrote

Sheeo:
> Well right — I suggest you keep the namespace correct, though
> `.ldoc/lua/ai.opai.lua` should describe the `lua.ai.opai` module
> This way you even get .ldoc being excluded from the game distribution for free

So, just to make sure I get this right: I understood it that way that I just have to move those files from `fa/lua/.ldoc/` to `fa/lua/.ldoc/lua`. 

I am not certain what we call a "top-level" folder. If a top-level folder is a folder in `lua/` then it would mean that we have `fa/.ldoc/lua/` instead.

Sheeo:
I did indeed mean `.ldoc` at the very top-level, representing the directory tree of the rest of the repo.

E.g. instead of this move: `lua/.ldoc/ai.aibuilders.lua → lua/.ldoc/lua/ai.aibuilders.lua`, the following: `lua/.ldoc/ai.aibuilders.lua → .ldoc/lua/ai.aibuilders.lua`

You could do `.ldoc/lua/ai/aibuilders.lua` too, but allowed dotted shorthand names for succinctness may be a good idea.

silentsnooc:
Oh okay - I think this might be a problem. LDoc is not very flexible. You're passing just a root directory and that's basically it. The problem with moving `.ldoc` out of the `lua` directory would not work unfortunately.

Sheeo:
Isn't that merely a question of what directory you invoke ldoc on?

silentsnooc:
Oh, yes of course, we could run `ldoc` on

    ldoc fa/ -d output-dir

I'd have to check if LDoc likes that, but yes, might work. :)

silentsnooc:
okay - the documentation gets _a lot_  bigger with that move and it looks a bit more chaotic I guess :D but well, we'd have to sort it out I guess how to structure the documentation anyway.

silentsnooc:
Well .. as I suspected - trial & error traces :D

TheKeyBlue:
You can rewrite history, so its not that big of a deal.
Could use `git commit --amend` incase you're doing/undoing single lines each time.
(this adds the changes to the last commit)

Sheeo:
You'll want to familiarize yourself with the command `git rebase -i` :)

silentsnooc:
Alright - that should do it I guess :D I didn't realize I have to force push this to the remote branch ^^

silentsnooc:
I have now also created a documentation for the engine which I parsed from Wikia. 

Again, a preview can be found [here](https://www.student.tugraz.at/s.falk/engine-docs/index.html). The question here would be if the generated code (download [here](https://www.student.tugraz.at/s.falk/engine.zip)) should be added to this repository. Imho it _should_ because as people add documentation it will come in with just one PR. 

The remaining question is if the engine documentation should be _separate_ or if it would go into the current documentation (currently seen [here](https://www.student.tugraz.at/s.falk/fa/)), which is already quite huge.

aeoncleanse:
I think it should all be in one place, navigable one to the next. The "sim.XXX" should be enough to tell people this is an engine command. Or perhaps it could be "engine.sim.XXX() and engine.user.XXX()" ?

silentsnooc:
@aeoncleanse It would not be a problem to add `engine` to the namespace for engine entities. However, I decided to remove it because it was a but "too much". I agree, that everything should be in one place if possible.

silentsnooc:
I was able to create the documentation with a doxygen filter - although I had to change some stuff from this very old thing it finally worked. I need to do some testing but there is a chance that we can get something like this: http://mirror.easyname.at/ctan/web/lua2dox/examples/lua/docs/html/classAnimal.html

PR #1749, opened 2016-12-25
Use default scenario from uiutil if map is missing

speed2CZ:
This way the default map can be easily changed, without complicated
hooking, for mods like coop, tutorials or any mod that doesn't use
skirmish maps.


TheKeyBlue:
Nice. Really usefull for coop :)

PR #1750, opened 2016-12-26
include hook directory in mod_info.lua

Exotic-Retard:
so some uplords working on a git patcher and since faf wants to have
schook defined, this is it happening. no other changes bscly, cannot possible fail right?

### DIFF
--- a/mod_info.lua
+++ b/mod_info.lua
@@ -27,3 +27,6 @@ mountpoints = {
     textures = '/textures',
     units = '/units'
 }
+hooks = {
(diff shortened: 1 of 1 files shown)


speed2CZ:
ehm `schook` is already on line 21 in `moutpoints`, are you sure it should be there twice?

Exotic-Retard:
yes. a mounted directory is not the same as one specified for hooking. fir you need to mount it to do anything at all and then say that its for hooking and not just shadowing

PR #1758, opened 2016-12-30
Feature/transport wrecks

Exotic-Retard:
So now when a transport dies the units its carrying arent deleted. they fall with the transport and die on impact, leaving wrecks. neato
should look like this:
https://puu.sh/t66hI/7caa9b7954.jpg
http://puu.sh/t66oa/1d847ce574.jpg

note: this uses some absurd black magic, by hooking the Kill function which i didnt even know existed, dunno what it even does but this works. this catches the transport before its table is cleared and detaches all the units, and flags them to die on layer change.
for some reason OnImpact didnt work when i tried it but yeah.
monor bug - things like the acu and units that walk on the seabed - collide with that and die on that and not the water surface, which i why i tried to use OnImpact. ah well.

#FixedInEquilibrium
#BalanceOnslaughtNeverStups

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2778,6 +2778,12 @@ Unit = Class(moho.unit_methods) {
         if self.LayerChangeTrigger then
             self:LayerChangeTrigger(new, old)
         end
+        
+		--for units falling out of a dead transport - they are destined to die, so we kill them and leave the wreck.
+		if self.falling and (new == 'Land' or new == 'Water') and old == 'Air' then
+			self.falling = nil
+            self:Kill()
+		end
     end,
 
     OnMotionHorzEventChange = function( self, new, old )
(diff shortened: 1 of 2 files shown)


Exotic-Retard:
doin that thing where you reference an issue and its super fancy and all that https://github.com/FAForever/fa/pull/1663

lua/sim/Unit.lua
```diff
@@ -2778,6 +2778,12 @@ Unit = Class(moho.unit_methods) {
         if self.LayerChangeTrigger then
             self:LayerChangeTrigger(new, old)
         end
+        
+		--for units falling out of a dead transport - they are destined to die, so we kill them and leave the wreck.
+		if self.falling and (new == 'Land' or new == 'Water') and old == 'Air' then
```
aeoncleanse: Fix whitespace please :)

PR #1760, opened 2016-12-31
Remove debug keys for default layout

speed2CZ:
Since we've find out the hard way that this changes the layout for anyone who never changed a key, we have a change to remove debug keys for quite a few players.

Changing the rest of the keys for hotbuild will need a different approach so we don't blow up aeolus again.

Most of the keys are not needed for casual users.
Moved log to `Ctrl-F9` so it's less likely to be pressed by accident.

### DIFF
--- a/lua/keymap/defaultKeyMap.lua
+++ b/lua/keymap/defaultKeyMap.lua
@@ -159,54 +159,13 @@ defaultKeyMap = {
 }
 
 debugKeyMap = {
-    ['Ctrl-Alt-P']          = 'debug_navpath',
     ['Alt-F2']              = 'debug_create_unit',
     ['Alt-T']               = 'debug_teleport',
-    ['Alt-A']               = 'debug_run_opponent_AI',
     ['Ctrl-Alt-B']          = 'debug_blingbling',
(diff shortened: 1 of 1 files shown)


aeoncleanse:
I'd prefer we just work this into the entire menu overhaul, so this PR will sit for a while

speed2CZ:
If that is not ready for the next patch, this can be merged to remove debug keys for majority of ppl that tend to press them by accident

aeoncleanse:
Good thing is that IS the next patch. Patch will not ship until this is done.

speed2CZ:
If it's ready, we don't want to delay patch cause of feature that is only nice to have ;)

aeoncleanse:
I have decided to approach the debug keys in a different way. Since the debug keys are those which can only be used with cheats on, and are the same for default and for hotbuild (People generally don't change them) I would like to unbind/bind them depending on the cheat status of the game, as well as hiding the section in the keybind UI when cheats are off. We'll use a button for debug keybinds that hides, but when cheats are on it shows up, giving a popup allowing you to view/rebind them.

PR #1763, opened 2017-01-01
Feature/transport wreckage - Fixes #1432

### LINKED ISSUE #1432: No wrecks are created for units in a shot down transport
It would be great if you get the wrecks from shooting down drops. Right now only the transport wreck is left, the units inside the transport leave no wrecks.

aeoncleanse:
Enhanced version of Exotic's transport death that we think has ironed out the bugs with version 1. #1432


Exotic-Retard:
wait wait what about the transports landed while having units in them?

aeoncleanse:
That's why it's a PR and not merged :) Not sure what we can do about that though... Gonna be interesting.

aeoncleanse:
@Exotic-Retard - I did it :)

PR #1764, opened 2017-01-02
Remove loyalist ability to stun air units

JaggedAppliance:
Remove the loyalist ability to stun air units


aeoncleanse:
This will prevent it from stunning landed air units. I'm not sure that's what you're really after. The proper solution would be to use an alternative buff function for Loyalist which filters out units in the Air layer.

Exotic-Retard:
i would like to see the proper alternative solution :>

PR #1765, opened 2017-01-02
Rebalance Janus bomber

JaggedAppliance:
Increase the burn time and reduce the initial damage of the bombs.


aeoncleanse:
This seems likely to make Janus too weak again. Right now, it's good, perhaps a _tad_ OP. But this is a _significant_ nerf. I reckon it's too much.

PR #1766, opened 2017-01-02
Optimization for Blueprints Loading

HUSSAR-mtrela:
@aeoncleanse _this PR is replacement (re-base) for PR# #1420_

- [x] optimize table.hash function (+30% performance)
- [x] optimize bluprints.lua to save bp.CategoriesHash instead of discarding it (+32% performance)
- [x] optimize sim lua files to use bp.CategoriesHash instead of `for loops` to look up bp.Categories
- [x] optimize UnitsAnalyzer to use bp.CategoriesHash instead of looking up bp.Categories
- [x] moved blueprint pre-loading from startup of lobby to UnitsManager
- [x] added progress bar when loading blueprints in UnitsManager
- [x] added support for factions 4+ (e.g. nomads mod does not have to override UnitsManager.lua)
- [x] optimized layout of UnitsManager to maximize space for units icons (tested on different screen sizes)

  **Note this PR is ready**


HUSSAR-mtrela:
that's alright and this PR is not complete yet a few more changes in Units Manager are needed.

duk3luk3:
I'll take a look as soon as I can!

lua/system/Blueprints.lua
```diff
@@ -84,7 +84,6 @@ local function GetSource()
     return DiskToLocal(there)
 end
 
-
```
duk3luk3: :+1:

lua/AI/aibuildstructures.lua
```diff
@@ -610,14 +610,12 @@ end
...
     local bp = ArmyBrains[1]:GetUnitBlueprint(unitName)
-    for k,v in bp.Categories do
-        if v == 'BUILTBYTIER1ENGINEER' then
-            return true
-        elseif v == 'BUILTBYTIER2ENGINEER' then
-            return true
-        elseif v == 'BUILTBYTIER3ENGINEER' then
-            return true
-        end
+    if bp.CategoriesHash.BUILTBYTIER1ENGINEER then
```
duk3luk3: :+1:

lua/ui/game/unitview.lua
```diff
@@ -213,9 +213,9 @@ function UpdateWindow(info)
         end
         local techLevel = false
         local levels = {TECH1 = 1,TECH2 = 2,TECH3 = 3}
-        for i, v in bp.Categories do
-            if levels[v] then
-                techLevel = levels[v]
+        for cat, level in levels do
```
duk3luk3: 👍

lua/system/utils.lua
```diff
@@ -384,12 +384,12 @@ end
...
 --- t = { [A] = true, [B] = true, [C] = false }
---- table.hashkeys(t, true)  =>  'A, B'
---- table.hashkeys(t, false) =>  'C'
+--- table.hashkeys(t, true)  =>  { 'A', 'B' }
+--- table.hashkeys(t, false) =>  { 'C' }
 function table.hashkeys(t, value)
     if value == nil then value = true end -- defaulting to true
     local r = table.filter(t, function(v) return v == value end)
-    return table.concatkeys(r)
+    return table.keys(r)
```
duk3luk3: Comment above function still says "Concatenate keys of hash table". Does the comment need to be fixed or are there any other callsites of this function that expect the function to concatenate keys?

lua/system/utils.lua
```diff
@@ -488,15 +488,14 @@ end
 --- @param t is a table to filter
 --- @param fn is decision function to use to filter the table.
 --- @return A new table containing every mapping from t for which fn function returns `true` when passed the value.
-function table.filter(t, filterFunc)
+function table.filter(t, fn)
```
duk3luk3: 👍

lua/ui/lobby/UnitsAnalyzer.lua
```diff
@@ -668,13 +668,13 @@ end
 -- Returns unit's categories that should not be hidden in tooltips
 function GetUnitsCategories(bp, showAll)
     local ret = {}
-    if bp.Categories then
+    if bp.CategoriesHash then
```
duk3luk3: What does this function do if `bp.CategoriesHash` is missing? Shouldn't it panic / log an error / raise an exception?

lua/ui/lobby/UnitsAnalyzer.lua
```diff
@@ -1055,11 +1056,11 @@ local function CacheProjectile(bp)
...
-        local categories = {}
-        for _, category in bp.Categories do
-            categories[category] = true
-        end
-        bp.Categories = categories
+        --local categories = {}
+        --for _, category in bp.Categories do
+        --    categories[category] = true
+        --end
+        bp.CategoriesHash = table.hash(bp.Categories)
```
duk3luk3: Please don't leave commented-out code

lua/ui/game/gamemain.lua
```diff
@@ -630,7 +630,7 @@ function HideGameUI(state)
     end
 end
 
--- Given a userunit that is adjacent to a given blueprint, does it yield a
+-- Given an UserUnit that is adjacent to a given blueprint, does it yield a
```
duk3luk3: The 'an' is wrong here because "User" isn't actually pronounced with a vowel at the beginning. :P

lua/ui/game/gamemain.lua
```diff
@@ -536,12 +535,15 @@ end
 
 local _beatFunctions = {}
 
--- throttle means never run function more than 10 times per second to reduce
--- UI load when speeding up sim / replay
+-- Adds a function callback that will be called on sim beats
+-- @param fn       - specifies function callback
+-- @param throttle - specifies whether never to run a function more than 10 times per second 
+--                   to reduce UI load when speeding up sim / replay
```
duk3luk3: :+1:

duk3luk3:
I did a review because I was feeling motivated on the train this morning. I hope it's helpful to you @HUSSAR-mtrela !

lua/system/utils.lua
```diff
@@ -237,24 +237,35 @@ function sort_down_by(field)
...
     local r = {}
     if not t then return r end -- prevents looping over nil table
+    local n = 1
     for k,v in t do
-        table.insert(r,k)
+        r[n] = k -- faster than table.insert(r,k)
+        n = n + 1
     end
-    table.sort(r, comp)
+    if not skip then table.sort(r, comp) end
```
HUSSAR-mtrela: the skip is an optional parameter of this function and I will rename it to  skip_sort because it's better name for it.

lua/system/utils.lua
```diff
@@ -444,30 +485,39 @@ function table.print(tbl, tblPrefix, printer)
...
 
-    return newTable
+    return r
+end
+--- Returns total count of values that match fn function or if values exist in table
+--- @param fn is optional filtering function that is applied to each value of the table
+function table.count(t, fn)
+    if not t then return 0 end -- prevents looping over nil table
+    if not fn then fn = function(v) return v end end  
+    local r = table.filter(t, fn)
```
HUSSAR-mtrela: The table.count has an optional filtering function fn which defaults to a function that checks if a table has a value for all its keys. We do not want to default to getting size of the table whole table if the fn parameter is not specified because it would be same as table.getsize(). The main purpose of this function is to work with hashed tables, e.g.  local t = { [A] = true, [B] = false, [C] = true }
table.count(t) -- returns 2
table.getsize(t) -- returns 3
Yes, this function might be faster without using table.filter and I will test it with a few tables with varying sizes

lua/system/utils.lua
```diff
@@ -444,30 +485,39 @@ function table.print(tbl, tblPrefix, printer)
...
---- @param filterFunc Decision function to use to filter the table.
---- @return A new table containing every mapping from t for which filterFunc 
---- returns `true` when passed the value.
+--- @param t is a table to filter
+--- @param fn is decision function to use to filter the table.
+--- @return A new table containing every mapping from t for which fn function returns `true` when passed the value.
 function table.filter(t, filterFunc)
-    local newTable = {}
+    local r = {}
+    if not fn then fn = function(v) return v end end
```
HUSSAR-mtrela: - corrected `filterFunc` to `fn` in my next commits
- table.copy(t) would return the whole table and it would be wrong for a table like this one:
```
local t = { [A] = true, [B] = false , [C] = true }
table.copy(t) -- returns all 3 items
table.filter(t) -- returns all only 2 items and item B is skipped 
```
Note that `fn = function(v) return v end` is the basic filtering function out there and it should be used as default for the `fn` of `table.filter` As its name suggest this function is supposed to filter a table rather than just return a copy of it.

lua/system/utils.lua
```diff
@@ -444,30 +485,39 @@ function table.print(tbl, tblPrefix, printer)
...
+--- Returns total count of values that match fn function or if values exist in table
+--- @param fn is optional filtering function that is applied to each value of the table
+function table.count(t, fn)
+    if not t then return 0 end -- prevents looping over nil table
+    if not fn then fn = function(v) return v end end  
+    local r = table.filter(t, fn)
+    return table.getsize(r)
 end
 
 --- Returns a new table with unique values
```
HUSSAR-mtrela: yes, I will add this to comment for this function

lua/system/Blueprints.lua
```diff
@@ -402,44 +411,39 @@ function PreModBlueprints(all_bps)
...
-            cats[cat] = true
-        end
-
-        if cats.ENGINEER then -- show build range overlay for engineers
+        if bp.CategoriesHash.ENGINEER then -- show build range overlay for engineers
             if not bp.AI then bp.AI = {} end
             bp.AI.StagingPlatformScanRadius = (bp.Economy.MaxBuildDistance or 5) + 2
-            if not cats.OVERLAYMISC and not cats.POD then -- Exclude Build Drones
-                table.insert(bp.Categories, 'OVERLAYMISC')
+            if not bp.CategoriesHash.OVERLAYMISC and not bp.CategoriesHash.POD then -- Exclude Build Drones
```
HUSSAR-mtrela: yes, you are correct - it's not necessary here anymore and it will remove it even thou it will be very very slightly improvement without: bp.CategoriesHash.OVERLAYMISC in the if statement.

lua/system/Blueprints.lua
```diff
@@ -345,7 +348,7 @@ function HandleUnitWithBuildPresets(bps, all_bps)
                         t = t + (tempBp.Enhancements[enh].BuildTime or 0)
                         -- HUSSAR added name of the enhancement so that preset units cannot be built 
                         -- if they have restricted enhancement(s)
-                        table.insert(tempBp.Categories, enh) -- do not change case of enhancements
+                        bp.CategoriesHash[enh] = true -- hashing without changing case of enhancements
```
HUSSAR-mtrela: Nice catch of my mistake of changing from tempBp to bp. i will make sure to correct it and follow your suggestions for commenting code.

lua/ui/lobby/UnitsAnalyzer.lua
HUSSAR-mtrela: hmm, bp.CategoriesHash should always exists because it is created in Blueprints.lua and that's only file that loads bp objects.  Previously, I was creating it in UnitManager. I will remove this if condition and test it of course on all bp objects.

lua/ui/lobby/UnitsAnalyzer.lua
HUSSAR-mtrela: yes, I will correct it and do another pass through all files that I changed to check for commented-out code

HUSSAR-mtrela:
@duk3luk3 yes, this is very helpful and I really appreciate your review.

lua/system/utils.lua
```diff
@@ -339,11 +339,14 @@ function table.indexize(t)
...
 --- Converts a table to a new table with values as keys and values equal to true
+--- it is useful for quickly looking up values in tables instead of looping over it 
+--- table.hash { [1] = 'A',  [2] = 'B',  [3] = 'C' } => 
+--             { [A] = true, [B] = true, [C] = true } 
 function table.hash(t)
     if not t then return {} end -- prevents looping over nil table
     local lookup = {}
     for k, v in t do
-        lookup[tostring(v)] = true
+        lookup[v] = true
```
HUSSAR-mtrela: I added 2nd parameter that determines whether or not perform `tostring(v) `
With my latest changes:
- there is not change to the semantics of` table.hash` function
- yes, this function is used only with tables that have strings as values  
- no, there are no tables this function will not work with

lua/system/utils.lua
HUSSAR-mtrela: yes, the comment should be corrected and I have just done so

HUSSAR-mtrela:
@aeoncleanse this PR is ready and let me know if you cannot resolve above conflict in gamemain.lua

aeoncleanse:
👍 Waiting on rebase onto deploy/fafdevelop. I have already retargeted the branch.

aeoncleanse:
Merged as 797050aae7d38c427c7eb950a5da5a79f6605351

PR #1771, opened 2017-01-04
Allow Novax Satellite to be rebuilt

aeoncleanse:
TODO: Decide on how much the Satellite should cost, and how much reclaim it should drop


HUSSAR-mtrela:
@aeoncleanse IMO, we should first consider purpose of this unit and then it will be easier to decided how much it should cost. As we know it is called Defensive Satellite System but most people use it as an offensive weapon for sniping high-value targets or blocking nuke launches by hovering over enemy nukes. As its name suggest, this unit should be primary use for defense. That's why I suggest the following:
- add offensive weapon where the satellite can move and fire as it currently does
- add defensive weapon where the satellite can move within some limited range from Novax center (e.g. 200) but it's fire-rate and damage is increased 
This could be done either by implementing a toggle on the satellite for changing firing mode (e.g. something like sniper bots have or like anti-air units have 2 weapons that they can switch between)

Alternatively (and most likly easier to implement), Novax center could build 2 different versions of satellites: 
- offensive satellite with unlimited range of movement but with higher unit cost
- defensive satellite with limited range of movement but with lower unit cost

Exotic-Retard:
to make any sense of how much it should cost you first need to decide when it would be (re)built. currently that's only in the case when its hit by a nuke, so you can have the price pretty low for that, since you can only do that in extremely specific circumstances

so you also need to decide what would be able to kill it. only semi reasonable choice i see is smd, but even then i dont think its that reasonable if you can use sats to deplete antinukes, when the sat itself is defendable with shields.

personally i think it doesnt make too much sense to change the sats invincible status and instead just make the rebuild feature in the case of nukes. atm the sat is balanced like an artillery piece and its not far off: 28k mass for ~200dps so if you would build multiple sats from the same centre you would need a similar cost ratio and therefore make the first sat free or the centre cost a lot less in comparison, and then you would have the (perhaps interesting) choice of building more sats from the same centre or building another centre

i would get it working first with a cost of ~3000 mass and see what happens. to maintain the closest behaviour to current it needs to cost exactly 19% of the centres mass (the best you can get with reclaim+ rebuild)

PR #1776, opened 2017-01-14
add the outdated adaptive maps to blacklist.lua

CookieNoob:
due to a small and very unfortunate bug I had to update the adaptive maps once...


CookieNoob:
ups too much cleanup xD

PR #1783, opened 2017-01-22
Doubles reclaim time of wrecks. Fixes #1484

### LINKED ISSUE #1484: Increase reclaim time of wrecks
With reclaim values being quite high (81% for stuff on land) a major increase of the time in takes to reclaim the fields of battle to mitigate the negative effect of a failed attack.

Maybe also introduce a way to reclaim faster by either moving in a "reclaimer unit" (high reclaim rate but slow/weak) or put down some kind of mass extractor building which then automatically reclaims wreck in a radius while it is active.

MoltenMan:
Makes feature #1484


aeoncleanse:
Implemented slightly differently here: 41e157f5c53ba2c8b61c6f1226b5604211fbecc4

PR #1785, opened 2017-01-24
UI refactoring + improvement to accommodate more factions

TheKeyBlue:
Everything relating unitView is a simple refactor to allow easy hooking.

In avatars it contains some small changes to remove 4 faction hardcoding.

### DIFF
--- a/lua/ui/game/avatars.lua
+++ b/lua/ui/game/avatars.lua
@@ -612,7 +612,8 @@ function CreateIdleEngineerList(parent, units)
 
         for index, units in engineers do
             local i = index
-            if i == 3 and currentFaction ~= 1 then
+            -- ADDED SUPPORT FOR CUSTOM FACTIONS HAVING FIELD ENGINEERS
+            if i == 3 and (not Factions[currentFaction].IdleEngTextures or not Factions[currentFaction].IdleEngTextures.T2F) then
                 continue
             end
             if not self.icons[i] then
@@ -766,7 +767,7 @@ function AvatarUpdate()
     -- Find the faction key (1 - 4 valid. 5+ happen for Civilian, default to 4 to use Seraphim textures)
     -- armiesTable[GetFocusArmy()].faction returns 0 = UEF, 1 = Aeon, 2 = Cybran, 3 = Seraphim, 4 = Civilian Army, 5 = Civilian Neutral
     -- We want 1 - 4, with 4 max
-    currentFaction = math.min(GetArmiesTable().armiesTable[GetFocusArmy()].faction + 1, 4)
+    currentFaction = math.min(GetArmiesTable().armiesTable[GetFocusArmy()].faction + 1, table.getn(Factions))
     
     if avatars then
         for _, unit in avatars do
--- a/lua/ui/game/layouts/unitview_left.lua
+++ b/lua/ui/game/layouts/unitview_left.lua
@@ -6,13 +6,14 @@ local options = Prefs.GetFromCurrentProfile('options')
 local NinePatch = import('/lua/ui/controls/ninepatch.lua').NinePatch
 
 local iconPositions = {
-    [1] = {Left = 70, Top = 60},
-    [2] = {Left = 70, Top = 80},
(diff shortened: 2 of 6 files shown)


lua/ui/game/avatars.lua
```diff
@@ -763,10 +764,11 @@ function AvatarUpdate()
...
 
-    -- Find the faction key (1 - 4 valid. 5+ happen for Civilian, default to 4 to use Seraphim textures)
-    -- armiesTable[GetFocusArmy()].faction returns 0 = UEF, 1 = Aeon, 2 = Cybran, 3 = Seraphim, 4 = Civilian Army, 5 = Civilian Neutral
-    -- We want 1 - 4, with 4 max
-    currentFaction = math.min(GetArmiesTable().armiesTable[GetFocusArmy()].faction + 1, 4)
+    currentFaction = 0
+    local army = GetFocusArmy()
+    if army >= 0 then
+        currentFaction = GetArmiesTable().armiesTable[GetFocusArmy()].faction + 1
+    end
```
aeoncleanse: What happens here for the civs? Reason it was limited to 4 was to use Sera index for the civs. Do we not still need some kind of limit?

TheKeyBlue: I'm thinking that was an old version that had a different faction for civs.
Because currentfaction is used like this: `Factions[currentFaction]`
And since there is no civ faction in Factions.lua, i don't think thats an issue.

TheKeyBlue:
commented out code?
do you mean this?
https://github.com/TheKeyBlue/fa/blob/03ea71c30da73fc8e3c1502081082c86a39c930b/lua/ui/game/unitview.lua#L38

cause i can't see much else at first glance.

aeoncleanse:
That's some, unless it's useful info for some reason, but I was mainly talking about this kind of thing: https://github.com/FAForever/fa/pull/1785/commits/03ea71c30da73fc8e3c1502081082c86a39c930b#diff-9a67954bde3191933926b5c2339b46a5R119

lua/ui/game/avatars.lua
aeoncleanse: Please test it.

TheKeyBlue: Woops, you were right.
I'll get that fixed.

TheKeyBlue:
Removed the comments.
Didn't remove the ones i pointed out, because it was an example of what the `info ` parameter contains. Which is usefull when you want to understand/change it.

TheKeyBlue:
Well since i'm working with unitview anyway, i quickly added an icon for buildrate in this PR too.

aeoncleanse:
Could I get a screenshot of that new icon?

TheKeyBlue:
its an scfa original icon from the same folder as the others.
buildrateicon

speed2CZ:
It's always showing the new icon for build power
For units that are not finished yet and for units that don't have any build power (pgen, intie, ...)

TheKeyBlue:
Okay i integrated it more into how the other stats work, but i can't remove it from units that are being built. (this is how it has always been btw)
There doesn't seem to be any simple ui call for that.

aeoncleanse:
It should show only when the Mass and Energy signs are also showing. Is that the case?

TheKeyBlue:
euh no. production/consumption of mass/energy depends on the unit being alive producing or building something. So it actually using/producing the resource.
While buildpower is constant, so its always available. This means it will be showed even when its not finished.
I tried to add an exception for unfinished buildings but as mentioned i haven't figured out how to do that.

aeoncleanse:
Use self:GetFractionComplete() - Or something similar, that's off the top of my head. It's used all over unit.lua and defaultunits.lua

TheKeyBlue:
yeah this is ui stuff, so you can't get use sim methods.
Washy also couldn't figure out how to do it. So doubt there is a way unless we add something new. (which is kinda overkill in my opinion, since nobody has complained about it really)

aeoncleanse:
01ec456d8caf4c7232c9bd346bc95f1dd3eb11d2

PR #1786, opened 2017-01-25
Add support factories to All Race Templates feature

speed2CZ:
Im not sure what the "b" in 
```LUA
local FirstLetterArray = { "%1", "x", "u", "b", "z" }
```
stands for since I dont recall any unit ID starting with b

### DIFF
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -133,6 +133,10 @@ local enhancementTooltips = {
     Back = 'construction_tab_enhancment_back',
 }
 
+-- First and second letter of blueprint ID for faction conversion
+local BPID_FirstLetterArray = { "%1", "x", "u", "b" }
+local BPID_SecondLetterArray = { ["Aeon"] = "a", ["UEF"] = "e", ["Cybran"] = "r", ["Seraphim"] = "s" }
+
 --
 -- Workaround for an apparent engine bug that appeared when engymod was deployed (by Rienzilla)
 --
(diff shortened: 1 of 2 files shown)


aeoncleanse:
BlackOps?

lua/ui/game/construction.lua
```diff
@@ -2322,7 +2322,7 @@ function OnSelection(buildableCategories, selection, isOldSelection)
                 if currentFaction then
                     sortedOptions.templates = {}
                     local function ConvertID(BPID)
-                        local FirstLetterArray = { "%1", "x", "u", "b" }
+                        local FirstLetterArray = { "%1", "x", "u", "b", "z" }
                         local SecondLetterArray = { ["Aeon"] = "a", ["UEF"] = "e", ["Cybran"] = "r", ["Seraphim"] = "s" }
```
aeoncleanse: I don't suppose you could make this non-hardcoded to 4 factions, similar to what theKeyBlue's PR is doing elsewhere? Perhaps keep it hardcoded but move it outside the function to at least make it easy to hook?

Exotic-Retard: perhaps using this would work: 
```
                    local function ConvertID(BPID)
                        local prefixes = currentFaction.GAZ_UI_Info.BuildingIdPrefixes or {}
                        for k, prefix in prefixes do
                            if table.find(buildableUnits, string.gsub(BPID, "(%a+)(%d+)", prefix .. "%2")) then
                                return string.gsub(BPID, "(%a+)(%d+)", prefix .. "%2")
                            end
                        end
                        return false
                    end 
```

also on that note i would just take a look at the whole nomads construction.lua, maybe there are more goodies to integrate from there

speed2CZ:
Is this what you're looking for?

Exotic-Retard:
looks much better i think

Exotic-Retard:
actually, lets try this: https://github.com/FAForever/fa/pull/1787

aeoncleanse:
We'll use retard's

speed2CZ:
Don't you want to extrack that cleanup commit from it? Cause that was some honourable work on a very ugly file

PR #1788, opened 2017-01-25
Increase kills unitstat of killer

TheKeyBlue:
I am assuming you have to give 'kills' to the killer instead of to the victim.
If not, then this shouldn't really be in OnKilledUnit..


aeoncleanse:
You've misunderstood this code I think. This code stores the bp id of the unit killed on the army which killed it, so that the score later can list "X of unit Y killed".

TheKeyBlue:
oh yeah i confused unitid with entityid...

PR #1793, opened 2017-01-27
Increase beetle hitbox size and put torso as primary target

aeoncleanse:


aeoncleanse:
@CookieNoob What do you think, can you test these numbers locally? Works for me.

CookieNoob:
@aeoncleanse yea I can test it locally, I am just not home atm xD will need to wait a few hours

PR #1795, opened 2017-01-27
Fix cybran civilian buildings BP categories

speed2CZ:
This was throwing an error in log


aeoncleanse:
That's excellent, thanks :)

PR #1805, opened 2017-02-03
Coop improvements

speed2CZ:


aeoncleanse:
This looks fab. We'll just have to remember to triple check hotstats in our testing before release.

PR #1284, opened 2016-05-17
Hide start position and faction when they are random

Crotalus:
Rebased #1215 

TL;DR
- Shuffle positions within a team (sim-side) to prevent UI mods of showing start position of random spawns
- Hide random faction until a player is scouted

There are UI mods floating around which shows start position of a player by looking at the scenario files for ArmyID ->  position. Another easy way is to press F11 if you memorized the positions (Setons - 1, 2 for front .. 7, 8 air)

<Link href="http://forums.faforever.com/viewtopic.php?f=41&t=11577" target="_blank">http://forums.faforever.com/viewtopic.php?f=41&t=11577</Link>

With random / balanced spawns this should be considered an exploit. By shuffling the team positions sim side we prevent user side to pin down the start positions.

Another improvement could be a game option "Incognito" which completely hide the names of the enemies (show in replay / if players choose to reveal themselves during game). Random colors then also ofc. Currently high ranked players are ganked (and vice versa with weak players). Would be better if such strategies were a result of scouting an opponent's gameplay, not simply by looking at the commander name.

PR #1810, opened 2017-02-07
Hide start position and faction when they are random

aeoncleanse:
Rebased for  @Crotalus 

Rebased #1284 

TL;DR

Shuffle positions within a team (sim-side) to prevent UI mods of showing start position of random spawns
Hide random faction until a player is scouted
There are UI mods floating around which shows start position of a player by looking at the scenario files for ArmyID -> position. Another easy way is to press F11 if you memorized the positions (Setons - 1, 2 for front .. 7, 8 air)

http://forums.faforever.com/viewtopic.php?f=41&t=11577
With random / balanced spawns this should be considered an exploit. By shuffling the team positions sim side we prevent user side to pin down the start positions.

Another improvement could be a game option "Incognito" which completely hide the names of the enemies (show in replay / if players choose to reveal themselves during game). Random colors then also ofc. Currently high ranked players are ganked (and vice versa with weak players). Would be better if such strategies were a result of scouting an opponent's gameplay, not simply by looking at the commander name.


loc/DE/strings_db.lua
```diff
@@ -4682,6 +4682,7 @@ lob_0004="Kein Zuschauer erlaubt."
 lob_0005="Vom Host rausgeworfen."
 lob_0006="Keine Clients in Warteschlange des Starts erlaubt."
 lob_0007="Host hat die Lobby verlassen"
+lob_0008="Spiel begonnen hat."
```
CookieNoob: wtf xD who translated this? If you translate that to English this means "Game started has."

if this is supposed to be the translation of "Game was launched." it should read 
"Spiel bereits gestartet." = "Game was already launched"

aeoncleanse:
OK, I rebased this after having fixed the global reference error in an early commit elsewhere. It now launches. We will need to run games of this with someone keeping an eye on the server logs, to make certain shuffling these teams around does not interrupt the server's keeping track of player IDs, and re-introduce the old rating bug

PR #1821, opened 2017-02-08
Lobby: remove unnecessary and harmfull SetPlayerOption

TheKeyBlue:
Remove SetPlayerOptions which should only be used for local slots or hosts.
Every player will do SetPlayerOption for himself so not everyone else should do it for others anyway.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -1690,7 +1690,6 @@ function UpdateAvailableSlots( numAvailStartSpots, scenario )
                 end
             else
                 GUI.slots[i].faction:SetItem(playerFactionIndex)
-                SetPlayerOption(i, 'Faction', playerFactionIndex)
                 gameInfo.PlayerOptions[i].Faction = playerFactionIndex
             end
         end


aeoncleanse:
What  does this introduce or remove, that we can test?

TheKeyBlue:
Use this map to host a game and let someone else join you: 
[Theta.4.Players.v0001.zip](https://github.com/FAForever/fa/files/762512/Theta.4.Players.v0001.zip)
Then the joiner will get an error message in his log + some weird lobby behaviour: 
`\lua\ui\lobby\lobby.lua(4284): access to nonexistent global variable "key"`
Which is due to the bad warning message from the other PR i made #1823

But the issue that is fixed here is the fact that the warning message shouldn't even happen in the first place.

aeoncleanse:
I'm still not sure about this. I need to know what this line is there for before removing it. Why is it harmful? Does it allow cheating? Or is this just how the local player sets the right faction for other players based on sent data?

TheKeyBlue:
Okay. This was recently added with my restricted factions lobby change.

SetPlayerOption sends the value of an option for a player to the other players. In this case their faction.
Currently in the implementation everyone does SetPlayerOption for all players whenever a new map with different restrictions happens. Basicly it tries to change the option value of another player. This isn't allowed for obvious reasons (can't let you change other people's colors or teams etc), only the host has the power to do that. So since SetPlayerOption only sends out the values, because setting the appropriate values locally is done by the remaining 2 lines, we should be able to simply remove this line without any issue.

Basicly it tries to send changed values to other players, which isn't allowed except for your own slot and thus it triggers warnings down the line which isn't what we want.

Look here: 
SetPlayerOption : https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/lobby.lua#L4354
SetPlayerOptions : https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/lobby.lua#L4332

PS: maybe not really harmful. If the warning message didn't break, then this line would only trigger a warning message and do nothing.

aeoncleanse:
So there is no reason it was added?

TheKeyBlue:
Well it was part of the entire addition. So yeah mistakes were made.  :/

PR #1836, opened 2017-02-21
Allow for some more customizability of AbilityButtons

TheKeyBlue:
Allow some more customizability for custom AbilityButtons.
I needed this to make an Overcharge like button that gets unlocked with an ACU upgrade.

- AbilityRequirement : checks if the requirements for this ability are met. So you can enable/disable it easily.
- behavior : allow for a button behavior that doesn't require a command mode (which is what AbilityButtonBehavior does)

GetCurrentSelection() : so you can query the current selection in the onframe function of your button.

### DIFF
--- a/lua/ui/game/orders.lua
+++ b/lua/ui/game/orders.lua
@@ -150,6 +150,10 @@ end
 local orderCheckboxMap = false
 local currentSelection = false
 
+function GetCurrentSelection()
+    return currentSelection
+end
+
 -- helper function to create order bitmaps
 -- note, your bitmaps must be in /game/orders/ and have the standard button naming convention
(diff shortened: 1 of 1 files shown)


aeoncleanse:
Does anything in FAF actually use this??

TheKeyBlue:
Well as far as i know only Eye of Rhianne, but since it doesn't use those values, it shouldn't be an issue.

And i hoped it could get put here instead of some big ass hook. If you prefer i don't add it here, I could PR a refactor so its easier to hook. Can do that too.

TheKeyBlue:
Alright i changed it into a refactor PR : #1839

PR #1849, opened 2017-02-27
update cybran mass storage regen

JaggedAppliance:
regen 2 -> 4


aeoncleanse:
Merged the change into an earlier commit already :)

PR #1851, opened 2017-02-28
Change Attack move to be more beautiful

Exotic-Retard:
more attacky and more movy now - has animated arrows along the order
line and a red move order rather than attack order

looks just like patrol but red now.


aeoncleanse:
da7a0dc21f76984a0226dc3e383f47a04f99a9d9

PR #1852, opened 2017-02-28
Create a new button for attack move command

TheKeyBlue:
As described here is a new button that gives you attack move.
It uses a dummy ScriptTask AttackMove so we use the RULEUCC_Script's commandmode.
Then when you've clicked on the location the command will be adjusted similarly to how the new AssistMex to cap it functionality works.

Currently it should be available for all mobile units that either have the RetaliateToggle or are engineers.
This requires some testing to see if there are units included in the above description that shouldn't have it or units that are missing from it.


lua/sim/tasks/AttackMove.lua
```diff
@@ -0,0 +1,24 @@
...
+--*
+--* Dummy task for the Attack Move button
+--*****************************************************************************
+local ScriptTask = import('/lua/sim/ScriptTask.lua').ScriptTask
+local TASKSTATUS = import('/lua/sim/ScriptTask.lua').TASKSTATUS
+local AIRESULT = import('/lua/sim/ScriptTask.lua').AIRESULT
+
+AttackMove = Class(ScriptTask) {
+    
+    OnCreate = function(self,commandData)
```
aeoncleanse: We don't need this function.

lua/SimCallbacks.lua
```diff
@@ -231,3 +231,11 @@ function IsInvalidAssist(unit, target)
         return IsInvalidAssist(unit, target:GetGuardedUnit())
     end
 end
+
+Callbacks.AttackMove = function(data, units)
+    if data.Clear then
+        IssueClearCommands(units)
+    end
+    IssueFormAggressiveMove(units, data.Target,'AttackFormation' , data.Rotation)
+
```
aeoncleanse: Dead whitespace

TheKeyBlue:
Why was this merged if I still needed to remove some stuff?

aeoncleanse:
We can fix them later :)

PR #1865, opened 2017-03-06
Check if the lastscenario is of the correct type for this host

TheKeyBlue:
Fixes some bugs for this https://github.com/FAForever/fa/commit/127608ceb5de4420c67e07323de94bc70219228b

In normal play all maps are of type skirmish, right?


speed2CZ:
Yes, only skirmish

PR #1884, opened 2017-03-14
Refactor BaseManager's default templates creation

speed2CZ:
Using existing function for converting template to other factions rather than milions IFs.

This is 100% campaign stuff


aeoncleanse:
b61b455e7db749531e2b22e1df70debbe59ae388

PR #1894, opened 2017-03-25
Added Pixel Shader for Nomads and tweaked bloom values

ThomasHiatt:
Copied and renamed the old NormalMappedPS function to
NomadNormalMappedPS and updated all nomad techniques that were using the
old version to use the new version. The new version adjusts the color
and alpha values to give things a bit more glow while not desaturating
the color.

### DIFF
--- a/effects/mesh.fx
+++ b/effects/mesh.fx
@@ -2219,10 +2219,10 @@ float4 NormalMappedPS( NORMALMAPPED_VERTEX vertex,
 	return float4( color.rgb, alpha );
 }
 
-/// NomadNormalMappedPS
+/// NomadsNormalMappedPS
 ///
 /// Lighting using normal maps
-float4 NomadNormalMappedPS( NORMALMAPPED_VERTEX vertex, 
+float4 NomadsNormalMappedPS( NORMALMAPPED_VERTEX vertex, 
                        uniform bool maskAlbedo, 
                        uniform bool glow, 
                        uniform bool hiDefShadows,
@@ -6770,7 +6770,7 @@ technique GlassAlpha_LowFidelity
 
 // STRUCTS
 
-struct NOMADBUILD_VERTEX
+struct NOMADSBUILD_VERTEX
 {
     float4 position : POSITION0;
     float3 normal : TEXCOORD3;
@@ -6786,7 +6786,7 @@ struct NOMADBUILD_VERTEX
 
 // VERTEX SHADERS ---------------------------------------------------------------------------------------------
 
-NOMADBUILD_VERTEX NOMADBuildVS(
+NOMADSBUILD_VERTEX NOMADSBuildVS(
     float3 position : POSITION0,
     float3 normal : NORMAL0,
     float3 tangent : TANGENT0,
@@ -6803,7 +6803,7 @@ NOMADBUILD_VERTEX NOMADBuildVS(
     float1 colorLookup : TEXCOORD7
 )
 {
-    NOMADBUILD_VERTEX vertex = (NOMADBUILD_VERTEX)0;
+    NOMADSBUILD_VERTEX vertex = (NOMADSBUILD_VERTEX)0;
     CompatSwizzle(color);
 
     float4x4 worldMatrix = ComputeWorldMatrix( anim.y + boneIndex[0], row0, row1, row2, row3);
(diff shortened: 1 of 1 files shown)


CookieNoob:
uhm. Thats not what we wanted actually... You change the current shader instead of introducing a new shader to which we would switch the units to that should use it.

What it should instead do:
introduce a new technique NomadsUnit2 that uses the new pixelshader
leave the old function NomadUnit unchanged, but rename it to NomadsUnit1

Exotic-Retard:
actually what cookie wrote is not quite right either xD
we want to keep a set of not renamed techniques temporarily so that when its updated in base faf it doesnt break nomads so we dont have to immediately release a patch

but we DO want to eventually rename all NomadSomething to NomadsSomething
on top of that we want 2 versions of NomadUnit - NomadsUnit and NomadsUnit2 using the normal and new pixel shader respectively.

CookieNoob:
you could also use the new naming convention for the pixel shader: `NomadNormalMappedPS` -> `NomadsNormalMappedPS`

effects/mesh.fx
```diff
@@ -7727,6 +7781,960 @@ technique ShieldNomadStealth_LowFidelity
...
+    // A pass afterwards must write the depth buffer
+    pass P0
+    {
+        RasterizerState( Rasterizer_Cull_None )
+        AlphaState( AlphaBlend_SrcAlpha_InvSrcAlpha_Write_RGBA )
+        ZEnable = true;
+        ZWriteEnable = false;
+
+        VertexShader = compile vs_1_1 NOMADBuildVS();
+        PixelShader = compile ps_2_a NomadBuildHologramPS(true);
```
CookieNoob: why is the nomads technique using the nomad (no s) shaders?

aeoncleanse:
Or we could ensure that faf and nomads patches are pushed at the same time like sensible people, so this is totally not needed

CookieNoob:
By leaving the not renamed shader techniques in the mesh file, we are independent from the FAF patch cycle for the time until we renamed all shaders in nomads.

Having a wrong name for the technique for a single unit that's on the field will cause a hard game crash. A smooth transition (adding new techniques, then rename them in the units, then remove the old techniques) is the best way to do it.

aeoncleanse:
Would someone like to tell me why code dedicated to and used only by a mod is found in the primary game files?

CookieNoob:
as you probably saw, this code was added a loooonnng time ago by ZeP as it was intended to implement nomads into the main game as soon as the closed beta phase was over. It wasn't directly merged to allow for a faster update and patch cycle during development.

There are more parts in the code that are only there to allow for nomads support. For example the movies folder in the ProgramData/FAForever including the updater was added to make the nomads loading screen work (it was added by ZeP but broke/never worked).

one of the many mentions of nomads being implemented in main FAF: http://direct.faforever.com/2013/07/patch-3626-next/#.WOYTvSf6xFQ

aeoncleanse:
I'm gonna have to discuss which way to go about this in the next Dev conference. Personally I'd rather pull all Nomads code out of the base game, a mod is a mod.

CookieNoob:
sure, dont forget to make a PR to add all of this code to nomads then.

CookieNoob:
sure, dont forget to make a PR to add all of this code to nomads then.

It's probably unnecessary to remind you that FAF is also a mod, and some mod compatibility doesnt hurt. So if you remove the code, make sure nomads still works.

aeoncleanse:
Technically FAF is a mod. Logistically FAF has no more been a mod than the core game for some years now. 

Removing dedicated code not ever accessed by 99% of the userbase is not removing compatibility. Removing compatibility to block Nomads working would be removing compatibility.

Exotic-Retard:
removing shader changes from here would block nomads working. until you put a pr with the shader changes in nomads working from their own repo, that would be breaking compatibility. and yes i have already tried putting them in there but it didnt work, so its your turn.

same goes for all the other parts of nomads in faf repo. theyre not there just because we felt like it.

aeoncleanse:
They can be modded, you have a full sim hook. Just shadow the current repo if you must, but this code doesnt belong here. Im not suggesting a patch removing this be pushed before nomads is ready to or already has adapted

aeoncleanse:
I will merge this for now. I have also created a new issue: #1906

PR #1898, opened 2017-04-06
Manually filter for nearby overlapping structures for Mex Cap

aeoncleanse:
This _should_ fix the issue we've been having with the engine for some reason letting the mex cap storages overlap other structures. Needs THOROUGH testing!!!!!

### DIFF
--- a/lua/utilities.lua
+++ b/lua/utilities.lua
@@ -9,6 +9,56 @@ function GetDistanceBetweenTwoEntities(entity1, entity2)
     return VDist3(entity1:GetPosition(),entity2:GetPosition())
 end
 
+-- Function originally created to check if a Mass Storage can be queued in a location without overlapping
+function CanBuildInSpot(originUnit, unitId, pos)
+    local bp = __blueprints[unitId]
+    local mySkirtX = bp.Physics.SkirtSizeX / 2
+    local mySkirtZ = bp.Physics.SkirtSizeZ / 2
+
+    -- Find the distance between my skirt and the skirt of a potential Quantum Gateway
+    local xDiff = mySkirtX + 5 -- Using 5 because that's half the size of a Quantum Gateway, the largest stock structure
+    local zDiff = mySkirtZ + 5
+
+    -- Full extent of search rectangle
+    local x1 = pos.x - xDiff
+    local z1 = pos.z - zDiff
+    local x2 = pos.x + xDiff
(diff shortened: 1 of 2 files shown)


PhilipJFryFAF:
tested and doesn't fix the issue

[replay-id](replay.faforever.com/6031267)

TheKeyBlue:
I checked the code and that looks good to me.
I also did some ingame tests and was unable to reproduce the bug using the same methods I can currently use in the develop version.

lua/utilities.lua
```diff
@@ -9,6 +9,56 @@ function GetDistanceBetweenTwoEntities(entity1, entity2)
...
+            local zDist = math.abs(pos.z - structPos.z)
+
+            local skirtDiffx = mySkirtX + (structPhysics.SkirtSizeX / 2)
+            local skirtDiffz = mySkirtZ + (structPhysics.SkirtSizeZ / 2)
+
+            -- Check if the axis difference is smaller than the combined skirt distance
+            -- If it is, we overlap, and can't build here
+            if xDist < skirtDiffx and zDist < skirtDiffz then
+                allowed = false
+                break
```
HUSSAR-mtrela: you could just `return false` here

lua/utilities.lua
```diff
@@ -9,6 +9,56 @@ function GetDistanceBetweenTwoEntities(entity1, entity2)
...
+    local x1 = pos.x - xDiff
+    local z1 = pos.z - zDiff
+    local x2 = pos.x + xDiff
+    local z2 = pos.z + zDiff
+
+    -- Find all the units in that rectangle
+    local units = GetUnitsInRect(Rect(x1, z1, x2, z2))
+
+    -- Filter it down to structures only
+    units = EntityCategoryFilterDown(categories.STRUCTURE, units)
```
HUSSAR-mtrela: IMO, you should also check for units that can be build outside of factories. For example. building Megalith next to mex should prevent building of a mass storage. In other words filtered `units `table should contain `STRUCTURES` or `EXPERIMENTAL` units

aeoncleanse: That would be categories.STRUCTURE + categories.EXPERIMENTAL right ?

speed2CZ: yes

HUSSAR-mtrela: yes, that's correct.

PR #1903, opened 2017-04-09
Fix Aeon shields not rotating

### LINKED ISSUE #1902: Aeon Shield missing animation
The static aeon t2/t3 shield generator is missing an animation that is present in the steam version of the game
i assume this is not intended
[steam version](https://puu.sh/veJRg/61ba067cc4.gif)
[faf version](https://puu.sh/veJSb/a69b81c97f.gif)

speed2CZ:
Fixes #1902
As the open animations don't exists, this can be removed so the code can get tocreating the rotators (as I dont expect anyone to make the open animations).
It got broken by this commit https://github.com/FAForever/fa/commit/252ede50c378c40fc0bba848ccb32b81b9b256b6 which removed reference to non-existant animation from the BPs.

### DIFF
--- a/lua/aeonunits.lua
+++ b/lua/aeonunits.lua
@@ -190,27 +190,16 @@ AShieldStructureUnit = Class(ShieldStructureUnit) {
     OnShieldEnabled = function(self)
         ShieldStructureUnit.OnShieldEnabled(self)
         local bp = self:GetBlueprint()
-        local bpAnim = bp.Display.AnimationOpen
-        if not bpAnim then return end
-        if not self.OpenAnim then
-            self.OpenAnim = CreateAnimator(self)
-            self.OpenAnim:PlayAnim(bpAnim)
-            self.Trash:Add(self.OpenAnim)
-        end
         if not self.Rotator then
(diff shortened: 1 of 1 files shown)


aeoncleanse:
@speed2CZ Is anyone able to see where/how this stopped working? Clearly in Steam the code works...

MrNukealizer:
It stopped working in commit 252ede5 because it removed AnimationOpen from the blueprint. While the file didn't actually exist, removing that attribute caused this code to exit early instead of trying to create the missing animation then starting the rotation.

lua/aeonunits.lua
```diff
@@ -190,27 +190,16 @@ AShieldStructureUnit = Class(ShieldStructureUnit) {
     OnShieldEnabled = function(self)
         ShieldStructureUnit.OnShieldEnabled(self)
         local bp = self:GetBlueprint()
-        local bpAnim = bp.Display.AnimationOpen
-        if not bpAnim then return end
```
speed2CZ: @aeoncleanse the function returned here if the animation was in the blueprint, even if it was a string to non existant one. With that being removed from the blueprint, `bpAnim` is nil

PR #1907, opened 2017-04-12
Hotbuild Preset In Key Binding UI

HUSSAR-mtrela:
This PR makes the following changes:

- added a button for resetting keybinding to hotbuild key mapping in the Key Binding UI
- improved stats for category of keys (unbound and bound actions)
- improved arrangement of UI elements in Key Binding dialog
- improved logic for generating key binding table


HUSSAR-mtrela:
image
@aeoncleanse let me know what do you think about these changes.

aeoncleanse:
What happened to the other buttons?

HUSSAR-mtrela:
I moved them next to each action line (gray buttons with + and x symbols) so it is faster to assign and unbind key bindings.

PR #1912, opened 2017-04-14
Cleanup files

CookieNoob:
this changes all tabs into spaces and removes spaces at the end of lines.

@aeoncleanse started doing this for all files separately so this will shorten the process.


aeoncleanse:
You'll see from the commit history that I did it one file at a time. I did this to avoid polluting the commit history with a big commit like this. I proposed exactly this change a long time ago, but others pointed this issue out. Unfortunately, it means we have to keep going slowly, one file at a time :)

CookieNoob:
ok then, you have a lot of stuff to do :D 635 more files to go!

aeoncleanse:
Tell me about it lol

CookieNoob:
but tbh, I dont see how 637 separate commits polute the commit history less than one where all white spaces are replaced...

aeoncleanse:
It's also to do with polluting git blame. If you look back at most of the files, you eventually hit one commit where sheeo unified the line endings right when we took over. It blocks the blame history, and is really, really annoying.

CookieNoob:
git blame will be exactly the same if its one or many commits, since it will change the exact same lines.

aeoncleanse:
You know what, you're right. Lets get this done.

PR #1923, opened 2017-04-17
fix/ ships with death animation sinking on land

CookieNoob:
sometimes ships glitch on land and then get destroyed. When they have a death animation specified they start sinking through the ground.

This PR doesnt fix the glitch that they get on land, but it fixes the sinking.


aeoncleanse:
What's the glitch they get on land, is this just salem, and is it just this patch?

CookieNoob:
sometimes when you build a navy factory close to land and the ship rolls off on land, any ship can get on land this way.

But its better to check if a unit can sink before starting the sinking animation anyway.

CookieNoob:
sometimes it happens due to bad pathfinding as well...

PR #1938, opened 2017-04-24
Improve FR localization

speed2CZ:
Just a few things, but better than nothing


aeoncleanse:
d524bb53c4299b0889aa02bdf0732ddadf994d92

PR #1941, opened 2017-04-25
Add hiden tag for objectives

speed2CZ:
Additing to the bonus objective but works with all types.
Hidden objective doesnt show up in the UI, only announcement about
completing them.
Incomplete ones also dont show up on the score screen, instead of that
there's "Not Discovered" title for them.


speed2CZ:
@aeoncleanse it's ready

Issue #1934, opened 2017-04-24
Game needs 2 restarts to change the language


speed2CZ:
Changing language and clicking on Apply shows https://puu.sh/vuwMV/13eb2360eb.png
which closes the game but doesnt change the language yet.
It sets the language into profile table
```LUA
profile = {
    profiles = {
        {
            options = {
                selectedlanguage = 'CZ',
            },
        },
    },
}
```
After starting the game again, the language option is showing `CZ` as selected but the game is still in english.
Another restart does the job by setting:
```LUA
options_overrides = {
    language = 'CZ',
}
```

Uveso:
Looks like a job for me :)

aeoncleanse:
2d302b15c7a33a46edd54d0b7a0ff3e5402beab9

PR #1948, opened 2017-04-27
BugFix loading right language after gamerestart

Uveso:
fix for #1934

After changing the option for language and restarting the game, the new language will not appear.
That is caused by not saving options_overrides before gamerestart.

To solve this i modified the Localization.lua to get the actual language from useroptions instead of options_overrides.

### DIFF
--- a/lua/options/options.lua
+++ b/lua/options/options.lua
@@ -280,9 +280,6 @@ options = {
                 type = 'toggle',
                 restart = true,
                 default = __language,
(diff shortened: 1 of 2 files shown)


aeoncleanse:
@HUSSAR-mtrela - Please review this change. 

@Uveso - Great to have you back my friend :)

HUSSAR-mtrela:
@aeoncleanse this PR looks good 
@Uveso welcome back!

PR #1953, opened 2017-05-01
Fix/1777 hashmap iteration

### LINKED ISSUE #1777: ArmyIndex != StartPos leads to wrong scores reported
When a game is started, the host sends the game options for every player to the server. That includes the `army` index, which is just set to the `slotNum` number.

This is incorrect when there are free slots, because the engine always indexes armies starting from 1.

This means that when the game does score reporting based on the aibrain's `GetArmyIndex` function, it will report for the wrong army.

More details here: https://docs.google.com/document/d/1uuDYLmVisrN0GxWSJOWIjSlS1rPtWTtL5fTQL3OPVS4/edit

This almost certainly causes rating bugs.

It is actually hard to link players between lobby and game. The easiest way to associate them for the server is using the connection ID.

So we need to find a way to make sure this stuff is synchronized.

duk3luk3:
First commit adds logging for amy index calculation

Second commit fixes bug, maybe

Please deploy to fafdevelop for testing

Hopefully fixes #1777

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -5956,10 +5956,15 @@ function InitHostUtils()
         -- Call this function once just before game starts
         SendArmySettingsToServer = function()
             local armyIdx = 1
-            for slotNum, playerInfo in gameInfo.PlayerOptions do
-                LOG('>> Setting army '..armyIdx..' for slot '..slotNum)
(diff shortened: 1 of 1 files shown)


aeoncleanse:
19ef425fb3782838e2fb03c69087e26d38d129b2

PR #1954, opened 2017-05-01
BugFix loading right language after gamerestart II

Uveso:
fix for #1934

After changing the option for language and restarting the game, the new language will not appear.
That is caused by not saving options_overrides before gamerestart.

New language is now set inside userinit.lua


aeoncleanse:
f5d488979642a114b3fdef1fe9899b7b12dbcf21

PR #1975, opened 2017-05-12
test folding

IDragonfire:


IDragonfire:
ignore this ^^ only for testing

PR #1979, opened 2017-05-13
Add an option to have a factional UI font color

TheKeyBlue:
Can change it both in lobby (settings) and ingame (options)

Still needs translations, but i'll wait for some feedback regarding the text before asking for it.


lua/ui/uiutil.lua
```diff
@@ -245,7 +245,7 @@ function UpdateWorldBorderState(skin, isOn)
 end
 
 --* skin control, sets the current skin table
-function SetCurrentSkin(skin)
+function SetCurrentSkin(skin, overrideTable)
```
TheKeyBlue: I added this cause I couldn't get it to update ingame. 
(Due to set function from the option only happening before the prefs get updated)
So if somebody has a better way to do this, please share.

lua/ui/uiutil.lua
```diff
@@ -265,7 +265,11 @@ function SetCurrentSkin(skin)
     bodyColor:Set(skinTable.bodyColor)
     factionTextColor:Set(skinTable.factionTextColor)
     factionBackColor:Set(skinTable.factionBackColor)
-    fontColor:Set(skinTable.fontColor)
+    if ((overrideTable.faction_font_color == nil) and Prefs.GetOption('faction_font_color') ) or overrideTable.faction_font_color then
```
aeoncleanse: Extra brackets and spaces here...

lua/ui/lobby/lobby.lua
```diff
@@ -5123,6 +5123,17 @@ function ShowLobbyOptionsDialog()
         end
         RefreshLobbyBackground()
     end
+    --
```
aeoncleanse: Spare hyphens?

lua/options/options.lua
```diff
@@ -433,6 +433,21 @@ options = {
...
+                title = "<LOC OPTIONS_0213>Use Factional UI Font Color",
+                key = 'faction_font_color',
+                type = 'toggle',
+                default = false,
+                set = function(key,value,startup)
+                    import('/lua/ui/uiutil.lua').UpdateCurrentSkin({faction_font_color = value})
+                end,
+                custom = {
+                    states = {
+                        {text = "<LOC _On>", key = true },
```
aeoncleanse: Odd spaces here inside the {}

lua/ui/uiutil.lua
TheKeyBlue: it looks incredibly ugly, but couldn't think of a better way to do it.
Basicly it needs to use the override if there is one, and otherwise use the prefs.

lua/options/options.lua
TheKeyBlue: what spaces?

aeoncleanse: There's one after true before }

lua/ui/uiutil.lua
aeoncleanse: Logic is fine. I just mean that (overrideTable.faction_font_color == nil) doesn't need the brackets, and there's a space after colour')

lua/options/options.lua
TheKeyBlue: thats also the case in a bunch of the others. (so my confusion)
But i'll remove it.

lua/ui/uiutil.lua
TheKeyBlue: ah i thought you wanted me to add more brackets.
I'll remove this too then.

lua/options/options.lua
```diff
@@ -433,6 +433,21 @@ options = {
                 },
             },
             {
+                title = "<LOC OPTIONS_0213>Use Factional UI Font Color",
+                key = 'faction_font_color',
+                type = 'toggle',
+                default = false,
```
speed2CZ: I'd like to point out that as every feature we introduce with an option, we want to have it enabled by default with an option to disbale for crybabies, not the other way around, else no one will know about it.

lua/ui/lobby/lobby.lua
TheKeyBlue: well I added them since this was used to seperate the code for the other checkboxes.
In an attempt to keep it consistent.
Remove it anyway? (or even remove all of them?)

lua/options/options.lua
TheKeyBlue: Well since there was so much complaining and hate, I thought just having it off would be better.
But if thats the agreed policy, I'll put it on by default.

aeoncleanse: On by default please

TheKeyBlue:
Okay fixed the spacing,hyphen, ... issues and set it to default on.
I'll ask for translations now

TheKeyBlue:
Just added the translations I was able to get.
RU, CZ, DE and FR

TheKeyBlue:
Added ES too

PR #1980, opened 2017-05-13
Update loadingtips.lua

Fast-Thick-Pants:
I have proposed some changes, I feel that the instructional language used previously drew too much attention to the advice-giver, if we want this advice to feel integrated it must try to avoid breaking the immersion of the game. Basically I want to replicate the tone and language of an in-game tutorial rather than an internet forum with advice for players. I have also tried to include the reason with the advice to avoid condescension.

Blimey.


lua/ui/help/loadingtips.lua
```diff
@@ -1,38 +1,39 @@
...
     "<LOC loadingtip_0017>Rebuilding structures on their wrecks saves you 50% of construction cost and time",
-    "<LOC loadingtip_0018>An un-upgraded ACU is equivalent to 20 T1 tanks and 2 T1 Engineers",
-    "<LOC loadingtip_0019>Always build Tactical Missile Defense as soon as you have T2",
-    "<LOC loadingtip_0020>Hydrocarbon generator has the same power outcome as 5 T1 Power Generators but it’s much cheaper",
-    "<LOC loadingtip_0021>Aeon T3 Mobile Artillery can shoot while it’s moving",
-    "<LOC loadingtip_0022>Include reclaiming rocks around your start position into your build order",
+    "<LOC loadingtip_0018>An un-upgraded ACU has a combat potential equivalent to 20 T1 tanks",
+    "<LOC loadingtip_0019>It is sensible to build Tactical Missile Defense as soon as you have T2",
+    "<LOC loadingtip_0020>The Hydrocarbon's power production is equivalent to five T1 Power Generators but costs 58% less",
+    "<LOC loadingtip_0021>The Aeon T3 Mobile Artillery is called Serenity and is unique in being able to fire while moving",
```
TheKeyBlue: I don't think mentioning the units name is really necessary. Especially since it isn't really used in conversation much, in contrast to the common usage of the names of t1 tanks.

lua/ui/help/loadingtips.lua
```diff
@@ -1,38 +1,39 @@
...
-    "<LOC loadingtip_0011>Loyalists can deflect tactical missiles",
-    "<LOC loadingtip_0012>The Monkeylord is the cheapest land experimental unit",
-    "<LOC loadingtip_0013>Start building a second Paragon as soon as you finish your first one",
-    "<LOC loadingtip_0014>Place a move order before building a structure to make use of the build range",
-    "<LOC loadingtip_0015>The Seraphim destroyer is the only destroyer that can be submerged",
-    "<LOC loadingtip_0016>Tech 1 Mass Extractors only need a few seconds to pay back their investment",
+    "<LOC loadingtip_0001>The ACU’s Overcharge weapon uses 5000 energy requiring at least one Energy Storage",
+    "<LOC loadingtip_0002>Consider the placement of Energy Storage, it explodes violently if destroyed",
+    "<LOC loadingtip_0003>An ACU Overcharge deals 12 000 damage to units, 400 damage to an ACU and 800 damage to buildings",
+    "<LOC loadingtip_0004>Mass Fabricators are inefficent unless you have upgraded every available Mass Extractor to T3",
```
TheKeyBlue: This seems a little off? Like they suddenly become more efficient after getting all t3 mex.

lua/ui/help/loadingtips.lua
Fast-Thick-Pants: Well this was my point, this advice should sound like it is coming from the game right?
not from the players
In which case either the units are named or they are not

lua/ui/help/loadingtips.lua
Fast-Thick-Pants: Mass Fabricators are inefficient units and are best used only after upgrading every available Mass Extractor to T3.... bit long?

TheKeyBlue:
I'm liking the suggestions so far.
Feels a little less commanding and more friendly/gentle(?).

lua/ui/help/loadingtips.lua
aeoncleanse: Things shouldn't be named.

lua/ui/help/loadingtips.lua
aeoncleanse: "Mass Fabricators are inefficient compared with T3 Mass Extractors. Try to upgrade safe Mass Extractors before advancing to Fabricators"

lua/ui/help/loadingtips.lua
Fast-Thick-Pants: So that would include using T3 Siege Assault Bot for Loyalist and Experimental Spiderbot for Monkeylord, sounds fine.

lua/ui/help/loadingtips.lua
Fast-Thick-Pants: That is good. It is also, I think, the longest one so far, are there any kinds of limits to the length of these?

aeoncleanse: Not that I can think of, within reason.

lua/ui/help/loadingtips.lua
aeoncleanse: Correct

lua/ui/help/loadingtips.lua
speed2CZ: Well it's not using multiline text, so probably if it's too long it will just be cut off on smaller screens. Someone can test it.

lua/ui/help/loadingtips.lua
```diff
@@ -1,38 +1,39 @@
...
+    "<LOC loadingtip_0005>It is most efficent to add Mass Storages to T2 Mass Extractors before upgrading them to T3",
+    "<LOC loadingtip_0006>It is more efficent to upgrade a Mass Extractor to T2 before adding adjacent Mass Storages",
+    "<LOC loadingtip_0007>Submerged ACUs are harder to locate and kill, this is especially valuable in the late-game",
+    "<LOC loadingtip_0008>An un-assisted Nuke silo will take five minutes to build a missile, Anti-Nukes take four",
+    "<LOC loadingtip_0009>The attack-move command from a factory increases the engineers’ reclaim range",
+    "<LOC loadingtip_0010>Engineers given the attack-move command will reclaim along that path",
+    "<LOC loadingtip_0011>Cybran Loyalists can reprogram incoming tactical missiles and return them to the sender",
+    "<LOC loadingtip_0012>The Monkeylord is a relatively inexpensive experimental unit and can be built in a hurry",
```
Fast-Thick-Pants: This is a reply to Speed's point "in your version the main point seems to be that the mobile arty is called Serenity" Well why not have examples like this, we should remove the "Advice:" part and include fun facts about the game qoutes from Brackman or whomever says "The Gloves are coming off" Or lines from the Art of War, motivational lines, humour, things like that. It would help to give it that game feel.

lua/ui/help/loadingtips.lua
Fast-Thick-Pants: Well why not have examples like this, this isn't the most interesting but I think we should remove the "Advice:" part and include fun facts about the game qoutes from Brackman or whomever says "The Gloves are coming off" or lines from the Art of War, motivational lines, humour, things like that. It would help to give it that game feel. Like the Total War games do.

Fast-Thick-Pants: Some of these do get rather long, for example if we use Aeon Experimental Resource Generator instead of Paragon. What do you think aeoncleanse?

PR #1986, opened 2017-05-14
Lobby: fix issue with offline coop lobby breaking again.

TheKeyBlue:
#1973 fixed it for online, but in reality it just doesn't check the requiredType at all (only necessary in offline mode)

Couldn't replicate the way client launches the gamelobby with a certain map.
But the logic seems to be sound so I expect it to work.


aeoncleanse:
So... what's actually broken? And why?

TheKeyBlue:
Well there is a system in place with default maps for each mod (see uiutil.defaultScenario)
This was needed since coop uses different type of map (see uiutil.requiredType)
See https://github.com/FAForever/fa/pull/1865

So if you host a game on coop but had a normal map selected previously then lobby will break since coop can't handle a non-coop map.
So it was setup to use the defaultmap if the selected map's type doesn't match with the requiredtype.
But this was bugged so it always used the defaultmap instead. #1973 tried to fix it but instead just removed the feature completely. 

Now i'm back here with a version that should make offline switching between coop and faf work without issue, while working with the client hosting too.

MrNukealizer:
I'm confused what this change accomplishes. If desiredScenario is nil or empty it should work exactly the same. If it gives a scenario even when not launching through FAF, the references to the last scenario are pointless.

TheKeyBlue:
It does seem to always give a scenario even when you launch offline.
In your fix it was added to check for an empty desiredScenario before running that piece of code. 
But since this never happens the important check to see if the map is of required type never happens either. 

So it had to be taken out of the if again.

This part could be removed:
```
if not self.desiredScenario or self.desiredScenario == "" then
    self.desiredScenario = Prefs.GetFromCurrentProfile("LastScenario")
end
```
I didn't see the harm of that check, but its not really necessary probably.

aeoncleanse:
@MrNukealizer you happy with this now. I'm not merging until you two agree soooo... FIGHT! 🗡

MrNukealizer:
I haven't tested it, but it looks fine. My only real concern is if every mod requires the right map types so this doesn't accidentally break anything. It's their problem if they don't though.

TheKeyBlue:
Well test it then please :) This went wrong twice already due to bad testing xD
Well as far as i know only faf, coop and tutorials use different maps. And its setup properly for them already. 
Don't think there are any others?

MrNukealizer:
Unfortunately due to my setup I can't really test this.

PR #1987, opened 2017-05-14
remove old nomad shader

CookieNoob:
in nomads the shader techniques are now swiched to the new naming
convention "NomadUnit" -> "NomadsUnit" therefore the old techniques are
not needed anymore.

(this is part of the renaming of "nomad" -> "nomads" in the nomads mod,
as both names were used very inconsistently)


CookieNoob:
@Exotic-Retard @ThomasHiatt

Issue #1981, opened 2017-05-13
Unit Manager restricts Fatboy when No Air preset is enabled


HUSSAR-mtrela:
Reported by Masyaka
http://forums.faforever.com/viewtopic.php?f=2&t=14550

No Air preset should exclude (LAND * AIRSTAGINGPLATFORM) units

MrNukealizer:
Also check Cybran cruiser, Atlantis, and possibly carriers.

MrNukealizer:
I looked at some of the restriction presets and it seems the problems are a lot more extensive than just this. It seems like the the restriction category system could use some work. Some examples:
* Restricting amphibious units doesn't affect Fatboy or Scathis and does restrict Harbinger Mark IV.
* Restricting air does restrict the Fatboy as noted, but it's inconsistent in other ways too. It restricts the Cybran T3 mobile AA which can attack air and land, but not the Seraphim version which can also attack air and land. It also allows the Shard while restricting all other AA only units.
* Restricting stealth aircraft only affects the Cybran spy plane, not the ASF or strategic bomber.
* Restricting stealth boats affects UEF and Cybran frigates but not the actual stealth naval units like the Mermaid, Baracuda, and Plan B.
* Restricting mobile stealth affects Rhino, Fire Beetle, Loyalist, and Brick. The description also seems to apply only to stealth fields but it affects personal stealth.
* Restricting air snipe units affects T2 torpedo bombers but not T3. and T2/T3 gunships but not T1.
* Restricting AA navy affects the Cooper, Mermaid, and Cybran frigate.
* Restricting torpedo boats affects Aeon and Seraphim frigates, does not affect Cooper, affects only the Cybran destroyer, and does not affect the Cybran attack sub.
* Restricting torpedo land units does literally nothing even though Fatboy, Megalith, Monkeylord, and Brick are land units with torpedoes. 
* Direct fire land includes the Seraphim T3 mobile AA but not the Cybran version.
* Restricting mass extractors also restricts mass storage for some reason.

HUSSAR-mtrela:
I added my comment above (in bold text). Note that restriction system could be improved by checking type of weapon units have because categories in blueprints are not always reliable to determine that. Although, I think that blueprints of these units should be corrected because the game engine is also using them for other reason.

MrNukealizer:
A lot of the comments make sense, but a couple don't.
[...]

PR #1988, opened 2017-05-14
fix fatty being restricted by no-air-restriction

PhilipJFryFAF:


PhilipJFryFAF:
not 100% sure if this is the best way to fix the issue maybe putting a unique category for the fatty in the bp file and excluding him like that would be better...

PhilipJFryFAF:
Never mind just double checked and without the PR the no-air restriction affects 106 units and with the PR it affects 105 (the one difference being the Fatboy).

lua/ui/lobby/UnitsRestrictions.lua
```diff
@@ -78,7 +78,7 @@ Expressions = {
     -- added exclusion of engineers and structures because they are restricted by other presets
     LAND        = "(LAND - ENGINEER - STRUCTURE)",
     -- added restriction of AA structures because they are not needed when all air units are restricted
-    AIR         = "((STRUCTURE * (ANTIAIR + AIRSTAGINGPLATFORM)) + (AIR - POD - SATELLITE) + (LAND * ANTIAIR - DIRECTFIRE) + (TECH3 * NAVAL * CARRIER * AIRSTAGINGPLATFORM))",
+    AIR         = "((STRUCTURE * (ANTIAIR + AIRSTAGINGPLATFORM)) + (AIR - POD - SATELLITE) + (LAND * ANTIAIR - DIRECTFIRE - INDIRECTFIRE) + (TECH3 * NAVAL * CARRIER * AIRSTAGINGPLATFORM))",
```
HUSSAR-mtrela: nice try of fixing it but there is actually a better way of fixing it and I will submit PR that also fixes some other issues mentioned in #1981

MrNukealizer:
It still seems like a good change in case any mods add units with ANTIAIR and INDIRECTFIRE but not DIRECTFIRE.

HUSSAR-mtrela:
@aeoncleanse this PR can be close because I have better fix in PR #1990

PhilipJFryFAF:
thanks Hussar i'll close the PR

PR #1990, opened 2017-05-15
Unit restrictions Hotfixes

HUSSAR-mtrela:
Fixes a few items reported in issue #1981 

**1 -** Restricting air does restrict the Fatboy as noted, but it's inconsistent in other ways too. It restricts the Cybran T3 mobile AA which can attack air and land, but not the Seraphim version which can also attack air and land. It also allows the Shard while restricting all other AA only units.
- [x] include Seraphim T3 mobile AA in the restriction

**2 -** Restricting amphibious units doesn't affect Fatboy or Scathis and does restrict Harbinger Mark IV.
- [x] include Fatboy and Scathis in the restriction and exclude Harbinger

**5 -** Restricting mobile stealth affects Rhino, Fire Beetle, Loyalist, and Brick. The description also seems to apply only to stealth fields but it affects personal stealth.
**BLUEPRINTS for these units should not have COUNTERINTELLIGENCE category**
- [x] update description with personal stealth**

**6 -** Restricting air snipe units affects T2 torpedo bombers but not T3. and T2/T3 gunships but not T1.
- [x] include T3 bombers in the restriction, T1 gunships are excluded by design


PhilipJFryFAF:
you might want to take a look at nomads too since they have some multi purpose units like mobile arty with AA

HUSSAR-mtrela:
Yes, I will and nomands mod should just re-use these unit restriction instead of replacing them.

CookieNoob:
nomads is just using the file from FA, the unitmanager and unitrestrictions files were removed in the last nomads upgrade.

aeoncleanse:
81c36820094d9f4f99f2921a646f4a2dc5a72063

PR #1993, opened 2017-05-15
minor hotbuild fixes

PhilipJFryFAF:
puts the energy storage in on the same list of buildings as power generators in the hotbuild unit group
saves you one additional hotkey to bind


MrNukealizer:
I can see how that could be nice, but it also seems like it might confuse people into making energy storage instead of T1 pgens...

TheKeyBlue:
I don't think that's a real concern, but I can't be sure.

I like it, Now I don't need to learn another key.

PhilipJFryFAF:
i think it fits and since you have to cycle through to get to the storage it shouldn't cause too many issues if any

PhilipJFryFAF:
the select next idle air scout function did select every single air scout instead of just the next

PhilipJFryFAF:
changed scout to intelligence so that idle spy planes will be selected and not just scout planes

Issue #2003, opened 2017-05-19
switching from player to observer lose data (veterancy) 


shalkya:
there is a lose of UI data, when switching from player to observer while looking at a replay.
here it is veterancy : wrong regen is showing + wrong number of stars
https://files.slack.com/files-pri/T02T320UA-F5FFETUAF/image.png
https://files.slack.com/files-pri/T02T320UA-F5FK1GN3W/image.png

aeoncleanse:
Fixed in #2004

PR #2004, opened 2017-05-19
Fix observers not seeing correct unit regen

### LINKED ISSUE #1676: Wrong ACU regen for observers
explained with pictures here: http://forums.faforever.com/viewtopic.php?f=3&t=13462

MrNukealizer:
Fixes #1676, #2003


aeoncleanse:
https://github.com/FAForever/fa/issues/1676

Did this actually fix this one then?

MrNukealizer:
Yes, it fixes that bug. I don't know what you're asking about how the sync table gets the data. I literally changed the code that puts the data in the sync table.

PR #2014, opened 2017-05-20
Make loading tips more readable

Crotalus:
Add dropshadow so the loading tips text gets more constrast compared to the background

### DIFF
--- a/lua/ui/game/gamemain.lua
+++ b/lua/ui/game/gamemain.lua
@@ -280,14 +280,15 @@ local function LoadDialog(parent)
     local text = '::  ' .. LOC('<LOC LOAD_0000>IN TRANSIT') .. '  ::'
     local textControl = UIUtil.CreateText(movie, text, 20, UIUtil.bodyFont)
     textControl:SetColor(color)
+    textControl:SetDropShadow(true)
     LayoutHelpers.AtCenterIn(textControl, parent, 200)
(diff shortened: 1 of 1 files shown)


lua/ui/game/gamemain.lua
```diff
@@ -385,6 +386,7 @@ function CreateWldUIProvider()
         local text = '::  ' .. LOC('<LOC LOAD_0000>IN TRANSIT') .. '  ::'
         local textControl = UIUtil.CreateText(background, text, 20, UIUtil.bodyFont)
         textControl:SetColor(color)
+        textControl:SetDropShadow(true)
```
speed2CZ: This one just flashes for a split of the second when the loading is done

Crotalus: Good catch, added it via github's web editor. Should probably be squashed into one commit (not at home atm)

micheljung:
And are they still flashing? If so, I think that should be removed, too :)

HUSSAR-mtrela:
why dont you use colors from UIUtil which are also used in existing UI elements?
```
textControl:SetColor(UIUtil.bodyColor) 
textControl:SetColor(UIUtil.tooltipTitleColor) 
textControl:SetColor(UIUtil.fontColor) 
```

speed2CZ:
@micheljung well IN TRANSIT is flashing
so I made the tips flash as well cuz it looked bad if one was flashing and second didnt.
Maybe slow donw your PC and you'll have more than enough time to read it even backwards!

PR #2018, opened 2017-05-21
increase harb buildrate

JaggedAppliance:
Increase buildrate to compensate for the global reclaim speed nerf.


aeoncleanse:
2a34a57fefbaf67f3b39821da5bf84371d682304

Issue #1982, opened 2017-05-13
Pings are inconsistent and unreliable and possibly abusable


TheKeyBlue:
I've heard about people not being able to ping or their pingmarkers not appearing on the map during games.
This is obviously annoying as you don't want to spend time trying to write the same marker multiple times.

It happened in giebs stream:
https://www.twitch.tv/videos/142915443 at 03:08:08

I'm thinking the Ping lock could have something to do with it.
https://github.com/FAForever/fa/blob/eb7c51f0ff647dbed415c61d4fd6e90f9ae6e985/lua/SimPing.lua#L21
It even blocks pings of your opponents I think. (since its called from a Simcallback every player gets it)
So in theory you could block your opponents pinging possibility by spamming pings yourself.
(needs verification though)

(sim and ui since its based on SimCallbacks and UserSyncs)

TheKeyBlue:
Okay I just did some ingame test with somebody spamming pings.
If both are spamming it looks like its quite random who's ping actually happens. 

This is especially bad for markers since they require substantial time investment.
I suggest we just make a cooldown for each player individually instead. 
So you can ping without getting fucked by ally or enemy pinging aswell.

aeoncleanse:
That seems sensible. A per-player cooldown, 2 seconds?

Exotic-Retard:
if possible can it be something like you get 3 pings every 15 seconds because sometimes you want to ping twice for two places or for sth very important

MrNukealizer:
I have a PR in that changes the limit to 3 pings every 15 seconds plus one additional marker, also with a cooldown. My first impression is that 15 seconds feels a bit long when spamming pings, but I'm sure the player's allies would be glad the limit is lower than one a second...

PR #2021, opened 2017-05-22
Change ping lockout to be per player instead of global

MrNukealizer:
Addresses #1982 by changing the limit to be per player. It allows 3 pings with a 15 second cooldown. It also allows an additional marker after reaching the limit, to be a little more accommodating if someone takes the time to type out a message. The marker still adds to the cooldown to prevent low effort marker spam.

### DIFF
--- a/lua/SimPing.lua
+++ b/lua/SimPing.lua
@@ -1,6 +1,9 @@
 local SUtils = import('/lua/ai/sorianutilities.lua')
 
-local PingLocked = false
+local PingLimit = 3 -- Maximum number of pings per army, within PingTimeout seconds
+local ExtraMarkers = 1 -- How many extra markers an army can make beyond the normal ping limit
+local PingTimeout = 15 -- How long a ping counts toward an army's limit
+local PingsRemaining = {}
 local PingMarkers = {}
 MaxPingMarkers = 15
 --On first ping, send data to the user layer telling it the maximum allowable markers per army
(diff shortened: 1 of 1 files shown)


TheKeyBlue:
Yeah 3 / 15s sounds way too long. Just imagine people quikly pinging 3 location to alert somebody of something and then having to wait 15s while being unable to alert team members about incoming snipes.
I think just keeping the 1 / 1s ping would be best at first, atleast for testing on develop.
Then if ping spam by all players is an issue we can increase the cooldown.

MrNukealizer:
I was following @Exotic-Retard's suggestion, but I agree it might need some refinement. What situation comes to mind where you'd need to quickly ping 3 locations in quick succession?

aeoncleanse:
Responding to enemy joint attacks. Every 5s will be fine.

CookieNoob:
I think 3/15sec is not enough. In the rare cases that someone scouts in a teamgame, he usually pings everything important that he can see. For example started nukes/experimentals/stratbomber... or just finished experimentals (might get more than one ping). Being limited to 3 pings would mean that you run out of pings quickly there. So I agree with @TheKeyBlue 1 / 1s is good for now.

also in the case of 3 / 15s you need to communicate it in a very good way or you will see lots of posts in tech support "ping feature broken" and a way to display the cooldown ingame (that could be done similar to the nomad intelprobe UI element)

PR #2022, opened 2017-05-22
Fix missing reference in Ythotha script

MrNukealizer:
Removing this causes an error in StartBeingBuiltEffects where EffectUtil is used.


aeoncleanse:
Damn, sorry about this, but this one's already covered in deploy/fafdevelop 3ff2030283377f39f620b2817a339a19c79c98ff

PR #2025, opened 2017-05-24
Base manager improvements

### LINKED ISSUE #1417: Coop: ACU in BaseManager doesn't OC, Upgrade
This code that sets platoon for ACU should add something like overcharging and upgrading.
https://github.com/FAForever/fa/blob/develop/lua/AI/OpAI/BaseManager.lua#L1476

Instead the only thing that ACU does is follows
`PlatoonAIFunction = {'/lua/ai/opai/BaseManagerPlatoonThreads.lua', 'BaseManagerEngineerPlatoonSplit'},`

Is there anyone who knows how excatly `PlatoonAddFunctions` and `PlatoonAIFunction` work?

When I commented out `PlatoonAIFunction` I made ACU upgrade (with few fixes to the `UnitUpgradeBehavior` function) 
The thing is that if `PlatoonAIFunction` is set, then niether `UnitUpgradeBehavior` or `CDROverchargeBehavior` is called

speed2CZ:
Work in progress.
#JustCoopThings
- [x] Loads default scouting chain if they exist
- [x] Basic TML control
* Base manager can now operate TML launchers, even if they are rebuilt/newly built.
* They don't need to be activated via script anymore.
* Enabled by default, can be disabled via `bm:SetActive('TML', false)`
* On `Hard` difficulty they use sorian function for predicting mobile target location
- [x] Fixed issue with engineers trying to assist factories where they are non in the base
* Engineers wont initiate permanent assist on factories if there are no factories in the base
* Extra check in the perma assist function to not assist factory if there isnt one, the function should no longer end up with an error in that case.
- [x] Fixed ACU/sACU upgrading not working
* Upgrade thread is not properly set on the command units. If there are upgrades set in the base manager via `bm:SetUnitUpgrades` or `bm:SetACUUpgrades` or `bm:SetACUUpgrades` they will get upgraded.
- [x] Mercy Platoon for OpAI
- [x] Add new OpAI Naval Platoons
- [x] Fixes for OpAI Land Platoons
* Fixed wrong order of the units/children in the platoon.
* Added few new platoon templates
- [x] Fixed OpAI Childer check
* The Platoon build will now correctly keep only platoons that have all children
* Example: Desired platoon with children types `{'LightTanks', 'HeavyTanks'}` used to keep platoons with just 1 childer of that type, so platoon that has just `'LightTanks'` or `'HeavyTanks'` was kept enabled.
- [x] More control over OpAI platoon size
- [x] Units of all human armies are properly set invincible during cinematics

Fixes #1417

### DIFF
--- a/lua/AI/OpAI/BaseManagerPlatoonThreads.lua
+++ b/lua/AI/OpAI/BaseManagerPlatoonThreads.lua
@@ -1247,25 +1247,14 @@ function UnitUpgradeThread(unit)
         unitType = 'DefaultSACU'
     end
 
-    local pool = aiBrain:GetPlatoonUniquelyNamed('ArmyPool')
     while not unit:IsDead() do
         if not bManager then
             bManager = aiBrain.BaseManagers[unit.PlatoonData.BaseName]
         end
         if bManager then
-            -- See if unit is in the pool
-            local found = false
-            for k, v in pool:GetPlatoonUnits() do
-                if v == unit then
-                    found = true
-                    break
-                end
-            end
-
-            -- If its in the pool and needs an upgrade
             local upgradeName = bManager:UnitNeedsUpgrade(unit, unitType)
 
-            if found and upgradeName then
+            if upgradeName and not unit:IsUnitState('Building') then
                 local platoon = aiBrain:MakePlatoon('', '')
                 aiBrain:AssignUnitsToPlatoon(platoon, {unit}, 'support', 'none')
 
@@ -1273,6 +1262,8 @@ function UnitUpgradeThread(unit)
                     TaskName = "EnhanceTask",
                     Enhancement = upgradeName
                 }
+                IssueStop({unit})
+                IssueClearCommands({unit})
                 IssueScript({unit}, order)
 
                 repeat
--- a/lua/AI/OpAI/BaseManagerPlatoonThreads.lua
+++ b/lua/AI/OpAI/BaseManagerPlatoonThreads.lua
@@ -1151,6 +1151,9 @@ function BaseManagerTMLAI(platoon)
     end
 end
 
+function BaseManagerNukeAI(platoon)
+end
+
 function AMUnlockBuildTimer(platoon)
     ForkThread(AMPlatoonHelperFunctions.UnlockTimer, platoon.PlatoonData.LockTimer, platoon.PlatoonData.PlatoonName)
 end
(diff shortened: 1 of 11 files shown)


lua/AI/OpAI/BaseManagerPlatoonThreads.lua
```diff
@@ -1084,6 +1085,72 @@ function BaseManagerScoutingAI(platoon)
...
+        categories.EXPERIMENTAL,
+        categories.ENERGYPRODUCTION,
+        categories.STRUCTURE,
+        categories.TECH3 * categories.MOBILE})
+
+    while aiBrain:PlatoonExists(platoon) do
+        if BMBC.TMLsEnabled(aiBrain, baseName) then
+            local target = false
+            local blip = false
+            while unit:GetTacticalSiloAmmoCount() < 1 or not target do
```
TheKeyBlue: Shoulnd't you only look for a target if your silo is filled?

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
```diff
@@ -1084,6 +1085,72 @@ function BaseManagerScoutingAI(platoon)
...
+                WaitSeconds(7)
+                target = false
+                while not target do
+                    target = platoon:FindPrioritizedUnit('Attack', 'Enemy', true, unit:GetPosition(), maxRadius)
+
+                    if target then
+                        break
+                    end
+
+                    WaitSeconds(3)
```
TheKeyBlue: if you put this at the start of the while, you won't need the 'if .. then break' statement, right?
(and compensate by setting the waitseconds(7) before this while to waitseconds(4)

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
```diff
@@ -1084,6 +1085,72 @@ function BaseManagerScoutingAI(platoon)
...
+
+                    if not aiBrain:PlatoonExists(platoon) then
```
TheKeyBlue: shouldn't you test for existence before finding the target?

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
speed2CZ: This is actually just copy pasted TacticalAI from platoon.lua with only addition of shooting in front of moving units on hard diff. I know it works, but not saying it can't be improved. I'll look into that later.

speed2CZ:
@aeoncleanse or @ChrisKitching I think Im done with this

aeoncleanse:
I'll look it over this weekend, unless Chris decides to rip it apart ;)

speed2CZ:
Nothing to rip just merge :P its coop only

PR #2027, opened 2017-05-25
Notify: Finished support for disabling custom messages and fixed some bugs

MrNukealizer:


lua/ui/notify/notify.lua
```diff
@@ -33,7 +33,7 @@ function setupStartDisables()
     for key, data in messages do
         local category = key
         if factions[category] then -- Don't make a disabler per-faction
-            category = 'acus'
```
aeoncleanse: Not sure what you're doing here mate. We definitely need this, as explained in the comment.

lua/ui/notify/notify.lua
```diff
@@ -200,14 +228,13 @@ end
...
 function onCompletedEnhancement(id, category, source)
-    local msg = {to = 'allies', Notify = true, category = category, text = messages[category][source] .. ' done!'}
+    local msg = {to = 'allies', Chat = true, Notify = true, text = messages[category][source] .. ' done!', data = {category = category, source = source, trigger = 'completed'}}
 
     if id then
         local data = ACUs[id]
         if data then
             msg.text = messages[category][source] .. ' done! (' .. round(GetGameTimeSeconds() - data.startTime, 2) .. 's)'
             killWatcher(data)
-            NotifyOverlay.sendDestroyOverlayMessage(id)
```
aeoncleanse: We don't need this any more??

lua/ui/notify/defaultmessages.lua
```diff
@@ -87,12 +87,12 @@ defaultMessages = {
         xsl0401 = "Ythotha",
     },
     tech = {
-        researchlandtech2 = "Tech 2 Land HQ upgrade",
-        researchlandtech3 = "Tech 3 Land HQ upgrade",
-        researchairtech2 = "Tech 2 Air HQ upgrade",
-        researchairtech3 = "Tech 3 Air HQ upgrade",
-        researchnavaltech2 = "Tech 2 Naval HQ upgrade",
-        researchnavaltech3 = "Tech 3 Naval HQ upgrade",
+        researchLANDTECH2 = "Tech 2 Land HQ upgrade",
```
aeoncleanse: Sources should all be lower case. Just trying to keep things consistent.

lua/ui/notify/notify.lua
MrNukealizer: Oops, I misread something. There's no need to make the acus setting 4-5 times, but I mistakenly thought the acus category actually existed on its own.

lua/ui/notify/notify.lua
MrNukealizer: I moved it into killWatcher with the more closely related stuff.

lua/ui/notify/defaultmessages.lua
MrNukealizer: In that case something needs to convert it to lowercase elsewhere, because it currently uses this as the key.

PR #2034, opened 2017-05-26
GameUI: put silo missile count and max size on top of order button

### LINKED ISSUE #2033: can't see how much a nuke or a tml can store because of shortcut
http://imgur.com/4ltmhk7
http://imgur.com/ZwzK6zt

here are the screenshot, you can still see this in different way, but i guess it's not how we wanted that to happen.

TheKeyBlue:
Currently hotkeylabel covers silo size, fixes #2033
missiles


aeoncleanse:
Hmm... If it is possible, could we have it in the normal position if labels are off, and this position if they're on, with proper live updating etc?

TheKeyBlue:
yeah i think so. It gets completely remade everytime you select a new unit as far as i know.
So i could add a condition based on the hotkeylabel settings.
I'll try adding it tomorrow.

TheKeyBlue:
Labels are always on? Couldn't find anything that suggested you could.
So now its only moved to the top if there is a key to display.

TheKeyBlue:
Only needs proper loc and maybe a better title and description. 
(didn't copy/paste title and description in locs yet incase of improved text)

TheKeyBlue:
Not sure why this got merged?
I mentioned in the last comment that I still needed to adjust text and LOC..
I'll just make a new PR with those changes then?

aeoncleanse:
Yup, sure :)

PR #2039, opened 2017-05-28
Various Hotbuild fixes

MrNukealizer:
- [X] Fix the upgrade key not working for engineering stations and mod structures.
- [X] Fix the key for T3 torp bombers and transports doing nothing because it was trying to upgrade.
- [X] Add a keybinding for T3 mobile AA.
- [X] Fix upgrade key label always going on HQ when the key actually upgrades to support factory.
- [X] Add keybinding entries for support factories.


lua/keymap/hotbuild.lua
```diff
@@ -479,24 +477,19 @@ end
...
     local selectedUnits = GetSelectedUnits()
     local availableOrders, availableToggles, buildableCategories = GetUnitCommandData(selectedUnits)
-    local bpTypes = {}
+    local result = true
 
     for index, unit in selectedUnits do
-        local bpId = unit:GetBlueprint().BlueprintId
-        local cmd = upgradeTab[bpId]
+        local bp = unit:GetBlueprint()
+        local cmd = bp.General.UpgradesTo
```
CookieNoob: are you sure this wont break upgrades of factories to support/HQ factories?

Currently a t1 factory upgrades to a support factory as long as there is an HQ factory.

MrNukealizer: Ok, it actually does break that. If that functionality is to be preserved I'll also need to change the hotkey label code, because it puts the upgrade key label on the HQ, yet pressing the key builds the support factory.

lua/ui/game/gamemain.lua
```diff
@@ -477,6 +478,9 @@ function OnSelectionChanged(oldSelection, newSelection, added, removed)
     if not hotkeyLabelsOnSelectionChanged then
         hotkeyLabelsOnSelectionChanged = import('/lua/keymap/hotkeylabels.lua').onSelectionChanged
     end
+    if not upgradeTab then
+        upgradeTab = import('/lua/keymap/upgradeTab.lua').upgradeTab
```
TheKeyBlue: What is the goal/point of having it initialize here and not just have it be initialized from the start?
Some kind of efficiency improvement?

MrNukealizer: Many, many files import /lua/ui/game/gamemain.lua. At the moment there's no real need to avoid importing upgradeTab in the main section, but it goes with the other file imported right above it, and circular dependencies are all too common with gamemain.lua.

TheKeyBlue: ah I see. I guess that makes sense.

PR #2043, opened 2017-05-29
Make sure ShareOption is set to a proper value

### LINKED ISSUE #2042: Invalid shareconditions still causes issues.
Games with invalid shareconditions should work as shareUntilDeath, but due to some bug all shared units stay shared.
The issue is caused in this part of the code: https://github.com/FAForever/fa/blob/develop/lua/aibrain.lua#L649

TheKeyBlue:
To avoid issues with invalid ShareOption later on in the game. Fixes #2042

### DIFF
--- a/lua/simInit.lua
+++ b/lua/simInit.lua
@@ -57,6 +57,22 @@ function SetupSession()
     -- from the scenario script.
     ScenarioInfo.Env = import('/lua/scenarioEnvironment.lua')
 
+    --Check if ShareOption is valid, and if not then set it to ShareUntilDeath
+    local shareOption = ScenarioInfo.Options.Share
+    local globalOptions = import('/lua/ui/lobby/lobbyOptions.lua').globalOpts
+    local shareOptions = {}
+    for _,globalOption in globalOptions do
+        if globalOption.key == 'Share' then
+            for _,value in globalOption.values do
+                shareOptions[value.key] = true
+            end
+            break
+        end
+    end
+    if not shareOptions[shareOption] then
+        ScenarioInfo.Options.Share = 'ShareUntilDeath'
+    end
+
     -- if build/enhancement restrictions chosen, set them up
     local buildRestrictions, enhRestrictions = nil, {}
 


lua/simInit.lua
```diff
@@ -57,6 +57,23 @@ function SetupSession()
...
     ScenarioInfo.Env = import('/lua/scenarioEnvironment.lua')
 
+    --Check if ShareOption is valid, and if not then set it to ShareUntilDeath
+    local shareOption = ScenarioInfo.Options.Share
+    local globalOptions = import('/lua/ui/lobby/lobbyOptions.lua').globalOpts
+    local shareOptions = {}
+    for _,globalOption in globalOptions do
+        if globalOption.key == 'Share' then
+            for _,value in globalOption.values do
+                WARN(value.key)
```
aeoncleanse: Probably don't need this WARN. Otherwise looks good.

TheKeyBlue: woops i thought i removed it all

Uveso:
I am sorry, but i don't understand this.

First, i can't reproduce a wrong sharecondition. Even if i change Game.prefs before gamestart and while the game is running.
`ScenarioInfo.Options.Share` will allways fall back to optionsdefault `ShareUntilDeath`.

And if i force to change ScenarioInfo.Options.Share to an unknown sharecondition, it will be caught by the failsave here: https://github.com/FAForever/fa/blob/develop/lua/aibrain.lua#L663

Even if we force an invalid sharecondition saved inside `ScenarioInfo.Options.Share`, it will fall back to 'ShareUntilDeath' or catched inside aibrain.lua.

I really don't understand how this fix has any effect.

And how do you know, that ScenarioInfo.Options.Share will store sometimes a wrong sharecondition ?

How can i test/reproduce it ?

TheKeyBlue:
well the failsafe doesn't work properly actually. Having `GetBackUnits(BrainCategories.Allies)` before the `else` fucks it up apparantly.
And then icedreamer suggested we should try to make sure its a correct shareoption at the beginning of the game to avoid having to deal with any other possible bugs or problems there could be concerning shareoption.

I can't really reproduce it either, now. :/ I thought I could.
We know it can store a wrong value cause there are replays with a wrong value: 
https://faforever.slack.com/files/keyblue/F5KU6F9HU/6279043.fafreplay

I just applied my changes to that the code and ran the replay. And it behaved like it should.
(desync when the units died like they should've instead of surviving)


Maybe its better if I also fix the bug with the GetBackUnits() too.

aeoncleanse:
There is no bug with getbackunits. The error is in my logic when i wrote that code, i didnt anticipate properly the behaviour of nonstandard modes.

Uveso:
Ah i see, thanks for the explanation!

PR #2049, opened 2017-05-31
Only show reclaim in the playable area

### LINKED ISSUE #2032: Show reclaim labels only in the playable area
I think there is no need to show reclaim that is outside of the playable area. I guess this mainly affects coop as custom maps dont change the are that often.

https://puu.sh/w1Lzp/adfc480634.png

TheKeyBlue:
fixes #2032

I couldn't find anyway to get playable area from UI, so I made a new Sync for UserSync.
If anyone has a better way to do it, then let me know.

reclaimoutside
playableareaonly

### DIFF
--- a/lua/UserSync.lua
+++ b/lua/UserSync.lua
@@ -10,8 +10,10 @@ PreviousSync = {}
 -- the Sync.UnitData table into this table each sync (if there's new data)
 UnitData = {}
 
-local UpdateReclaim = import('/lua/ui/game/reclaim.lua').UpdateReclaim
+local reclaim = import('/lua/ui/game/reclaim.lua')
+local UpdateReclaim = reclaim.UpdateReclaim
 local sendEnhancementMessage = import('/lua/ui/notify/notify.lua').sendEnhancementMessage
+local SetPlayableArea = reclaim.SetPlayableArea
 
 -- Here's an opportunity for user side script to examine the Sync table for the new tick
 function OnSync()
@@ -91,5 +93,9 @@ function OnSync()
         for _, messageTable in Sync.EnhanceMessage do
             sendEnhancementMessage(messageTable)
         end
+    end
+    
+    if Sync.NewPlayableArea then
+        SetPlayableArea(Sync.NewPlayableArea)
     end
 end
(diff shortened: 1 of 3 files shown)


lua/UserSync.lua
```diff
@@ -92,4 +94,8 @@ function OnSync()
             sendEnhancementMessage(messageTable)
         end
     end
+    
+    if Sync.NewPlayableArea then
+        SetPlayableArea(Sync.NewPlayableArea)
```
MrNukealizer: This can be done in /lua/usercamera.lua line 28 without the need for any new sync data.

TheKeyBlue: Mmh, I would say this is better.
But if people prefer to use usercamer.lua, then I will.

aeoncleanse:
Pretty sur eonly on-map data is even being sent from sim-side... Please check the part where it's sent to Sync.

TheKeyBlue:
? its not about on map data. It's about in playable area data. 
As you can see in the pics i've posted.

And since playable area can change, i assumed it would be better to just have the ui choose which ones to display, instead of having the sim send removal messages whenever a reclaim suddenly falls outside the playable area, or vice versa. Which would be more complicated and bug prone.

speed2CZ:
Yes, current solution is good

aeoncleanse:
OK having looked at this more closely I see what's trying to be done. I am very concerned about the fact that we're running the position check every marker, every time. When developing this we found that odd things could severely affect smoothness and FPS. If possible, it might be better to do it in camera...

TheKeyBlue:
How should we do it in camera?
We were talking about getting the playable area from the camera. Instead of with `Sync.NewPlayableArea`

aeoncleanse:
Yeah, that. Just whatever nuke said.

MrNukealizer:
What I said was only about getting the map bounds without adding anything more to the Sync table. That wouldn't affect the performance of the position check at all.

TheKeyBlue:
I should probably change the check to do it when the reclaim gets synced and when a new playable area gets synced. 
Then it only happens once for each reclaim creation, and once for each map expansion.

MrNukealizer:
Yeah, that's a good idea. That should prevent any performance concerns.

TheKeyBlue:
Alright I did some changes.
Now the playable area check only happens when the area changes and when the reclaim gets updated.
It seperates the visible from the invisible reclaim.
Now there is no check necessary when the labels are created.

PR #2063, opened 2017-06-08
Fix overcharge queuing

### LINKED ISSUE #2057: can't queue up several OC
since now auto OC is 5 sec, and normal one is 3.3 sec; we would expect queuing OC would allow to to shoot at every 3.3 sec (more or less). 
But despite having the order queued up, the ACU fire only 1 shoot.

MrNukealizer:
This fixes #2057. It works correctly in my tests but something doesn't feel quite right, so some public testing might be wise.

### DIFF
--- a/lua/sim/defaultweapons.lua
+++ b/lua/sim/defaultweapons.lua
@@ -980,7 +980,9 @@ OverchargeWeapon = Class(DefaultProjectileWeapon) {
         if self:BeenDestroyed() then return end
         DefaultProjectileWeapon.OnEnableWeapon(self)
         self:SetWeaponEnabled(true)
-        self.unit:SetWeaponEnabledByLabel(self.DesiredWeaponLabel, false)
+        if self:CanOvercharge() then
+            self.unit:SetWeaponEnabledByLabel(self.DesiredWeaponLabel, false)
+        end
         self.unit:BuildManipulatorSetEnabled(false)
         self.AimControl:SetEnabled(true)
         self.AimControl:SetPrecedence(20)
@@ -1018,7 +1020,15 @@ OverchargeWeapon = Class(DefaultProjectileWeapon) {
             if self:CanOvercharge() then
                 DefaultProjectileWeapon.IdleState.OnGotTarget(self)
             else
-                self:OnDisableWeapon()
+                self:ForkThread(function()
+                    while self.enabled and not self:CanOvercharge() do
+                        WaitSeconds(0.1)
+                    end
+                    
+                    if self.enabled then
+                        self:OnGotTarget()
+                    end
+                end)
             end
         end,
 


lua/sim/defaultweapons.lua
```diff
@@ -980,7 +980,9 @@ OverchargeWeapon = Class(DefaultProjectileWeapon) {
         if self:BeenDestroyed() then return end
         DefaultProjectileWeapon.OnEnableWeapon(self)
         self:SetWeaponEnabled(true)
-        self.unit:SetWeaponEnabledByLabel(self.DesiredWeaponLabel, false)
+        if self:CanOvercharge() then
```
aeoncleanse: This will disable the main weapon if OC isn't ready... ?

aeoncleanse:
Can you be more specific about what doesn't feel right? How does it feel? What should testers look for?

lua/sim/defaultweapons.lua
MrNukealizer: Currently the main weapon is always disabled when the overcharge weapon is enabled. For auto overcharge that's only when it's ready to fire, but manual overcharge is enabled when it gets a target. What this does is not disable the main weapon if the overcharge weapon is enabled but isn't ready to fire yet. Otherwise the ACU stands there not shooting while waiting for it to be ready.

I'm honestly not sure why the main weapon is disabled or what that's supposed to accomplish. In my testing there was no noticeable difference between disabling it and leaving it enabled.

MrNukealizer:
It feels like there's a bug in here just waiting to be found, but I tried every situation I could think of and couldn't find it. There are two main things I could see going wrong:

1. The ACU overcharges the wrong target or goes through with a cancelled overcharge order.

2. The ACU doesn't overcharge at all because the main weapon stays enabled and messes up targeting or something.

I tested for both to the best of my ability but this is the kind of thing I use differently than other people so I may not have done the right tests.

aeoncleanse:
Hmmm... You're right, I'm not sure why the normal weapon needs disabling either :/ Is it worth exploring a redesign of Overcharge? The two modes have different weapons, with the manual fire one specifying ManualFire = true in bp... Is there a reason we need to enable/disable the auto one except on the button toggle? Can't it just be on and behave like a normal, 5s reload, energy requiring weapon?

MrNukealizer:
I honestly don't understand what half the overcharge code is for, so there might be something I'm missing. It does look like the auto version could be made more "normal" and still function the same. It seems like the only significant difference between overcharge and a normal weapon is that it tells the engine to disable and reenable the overcharge order.

I don't think it's worth a resesign at the moment, since it seems to mostly work as intended. It would be worth looking into if something more does need to be changed though.

aeoncleanse:
It's always worth a redesign if it fixes a bug and simplifies code. Always. I see no reason not to try splitting the Auto into a new weapon class, having the button enable/disable stuff on the manual one only, and have the auto act like a normal weapon except it drains E and checks E storage on OnGotTarget or somewhere in that sequence. Also has to check not upgrading etc, but other than that, just have the button right click properly enable/disable that weapon.

PR #2064, opened 2017-06-09
Fix network stats opening but not closing when alone

MrNukealizer:
When pressing F11 in a replay or singleplayer game, the network stats overlay opens, but pressing the key again doesn't close it.


ChrisKitching:
Bonus points for utter obscurity. :D

PR #2065, opened 2017-06-09
hide maps if newer versions are available locally

CookieNoob:
extends the "hide obsolete" function in the map select window to hide maps where higher versions with the same name are available locally.
Currently it only hides maps on the blacklist.

### DIFF
--- a/lua/ui/dialogs/mapselect.lua
+++ b/lua/ui/dialogs/mapselect.lua
@@ -238,19 +238,21 @@ mapFilters = {
         FilterFactory = {
             SelectedKey = 1,
             Filters = {
-                function(scenInfo) 
-                    local newestVersion = true
-                    if scenInfo.map_version ~= nil then
-                        for _,comparisionlist in scenarios do
-                            if LOC(scenInfo.name) == LOC(comparisionlist.name) then
-                                if scenInfo.map_version < (comparisionlist.map_version or 1) then
-                                    newestVersion = false
-                                    break
+                function(scenInfo)
+                    if scenInfo.Outdated then
+                        return false
+                    end
+                    local version = scenInfo.map_version or 0
+                    for _,comparisionlist in scenarios do
+                        if scenInfo.name == comparisionlist.name then
+                            if comparisionlist.map_version then
+                                if version < comparisionlist.map_version then
+                                    return false
                                 end
                             end
                         end
                     end
-                    return not (scenInfo.Outdated or not newestVersion)
+                    return true
                 end
             },
             Build = function(self)
--- a/lua/ui/dialogs/mapselect.lua
+++ b/lua/ui/dialogs/mapselect.lua
@@ -238,7 +238,20 @@ mapFilters = {
         FilterFactory = {
             SelectedKey = 1,
             Filters = {
-                function(scenInfo) return not scenInfo.Outdated end
+                function(scenInfo) 
+                    local newestVersion = true
+                    if scenInfo.map_version ~= nil then
+                        for _,comparisionlist in scenarios do
+                            if LOC(scenInfo.name) == LOC(comparisionlist.name) then
+                                if scenInfo.map_version < (comparisionlist.map_version or 1) then
+                                    newestVersion = false
+                                    break
+                                end
+                            end
+                        end
+                    end
+                    return not (scenInfo.Outdated or not newestVersion)
+                end
             },
             Build = function(self)
                 return self.Filters[self.SelectedKey]
@@ -1015,17 +1028,6 @@ function PopulateMapList()
             -- Name filter needs special treatment
             if nameFilter and nameFilter:GetText() ~= "" then
                 passedFiltering = passedFiltering and string.lower(sceninfo.name):find(string.lower(nameFilter:GetText()))
-            end
-        end
-        -- hide maps that are not on the blacklist but have newer versions available if obsolete maps are hidden
-        if currentFilters['map_obsolete'] ~= nil and sceninfo.map_version ~= nil then
-            for _,comparisionlist in scenarios do
-                if LOC(sceninfo.name) == LOC(comparisionlist.name) then
-                    if sceninfo.map_version < (comparisionlist.map_version or 1) then
-                        passedFiltering = false
-                        break
-                    end
-                end
             end
         end
         
--- a/lua/ui/dialogs/mapselect.lua
+++ b/lua/ui/dialogs/mapselect.lua
@@ -1030,7 +1030,7 @@ function PopulateMapList()
                 passedFiltering = passedFiltering and string.lower(sceninfo.name):find(string.lower(nameFilter:GetText()))
             end
         end
-        
+
         if passedFiltering then
             -- Make sure we finish up with the right map selected.
             scenarioKeymap[count] = i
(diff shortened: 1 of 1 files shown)


TheKeyBlue:
Its probably better if you add that filter to the mapFilter table:
https://github.com/CookieNoob/fa/blob/develop/lua/ui/dialogs/mapselect.lua#L227

So it fits in the used filter system.

ChrisKitching:
Probably worth creating a ticket on the client to automatically delete such maps, too.

The map versioning system sucks. Right now every version ever just accumulates, filling your hard drive. It'd be rather useful for the client to have a system to garbage collect outdated ones, in addition to simply hiding them in UI.

CookieNoob:
It would be great to have an auto update function in the client which checks if the map you want to host is the most recent version of that map and automatically download the new version if available. Afterwards the old version could be deleted.

I requested it here: https://github.com/FAForever/client/issues/418 (september last year)

ChrisKitching:
> I requested it here: FAForever/client#418 (september last year)

Developer time is scarce, alas.

CookieNoob:
yes I know :)

lua/ui/dialogs/mapselect.lua
```diff
@@ -238,7 +238,20 @@ mapFilters = {
...
             SelectedKey = 1,
             Filters = {
-                function(scenInfo) return not scenInfo.Outdated end
+                function(scenInfo) 
+                    local newestVersion = true
+                    if scenInfo.map_version ~= nil then
+                        for _,comparisionlist in scenarios do
+                            if LOC(scenInfo.name) == LOC(comparisionlist.name) then
+                                if scenInfo.map_version < (comparisionlist.map_version or 1) then
+                                    newestVersion = false
```
ChrisKitching: This boolean is unnecessary: just return false right away.

lua/ui/dialogs/mapselect.lua
```diff
@@ -238,7 +238,20 @@ mapFilters = {
...
+                                    break
+                                end
+                            end
+                        end
+                    end
+                    return not (scenInfo.Outdated or not newestVersion)
```
ChrisKitching: Nested nots - unnecessarily confusing formulation.

ChrisKitching: Consider pulling the Outdated check up to the top:

```lua
if scenInfo.Outdated then
    return false
end

... stuff
```

ChrisKitching: (Then this bit just becomes `return true`)

lua/ui/dialogs/mapselect.lua
```diff
@@ -238,7 +238,20 @@ mapFilters = {
         FilterFactory = {
             SelectedKey = 1,
             Filters = {
-                function(scenInfo) return not scenInfo.Outdated end
+                function(scenInfo) 
+                    local newestVersion = true
+                    if scenInfo.map_version ~= nil then
```
ChrisKitching: Explicit comparison to nil is a definite antipattern. You can just do `if scenInfo.map_version ...`.

lua/ui/dialogs/mapselect.lua
```diff
@@ -238,7 +238,20 @@ mapFilters = {
...
+                        for _,comparisionlist in scenarios do
+                            if LOC(scenInfo.name) == LOC(comparisionlist.name) then
+                                if scenInfo.map_version < (comparisionlist.map_version or 1) then
```
ChrisKitching: I've seen maps with a version number of zero.

I'd suggest rephrasing this to extract the map_version null check.

lua/ui/dialogs/mapselect.lua
```diff
@@ -238,7 +238,20 @@ mapFilters = {
         FilterFactory = {
             SelectedKey = 1,
             Filters = {
-                function(scenInfo) return not scenInfo.Outdated end
+                function(scenInfo) 
+                    local newestVersion = true
+                    if scenInfo.map_version ~= nil then
+                        for _,comparisionlist in scenarios do
+                            if LOC(scenInfo.name) == LOC(comparisionlist.name) then
```
ChrisKitching: Doesn't `scenInfo` have an `id` field or something you can use to do this comparison?

Also - aren't version numbers typically encoded in the names of maps? That's certainly something that had to be dealt with when making the map blacklist (indeed, that's much of how it worked). That was how ZeP implemented map versioning, and nobody got around to fixing it yet.

Does this code actually work on real-world examples? I'd expect it to fall over when faced with Zep-insanity versioned maps (which, unless someone changed something, is all of them?)

ChrisKitching: You can, naturally, use a regular expression and the lua string library to decode the version number from the name, if you please. Or perhaps something somewhere else did that already? That'd be sensible...

lua/ui/dialogs/mapselect.lua
ChrisKitching: This and the comparison above could be joined with logical or, and then just do an early continue, instead of making the nesting depth greater and greater,

lua/ui/dialogs/mapselect.lua
CookieNoob: yes it works, I tested it. I realized that only one of my maps used a LOC tag in the map name and since the updater renames the map files based on the name, I dont think its possible to have language specific names. I remove these LOCs.

Version numbers are not in the names of the maps.

lua/ui/dialogs/mapselect.lua
CookieNoob: map_version = 0 is not allowed anymore with the map uploader afaik. I can switch the `or 1` to `or 0` for maps that have map_version = 0.

Do you mean I should add another if before the map version check?

lua/ui/dialogs/mapselect.lua
ChrisKitching: > I dont think its possible to have language specific names. I remove these LOCs.

This should be possible: if it isn't, you just need to add a LOC() call to something, somewhere. That said, few maps do this, and your map will have to provide a hook to the language files for this to work: something even fewer people will know how to do.

> Version numbers are not in the names of the maps.

Right, sorry, the map loader presumably parses and removes that, somewhere, then... The _filenames_ definitely do contain version numbers...

lua/ui/dialogs/mapselect.lua
ChrisKitching: The point I was making is that you can cheat and use early-continue to simplify things a little:

```lua

if LOC(scenInfo.name) == LOC(comparisionlist.name) or 
   scenInfo.map_version < (comparisionlist.map_version or 1) then
       continue
end
```

This approach means your indentation doesn't grow and grow as you add more conditions. This is a particularly powerful technique if you have else blocks (because this stops them from all ending up in a confusing - possibly duplicated - pole at the bottom.


But yes. not super important ;)

lua/ui/dialogs/mapselect.lua
speed2CZ: I think only original maps and coop missions use localized name. As you cant hook the loc files from the map folder.

ChrisKitching: There's probably value in changing that.

PR #2066, opened 2017-06-09
add outdated map warning after map selection

### LINKED ISSUE #2024: outdated map warning
since we dont have an auto-update function for maps/mods, the current system works with blacklists. It is working properly for mods (i.e. it disables them and shows the player the reason why its disabled...), for maps however...

There is an option in the game lobby to "hide outdated maps". Enabling it will prevent the blacklisted maps from showing up in the map selection window. However, once this option is disabled all available maps are shown and it has to be enabled manually again. These maps can still be hosted/played (this is a good thing).

I would like to display a warning when a map is blacklisted and selected (i.e. when the option window is closed) `The currently selected map is outdated and/or unbalanced. Please download the latest version from the map vault.`
It should only be shown when a new map is selected and it is on the blacklist (i.e. dont show it when only some game options are changed).

Furthermore, the option "hide outdated maps" should be set to "enabled" every time the lobby is started.

CookieNoob:
if an outdated map is selected, a popup is shown to warn the host and asks him to update the map
fixes #2024


TheKeyBlue:
You have copypasted code in `selectButton.OnClick` and `selectButton.OnDoubleClick` which should be refactored into a function. 
Copypasted code is simply bad practice.

PR #2067, opened 2017-06-09
fix "hide_obsolete" filter in map selection disabled by default

CookieNoob:
if the game prefs file contains a section (usually empty) of the form
stored_filter_state = {  },
then the map_obsolete flag is not set to 1 and thus all maps are shown
even if the option for the filter sais "hide outdated maps".

part of #2024

### DIFF
--- a/lua/ui/dialogs/mapselect.lua
+++ b/lua/ui/dialogs/mapselect.lua
@@ -264,6 +264,9 @@ function InitFilters()
             SelectedKey = 1 -- Enable obsolete map filtering by default.
         }
     }
+    savedFilterState['map_obsolete'] = {
+        SelectedKey = 1 -- Enable obsolete map filtering by default.
+    }
 
     -- savedFilterState is an array of tables of filter options
     for filterKey, v in savedFilterState do


lua/ui/dialogs/mapselect.lua
```diff
@@ -264,6 +264,9 @@ function InitFilters()
             SelectedKey = 1 -- Enable obsolete map filtering by default.
         }
     }
+    savedFilterState['map_obsolete'] = {
+        SelectedKey = 1 -- Enable obsolete map filtering by default.
+    }
```
ChrisKitching: This will unilaterally enable the `map_obsolete` filter when filter state is loaded.

This means that if someone turns off the filter, it will be turned by on again as soon as they reload. Surely that's not what you want?

The config file shouldn't ever end up with a `map_obsolete = {}` under `stored_filter_state` - the default initialisation is just above, and (hopefully) it's getting properly saved somewhere else.

You could defensively check for that case here and then set the default, but the real solution is to stop it from getting into this mess to begin with...

CookieNoob: imo the filter should be enabled with every restart of the game lobby. The maps are blacklisted for a reason (new version/unbalanced/....). In my prefs file the savedFilterState was always empty. When opening the map selection window the "hide obsolete" filter was shown as "yes" (hide obsolete maps) but the maps were displayed anyway.

PR #2072, opened 2017-06-13
Add support UTF (all languages) to lobby and ingame chat

dk0x:
Add support UTF (all languages) to lobby and game chat.

Problem: 
A: lua 5.1 dont support UTF conversion (only 5.3+ v support it) 
B: table of available symbols to type in chat hardcoded in exe class (moho.edit_methods OnCharPressed, OnNonTextKeyPressed). Its only Symbols Aa-Zz, 0-9, and !@#$%^&*()_+-=`~ etc.

Issue:
A: i add one file (UTF.lua) and one function UTF(charCode) to convert all charcodes directly to symbols in UTF format (all charcodes 128-65534)
B: because all unavailable to type charcodes call OnNonTextKeyPressed, 
i add 4 line of code 
```
	local UTF = import('/lua/UTF.lua').UTF
	local ccc = UTF(charcode)
	if ccc ~= '' then
		self:SetText(self:GetText() .. ccc)
	end
```
in 2 files 
```
	\lua\ui\game\chat.lua
	\lua\ui\lobby\lobby.lua
```
in function OnNonTextKeyPressed

Result: 
Easy to type any language in chat directly from your hands to game without Notepad (ctrl+c, ctrl+v) 

Regards -_o, dk0xFFFF


lua/ui/game/chat.lua
```diff
@@ -648,6 +648,11 @@ function CreateChatEdit()
     LayoutHelpers.AtVerticalCenterIn(group.chatBubble, group.edit)
 
     group.edit.OnNonTextKeyPressed = function(self, charcode, event)
+        local UTF = import('/lua/UTF.lua').UTF
+        local ccc = UTF(charcode)
+        if ccc ~= '' then
+		self:SetText(self:GetText() .. ccc)
```
ChrisKitching: Do not indent with tabs

lua/ui/lobby/lobby.lua
```diff
@@ -2947,6 +2947,11 @@ function CreateUI(maxPlayers)
 
     --- Handle up/down arrow presses for the chat box.
     GUI.chatEdit.OnNonTextKeyPressed = function(self, keyCode)
+		local UTF = import('/lua/UTF.lua').UTF
+		local ccc = UTF(keyCode)
+		if ccc ~= '' then
+			self:SetText(self:GetText() .. ccc)
+		end
```
ChrisKitching: This code is copy-pasted in two places: introduce a function for this.

lua/ui/lobby/lobby.lua
```diff
@@ -2947,6 +2947,11 @@ function CreateUI(maxPlayers)
 
     --- Handle up/down arrow presses for the chat box.
     GUI.chatEdit.OnNonTextKeyPressed = function(self, keyCode)
+		local UTF = import('/lua/UTF.lua').UTF
```
ChrisKitching: You could move this import to the top of the file, instead of doing it inline everywhere.

After all, it seems likely that this function will come in handy in a few other places in the UI.

lua/ui/game/chat.lua
ChrisKitching: This line assumes that the caret is at the end of the line. This isn't necessarily true! It'll add characters to the end always, regardless of caret position.

ChrisKitching:
Oooh, a much requested feature, yay.

You need to rebase your branch: those merge commits are tangly. If you're new to git, I recommend the [Atlassian](https://www.atlassian.com/git/tutorials) tutorials. Especially, in this case, the part about [rebasing](https://www.atlassian.com/git/tutorials/rewriting-history).

 The UTF-8 lookup table is a bit concerning. That's over a megabyte of lua for the parser to handle, and most worryingly it procedurally builds the lookup table:

```
chars[128]=''
chars[129]=''
chars[130]=''
chars[131]=''
```

I suspect you'll find you get rather better performance if you instead make it a giant array literal: `["a","b","c",...]` - you'll certainly get a much smaller lua file. I realise that the indexes don't start from zero, but since they are contiguous you can fix that by simply subtracting an offset whenever you look into the table.

You might also have success representing your lookup table as a giant string, and extracting a character from it at the appropriate index.

I'm assuming you've tried using the rather promising-looking `string.char()` function documented [here](http://lua-users.org/wiki/StringLibraryTutorial)? It'd be nice to avoid having the huge lookup table.

lua/ui/lobby/lobby.lua
```diff
@@ -2947,6 +2947,11 @@ function CreateUI(maxPlayers)
...
+		local ccc = UTF(keyCode)
+		if ccc ~= '' then
```
ChrisKitching: Since appending the empty string is a no-op, you don't really need to bother with this check.

ChrisKitching:
Oh, and, if you do keep the lookup table, _make it global_.

Currently, every time you press a key, it allocates a new lookup table, procedurally fills it, looks up one element, and then it becomes eligible for GC.

Eep.

Sheeo:
Rather than the procedurally generated UTF-8 lookup function, you may want to take a look [here](https://github.com/Stepets/utf8.lua/blob/master/utf8.lua#L311).

dk0x:
> Sheeo commented 6 hours ago
> Rather than the procedurally generated UTF-8 lookup function, you may want to take a look here.

i tested it before, but lua throw exception on line 
`		local Byte1 = 0x80 + (unicode % 0x40);`
` failed: c:\programdata\faforever\gamedata\!lua.nx2\lua\utf.lua(6): `)' expected near `%'`

i think lua 5.1 dont support %

dk0x:
> Oooh, a much requested feature, yay.
> 
> You need to rebase your branch: those merge commits are tangly. If you're new to git, I recommend the Atlassian tutorials. Especially, in this case, the part about rebasing.
> 
> The UTF-8 lookup table is a bit concerning. That's over a megabyte of lua for the parser to handle, and most worryingly it procedurally builds the lookup table:
> 
> chars[128]='�'
> chars[129]='�'
> chars[130]='�'
> chars[131]='�'
> I suspect you'll find you get rather better performance if you instead make it a giant array literal: ["a","b","c",...] - you'll certainly get a much smaller lua file. I realise that the indexes don't start from zero, but since they are contiguous you can fix that by simply subtracting an offset whenever you look into the table.
> 
> You might also have success representing your lookup table as a giant string, and extracting a character from it at the appropriate index.
> 
> I'm assuming you've tried using the rather promising-looking string.char() function documented here - that'd at least avoid the need for the first 256 elements of the table :P

i tied array method ["a","b","c",...] but it is so borning with debug becouse lua tell me only line of error, not a position in line (some symbols generated(i generate this file in c# miniprogramm) with errors like 'Æ', 'Æ, 'Æ, 'Æ', 'Æ' without ' symbol :-/)

But ok you are right about this. I try now generate better and smaller UTF file without errors.

and yes, i'm new on git. its my first pull request =)

lua/ui/game/chat.lua
dk0x: i love tabs! ;-D HOLYWAR!

lua/ui/lobby/lobby.lua
dk0x: ok
like 		import('/lua/UTF.lua').AddUnicodeCharToEditText(self, charcode)

lua/ui/lobby/lobby.lua
dk0x: > ok
> like import('/lua/UTF.lua').AddUnicodeCharToEditText(self, charcode)

or better 
top of file 
local AddUnicodeCharToEditText = import('/lua/UTF.lua').AddUnicodeCharToEditText

in function
AddUnicodeCharToEditText(self, charcode)

lua/ui/game/chat.lua
MrNukealizer: Tabs are amazing, but unfortunately the entire codebase uses 4-space tabs.

dk0x: hm. absolytely true!

work in progress. how can i get position of carret?

lua/ui/lobby/lobby.lua
dk0x: but empty string anyway call edit:SetText(edit:GetText() .. unicodeChar)

better skip this if string is empty

dk0x:
> Oh, and, if you do keep the lookup table, make it global.
> 
> Currently, every time you press a key, it allocates a new lookup table, procedurally fills it, looks up one element, and then it becomes eligible for GC.
> 
> Eep.


i realize table of symbols and make this table global in _G global table like _G.UnicodeTable = {}

lua/ui/game/chat.lua
dk0x: okay :-(

:-D

aeoncleanse:
Push additional changes to the same branch and keep this pr open. The pr will update. This way we keep the comment history in place

dk0x:
onr problem for me now

how can i get position of carret?

ChrisKitching:
> i tied array method ["a","b","c",...] but it is so borning with debug becouse lua tell me only line of error, not a position in line (some symbols generated(i generate this file in c# miniprogramm) with errors like 'Æ', 'Æ, 'Æ, 'Æ', 'Æ' without ' symbol :-/)

Definitely, the only sane way to do this is use a script to generate the table.



> Push additional changes to the same branch and keep this pr open. The pr will update. This way we keep the comment history in place

I linked you earlier to the Atlassian tutorial on history rewriting.

A PR is basically an indicator of "Please merge my branch into master". You can freely force-push your branch to change the content of your request: no need to spam more commits or make new PRs.

It's a bit tricky to get the hang of at first, but it really makes a great workflow once you've got it down.



> i realize table of symbols and make this table global in _G global table like _G.UnicodeTable = {}

To be more precise, you want it to be module-local to your UTF8 module. The key thing you want to avoid is having it function-local as it now is.

> i think lua 5.1 dont support %

Normal lua 5.1 actually does, but the variant of it in Supreme Commander doesn't. It does have `math.mod()`, however, which is equivalent.

lua/ui/game/chat.lua
ChrisKitching: Either can be worked with, but:

- All the existing code uses 4-spaces.
- Files that contain a mixture often render incorrectly. Editors need to be told how many "spaces-worth" of width to use to render a tab. This needs changing when you switch between projects that use a mixture of 2-space/tabs and a project that uses a mixture of 4-space/tabs.

So - it's usually simpler just to configure you editor to insert the correct number of spaces when you press tab. Basically any code editor ever has this feature.

On a related note: make sure to use `\n` line endings. We've had historical trouble with git churn caused by this silliness (since many editors normalise line endings at save-time... :/ )

dk0x:
inserting in string in lua so borning =)
```
function InsertChar(str, chr, pos)
        return str:sub(1, pos) .. chr .. str:sub(pos+1)
end
```

Problem with insert symbol to carret pos:

lua string.sub work nice if string no contain unicode sumbols like 
string.sub("1234567890", 7) == 1234567
but if string contain >1 unicode symbol, string.sub(s, start, end) broke the string like this

```
pos = 2
str = "12345й67890"
str:sub(1, pos) .. 'ё' .. str:sub(pos+1) == "1ё2345й67890" -- work normaly

str = "12345й67890"
pos = 8
str:sub(1, pos) .. 'ё' .. str:sub(pos+1) == "12345й6ё7890" -- broke, need 12345й67ё890

str = "12345й67890"
pos = 6
str:sub(1, pos) .. 'ё' .. str:sub(pos+1) == "12345Ѐёﾹ67890" -- absolytely broke, need 12345йё67890
```
if string contain unicode symbol str:sub(1, pos) broke string
i think is because unicode symbol is 2 bytes and string.sub(s, start, end) take first byte, add as symbol, and broke second byte and add as symbol. 
but copypaste й from notepad work perfect

example:
й is 2 bytes 04 and 39 (in hex ofc)
in string 12345Ѐёﾹ67890 we see 2 symbols Ѐ and ﾹ. String.sub divided 2 byte of й to Ѐ == 0x0400 and strange ﾹ == 0xFFB9 

well I have not yet solved this problem. carret insert dont work now.
except it all work normaly.

ChrisKitching:
> i think is because unicode symbol is 2 bytes

See the section [here](http://lua-users.org/wiki/LuaUnicode) about string length.

tl;dr: yes, lua string length is in terms of bytes. You'll need to account for that. Note that this is not necessarily as simple as checking the size of the char code (and seeing how many bytes are needed to represent it), but you might be able to get away with that,.

lua/ui/game/chat.lua
dk0x: i'm use notepad++ for edit lua files
line endings end with Dos\Windows CR LF
i can change to LF or CR or leave as is.

anyway i replace tabs to 4 spaces.

dk0x:
ok. i can get size of string contain unicode symbols
123ё456 size = 7  (symbols [1][2][3][р][ﾑ][4][5][6])
but... example:
123ё4|56 ( | is carret pos)
fa exe tell lua - carretpos = 5
but string.sub(1, pos) need real byte carret pos for insert (is 6 string.sub(1, 6) + chr + string.sub(8))
first problem is convert unicode carret pos (5) in byte carret pos (6) (123рﾑ|456])
i need foreach symbols, and if symbol is unicode symbol - add to real carretpos 2, Otherwise 1
but if i convert text 123ё456  to array, result is [1][2][3][р][ﾑ][4][5][6]
all string functions(gsub, sub etc) work only with 1 by 1 bytes (like [1][2][3][р][ﾑ][4][5][6] array)
How can i extract symbols without like [1][2][3][ё][4][5][6] (no [1][2][3][р][ﾑ][4][5][6])
if i achive it, i can foreach it and calculate real carret pos for insert new unicode char.

(i broke my brain =D)

dk0x:
rdy to merge

MrNukealizer:
You're fine. The reason you can't create that pull request is because this one is from the same branch to the same branch. You just have to wait for someone to come around and merge it.

lua/UTF.lua
```diff
@@ -0,0 +1,1346 @@
+-- dk0x unicode minilib
+
+_G.UnicodeTable = {
```
ChrisKitching: As I mentioned before, you want this to be module-local, not function-local. You really, really do not want this to stay in memory forever. :P

lua/UTF.lua
```diff
@@ -0,0 +1,1346 @@
+-- dk0x unicode minilib
+
+_G.UnicodeTable = {
+"","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","",
+"","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","",
+"","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","","",
+"","","","","","","","",""," ","¡","¢","£","¤","¥","¦","§","¨","©","ª","«","¬","­","®","¯","°","±","²","³","´","µ","¶","·","¸","¹","º","»","¼","½","¾","¿","À","Á","Â","Ã","Ä","Å","Æ","Ç","È",
```
ChrisKitching: You seem to have padded with empty-string instead of just adding an offset to the value when you look into the table. This seems unwise.

lua/UTF.lua
```diff
@@ -0,0 +1,1346 @@
...
+"㸝","㸞","㸟","㸠","㸡","㸢","㸣","㸤","㸥","㸦","㸧","㸨","㸩","㸪","㸫","㸬","㸭","㸮","㸯","㸰","㸱","㸲","㸳","㸴","㸵","㸶","㸷","㸸","㸹","㸺","㸻","㸼","㸽","㸾","㸿","㹀","㹁","㹂","㹃","㹄","㹅","㹆","㹇","㹈","㹉","㹊","㹋","㹌","㹍","㹎",
+"㹏","㹐","㹑","㹒","㹓","㹔","㹕","㹖","㹗","㹘","㹙","㹚","㹛","㹜","㹝","㹞","㹟","㹠","㹡","㹢","㹣","㹤","㹥","㹦","㹧","㹨","㹩","㹪","㹫","㹬","㹭","㹮","㹯","㹰","㹱","㹲","㹳","㹴","㹵","㹶","㹷","㹸","㹹","㹺","㹻","㹼","㹽","㹾","㹿","㺀",
+"㺁","㺂","㺃","㺄","㺅","㺆","㺇","㺈","㺉","㺊","㺋","㺌","㺍","㺎","㺏","㺐","㺑","㺒","㺓","㺔","㺕","㺖","㺗","㺘","㺙","㺚","㺛","㺜","㺝","㺞","㺟","㺠","㺡","㺢","㺣","㺤","㺥","㺦","㺧","㺨","㺩","㺪","㺫","㺬","㺭","㺮","㺯","㺰","㺱","㺲",
+"㺳","㺴","㺵","㺶","㺷","㺸","㺹","㺺","㺻","㺼","㺽","㺾","㺿","㻀","㻁","㻂","㻃","㻄","㻅","㻆","㻇","㻈","㻉","㻊","㻋","㻌","㻍","㻎","㻏","㻐","㻑","㻒","㻓","㻔","㻕","㻖","㻗","㻘","㻙","㻚","㻛","㻜","㻝","㻞","㻟","㻠","㻡","㻢","㻣","㻤",
+"㻥","㻦","㻧","㻨","㻩","㻪","㻫","㻬","㻭","㻮","㻯","㻰","㻱","㻲","㻳","㻴","㻵","㻶","㻷","㻸","㻹","㻺","㻻","㻼","㻽","㻾","㻿","㼀","㼁","㼂","㼃","㼄","㼅","㼆","㼇","㼈","㼉","㼊","㼋","㼌","㼍","㼎","㼏","㼐","㼑","㼒","㼓","㼔","㼕","㼖",
+"㼗","㼘","㼙","㼚","㼛","㼜
```
ChrisKitching: This commented line is not adding any value.

ChrisKitching: This file contains zero comments. Given the funky hacks involved, that's not a good sign.

ChrisKitching: This uses gsub to turn the string into a list of characters. This isn't really necessarily, since you read every character _at most once_, meaning much of the work done by this is redundant.

Why can't you just use str:sub to extract the character you want inside the while loop? That way you only read out the characters you actually need, and you don't iterate the string twice.

ChrisKitching: Sudden thought.

Instead of exploiting gsub with a lambda to build this table of characters, why don't you use it to directly calculate the offset you want?

```lua
local offset = 0

local function accumulateOffset(c)
    if offset < pos && string.byte(c) > 200 then
        offset = offset + 1
    end
end
std:gsub(".", accumulateOffset)

-- Tada, offset is now equal to the number of multibyte characters before `pos`, and you don't need the while loop at all.
```

ChrisKitching:
Good progress!

Before we merge, I'd like you to attempt to use the unicode function sheeo referenced earlier:
> Rather than the procedurally generated UTF-8 lookup function, you may want to take a look here.

This would allow us to avoid having the nasty lookup table entirely.

Note that you can replace uses of the infix modulo operator `A % B` with calls to the function `math.mod(A, B)`.

dk0x:
```
local function UTF(unicode)
	if unicode <= 0x7F then return string.char(unicode) end

	if (unicode <= 0x7FF) then
		local Byte0 = 0xC0 + math.floor(unicode / 0x40);
		local Byte1 = 0x80 + math.mod(unicode, 0x40);
		return string.char(Byte0, Byte1);
	end;

	if (unicode <= 0xFFFF) then
		local Byte0 = 0xE0 + math.floor(unicode / 0x1000);
		local Byte1 = 0x80 + math.mod(math.floor(unicode / 0x40), 0x40);
		local Byte2 = 0x80 + math.mod(unicode, 0x40);
		return string.char(Byte0, Byte1, Byte2);
	end;

	if (unicode <= 0x10FFFF) then
		local code = unicode
		local Byte3= 0x80 + math.mod(code, 0x40);
		code       = math.floor(code / 0x40)
		local Byte2= 0x80 + math.mod(code, 0x40);
		code       = math.floor(code / 0x40)
		local Byte1= 0x80 + math.mod(code, 0x40);
		code       = math.floor(code / 0x40)
		local Byte0= 0xF0 + code;

		return string.char(Byte0, Byte1, Byte2, Byte3);
	end;

	error 'Unicode cannot be greater than U+10FFFF!'
end

```

worked. table deleted. use func now.

goin to next optimization. work in progress...

lua/UTF.lua
dk0x: because only iterate array can skip bytes (t[i+addPos])
skip bytes need because only first byte of unicode char need to check [р]>200, and ignore second [ﾑ]. ([1][2][3][р][ﾑ][4][5] - [р], not [ﾑ]. if do not skip, this [ﾑ] can contain byte [1-255] and iterate addPos to)

let's check ur code with pos = 5
on monitor [1][й][2][й][3] | [й][4]
in memory  [1][р][ﾑ][2][р] | [ﾑ][3][р][ﾑ][4]
need           [1][р][ﾑ][2][р][ﾑ][3] | [р][ﾑ][4]
need offset = 2 and newPos = 7

pseudocode
```
if offset(0) < pos(5) (true) and c[1] > 200 (false) offset = 0
if offset(0) < pos(5) (true) and c[р] > 200 (true)  offset = 1
if offset(1) < pos(5) (true) and c[ﾑ] > 200 (false) offset = 1 (potentially c > 200 (true) and offset inc twice, need skip this check)

if offset(1) < pos(5) (true) and c[2] > 200 (false) offset = 1
if offset(1) < pos(5) (true) and c[р] > 200 (true)  offset = 2
if offset(2) < pos(5) (true) and c[ﾑ] > 200 (false) offset = 2 (potentially c > 200 (true) and offset inc twice, need skip this check)
/*/* and now we need stop iteration, but not */*/

if offset(2) < pos(5) (true) and c[3] > 200 (false) offset = 2
if offset(2) < pos(5) (true) and c[р] > 200 (true)  offset = 3
if offset(3) < pos(5) (true) and c[ﾑ] > 200 (false) offset = 3 (potentially c > 200 (true) and offset inc twice, need skip this check)

if offset(3) < pos(5) (true) and c[4] > 200 (false) offset = 3
if offset(3) < pos(5) (true) and c[р] > 200 (true)  offset = 4
if offset(4) < pos(5) (true) and c[ﾑ] > 200 (false) offset = 4 (potentially c > 200 (true) and offset inc twice, need skip this check)
```

result offset = 4 and newPos = 9

i improve ur idea

```
    local offset = 0
    local posWithUnicodeChars = 1
    local skipSecondByteOfUnicodeChar = false
    str:gsub(".", 
        function(c) 
            if posWithUnicodeChars <= pos then 
                -- first byte always >=0xC0 (192). look at function UTF(unicode) local Byte0 = 0xC0 + ...
                if string.byte(c) >= 192 and not skipSecondByteOfUnicodeChar then -- if byte is first part and not second part of unicode char 
                    offset = offset + 1     -- increase offset
                    skipSecondByteOfUnicodeChar = true -- and skip next byte because is second part of unicode char
                else        -- otherwise skip second byte and reset skipSecondByteOfUnicodeChar to false
                    skipSecondByteOfUnicodeChar = false
                    posWithUnicodeChars = posWithUnicodeChars + 1  -- this inc fires all iteration of bytes one by char except unicode char (fires one by two byte)
                end
            end
        end
    )
   return str:sub(1, pos+offset) .. chr .. str:sub(pos+offset+1)

```


BUT!
this code worse for understanding logicaly
i think better use array instead
```
    local byteArrayOfText={}
    str:gsub(".", function(c) table.insert(byteArrayOfText,c) end)
    local addPos = 0
    local posWithUnicodeChars = 1
    while posWithUnicodeChars <= pos do
        -- first byte always >=0xC0 (192). look at function UTF(unicode) local Byte0 = 0xC0 + ...
        if string.byte(byteArrayOfText[posWithUnicodeChars+addPos]) >= 192 then 
            addPos = addPos + 1
        end
        posWithUnicodeChars = posWithUnicodeChars + 1
    end
    return str:sub(1, pos+addPos) .. chr .. str:sub(pos+addPos+1)

```

lua/UTF.lua
dk0x: use func instead of table now. not relevant

lua/UTF.lua
dk0x: use func instead of table now. not relevant

lua/UTF.lua
dk0x: use func instead of table now. not relevant

dk0x:
well. need choice with two variants of inserting char to string with unicode chars

```
function InsertChar1(str, chr, pos)
    local byteArrayOfText={}
    str:gsub(".", function(c) table.insert(byteArrayOfText,c) end)
    local addPos = 0
    local posWithUnicodeChars = 1
    while posWithUnicodeChars <= pos do
        -- first byte always >=0xC0 (192). look at function UTF(unicode) local Byte0 = 0xC0 + ...
        if string.byte(byteArrayOfText[posWithUnicodeChars+addPos]) >= 192 then 
            addPos = addPos + 1
        end
        posWithUnicodeChars = posWithUnicodeChars + 1
    end
    return str:sub(1, pos+addPos) .. chr .. str:sub(pos+addPos+1)
end

function InsertChar2(str, chr, pos)
    local offset = 0
    local posWithUnicodeChars = 1
    local skipSecondByteOfUnicodeChar = false
    str:gsub(".", 
        function(c) 
            if posWithUnicodeChars <= pos then 
                -- first byte always >=0xC0 (192). look at function UTF(unicode) local Byte0 = 0xC0 + ...
                if string.byte(c) >= 192 and not skipSecondByteOfUnicodeChar then -- if byte is first part and not second part of unicode char 
                    offset = offset + 1     -- increase offset
                    skipSecondByteOfUnicodeChar = true -- and skip next byte because is second part of unicode char
                else        -- otherwise skip second byte and reset skipSecondByteOfUnicodeChar to false
                    skipSecondByteOfUnicodeChar = false
                    posWithUnicodeChars = posWithUnicodeChars + 1  -- this inc fires all iteration of bytes one by char except unicode char (fires one by two byte)
                end
            end
        end
    )
   return str:sub(1, pos+offset) .. chr .. str:sub(pos+offset+1)
end

```

dk0x:
rdy to merge

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
...
+
+	if (unicode <= 0xFFFF) then
+		local Byte0 = 0xE0 + math.floor(unicode / 0x1000);
+		local Byte1 = 0x80 + math.mod(math.floor(unicode / 0x40), 0x40);
+		local Byte2 = 0x80 + math.mod(unicode, 0x40);
+		return string.char(Byte0, Byte1, Byte2);
+	end;
+
+	if (unicode <= 0x10FFFF) then
+		local code = unicode
```
Sheeo: Please consistently use spaces for indentation

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
+-- dk0x unicode minilib
```
Sheeo: Please avoid using author comments, we have the git history to drill down and find out who wrote the code :)

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
...
+        -- first byte always >=0xC0 (192). look at function UTF(unicode) local Byte0 = 0xC0 + ...
+        if string.byte(byteArrayOfText[posWithUnicodeChars+addPos]) >= 192 then 
+            addPos = addPos + 1
+        end
+        posWithUnicodeChars = posWithUnicodeChars + 1
+    end
+    return str:sub(1, pos+addPos) .. chr .. str:sub(pos+addPos+1)
+end
+
+function InsertChar2(str, chr, pos)
```
Sheeo: This function appears to be unused?

dk0x: we need do a choice from InsertChar1 or InsertChar2 method.

lua/UTF.lua
dk0x: rdy

lua/UTF.lua
dk0x: rdy

ChrisKitching:
Very good news that the lookup table has gone away. Time-constrained right now, so no review at this moment.

dk0x:
ok. i'm vote to InsertChar1 (array) method.
i comment InsertChar2 for possible future use.
pull request completely ready to merge.
thx for work guys.

ChrisKitching:
You need to history-rewrite your branch to tidy up the git history.

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
...
+                    skipSecondByteOfUnicodeChar = true -- and skip next byte because is second part of unicode char
+                else        -- otherwise skip second byte and reset skipSecondByteOfUnicodeChar to false
+                    skipSecondByteOfUnicodeChar = false
+                    posWithUnicodeChars = posWithUnicodeChars + 1  -- this inc fires all iteration of bytes one by char except unicode char (fires one by two byte)
+                end
+            end
+        end
+    )
+   return str:sub(1, pos+offset) .. chr .. str:sub(pos+offset+1)
+end
```
ChrisKitching: Don't leave commented code here... what value is there?

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
...
+    local posWithUnicodeChars = 1
+    while posWithUnicodeChars <= pos do
+        -- first byte always >=0xC0 (192). look at function UTF(unicode) local Byte0 = 0xC0 + ...
+        if string.byte(byteArrayOfText[posWithUnicodeChars+addPos]) >= 192 then 
+            addPos = addPos + 1
+        end
+        posWithUnicodeChars = posWithUnicodeChars + 1
+    end
+    return str:sub(1, pos+addPos) .. chr .. str:sub(pos+addPos+1)
+end
```
ChrisKitching: Does this perform acceptably when typing a long message? Or does it start to feel "choppy"?

lua/ui/lobby/lobby.lua
```diff
@@ -46,6 +46,9 @@ local UnitsAnalyzer = import('/lua/ui/lobby/UnitsAnalyzer.lua')
 
 local IsSyncReplayServer = false
 
+-- dk0x unicode
```
ChrisKitching: In-code credits aren't a thing. We've got `git blame` for that - don't introduce clutter.

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
...
+]]
+
+function AddUnicodeCharToEditText(edit, unicode)
+    if unicode <= 0x7F then return "" end
+    local unicodeChar = UTF(unicode)
+    if unicodeChar ~= "" then 
+        local text = edit:GetText()
+        local charLim = edit:GetMaxChars()
+        if STR_Utf8Len(text) >= charLim then
+            PlaySound(Sound({Cue = 'UI_Menu_Error_01', Bank = 'Interface',}))
```
ChrisKitching: Does this reflect what happens when editing normally, then?

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
...
+            return
+        end
+        local pos = edit:GetCaretPosition()
+        text = InsertChar(text, unicodeChar, pos)
+        edit:SetText(text)
+        --edit:SetText(text .. unicodeChar)
```
ChrisKitching: What value does this commented line add?

Consider adding some descriptive comments, too? There are currently none!

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
...
+        local Byte2= 0x80 + math.mod(code, 0x40);
+        code       = math.floor(code / 0x40)
+        local Byte1= 0x80 + math.mod(code, 0x40);
+        code       = math.floor(code / 0x40)
+        local Byte0= 0xF0 + code;
+
+        return string.char(Byte0, Byte1, Byte2, Byte3);
+    end;
+
+    error 'Unicode cannot be greater than U+10FFFF!'
```
ChrisKitching: You probably want to print this to the debug window using the builtin WARN() function. Perhaps include the problematic value in the printed message?

lua/UTF.lua
```diff
@@ -0,0 +1,87 @@
...
+end
+
+function InsertChar(str, chr, pos)
+    local byteArrayOfText={}
+    str:gsub(".", function(c) table.insert(byteArrayOfText,c) end)
```
ChrisKitching: It feels like what you really want here is a UTF_strlen() function that gives you the byte-length of a UTF string. This InsertChar function is basically a mashup of that with append, which makes it a bit tangly.

ChrisKitching: (That function would probably come  in handy elsewhere...)

lua/ui/lobby/lobby.lua
dk0x: sry. forgot. clear now.

lua/UTF.lua
dk0x: no. its not happen when editing normally.
for insert uniode and normal char need 2 checks in two function.
Did I understand the question correctly?

lua/UTF.lua
dk0x: rdy.
    WARN('Unicode cannot be greater than U+10FFFF! (in utf.lua with param UTF('.. unicode ..'))')

lua/UTF.lua
dk0x: (гu mean STR_Utf8Len)
u dont understand me)
read my comment [here](https://github.com/FAForever/fa/pull/2072#issuecomment-308483849)
we need calculate how many unicode chars (with 2 bytes offset) and normal chars (with 1 byte) string contain BEFORE carret position.
STR_Utf8Len() return size in bytes of ALL string, not before carret pos.

but! You made me think some optimization.

more often user type the message and insert unicode in last position. Then we dont need any operation of inserting. Just glue text and char like text .. unicodechar.

lua/UTF.lua
dk0x: rdy. add some comments. remove --edit:SetText(text .. unicodeChar)

lua/UTF.lua
dk0x: rdy

lua/UTF.lua
dk0x: its does.
but i noticed some strange.
if we type long msg like йячсмицукенгшщзхъ\эждлорпайцукорпровыфячффывааппррсмиячсмиттячтьчсчясчывбю.апрвапуке
and press enter in chat this message will look like 
oqvb9dv_szocfxfma5rhxw 1
its happen because hardcoded moho.edit_methods split long message like 
(after pressing Home and carret go to first pos)
5gjsxuhwst6rgjx3odzv7g 1
(after pressing End and carret go to last pos)
hnggqc9gqygqceqjkruq_g 1

bacause its operation hardcoded i think i dont fix it.

lua/UTF.lua
dk0x: rdy

dk0x:
add some optimization and comments. clear author.
my bad descrition of change :DDD
d37zvo-2sta_-wvdtnqbla 1

some needs to fix [comment here](https://github.com/FAForever/fa/pull/2072#discussion_r122916949)
but maybe its imposible to fix because its hardcoded in moho.edit_methods.
let me time to think out it.

dk0x:
OH GOD!
i find some! STR_Utf8SubString!
time to test InsertChar with STR_Utf8SubString function.
wait...

ChrisKitching:
> i find some! STR_Utf8SubString!

That was in the binary all along?

ChrisKitching:
Oh yeah. Right here:
http://supcom.wikia.com/wiki/LUADOC_1.0.3223

Derp.

dk0x:
ok. some problem with string.sub and string.gfind in \lua\maui\text.lua
busy now. when I'll come back i think i'l fix it.
wait.

dk0x:
fuck :D I don't understand github
well... maybe i'm just open new pull request? so many changes.
i'm fix text.lua problem with unicode word splitting.

dk0x:
m3ru5k7_qeo3ufpaj8wcia 1

test on
`asd1asd2asd3asd4asd5asd6asd7asd8asd9asd0asd1asd2asd3asd4asd5asd6asd7asd8asd9asd0asd1asd2asd3asd4asd5asd6asd7asd8asd9asd0asd1asd2asd3asd4asd5asd6asd7asd8asd9asd0asd1asd2asd3asd4asd5asd6asd7asd8asd9asd0`
`фыв1фыв2фыв3фыв4фыв5фыв6фыв7фыв8фыв9фыв1фыв2фыв3фыв4фыв5фыв6фыв7фыв8фыв9фыв1фыв2фыв3фыв4фыв5фыв6фыв7фыв8фыв9фыв1фыв2фыв3фыв4фыв5фыв6фыв7фыв8фыв9фыв1фыв2фыв3фыв4фыв5фыв6фыв7фыв8фыв9фыв0`

all work correct

ChrisKitching:
Great work so far :D

Very happy to see this long-awaited feature on the verge of completion.

It looks like you had a git accident and now you need to read this:
https://www.atlassian.com/git/tutorials/rewriting-history

dk0x:
thx for reply 👍 
i'm already rebase (history-update?) my dk0x/fa fork? or no? i'm merge all changes from FAForever/fa to dk0x/fa after my comment [this](https://github.com/FAForever/fa/pull/2072#issuecomment-309716243)

or just tell me click-click instructions :D

ChrisKitching:
A GitHub pull request is just a request to merge from one branch into another. The diff shown is the difference between the feature branch and the base branch you've requested merge with.

You can force-push the feature branch as much as you want and the PR will update.


For some reason, you've sucked in history-rewritten copies of 130-ish commits from develop onto your branch. You will need to interactively rebase your feature branch on top of the target branch.

dk0x:
just tell me click-click! :D
i'm install git desktop, clone dk0x/fa to desktop.
maybe u ell me cmd line?

dk0x:
somebody help me with step-by-step instruction plz.

aeoncleanse:
Just leave it, i'll rebase it for you tonight

aeoncleanse:
https://github.com/FAForever/fa/pull/2105

Rebased onto develop and pushed to an fa repo branch. 

@dk0x - Can you double check all your changes are here, that I've not missed something?
@Sheeo @ChrisKitching - Can you check if there was anything you wanted done that got lost in the commit chaos? :)

Issue #2040, opened 2017-05-28
Collection of formation issues


MrNukealizer:
Here's a place to keep track of problems with the current formations that might need work:
Edit: list moved to PR #2073.

PhilipJFryFAF:
If you change the t3 sub you might want to watch for where it ends up in the formation to avoid it taking damage from surface units getting hit by projectiles with enough aoe to harm a submerged sub

Exotic-Retard:
actually i think thats optimal behavior - the issue with the t3 sub is a separate one and can be addressed separately as well, while in terms of formation any t3 subs need to be at the front (otherwise they cant fire their torpedoes), and t3 strategic subs need to be at the very back, while the cyb one should be somewhere at the back middle since it has torpedoes.

in any case high level games involve a lot of micro which separates the subs.

it would be cool if the subs were offset by half a unit width so they are not under the ship directly but in between them, it also makes it easier for strategic icons. but im not sure how difficult that is

MrNukealizer:
It shouldn't be too hard to remove one column of subs to stagger them between the surface units.

MrNukealizer:
Status update: I've been digging around in the formations a bit and there's a lot we can do here. All the stuff on the checklist is doable, but we can also make a lot of more significant changes if needed, such as adding new formations, changing the shape of existing formations (like making air formations actually useful) and more.

speed2CZ:
We should spread out the air formations a bit, so the air units wont get killed by flak and sams so easily

MrNukealizer:
Status update: After a couple failed attempts at fixing unit spacing I've come up with [this](http://i.imgur.com/Lkdsj6Z.jpg), which looks promising. Now to make it work for naval units and tackle the other problems.

PR #2073, opened 2017-06-14
Formation Improvements

MrNukealizer:
Here's my work on improving formations. This is a work in progress and addresses the points in #2040. Any comments or suggestions are welcome, but please look at the checklist first.

Progress:
- [x] Stop large units from causing smaller units to spread out too much:
    - [x] Land
    - [x] Sea
    - [x] Air (done but they share space with the main formation since they normally have different altitude.)
    - [x] Guard (done but large units can get kind of cramped if there are also smaller units.)
    - [x] Make mixed-size formations look reasonably good in all situations (needs testing).
- [x] Update guard formation:
    - [x] Decrease the unit spread so they're not all over the place.
    - [x] Improve spacing of units in the outermost ring when there aren't enough for the full ring.
    - [x] Spread shields more evenly across the formation to protect the other units instead of clumping up.
- [x] Update sub formations to put them between surface units instead of directly under them.
- [x] Update air formations:
    - [x] Make formations centered instead of offset to the right of the destination.
    - [x] Adjust spacing (maybe even shape?) to reduce vulnerability to splash.
    - [x] Increase formation width (3/2 chevrons -> 3/4 or 5/4?) to accommodate the large groups often used at once.
- [x] Cache formation results so they aren't recalculated many times in a row for the same units.
    - [x] Extend caching to include when mixed air/land formations are split into two calls (might not be possible or efficient).
- [x] Fix unit categories:
    - [x] T3 sub hunter counts as a nuke sub rather than a torpedo sub.
    - [x] Scathis is excluded from all groups and thus never gets a spot in the formation.
    - [x] Seraphim sniper bots seem to be convinced they're T3 mobile arty, so they're assigned bot spots but steal arty spots and leave the arty with nowhere to go. (FAForever/faf-coop-maps#215)
    - [x] Sniper bots get normal bot positions, which are often too close to the front for such squishy units.
    - [x] Experimentals without DIRECTFIRE generally don't match any of the categories and get stuck at the back.
    - [x] Bouncer has a 2x2 footprint unlike every other T1-T3 land unit.
- [x] Cleanup:
    - [x] Remove unnecessary logging.
    - [x] Fix comments and formatting.

### DIFF
--- a/lua/formations.lua
+++ b/lua/formations.lua
@@ -369,12 +369,27 @@ local AttackChevronBlock = {
     RepeatAllRows = false,
     HomogenousBlocks = true,
     { ChevronSlot, },
-    { ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, }, -- 1 -> 3 at 20 units
     { ChevronSlot, ChevronSlot, ChevronSlot, },
-    { ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, }, -- 3 -> 5 at 60 units
     { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
     { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, }, -- 5 -> 7 at 170 units
     { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, }, -- 7 -> 9 at 390 units
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, },
+    { ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, ChevronSlot, }, -- 9 -> 11 at 760 units
 }
 
 local GrowthChevronBlock = {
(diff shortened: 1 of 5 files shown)


lua/formations.lua
```diff
@@ -834,75 +910,141 @@ end
...
+						else
+							if whichCol == 1 then
+								xPos = 0
+							else
+								xPos = math.ceil(((whichCol-1) /2)) + offsetX
+								if not (math.mod(whichCol, 2) == 0) then
+									xPos = xPos * -1
+								end
+							end
+						end
```
CookieNoob: convert these tabs to 4 spaces to match the rest of the codebase

MrNukealizer: That's what the "Fix comments and formatting" checkbox is for. I can't work with 4 tabs so I have to do a replace all every time I'm done editing, which is best done when I'm actually done editing.

CookieNoob: ah ok nvm then :)

ChrisKitching:
> The whole file needs a formatting pass to match the rest of the codebase, which I'll do after getting things working.

Make sure that doesn't end up in its own commit.

lua/formations.lua
```diff
@@ -1,15 +1,15 @@
...
+-- **
+-- **  File     :  /cdimage/lua/formations.lua
+-- **  Author(s):
+-- **
+-- **  Summary  :
+-- **
+-- **  Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
+-- ****************************************************************************
+--
+-- Basic create formation scripts
```
ChrisKitching: It's helpful to do this sort of syntax conversion stuff in its own commit so reviewers can clearly see what your functional changes are.

This stuff is nice to have (especially when you start editing a new file), but when it's mixed in with a functionally-significant commit it really makes review harder. This diff is _massive_, and most of it is non-functional changes.

lua/formations.lua
```diff
@@ -729,13 +749,23 @@ function GrowthFormation(formationUnits)
 end
 
 function BlockFormation(formationUnits)
+	LOG("BlockFormation")
```
ChrisKitching: Accidentally left-in debug logging?

lua/formations.lua
```diff
@@ -729,13 +749,23 @@ function GrowthFormation(formationUnits)
...
+	if LastFormation == 'BlockFormation' and CheckSameUnits(formationUnits) then
+		return FormationPos
+	end
+	
+    FormationPos = {}
```
ChrisKitching: You indented with spaces :astonished:

lua/formations.lua
```diff
@@ -793,39 +831,85 @@ function CircleFormation(formationUnits)
...
+				unitsPerShield = ringChange / shieldsInRing
+				nextShield = unitsPerShield - 0.01 -- Rounding error could result in missing a shield if nextShield is supposed to equal ringChange.
+			end
         end
-        offsetX = sizeMult * math.sin(lerp(unitCount/ringChange, 0.0, math.pi * 2.0)) # + math.pi / 16)
-        offsetY = sizeMult * math.cos(lerp(unitCount/ringChange, 0.0, math.pi * 2.0)) # + math.pi / 16)
-        #LOG('*FORMATION DEBUG: X=' .. offsetX .. ', Y=' .. offsetY)
-        table.insert(FormationPos, { offsetX - 10, offsetY, categories.ALLUNITS, 0, rotate })
+		local ringFraction = unitCount/ringChange
+        offsetX = sizeMult * math.sin(lerp(ringFraction, 0.0, math.pi * 2.0))
```
ChrisKitching: Even github is rendering this indentation wrong o.0 .

lua/formations.lua
```diff
@@ -834,75 +918,141 @@ end
...
     while unitsList.UnitTotal >= i do
         if whichCol > currRowLen then
             if whichRow == numRows then
                 whichRow = 1
-                if formationBlock.RowBreak then
-                    formationLength = formationLength + 1 + formationBlock.RowBreak
-                else
-                    formationLength = formationLength + 1
-                end
+                formationLength = formationLength + 1 + (formationBlock.RowBreak or 0) + (formationBlock.LineBreak or 0)
```
ChrisKitching: This change may make it shorter, but it definitely doesn't make it clearer.

Fewer lines isn't necessarily better. If you want to simplify the original expression, subject it to common subexpression elimination, like this:

```lua
formationLength += 1
if formationBlock.RowBreak then
     formationLength += formationBlock.RowBreak
end
```

At this point, it becomes apparent that we could further simplify it by making `RowBreak` always exist, but sometimes be zero (then it'd just be `+= 1 + formationBlock.RowBreak`). That might be worth investigation. The code has enough nasty null checks already ;)

lua/formations.lua
```diff
@@ -834,75 +918,141 @@ end
...
+								end
+								if blocked then
+									continue
+								end
+							end
+							fs = k
+							groupData = v
+							break
+						end
+					end
```
ChrisKitching: Nesting in the above block gets extremely deep. Consider introducing a function and using early returns to untangle it a bit.

lua/formations.lua
```diff
@@ -834,75 +918,141 @@ end
...
+                    if groupData then
+						local offsetX = 0
+						local offsetY = 0
+						
+						if size > 1 then
```
ChrisKitching: This one would be a great candidate for early return.

lua/formations.lua
ChrisKitching: FYI, you can achieve that roughly as follows (assuming this is the top commit)

```
git reset --soft HEAD~
git reset
git add -p
# Select non-functional changes in the ensuing interactive mode, use `h` for help

git commit
git add . # selects everything else
git commit
```

A similar approach can be employed in combination with `git rebase -i` to apply that to commits behind the top commit.

lua/formations.lua
CookieNoob: `+=` is not defined in LUA, is it?

ChrisKitching:
Woot, another much-requested feature.

lua/formations.lua
MrNukealizer: Yeah, I originally tried to work on the file as it was. Eventually the formatting and # comments became too hard to deal with, so I reformatted it after already making significant functional changes. If I had known it would be necessary, I definitely would've done it separately first.

lua/formations.lua
MrNukealizer: Yep. There's a bit more of that commented out in the places I'm currently working, but I completely forgot about BlockFormation and CircleFormation because they're never used.

lua/formations.lua
MrNukealizer: I'm not sure what you mean but it probably wasn't my doing.

lua/formations.lua
MrNukealizer: So you would rather have ```formationLength = formationLength + 1
if formationBlock.RowBreak then
     formationLength = formationLength + formationBlock.RowBreak
end
if formationBlock.LineBreak then
     formationLength = formationLength + formationBlock.LineBreak
end
``` instead of `formationLength = formationLength + 1 + (formationBlock.RowBreak or 0) + (formationBlock.LineBreak or 0)`? It seems like a far more verbose way of saying what's obvious in the one line.

lua/formations.lua
MrNukealizer: Yeah, I'll look into it. That part seemed kind of awkward because most such functions would need quite a few variables from inside the main function.

MrNukealizer:
Thanks for the review. I'm good at making code that works; not so much for making code that's nice to read. I'm trying to work on that and your advice should be helpful.

lua/formations.lua
ChrisKitching: Does lua really not have +=? Eep. It's been a while ;)

I seem to have derped and not noticed that both RowBreak and LineBreak are a thing, either.

I consider the "or 0" thing a bit of an antipattern - this notion of having nils all over the place so the code needs to be littered with null checks is a bit icky. Doesn't it make sense to just insist that RowBreak/LineBreak are always present, but possibly zero? (unless you can make them obsolete and remove them outright).

But yes, your expression is basically fine, then, sorry for the extreme derp. Bonus points for eliminating nullity so you can simplify it further, and even more bonus points for making RowBreak go away - if that turns out to be possible.

One thing I've seen before, and which is what caused me to jump on your change in the first place, is where people think it's "better" to replace a sequence of ifs with a horrifying nested ternary expression. Null-safe arithmetic in lua isn't that, however.

MrNukealizer: Yeah, I know what you mean. I might get a little close by performing math on the result of `math.mod(var, 2)` but that's normally when there's a very simple mathematical difference between the even and odd versions.

MrNukealizer:
I think that's it. All my goals have been achieved except giving sniper bots positions farther back, but that won't work because they have the same categories as frontline bots.

Time for review and public testing, unless I missed something obvious...

aeoncleanse:
Feel free to give them a new "SNIPER" category if that will work.

MrNukealizer:
That should work. I'll try it later.

MrNukealizer:
Ok, adding the SNIPER category worked great. That allows for decent positioning of long range direct fire units. I believe this PR is once again ready for review and testing.

PR #2074, opened 2017-06-14
Make construction pause button change with skin

### LINKED ISSUE #2031: Game: The pause button color doesn't change when you switch faction UI layout
pause
So when you switch faction ui layout in game the two vertical lines don't chane colour.

speed2CZ:
Fixes #2031


ChrisKitching:
You don't seem to have added per-faction `/game/construct-sm_btn/pause_on.dds` files. Does the game already contain those?

CookieNoob:
yes, this problem only occurred when switching the skin of the UI ingame (with the debug keys ALT+left/right arrow)

speed2CZ:
Ofc it does, the different color buttons exist and it works fine if you dont change the skin during the game. So it just needed to load it via the right function

PR #2078, opened 2017-06-19
Fix CZAR targeting hover units with depth charges

MrNukealizer:
Currently the CZAR launches depth charges at hover units, which then just circle around harmlessly below the water.


MrNukealizer:
I don't know. I just changed the one thing that was different from torpedo bombers. I suppose you could exclude AIR as well.

PR #2091, opened 2017-06-28
brings names in line

PhilipJFryFAF:
http://spooky.github.io/unitdb/#/URA0303,UEA0303,UAA0303,XSA0303
the missing "-" for the cybran ASF bothers me more than i'd like to admit


speed2CZ:
You need to change it in the loc files as well, since this string gets replaced by the localized version
https://github.com/FAForever/fa/blob/develop/loc/US/strings_db.lua#L6027-L6028

PhilipJFryFAF:
Thank you for the info. Changed the localized file as well.

Uveso:
@PhilipJFryFAF 
You added the "-" to the unitblueprint BUT you removed it from the localisation file.
Now we have a different translation then the original text.

I gues it makes more sense to add the minus in the localisation file also.
Then it will like the other.
Just to compare:
```
uaa0303_desc="Air-Superiority Fighter"
uaa0303_help="Air-Superiority Fighter"

uea0303_desc="Air-Superiority Fighter"
uea0303_help="Air-Superiority Fighter"

xsa0303_desc="Air-Superiority Fighter"
xsa0303_help="Air-Superiority Fighter"

ura0303_desc="Air Superiority Fighter"
ura0303_help="Air Superiority Fighter"
```

PhilipJFryFAF:
thanks for checking
reverted my removal of the dash for aeon ASF and added it for cybran in the localization file

PhilipJFryFAF:
removed all hyphens from the ASF since both versions seem to be ok and the hyphen messes up the cheat menu search bar so not having it seems to be the better choice

PR #2094, opened 2017-06-29
Fix for factory build templates icons

Uveso:
Fixed: Factorytemplates was not supporting icons from mod units.

factorytemplateicon

Fixed: Selecting a new icon to the template was not working because of a
different array structure from normal buildtemplates.

Structure of normal engineer buildtemplate:
```  templateData={ 3, 3, { "ueb2101", 936, 0, 0 }, },```

And this is how the factory buildtemplate looks:
```templateData={ { count=5, id="brnt1mt" } },```

To get the icon from the factory buildtemplate we need to use `templateData.id` instead of getting the first array entry with `templateData[1]`

### DIFF
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -1587,8 +1587,14 @@ function CreateTemplateOptionMenu(button, templateObj)
                 local controls = {}
                 for _, entry in theTemplate.templateData do
                     if type(entry) == 'table' then
-                        if not contents[entry[1]] then
-                            contents[entry[1]] = true
+                        if entry.id then
+                            if not contents[entry.id] then
+                                contents[entry.id] = true
+                            end
+                        else
+                            if not contents[entry[1]] then
+                                contents[entry[1]] = true
+                            end
                         end
                     end
                 end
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: lua/ui/game/construction.lua [hunk 1]
```diff
@@ -1587,8 +1587,14 @@ function CreateTemplateOptionMenu(button, templateObj)
...
-                            contents[entry[1]] = true
+                        if entry.id then
+                            if not contents[entry.id] then
+                                contents[entry.id] = true
+                            end
+                        else
+                            if not contents[entry[1]] then
+                                contents[entry[1]] = true
+                            end
                         end
```
ChrisKitching: You could rephrase this more simply:

```lua
let targetID = entry.id or entry[1]
if not contents[targetID] then
    contents[targetID] = true
end
```

And, since `contents` is being used as a set, the only value it'll ever have is `true`, so you don't even need the if:

```lua
contents[entry.id or entry[1]] = true
```

Uveso: Thats a good idea!
rewritten, rebased and tested. Works fine :D

PR #2096, opened 2017-07-01
Notify: Add the ability to limit the number of times a message is shown and fix mod breaking

MrNukealizer:
This stops showing non-enhancement messages after a certain number. That way if you start mass producing experimentals or factory HQs it won't spam allies for every single one completed.

Currently the limits are hard-coded at 1 message per factory HQ and 3 per nuke/arty/experimental. If the numbers should be changed let me know. I looked at allowing the numbers to be changed in the customizer, but I can't think of a way to do it that wouldn't just clutter up the window.

### DIFF
--- a/etc/faf/blacklist.lua
+++ b/etc/faf/blacklist.lua
@@ -120,10 +120,6 @@ Blacklist = {
     ['ecbf6277-24e3-437a-b968-EcoManager-v6'] = UPGRADE,
     ['ecbf6277-24e3-437a-b968-EcoManager-v7'] = UPGRADE,
 
-    -- Selection Cost UI (old versions that break the UI)
-    ['2018eaac-e8ed-11e4-b02c-1681e6b88ec4'] = UPGRADE,
-    ['2018eaac-e8ed-11e4-b02c-1681e6b88ec5'] = UPGRADE,
-
     -- Supreme economy (old ones)
     ['f8d8c95a-71e7-4978-921e-8765beb328e8'] = UPGRADE,
     ['89BF1572-9EA8-11DC-1313-635F56D89591'] = UPGRADE,
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2047,6 +2047,7 @@ Unit = Class(moho.unit_methods) {
     end,
 
     OnFailedToBeBuilt = function(self)
+        self:SendNotifyMessage('cancelled')
         self:ForkThread(function()
             WaitTicks(1)
             self:Destroy()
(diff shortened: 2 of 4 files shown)


aeoncleanse:
I think I know how we'll do this:

For Factories, no limit, but it should only trigger if you're _upgrading_ not if you're building a HQ from scratch. Most people only want one or two HQs after all.

For Exps, I think on a per-unit-type basis, can we limit it to say, 3 notifications every 5 mins of gametime? So if you start/finish the 4th ML inside a 5 minute period, it doesn't spam.

The alternative is that we just tell people to toggle it if it annoys them.

MrNukealizer:
How do you build an HQ from scratch? As far as I'm aware, they're always made by upgrading from a T1 factory. The reason I added this limit is specifically to address complaints about getting too many notifications when someone upgrades all their factories to HQs instead of support factories.

As for the 5 minute timer, that seems like it could still get spammy, but much better than nothing.

On another note, it seems I was barking up the wrong tree with mods breaking the UI. Simply moving an import fixes all that.

CookieNoob:
@aeoncleanse you cant build an HQ directly. Building higher tier factories (besides support factories) is only possible in BlackOps. I agree with MrNukealizer that there should be a limit. It is completely irrelevant if someone upgrades an air factory to t2 for some transports 40 minutes into a 5v5 game. Also these notifications for experimentals are nice for early experimentals (i.e. rushed or counter), but later in the game they are not necessary or useful anymore.

About the limits: 3 messages per 5 minutes? I.e. when you build more than 3 t3 arties in 5 minutes it doesnt show a message anymore? That limit seems kinda useless.

I suggest to show a message for:
- the first HQ factory upgrade per type (not necessarily per player), notifications about t2 air or t3 air are really important in teamgames.
- all nukes/arties or the first 3 (there are not that many)
- the first 3 experimentals

after these, the game progressed to a point where further notifications dont give any additional information.

MrNukealizer:
I'd say some more experimental notifications could be useful. It's not unreasonable for your team to have 3 Monkeylords or the like then later start production of a Paragon or some other change of plans.

CookieNoob:
sure a notification for a game ender should always be shown (paragon/mavor/yolo/salvation). However, these notifications have the purpose to tell your teammates that you started something and that they should assist it/act accordingly. The purpose is to avoid 3 monkeys being build at the same time. Similarly, the original idea of notify was to avoid having several ppl upgrading RAS at the same time.

Imo we dont need infinite notifications about experimentals, only the first ones are important.

aeoncleanse:
Been thinking about this all day. So... What about this:

For HQs, have a per-player limit of 1 per type. So it only notifies the first time a player unlocks one of the 6 techs. The purpose of Notify is to _notify_, any fewer than this and we may as well remove the feature. Remember --- Showing Factory Upgrades defaults to OFF

For Exps, have a per-player, per-Exp limit of 1, with the exception of the game-enders. Same as above I guess, and FYI it also defaults off. T3 Arty can have the same. I suppose in the end, it's a limit of 1 for anything but ACU Upgrades.

PR #2101, opened 2017-07-04
fixes the distance for attackmove

PhilipJFryFAF:
Currently the T3 Subhunter tends to stop at 32 range (it's torp def range) when using the attack move command which prevents it from using it's range properly.
This change should fix that problem.


PhilipJFryFAF:
before: https://puu.sh/wBuzL/7591a34246.png
after: https://puu.sh/wBuGH/cac1be9fe9.png

PR #2123, opened 2017-07-25
Add random modes with revealed spawn locations

MrNukealizer:


MrNukealizer:
It seems the automated checks don't like the C-style '|' bitwise OR. Is there a better way to do that in lua?

PR #2124, opened 2017-07-25
remove "quick tip" prefix

### LINKED ISSUE #2048: Improve quick tip
I love quick tips, but there are two things I think can/should be improved:

1. Don't flash. Not only is it very annoying (to me), but also does it reduce the time the quick tip is actually readable to about 50%.
2. Remove the "Quick tip:" prefix. Not only are Players used to the concept of tips on loading screens, but it is also very obvious from the text itself. So announcing "the following text is a quick tip" is pretty redundant thus increasing read time and consuming precious space without any use.

CookieNoob:
removes the quick tip prefix of the quick tips since its quite obvious
that they are tips.
fixes #2048

### DIFF
--- a/lua/ui/game/gamemain.lua
+++ b/lua/ui/game/gamemain.lua
@@ -303,8 +303,7 @@ local function LoadDialog(parent)
                 local tipsTbl = import('/lua/ui/help/loadingtips.lua').Tips
                 local tipsSize = table.getn(tipsTbl)
                 while WorldIsLoading() do
-                    local tipText = LOCF('%s', tipsTbl[Random(1, tipsSize)])
-                    control:SetText(tipText)
+                    control:SetText(LOC(tipsTbl[Random(1, tipsSize)]))
                     control:SetDropShadow(true)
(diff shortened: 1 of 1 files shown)


lua/ui/game/gamemain.lua
```diff
@@ -303,7 +303,7 @@ local function LoadDialog(parent)
                 local tipsTbl = import('/lua/ui/help/loadingtips.lua').Tips
                 local tipsSize = table.getn(tipsTbl)
                 while WorldIsLoading() do
-                    local tipText = LOCF('%s: %s', '<LOC loadingtip_0000>Quick Tip', tipsTbl[Random(1, tipsSize)])
+                    local tipText = LOCF('%s', tipsTbl[Random(1, tipsSize)])
```
speed2CZ: This will do the job,
```LUA
LOC(tipsTbl[Random(1, tipsSize)])

CookieNoob: actually, storing the text in the local isnt needed either

PR #2130, opened 2017-07-30
fix modmanager handling of incomplete modinfo files

### LINKED ISSUE #2128: bugged mod manager
noticed that the mod manager in the lobby in the 3686 patch bugs out (shows no mods at all or the list expands over the borders)

forgedalliance_2017-07-27_20-44-02

[logs.zip](https://github.com/FAForever/fa/files/1183958/logs.zip)

CookieNoob:
the mod manager sorts the mods by UI and SIM mods. That requires to have the UI_Only parameter set in the modinfo file.

Some mods dont have this parameter set at all and thus break the modmanager.

fixes #2128

### DIFF
--- a/lua/ui/lobby/ModsManager.lua
+++ b/lua/ui/lobby/ModsManager.lua
@@ -487,7 +487,7 @@ function RefreshModsList()
         mod.title = GetModNameVersion(mod)
         if mod.ui_only then
             mod.type = 'UI'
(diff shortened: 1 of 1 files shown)


ChrisKitching:
Bonus points for finding and exterminating all instances of `x == false` and `x == true`. It's such an antipattern... :D

PR #2144, opened 2017-08-13
Fix Percival LOD model not turning gun. Fixes #2075

### LINKED ISSUE #2075: LOD cutoff for percy too small
the point where the high quality model is swiched to the low quality one is on a too low zoom level, so that the actual model is really visible (and not hidden by strategic symbols). Since the low quality model doesnt turn its gun and the barrel points always forward, it looks really silly when its in combat.

grafik

one zoom level closer:
grafik

after increasing the LOD cutoff values and the icon fade in level (to see the model better) it seems the model itself is turning, but the weapon isnt. (all of these percies are looking downwards, the barrel is shown towards the top though)
grafik

from a bit closer:
grafik

aeoncleanse:


CookieNoob:
works nicely gj!

PR #2150, opened 2017-08-16
Fix for modded units with no factionname

Uveso:
http://forums.faforever.com/viewtopic.php?f=3&t=15002&p=153183#p153183
One of the last patch broke the BrewLan mod.

Just a small issue;
some brewLan units don't have a factionname, and causes an error in this line:
`bp.FactionCategory = string.upper(bp.General.FactionName)`

### DIFF
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -493,7 +493,7 @@ function PreModBlueprints(all_bps)
             end
         end
 
-        bp.FactionCategory = string.upper(bp.General.FactionName)
+        bp.FactionCategory = string.upper(bp.General.FactionName or 'Unknown')
 
         -- Mod in AI.GuardScanRadius = Longest weapon range * longest tracking radius
         -- Takes ACU/SCU enhancements into account


aeoncleanse:
Fix the mod... No unit should EVER have no faction. That's completely dumb.

speed2CZ:
I have to agree with icedreamer here, there is some minimum that needs to be done to make the unit working and in my opinion, defining faction in the blueprint should be one of those.
And from what Ive seen in the code, the unit's faction is used in more than one place and it can break some other things anyway.

The-Balthazar:
Except when it's a dummy unit, isn't supposed to be buildable or interactable by players, and doesn't need said data key. You shouldn't just assume things.

aeoncleanse:
FA has dummy units. They have factions.

The-Balthazar:
Which is fine if they are factional dummy units, and you don't mind them having defined meshes for when they are under construction and other things.

Uveso:
Well, the game don't need any faction for a unit. (Why forcing it after 10 years ?)

The only reason for this is the unitanalyzer.

And the unitanalyzer is checking the faction in function 'GetUnitFaction(bp)'

So this is the way we "analyze" a unit's faction now:
1. It will print a debugtext to inform us, if the faction and category.faction is different
2. it will crash to inform us, there is no faction.

Isn't this a little bit pointless ?

In my eyes clearly our fault, not from the mod.

aeoncleanse:
The change was made with the assumption that no unit should ever have no faction, because there are about 50 places in the unit code which would just outright break if it didn't. Since dummy units have no reason to not have a faction, may as well enforce it for consistency. 

If he insists that his dummy units can't have a faction set (Could be right, I've not checked)... It's a mod. He can just override the function like this. Doesn't belong maingame IMO, and there's no handling for an 'unknown' down the chain. It expects a faction value.

Uveso:
'unknown' is just a placeholder. You can write in whatever you want, because it will not be used on those units.

BrewLan was running for ages with those units without a factionName.
So we can assume that units don't necessarily need a faction name.

And since BrewLan, units have a reason to not have a faction. We just don't knew it ;)

Sure, we can ask The-Balthazar that he needs to change his mod.

To be compatible with the unitanalyzer!
(and not the game itself)

But this is the point where i say, we have to fix the unitanalyzer, because only the unitanalyzer needs a faction name, not the game.

Uveso:
While testing a patch for this inside the BrewLan mod i stucked.

The Unitmanager is also using FactionNames and i can't hook into it because the mod is not active game-lobby.

speed2CZ:
Alright, I've changed my mind, the change sounds reasonable

aeoncleanse:
So be it. If this breaks things, it's not on me :p

Issue #2149, opened 2017-08-16
Use widescreen faction logos in the lobby


speed2CZ:
http://forums.faforever.com/viewtopic.php?f=2&t=7032&start=110#p99675

PR #2151, opened 2017-08-16
replace background faction images with widescreen versions

PhilipJFryFAF:
got 4 images from Blackster aka Ninrai from http://forums.faforever.com/viewtopic.php?f=2&t=7032&start=110#p99675
need to make/get one more for random factions


CookieNoob:
fixes #2149

speed2CZ:
The best way is having that directly in the comment of the commit

PhilipJFryFAF:
ty for the info about the issue reference i'll think of it next time i make an issue
also added the random image
thanks again to ninrai for making it

PR #2152, opened 2017-08-16
change salvation to be experimental

### LINKED ISSUE #1052: Aeon Rapid Arty (Salvation) should be categorized as Experimental instead of Tech3
Aeon Rapid Arty (Salvation) should be categorized as Experimental instead of Tech3 because it costs much more than Tech3 Heavy Arty and it does damage roughly equallyvalent to other experimental arty units in other factions: Scathis and Mavor

Labels UI, bug, low

CookieNoob:
fixes #1052

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,3 +1,12 @@
+Patch 3688
+============================
+### Other
+- Changed Salvation to be categorized as an experimental instead of a t3 building
+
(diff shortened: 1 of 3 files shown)


speed2CZ:
Hopefully this won't mess with AI

PR #2153, opened 2017-08-16
counterintel disableing intel unintentionally

### LINKED ISSUE #896: Monkeylord sonar turns off when out of energy
This is probably a problem with all units that have intel and some form of anti-intel (stealth, cloak, etc.). Because there is only one FreeIntel property in the blueprint, all forms of Intel (and, stupidly, anti-intel) are caught in it.

Turning off the Monkeyloard's stealth turns off all of its energy cost and the sonar comes back on. I believe that this is at least extremely counter-intuitive and should be changed.

It would be very silly if a cybran ACU's omni turns off because the player is out of energy. I haven't verified that that happens but it probably does.

CookieNoob:
This commit fixes the following two issues:

The monkeylord loses its sonar capabilities in case of a powerstall when
the stealthfield is active. If it is deactivated, the sonar continues to
work. fixes #896

If the cybran acu, sacu or the uef sacu has a counter intel upgrade
(stealth, cloak, jamming) and the player powerstalls, the omni of the
unit will be deactivated.

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2396,11 +2396,9 @@ Unit = Class(moho.unit_methods) {
         local function DisableOneIntel(disabler, intel)
             local intDisabled = false
             if Set.Empty(self.IntelDisables[intel]) then
-                local intel_bp = self:GetBlueprint().Intel
-                for _, inteltype in intel_bp.ActiveIntel do
-                    if intel == inteltype then
-                        return
-                    end
+                local active = self:GetBlueprint().Intel.ActiveIntel
+                if active and active[intel] then
+                    return
                 end
                 self:DisableIntel(intel)
 
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2397,11 +2397,9 @@ Unit = Class(moho.unit_methods) {
             local intDisabled = false
             if Set.Empty(self.IntelDisables[intel]) then
                 local intel_bp = self:GetBlueprint().Intel
-                if intel_bp.AlwaysActiveIntel then
-                    for _, inteltype in intel_bp.ActiveIntel do
-                        if intel == inteltype then
-                            return
-                        end
+                for _, inteltype in intel_bp.ActiveIntel do
+                    if intel == inteltype then
+                        return
                     end
                 end
                 self:DisableIntel(intel)
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2396,11 +2396,19 @@ Unit = Class(moho.unit_methods) {
         local function DisableOneIntel(disabler, intel)
             local intDisabled = false
             if Set.Empty(self.IntelDisables[intel]) then
+                local intel_bp = self:GetBlueprint().Intel
+                if intel_bp.AlwaysActiveIntel then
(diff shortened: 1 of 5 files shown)


units/UEL0301/UEL0301_unit.bp
```diff
@@ -694,6 +694,8 @@ UnitBlueprint {
         UnitWeight = 0,
     },
     Intel = {
+        AlwaysActiveIntel = true,
```
aeoncleanse: Do we need this flag? If the ActiveIntel entry exists, it will return true. If not, false.

CookieNoob: I introduced it for performance reasons as a simple if is faster than a loop... but I can change it easily :)

lua/sim/Unit.lua
```diff
@@ -2397,11 +2397,9 @@ Unit = Class(moho.unit_methods) {
             local intDisabled = false
             if Set.Empty(self.IntelDisables[intel]) then
                 local intel_bp = self:GetBlueprint().Intel
-                if intel_bp.AlwaysActiveIntel then
-                    for _, inteltype in intel_bp.ActiveIntel do
-                        if intel == inteltype then
-                            return
-                        end
+                for _, inteltype in intel_bp.ActiveIntel do
```
aeoncleanse: OK, I actually meant doing this:
```
if intel_bp.ActiveIntel then
    for _, inteltype in intel_bp.ActiveIntel do
... ...```

That is then cleaner. You can do it this way, but you have to actually do 
```
for _, inteltype in intel_bp.ActiveIntel or {} do```

To avoid crashing when the table isn't there, but in those cases creating a table is slower than checking the existence of it. So yeah, just use my suggestion at the top :)

aeoncleanse: Even _faster_ is to make a local of intel_bp.ActiveIntel to drill down another layer.

aeoncleanse:
See slack for ffinal feedback

units/UEL0301/UEL0301_unit.bp
```diff
@@ -694,7 +694,9 @@ UnitBlueprint {
         UnitWeight = 0,
     },
     Intel = {
-        ActiveIntel = {'Omni' = True},
+        ActiveIntel = {
+            Omni = True,
```
speed2CZ: `true` not `True`

units/URL0001/URL0001_unit.bp
```diff
@@ -697,7 +697,10 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        ActiveIntel = {'Omni' = True, 'Sonar' = True},
+        ActiveIntel = {
+            Omni = True,
+            Sonar = True,
```
speed2CZ: `true` not `True`

CookieNoob:
I should really stop doing stuff so early in the morning

units/URL0402/URL0402_unit.bp
```diff
@@ -262,7 +262,9 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        ActiveIntel = {'Sonar' = True},
+        ActiveIntel = {
+            Sonar = True,
```
speed2CZ: `true` not `True`

PR #2154, opened 2017-08-16
Remove air attack formation and decrease width by 20%

MrNukealizer:


aeoncleanse:
In time, it seems things aren't as bad as initially thought. Closing for now.

PR #2157, opened 2017-08-17
fix acu consumption not correct after overcharge

### LINKED ISSUE #717: ACU incorrectly displays mass+energy spend
Symptom: ACU economy always shows default value (or increased if it has ras upgrade)

This means it doesn't display as using mass/energy when you build something.

To reproduce:  Use acu overcharge.  This causes the bug.

CookieNoob:
the engine function GetRolloverInfo() provides the UI with values for
massRequested and massConsumed. Before overcharging, these values
usually coincide. However, after firing an overcharge, the function
returns for massRequested only 0 independently of the actual
consumption. This issue can be fixed by using massConsumed instead. (the
same applies to energy as well)

fixes #717

### DIFF
--- a/lua/ui/game/unitview.lua
+++ b/lua/ui/game/unitview.lua
@@ -86,8 +86,8 @@ end
 
 local statFuncs = {
     function(info)
-        if info.massProduced > 0 or info.massRequested > 0 then
-            return string.format('%+d', math.ceil(info.massProduced - info.massRequested)), UIUtil.UIFile('/game/unit_view_icons/mass.dds'), '00000000'
+        if info.massProduced > 0 or info.massConsumed > 0 then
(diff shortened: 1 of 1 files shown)


CookieNoob:
this doesnt fix the economy overlay, that is still incorrect. I dont know where that is handled.

MrNukealizer:
I'm fairly sure the data comes from the engine. I'll look into it, but in what situations do ACUs use mass while overcharging?

CookieNoob:
just check the acu consumption values of the acu after using an overcharge. The massRequested values are always returned as 0 from the GetRollOverInfo() function. I think the CreateEconomyEvent() function messes something up with the value since the effect of the overcharge is only temporary. But all of these functions are engine stuff...

MrNukealizer:
Why was this merged? The discussion said it doesn't actually work yet. I also have a strong feeling this will make the numbers inaccurate when stalling on mass or energy.

aeoncleanse:
It said it doesn't fix one thing, but fixes something. So I merged for testing.

PR #2158, opened 2017-08-17
Restrict console commands

CookieNoob:


MrNukealizer:
The warning should probably include what the command was to avoid confusion when it shows up in the log.

PR #2159, opened 2017-08-17
disable vision for submerged amphibious units

### LINKED ISSUE #2056: make submerge unit not able to see unit on land
this is more a balance issue.
as unit on land can't see submerge unit, would be great that submerge unit can't see unit on land.
would remove some unit shooting at other while the other one can't fight back.
that won't really change anything to ACU since theyhave omni anyway.
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=6317487

CookieNoob:
fixes #2056

the vision is still enabled for acus and sacus (and other units with
omni sensors)

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2840,6 +2840,14 @@ Unit = Class(moho.unit_methods) {
         if self.LayerChangeTrigger then
             self:LayerChangeTrigger(new, old)
         end
+        
+        if new == 'Seabed' then
+            if self:GetBlueprint().Intel.OmniRadius == 0 then
+                self:DisableIntel('Vision')
+            end
+        else
+            self:EnableIntel('Vision')
+        end
     end,
 
     OnMotionHorzEventChange = function(self, new, old)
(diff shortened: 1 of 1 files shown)


lua/sim/Unit.lua
```diff
@@ -2840,6 +2840,14 @@ Unit = Class(moho.unit_methods) {
         if self.LayerChangeTrigger then
             self:LayerChangeTrigger(new, old)
         end
+        
+        if new == 'Seabed' then
+            if self:GetBlueprint().Intel.OmniRadius == 0 then
```
aeoncleanse: What if OmniRadius is nil?

CookieNoob: I logged the radius for all units that can go underwater and have no omni radius defined in the blueprint and it always came back as 0. It is working properly for all amphibious land units... so I guess this is fine.

I can change it to `not ... > 0` though

CookieNoob: this is actually a bad idea and doesnt work for some reason...

PR #2160, opened 2017-08-18
add dummy weapon to sera sniper bot

### LINKED ISSUE #1319: Sera sniper incorrect attack move behaviour
Sera snipers in sniper mode do not stop walking when attack move is used, they just continue walking towards the enemy rather than stopping when they get into range of an enemy. Attack move works correctly when they are in their default mode.

CookieNoob:
fixes #1319

the main weapon (the first one in the weapons array) is disabled when it is used in sniper mode, thus it wont stop when it gets in range. This commit adds a new weapon that does no damage but stops the unit at the correct range due to being enabled all the time.


aeoncleanse:
More than this need to be done. To work, these weapons have their range shifted dynamically depending on active weapon. See the ACUs for details in their scripts.

CookieNoob:
I'll add the necessary code in the evening

PR #2161, opened 2017-08-18
increase target check frequency of sera destro torpedo

### LINKED ISSUE #1259: sera stop firing torpedo for sometime
After an attack order, if you give an other order, it will stop shootibg torpedo for 2 sec.
The solution is too reduce checkscanradius (something like that) from 2s to 0.2s.

I will post replay (log if neccessary) later

CookieNoob:
fixes #1259

@shalkya wtf is that issue description? it took me twice as long to figure out what you mean than actually fixing it...


shalkya:
i should have said "sera destro stop firing for periods of time"; i'm sorry, i thought the description was good enough to explain the title

PR #2163, opened 2017-08-18
add dummy weapon to cybran acu for torpedo targeting

### LINKED ISSUE #1105: targetting with torpedo com bug
http://forums.faforever.com/forums/viewtopic.php?f=3&t=9699

maybe link with this : https://github.com/FAForever/fa/issues/1090

CookieNoob:
fixes #1105

also sanitize the "CoolingUpgradeRemove" part of the script of the cybran acu

# THIS REQUIRES TESTING!!

I tested it and it worked nicely, but this is a really important unit...


MrNukealizer:
This seems to change the ACU behavior slightly when overcharging. 

1. It fixes the problem of the ACU refusing to fire at a wall or the ground when auto-overcharge is enabled.
2. When the ACU is commanded to manually overcharge a wall it always fires a shot from the main cannon along with the overcharge shot, whereas the other ACUs only do that some of the time. 
3. When executing multiple queued overcharge commands on the same target, the ACU doesn't try to turn its upper body to face forward between shots, potentially reducing delay between shots.

PR #2171, opened 2017-08-19
add failsafe to Buildstate of cybran drones

### LINKED ISSUE #2167: cybran build drones dont activate idle state when target is destroyed
warning: Error running lua script from destroyed thread: IssueMoveOffFactory: Passed in an invalid target point.
         stack traceback:
         	[C]: in function `IssueGuard'
         	...\gamedata\units.nx2\units\ura0001\ura0001_script.lua(111): in function <...\gamedata\units.nx2\units\ura0001\ura0001_script.lua:103>

CookieNoob:
add additional check to the cybran drones if the assisted unit is dead
or not

potentially fixes #2167

### DIFF
--- a/units/URA0001/URA0001_script.lua
+++ b/units/URA0001/URA0001_script.lua
@@ -103,7 +103,7 @@ URA0001 = Class(CAirUnit) {
         Main = function(self)
             local focus = self.spawnedBy:GetFocusUnit()
 
-            if not focus then
+            if not focus or focus.Dead then
                 ChangeState(self, self.IdleState)
             end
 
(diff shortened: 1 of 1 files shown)


units/URA0001/URA0001_script.lua
```diff
@@ -103,7 +103,7 @@ URA0001 = Class(CAirUnit) {
         Main = function(self)
             local focus = self.spawnedBy:GetFocusUnit()
 
-            if not focus then
+            if not focus or focus.Dead then
```
aeoncleanse: please add "or focus.BeenDestroyed()" to check the engine entity too.

MrNukealizer: When a unit is destroyed it should always set Dead. I've never seen a unit that was destroyed but didn't have Dead set to true.

aeoncleanse: It happens. Things can be wiped engine side _before_ the lua is told about it. Causes issues like this sometimes.

CookieNoob: it should probably be focus:BeenDestroyed() instead of focus.BeenDestroyed()

PR #2181, opened 2017-08-19
Feature/newvet lite

aeoncleanse:
Simplified version of vetmod owing to the fact that vet on experimentals was recently changed to be weaker when a bug was fixed.

This affects only the distribution of veterancy. That is, the HP and regen rewards stay the same, but we move from a very basic xp system (T1 give 1, T2 give 3, etc) to a mass-value-based system (Units gain xp based on the _value of mass_ they have contributed _damage_ to). It's powerful, flexible, and has been tested in EQ for a good 2 years now (Alongside other changes to the rewards, again, not copied here)


shalkya:
what about you kill a transport full of unit, will this count the unit inside, or just be as it is right now ?

aeoncleanse:
Nice point. Currently it would only count the transport.

CookieNoob:
good idea to start a huge fire somewhere else to distract from the salvation change :D

PR #2192, opened 2017-08-21
improve consumption display on units

### LINKED ISSUE #2197: cybran nuke subs shows wrong ui e consumtion when e stall
it drains -200, but when you reach e stall it show -20 e on the cybran nuke although it does drain -200 still
image

CookieNoob:
use energyRequested intead of energyConsumed if energyRequested is available (same for massRequested and massConsumed)

improves the fix for #717 #2157

fixes #2197

### DIFF
--- a/lua/ui/game/unitview.lua
+++ b/lua/ui/game/unitview.lua
@@ -87,11 +87,11 @@ end
 local statFuncs = {
     function(info)
         if info.massProduced > 0 or info.massConsumed > 0 then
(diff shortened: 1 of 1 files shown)


MrNukealizer:
I know I was the one who suggested the current version, but after looking more closely at the code behind it, `massUsed = math.max(info.massRequested, info.massConsumed)` seems better than checking if one is 0.

PR #2193, opened 2017-08-22
 	Make Aeon T2 generator ball spin like the T3 version

MrNukealizer:
The Aeon T2 generator has a little ball hovering above it much like the T3 generator, but that ball is static. Given the Aeon love of spinning balls, it seemed like it should move. The T1 version also has a static ball that can be made to spin, but it's a solid color so the spinning is practically unnoticeable.

### DIFF
--- a/units/UAB1201/UAB1201_script.lua
+++ b/units/UAB1201/UAB1201_script.lua
@@ -11,6 +11,13 @@ local AEnergyCreationUnit = import('/lua/aeonunits.lua').AEnergyCreationUnit
 
 UAB1201 = Class(AEnergyCreationUnit) {
     AmbientEffects = 'AT2PowerAmbient',
+
+    OnStopBeingBuilt = function(self, builder, layer)
+        AEnergyCreationUnit.OnStopBeingBuilt(self, builder, layer)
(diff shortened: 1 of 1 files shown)


Exotic-Retard:
um, uh, i think this is one of those things that should stay like it was probably.

the t1 doesnt spin and its a simple ball,
the t2 doesnt spin but its a more fancy ball,
the t3 is even more fancy and spins

im fairly sure this is how it was intended to be :\

aeoncleanse:
All the same, no reason to not do this, really. I'll see how good it looks, maybe do a quick poll.

MrNukealizer:
Literally every other hovering ball on any Aeon structure spins. It seems quite inconsistent that either of these don't, but it's really hard to see on the T2 version so I could see that being static for performance reasons.

MrNukealizer:
Literally every other hovering ball on any Aeon structure spins. It seems quite inconsistent that the T2 and T1 generators break that pattern. The T1 version is understandable because you'd have to look very closely to see if the ball is spinning or not, and if the model were a perfect sphere it would be impossible. The T2 version, on the other hand, is quite visible.

Issue #2087, opened 2017-06-26
New map filter hides maps it shouldn't


speed2CZ:
So I have 3 version of first Aeon mission in my map folder
first one you see is my dev version with `map_version = 3,` in the scenario file, others are v8 and 9 but the game isn't showing any
![forgedalliance_2017-06-26_12-48-32](https://user-images.githubusercontent.com/10947024/27536090-eadaf64c-5a6d-11e7-86e4-9f999f75067e.png)

And same applies for FA mission 1 and 5, also not showing

![explorer_2017-06-26_12-50-58](https://user-images.githubusercontent.com/10947024/27536159-3eaf2aae-5a6e-11e7-9979-bf6d298dc11c.png)

CookieNoob:
this is not a bug, these maps are blacklisted.
https://github.com/FAForever/fa/blob/develop/etc/faf/mapblacklist.lua#L900

If these are not supposed to be blacklisted, then I suggest to remove them from the blacklist.

PR #2201, opened 2017-08-22
fix blacklist for coop maps

### LINKED ISSUE #2194: Lobby map version check incorrectly says map is outdated
Something is wrong with the map version check when leaving the options screen. I just reproduced it with FA mission 5 in coop. I downloaded version 16 when creating the game, went to the options, clicked OK, and the game said `The currently selected map is outdated and/or unbalanced. Please download the latest version from the map vault.` I have versions 13 and 16 on my computer and in the options it clearly said version 16 was selected.

CookieNoob:
when the automated script generated the map blacklist it used the maps on the test server. However, the map version on the test server severely got out of sync from the map versions in the production enviroment. This lead to several coop missions shown as outdated even though the mission itself was the latest version.

fixes #2194 and #2087


ChrisKitching:
Bonus points for actually re-running the script, too. Plenty of non-coop maps are presumably obsolete by now...

CookieNoob:
not really needed, the outdated map filter is hiding older versions of maps when newer versions are present

CookieNoob:
@ChrisKitching where exactly is the script stored? (not that I had server access to actually run it)

ChrisKitching:
I think it was on the test server. Perhaps in my home directory there.

Eep. Why didn't this end up in git.

@sheeo - do you know?

... And do we actually care? At some point we might develop a proper map versioning scheme. The script was just an elaborate system for constructing canonical map versions from the many conflicting name-based versioning schemes. You end up with maps like `something-v2.v0003` and `something-v1.v0001`, and `something2.v0004`: the script constructed a strict ordering on these and then blacklisted everything except the newest one.

If you've implemented comparable logic in the lobby then maybe it's okay: but I'm not sure how you can have achieved this, since it requires knoweldge of the set of maps that exist on the server. At least when I implemented it, the idea was to rerun the script regularly and commit the results to the blacklist lua file. That way, you'd at least get outdated map notifications each time a new lobby version shipped. This was intended as an interim solution until map versioning stopped being insane.

In any case, it shouldn't be super hard to reconstruct it. It's just a bit of bash-fu ;) .

CookieNoob:
the ideal way to do it would be with the client automatically updating the maps if new versions are available and deleting old versions.

What I did (and you reviewed the PR actually) is to check if a map that is available has multiple versions and hides all except the latest one. https://github.com/FAForever/fa/pull/2065

ChrisKitching:
> What I did (and you reviewed the PR actually) is to check if a map that is available has multiple versions and hides all except the latest one. #2065

Yes - this is great as far as it goes, but what this can't do is detect that a map has a newer version on the server. It will, however, usefully detect when you have multiple versions installed locally.

As you say, the right approach involves something outside of lua-land sorting it out. Lots of sensible approaches exist. I would advise making the map versioning not be reliant on filename in the current way, too - have an actual version field in the database.

ChrisKitching:
Oh, this is a PR, too. Derp. Let's merge it :P

PR #2202, opened 2017-08-22
round clocktime for reload bar

### LINKED ISSUE #2198: missile loading at 99% on automatic launching unit
image
would it be better than missile unit finish their loading cycle. Right now it needs to activate the firing cycle to finish the loading cycle (which stay at 99%).

CookieNoob:
prevents early exits of the reload bar rendering which caused the bar to stop at 98%

fixes #2198


MrNukealizer:
There's no need to use math.floor every loop. I'd switch line 593 to `local clockTime = math.round(10*rof)` and not worry about precision after that since it's always an integer.

Issue #2216, opened 2017-08-27
Shield recharge timings make no sense at all


aeoncleanse:
```    -- Basically run a timer, but with visual bar movement
    ChargingUp = function(self, curProgress, time)
        local owner = self.Owner
        local position = owner:GetPosition()
        local shieldbp = self.Owner:GetBlueprint().Defense.Shield
        local shieldRadius = shieldbp.ShieldSize
        local aiBrain = owner:GetAIBrain()
        local otherShields = aiBrain:GetUnitsAroundPoint((categories.SHIELD * categories.DEFENSE), position, shieldRadius, 'Ally')
        local rechargeTime = time + ((table.getn(otherShields) - 1) * .2 * time)
        if rechargeTime > time * 3 then
            rechargeTime = time
        end
        while curProgress < rechargeTime do
            local fraction = self.Owner:GetResourceConsumed()
            curProgress = curProgress + (fraction / 10)
            curProgress = math.min(curProgress, rechargeTime)

            local workProgress = curProgress / rechargeTime

            self:UpdateShieldRatio(workProgress)
            WaitTicks(1)
        end
    end,```

aeoncleanse:
We increase charge time if there are nearby shields? Since when? We have Shield HP Spillover for that. 
If the time calculated from nearbys is bigger than 3*input, just use input? Wut?

MrNukealizer:
I have to agree. That's a really weird way to go about it. It has been in there since the file was [added to FAF](https://github.com/FAForever/fa/commit/f6feed3615bee3ab172f7700d179d073ca706ccd) but I'm not seeing the logic there. It does explain a lot of weirdness I've seen where one shield can stop artillery but two can't. For reference the original GPG code is: ```ChargingUp = function(self, curProgress, time)
        while curProgress < time do
            local fraction = self.Owner:GetResourceConsumed()
            curProgress = curProgress + ( fraction / 10 )
            curProgress = math.min( curProgress, time )
            
            local workProgress = curProgress / time
            
            self:UpdateShieldRatio( workProgress )
            WaitTicks(1)
        end    
    end,```

CookieNoob:
my guess is that this is the first attempt to nerf overlapping shields before damage transfer was introduced. After that was added, no one removed that code.

aeoncleanse:
I agree. I'm going to move for removing this to simplify shield mechanics for players, maybe publish a bullet point guide of how shields work afterwards.

JaggedAppliance:
[...]

PR #2217, opened 2017-08-28
Fix/shieldtimer

aeoncleanse:
Removes ZeP's original attempt at curtailing stacked shields by affecting their recharge cycles. This code is unknown, unpredictable, badly designed, and has been long superceded by shield damage spillover.

Also makes shields regen per 0.2s from per 1s, just for smoothness.


CookieNoob:
you should add `fixes #2216` in the commit message, not the title. It didnt close the issue automatically

PR #2225, opened 2017-08-29
Balance:T3gunshipchanges

PhilipJFryFAF:
adds stealth to the Wailer and removes its radar jamming ability
adds radar jamming to the Broadsword


CookieNoob:

PR #2226, opened 2017-08-29
Balance:NovaxBuff

PhilipJFryFAF:
increases the radar range from 70 to 150
adds omni range of 50
increases movment speed from 6 to 9


CookieNoob:

PR #2227, opened 2017-08-29
Massfabrebalance

PhilipJFryFAF:
rebalance of t2 and t3 mass fabricators
values are stolen from EQ for now
might have to make adjustments according to feedback in time

### DIFF
--- a/lua/sim/AdjacencyBuffs.lua
+++ b/lua/sim/AdjacencyBuffs.lua
@@ -39,7 +39,7 @@ local adj = {           -- SIZE4     SIZE8   SIZE12    SIZE16   SIZE20
         MassActive=         {-0.025, -0.0125, -0.008333, -0.0125, -0.0075},
     },
     T3MassFabricator={
-        MassActive=         {-0.2, -0.2, -0.2, -0.2, -0.0225}, -- is the SIZE16 value really correct?
+        MassActive=         {-0.2, -0.2, -0.2, -0.2, -0.0375},
     },
     T1EnergyStorage={
         EnergyProduction=   {0.125, 0.0625, 0.041667, 0.03125, 0.025},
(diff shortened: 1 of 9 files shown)


PhilipJFryFAF:
buffed the adjacency for t3 fabs+size20 buildings (QGates) to match up with t2 fabs

CookieNoob:

HUSSAR-mtrela:
Can we make the T3 mass fabs be a bit more efficient then T2 mass fabs
With this change 
```
15 x T2 fabs produce 15m for 1500e with 3000m build cost
01 x T3 fab produces 16m for 1500e with 4000m build cost
```
So T3 fabs have 1 extra mass bonus over T2 fabs but it cost 1000m more. can T3 fabs produce a little more mass, e.g. 18 or 20m?

duk3luk3:
They're already more space efficient, which is a big advantage in and of itself.

MrNukealizer:
Then why does nobody ever use them? I've seen a total of 0 T3 mass fabs created by players with over 600 rating.

PhilipJFryFAF:
The t3 fabs become more efficient due to the adjacency bonuses
eg with those numbers one t3 pgen provides enough power to run 2 t3 fabs that are adjacent to it which means that for 10 400 Mass invested you get 32 back so they pay back within 325 seconds
at the same time a t3 pgen can run "only" 28 t2 fabs with the power it provides  which means those pay back within 314 seconds
you can also use the t3 fabricator for its adjacency bonus factories for a bonus of 20% of whatever it produces atm which makes them pretty good (eg a UEF t3 land factory building percies drains 19.2 mass/second so the bonus there would be 3.84 mass (in a scenario where you have 3 landfacs adjacent to the t3 fab you'd get a cost reduction of 11.52 mass/second thus increasing its effective output by 72%)

PR #2240, opened 2017-09-02
fix:Tooltip

PhilipJFryFAF:
add stealth to the ability list of the HARMS since it does have it


Uveso:
Maybe it's also missing 'COUNTERINTELLIGENCE' as Category ?

PhilipJFryFAF:
ty for pointing that out
would have missed it otherwise

PR #2245, opened 2017-09-06
Improve army colors

speed2CZ:
This colous is better for recognizing strategic icons.
In additing it is the true UEF blue from the FA campaign

Screenshots here http://forums.faforever.com/viewtopic.php?f=42&t=15005#p153902


speed2CZ:
This is the sera one, I was always missing it <3
It should not conflict with the others
forgedalliance_2017-09-06_18-21-52

HUSSAR-mtrela:
can we change order of colors displayed in the comb-box? I mean that gray color and other dark colors  (e.g. dark green, dark purple) should be listed as last because they are harder to see in the game.

speed2CZ:
I never really had problems with any of those colours so Im not be the best one to sort them

aeoncleanse:
Replaced by #2272, a rebase.

PR #2248, opened 2017-09-07
Fix veterancy desync

MrNukealizer:


Crotalus:
If you index the instigator table with the unit ID instead of the actual unit object you get an ordered table/array and can skip the later sorting.

Someone should probably go through all lua code and make sure tables aren't indexed by objects, seems risky.

MrNukealizer:
I thought about that, but it seems like for performance purposes it's probably better to get and sort the IDs once than to have to get the attacker's ID every time a unit takes damage. I can definitely change it to that method if necessary though.

aeoncleanse:
Nah, this way we are creating a lot of tables in the long run. It's an effective doubling. Might I suggest storing entity ID on self in OnCreate and using it to index? Would be useful to have it around like that in many places.

MrNukealizer:
Done. Good point about saving the ID. That can reduce calls to GetEntityId by 13 just in the one file.

ChrisKitching:
[GetEntityId()](http://supcom.wikia.com/wiki/LUADOC_1.5.3596) is a native call, too, so this is less of a micro-optimisation than it appears. Purging it completely might be nice, but you might run into trouble if the native code is ever changing IDs - since calls to `GetEntityId()` would return the new one, but this won't.

aeoncleanse:
Native calls are still more than 3 times slower than fetching a self.thing. While IDs are recycled, they are never reassigned during the lifespan of a given C object.

PR #2266, opened 2017-09-26
Feature: Unit Skins Manager

HUSSAR-mtrela:
This PR adds a new dialog for selecting unit skins in the Lobby/Game options:

image

COMPLETED tasks by HUSSAR
- [x] add logic for managing unit skins (load, verify, save)
- [x] add dialog for selecting unit skins
- [x] add logic for verifying selection of 1 unit skin for each faction 
- [x] add logic for verifying selection of 1 unit skin for each unit type (ACU, SCU, experimental, etc) 
- [x] add filter of unit skins base on each faction
- [x] add temporary unit skins for testing Unit Skins Manager

TODO tasks for @JJ173 
- [ ] remove temporary unit skins when we have skins for all ACUs
- [ ] add a few unit skins for each ACU and also SCU (if we have time) base on below specification
- [ ] add logic to apply selected skins to unit(s) when the game is loading
- [ ] add a function for checking with FAF server which skins a player can select

In my onion, there should be two types of unit skins:
1. unit skins for all players can select 
2. unit skins that can be won or rewarded for special contributions to FAF community

Each unit skin should contain these files:
```
 /textures/ui/unitSkins/SKIN_NAME/UEL0001_Albedo_2.dds (name might differ per faction/unit)
 /textures/ui/unitSkins/SKIN_NAME/UEL0001_SpecTeam_2.dds (name might differ per faction/unit)
 /textures/ui/unitSkins/SKIN_NAME/skin_icon.dds
 /textures/ui/unitSkins/SKIN_NAME/skin_info.lua
```

Example of skin_info.lua file:

```
uid = "4dfc9962-a001-4a86-bb4f-3d8b4ecc04b0"
name = "AEON Rusty Skin"
description = "A Rusty skin variant for the AEON ACU."
author = "Exotic_Retard"
faction = "AEON" -- or UEF, CYBRAN, SERAPHIM, NOMADS
version = 1.0 
target = "ACU" -- or SCU, Experimental_Name
-- note there is no setting for icon because SkinManager will use skin_icon.dds
```

Preview of Unit Skin Manager
image


lua/ui/lobby/SkinsManager.lua
```diff
@@ -0,0 +1,235 @@
...
+    name = name:gsub(" %d", "")
+    -- cleanup name
+    name = name:gsub(" %(%)", "")
+    name = name:gsub("%)", "")
+    name = name:gsub(" %-", " ")
+    name = name:gsub("%- ", "")
+    name = name:gsub("%-", " ", 1)
+    name = name:gsub("%_", " ")
+    name = name:gsub(" %(", " - ")
+    name = StringCapitalize(name)
```
ChrisKitching: You probably want to read the [patterns tutorial](https://www.lua.org/pil/20.2.html). You should be able to merge many of these gsubs together - and provide a nice comment explaining what it's for? ;)

Exotic-Retard:
`target = "ACU" -- or SCU, Experimental_Name`
could you explain how this works? i think a clearer way of doing this is to put the blueprint ID in here instead, such as UEL0001 for the uef acu. That way, its very specific, doesnt need translation/interpretation to make sure we only apply that skin to the right factions acu, ect. Also it may let us make more general skins in the future for any unit. in theory that is.

duk3luk3:
hey @ChrisKitching do you have any idea how to manage the "add a function for checking with FAF server which skins a player can select" part? There's the "[CreateLobby hack](https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/lobby.lua#L628)" but I haven't actually managed to get that to work - FA complains if it receives more than one CreateLobby GPG message and I have no idea how to massage it so it doesn't.

What I want in the most basic form is a channel that allows me to send arbitrary data from client to ingame lobby (or, ideally, also from client to game) so that we can use this information to check whether other players are lying to us (about things like their name, their rank, their skins).

HUSSAR-mtrela:
@Exotic-Retard you can set the `target `value to any string. Using blueprint ID sounds like the best choices because this guarantees unique value and it will be useful to determine which skin should be applied to a unit. The Skin Manager uses the `target` as string value to determine if a player can select multiple skins for the same faction. For example, it allows selecting one `AEON` skin for `ACU` and one `AEON` skin for `SCU` (as it is displayed in preview above). However, the Skin Manager does not allow selecting 2 AEON skins for an `ACU` unit because then the game would not know which skin should be applied.

lua/ui/lobby/SkinsManager.lua
HUSSAR-mtrela: yes, I should really learn how to use these patterns. Thanks for catching it :)

ChrisKitching:
> hey @ChrisKitching do you have any idea how to manage the "add a function for checking with FAF server which skins a player can select" part? There's the "CreateLobby hack" but I haven't actually managed to get that to work - FA complains if it receives more than one CreateLobby GPG message and I have no idea how to massage it so it doesn't.
> 
> What I want in the most basic form is a channel that allows me to send arbitrary data from client to ingame lobby (or, ideally, also from client to game) so that we can use this information to check whether other players are lying to us (about things like their name, their rank, their skins).

Prettymuch your only choice (that I know of) here is something like the `CreateLobby` hack.

The only way you can get data from the outside world into Lua is via GPGNet calls or command line arguments. The GPGNet entry points are hardcoded with specific signatures, and you can't add more without some relatively challenging binary patching (I tried).

It sounds like the `CreateLobby` implementation is flawed, but I think the underlying idea should work. You need to find a different GPGNet entry point - one that you _can_ call repeatedly - and then rewrite its Lua handler to have two modes: regular, and message-handler. You'll probably use one of the arguments as some kind of sentinel value to detect "message handler" mode.

Going the other way is much easier: the GPGNetSend function we already use to talk to the local relay (and, via it, the server) lets you send arbitrary data out.

In the specific case of skins: since that's static for the lifetime of `fa.exe`, you could do it with command-line arguments. But it'd be pretty easy to cheat, just like it is for all the other things we send via command line arguments.

As a last resort, you could implement a kind of sanity-check: you can have the client GPGNetSend all its data (rating, skins, whatever else) to the server, and have the server punish you in some way - perhaps via disconnect - if you have manipulated the command line arguments. You'd also want to implement a timeout so that if this doesn't occur within a certain time limit after you join a game, you get booted anyway (otherwise someone can just patch their lua not to send the message at all).

JJ173:
This PR still needs rebasing @HUSSAR-mtrela I can do the logic for this next week though. I just want to get all of these commits removed that are irrelevant to our task. I'm eager to finally get this working after discussions about it last year. The idea is quite simple. Exotic and I will take a look some time. If he's not around, I'll give it a go myself. We have some what of an idea of where to go with this but we need to make sure it's a route that is secure and doesn't cause desync.

JJ173:
Can I also request that you ensure the UI can only choose 1 skin per faction? If a user picks one, it needs to deactivate the other selections. Thanks :)

Exotic-Retard:
technically no need for that, you can only set one mesh at a time anyway. so if a user picks multiple skins per faction it wouldnt matter. we rely on this to be able to get the acus looking different in the same game already so kinda ho harm done i guess. if its put in just make it not per faction but per unit type so that we could extend the system in the future to non-acus as well in theory

JJ173:
@HUSSAR-mtrela Still waiting for this to be rebased.

PR #2285, opened 2017-10-18
Remove TAirUnit from the UEF Gunship class

speed2CZ:
TAirUnit is inside AirTransport class so it's not needed here.
I don't see any functionality removed by this.

Ive encountered ambiguous class definition error with RK explosion mod, it's hooking AirUnit class inside defaultunits.lua . The mod worked fine for all units except the UEF gunhips. Removing this ^^ seems to fix the problem and I dont think it's removing anyting from the gunship.
More tests are welcome.


Uveso:
I tested this with Speed2CZ.

TAirUnit is indeed unnecessary, it's already defined inside AirTransport class:
```LUA
--- Base class for air transports.
AirTransport = Class(AirUnit, BaseTransport) {

```

PR #2306, opened 2017-11-09
Blacklist Icon Support

PhilipJFryFAF:
since we have integrated icon support and this mod just breaks the game for the user it should be blacklisted


Uveso:
Well there is still a code part that is not covered by our own icon support. (ACU enhancements)
I will make a PR for it, then we can block Black Ops icon support :)

PhilipJFryFAF:
ty for noticing (and fixing)
i was under the impression that everything the mod does is already covered by faf

Uveso:
FAF can more then the old BO Icon support.

-Buildtemplates
-Unitgroups ([CTRL]+[1] etc.)
-ACU Enhancements (buttons are implemented way after BO icon support)

PR #2308, opened 2017-11-10
Rebalance Civilian Structures

speed2CZ:
WIP
- [x] Fix factions in BPs
- [x] Adjust HPs
- [x] Regen for Cybran structures
- [x] Adjust M cost


JaggedAppliance:
Looks good, tested on theta, saltrock, and ambush pass where there were only slight imbalances which can be rectified. Nothing gamebreaking for these maps though with the mass amounts.

Issue #2325, opened 2017-12-02
Bug: Cheat AI does not build intel units if omni cheat is disabled.


Uveso:
A cheating AIx normaly don't need intel units or structures because it has Omni-View.

But if the option "AIx Omni Setting" is disabled, the AIx has no Omni-View AND is not building intel Units.

This way we have a chating but weak AIx.

MrRowey:
@Uveso Do you know if this Bug is fixed or not ?

MrRowey:
@Azraeel you know if this is still a bug ?

Garanas:
This is likely still a bug. I'll look into it for the patch of January.

Uveso:
I fixed this on  11 Jun 2018
https://github.com/FAForever/fa/commit/b9a348b881c74e86f74bbab51c7d9e0a6c94898a

AIs are now always building NonCheatBuilders

PR #2326, opened 2017-12-02
Fix Cheat AI does not build intel units if omni cheat is disabled.

Uveso:
Issue  #2325

To fix this i added a if-then statement to check the options for "AIx Omni Setting".

Now "NonCheatBuilders" (Intel Units) will be build in case we have a normal AI, or an AIx with no omni View.


Uveso:
This fix is already outdated :)

If you don't play in Assassination mode, and your commander is gone, then also Omni View is off.
And you need again intel builders. So you need intel builders in every game except Assassination.

At the end this will be fixes by a new BuilderCondition: "HasNotOmniView".
For now, we build intel in every game, even if we are cheating with Omni View.

But this will be part of another PR :)

PR #2330, opened 2017-12-04
Improve map selection window

CookieNoob:
this PR contains two changes to the map selection window:

1. the way the AI compatibility is tested for maps
instead of searching for a land path node with a specific name, search for a marker that has other markers listed as "adjacent" markers. These markers are usually path nodes (not necessarily land pathnodes). Markers with only one entry for this property correspond to chains, so searching for the separator (space) in this property yields markers with multiple adjacent markers.

2. fixes official maps being hidden by the outdated map filter

### DIFF
--- a/lua/ui/maputil.lua
+++ b/lua/ui/maputil.lua
@@ -237,17 +237,21 @@ function CheckMapHasMarkers(scenario)
     doscript('/lua/dataInit.lua', saveData)
     doscript(scenario.save, saveData)
 
-    if saveData and saveData.Scenario and saveData.Scenario.MasterChain and
-            saveData.Scenario.MasterChain['_MASTERCHAIN_'] and saveData.Scenario.MasterChain['_MASTERCHAIN_'].Markers then
-        for marker, data in saveData.Scenario.MasterChain['_MASTERCHAIN_'].Markers do
-            if data.adjacentTo then
-                if string.find(data.adjacentTo, ' ') then
-                    return true
-                end
-            end
-        end
-    else
-        WARN('Map '..scenario.name..' has no marker chain')
+    
+    local markers = saveData and 
+                    saveData.Scenario and 
+                    saveData.Scenario.MasterChain and
+                    saveData.Scenario.MasterChain['_MASTERCHAIN_'] and
+                    saveData.Scenario.MasterChain['_MASTERCHAIN_'].Markers or false
+ 
+    if not markers then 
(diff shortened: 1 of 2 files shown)


lua/ui/maputil.lua
```diff
@@ -240,8 +240,10 @@ function CheckMapHasMarkers(scenario)
     if saveData and saveData.Scenario and saveData.Scenario.MasterChain and
             saveData.Scenario.MasterChain['_MASTERCHAIN_'] and saveData.Scenario.MasterChain['_MASTERCHAIN_'].Markers then
         for marker, data in saveData.Scenario.MasterChain['_MASTERCHAIN_'].Markers do
-            if string.find(string.lower(marker), 'landpn') then
-                return true
+            if data.adjacentTo then
+                if string.find(data.adjacentTo, ' ') then
+                    return true
+                end
```
HUSSAR-mtrela: you should combine both if statments:
```
if data.adjacentTo and string.find(data.adjacentTo, ' ') then
   return true
end
```

HUSSAR-mtrela: Also, saving master chains to a local variable will make this function run faster because it will not have to search for it multiple times:
local markers = SaveData and SaveData.Scenario and 
    SaveData.Scenario.MasterChain and
    SaveData.Scenario.MasterChain['_MASTERCHAIN_'] and
    SaveData.Scenario.MasterChain['_MASTERCHAIN_'].Markers or false
 
if not markers then 
   WARN('Map '.. scenario.name..' has no markers') return false 
else 
   for marker, data in markers do
      if data.adjacentTo and string.find(data.adjacentTo, ' ') then
         return true
      end
   end
end
return false

speed2CZ: ```local markers = SaveData.Scenario.MasterChain['_MASTERCHAIN_'].Markers or false```
well this should do the same job

speed2CZ: or maybe you don't even need `or false` since if it happens to be `nil` it will pop out the warning as well

HUSSAR-mtrela: ```local markers = SaveData.Scenario.MasterChain['_MASTERCHAIN_'].Markers or false```
would crash the function if either the following values are nil:
```SaveData, 
SaveData.Scenario, 
SaveData.Scenario.MasterChain, 
SaveData.Scenario.MasterChain['_MASTERCHAIN_']```

PR #2331, opened 2017-12-04
remove "8 - Badlands_v4.v0001" from the mapblacklist

shalkya:
remove "8 - Badlands_v4.v0001" from the mapblacklist as it is the most up-to-date version.

Btw i have not done any rebase or shit since ages. Hope won't be an issue.


CookieNoob:
the entry in the blacklist is definetely a mistake... 

however, I kinda dislike the naming of the map. Starting with a number is a problem for the map files and the versioning of the map is terrible too. A simple solution would be to upload a new version of the map with the name "Badlands" and give it a proper version (version 5). Then hide the other version in the vault.

PR #2334, opened 2017-12-14
Aeon Hydro targetting fix

### LINKED ISSUE #2333: Aeon hydro not being targeted while in build
[forum thread](http://forums.faforever.com/viewtopic.php?f=3&t=15588)

PhilipJFryFAF:
for reasons unknown to me the AI table in the unit bp messes up the targeting for units so i just removed it
(couldn't find any negative impact when doing so but feel free to double check)

fixes #2333

### DIFF
--- a/units/UAB1102/UAB1102_unit.bp
+++ b/units/UAB1102/UAB1102_unit.bp
@@ -1,11 +1,4 @@
 UnitBlueprint {
-    AI = {
-        TargetBones = {
-            'Blade03',
-            'Extension01',
-            'Extension04',
-        },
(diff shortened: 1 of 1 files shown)


CookieNoob:
> for reasons unknown to me the AI table in the unit bp messes up the targeting for units so i just removed it
the reason is that these bones are below the surface during the build of the hydro (aeon build animation) and thus the units hit the ground instead of the part which is already above the surface. Removing this part will make all units only target the central bone (I think) of the hydro instead of several targets inside the model.

PR #2338, opened 2017-12-21
temporarily add christmas objects to the game

CookieNoob:
includes:
some props being replaced by gifts (9% of the rocks and 3.5% of the trees)
Santa hat on GC
GC rename
Snowman


CookieNoob:
I suggest to use "squash and merge" so that all this is only a single commit

PR #2343, opened 2017-12-26
diversify arty

PhilipJFryFAF:
more range for sera/aeon
less range for cybran/uef
higher rate of fire/less dmg for cybran


aeoncleanse:
Didn't we literally just UNdiversify arty range, cos differences are stupid?

PhilipJFryFAF:
sry didn't check on the PR lately
we decided to keep some differences in range but not enough to make one arty incredibly more useful than another

Issue #2312, opened 2017-11-12
fuel bar hide shield bar on T3 transport


shalkya:
this is the current situation : 
![image](https://user-images.githubusercontent.com/11618427/32704386-6f81c5ca-c804-11e7-9889-2153391fd087.png)
![image](https://user-images.githubusercontent.com/11618427/32704539-22de60b8-c807-11e7-82eb-50830c62976d.png)

and the fix would end up giving something similar to this : 
![image](https://user-images.githubusercontent.com/11618427/32704546-5f653dae-c807-11e7-8ed0-9887ca9d5d59.png)

PS : you need to look at bottom left

PhilipJFryFAF:
fixed with #2345

PR #2345, opened 2017-12-31
Fix issue #2312

FAETHER:
Create proper fuelbar display for shielded air units. https://github.com/FAForever/fa/issues/2312
Also will work for any mods that add shield air units.

The end result:

### DIFF
--- a/lua/ui/game/layouts/unitview_mini.lua
+++ b/lua/ui/game/layouts/unitview_mini.lua
@@ -158,11 +158,11 @@ function PositionWindow()
 end
 
 function UpdateStatusBars(controls)
-	if options.gui_detailed_unitview != 0 and controls.store == 1 then
-		LayoutHelpers.CenteredBelow(controls.fuelBar, controls.shieldBar,3)
- 	    LayoutHelpers.CenteredBelow(controls.shieldText, controls.fuelBar,-2.5) 
- 	elseif options.gui_detailed_unitview != 0 then
- 	    LayoutHelpers.CenteredBelow(controls.fuelBar, controls.shieldBar,0)
- 	    LayoutHelpers.CenteredBelow(controls.shieldText, controls.shieldBar,0)
- 	end
+    if options.gui_detailed_unitview != 0 and controls.store == 1 then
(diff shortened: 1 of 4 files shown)


speed2CZ:
Im not sure if this is the right place to set it. The position should be set in the layout file.

FAETHER:
I will look into it.

lua/ui/game/layouts/unitview_mini.lua
```diff
@@ -156,3 +156,13 @@ function PositionWindow()
     end
     LayoutHelpers.AtLeftIn(controls.bg, controls.parent, 17)
 end
+
+function UpdateStatusBars(controls)
+	if options.gui_detailed_unitview != 0 and controls.store == 1 then
+		LayoutHelpers.CenteredBelow(controls.fuelBar, controls.shieldBar,3)
```
speed2CZ: the position should be set on line 72
does it work if you move it there

just setting the positions without adding any new funcitons or variables

FAETHER: No, unfortunately it will not work on line 72 because it is encapsulated inside Setlayout () which is called only 3 times on the game start from unitview_detail file. Because the part that that I am making is a dynamic; the game needs to check every unit that is being displayed in real time (if you worry about performance, it is light because it updates only when the mouse hovers over a unit) . The reason behind this is simple;  faf unitview was always static, so there is no code where I can place it except to make a new function.

HUSSAR-mtrela:
I recommend moving the fuel bar below the text with remaining fuel time and making the bar shorter. (something like the veterancy bar) Stacking many small horizontal bars next to each other, makes it harder to read them. IMO, shield bar should have same height as the HP bar.

Exotic-Retard:
hmm i think the layout is fine actually, its in the same position as you would see in the other spots i.e. just below the health bar and doesnt take up a lot of ui space, so its not cluttering things up much

FAETHER:
I am just going to leave it at that for now unil you guys decide that to do with it. Different layouts work fine.

PR #2347, opened 2018-01-02
Walls double click fix

FAETHER:
Issue: FAForever#2292

End result:
All walls in the camera view will be selected by double click. I have tested this and can confirm that there are no drawbacks in removing the 'Wall' flag.

P.S I think that GPG have added this flag intentionaly, but if we want to fix this, this is the only solution possible.


speed2CZ:
You can't do this, there are many parts of the code that use this to recognize walls.

Issue #2257, opened 2017-09-22
give rally point to novax


shalkya:
there is no rally point for novax, even if they can re-build sat'; feels a bit weird.

PR #2352, opened 2018-01-06
Give Rallypoint to Novax

FAETHER:
Issue: https://github.com/FAForever/fa/issues/2257
The rallypoint will only work for rebuilt satellite. Still better then nothing. 
This line has been killing me. That is why the first satellite does not follow the rallypoint. 
https://github.com/FAForever/fa/blob/develop/units/XEB2402/XEB2402_Script.lua#L58
I am trying to rework the station script so that once it is built, the station will insta build a satelite (so that way it will follow rallypoint) 
Some problems it run into: 
1. how to make factory autobuild (best example kennels code: https://github.com/FAForever/fa/blob/develop/lua/terranunits.lua#L405 but it is unreadable to me. )
2. The fist satellite should have separate bp (to have economy categories = 0 ). Because once loaded bp's cant be changed.  

If i figure this out i will do another PR, but for now only this.


FAETHER:
I would greatly appreciate any help in that regard.

CookieNoob:
why dont you just issue a move order to the satellite?

FAETHER:
I tried doing issue move but i made it like this: 
local place = {200, 50, 0}
IssueMove(self.Satellite, place)

Nothing happened, so i guess i used engine function incorrectly. 
We need better engine documentation XD.

Imma test this quickly, Cookienoob.

FAETHER:
Where self.Rallypoint suppose to come from? 
log gives: 
`WARNING: Error running lua script: IssueMove: Passed in an invalid target point.`

speed2CZ:
If you use the function incorrectly, it'll break.
About the documentation, if you feel it's not enough, you can help improving it https://github.com/FAForever/fa/blob/develop/engine/Sim.lua#L631-L636

FAETHER:
Additional thanks to Speed2 for helping me with this!

PR #2354, opened 2018-01-07
Update french translation

PhilipJFryFAF:
Translation provided by discord user `Rackover#2107` 
I don't know any french so if someone could read over it and correct stuff that would be nice


PhilipJFryFAF:
... gotta fix my files first i guess

PR #2355, opened 2018-01-07
more french

PhilipJFryFAF:
translation provided by discord user `Rackover#2107`
please check for errors if you can (i can't read any french)


PhilipJFryFAF:
some lines are still missing - will update when i get them

loc/FR/strings_db.lua
```diff
@@ -7769,229 +7769,232 @@ MAPSEL_0044="13"
...
+loadingtip_0000="Astuce"
+loadingtip_0001="La surcharge de l'UBC nécéssite 5 000 d'énergie. Il faut au moins 1 stockage d'énergie pour l'utiliser."
+loadingtip_0002="Construisez chaque stockage d'énergie de façon à ce que sa destruction n'endommage aucun autre bâtiment."
+loadingtip_0003="La surchage de l'UBC inflige 400 points de dégâts à un autre UBC, 800 aux bâtiments et 12 000 aux autres unités."
+loadingtip_0004="Ne construisez pas de fabricateur de masse avant d'avoir complètement amélioré tous vos extracteurs."
+loadingtip_0005="Entourez vos extracteurs de stockages de masse afin d'améliorer leur production de masse."
+loadingtip_0006="Améliorez vos extracteurs en Tech 2 avant de construire des stockages de masse."
+loadingtip_0007="Un UBC immergé est toujours plus difficile à repérer ou à assassiner. Si vous ne pouvez pas protéger l'UBC en fin de partie, cachez-le."
+loadingtip_0008="Un missile stratégique (\"nuke\") se construit en 5 minutes : l'anti-missile, lui, se construit en 4 minutes."
+loadingtip_0009="Utilisez l'ordre \"Se déplacer en attaquant\" depuis une usine augmente la portée de recyclage des ingénieurs produits, le temps de l'execution de l'ordre." ###
```
speed2CZ: remove `###` at the end of this line

loc/FR/strings_db.lua
```diff
@@ -7769,229 +7769,232 @@ MAPSEL_0044="13"
...
+sorian_0027="IAx: IA Sorian Aérien"
+sorian_0028="IAx: IA Sorian Naval"
+sorian_0029="IAx: IA Sorian Tortue"
+sorian_0030="IAx: IA Sorian Adaptative"
+teamkill_0001 = "Fratricide détecté"
+teamkill_0002 = "Vous avez été tué par un tir allié. L'assassinat délibéré de coéquipiers est proscrit par le règlement de FA Forever. Si vous considérez que votre défaite était délibérée ou déloyale, appuyez sur le bouton ci-dessous pour le signaler."
+teamkill_0003 = "Signaler"
+teamkill_0004 = "Pardonner"
+
 # part added by retard for additional completion and stuff
```
speed2CZ: you could also change all commends from `#` to `--`

Rackover:
Here is the definitive version ! 

[strings_db.zip](https://github.com/FAForever/fa/files/1611292/strings_db.zip)

PhilipJFryFAF:
took rackovers new version and updated the PR

PhilipJFryFAF:
updated the PR with more suggested changes

shalkya:
i've done another checking, will need one more checking/testing by changing loc.nx2 file (and putting them as read only) since i can't get the faf setting by using the console + starting the game localy.
will wait until new change are being done (some change are still left to be done)
so if another french wanna test, it's better to do as i just explained (changing loc.nx2 way)

Rackover:
Rechecked along with @shalkya , new version here, awaiting another check.
[strings_db.zip](https://github.com/FAForever/fa/files/1678029/strings_db.zip)

Also know that a lot of faf-content cannot be translated as it's not present in the loc files. This represents a handful of lines, maybe 10 or 15. Not much compared to the 700+ translated lines here, so it shouldn't be any problem.

I contacted keyser for re-check, but anyone else can do it too. But please **please** don't send me lines to translate that are not present in the locfile. Check for line presence in the US / FR loc file before sending it to me : if the line is not there, i cannot translate it.

loc/FR/strings_db.lua
```diff
@@ -1,6 +1,10 @@
 # autogenerated file - DO NOT EDIT THIS FILE UNDER ANY CIRCUMSTANCES
 # USE LocStrings.xls INSTEAD TO EDIT STRINGS
 
+--[[
+	r: FAF Translation starts on line 7656
+--]]
```
ChrisKitching: This comment doesn't seem pointful?

Git blame can tell you everything you need to know, and that line number seems likely to end up being made incorrect as time goes on...

Rackover: Remove it then. I just put it here to make reviewing the translation easier, as the translation I made is mixed up with the official FA translation, itself mixed up with the original SC1 translation. This notice allows to distinguish the fan-made translation from the rest : i even added a banner around that line to help acknowledging the separation.

ChrisKitching: > Dont hesitate to contact me on slack if you have any other question or remark regarding this file, by the way.

Just a drive-by review, I'm afraid. I know exactly enough French to ask people if they speak English ;)

Issue #2351, opened 2018-01-06
Loyalist Missile Deflection


OCGDFS:
Im not sure if this feature is intended like it is at the moment. However, if the deflected missile should target the unit which fired the missile, it doesnt work. Though the missile flies into the opposite direction, it ends up on an unspecified point.

JaggedAppliance:
Can you post a replay of this? As far as I know the missile travels directly towards where the missile was shot from with no arc, so if it was a cruiser missile for example it will often hit the ground on the way back and if not the cruiser will likely have moved since shooting anyway.

OCGDFS:
I tested it with com tml (uef). It doesnt work here. I cant say somethin about other missiles.

OCGDFS:
I've tested it with a tml com. The missile flies back somewhere behind the com, so there was no targeting. I'm not sure behind the intention behind that feature though, but i think i remember that it worked in the past, vanilla maybe.

FAETHER:
Proof please.

FAETHER:
![123](https://user-images.githubusercontent.com/33268383/34652926-dc395db6-f3b2-11e7-9b5a-2dedecec47a8.png)

Yes, every tactical missle after being deflecled looses its collision with any units; in pic it goes through the shields. It has only collision  with the mapheight.

speed2CZ:
This should be fixed already

PR #2356, opened 2018-01-08
Loyalist deflection bug fix.

FAETHER:
Issue: https://github.com/FAForever/fa/issues/2351
Description:  Because of this commit, https://github.com/FAForever/fa/commit/d59cbecb4acdb22d310e83d42fb6f4318026e314#diff-460d6c7c63eb0ddeafeda7bc31da0972
the DamageData variable is no longer used in the GetCollideFriendly function. Whoever made this commit did not consider all the consequences of this change. This led to bug mentioned above.
The only real change i made is on the line 128.
End result:
Bug fixed, everybody happy.

Also, please test this! Dont just throw the work it into garbage because of looks!


CookieNoob:
when you upload an entire file with changes, it is really hard to see what you changed. So first upload the unchanged file and then in a separate commit do your changes.

there are also LOGs left in the file, you probably forgot to remove them.
it also looks like you copied over the hook into the newly uploaded file? Plz separate these things into different commits!

lua/defaultantiprojectile.lua
```diff
@@ -0,0 +1,156 @@
...
+           return
+        end
+            
+    local beams = {}
+    for k, v in self.RedirectBeams do               
+        table.insert(beams, AttachBeamEntityToEntity(self.EnemyProj, -1, self.Owner, self.AttachBone, self:GetArmy(), v))
+    end
+        if self.Enemy then
+         -- Set collision to friends active so that when the missile reaches its source it can deal damage. 
+            self.EnemyProj.CollideFriendly = true         --Change by Kyryl Melekhin 1/8/2018
```
CookieNoob: comments like this are not necessary, you can use git blame for that

lua/defaultantiprojectile.lua
```diff
@@ -0,0 +1,156 @@
...
+WaitingState = State{
+    OnCollisionCheck = function(self, other)
+        --LOG('*DEBUG MISSILE REDIRECT COLLISION CHECK')
+        if EntityCategoryContains(categories.MISSILE, other) and not EntityCategoryContains(categories.STRATEGIC, other) 
+           and other != self.EnemyProj and IsEnemy( self:GetArmy(), other:GetArmy() ) then
+           self.Enemy = other:GetLauncher()
+           self.EnemyProj = other
+           --NOTE: Fix me We need to test enemy validity if there is no enemy 
+           --      set target to 180 of the unit
+           if self.Enemy then
```
CookieNoob: maybe try to change the army of the missile here? that will make the tmd of the original owner of the missile attack the missile

FAETHER:
Some of those comments and logs were left by gpg developers. Should i remove them all? Or leave some that explain how the code works?

CookieNoob:
The comment I mean was this:
`--Change by Kyryl Melekhin 1/8/2018`

the LOG I was talking about is this one:
`       LOG('*DEBUG MISSILEREDIRECT START BEING CREATED fa')`

the best way would be to redo the PR by first adding the defaultantiprojectile.lua file (first commit)
then remove the hook and add that code to the defaultantiprojectile.lua file (second commit)
then do the bugfix and add your changes as the last commit (dont leave debug logging or comments like "added/changed by... on..." in the code)

FAETHER:
Alright, i will redo it when i get home, in ~5hours.

lua/defaultantiprojectile.lua
Exotic-Retard: how do you do that? im not sure its possible :/

CookieNoob: if there are no engine functions for that, maybe just create a new projectile (with the new army ofc), copy the damagedata and use the original blueprint

FAETHER: That would be a good thing to make as enchancement issue on github. At some point i will took into this.

PR #2358, opened 2018-01-11
Loyalist deflection bug fix

FAETHER:
Issue: #2351
Description: Because of this commit, d59cbec#diff-460d6c7c63eb0ddeafeda7bc31da0972
the DamageData variable is no longer used in the GetCollideFriendly function. Whoever made this commit did not consider all the consequences of this change. This led to bug mentioned above.
The only real change i made is on the line 128.
End result:
Bug fixed, everybody happy.

I have tried my best.


PhilipJFryFAF:
the aeon tac missile launcher missle and mobile missile launcher missile don't get directed back to the sender

aeon mml missile just goes straight for the ground and the aeon tml missile misses the origin by something like the min range of the tml

aeoncleanse:
Reworked by @Crotalus and merged as 733bd28f48e070a9880ea6ecc9ff8cd8b919b5ee

Issue #2253, opened 2017-09-19
show shield with the ACU UI icon


shalkya:
![image](https://user-images.githubusercontent.com/11618427/30605189-7644eaf2-9d6d-11e7-9fb2-87de50f29fb7.png)
would look like something like that, and ofc it would update like the life does update.
before : ![image](https://user-images.githubusercontent.com/11618427/30605329-f725e93c-9d6d-11e7-8439-c646d8da5dee.png)

after : ![image](https://user-images.githubusercontent.com/11618427/30605245-a7c30262-9d6d-11e7-9771-51fe0982d7b3.png)

PR #2360, opened 2018-01-14
Acu icon shieldbar

FAETHER:
Issue: https://github.com/FAForever/fa/issues/2253
Description:

The first commit fixes invalid bitmap dimensions involving different color bars.
Before:
default
After:
change

The second commit adds shieldbar file.

The third commit expands the UEF avatar tray, to accomodate shieldbar. Note: it will always look like that.
capture

The forth commit creates shieldbar mechanics. 
shield
There are only 2 commanders that can have shields. Aeon is always the broken one, but it would require too much work to remake the entire bitmap of self. It is very convoluted code and would suck hours of testing, so I left it at default.


JaggedAppliance:
I like this a lot, could you make this pr for development/balance? It will get in the game a lot faster :)

Issue #2322, opened 2017-11-26
fighter bomb slowing down in move order when they are nearby opponent land/naval unit


shalkya:
they are speeding the shit out of them when they got the attack order. 
btw testing on T1 and T4 bomber + mercy and cybran gunship, this behaviour doesn't happen.
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=7106218

shalkya:
In order to keep some archive about the investigation : https://www.youtube.com/watch?v=CSB_bv6H-QU

`DRA0202 = Class(CAirUnit) {

    Weapons = {

        AntiAirMissiles = Class(CAAMissileNaniteWeapon) {},

        GroundMissile = Class(CIFMissileCorsairWeapon) {

        

            IdleState = State (CIFMissileCorsairWeapon.IdleState) {

                Main = function(self)

                    CIFMissileCorsairWeapon.IdleState.Main(self)

                end,

                

                OnGotTarget = function(self)

                    self.unit:SetBreakOffTriggerMult(2.0)

                    self.unit:SetBreakOffDistanceMult(8.0)

                    self.unit:SetSpeedMult(0.67)

                    CIFMissileCorsairWeapon.IdleState.OnGotTarget(self)

                end,            

            },

        

            OnGotTarget = function(self)

                self.unit:SetBreakOffTriggerMult(2.0)

                self.unit:SetBreakOffDistanceMult(8.0)

                self.unit:SetSpeedMult(0.67)

                CIFMissileCorsairWeapon.OnGotTarget(self)

            end,

        

            OnLostTarget = function(self)

                self.unit:SetBreakOffTriggerMult(1.0)

                self.unit:SetBreakOffDistanceMult(1.0)

                self.unit:SetSpeedMult(1.0)

                CIFMissileCorsairWeapon.OnLostTarget(self)

            end,

        },

    },`

should be linked to that.

Issue #2321, opened 2017-11-26
don't allow corsair to fire his ground attack while in move order


shalkya:
doesn't allow the micro when you want to snipe something, since they will randomly shoot at stuff in the path and fuck up their fire cycle.
other fire fighter doesn't fire like this (janus/sera one)
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=7106185

PR #2361, opened 2018-01-15
Bomber speed fix.

FAETHER:
Issue: https://github.com/FAForever/fa/issues/2321 and https://github.com/FAForever/fa/issues/2322

Description: The Notha, Corsair and Janus no longer slow down on the move order. Other commands will cause slow down. 
There is also similar script for t1 bombers, however I did not touch this in currect PR. 
Corsairs will no longer attack ground units on the movem whence not slow down. They will attack with patrol, attackmove, attack, or idle. 
Other bombers did not have corsair problem, but they were still slowing down in enemy range for no reason (on move command). 

I know, that you guys dont have time to review my PRs or test them, so i made i video to enhance the first one.
https://www.youtube.com/watch?v=yqKmcBocXDg&feature=youtu.be

### DIFF
--- a/units/DEA0202/DEA0202_Script.lua
+++ b/units/DEA0202/DEA0202_Script.lua
@@ -24,10 +24,14 @@ DEA0202 = Class(TAirUnit) {
     end,
                 
     OnGotTarget = function(self)
-        self.unit:SetBreakOffTriggerMult(2.0)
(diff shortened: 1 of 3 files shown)


aeoncleanse:
Nice job this.

PR #2392, opened 2018-02-28
OC AoE fix

Strogoo:


shalkya:
it does apply on ACU and building, but doesn't get in count the armor of this unit.
so it tells you that you can kill them with OC, while you actually can't kill them.
either we remove the color for this units, or we take in count the armor. (400 damage max vs ACU, and % reduction vs building)

Strogoo:
Added code for ACU and structures. OC damage vs structures is kinda broken at this moment. So I adapted my code to it broken state. Will change it when/if OC damage will be fixed.

shalkya:
your aoe doesn't work if : you OC a low life building, and there is a brick next to it.
drain and damage will be calculated on the building and not on the brick. Leading to the brick take only few damage, and e drain being low.
same goes for ACU (the amount of life doesn't matter in this case). if you target the ACU, it will drain 5k e and do the minimum damage. Leaving the brick nearly full life.

lua/sim/DefaultProjectiles.lua
```diff
@@ -411,4 +420,44 @@ OverchargeProjectile = Class() {
...
+                else
+                    table.insert(unitsHP, unit:GetHealth())
+                end
+            end
+        end
+               
+        for _, unit in UnitsInSphere(launcher, self:GetPosition(), 13.2, categories.EXPERIMENTAL*categories.LAND*categories.MOBILE) do
+            -- Special for fatty's shield
+            if EntityCategoryContains(categories.UEF, unit) and unit.MyShield._IsUp then
+                table.insert(unitsHP, unit.MyShield:GetMaxHealth())
```
aeoncleanse: I suspect this won't work. What if the OC is fired at a unit near a fatboy from inside the shield? Why do we want to deal damage to the body according to the shield HP?

I'm also not a fan of doing two searches, though I see why you did. This method currently detects any mobile experimentals twice if they're inside the initial radius...

lua/sim/DefaultProjectiles.lua
```diff
@@ -411,4 +420,44 @@ OverchargeProjectile = Class() {
...
+            else
+                local distance = math.min(unit:GetBlueprint().SizeX, unit:GetBlueprint().SizeZ)
+                if GetDistanceBetweenTwoEntities(unit, self) < distance + self.DamageData.DamageRadius then
+                    table.insert(unitsHP, unit:GetHealth())
+                end
+            end
+        end
+        
+        if EntityCategoryContains(categories.EXPERIMENTAL, targetEntity) then
```
aeoncleanse: This could be solved by using the cylindrical unit search function instead of the spherical one, possibly?

lua/sim/DefaultProjectiles.lua
```diff
@@ -411,4 +420,44 @@ OverchargeProjectile = Class() {
...
+            table.insert(unitsHP, targetEntity:GetHealth())
+            --[[ we need this because if OC shell hitted top part of GC model its health won't be in our table
+            Bug appeared since we use shell.pos in getUnitsInSphere instead of target.pos.
+            Shell is too far from actual target.pos(target pos is somewhere near land and shell is near GC's head)
+            and getUnits returns nothing. Same to GetDistance. Distance between shell and GC pos > than math.min (x,z) size]]
+        end
+        if unitsHP[1] then
```
aeoncleanse: Instead of storing all this in a table etc, can't we do all the distance checks etc in one loop and simply do if newhp > returnhp then returnhp = newhp ?

lua/ui/game/unitview.lua
```diff
@@ -22,9 +22,27 @@ local GetUnitRolloverInfo = import("/lua/keymap/selectedinfo.lua").GetUnitRollov
 
 local selectedUnit = nil
 local updateThread = nil
-
+local unitHP = {}
```
aeoncleanse: We only store one value. Why a table and not a simple int variable?

lua/ui/game/unitview.lua
```diff
@@ -22,9 +22,27 @@ local GetUnitRolloverInfo = import("/lua/keymap/selectedinfo.lua").GetUnitRollov
...
 controls = import('/lua/ui/controls.lua').Get()
 
+function OverchargeCanKill()
+    local bp = __blueprints["ual0001"].Weapon[2].Overcharge --taking Overcharge values from Aeon ACU bp.
+    
+    -- this one is from DefaultProjectiles.lua OverchargeProjectile EnergyAsDamage()
+    local damage = (math.log((GetEconomyTotals().stored.ENERGY * bp.energyMult + 9700) / 3000) / 0.000095) - 15500
+	
+    if unitHP[1] and string.find(unitHP.blueprintId, "0001") and unitHP[1] > bp.commandDamage then
```
aeoncleanse: Are we definitely forced to use string ops and not just check the unit category? How does the mouseover info panel harvest data on the unit? Can we copy that?

lua/shield.lua
```diff
@@ -169,6 +169,15 @@ Shield = Class(moho.shield_methods, Entity) {
     end,
 
     ApplyDamage = function(self, instigator, amount, vector, dmgType, doOverspill)
+        if dmgType == 'Overcharge' and instigator.EntityId then 
+            local wep = instigator:GetWeaponByLabel('OverCharge')
+            if EntityCategoryContains(categories.COMMAND, self.Owner) then --fixed damage for all ACU shields
```
aeoncleanse: We have implemented faster single-category checks in FAF. To check a unit has, for example, COMMAND, use:

if unit:GetBlueprint().CategoriesHash['COMMAND'] then

It's faster than calling the function. Only works for checking single categories at a time, not categories.COMMAND*categories.LAND for example.

lua/sim/DefaultProjectiles.lua
Strogoo: Well, as you said "we can't be 100% accurate". And what is the point to OC some unit near fatty if you are already inside the shield and can just OC... mmm fatty? And if you firing at fatty from inside the shield, then you probably don't care about e drain (you should be prepared for the fight vs t4, it is not like some early fights vs t1/t2 spam when you really care about e). Anyway, we can still fix it (if distanceBetween ACU and fatty < X etc etc) but do we actually want to complicate the code for this pretty rare situation.

And as for searches, first one have only 2.7 radius and used for small units (yes, sometimes it can detects t4 twice)

lua/sim/DefaultProjectiles.lua
Strogoo: Yes, probably. But this way is simplier and faster. We don't need cylindrical search for other calculations, so just single "if" is better, I think.

lua/sim/DefaultProjectiles.lua
Strogoo: mmm Yeah, I will do that.

lua/ui/game/unitview.lua
Strogoo: yeah, it will be int when I implement your previous suggestion

lua/ui/game/unitview.lua
Strogoo: Yep, we can't just check it. As we are in UI and you can do :isInCategory only for user.Unit (but we need this for enemys):
this information from UpdateWindow(info) function is about your unit (it contains user.Unit table which you can use with isInCategory):

screenshot_3
 
and this is for enemy (there is no any "unit" table in it) :

screenshot_4

So I think it is hardcoded (as there should be only limited information about enemy units in UI).

lua/shield.lua
Strogoo: Ok, nice. I will do that

lua/sim/DefaultProjectiles.lua
aeoncleanse: I'm most worried about people aiming a shot at the fatty and having it intercepted. I suspect we can handle all these cases much better by looking at the shield type (Personal, personalbubble, bubble).

lua/ui/game/unitview.lua
aeoncleanse: I just realised this takes damage values from the Aeon ACU, hardcoded, which is really not OK. This is an easy way to let everything go wrong in future. 

We need a better way to approach this, for sure. We have the blueprint ID. We have __blueprints... A lookup in Blueprints is almost certainly better than stringop (lua hates stringops). That would allow us to check the unit category.

Given we only have this code running when a unit with OC is selected, we definitely should have data on the selected unit too, so we can get real weapon data rather than hardcoding the aeon ACU...

lua/sim/DefaultProjectiles.lua
```diff
@@ -411,4 +420,44 @@ OverchargeProjectile = Class() {
...
         return (math.log((energy + 9700) / 3000) / 0.000095) - 15500
     end,
+    
+    UnitsDetection = function(self, targetType, targetEntity)
+     -- looking for units around target which are in splash range
+        local launcher = self:GetLauncher() 
+        local unitsHP = {}
+        
+        for _, unit in UnitsInSphere(launcher, self:GetPosition(), 2.7, categories.MOBILE) do
+            if not EntityCategoryContains(categories.COMMAND, unit) then
```
ChrisKitching: Couldn't you just make the `UnitsInSphere` call above filter by `categories.MOBILE - categories.COMMAND` and do away with this check?

lua/sim/DefaultProjectiles.lua
ChrisKitching: The cylindrical search is implemented in native code (riiiight?). If so, and if using it lets you do fewer Lua<->native calls, I think you will find that overall it's a net speedup. Lua calls are vrey slow, and lua execution is slow. Picking a slightly more expensive native code function and simplifying a ton of Lua as a result is definitely a good choice.

lua/ui/controls/worldview.lua
```diff
@@ -215,6 +225,12 @@ WorldView = Class(moho.UIWorldView, Control) {
         elseif command_mode == "order" then
             if self:ShowConvertToPatrolCursor() then
                 self.Cursor = {UIUtil.GetCursor("MOVE2PATROLCOMMAND")}
+            elseif command_data.name == "RULEUCC_Overcharge" then
+                if OverchargeCanKill() == false then
```
ChrisKitching: You know you can just invert this if-block and get rid of the `== false`, right? Reads better, too: the reader doesn't have to perform a negation.

```lua
if OverchargeCanKill() then
    self.Cursor = {UIUtil.GetCursor(command_data.name)}
else
    self.Cursor = {UIUtil.GetCursor("OVERCHARGE2")}
end
```

It's especially worth noting that in lua, `nil` is sometimes used to represent non-truthiness. Explicitly checking against `true` and `false` is generally discouraged as a result: you almost always want to go for `if foo` and `if not foo`.

lua/ui/game/unitview.lua
```diff
@@ -22,9 +22,27 @@ local GetUnitRolloverInfo = import("/lua/keymap/selectedinfo.lua").GetUnitRollov
...
 local selectedUnit = nil
 local updateThread = nil
-
+local unitHP = {}
 controls = import('/lua/ui/controls.lua').Get()
 
+function OverchargeCanKill()
+    local bp = __blueprints["ual0001"].Weapon[2].Overcharge --taking Overcharge values from Aeon ACU bp.
+    
+    -- this one is from DefaultProjectiles.lua OverchargeProjectile EnergyAsDamage()
```
ChrisKitching: So you copy-pasted a gnarly equation from somewhere else instead of introducing a function? :D

What happens when someone wants to change this equation, and inevitably forgets to change one of the copies? A bug, that's what ;)

lua/ui/game/unitview.lua
Strogoo: "A lookup in Blueprints is almost certainly better than stringop (lua hates stringops)." 

Didn't know that. Ok, if ID -> __bp etc. way is faster, I will change it.


"Given we only have this code running when a unit with OC is selected, we definitely should have data on the selected unit too"

mmm, yep, it's doable via selected unit. But you can also OC if ACU + some tanks are selected (it is very very rare case, but still). So we need to check every selected unit (:IsInCategorey etc.) It will be a bit slower, but still better than hardcoded, I think.

lua/sim/DefaultProjectiles.lua
Strogoo: Honestly, I just didn't know syntax for category exclusion :D "-" is obvious, but still =)

lua/ui/controls/worldview.lua
Strogoo: Yes, you are right.

lua/ui/game/unitview.lua
Strogoo: Yep=) Will be changed.

lua/sim/DefaultProjectiles.lua
Strogoo: Do you mean "GetUnitsInRect"? Yes, we can use it. But after this we should exclude air units somehow. It will be VDist3(unitpos, shellpos) or ... :EntityCategoryContains so idk, still slower as for me. + One more thing, we already use "EntityCategoryContains" in one loop + in "shields" part + in "ACU" part. I doubt that removing it in "t4" part will give us noticeable performance boost (+ the fact that we still have to exclude air units if we use cylindrical search)

lua/sim/DefaultProjectiles.lua
Strogoo: In all of these cases OC deal full damage to fatty:
case 1
case 2
case 3
case 4

So idk what else do we need

lua/ui/game/unitview.lua
Strogoo: Ok, I tried 
local targetBp = __blueprints[unitHP.blueprintId].Categories
instead of
string.find(unitHP.blueprintId, "0001") / string.byte(unitHP.blueprintId, 3) == 98
Now i have table with all Categories:
1 - land
2 - mobile etc.


Next step will be

for _, value in pairs(targetBp) do
     if value == STRUCTURE/MOBILE/LAND/etc then
       return true
     end
 end

or what? Are you sure this way is faster than string.find/string.byte for unitId without all these blueprints, tables, loops etc. 

I'm a bit tired of this minor changes tbh, can we just left it as it is? I'm not talking about hardcoded aeon acu now, just about string.find/byte. As for me, hardcoded Aeon ACU is fine too, even if there will be some changes to OC in future, and even if someone will forget to change aeon bp, then not correct OC symbol won't be your main problem (and it will be solved right after bp fix). Anyway, I'm not "a boss" here, but all this minor stuff is pretty annoying.

aeoncleanse: Use CategoriesHash table.

And no, we can't. It's about keeping code quality, maintainability etc good. This is what code review is about.

lua/ui/controls/worldview.lua
Strogoo: Nvm. Since we want default green symbol for "idle" state (no unit under cursor), we can't just invert this. Otherwise I have to add some additional "if" to OverchargeCanKill() and this will ruin some logic there etc etc.

JaggedAppliance: We want the greyed OC icon for idle, green when you can kill the unit under the cursor.

Strogoo: *sigh*                                                                                                                                :D

Strogoo: Actually, with third symbol, code become really more complex. Better leave it as it is. (idle+can kill - green, can't kill - orange)

aeoncleanse: Shouldn't be too bad. UI already knows when it is active/inactive for the UI button to light up,so you should be able to easily check that.

Strogoo: Yeah, nvm. My mind is rusty today =( Will upload version with third symbol soon.

PR #2403, opened 2018-03-05
send message to server when the simulation ends

### LINKED ISSUE #2378: Send `GameEnded` on simulation end
Currently, it's pretty difficult for the server to figure out when a game has actually ended. The current rule is:

* Every connected player sent a game outcome (victory, defeat) for every other connected player
* Every victorious player sent game scores (0, 10) for every other connected player
* Every player disconnected

This isn't nearly implemented in the Python server; it just assumes a game to have ended when all players disconnected or when the **client** sends `Ended` - and the client does so when the game **process** closes. This causes games to last as long as the last player doesn't close his process - sometimes hours (the longest game in hour database supposedly lasted 216h).

By sending `GameEnded` when the simulation ends, this can easily be improved. I suggest not using `Ended` as this would have unwanted side effects with the current server implementation.

CookieNoob:
fixes #2378


micheljung:
This can be merged safely, the python server just ignores the command.

https://github.com/FAForever/server/blob/e930381ccaa63bdb05f91ecdd9ad557f6b8daa5f/server/gameconnection.py#L254

PR #2406, opened 2018-03-09
beta tweaks

Petricpwnz:
miss me with that balance shit
also somehow UEB4202 snuck in, 2lazy to fix but it doesnt seem any different from how it is so i dunno whats that about


CookieNoob:
why are you adding a new unit folder with the blueprint of the t2 UEF shield?
You added `units/UEB0402/UEB4202_unit.bp`

I suggest not to merge this PR without removing it first.

Petricpwnz:
I wonder how that happened considering i didnt even look at the shield, didnt rename any folders and surely didnt randomly rename any, anyway fixed

PR #2409, opened 2018-03-13
remove Campaign

PhilipJFryFAF:
changes the Campaign button to load the Skirmish menu since the Campaign
doesn't work anyway and coop missions should be loaded via the
skirmish/multiplayer menu

### DIFF
--- a/lua/ui/menus/main.lua
+++ b/lua/ui/menus/main.lua
@@ -105,7 +105,7 @@ function CreateUI()
         {
             name = '<LOC _Campaign>',
             tooltip = 'mainmenu_campaign',
-            action = function() ButtonCampaign() end,
+            action = function() ButtonSkirmish() end,
         },
         {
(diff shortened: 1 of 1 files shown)


CookieNoob:
coop missions need the coop mod though, when you start the game from the exe (which is the only situation when you see the main menu), this wont be loaded automatically. Thus, a coop mission wont work when it is loaded from the skirmish menu without additional modifications...

speed2CZ:
That's true, but it wont launch broken campaign anymore.

CookieNoob:
maybe just remove the button instead of replacing what it does?

PhilipJFryFAF:
Rackover suggested to create a popup that informs the user about campaign not working and explaining how to get it to work instead of just removing/changing the button
would save us time and energy explaining ppl the same stuff over and over
gonna close this PR for now

PR #2416, opened 2018-03-18
make tooltips generic

PhilipJFryFAF:
this PR adds some more information to the acu upgrade tooltips for the english version


HUSSAR-mtrela:
With my improved unitViewDetails.lua we will not need any values in tooltips because they will be calculated from blueprints. This way, we will never have outdated/wrong tooltips.

PhilipJFryFAF:
i'll remove my hardcoded stuff and also gonna change all the already existent values in the tooltips to something more generic

only thing that bothers me atm is that there are no localisations for the tooltips afaik

HUSSAR-mtrela:
@PhilipJFryFAF please change title of this PR to "remove hardcode..." since you are going to remove these stuff from tooltips.

PhilipJFryFAF:
changed the title to reflect its new purpose
also made a commit to remove all the numbers and to make the text more generic in general

HUSSAR-mtrela:
let's hold on with merging this PR until I create a new PR for generating values of upgrades/units in tooltips

PhilipJFryFAF:
i don't really see an issue with merging atm but we might as well wait for your feature to be done
maybe i can find some time to get some work slaves to change the other localization files too

PR #2419, opened 2018-03-22
Enhancement: This will enable the game to load custom AIs from Mod folder

Uveso:
Added function to load AI-data from AIs placed inside modfolder
Added tooltip and language loader function for AI lobby tooltips
Added preloading for AI lobby options

Updated lobby option behavior in case a new color/team/faction/AI is selected.
The old system was reloading the AI data on every option change.
Since i added disc access for finding AI-Mods this was getting to slow.
Now all AI data is loaded on the beginning of the lobby.lua file and stored inside a local for faster access.
Now there is less lag then before, so we can call this update or enhancement :)

### DIFF
--- a/lua/SinglePlayerLaunch.lua
+++ b/lua/SinglePlayerLaunch.lua
@@ -1,6 +1,7 @@
 -- Logic and defaults for launching non-skirmish sessions
 local Prefs = import('/lua/user/prefs.lua')
 local MapUtils = import('/lua/ui/maputil.lua')
+-- Uveso - aitypes inside aitypes.lua is now a function, not a table.
 local aiTypes = import('/lua/ui/lobby/aitypes.lua').aitypes()
 
 function GetRandomName(faction, aiKey)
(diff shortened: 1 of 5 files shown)


HardlySoftly:
11/10 rates this functionality

aeoncleanse:
Can this be expanded to handle factions too, easily?

Uveso:
This can't be use for factions.
Actual it works without modifying the lobby itself.
I am only gathering all AI data before the lobby is displayed and put the data in the right place.

(theoretical it should be possible with factions, but in this case its not only preloading some data. But i will write this on my ToDo list for 2018. )

CookieNoob:
@aeoncleanse I only know one mod that adds a faction (nomads) and that needs a custom init file which includes a function to clear the cached shader since supcom doesnt always realize that there are new shader methods... so for nomads, this can't work.

Uveso:
I also only know Nomads, and i doubt there will ever be another new faction.

Uveso:
Update:
Disc access for finding AI files is minimized (from up to 16+16 calls to 1+1 call)
All unneeded (debug) variables removed.

Uveso:
Update:
Added preloading for lobby AI options

PR #2421, opened 2018-03-25
Tele delay

Strogoo:
Ok, I made:
git reset --hard (first commit)
git revert (first commit)

Still not perfect, but good enough, I think=)


JaggedAppliance:
Update your development/balance and then git rebase onto that to remove the aoe commits

PR #2425, opened 2018-03-27
Remove selen maintenance cost

JaggedAppliance:
Remove the maintenance cost for stealth+cloak and fix the selen radar being disabled during power stalls


CookieNoob:
I already made a PR to dix this properly, it's just not merged yet
https://github.com/FAForever/fa/pull/2350

JaggedAppliance:
Yeah I thought I would put it here while I was changing the selen, it will get into the game faster this way.

PR #2431, opened 2018-03-28
Beetle rebalance

JaggedAppliance:
Beetleful rebalance.

WIP


units/XRL0302/XRL0302_unit.bp
```diff
@@ -213,7 +213,7 @@ UnitBlueprint {
         {            
             AboveWaterTargetsOnly = true,
             CollideFriendly = false,
-            Damage = 1999,
+            Damage = 1499,
```
speed2CZ: why such an ugly number?

JaggedAppliance: Is this a joke? xD The death weapon also does 1 damage.

PR #2432, opened 2018-03-28
ML Laser Tracking fix

Petricpwnz:
steal exotics eq ml tracking fix cause he is too much of a плохой человек to add it himself


speed2CZ:
I would prefer not having such commit names on git for sanity of future contributors

Petricpwnz:
Future contributors? Whats that? 🤔

speed2CZ:
just change it

PR #2443, opened 2018-04-19
HotFix for AI mod loader

Uveso:
This was a hook inside my AI mod but should be part of the FAF code.
It's preloading templates for the AI from AI mods.


lua/simInit.lua
```diff
@@ -186,6 +186,9 @@ function SetupSession()
     LOG('Loading script file: ', ScenarioInfo.script)
     doscript(ScenarioInfo.script, ScenarioInfo.Env)
 
+    -- Preloads AI telplates from AI mods
```
Sheeo: Wasn't this meant to say "templates"?

Uveso: oops :)

PR #2448, opened 2018-05-03
Tele delay value in bp

Strogoo:
Added TeleportDelay (seconds) to bp. Also changed delay from 15 to 10 seconds.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2266,9 +2266,11 @@ CommandUnit = Class(WalkingLandUnit) {
             self.TeleportThread = nil
         end
         
-        local teleDelay = 150 --Ticks
+        local teleDelay = self:GetBlueprint().General.TeleportDelay or 0
         local tick
         local seconds
+        
+        teleDelay = teleDelay * 10 -- We have seconds in bp (easier to use), so need to convert value in ticks.
         
         if teleportTime[self.EntityId] then
             tick = GetGameTick() - teleportTime[self.EntityId]
--- a/units/UAL0001/UAL0001_unit.bp
+++ b/units/UAL0001/UAL0001_unit.bp
@@ -802,6 +802,7 @@ UnitBlueprint {
         SelectionPriority = 3,
         TechLevel = 'RULEUTL_Experimental',
         UnitWeight = 1,
+        TeleportDelay = 10,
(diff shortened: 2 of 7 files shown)


lua/defaultunits.lua
```diff
@@ -2266,10 +2266,14 @@ CommandUnit = Class(WalkingLandUnit) {
             self.TeleportThread = nil
         end
         
-        local teleDelay = 150 --Ticks
+        local teleDelay = self:GetBlueprint().General.TeleportDelay or 0
         local tick
         local seconds
         
+        if teleDelay ~= 0 then
```
ChrisKitching: I don't see why you need to avoid doing the multiplication when the delay is zero? 0 * 10 is still zero, so nothing will break...

Strogoo: Yeah, you are right =)

lua/defaultunits.lua
```diff
@@ -2266,10 +2266,12 @@ CommandUnit = Class(WalkingLandUnit) {
             self.TeleportThread = nil
         end
         
-        local teleDelay = 150 --Ticks
+        local teleDelay = self:GetBlueprint().General.TeleportDelay or 0
```
speed2CZ: Shouldnt you keep it ...or 15... So it doesnt break modded or whatever units that rely on this?

Strogoo:
Ok, I combined commits (if that is what you wanted me to do;)

You can do a lot of stuff with that "git rebase", according to wiki. But there were no commits in fa:dev/balance after I made this branch, so I don't need to update my dev/balance (and this branch too, as it based on dev/balance), right?

Anyway, just let me know if I did smth wrong.

ChrisKitching:
Either I've completely lost the plot,or it looks like you still have the issue Speed pointed out. Mods that don't provide this blueprint value (AKA: All of them) will end up with zero teleport delay due to this change. Insta-teleport is unlikely to be the intended default behaviour...?

Strogoo:
Current tele has 15 seconds animation. We add additional timer (10 sec) with text notification, so after the patch you need 15+10=25 sec to jump. If there is no TeleportDelay in mod's blueprints then tele will perform same as before the patch (15 sec animation, not insta), without any text notifications etc.

PR #2450, opened 2018-05-05
t3 and navy

Petricpwnz:
Goverment certified.
something buffs the health of shards so they won't die to a single torp bomber in one shot
improves the maneuverability and range of Omen class battleships
health nerf/range buff for titans
health buff/range nerf for loyas
accuracy/muzzle velocity buff for sera battleships something


PhilipJFryFAF:
buffs the health of shards so they won't die to a single torp bomber in one shot
improves the maneuverability and range of Omen class battleships
health nerf/range buff for titans
health buff/range nerf for loyas
accuracy/muzzle velocity buff for sera battleships

speed2CZ:
well that info could be in the commit message

PR #2081, opened 2017-06-21
Fixed exit game for FAF and desktop launch

Uveso:
Related issues: #248 #619 #790

This is a fix for exiting the game.
FAF needs the game to fall back to the desktop (FAF Lobby).
But without FAF the game should exit to the main menu, not the desktop.

This is how it works now:

Game launched from FAF will return to desktop on exit.(SafeQuit()->ExitApplication())
Game launched from Desktop will return to the ingame menu on exit.(ExitGame())

Also fixed:
Ingamelobby shows now "exit" instead of "back" button if launched from FAF.
(added tooltip to exit and back button)
Scorescreen shows now "exit to FAF" instead of "continue" if launched from FAF
(fixed tooltip for "exit to FAF" button, added tooltip to "continue" button)
Removed Rehost button from LAN Multiplayer menu.
Added EscapeHandler.SafeQuit() to singleplayer game exit.

All tests are done with gameoption "Quick Exit" ON and OFF (different function calls!!!)
(FormClosingEvent == means klicking "x" in the right top corner of the game window.)

LOCATION Main menu:

ACTION menu "Exit" button:
- local start: Back to Desktop - ExitApplication()
- FAF-Lobby start:  Back to Desktop - ExitApplication()

ACTION FormClosingEvent:
- local start: Back to Desktop - SafeQuit()->ExitApplication()
- FAF-Lobby start:  Back to Desktop - SafeQuit()->ExitApplication()

ACTION [ALT]+[F4]
- local start: Back to Desktop - SafeQuit()->ExitApplication()
- FAF-Lobby start:  Back to Desktop - SafeQuit()->ExitApplication()

LOCATION Playing Multiplayer:

ACTION menu "Exit to Windows" / "Exit to FAF" button
- local start: Back to Desktop - SafeQuit()->ExitApplication()
- FAF-Lobby start:  Back to Desktop - SafeQuit()->ExitApplication()

ACTION menu Rehost
- local start: Rehost button removed.
- FAF-Lobby start:  Back to Desktop - SafeQuit()->ExitApplication()

ACTION FormClosingEvent:
- local start: Back to Desktop - SafeQuit()->ExitApplication()
- FAF-Lobby start:  Back to Desktop - SafeQuit()->ExitApplication()

ACTION [ALT]+[F4]
- local start: Back to Desktop - SafeQuit()->ExitApplication()
- FAF-Lobby start:  Back to Desktop - SafeQuit()->ExitApplication()

LOCATION Scorescreen:

[...]

PR #2455, opened 2018-05-07
Remove SessionEndGame when leaving game

Crotalus:
Found this old branch sitting around, not sure it's completely legit though. Have a vague memory that I saw illegal command (OP_CallEndGame) inserted in replay command stream, or at least sent via network command, can't remember which.

In any case, it could be worth a shot to try it out, the code it removes certainly didn't mitigate the desync issues (but maybe not causing any either)


ChrisKitching:
The original goal of the code that this removes is to make disconnects never lead to a timeout countdown. Before I added this, if someone just does ALT+F4 (ragequit :D) then the game would often behave as if their connection timed out. Everyone would stop, have to wait through the 1:45 countdown, and then eject them. It was extremely annoying because people would routinely do this when they were close to losing, making completing a game a painful experience.

I don't know what `ConExecute('ren_oblivion true')` or `ConExecute('ren_ui false')` are doing here: those were added since I originally did the work and I'm not sure what they're aiming for. I guess preventing people from pressing buttons?

Anyway: this code was never intended to "mitigate desync". We should probably become more confident about whether or not it's doing harm before we straight-up remove it, because (unless something else has changed...?) as written this PR will cause a rather annoying regression.

aeoncleanse:
Needs rebasing and further investigation as per @ChrisKitching

Uveso:
Since i updated the game behavior on exit to FAF / desktop, the function SafeQuit() is used in 19 places.
Are you sure the "safe" way is not needed in all 19 cases ?
#2081

Strogoo:
```I saw illegal command (OP_CallEndGame) inserted in replay command stream```

@Crotalus what parser do you use to see this stuff in replay? Contact Dragonite (FAF slack) if you have time and desire to help. He is working on new replay server and your advices would be really useful. ty

PR #2456, opened 2018-05-07
Trigger game full check when closing slot

Crotalus:


ChrisKitching:
Do you also need to call this from `ConvertObserverToPlayer`?

aeoncleanse:
Please rebase and check @ChrisKitching 's comment about moving observer to slot.

KionX:
Already in the lobby.

PR #2469, opened 2018-05-20
more faction diversity - more aoe for aeon and less for sera destro aoe

PhilipJFryFAF:
less aoe on the main weapon for sera destros and more for aeon destros


speed2CZ:
sera destro has 2 guns

PR #2471, opened 2018-05-24
HARMS fix

Strogoo:
Previous sinking function is commented (I'm not sure, should it be deleted or not)

Elevation = -3.8 is similar to sinkFor = 3.4 from old code (don't ask me why). -3.7 can be groundfiered by cyb destro.

Every single line of code in unit bp is needed (yes, even "EPERIMENTAL" category). You can remove smth and test it out. You will be surprised after every step, I swear :D


Strogoo:
Oh, didn't notice that HARMS is in both tabs now (t3 and exp). That's obvious, but still =( Ok, I'll try smth else tomorrow, because removing EXP category causes some bugs. Any suggestions are welcome.

shalkya:
fix #2472

Strogoo:
Ok, I made a lot of tests, tried different combinations and have to say, that it can not be done without 'EXPERIMENTAL' category. That's all about engine from this point. So I just removed HARMS from t4 tab (hotbuild works fine and doesn't need fixes).

ChrisKitching:
What does adding the experimental category achieve?

units/XRB2308/XRB2308_script.lua
```diff
@@ -16,7 +16,10 @@ XRB2308 = Class(CStructureUnit) {
 
     OnStopBeingBuilt = function(self, builder, layer)
         CStructureUnit.OnStopBeingBuilt(self, builder, layer)
-        self:StartSinkingFromBuild()
+        --self:StartSinkingFromBuild() --previous sinking code
```
ChrisKitching: ... Just delete it then? This is what version control is for.

Strogoo:
Made one more test, removed 'EXP' from Atlantis bp. Same 'jump' when dive command is executed.

https://youtu.be/QxZv8mq4ayE

units/XRB2308/XRB2308_unit.bp
```diff
@@ -59,6 +59,9 @@ UnitBlueprint {
         'OVERLAYANTINAVY',
         'SORTDEFENSE',
         'COUNTERINTELLIGENCE',
+        'EXPERIMENTAL',
```
speed2CZ: this will mess with AI, you cant just add exp cat to non exp unit

Strogoo:
Well, then we need another fix (smth with eggs or whatever was suggested in slack).

FAETHER:
I could integrate this by patching some category in the engine. So that there will be no need to do any bp or lua file changes. I put this on my TODO this.

PR #2474, opened 2018-05-25
 fix nil Entity,  rem dublicate

dk0x:
https://github.com/FAForever/fa/blob/dd04a04242bbdeb1a9774ec82da06e5cb764c621/lua/spreadattack.lua#L444

Need check for nil "if target then" like
`if target then
	Function({ unit }, target)
end`

because ShadowCopies of orders can be to old and Entity by Id not exist anymore(destroyed/reclaimed etc). 
Example IssueReclaim(tblUnits, target) looks like IssueReclaim({t1engi}, nil)
And its situation throw this exception

> WARNING: Error running sim lua callback function 'GiveOrders':
>          d:\fa-develop\lua\spreadattack.lua(444): Expected a game object. (Did you call with '.' instead of ':'?)
>          stack traceback:
>          	[C]: in function `Function'
>          	d:\fa-develop\lua\spreadattack.lua(444): in function `fn'
>          	d:\fa-develop\lua\simcallbacks.lua(16): in function <d:\fa-develop\lua\simcallbacks.lua:13>

and
make FixOrders function public(remove local). Its can be (and need) used outside spreadattack.lua with ShadowOrders
like

local ShadowOrders = import("/lua/spreadattack.lua").ShadowOrders
local FixOrders = import("/lua/spreadattack.lua").FixOrders
FixOrders(unit)
local unitOrders = ShadowOrders[unit:GetEntityId()]


aeoncleanse:
This is great. Can you rebase your branch onto the new develop, since we pushed a balance patch underneath it? This PR will auto-update.

aeoncleanse:
Re-rebased, squashed, and merged as this commit: cea6783c69742c495c4184c9e22632de84ea4cfb

PR #2487, opened 2018-06-07
fix OC wrong energy drain

Strogoo:
https://forums.faforever.com/viewtopic.php?p=164607#p164607

### DIFF
--- a/lua/sim/DefaultProjectiles.lua
+++ b/lua/sim/DefaultProjectiles.lua
@@ -6,7 +6,8 @@
 -----------------------------------------------------------------
 local Projectile = import('/lua/sim/Projectile.lua').Projectile
 local UnitsInSphere = import('/lua/utilities.lua').GetTrueEnemyUnitsInSphere
-local GetDistanceBetweenTwoEntities = import('/lua/utilities.lua').GetDistanceBetweenTwoEntities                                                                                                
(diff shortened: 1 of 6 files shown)


aeoncleanse:
@JaggedAppliance - Has this been tested working as intended?

JaggedAppliance:
@Strogoo @aeoncleanse 
It works correctly for the most part, however if you have multiple ocs happening and not enough energy to meet the minimum required energy consumption for oc, then it will fail and you will go into negative energy.

For example, you fire three ocs at once but only have 12k energy in storage when they land. The last one to land will drain almost 54k energy, putting you deep into the negative.

Strogoo:
Yeah, it is still possible to use multiple ocs even if you have less than '5k*number of OCs ' energy in storage. There was same problem in prev patch with old OC system, and with current 'energy darin on impact' system it's even harder to fix (if possible at all). But in this case (when you have not enough e) every OC should have min damage and e drain (1500dmg/~5k e). So '54k energy' for last shot sounds like a bug.

I'll test it and try to fix.

Strogoo:
works fine. Here is test with explanations:

https://youtu.be/3zaRyNFJr7o

JaggedAppliance:
Yes that clears it up, this is working as intended and can be merged @aeoncleanse

PR #2492, opened 2018-06-09
Add new props, textures and decals

CookieNoob:
using the material that Deep_ made in 2013, these additional map materials will help to create a new set of different looking maps.

http://forums.faforever.com/viewtopic.php?f=53&t=4651&p=49933#p50157

He created all of these assets by himself and ok'd its use.
> Yes all used textures and meshes are done by myself and sure i would like to share my stuff.


aeoncleanse:
Merged as 985f91e19e664847683e346d0b6ab7baab2225ea

PR #2494, opened 2018-06-10
Fix zero mass BC's wreck

Strogoo:


aeoncleanse:
Nice catch

PhilipJFryFAF:
tested and it works
nice and fast fix!

PR #2497, opened 2018-06-17
Disruptable tele

Exotic-Retard:
ok so Jagged promised me UNLIMITED POWER which is why im here.

also this change had his approval so if its shit blame him!

anyway:
Teleport was a bit of a strange mechanic because it often results in cheese and noob stomping because they dont know how to build pd in their base. This is mostly a result of telemaser, while non-telemaser teleport strategies are pretty weak in comparison.

This change addresses the root of the issue - the fact that a cybran acu can pop into your base and you cant stop it getting there, only kill it afterwards, and also its usually surprising if you arent paying attention.

By adding a target unit at the destination that allows the teleportation to be disrupted, nearby units will attack the target and if they kill it, the order is cancelled. This way, everyone gets out alive, and the target is much easier to kill harmlessly than some ultra t3 5vet insane hp ACU who is killing you with a laser.

The target removes the element of having to pay attention, since units will auto target it. Also, you need a lot fewer PD to kill it because the target has 10k HP and you have the entire duration of the teleport charge to take it down, not the time it takes for your acu to die.

this is a pretty large nerf for telemaser when teleporting into someones base, but if you go to spots that are safer then the upgrade is better, because its cheaper and doesnt get cancelled in safe areas. This way, teleporting in is more difficult and requires better placement of targets, making things like non-maser acus more viable for running around and building pds near enemy mexes and such.

what this change does is make teleportation more accessible to everyone (by buffing teleport for 3/4 factions), and in doing so gives more opportunities and valid strategies to the player, increasing the depth of the game and "skill ceiling" while also making it less hard on newer players.


Strogoo:
wow, Exotic is back in action, nice!

Only thing I don't like with in this system is that you can't do suicidal jump into enemy SMD surrounded by tons of t1/t2 pd, so your teammate can nuke his base later. I'll go and blame Jagged for that :D

*Is there any hope that you will make prs with some fixes too? At least the one which fixes asist for mobile shields/aa (dummy weapon problem), would be awesome. I like all EQ's fixes, sadly, most of them are not in faf balance =(

shalkya:
It gives the ability to spawn something that have 10k hp for 10k e drain.
this way you can try to distract unit from real threat or block shot toward an unit.
especially since you can respawn it all the time => You can create something that has virtually illimited health.

5cript:
There is one reason its abusable AF: Friendly fire. Have enemy strats/PD nearby and you can use those to kill targets instead etc. Even if it kill the teleporting ACU, it might still be worth it.

shalkya:
yes we already mentionned it to exotic retard, + some other issue eg : strat getting lured out of their base, so they can get catch easily by asf etc etc etc
this is mostly abandoned feature/idea afaik, just need jag call to close it

PR #2507, opened 2018-07-05
Various balance stuff

Petricpwnz:
Includes:
-TMD build time decrease
-Aeon destro splash increase
-Tempest cost decrease and hit consistency improvement
-Beetle remove ACU stun and building stun
-Shield boat stuff
-Swiftwind cost increase
-Aeon BS range fix and slight buff


units/UAS0401/UAS0401_unit.bp
```diff
@@ -430,8 +430,8 @@ UnitBlueprint {
             TurretBonePitch = 'Turret_Barrel01',
             TurretBoneYaw = 'Turret',
             TurretDualManipulators = false,
-            TurretPitch = 40,
-            TurretPitchRange = 50,
+            TurretPitch = 15, 
+            TurretPitchRange = 5,
```
Exotic-Retard: not sure whats going on there really, the low pitch range sounds like a great way to make it fail at shooting up nearby hills, i dont see how this fixes anything, at having it at 25 or so would be more sensible probably

Petricpwnz: it hit just fine in my tests so until you prove otherwise i say yes to this, it makes hit consistency MUCH better and allows muzzle velocity increase

Petricpwnz:
there, now instead of all relevant angles it will also hit the top of an astro crater edge while doing a barrel roll (yea it actually hit suprisingly well with just 5 degrees of wiggle).

Exotic-Retard:
im gonna be that guy and say that this needs splitting up into sensible commits with one change per commit (torpedo bombers together, exps together, ect.)

right now if youre not some game and github pro its hard to work out whats going on and it would be a pain in the ass to go back and change things, rearrange and so on later if theres a need for it.

get your local branch, do a soft reset then remake all the commits with sensible names, then do a force push to the branch and this pr should update, takes about 15 min, and in the future spend about 5 min writing good commit messages :>

speed2CZ:
As retard said, one change per commit please.

speed2CZ:
Im just gonna not care anymore, seems like a better option

PR #2509, opened 2018-07-05
adding "SPECIALLOWPRI" category to T2/T3 mexx/pgen

shalkya:
this allow T2/T3 Pgen/Mexx to be targetted in priority over other building

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,4 +1,4 @@
-Hotfix Patch 3698 (23 August, 2018)
+Hotfix Patch 3698 (24 August, 2018)
 ============================
 - Remove AI Takeover from ladder as it is intended to be an option for custom games
 


HUSSAR-mtrela:
I suggest using "TARGETLOWPRI" category which is more descriptive and explains that it is used for targeting.

shalkya:
if we rename the cat in the exe (that's what you want right ?), we also need to rename the cat in all the blue print. I may also appear several time per BP.
the benefice is not worth the amount of work here.

HUSSAR-mtrela:
Oh sorry, I thought that this was a new category and not existing one.

shalkya:
yes we are using a handy cat, that was nearly not used to categorize unit but that is already in place in all weapon target priorities of all unit, at the good spot for the targetting of mexx/pgen.

shalkya:
that's what we are doing.... the name is just miss leading

(read back the discussion if you wonder why we used this name)

shalkya:
the targetting of ACU falls into the 'ALLUNITS' cat, which lead ACU to be targeted after T2/T3 mexx/pgen. I'm not sure we want this.
Any fix for this that i can think of, would require massive change in the BP of hundred of unit, which i'm not willing to do.

shalkya:
fix : 
#2546 #2547 #2548 #2549 #2550 #2551 #2552 #2553 #2554 #2555 #2556 #2557 #2558 #2559 #2560 #2561 #2562 #2563 #2564 #2565

Issue #2288, opened 2017-10-23
Move command does not work on some T4 that are under construction


HUSSAR-mtrela:
These T4 units can queue move command when they are under construction:
- CYBRAN Megalith: Experimental Megabot
- AEON Czar: Experimental Aircraft Carrier
- AEON Tempest: Experimental Battleship
- UEF Fatboy: Experimental Mobile Factory
- UEF Atlantis: Experimental Aircraft Carrier

These T4 units can NOT queue move command when they are under construction:
- SERAPHIM Ahwassa: Experimental Bomber
- SERAPHIM Ythotha: Experimental Assault Bot
- CYBRAN Monkeylord: Experimental Spiderbot
- CYBRAN Soul Ripper: Experimental Gunship
- CYBRAN Scathis: Experimental Mobile Rapid-Fire Artillery
- AEON Galactic Colossus: Experimental Assault Bot

Probably, these units are missing selectable category/functionality.

PR #2510, opened 2018-07-07
added "CQUEMOV" cat

shalkya:
allow the ability to select experimental and offensive/defensive structure while being built.
need to be use in combination of the https://github.com/FAForever/FAPatch/pull/6

### DIFF
--- a/units/UAB2101/UAB2101_unit.bp
+++ b/units/UAB2101/UAB2101_unit.bp
@@ -50,6 +50,7 @@ UnitBlueprint {
         'RECLAIMABLE',
         'OVERLAYDIRECTFIRE',
         'SORTDEFENSE',
+        'CQUEMOV',
     },
     CollisionOffsetY = 0,
     Defense = {
--- a/units/UAB2101/UAB2101_unit.bp
+++ b/units/UAB2101/UAB2101_unit.bp
@@ -50,7 +50,7 @@ UnitBlueprint {
         'RECLAIMABLE',
         'OVERLAYDIRECTFIRE',
         'SORTDEFENSE',
-        'CQUEMOV',
+        'CQUEMOV', --"CQUEMOV" enables the selection and move commands during construction
     },
     CollisionOffsetY = 0,
     Defense = {
--- a/units/UAB2104/UAB2104_unit.bp
+++ b/units/UAB2104/UAB2104_unit.bp
@@ -50,6 +50,7 @@ UnitBlueprint {
         'RECLAIMABLE',
         'OVERLAYANTIAIR',
         'SORTDEFENSE',
+        'CQUEMOV',
     },
     CollisionOffsetY = -0.45,
     Defense = {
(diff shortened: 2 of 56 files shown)


Exotic-Retard:
is it possible to rename the cat to something more obvious?
i have no idea what CQUEMOV might be, but SELECTWHILEBUILDING is a lot more self explanatory

HUSSAR-mtrela:
yes, please use `SELECTWHILEBUILDING` for this new category

FAETHER:
No, I need the category to be the same length as factory.

Exotic-Retard:
in that case would this work? its the same length:
BUILDSLCT
CQUEMOV

HUSSAR-mtrela:
@FAETHER please explain why does it have to be the same length as `FACTORY `category?

FAETHER:
It has to be the same length because it is easier to implement, leaves no chance of a bug to occur. Because the stack would be on identical memory adresses as the previous call. My goal is to ensure 100% bug free solution. Also - I don't understand why cquemov is not descriptive for you. C - construction. Que - que a command. Mov - anything related to movement.

Exotic-Retard:
yeah its all a case of how people think i guess xD for example i had no idea what the C could stand for. its not super important as long as it works. i think the only way to get something that most can understand is simply to ask a lot of people what they think the best one is, for me its BLDSLCT since its to do with building and selecting, rather than queueing and moving, which are more specific uses of the category

HUSSAR-mtrela:
@shalkya thanks for explaining and maybe we should drop "C" and append "E":
`QUEMOVE` instead of
`CQUEMOV`

HUSSAR-mtrela:
btw, I think this PR should be merged to FAForever:development/balance  and not develop branch

FAETHER:
Actually. Should be mergedon both ends and also new .exe file as well. So that nothing will get regressed. Similar to what has happened to wall selection fix.

FAETHER:
It probably tested for lua or something, idk. This is only blueprint changes, so should not be a hassle. Keyser did this pr, so he probably tested this very well. Can we just ignore travis build? I don't see any errors.

shalkya:
there was indeed 4 missing commas, though i'm not entirely sure, it was my fault. (might be completely wrong though)

CookieNoob:
it would be a good idea to add a comment that explains this category somewhere. This can be in the blueprints (just do a file search and replace) or at some other location (where it is easy to find).

shalkya:
something like -- CQUEMOV means "construction queue movement"

Sheeo:
Where's the corresponding exe patch code?

I am strongly against using this category name and I'm not convinced the length of it makes the binary patch any more "bug free" than not -- if it relies on the length of the category name it is by definition bugged.

HUSSAR-mtrela:
@shalkya did you forget to reference issue # 2288 in this PR?

shalkya:
ah yeah, it fixes #2288

PR #2514, opened 2018-07-10
Added chat colors in the lobby depending on who's talking

Rackover:
This commit adds colors and styling options (different fonts) into the lobby by making use of a new class for this chat, defined in chatarea.lua.
With this commit, people in the lobby have their name colored with their currently selected faction color. Grey if observer.
image

(huge thanks to hussar for the help and reviewing provided)


HUSSAR-mtrela:
@Rackover you are welcome.

PR #2519, opened 2018-07-12
Avatar in lobby

Super4121:
Gets information from the client. Takes an avatar from the cache.

Lobby + Client "https://github.com/FAForever/downlords-faf-client/pull/942"

PR #2522, opened 2018-07-13
Fix var names, sys path... #2519 (avatar in lobby)

Super4121:
-now use fullname for vars (urlava->AvatarUrl etc.)
-Change CMDLINE_ARGUMENT_KEYS to  ["/avatartooltip"]
-variables in playerdata  changed from boolean (false) to string ("none")
-path to avatar files changed from system (C:/programmdata/faforever....) to local (/../../cache/avatars/)
#2519 @CookieNoob


PhilipJFryFAF:
i will just close this PR for now since we won't integrate avatars in the lobby for now and the other PR was reverted by icedreamer

Issue #2277, opened 2017-10-03
Dummy weapons on shields and AA causing problems for AI


speed2CZ:
AI moves around units in formation, before the shields and AA sticked in the formation, protecting what they should have, when other units stopped to shoot, the shields stopped with them and then continues to move together.

Now they stop long before that exposing all units they should be protecting.

I guess with AA it doesnt hurt that much if they stay behind, but shields are a problem.

MrRowey:
@Garanas Close has a Merged PR and everyone seemed to be happy to with the changes with the merge

PR #2524, opened 2018-07-14
Fix assist for mobile shields (EQ)

Strogoo:
---------------------------------------------------------------
Fix from EQ
https://github.com/Ithilis/Equilibrium/commit/e729f6d72244b7cf588c3767a723b330aac1a22c#diff-df0eb38b9064089cb5ba809139ab9e4e
All thanks to Exotic
---------------------------------------------------------------

#2277

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -250,3 +250,17 @@ Callbacks.AttackMove = function(data, units)
     end
     IssueAggressiveMove(units, data.Target)
 end
+
+--tells a unit to toggle its pointer
+Callbacks.FlagShield = function(data, units)
+    units = SecureUnits(units)
+    local target = GetEntityById(data.target)
+    if units and target then
+        for k, u in units do
+            if IsEntity(u) and u.PointerEnabled == true then
+                u.PointerEnabled = false --turn the pointer flag off
+                u:DisablePointer() --turn the pointer off
+            end
+        end
+    end
+end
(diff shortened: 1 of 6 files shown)


shalkya:
what does that fix ?

Strogoo:
Mobile shield stops moving if it's assisting friendly unit and there is enemy nearby. That's because of dummy weapon, which was added to fix attack-move for mobile shields.

Now both (attack-move and assit) work fine.

shalkya:
would it be possible to do the same on flak by any chance ?

Strogoo:
Yes. The thing is that this code starts additional thread for every mobile shield (when you give them "assist" command) and every second it checks if unit is still assisting. If not, then it turns on dummy weapon and kills the thread.

So this fix can have some impact on performance if you have too many shields/aa which are assisting other units.

Anyway, if IceDreamer doesn't mind, I can add it to mobile AA. (have to mention that Exotic did it only for mobile shields, so idk)

JaggedAppliance:
I think just doing it for shields is fine, they have to be in more precise positions than flak and I'm wary of performance impacts.

speed2CZ:
Can it be split into more commits, one where you add the original files, second where you clean them up and third where you actually make the fixes, so its possible to see what's actually changed?

Strogoo:
Done. I'm not sure how I should clean these files tbh ;) Probably some tab/4-spaces stuff etc? They all looks similar in Notepad++, so yeah, I definitely need some instructions =)

speed2CZ:
yeah, spaces, replacing # comments with --, != with ~= and such

shalkya:
reviewed and tested

PR #2530, opened 2018-07-21
Save HQ/mex upgrade progress after ownership transfer

Strogoo:
HQ/Mex and progress > 5%

Dying on Seton's mid with 3 80% mexes and half done t2 HQ is not so painful anymore ;)

Sadly, this trick doesn't work with Nuke/smd. Maybe I'll find another way.


shalkya:
tested and approved.

doesn't work for SCU upgrading. (maybe you want that only for building?)

PhilipJFryFAF:
questionable whether or not this will be accepted by the game/balance councilor

shalkya:
i mostly wondering about pausing the given unit, so it doesn't crash the eco of the guy you gave the unit to.

Strogoo:
I can add "Ultimate Full Share" option :D (also pause after share, if needed)

Yeah, SCUs use a bit different code for upgrades. I will add unfinished t4s and then take a look at scus.

PhilipJFryFAF:
changed the branch to deploy/fafdevelop since that's what we merge in from now on

lua/SimUtils.lua
```diff
@@ -182,10 +206,40 @@ function TransferUnitsOwnership(units, ToArmyIndex)
...
     end
+    
+    if upUnits[1] then
+        ForkThread(upgrade, upUnits)
+    end
 
     return newUnits
 end
 
+function upgrade(units)
```
CookieNoob: It would be helpful to give this function a more descriptive name...

Strogoo: yep, you are right. I'll change it to "upgradeTransferredUnits"

Strogoo:
Added unfinished EXPs and t3 arty. They will be transferred when owner died (no manual option). 

This transfer is pretty complex and we don't want to transfer too many units this way. So it doesn't contain nukes/smd buildings, as their cost are too low compare to t4/arty. 

As for sacus, they use some scripts for upgrades and I don't really want to spend  time on them. So let's just leave it as it is ;)

JaggedAppliance:
Is there a way to do this for nuke/smds? It would be great to include those if possible. If not this is still good. Leaving out SACUs is fine.

Strogoo:
Yep, all this stuff works only in "Full share" games.

As for nukes, the trick that I used for upgrades/t4 doesn't work with them. So no nuke/smd, at least for now :(

shalkya:
can you add that feature to hives/kennel ?
could you also be careful about it, since they already have a feature that pause them. (may not want to create duplicate of code, maybe fuse the both in your code, and remove previous one...)
also while you are at it can you fix that ? https://github.com/FAForever/fa/issues/2605 We get this issue from before, but it impact your pr if we put the kennel in your pr.

Strogoo:
Exotic suggested to make a separate function for actual transfer process so it can be used in other cases rather than after owner death. However, it's not so easy to do because of many reasons, so after short discussion I decided to give more specific name to transfer function and don't touch the code.

Strogoo:
Fixed engineer stations and #2605

shalkya:
image
image
gonna post that here in case you wanna take a look but it is in the actual game (not a regression)

shalkya:
tested works.

PR #2534, opened 2018-07-23
Wreck fix: default mass  from 81% to 90%

Strogoo:


PhilipJFryFAF:
while this PR would make the code more sensible it also messes with an established balance
so i'll close this PR for now since @JaggedAppliance doesn't support this idea
i recommend talking to the responsible people before investing time into a feature that might get outright rejected

JaggedAppliance:
@PhilipJFryFAF @Strogoo there's no need to contact people prior to doing something like this. I'm sure strogo knows that things can be rejected. For large features ofc it's better to do it in consultation with others but for smaller things it is actually preferable that people put up PRs.

CookieNoob:
what about changing the values in the blueprints to adjust the percentage? that way this bug is fixed and the value is kept the same

PhilipJFryFAF:
that would require us to have every unit bp on the repo
imo the code logic is not important enough in that case to justify such a change

PR #2537, opened 2018-07-26
SimSpeed patch.

Uveso:
Removed some functions like `:IsDead()` and used values directly (`.Dead`)

:IsDead() -> .Dead
:GetUnitStats() -> .UnitStats
:IsDefeated() -> .Result == "defeat"
:AIGetLayerPreference() -> .LayerPref
:PBMHasPlatoonList() -> .HasPlatoonList
:GetRadius() -> .Radius
:GetCurrentEngineerCount() -> .CurrentEngineerCount
:GetMaximumEngineers() -> .EngineerQuantity
:GetPriority() -> .Priority
:GetBuilderName() -> .BuilderName
:GetLocationRadius() -> .Radius

### DIFF
--- a/lua/AI/AIBehaviors.lua
+++ b/lua/AI/AIBehaviors.lua
@@ -583,7 +583,7 @@ CommanderOverrideCheck = function(self)
     local weaponRange = mainWeapon:GetBlueprint().MaxRadius + 50 -- Look outside range.
 
     local commanders = aiBrain:GetUnitsAroundPoint(categories.COMMAND, self:GetPlatoonPosition(), weaponRange, 'Enemy')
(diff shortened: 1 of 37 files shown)


speed2CZ:
How much does it help?

Uveso:
Well, hard to say.

I changed about 200 function calls.
Many function calls are inside while loops and are looping every tick overt all units.
This will affect about 2400-2600 calls every 0.1 second.

Benchmark shows a smal increase from 18725 to 18732

So the game is 0.037% faster :D

But if you use more units and the game is at its limit, then we have a slight improvement.
But never more then 1% game speed.

Issue #2472, opened 2018-05-24
harms aren't targeted by surface weapon while being built


shalkya:

PhilipJFryFAF:
fix for this issue is still needed and would be greatly appreciated (since Strogos attempt didn't work out)

shalkya:
faether did make an exe that allow harms to get targeted while being build, while not using 'exp' marker. Last time i checked harms was not targeted while settling at the bottom, and they was targetting only land unit (or being considered as a land unit, can't remember).

PR #2540, opened 2018-07-28
HARMS fix (dummy unit)

Strogoo:
Issue #2472

Attackers lose target when dummy HARMS is finished. But if you scouted dummy HARMS during construction, then real one will appear on your map after sinking animation.

If someone has any ideas how to improve this part in terms of performance, let me know.

https://github.com/Strogoo/fa/blob/HARMS-fix-dummy/units/XRB2308/XRB2308_script.lua#L36


PhilipJFryFAF:
well done!
gonna wait for uveso to take a look at the code to tell me if there is something that can be improved and ask jagged if he is ok with this sort of change

Exotic-Retard:
you dont need to add duplicates of the texture files.
in the bp specify the full path to all the textures, meshes in for all the LODs to the other harms and it will work without you bloating file size.

Exotic-Retard:
also to detect if its on intel you should use the units blip.
Sim.ReconBlip
is the class, you can find it with unit:GetBlip()
and it uses useful things like blip:IsOnRadar()
so you can ask it if its seen by something, since harms is a building you want IsSeenEver when constructions is complete and if the answer is yes then the newly made harms should be detected. prolly still with some quick vision blip but i dunno maybe can use some trickery there too.

shalkya:
assist harms won't rebuild it automatically.

Strogoo:
Yeah, I thought it works only for units that can be rebuilt by placing new project on wreck (and you can't do it with HARMS wreck, because it sinks). However, wreck just teleports and construction starts from 50%, nice :)

Strogoo:
Nvm, you can also rebuild it in a normal way. Just need to place new HARMS right above the wreck. That's smth new for me =) Yeah, need to fix it somehow then

shalkya:
yeah it is really annoying to do, especially that you need to wait the death animation.
That's why the auto-rebuild is pretty useful. Not sure if lot of people use it on harms though

projectiles/Sinker/Sinker_script.lua
```diff
@@ -56,7 +56,9 @@ Sinker = Class(Projectile) {
     OnImpact = function(self, targetType, targetEntity)
         if targetType == 'Terrain' then
             self:Destroy()
-            ForkThread(self.callback)
+            if self.callback then
```
Uveso: Why is this needed now? Maybe first check if one of your (too early) self:Destroy() is causing this ?

Uveso:
XRB2308_Script.lua

Obsolete line 10:
local utilities = import('/lua/utilities.lua')

You removed the DeathThread where this local was used. please also delete the local utilities line

units/XRB2308/XRB2308_script.lua
```diff
@@ -13,135 +13,51 @@ XRB2308 = Class(CStructureUnit) {
...
-        self:ForkThread(self.SinkDestructionEffects)
-
-        -- Avoid slightly ugly need to propagate this through callback hell...
-        self.overkillRatio = overkillRatio
-
-        local this = self
-        self:StartSinking(
-            function()
-                this:DestroyUnit(overkillRatio)
+        local x1 = pos[1] - 60
```
Uveso: There are some units with a higher visionradius then 60. Better use 100 here

units/XRB2308/XRB2308_script.lua
```diff
@@ -13,135 +13,51 @@ XRB2308 = Class(CStructureUnit) {
...
+        local z2 = pos[3] + 60
+        
+        
+        while true do
+            local UnitsInRec = GetUnitsInRect(Rect(x1, z1, x2, z2))
+            for _, unit in UnitsInRec do
+                local unitArmy = unit:GetArmy()
+                if unitArmy ~= army and not IsAlly(unitArmy, army) then
+                    local unitPos = unit:GetPosition()
+                    local visionRadius = unit:GetBlueprint().Intel.VisionRadius
```
Uveso: I am not sure, but we don't check `WaterVisionRadius`. Is this not needed ?

units/XRB2308/XRB2308_script.lua
```diff
@@ -13,135 +13,51 @@ XRB2308 = Class(CStructureUnit) {
...
-        self:StartSinking(
-            function()
-                this:DestroyUnit(overkillRatio)
+        local x1 = pos[1] - 60
+        local z1 = pos[3] - 60
+        local x2 = pos[1] + 60
+        local z2 = pos[3] + 60
+        
+        
+        while true do
```
Uveso: Even if the forked thread will be removed when a unit dies, it can still cause errors if you call it while the unit is beeing destroyed. Use `while not self.Dead do` instead.

Uveso:
After testing this with 500 Airplaines i got an buffer overrun Error:

```INFO: Wow!  Ran out of segment buffers from the pool, discarding segments!```

Please check this.

File: units/XRB2308/XRB2308_script.lua [hunk 5]
```diff
@@ -13,135 +13,51 @@ XRB2308 = Class(CStructureUnit) {
     Weapons = {
         Turret01 = Class(CKrilTorpedoLauncherWeapon) {},
     },
-
+    
+    OnStartBeingBuilt = function(self, builder, layer)
+        CStructureUnit.OnStartBeingBuilt(self, builder, layer)
+        self.SpotterThread = self:ForkThread(self.SpotterThread)
```
Uveso: Should be inside the trashbag in case the unit will be destroyed while building
`self.Trash:Add(self.SpotterThread)`

units/XRB2309/XRB2309_script.lua
```diff
@@ -0,0 +1,163 @@
...
+            local pos = self:GetPosition()
+            local spec = {
+                X = pos[1],
+                Z = pos[3],
+                Radius = 4,
+                LifeTime = 0.3,
+                Army = self.SpotterArmy,
+                Vision = true,
+            }
+            local vizEntity = VizMarker(spec)
```
Uveso: Why is the return value saved inside `vizEntity`  ? It's not used after this line.

units/XRB2309/XRB2309_script.lua
```diff
@@ -0,0 +1,163 @@
...
+        self.Trash:Add(CreateAttachedEmitter(self, 'xrb2308', army, '/effects/emitters/flash_04_emit.bp'):ScaleEmitter(2))
+
+        self:DestroyAllDamageEffects()
+        self:PlaySound(bp.Audio.Destroyed)
+
+        -- Here down is near-shadowing the function, all to change the entity subset. Dumb, right?
+        local isNaval = true
+        local shallSink = true -- This unit should definitely sink, no need to check cats.
+
+        WaitSeconds(utilities.GetRandomFloat(self.DestructionExplosionWaitDelayMin, self.DestructionExplosionWaitDelayMax))
```
Uveso: Why we need a random wait between 0 and 0.5 seconds and import `utilities` for this ?
Wouldn't it be better to just wait 0.5 seconds in any cases and don't need to import an extra function ?

projectiles/Sinker/Sinker_script.lua
Strogoo: We have this bug in current patch too. Appears when you build HARMS on shallow water. So this check is sort of lazy fix ;)

units/XRB2308/XRB2308_script.lua
Strogoo: Yeah. I will probably remake the whole system, since Exotic told me about some sweet functions

units/XRB2309/XRB2309_script.lua
Strogoo: Lazy copy paste, sry for that

units/XRB2309/XRB2309_script.lua
Strogoo: https://github.com/FAForever/fa/blob/deploy/faf/units/XRB2308/XRB2308_script.lua#L89 This is just a copy of current HARMS, so I have no idea tbh. I can replace it with 0.5 if you want

Strogoo:
Spotter thread and unnecessary textures have been removed. Thanks to Exotic.
Fixed rebuild from wreck and sacu assist.
WaitSeconds(GetRandomFloat) replaced by WaitSeconds(0.5)

units/XRB2309/XRB2309_script.lua
```diff
@@ -0,0 +1,155 @@
...
+
+        self.OpenAnim = CreateAnimator(self)
+        self.OpenAnim:PlayAnim(bpAnim)
+        self.Trash:Add(self.OpenAnim)
+        self:PlaySound(bp.Audio.Deploy)
+        
+        local pos = self:GetPosition()
+        
+        for _,army in self.SpottedByArmy or {} do
+            VizMarker({X = pos[1], Z = pos[3], Radius = 4, LifeTime = 0.3, Army = army, Vision = true,})
```
Exotic-Retard: why is this passed to the child harms? why not create the viz markers in the parent harms just as its being removed. that way there will be no delay between the old harms being spotted and the new one

units/XRB2309/XRB2309_script.lua
```diff
@@ -0,0 +1,155 @@
...
+        self.Trash:Add(CreateAttachedEmitter(self, 'xrb2308', army, '/effects/emitters/flash_04_emit.bp'):ScaleEmitter(2))
+
+        self:DestroyAllDamageEffects()
+        self:PlaySound(bp.Audio.Destroyed)
+
+        -- Here down is near-shadowing the function, all to change the entity subset. Dumb, right?
+        local isNaval = true
+        local shallSink = true -- This unit should definitely sink, no need to check cats.
+
+        WaitSeconds(0.5)
```
Exotic-Retard: i know uveso said not to do it but its part of the original death thread, and the variable delay is just a quality of life improvement, so that when you destroy many harms there is a variation in their death animations that just adds a little quality to the game. i would put it back tbh. its not a performance hog.

units/XRB2309/XRB2309_script.lua
Strogoo: Yeah, that's what I did in my first tries. But in that case spotted HARMS looks weird 
screenshot_3

So we have to spot it at its final position. Can also increase VizMarker lifetime, so it gives vision during all animation, but I think the short one in the end is better.

units/XRB2309/XRB2309_script.lua
Strogoo: Ok, I'll revert it then :)

PhilipJFryFAF:
cannot merge since there is some conflict that needs to be resolved @Strogoo

Strogoo:
Ok, don't merge it. I'll see what I can do

Strogoo:
*Doesn't work with shift-orders. Actual HARMS will attack first target from queue and then go random.

I will squash commits after code review.

PhilipJFryFAF:
@Strogoo ready for testing/merging?

Strogoo:
@PhilipJFryFAF Yes

PR #2543, opened 2018-07-29
remove player color that is to close to civilian one

Rackover:
… other colour

You can see the new colour on the right here : 
(img)https://i.imgur.com/P1iCPY0.jpg
I put it with the nearest colours available (light blue, grey, blue) and it visually conflicts with none.

### DIFF
--- a/lua/GameColors.lua
+++ b/lua/GameColors.lua
@@ -15,7 +15,6 @@ GameColors = {
     "FF2929e1",      -- UEF blue
     "FF5F01A7",      -- dark purple
     "ffff32ff",      -- new fuschia
-    "ffffbf80",      -- lt orange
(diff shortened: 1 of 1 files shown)


Uveso:
Well i am red/green color blind (Deuteranomaly -> green-weak)
and now civilians look like every other player for me. :(

I don't like this change.
Would be nice if you could make this barrier-free.

shalkya:
image
top left is the new player color.
on the right the civilian color that didn't change. 
all the 3 other colors are player colors that didn't change, but are close to the new one for a comparison.

gonna update the title so people don't get this wrong anymore

Uveso:
Oh yes, these screenshoots are way better.
Now i see the changes and know what you mean.

Thank's, can't wait for this PR :D

PR #2545, opened 2018-08-02
Remove reclaim preventing death explosions

Exotic-Retard:
#WhereWouldWeBeWithoutEQ
yeah turned out to be a quick fix this time xD
pretty much the title. when a unit dies it should die normally, and not
be able to prevent its death explosion by reclaiming it. now it dies
correctly. reclaiming without killing involved should result in no explosions, like normal. so that your base doesnt explode when you misplace a fab.

NOTE: im gonna be away for about a month after this commit so wont
respond to all the bugs that are gonna be dug up, also this is poorly
tested so some qa boys want to get on the case there. but yeah lets just
hope for the best.


PhilipJFryFAF:
just checked locally with some gunships shooting a t3 pgen while it is being reclaimed to death
if i reclaim it to death then no explosion 
if the gunship shot is the last hit then it explodes

looks good to me

PhilipJFryFAF:
jagged prefers having the ability to prevent pgen explosion damage by reclaiming the building

aeoncleanse:
Reopened pending council vote

PhilipJFryFAF:
has there been any vote on this yet?
last thing i heard was that it would make the game less predictable and thus it was not approved (by @JaggedAppliance )

PR #2570, opened 2018-08-18
An attempt to fix strats. Feel free to leave feedback

Strogoo:


shalkya:
tested, works

speed2CZ:
This should be either for all bombers or no bombers. T1/T2 bombers can kill Inties as well.

shalkya:
yeah it's so sexy when you ground fire an inte chasing down a T1 bomber :D

Strogoo:
Well, Jagged still not sure about this change for strats, so idk. Personally, I don't care too much about other bombers. A group of strats is more like a game ender (in terms of mass and their power) and other bombers are just... bombers :). But I can change them all if needed.

JJ173:
I disagree with this change entirely. It's a part of some crazy methods you find in the game, and is quite funny to play around with. I'm pretty sure this would have an effect on the ground fire capabilities of bombers that target other planes that are on the ground.

FAETHER:
But why is this needed? I disagree, hate when projectiles do not collide it makes the game look not realistic and ruins the immersion. Just imagine a projectile flying through the units texture. No matter how bad the balance can be affected, you can always fix this some other way, than just removing what was originally intended by creators is very dumb in my opinion. Its like cutting the code mechanic the game is known for, what is it guys, you are aiming to make it look like Starcraft 2 or something? Where everything is hardcoded to hit only intended type of units. Now in game this happens rarely because players know that its a thing, so why won't fly bombers and shoot into the wall (astrocarter map for ex). I dont see this a big balance issue either. Also if you were to ask my opinion on removing the novax beam collision with the nuke, my arguments will be the same as here. Peace out.

JaggedAppliance:
We'll try changing the strat bomber formation to be more spread out before looking at removing collision on these projectiles.

Strogoo:
mix
second

shalkya:
tested, works as expected. (strats are immune to owner other strats)

PR #2572, opened 2018-08-19
Fix units lose veterancy after transfer

Strogoo:
vet 2,3,4,5 units become vet 1 after transfer because of this code
```
-- Limit the veterancy gain from one kill to one level worth
    massKilled = math.min(massKilled, self.Sync.myValue)
```
now there is separate function for transferred units.

Also added massKilledTrue value (which is used by "Mass killed" feature) to the transfer process, so now it will be displayed for transferred unit too. And fixed little UI bug (If you select full veted unit and after that select smth else, then "Mass killed" text won't change to "Veterancy")


shalkya:
tested and approved

Issue #2571, opened 2018-08-19
AI Replacement in Ladder


JJ173:
Reports from various ladder players have concluded that AI Replacement is set to fire in Ladder by default for some unknown reason. The option has been set to `Disabled` by default to avoid any issues like this, but apparently it's not enough. We need a fix to ensure that this option is only used in Custom Games, and not ladder. 

https://i.gyazo.com/5e1c9af739c7c7cc54e0d5f254bfca2c.png

Recommended fix by Strogo: `if ScenarioInfo.Options.AIReplacement == 'AIReplacementOff' or ScenarioInfo.TeamGame == false then`

PR #2573, opened 2018-08-20
Hotfix for AI Replacement in Ladder

JJ173:
Possible fix provided by: `Strogo`. This should hopefully negate the AI takeover in ladder. This will need testing before merging. Not sure how we can test it but that should take care of the issue.

Addresses issue: https://github.com/FAForever/fa/issues/2571

### DIFF
--- a/lua/aibrain.lua
+++ b/lua/aibrain.lua
@@ -3421,9 +3421,7 @@ AIBrain = Class(moho.aibrain_methods) {
 
     AbandonedByPlayer = function(self)
         if not IsGameOver() then
-            if ScenarioInfo.Options.AIReplacement == 'AIReplacementOff' or not ScenarioInfo.Options.AIReplacement then
-                self:OnDefeat()
-            else
+            if ScenarioInfo.Options.AIReplacement == 'AIReplacementOn' then
                 ForkThread(function()
                     local oldName = ArmyBrains[self:GetArmyIndex()].Nickname
 
@@ -3450,6 +3448,8 @@ AIBrain = Class(moho.aibrain_methods) {
                     self:InitializeSkirmishSystems()
                     self:OnCreateAI()
                 end)
+            else -- If ScenarioInfo.Options.AIReplacement return nil or any other value, make sure the ACU explodes.
+                self:OnDefeat()
             end
         end
     end,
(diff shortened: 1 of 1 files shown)


lua/aibrain.lua
```diff
@@ -3421,7 +3421,7 @@ AIBrain = Class(moho.aibrain_methods) {
 
     AbandonedByPlayer = function(self)
         if not IsGameOver() then
-            if ScenarioInfo.Options.AIReplacement == 'AIReplacementOff' then
+            if ScenarioInfo.Options.AIReplacement == 'AIReplacementOff' or not ScenarioInfo.Options.AIReplacement then
```
speed2CZ: Simplier approach would be to check if  it's enabled, else ignore it.
That way you have just a single condition.

Uveso: We should first validate the existence of `ScenarioInfo.Options.AIReplacement` (in case this is the first game start with this option and without visiting the lobby before) then check the value:
```LUA
if ScenarioInfo.Options.AIReplacement and ScenarioInfo.Options.AIReplacement == 'AIReplacementOn' then
    ForkThread(function()
        [...]
    end)
else
    self:OnDefeat()
end
```

speed2CZ: It doesnt really need that much validating
ScenarioInfo.Options will always exist, else you wouldnt even lauch the game
And then you can check if its enabled right away. If the option doesnt exist and its nil, it wont break, or not?

JJ173: The issue is that Ladder is returning a nil value for `ScenarioInfo.Options.AIReplacement` because it simply doesn't exist in that scenario (Ladder is looking elsewhere for options I suppose). Changing the condition around and making it check if the flag is `AIReplacementOn` before executing the AI Replacement will fix it. If it returns a nil value, it will still execute the else statement, which in this case would be `self:OnDefeat()`.

JJ173:
I've made the changes as requested, but I'm guessing the PR has messed up and I have no idea on how to rebase it. Fresh commit is here: https://github.com/FAForever/fa/pull/2573/commits/e328c93fc1a5121f9a64453fd093963049c9703a

PR #2574, opened 2018-08-20
Add lobby option for player chat color

Crotalus:
Make the gay pride colors optional 🏳️‍🌈

### DIFF
--- a/loc/US/strings_db.lua
+++ b/loc/US/strings_db.lua
@@ -4364,6 +4364,7 @@ lobui_0403="Lobby is windowed until launch"
 lobui_0404="Chat Font Size:"
 lobui_0405="Lobby Background"
 lobui_0411="Faction Font Color"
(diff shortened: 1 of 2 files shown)


shalkya:
image
Should we also add an option (or use the same option) to revert the bold username ?
Could we make color "on" by default, and then people will turn it off if they don't like

aeoncleanse:
Great PR @Crotalus. I agree with shalkya, please include the bold-ness in the toggle, and default it to "On". Then we'll merge it in :)

Rackover:
The color + boldness should be default on because most users benefit from it. 
Also may you fix the faction color being overriden by the chat rewrite aswell please ? When the player colored names are disabled the faction color should technically take over. (Or else, be sure to make the chat white, not grey). Otherwise yeah it's really cool thanks for handling this

shalkya:
another PR has been set up to improve this one.

Issue #2575, opened 2018-08-21
aeon tmd deflect nuke


shalkya:
https://www.twitch.tv/videos/300117917?t=02h14m38s

known issue from long time, seems there were no issue made on github

FAETHER:
Actually, Nuke collides with aeon antiprojectiles that it fires in the air. And nuke gets distracted by smd because its clasified as 'MISSILE' in the blueprint.

FAETHER:
![hack](https://user-images.githubusercontent.com/33268383/44618741-c8d0b480-a849-11e8-9ab6-95a84341d12e.PNG)

As you can see when I have removed all MISSILE keywords, the nuke is no longer dragged in.

FAETHER:
![read](https://user-images.githubusercontent.com/33268383/44618787-d470ab00-a84a-11e8-8aee-8064a26f0efc.PNG)

Here is the current situation, nuke gets dragged in.

![bug](https://user-images.githubusercontent.com/33268383/44618790-e7837b00-a84a-11e8-9ad4-c9302146ec30.PNG)

And then either disappears with no explosion or explodes in the air. In both situations the gamage is still dealt.

shalkya:
does the nuke still get stopped by smd if it lose the missile label ?

FAETHER:
Here are the possible points of interest: 
https://github.com/FAForever/fa/blob/3b73337e27608d094fba5ed20c772833c9e21d1f/lua/sim/defaultweapons.lua#L100-L127
https://github.com/FAForever/fa/blob/3b73337e27608d094fba5ed20c772833c9e21d1f/lua/sim/Projectile.lua#L449-L472
https://github.com/FAForever/fa/blob/develop/lua/defaultantiprojectile.lua#L20-L50

FAETHER:
Yep, collision still applies even without missile keyword. Though damage is dealt in most cases. I had it once not deal damage at all, but im not confident enough to claim that its real and not just my messing around with code.

Strogoo:
Forgot to mention issue number in my PR. I fixed it without removing any category(this can potentially cause some problems). As for collission, this is fixable by DoNotCollideList. Should I add it to pr or we let nuke to collide with these spheres?

PR #2576, opened 2018-08-22
Fix Aeon tmd deflects nuke (EQ)

Strogoo:
Issue #2575

### DIFF
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -452,17 +452,20 @@ Projectile = Class(moho.projectile_methods, Entity) {
         self.MyFlare = Flare {
             Owner = self,
             Radius = tbl.Radius or 5,
+            Category = tbl.Category or 'MISSILE',  -- We pass the category bp value along so that it actually has a function.
         }
         if tbl.Stack == true then -- Secondary flare hitboxes, one above, one below (Aeon TMD)
             self.MyUpperFlare = Flare {
                 Owner = self,
                 Radius = tbl.Radius,
                 OffsetMult = tbl.OffsetMult,
+                Category = tbl.Category or 'MISSILE',
             }
             self.MyLowerFlare = Flare {
                 Owner = self,
                 Radius = tbl.Radius,
                 OffsetMult = -tbl.OffsetMult,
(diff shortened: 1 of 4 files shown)


HUSSAR-mtrela:
@shalkya @Strogoo this PR looks good code-wise.

Uveso:
Do i get this right, because we have a problem with the Aeon TMD collision detection, we are changing the flightpath from nukes ???

Did you test this with the UEF commander (tactical nuke) on short range ?
(It can't change flightangle on short distances)
nukeshort

(I know we don't care about mods, but this also won't work with moded missiles like
the Black Ops TECH3 tactical missile defence.)

In my opinion we should change the Aeon antimissile flare instead of (all) nuke flightpath's

Strogoo:
1. We aren't changing the flightpath, only turn rate at final phase.
nuke


2. This change doesn't affect tactical nuke, it has its own MovementThread.
movement

3. I don't know how t3 tmd works in BlackOps (will test it), but I doubt that this fix can break it.

Uveso:
Sorry i wrote to much, so just my first question;

"Why" we are fixing nukes when the Aeon tmd launcher is the problem ?"

Can't we change the Aeon tmd flares ?

Strogoo:
Well yes, you are right. Honestly, I just did it in easiest way (for me ;) Because defaultantiprojectile.lua is still not hooked, so I have to clean it up or make a new class based on it and then change tmd projectile blueprint etc etc. And here you just add some lines in one file and it works =)

Exotic just told me that he fixed it in EQ, so I will probably copy it.

shalkya:
need to code review the EQ change also then ;)

Strogoo:
https://github.com/FAForever/equilibrium/commit/36d7c31ce3788efaf63577328567ba3bfac05a1a#diff-4c50057a3c51be05ebd08578faafd4b2

https://github.com/FAForever/equilibrium/commit/ea8f7ec665bfc5caa46aa2ba44c4f78699f65618#diff-013dedf3e78323e4466cbe9353ba1a59

Uveso:
Tested the new changes ingame, and it's working flawless.

@Strogoo thank's a lot for making this new PR!!!

PR #2583, opened 2018-09-13
 added CQUEMOV for upgradable building

shalkya:
so that you can queue up upgrade while the unit is being built.

### DIFF
--- a/units/UAB1103/UAB1103_unit.bp
+++ b/units/UAB1103/UAB1103_unit.bp
@@ -48,6 +48,7 @@ UnitBlueprint {
         'RECLAIMABLE',
         'SHOWQUEUE',
         'SORTECONOMY',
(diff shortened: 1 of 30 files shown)


shalkya:
introduce a bug, some BP will keep "contructing" the building while it is finished and upgrading. Spending ressource on nothing (it doesn't actually assist the upgrade)
image
will let people that actually code take a look at that.

speed2CZ:
That doesnt look like a new bug. I saw this happen even with the vanilla FA. Easiest way to trigger this is to have a lot of build power and then qeue up all 4 upgrades on the shiels. Then some of the engineers will have the full construction bar, spending resources but not speeding up the upgrade anymore.

Strogoo:
Same bug in current patch:
https://youtu.be/kqCcd_H0wHI

Just open new issue and merge this awesome PR :) (and don't use 'repair' order with cybran shields :D)

PR #2588, opened 2018-09-23
Fix uef T3 MAA missing

Petricpwnz:


PhilipJFryFAF:
@Petricpwnz please make PRs against deploy/fafbeta in the future

PR #2593, opened 2018-09-25
Added toggle option to disable chat colors, fixed factional font colo…

Rackover:
…r, fixed observer color


shalkya:
image
can you pls remove the bold, and make the toggle on by default.

shalkya:
image
there is an issue with bold still ?

Rackover:
I thought you wanted boldness to disable only when the factional font color was enabled...

shalkya:
never said anything like that xD
i just said you make a toggle : 
if on => we get your full addition (color + bold)
if off => we let people get the lobby as before.

shalkya:
image
doesn't seems to work

shalkya:
tested and behave as expected.

PR #2544, opened 2018-08-02
Wreckage spawn like after death animation

c0okieZ:
This fix about spawn wreckage on game start
If unit have death animation and spawn on land (or in water on depth<5), we use this death animation and skip it to unit final condition, then create wreackage in this condition as after death in game (by default, function just spawn unit, create wreckage, destroy unit).
If unit is "naval" on water or "air", we disable deathAnimation (coz some naval units haven`t animation but have link for animation in blueprint, this glitched and hold it on surface) and use random rotation for simulate sinking/falling.
Else we just create wreckage and destroy unit like in default function

default
leftwreckage
this fix
rightrwreckage

[video](https://www.youtube.com/watch?v=3x-ChnKcpHQ)

PR #2596, opened 2018-09-27
Naval on shallow water doesn't need to rotate

c0okieZ:
Small fix for PR #2544 . Now naval on shallow water doesn't rotate


shalkya:
review and tested.

PR #2539, opened 2018-07-27
 [Issue#1260] Ythotha fix destroy

zkov96:
[Issue#1260](https://github.com/FAForever/fa/issues/1260) Ythotha fix destroy
Refactored code in fa/units/XSL0401/XSL0401_Script.lua
Changed:
1. Move lightning spawn from OnDestroy to OnReclaimed
2. Spawn lightning in DeathThread as original
3. Add defuse lightning from owner by reclaim
4. Call Unit:Destroy() not spawn lightning [as wanila]

This commits add lightning after reclaim Ythotha unit by any army (not owner). But it work more correctly then this [commit 1449dc5a3450b4cb497d20439969f0736ff6c784](https://github.com/FAForever/fa/commit/1449dc5a3450b4cb497d20439969f0736ff6c784) without using duct tape [IsBeingTransferred](https://github.com/FAForever/fa/commit/810ffcc30af98701374bd2588fbdc61d911a605b#diff-978e442e90e99e5cab1d4e27bd351351). Also not spawn lightning on Unit:Destroy() how it should be

PR #2608, opened 2018-10-02
make ally/owner triggers the storm by reclaiming

shalkya:
fix the balance change that was not approved from https://github.com/FAForever/fa/pull/2539


shalkya:
tested and reviewed.

PR #2612, opened 2018-10-08
Changing tml missle cost

TurinturambarFAF:
changes the cost of all tml missles from 180 mass to 250 mass.


PhilipJFryFAF:
why add the *.dds files or script files?
every file in the repo has to be distributed by the updater and stored locally in the gamedata folder so adding files that we don't change doesn't seem useful to me
Please remove all added files which you don't plan on changing @TurinturambarFAF

PR #2615, opened 2018-10-12
adding required bps

TurinturambarFAF:
adds the bps for all factions nuke missles.


PhilipJFryFAF:
already done in #2616 (one commit to add the BPs and one to change the bp/bt)

PR #2621, opened 2018-10-20
Use general on/off key for AI replacement option

### LINKED ISSUE #2610: Replace AIReplacementOn/AIReplacementOff with On/Off
Please change this;
1. It's inconsistent with all the other options, thus it causes confusion and introduces chaos
2. It does not hold any more (useful) information than simply On/Off

We'll make the server replace all incoming `AIReplacementOn` with `On` (and same for off) for now so you can change this w/o breaking compatibility. But we'd prefer to go back to a clean implementation soon :-)

speed2CZ:
Fixes #2610

Like this @micheljung ?

To the test team: I didnt test it, just replaced all code that the git found, so it should not break anything, but you know what to do :) test :))

### DIFF
--- a/lua/aibrain.lua
+++ b/lua/aibrain.lua
@@ -3431,7 +3431,7 @@ AIBrain = Class(moho.aibrain_methods) {
 
     AbandonedByPlayer = function(self)
         if not IsGameOver() then
-            if ScenarioInfo.Options.AIReplacement == 'AIReplacementOn' then
+            if ScenarioInfo.Options.AIReplacement == 'On' then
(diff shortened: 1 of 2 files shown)


PhilipJFryFAF:
@Rackover i assume this does mess with https://github.com/FAForever/server/pull/357
so i suggest we also change it on the server side to avoid having ai replacement rated

Rackover:
@PhilipJFryFAF 
https://github.com/FAForever/server/pull/357/files#diff-c83dea448bab57db88604d9e84c094ebR276
:D

PR #2627, opened 2018-10-21
update mod blacklist

PhilipJFryFAF:
adds the Better Allies and the No Friendly Fire Collision V2 mods to the mod blacklist since projectile collision with friendly units is removed anyway (since 3699)


shalkya:
tested and reviewed

PR #2633, opened 2018-10-25
Novax hotfix

Strogoo:
Moving code from terranweapons.lua to xea0002_script.lua


speed2CZ:
What exactly does it fix?

Strogoo:
novax

speed2CZ:
That's why making an issue first is a good thing :)

PhilipJFryFAF:
tested locally -> works
not sure how to test if it breaks mods since i don't know any mods that touch the script for the novax sat

PhilipJFryFAF:
fixed by #2641

PR #2636, opened 2018-10-25
scathis rebalance

shalkya:
some random change to work on.

medium range scathis design against T3 army/unshielded stuff.


aeoncleanse:
Top-down redesign - Mobile T4 anti-T3 artillery. Special features: High RoF, powerful EMP, moves faster underwater

PR #2641, opened 2018-10-28
Fix Novax crashing other mods

Exotic-Retard:
changing the script like that lead to mods that hook
defaultprojectileweapon to change something in there, have ambiguous
class definitions which is very difficult to fix if you dont know what
you are doing, and it locked out a bunch of modding opportunities. to
prevent this, there is a script free way of doing the same thing, with a
blueprint value.

what this really does is highlight the need to refactor the weapon
class. this only works because the Main function inside idle state waits
for the recoil manipulators, which lasts until the beam is destroyed via
the thread. its a complete mess really. but thats how all beam units
work.

there should be a proper blueprint value to end beam on retarget or not,
and the recoil should not stall the beam end sequence.

### DIFF
--- a/lua/terranweapons.lua
+++ b/lua/terranweapons.lua
@@ -262,12 +262,4 @@ TOrbitalDeathLaserBeamWeapon = Class(DefaultBeamWeapon) {
         end
         DefaultBeamWeapon.PlayFxWeaponUnpackSequence(self)
     end,
-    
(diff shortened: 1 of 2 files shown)


PhilipJFryFAF:
tested locally - works fine
appears to be the least mod breaking way of fixing the sat since no scripts are molested

PR #2644, opened 2018-10-29
Weapon priorities v2 + improvements from Exotic

Strogoo:
New branch - new PR =)

Big thanks to Exotic again.

All removed stuff (customizable presets/hotkeys, snipe mod from EQ etc) will be added after release via UI mod.

WeaponPriorities.lua is taken from here
https://github.com/FAForever/fa/commit/51d21cc20f97a6438056c885e520cac44c0e7d80
so no need to review this I think.

The rest of the code (mostly UI) is from my first PR.


Strogoo:
Fixed some bugs in sim code.

Also removed Exotic's code which add "ALLUNITS" to any preset because it kinda breaks presets with 'exlusive' flag. And you can't make special presets for asfs/snipers which contains only 1 or 2 category so unit can ignore unimportant targets and doesn't waste dps.

Non 'exclusive' presets use default priority table after user categories and this table already contains ALLUNITS in the end.

shalkya:
is that color on purpose when switching to civ ?
image

HUSSAR-mtrela:
@Strogoo let me know when this PR is ready for review. I might have some performance or UI improvements.

Strogoo:
Ok, gonna fix it today/tomorrow. also will add "sniper" mod from EQ

Strogoo:
"Shields" preset was broken and that was a reason of all UI bugs it seems

Strogoo:
"Units" (MOBILE - COMMAND - EXPERIMENTAL) will be more useful I think.

Strogoo:
Ehh, can't just copy-paste this "snipe" feature from EQ, doesn't work with presets. Right click on attack order -> snipe mode -> pick some preset from list -> Attack button doesn't return to default. Shouldn't be too hard to fix, but I'm sooooo tired of this PR =) Exotic will be upset if I won't add it ;) Any help is welcome.

How it works: http://forums.faforever.com/viewtopic.php?f=67&t=10004&p=167800#p167800

Code: https://github.com/FAForever/fa/commit/51d21cc20f97a6438056c885e520cac44c0e7d80#diff-e9ca75764cf18ddf719eb1f61d6587c4R189

Lines 189-300 and 1196 + toggle_red.dds

Mb I will add it by myself tomorrow, or the day after tommorow, or next week... :D Overall this PR is ready. @HUSSAR-mtrela

Nojoke185:
one other comment...
If for example you set a factory (such as an air factory) to power priority, Units coming out of the factory still have the default priority. If you click back on the factory it is also back to default.

Not sure if possible to code but could the factory retain its set priority and any units produced inherit the priority to save having to reset these?

Strogoo:
```Would it be worth looking into some specific priority variations for air/naval type units? (unless this is planned to be done as part of the UI Mod later?)```

Yeah, I had same idea. And yes, leave it for UI mod ;)

```If you click back on the factory it is also back to default.```

That's because factory doesn't have any weapon. "Target priority: name" means that sim code was triggered but all selected units without any weapon stays in default mode.

```If for example you set a factory (such as an air factory) to power priority, Units coming out of the factory still have the default priority.```

Yes, it's possible (sadly, not via UI mod). This requires some changes in sim code, I'll see what I can do. 

*for reviewers/repo admins: it doesn't mean that this PR shouldn't be merged before I add this feature (it can be added in other PR), so we are still in "Ready" state =))

**Comment by PhilipJFryFAF** [comment]
tested the PR locally
works as far as i can tell
@HUSSAR-mtrela would be nice if you can read through the code and give suggestions on how to improve it wherever you can

File: lua/WeaponPriorities.lua [hunk 1]
```diff
@@ -0,0 +1,160 @@
...
+            end
+            
+            for weaponNum, tbl in priorities do
+                if not finalPriorities[id] then finalPriorities[id] = {} end 
+                if not finalPriorities[id][weaponNum] then finalPriorities[id][weaponNum] = {} end
+                
+                if tbl[1] then
+                    local prioTbl = tbl
+                    
+                    for line, categories in prioTbl do
```
HUSSAR-mtrela: you can remove "`if tbl[1] then`" statement and "local prioTbl = tbl" assignment above by changing:
`for line, categories in prioTbl do`
to:
`for line, categories in tbl or {} do`

lua/keymap/keyactions.lua
```diff
@@ -429,6 +429,34 @@ keyActions = {
...
+    ['shift_Power'] = {action = 'UI_Lua import("/lua/keymap/misckeyactions.lua").SetWeaponPrioritiesHotkey("Power")',
+        category = 'Weapon priorities', order = 76,},
+    ['Shields'] = {action = 'UI_Lua import("/lua/keymap/misckeyactions.lua").SetWeaponPrioritiesHotkey("Shields")',
+        category = 'Weapon priorities', order = 77,},
+    ['shift_Shields'] = {action = 'UI_Lua import("/lua/keymap/misckeyactions.lua").SetWeaponPrioritiesHotkey("Shields")',
+        category = 'Weapon priorities', order = 78,},
+    ['EXP'] = {action = 'UI_Lua import("/lua/keymap/misckeyactions.lua").SetWeaponPrioritiesHotkey("EXP")',
+        category = 'Weapon priorities', order = 79,},
+    ['shift_EXP'] = {action = 'UI_Lua import("/lua/keymap/misckeyactions.lua").SetWeaponPrioritiesHotkey("EXP")',
+        category = 'Weapon priorities', order = 80,},
```
HUSSAR-mtrela: all these new key actions should have "target_" prefix in string keys because we use prefixes to identify similar key actions (e.g. see key actions  with "select_all_" prefix). Also these prefixes will ensure that we have unique string keys in this table. In other words:
please change:
```
['Default'] = ...
...
['EXP'] = ...
```
to
```
['target_Default'] = ...
...
['target_EXP'] = ...
```

lua/keymap/misckeyactions.lua
```diff
@@ -374,3 +374,36 @@ function CreateTemplateFactory()
...
+
+function SetWeaponPriorities(prioritiesString, name, exclusive)
+    local priotable
+    if type(prioritiesString) == 'string' then
+        priotable = prioritiesString
+    end
+    local units = GetSelectedUnits()
+    if units then
+        local unitIds = {}
+
```
HUSSAR-mtrela: you can remove "`if units then`" statement above by changing:
`for _, unit in units do`
to
`for _, unit in units or {} do`

lua/ui/game/orders.lua
```diff
@@ -551,12 +552,145 @@ local function CreateBorder(parent)
...
+            LayoutHelpers.AtLeftTopIn(UIUtil.CreateText(parent, name, 14, UIUtil.bodyFont), btn, 10, 0)
+            
+        else -- empty button
+            btn:DisableHitTest()
+        end   
+        
+        return btn
+    end
+    
+
```
HUSSAR-mtrela: please replace "while i < 8 do" with "for i, name in PrioritySettings.buttonLayout or {} do" and then you don't need these lines:
`local i = 1` 
`i = i + 1`
`local name = PrioritySettings.buttonLayout[i]`

lua/ui/game/orders.lua
```diff
@@ -551,12 +552,145 @@ local function CreateBorder(parent)
...
+    priorityTables = {
+        ACU = "{categories.COMMAND}",
+        Power = "{categories.ENERGYPRODUCTION * categories.STRUCTURE}",
+        PD = "{categories.DEFENSE * categories.DIRECTFIRE * categories.STRUCTURE}",
+        Units = "{categories.MOBILE - categories.COMMAND - categories.EXPERIMENTAL}",
+        Shields = "{categories.SHIELD}",
+        EXP = "{categories.EXPERIMENTAL}",
+    },
+    exclusive = {ACU = false, Power = false, PD = false, Units = false, Shields = false, EXP = false}, 
+    buttonLayout = {[1] = "Default", [2] = "ACU", [3] = "Units", [4] = "PD", [5] = "Power", [6] = "Shields", [7] = "EXP"},
```
HUSSAR-mtrela: the "buttonLayout" table can be simplified to:
`buttonLayout = { "Default", "ACU", "Units", "PD", "Power", "Shields", "EXP" },`

lua/keymap/misckeyactions.lua
```diff
@@ -374,3 +374,36 @@ function CreateTemplateFactory()
...
+function SetWeaponPriorities(prioritiesString, name, exclusive)
+    local priotable
+    if type(prioritiesString) == 'string' then
+        priotable = prioritiesString
+    end
+    local units = GetSelectedUnits()
+    if units then
+        local unitIds = {}
+    
+        for _, unit in units do
```
HUSSAR-mtrela: you can remove "`if units then`" statement above by changing:
`for _, unit in units do`
to
`for _, unit in units or {} do`

lua/ui/game/orders.lua
```diff
@@ -551,12 +552,145 @@ local function CreateBorder(parent)
...
+            
+        else -- empty button
+            btn:DisableHitTest()
+        end   
+        
+        return btn
+    end
+    
+  
+    while i < 8 do
```
HUSSAR-mtrela: please replace "while i < 8 do" with "for i, name in PrioritySettings.buttonLayout or {} do" and then you don't need these lines:
`local i = 1` 
`i = i + 1`
`local name = PrioritySettings.buttonLayout[i]`

HUSSAR-mtrela:
@PhilipJFryFAF please test if this PR still works with latest corrections and we are good to go.

Strogoo:
screenshot_2

Nojoke185:
Further suggestion:

Units preset :  can you specifically exclude engineers from this?
Seems current filter is categories.MOBILE - categories.COMMAND - categories.EXPERIMENTAL
(anything mobile but for command and experimental)
As we have new engineer preset we can use that to focus on engies. Units preset should focus fire on armed units

Strogoo:
Yep, good idea.

Strogoo:
Improved sim code a bit. 

Need some testing with big armies (20 pds + 20 tanks + 20 destros etc. at the same time)

Strogoo:
nvm, there will be UI changes too, so better test all these stuff together

Strogoo:
Ok, I found one more bug in sim code and rewrote it again :)

lua/WeaponPriorities.lua
```diff
@@ -67,33 +67,47 @@ function SetWeaponPriorities(data)
             table.insert(units, unit)
         end
     end   
+    
+    local preparedPrioTables = {}
 
     for _, unit in units do
         local blueprintId = unit:GetBlueprint().BlueprintId
         local weaponCount = unit:GetWeaponCount()
```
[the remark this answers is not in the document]
Strogoo: I'm sure it does. But I don't want to chnage smth in sim again tbh (every time I do it some bug appears =). Code is pretty fast now, changing preset for 1000 aeon gunships (3 weapons on board) takes 0.013 sec and 0.005 sec for 1000 asfs.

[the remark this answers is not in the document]
Strogoo: Yep, it counts dummy and death weapons. Some dummy weapons have priority tables in their blueprints btw, so we have to count them too. As for death weapons, current code use SetTargetingPriorities() with empty table for them which is not nice :(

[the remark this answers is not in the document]
HUSSAR-mtrela: thanks for looking into this.

PR #2649, opened 2018-11-01
Buffing bp of t2 and t3 engies

Tagada14:
Buffing bp of t2 and t3 engies to make them more cost efficient bp.

T2 engies

BuildRate = 12.5, -> BuildRate = 14.5,
Mass efficiency 10.4 mass for 1 bp -> 8.95 mass for 1 bp
around 10% increase
10% more efficient then t1 engie

T3 engies

BuildRate = 30, -> BuildRate = 38.5,
Mass efficiency 10.4 mass for 1 bp -> 8.1 mass for 1 bp
around 19.52% increase
10% more efficient then t2 engie


Apofenas:
...T2 build range 5->7.5
...T3 build range 5->10

JaggedAppliance:
We're going with this for now instead: https://github.com/FAForever/fa/pull/2697

HUSSAR-mtrela:
@JaggedAppliance can you explain why this PR was not used and instead you just changes speed of T2/T3 engineers to match T1 engineers ? IMO, this PR makes very good balance improvements and T2/T3 engineers would be worth building instead of spamming T1 engineers which causing slow down when they are a lot of engineers assisting a factory. At least, these changes should be made:
- T2 build range 5->7.5
- T3 build range 5->10

JaggedAppliance:
We are making them nicer to use rather than straight buffing them to be more efficient in terms of buildpower. Higher tech engineers already have big advantages of course, being able to build higher tech buildings, especialy defenses. 

Buffing the build range of higher tech engineers would encourage using them in conjunction with many t1 engineers as you can leave the t3 engy in one place and use t1 engies as the main buildpower.

HUSSAR-mtrela:
you will not be able to use T3 engineers in conjunction with many T1 engineers because T1 engineers will have smaller build range and power than T3 engineers. You said your self in a few casts that T1 engineers are often not doing anything because they stop building when a lot of engineers bump each other around a factory. At some point of a game, players stop building T1 airplanes (small DPS/range) and switch to T3 ASF (bigger DPS/range) and this should also be happening with engineers transition T1 -> T2 -> T3

PR #2651, opened 2018-11-04
ASF rebalance

shalkya:
- to reduce the need of ASF spam
- to make droping a bit more prevalente even if T3 air is in the game
- to make air exp less easy to snipe
- some speed tweaking might be needed for T3 scout/strat, although ASF still have 5 more speed than strat.

### DIFF
--- a/units/UAA0302/UAA0302_unit.bp
+++ b/units/UAA0302/UAA0302_unit.bp
@@ -15,7 +15,7 @@ UnitBlueprint {
         KTurn = 1,
         KTurnDamping = 1.5,
         LiftFactor = 7,
-        MaxAirspeed = 30,
+        MaxAirspeed = 27,
         MinAirspeed = 25,
         StartTurnDistance = 5,
         TightTurnMultiplier = 1.1,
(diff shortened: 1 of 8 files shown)


aeoncleanse:
This won't work fully to reduce numbers. The lower damage tells people they need more to do the same job, and the higher build time just means they'll spam more buildpower. It will take longer to achieve the swarms, but swarms remain the objective. To reduce numbers, you have to increase cost and effectiveness such that an air wing of equal strength is constructed of fewer actual planes.

shalkya:
i don't want to make the ASF arrive later in the game though.
i've been doing all this change especially in order to adjust ASFs to interceptor's stat.
i will maybe try to make people host a seton or 2 with these settings, and gather their pov afterward

JaggedAppliance:
This is something to consider for the next patch.

rjy7wb:
why not just lower flight times, then make them useless without fuel.

Techmind:
Also 2-nd the problem is that, asf as counter to air exps is too good, awasa/czar/donut need better air ehp, either raise their base hp x2 and lower their air dps / 2  or do some sort of 'flare', so 25-50% of attacks from asf can miss the experimental. Land based experimentals can only be countered by land units to some extent because of their speed and all exps having pretty great dsp & big ranged on weapons. Air is not the case. ASF is faster, takes way less damage (overkill from anti-air on experimentals) per mass spend.

PR #2655, opened 2018-11-11
Add scrolling text with announcement into the lobby

speed2CZ:
First try for adding the scrolling text with announcements into the in-game lobby.

Good for informing people about future tournaments or streams of some tournaments.

It looks like this: https://www.youtube.com/watch?v=sqQjGghdw9A
- [ ] Still need to figure out the way of sending this info into the game.
Which could be a text and a link (that way clicking on the text will open the web page)
- [ ] And slight problem that there's not enough space for the text with lower screen resolutions (height 768px)

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -2721,6 +2721,45 @@ function CreateUI(maxPlayers)
         end
     end
 
+    -- Scrolling text with live updates from he server.
+    -- TODO: Make it work with server (just the small things).
+    -- Uses command line argumets /newsText and /newsURL
+    local newsText = GetCommandLineArg('/newsText', 1)
+    if newsText then
+        -- Black background behind the text
+        local scrollingBG = Bitmap(GUI)
+        scrollingBG:SetSolidColor('ff000000')
+        scrollingBG.Left:Set(function() return GUI.Left() end)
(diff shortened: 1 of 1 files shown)


Strogoo:
> Still need to figure out the way of sending this info into the game.
> Which could be a text and a link (that way clicking on the text will open the web page)

cmd args? Idk how it works with long strings, should be fine I think ;)

https://github.com/FAForever/fa/blob/deploy/faf/lua/ui/lobby/lobby.lua#L125
&
https://github.com/FAForever/client/blob/stable/0.18/src/client/_clientwindow.py#L1449

```message["fafNews"] from server```

speed2CZ:
Ok, so the new version uses two command line arguments /newsText and /newsURL

I'm not sure about the lenght, my poor testing skill allowed me to try only some short ones.
So if someone can try to make it run with some longer text please.

axel1200:
If you can open links you could just request the news from API. That would be easy as .... for me then

axel1200:
If you can open links you could just request the news from API. That would be easy as .... for me then

axel1200:
Like I make a db schema add it to java-api and client can hand you over the news url where you request news from API... And we can set them in modor

axel1200:
I saw you got a json libary in there as well ;)

PR #2659, opened 2018-11-25
useful torpedo on amphibious unit

shalkya:
Making the torpedo worth not being a complete joke anymore.
ratio dps/mass : 0.15 for wagner; 0.1 for brick/othuums
comparison : 0.4 T1 subs, ~0.18 for destro (apart uef) but they also have good anti-torp, range and vision advantage; 0.4 T2 subs + they have antitorp, range/vision (possibly stealth) advantage

### DIFF
--- a/units/URL0203/URL0203_unit.bp
+++ b/units/URL0203/URL0203_unit.bp
@@ -362,7 +362,7 @@ UnitBlueprint {
             },
             BallisticArc = 'RULEUBA_None',
             CollideFriendly = false,
-            Damage = 22,
+            Damage = 15,
             DamageType = 'Normal',
             DisplayName = 'Meson Torpedo',
             FireTargetLayerCapsTable = {
--- a/units/URL0203/URL0203_unit.bp
+++ b/units/URL0203/URL0203_unit.bp
@@ -362,7 +362,7 @@ UnitBlueprint {
             },
             BallisticArc = 'RULEUBA_None',
             CollideFriendly = false,
-            Damage = 6,
+            Damage = 22,
             DamageType = 'Normal',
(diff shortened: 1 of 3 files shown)


Apofenas:
The problem with weak torpedoes for land units is not about just low DPS.

As I remember from test games with Softly a few years ago, Wagner with 10 DPS, 22 range and no Sonar/Water vision was strong enough to take on t1 subs on small maps with islands. You make it 11 DPS with old 32 range which should be well enough to take on...t1 subs?

When you face hover, you build frigate which is one of the most efficient units in game, if you need to counter subs, you build one of the most useless naval units. If you try to balance Wagner around t1 subs, it would still stay useless against anything stronger t1 subs.

Another problem about Wagner is its utility on land. Hovertanks are quite weaker than their land analogues. Wagner isn't as weak.

My suggestion on Wagner:
-HP: 1450->1200
-WaterVision:20->15

-Torp DPS: 3->20    --only one torp which can be countered by torp def
-MesonRocket DPS:50->20    --single rocket which matches torpedo fire circle(since it is same weapon)
-Torp/Rocket range: 32->20

-Bolter DPS: 50->80   --more consistent ground weapon
-Bolter range: 22->23

Same goes to Brick/Othuum. Reduce amount of torps to 1, reduce torp range and water vision, but increace their DPS.

shalkya:
i played a game lately, where wagner were raped by T1 subs, even if there was way more mass invested into wagner than subs. Doesn't mean i want to make wagner as strong as sub, but a buff won't mean that wagner will win against sub.
Ofc this need testing.

JaggedAppliance:
Needs rebasing

PR #2669, opened 2018-12-02
Fix Sera T2 PD Aim

PhilipJFryFAF:
changes the ballistic arc to be the same as on the cybran t2 pd
change was made by evildrew in his balance mod


shalkya:
no improvement from my testing

PhilipJFryFAF:
credits go to evildrew for changing this in his balance mod and Dro for making me aware of it

PR #2671, opened 2018-12-19
Increasing t2 arty minimum range

Farmsletje:
Making t2 arties a little less good and making sure they are used for their original purpose: Long range shelling. Min range = t2 pd max range
All t2 arties: Min range: 5 > 50


JaggedAppliance:
You have two wrong commits in this

PR #2673, opened 2018-12-23
Wrecks on game start

Strogoo:
t4 and naval wrecks spawn using this #2544

All other wrecks use old code.

Need testing on maps with exp and naval wrecks. I don't know any map like that and too lazy to make my own :D


PhilipJFryFAF:
closed cause it's just there to make the reclaim on setons look like it used to

PR #2678, opened 2018-12-30
Remove hardcoding on nuke weapon damage types

Exotic-Retard:
this allows nuke weapons to have damage types other than nukes. for mod
support reasons and for pretty code reasons


Apofenas:
Doesn't UEF billy ACU also use InnerRing and OuterRing damage values? Shouldn't it be changed too?

Exotic-Retard:
changed the billy, but as a note the acu death explosions already have a special damage type so they get to keep that

PR #2679, opened 2019-01-02
Change Target Bone

Uveso:
Changing salem's targetbone from 'Muzzle_Front' to 'Turret_Mount_Front'
http://forums.faforever.com/viewtopic.php?f=45&t=17066


PhilipJFryFAF:
tested and works

PR #2688, opened 2019-01-07
Scathis mobility increase

TurinturambarFAF:
increases scathis mobility to make it possible to dodge nukes


speed2CZ:
I think you want to have a single branch for all the scathis changes, each change in single commit, resulting in a single PR with multiple commits.

TurinturambarFAF:
what is the best way to fix it if it needs fixing?

PR #2701, opened 2019-01-11
reduces ED5 assist cost

TurinturambarFAF:
increases ED5 bt to other T3 shield level to fix assist cost


shalkya:
duplicate of #2703

PR #2702, opened 2019-01-13
Fix free assist for shields

Strogoo:
If you have any ideas how to improve performance for this hell or how to fix it in a different way, post it here.

*Only for static shields. We don't want to run it for mobile shields, right? :D Better hard nerf assist for them via bplueprint and RegenAssistMult.

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -76,6 +76,13 @@ Shield = Class(moho.shield_methods, Entity) {
         self.OffHealth = -1
 
         self.PassOverkillDamage = spec.PassOverkillDamage
+        
+        local ownerCategories = self.Owner:GetBlueprint().CategoriesHash
+        if ownerCategories.STRUCTURE then
+            self.StaticShield = true
+        elseif ownerCategories.COMMAND then
+            self.CommandShield = true
+        end    
 
         ChangeState(self, self.OnState)
     end,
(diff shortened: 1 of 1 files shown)


speed2CZ:
Will this mess with mods that add shield buildings?

Strogoo:
Well, without StaticShield in blueprint new shields will be created using main shield class (like before) and this means no fix for them. 

It can be fixed without adding new class, with more "if owner.catHash.Structure bla bla bla" in main shield class but idk :) Also no optimisations for overcharge then

Strogoo:
Actually, since I removed OC code from main class, there will be no damage reduction for new shields... Ok, better do it with more "if" I guess :D

Uveso:
You counting assisters with :GetGuards().
What about guards like Tanks who guard (but not assist) the shield ? Do they have a buildrate ?
What about Engineeers that are switched to pause ? They are guarding but not assisting while paused.

`Assist efficiency never drops, no matter what mass income you have`
Is this true ?
I have an mod with experimental shields, and assisting them mostly with 10 SubCommanders.
They can hold the shielddome easily under heavy fire.
But if i stall mass then they can't rep the shielddome anymore.
So the efficiency is almost 0 on low mass. ?!?

Sadly my girlfriend arrives in 30 minutes so i can't check this before wednesday :/

Strogoo:
All these cases are covered by ```if unit:GetFocusUnit() == self.Owner then```

Non-engineer units can't have shield as FocusUnit. Same for engineers on pause (FocusUnit = nil)

> Is  this true ?`


yep, in current faf version all shields work like that, sadly :(

>But if i stall mass then they can't rep the shielddome anymore.`

Hmm, intereseting. What mod is that?

Strogoo:
Ok, I've tested your mod. Same bug there with all shields, no easy fix :(

Uveso:
Well, then i have to fix it anyway :)

Tank's a lot for the info Strogoo !!!

Strogoo:
@Uveso Ok, paused engies still have shield as focus unit (idk how I tested it earlier) ;) Now it's fixed

Also fixed some minor bugs and improved performance a bit.

zlo-faf:
There was a small weirdness, gotta remember that  efficiency does drop significately (don't remember how much but more than 50% or something... ) when mass income drops from 1 to 0. I don't know why. I wanted to test fractional mass income, like 0.1, 0.5... but i forgot about it.

zlo-faf:
Ok i checked. with literally 0 mass income shield assist stop working, with 0.25 mass income it gets to full efficiency again, i was hoping to find something inbetween, but no.

shalkya:
tested.

PR #2710, opened 2019-01-20
Enable transport rates to be set vs blueprint

Exotic-Retard:
pretty much what it says on the tin. This allows blueprints to have
TransportAnimationSpeed and TransportDropAnimationSpeed which can
configure the rates at which the animations play. If they are missing,
the code assumes default behavior, so its backwards compatible.

this is for mod support reasons.


Exotic-Retard:
thats an unrelated issue, this allows extra mod support and doesnt affect the game by default. it will let you tune leg folding animations for all units, so it can improve that, but wont fix the sera transport thing.

PR #2711, opened 2019-01-20
Nuke missile kills non-experimental air units without explosion

Strogoo:
Not approved yet
!

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1779,6 +1779,52 @@ AirUnit = Class(MobileUnit) {
             MobileUnit.OnKilled(self, instigator, type, overkillRatio)
         end
     end,
+    
+    
+    -- Well, unit's OnCollisionCheck is called before projectile's OnCollisionCheck. That's why this check is here and not in NukeProjectile
+    -- Doing it for every projectile that hits air unit is so dumb, but what's other option? :(
+    OnCollisionCheck = function(self, other, firingWeapon)
+        -- Nuke doesn't explode on collide with air units
+        if other.Nuke and not self:GetBlueprint().CategoriesHash.EXPERIMENTAL then
+            if not IsAlly(self:GetArmy(), other:GetArmy()) then
+                self:Kill()
+                return false
+            end    
+        end
+        
+        -------- default part from unit.lua---------
+        --------------------------------------------
+        
+        if self.DisallowCollisions then
+            return false
+        end
+
+        if EntityCategoryContains(categories.PROJECTILE, other) then
+            if IsAlly(self:GetArmy(), other:GetArmy()) then
+                return other.CollideFriendly
+            end
+        end
+
+        -- Check for specific non-collisions
+        local bp = other:GetBlueprint()
+        if bp.DoNotCollideList then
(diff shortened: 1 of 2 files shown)


5cript:
So you can't intercept nukes with aircraft anymore (intentional or not)?
Why? I consider this is a feature. Is there any discussion to read up on?

Strogoo:
Yea, sry for bad description. If missile collide with enemy aircraft, it doesn't blow up or vanish, it kills this aircraft and continue its flight.

Other option is to turn off collisions so planes stay alive and nuke just go through. But there are a lot of people that always get triggered when we remove collisions for some units/projectiles, so this is more elegant fix for them :)

Personally I don't care, and happe with any fix for this annoying mechanics.

KionX:
Why is this done in AirUnit and not in TIFMissileNukeCDR ?
Why Copy-Paste ?
Why self.Nuke = true. In a shield.lua is written like this:
-- Allow strategic nuke missile to penetrate shields
if EntityCategoryContains(categories.STRATEGIC, other) and
    EntityCategoryContains(categories.MISSILE, other) then
    return false
end

lua/defaultunits.lua
```diff
@@ -1779,6 +1779,52 @@ AirUnit = Class(MobileUnit) {
...
         end
     end,
+    
+    
+    -- Well, unit's OnCollisionCheck is called before projectile's OnCollisionCheck. That's why this check is here and not in NukeProjectile
+    -- Doing it for every projectile that hits air unit is so dumb, but what's other option? :(
+    OnCollisionCheck = function(self, other, firingWeapon)
+        -- Nuke doesn't explode on collide with air units
+        if other.Nuke and not self:GetBlueprint().CategoriesHash.EXPERIMENTAL then
+            if not IsAlly(self:GetArmy(), other:GetArmy()) then
```
speed2CZ: this is unnecessary nesting, it should be all tested in a single condition

lua/defaultunits.lua
```diff
@@ -1779,6 +1779,52 @@ AirUnit = Class(MobileUnit) {
...
+        
+        -------- default part from unit.lua---------
+        --------------------------------------------
+        
+        if self.DisallowCollisions then
+            return false
+        end
+
+        if EntityCategoryContains(categories.PROJECTILE, other) then
+            if IsAlly(self:GetArmy(), other:GetArmy()) then
```
speed2CZ: same here

lua/defaultunits.lua
Strogoo: https://github.com/FAForever/fa/pull/2943

Strogoo:
@KionX I left a comment inside, also check this https://github.com/FAForever/fa/pull/2943

the reason why we do this in airUnit is becuse unit's onCollisionsCheck is called before prjectile's onCollisionCheck. If we put our check in projectile's code, then it won't even be triggered. Nuke will explode during unit.OnCollisionCheck

also I added self.Nuke true because it's faster than engine calls. Replacing engine calls with lua checks is a common practice in fa repo. Ideally, we should replace all of them, but for now it's 50/50 and depends mostly on programmer's tastes

PR #2712, opened 2019-01-21
More random falling animation for Novax

Strogoo:
Only for idle or slowly moving Novax, otherwise it uses default animation.


Strogoo:
Final version

https://youtu.be/82REbRPArN8

PR #2718, opened 2019-02-06
Adds exit button to endscreen

Uveso:
In case there is no data available at the endscreen, we have no exit button.

This PR will add one.
If the game was launched from FAF it will show "Exit to FAF".
If launched from desktop it will show "Continue".

newexitbutton

### DIFF
--- a/lua/ui/dialogs/hotstats.lua
+++ b/lua/ui/dialogs/hotstats.lua
@@ -1031,6 +1031,7 @@ function Set_graph(victory, showCampaign, operationVictoryTable, dialog, standar
             bar_btn:SetCheck(false)
             graph_btn:SetCheck(false)
             dual_btn:SetCheck(false)
+            exit_btn:SetCheck(false)
         end
     end
     -- main title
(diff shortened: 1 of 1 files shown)


speed2CZ:
I dont think the button is needed:

The exit button is in the standard tab, but by default it opens the graph.

I would just switch it to open the standard tab first, since thats the way the close the game and who wants to know something extra, can click on the graph tab or any other.

Uveso:
This button is for the case that we have no standard graph.

So i need this button about 30-50 times a day.

speed2CZ:
Even with no data, there is an exit button. Right here, in the standard tab.

Uveso:
Well, i dont know how i should explain this.

I want a button that is clickable without clicking another button before.

And making the Standard tab as default is not what i want to change,
because the history is one of our FAF features, and should be displayed first.

So it's not obviously at this point how you can continue.

(Also if we have the standard tab as default and i am viewing the charts, then i still have to click on standard to click on exit or continue)

And my personal reason is that i miss that button really several times a day.
It's just anoying to not have a fixed place to click for faster screen changing / continue testing.

And since i have this button local working i can't live without it ^^

Uveso:
rebased Mar 11, 2019

nyov:
+1
The Graph tab is the default, but sometimes rendering the graph, after a long big match, crashes my game. I'd also like a fast exit button (before it crashes), or at least make "Standard" tab the default.

PR #2719, opened 2019-02-07
Optimized unitcount for VictoryConditions.

Uveso:
Uncompleted units will no longer count for end game conditions.

Also we don't really need to know how many units are alive, we only need to know if at least 1 unit is not dead.
So i renamed the function CountCurrentUnits to AllUnitsInCategoryDead.
The function is now returning true or false instead of a unitcount.

This way we only need to find the first unit and break the search loop, instead of looping over all units.


Uveso:
rebased Mar 11, 2019

PR #2732, opened 2019-03-02
Check for dead status before killing beams

Exotic-Retard:
so, if you called OnKilled on the ML and battlecruiser after they were
killed, then bad things happened since GetWeapon doesnt work on dead
units.

this removes the possibility of that happening.

done for clean code and mod support reasons.


speed2CZ:
So whats the scenario in which OnKilled is called after the unit is already dead?

Exotic-Retard:
its for custom scripts and things. so, for example transports call onkilled after units die. if mods start doing things like that (nomads have an exp transport), then the units will bug out the game. in general we want to avoid this happening. if magic code is added to faf at an undefined point in the future, it would also be nice to have this here.

PR #2733, opened 2019-03-02
Immobilize ACU during upgrade

Strogoo:


HUSSAR-mtrela:
@Strogoo don't we need this run this command when work/upgrade is completed?
`self:SetImmobile(false)`

Strogoo:
@HUSSAR-mtrela I copied this from here https://github.com/FAForever/fa/blob/develop/lua/defaultunits.lua#L2146

I think this hack works like "Stop" order that doesn't ruin other orders and only affects movement. We tell engine that this unit can't move -> engine stops it -> next tick we revert unit state so it can move again. But unit stay in place as it doesn't have active move order.

PhilipJFryFAF:
works as advertised - the acu doesn't move anymore while upgrading (can't even bump it out of the way)

PR #2735, opened 2019-03-05
Remove "Assist" command from factory during upgrade

Strogoo:


speed2CZ:
Please don't, Quite often I set my factories to upgrade and then set them to assist (via shift) on some other factory

Uveso:
Please don't remove the guard while upgrading. 
My support factories are assisting different HQ factories.
And when i upgrade the support factories then i don't remember what factory it was assisting.

Strogoo:
hm, how about removing it only from HQ then? Canceling upgrade for support factories by misclick is not so critical, but HQs...

Uveso:
I tested this now for about 5 hours. (script search and played 2 games)

I walked thru the command mode and ValidateAssist callbacks, defaultunit.lua etc.
The cancel command on assist is invoked from the c-engine and can't be prevented from LUA.

Can't find another way to fix the upgrade cancel problem.

So removing the CommandCap is the only way i see here.
And it has almost no conflict with a normal game play.
(already existing guard commands will not be canceled if CommandCaps are removed)

From my side this PR is good to go now!

shalkya:
image
i can get 2 flags, if i give a rally point while upgrading, then i change it after upgrade without deselecting the HQ.
It's not big deal, just need to know why this happening, if that kind lead to bigger issue.
Otherwise i merge it

Uveso:
The 2 flags has nothing to do with this PR,
It's a c-engine bug, because the Tech1 building is repaced with the Tech2 building after upgrading.
So you have still the flag from factory tech1 and can set the flag for tech2.
Ich you reselect the building then the Tech1 entry is lost and the 2nd flag too.

shalkya:
just wondering can you make it so it's possible to shift-assist, but impossible to assist only ? 
that way you can apply it on HQ and assist facto upgrading

Strogoo:
I doubt so. Since the code that handle factory assist is in engine, we can't add any conditions to it or interrupt user clicks.

shalkya:
tested, reviewed and approved

PR #2742, opened 2019-04-05
Adjust static shields build cost and time

TurinturambarFAF:
updates values for static shields.


speed2CZ:
the commit names should be the summary of the changes
commit name "new values" to the file which houses all unit's stats and parameters is on the same information level as "my hands are typing words"

Try smth like: "Adjust mobile shields build cost and time"

PR #2744, opened 2019-04-10
buff mongoose damage

JaggedAppliance:
Damage increased 15  -> 18
Damage per gatling gun volley 225 -> 270


Apofenas:
I request MuzzleVelocity increace from 25 to 35 to improve its ability to hit targets at range

PR #2745, opened 2019-04-10
Fix T3 UEF MAA missing (for real this time)

JaggedAppliance:
Values courtesy of EQ.


speed2CZ:
Integrating EQ confirmed

PR #2747, opened 2019-04-11
remove anti-air from t3 gunships

JaggedAppliance:
Remove anti air from wailers and broadswords to stop them from chasing air units that fly past them.


speed2CZ:
The weapons are also in the units scripts, you might want to remove them from there as well

speed2CZ:
You can just set AutoInitiateAttackCommand to false and they will no longer chase.
No need to remove the weapon

PR #2750, opened 2019-04-13
Vet changes

Strogoo:


aeoncleanse:
What in the name of all that's holy is this lot even trying to do? Why are you massively overcomplicating the vet logic? Why are you introducing levers that don't need to be there? Why are we getting rid of OnKilledUnit?

Strogoo:
Better ask Jagged. That was his idea and I'm just an executor.

*I still agree with these changes, just don't want to argue about smth that wasn't initiated by me. sry :D

image

Strogoo:
```Why are we getting rid of OnKilledUnit```

ACUUnit.OnKilledUnit* Because it adjusts mass that units give to ACU. Jag wants to remove this exception, so now ACU uses OnKilledUnit from main class without any multipliers.

PR #2758, opened 2019-04-18
Aeon SCU Chrono

Apofenas:
1)Adds Chrono upgrade to aeon SCU at the back;
2)Moves Nano to left hand;
3)Removes Sacrifice upgrade and integrates it into Engineering upgrade
4)Adds Advanced Combatant Preset with Chrono
5)Adds nano to Rambo preset

https://youtu.be/853inM8uoGM


shalkya:
the chrono dampner is too good. We should nerf it to 1/3 and allow 3 SCU to stack up their chrono dampner to get the same result you get right now. Also it shouldn't affect opponent SCU.
Moving upgrades around seems ok.
We should use the sera 2nd gun symbol instead of the gun symbol of the aeon scu. It is confusing with the chrono.

Apofenas:
I just copied values from aeon SCU chrono in deploy/fafbeta branch. My intent is the upgrade, performance is up to you.

I'm not sure what do you mean by stun stack? I'm not sure it works this way from what i saw with Cybran SCUs and nomad EMPs. Being affected by stun weapon while already being stunned doesn't reset the stun duration. So the way i'm seeing it iso far is to nerf the ROF or the stun duration itself. I thought about range decrease but it seems I made a typo with `MaxRadius = 30,` and `Radius = 35,` . I'll fix tommorow.

About stunning other SCUs. Cybran EMP upgrade can stun other SCUs. Why chrono wouldn't? Idk. This can be done by adding `'SUBCOMMANDER'` category to `TargetDisallow = ` line.

Also the upgrade uses Chrono icon from ACU and the advanced combat preset uses former nano preset. I believe it should be consistent with ACU, shouldn't it?

shalkya:
What i mean about chrono stacking/syncing this way : https://github.com/FAForever/fa/pull/162
i'm not sure if that's already in, and just require rebalance of the ROF. But 3 aeon scu with chronos would fire in a desync way (with their ROF decreased), so using up to 3 scu at the same spot would be useful. But adding more chrono scu wouldn't matter, since they would fire at the same time as another scu.

Imo we shouldn't make the scu so good to kill each other. So stun (from cybran/aeon scu) should reduce/inefective on other scu, and OC from sera shouldn't be as deadly vs scu. But we need to be careful when balancing that, so uef scu aren't that deadly against other scu.

Yes the chrono icon should be the same as the ACU, but the gun upgrade icon on aeon scu is too similar to the chrono. (didn't talk about advanced combat preset icon) So we should use the icon from sera ACU for its 2nd gun (aoe) or simply gun icon from the aeon ACU.

Strogoo:
image


So we need some thread that constantly checks sacu positions, moves sacu from one sync to another depends on which sacus are near (and from what Sync they are) + ACU too. Honestly, this looks messy even on paper, still doable tho. But I'd rather find other way to balance sacu chrono tbh :)

Apofenas:
There is a way to limit chrono-spam (if it was a big be a problem) with power drain. There are lines atm for EnergyDrain, but Chrono doesn't actually need power to run. If it needed 15k e per shot, it would make usage of multiple SCUs to completely stun an army far more complicated as you'd have to heavily invest in power generators and energy storages just to run this.

PR #2776, opened 2019-04-28
rebalance Quantum Gateway

shalkya:
increase BP of QG

120 BP -> 180 BP

Make building several QG worth it, reduce the need of engie for SCU spam and makes adjacency relevant.


Techmind:
Its 120 bp for 3k metal (25 metal per bp), t3 engi is 40 for 300 metal (7.5 metal per bp). Engies are 3x more effective.  180 is still too low. It does not fix the problem.

PR #2777, opened 2019-04-28
rebalance beetle

shalkya:
make it the unit that cybran can rely on to defend the upgraded ACU push.
1 beetle kill a T2 pgen/mexx
3 beetle kill a T3 pgen/mexx

### DIFF
--- a/units/XRL0302/XRL0302_unit.bp
+++ b/units/XRL0302/XRL0302_unit.bp
@@ -36,20 +36,6 @@ UnitBlueprint {
             LodCutoff = 'UnitMove_LodCutoff',
         },
     },
-    Buffs = {
-        Stun = {   
-            Add = {
-                OnImpact = true,
-                OnDeath = true,
-            },
-            AppliedToTarget = false,
-            BuffType = 'STUN',
-            Duration = 2,
-            Radius = 6,
-            TargetAllow = 'MOBILE', --ALLUNITS
-            TargetDisallow = 'EXPERIMENTAL, WALL, COMMAND',
-        },
-    },
     BuildIconSortPriority = 80,
     Categories = {
         'PRODUCTFA',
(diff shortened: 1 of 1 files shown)


douglasg14b:
Shouldn't a PR of this sort be required to state the exact changes made in the post? The post is very misleading as this doesn't look like a trivial change to beetles, and has many other affects than just `make it the unit that cybran can rely on to defend the upgraded ACU push`...

Changes:

* Removes Stun effect
* Health: 500 -> 700
* Damage: 1499 -> 3500
* Damage Radius: 6 -> 2
* Build Costs:
  * Energy: 1800 -> 3000
  * Mass: 300 -> 400
  * Time: 1200 -> 1500
* Changes transport class from 1 to 2 (I asusme this means it requires T2+ to transport)

shalkya:
beetle are useless right now, and cybran has issues defending upgraded com push. Also it give the beetle the ability to be a tool to snipe eco/important structure as explained in the PR

douglasg14b:
>beetle are useless right now

Kinda, yeah. I suppose that I don't see reducing the damage radius, increasing the damage, and making a small increase to health as being large enough changes to fix that?

At least from my understanding, the damage radius is largely what made beetles effective in the limited roles they already have? Increasing damage is good and all, but it doens't help if they have to get even closer to their target to apply damage. The health buff does help this, but is 200hp enough to make the change drastic enough for them to be worth-while? i would think a marginal speed buff would do just as well, or better, than a health buff?

aeoncleanse:
IMO this is a bad design in the first place. This isn't really something Cybran needs, and it's not a satisfying weapon to use or have used against you. Just because people originally began to use beetles as a snipe weapon doesn't make it good design. Please see my beetle rebalance PR for an alternate proposal to give beetles an intuitive, coherant design across every aspect of the unit, which also integrates into the Cybran warfare style and fills a gap in their arsenal in a unique way.

PR #2785, opened 2019-05-13
improve targeting of torpedo bomber and subs (+ amphibous exp)

shalkya:
subs (+ amphibous exp) : 
doesn't focus AA anymore
tech 2 navy is targeted over other stage naval unit

torpedo bomber : 
they will focus AA (+ shield boat)
then they focus subs
then they focus T2 navy
then they focus T3 navy
then the rest


units/UAA0204/UAA0204_unit.bp
```diff
@@ -321,6 +321,13 @@ UnitBlueprint {
             TargetCheckInterval = 2.5,
             TargetPriorities = {
                 'SPECIALHIGHPRI',
+                'CRUISER',
+                '(NAVAL * TECH2 * SHIELD)',
+                '(STRUCTURE DEFENSE * ANTIAIR)',
```
Strogoo: '(STRUCTURE * DEFENSE * ANTIAIR)'

same for other blueprints

PR #2786, opened 2019-05-20
Sera ras scu

Apofenas:
Mass production: 2->1
Energy production: 200->20
Adds RAS upgrade
Adds RAS preset

Credit to Androish and Ulveso who made it for PhXModAlpha


Garanas:
Is this PR still valid? I'd like to close up if it is not going to be used anymore.

Basilisk3:
@Garanas I think this PR could be reopened, because the topic has gained new attention recently. I, for one, fully support this being added to the game.

- It does not make sense for the Seraphim to lack an eco-step.
- Right now, Seraphim players who really want to make RAS-SACUs just ask for an engineer. Not having this option simply results in them having to wait around.
- Increasing the base output of all SACUs to cope with the lack of a RAS option, opens up another can of worms balance wise, because now the combat-SACUs also generate more eco. In the PRs concerning the SACU-rebalance the base output of Seraphim SACUs is increased even more, which I think is overcomplicating things.

If you are looking for more arguments, I think @Apofenas went into detail in his posts on Discord.

Garanas:
@Basilisk3 I understand that there's a problem. 

I'd like to try the direction as proposed in https://github.com/FAForever/fa/pull/5805 first. It is not as good as RAS, but it gets a lot closer to it while giving the Seraphim faction other benefits such as having more total hit points, fire power and build rate because there will be more SACUs.

Basilisk3:
I will test it and let you know what I think.

PR #2787, opened 2019-05-23
Mass storage low target priority

shalkya:
make storage targeted after everything else


shalkya:
tested by nojoke, works.

PR #2793, opened 2019-06-17
Report time

Rackover:
The player needs to wait 3 seconds before clicking on Report. This change prevents people from making fake reports by accident.

### DIFF
--- a/lua/ui/dialogs/teamkill.lua
+++ b/lua/ui/dialogs/teamkill.lua
@@ -13,7 +13,6 @@ local TextArea = import('/lua/ui/controls/textarea.lua').TextArea
 local dialog = false
 local shouldReport = false
 local waitingTimeBeforeReport = 3;
-local waitedTime = 0;
 
 function CreateDialog(teamkill)
     if dialog then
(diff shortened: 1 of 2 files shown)


speed2CZ:
What happens if you click forgive right away? Does the forkthreaded function die or throw an error?

Rackover:
> 
> 
> What happens if you click forgive right away? Does the forkthreaded function die or throw an error?

It doesn't do anything as far as I've tested, but you may want to test it further

lua/ui/dialogs/teamkill.lua
```diff
@@ -58,4 +59,19 @@ function CreateDialog(teamkill)
 
     dialog.OnShadowClicked = function(self)
     end
+    
+    UIUtil.setEnabled(reportBtn, false)
+    ForkThread(WaitBeforeEnablingReportButton, reportBtn)
```
speed2CZ: Try smth like `dialog.CounDownThread - ForkThread(WaitBeforeEnablingReportButton, reportBtn)`
then if you Forgive before the coundown is finished, the thread should be destroyed together with the dialog window

Else it throws error on line 71, since you're trying to call a function on a object that doesn't exist anymore.

PR #2800, opened 2019-06-24
Fix Techlevel missmatch

Uveso:
Fixed wrong RULEUTL entries in BP.General.TechLevel


PhilipJFryFAF:
what's the difference between basic, advanced and secret?

Uveso:
Basic= Tech1
Advanced = Tech2
Secret = Tech3
Experimental = Tech4

HUSSAR-mtrela:
@Uveso  really good job on finally fixing BP.General.TechLevel in blueprints. Myyours upcoming blueprint verifier would catch this kind of issues

PR #2805, opened 2019-07-03
Modblacklist exp war

Uveso:
Changed label for Experiemtnal Wars from BROKEN to UPGRADE, because we have a newer version in our vault.


HUSSAR-mtrela:
@PhilipJFryFAF @aeoncleanse it's safe to merge this PR since it has very minimal impact on the game

PR #2806, opened 2019-07-04
update seraphim shader to be more realistic

BlackYps:
The seraphim shader looks weird at the moment. Especially on maps with very bright lighting. I gave the shader a look closer to the original SCFA shader while preserving the ability to have the teamcolor influence the shader. Here is the corresponding forum thread: https://forums.faforever.com/viewtopic.php?f=2&t=17676
The shader is tested on many maps. It reacts nicely to the different environments.

The build shader now also utilizes the teamcolor.

### DIFF
--- a/effects/mesh.fx
+++ b/effects/mesh.fx
@@ -2860,34 +2860,39 @@ float4 UnitFalloffPS_02( NORMALMAPPED_VERTEX vertex, uniform bool hiDefShadows) 
     float3 reflection = reflect( sunDirection, normal);
     float specularAmount = saturate( dot( reflection, -vertex.viewDirection));
     float3 phongAdditive = pow( specularAmount, 9) * specular.g * shadow * 0.7;
-	
+
     // Calculate environment map reflection
-    float reflectivity = saturate(specular.r * 2.5); //reduce artifacts of texture
+    float reflectivity = saturate(specular.r * 2.5); // Reduce artifacts of texture
     environment *= reflectivity * fallOff.a;
     float3 phongMultiplicative = light * environment * (1 - (diffuse.a * 0.5) ) * 0.7;
     
-    //make reflection more intense depending on the diffuse color
-    //could be cool but looks like shit because the diffuse texture is not interpolated for unknown reasons
-    //float Amount = (diffuse.r + diffuse.g + diffuse.b) / 3;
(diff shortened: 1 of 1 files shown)


aeoncleanse:
Picky, but could you go though and bring it fully up to style?
- `// Comments` should have a space after the `//`, be capitalised, and only have a `.` when doing complex sentences
- Please remove code, instead of commenting it out. Git provides history for us
- See if you can use blank lines (whitespace) to split chunks into sensibly divided sections with appropriate commentry. For most of us, shaders are a total mystery, so having some explanation for each logical block would be sweeeeeet

BlackYps:
Will do the style changes.

The commented out code is new code (actually rewritten old code, but the old one was immensely hard to understand) so it is not in the git history. But it has a bug and needs someone with in depth shading knowledge to fix. I wanted to preserve it in the future so people get reminded, that this is possible. If there is another option to save the code, I am also fine with that.

 I tried to not comment literally what the code does and tried to split it in chunks already, but maybe I assumed too much knowledge. It is hard to describe what the code does for someone with zero knowledge how rendering works, though, because the result is all visual. Can you give me a hint where you need more explanation?

aeoncleanse:
Usually it's good practise not to be too literal when programming. However, because this is a kind of open, community effort driven by novices without much experience, I encourage people who understand what's going on to put in full commentry whenever possible. Makes it easier for someone in future to maybe update or learn from it see?

For code that needs a more expert look that you've commented, maybe put a `TODO: ...` comment block above it, explaining why it's there and what needs looking at?

BlackYps:
I decided to write a Wiki page instead of spamming the code with endless comments. I think in the wiki the information is also more likely to be found. https://wiki.faforever.com/index.php?title=Shaders

speed2CZ:
@BlackYps can this be improved in the shader?
From certain angle the team color takes over the LOD1 model
ForgedAlliance_2019-07-22_18-42-35
with slightly different angle the unit looks fine
ForgedAlliance_2019-07-22_18-43-12
(those are gunship and ASF, but I think it happens on more units)

Some more pictures, with yellow it's really burning my eyes:
ForgedAlliance_2019-07-22_18-46-02
ForgedAlliance_2019-07-22_18-46-17

Could you investigate?

PR #2809, opened 2019-07-05
patch 3706 : reverting to stable 3704

shalkya:
revert all change of the game patch + update the changelogs


shalkya:
not needed

PR #2811, opened 2019-07-08
Fix invulnerable sub-carried units

aeoncleanse:
- Load unit into Stinger
- Load Stinger into any Naval Aircraft Carrier (Czar doesn't work)
- When Carrier dies, we get invulnerable, invisible units capable of attacking

This was caused by the engine Kill() command calling SetDead() before the detach/kill code was called, so the `if not self.Dead` safety check backfired and prevented the proper handling. 

This PR refactors some duplicated loop code, moves it into BaseTransport for inheritance, and adds a Kill() override to AircraftCarrier, fixing the bug.

Please test as many combinations of units inside units inside units as you can think of. Of concern I don't have time to check right now:
- Fatboy
- Air Staging

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2069,6 +2069,13 @@ SeaUnit = Class(MobileUnit){
 
 --- Base class for aircraft carriers.
 AircraftCarrier = Class(SeaUnit, BaseTransport) {
+    Kill = function(self, ...) -- Hook the engine 'Kill' command to flag cargo properly
+         -- The arguments are (self, instigator, type, overkillRatio) but we can't just use normal arguments or AirUnit.Kill will complain if type is nil (which does happen)
+        local instigator = arg[1]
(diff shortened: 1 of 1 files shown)


PhilipJFryFAF:
tested:
* loading an aurora into a stinger
* loading the stinger with the aurora into an atlantis or czar
* destroying the atlantis/czar via self destruct or instant self destruct

when killing the carrier via instant self destruct we still have the invisible/invincible aurora being left behind

when killing a transport with units loaded via self destruct then the loaded units will not die anymore

will need someone to fix both issues since a) the PR doesn't fix what it says it does and b) introduce a new bug

shalkya:
i cannot replicate the bug in normal faf.
1- load T2 tank into stinger
2- load stinger into atlantis or aircraft carrier
3- ctrl k aircraft carrier/atlantis (also during the unloading)

PR #2812, opened 2019-07-09
weapon category fix

shalkya:
- put the max OC dps for unit database calculation
- rename autoOC names
- rename experimental weapon

last change can be reverted, as "direct experimental" cat was used for high dps lazer/energy weapon. (cybran lazer included). Up for debate.


shalkya:
Actually I did the opposite, I replaced : 
```WeaponCategory = 'Direct Fire',```
with:
```WeaponCategory = 'Direct Fire Experimental',```

I putted that on all direct fire from experimental unit. But as I explained above:
>     - rename experimental weapon
>     last change can be reverted, as "direct experimental" cat was used for high dps lazer/energy weapon. (cybran lazer included). Up for debate.

HUSSAR-mtrela:
alright, it make sense now. Thanks

aeoncleanse:
I really don't think the cat rename is a good idea. That description is to let the player know what physical behaviour the weapon has, not how strong it is.

PR #2814, opened 2019-07-13
Fix/improve resource sharing

speed2CZ:
It's no longer possible to send resources to yourself or send negative
values

Untested

### DIFF
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -603,13 +603,15 @@ end
 
 function GiveResourcesToPlayer(data)
     SendChatToReplay(data)
-    if data.From != -1 then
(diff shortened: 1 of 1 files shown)


speed2CZ:
Maybe test what happens if you try sending to yourself 0 instead of negative values

norraxx:
And -0 too

Strogoo:
Works fine for me. Tried to send negative mass to myself and to other player, both were rejected.

Also sending `0` and `-0` doesn't affect anything

rjy7wb:
add a team resource feature that shows the amount of mass/energy being overflowed by the team

speed2CZ:
Make a new issue for other features/requests. This was a bug fix.

PR #2816, opened 2019-07-13
Improve team kill detection

### LINKED ISSUE #2815: Better team kill detection
Zulip_2019-07-13_14-08-25

- [x] Detect if the weapon that did the kill was a death weapon
- [x] Detect if the unit got self destructed
- [x] Report only if the unit killing the ACU with death explosion was self destructed by the player
- [ ] Obtain 1TB SSD

speed2CZ:
Adds a new DamageType to death weapons so we can tell if a unit got killed by the death explosion.
Marks units that got self destructed by the plyer.

Ignore kills caused by ally death explosions of units that were not self
destructed by the player.
Fixes #2815

Tested briefly.


PhilipJFryFAF:
@aeoncleanse can you do a quick review and give me your ok/merge it?
the auto report feature is generating way too many unwanted reports 90% of which are chain reactions which will be filtered out and thus not generate a report popup any longer with this PR

PR #2817, opened 2019-07-14
Fix target bones of Cybran T2 Land HQ

### LINKED ISSUE #2784: shots going through cybran T2 HQ while upgrading
[9655064-keyser.zip](https://github.com/FAForever/fa/files/3169494/9655064-keyser.zip)

at 17:45 min, i've 3 harbs shooting at the HQ upgrading to T3. the 3 of them are missing the HQ, shots go through

speed2CZ:
Could fix Fixes #2784

The BP has a target bone from a T1 factory.

Comparing visually the position of the bones, this could fix the issue.

Untested, go crazy.

### DIFF
--- a/units/URB0201/URB0201_unit.bp
+++ b/units/URB0201/URB0201_unit.bp
@@ -1,7 +1,7 @@
 UnitBlueprint {
     AI = {
         TargetBones = {
-            'URB0101',
+            'URB0201',
             'B07',
             'B08',
         },


aeoncleanse:
Hmmm. All these choices for target bone seem odd. May I ask that you increased the size of the collision box to better match the visuals please, it's much too small, and then add AttachPoint to the top of the bone list? That I'm confident will fully bring this up to standard.

speed2CZ:
The hitbox is not bigger in T3 hq so I rather not do that. I didnt add any bones or stuff, I just looked in the game what they names and positions and I think this could help

aeoncleanse:
Sounds to me like the boxes of all of them could use matching to their visuals then, if you're feeling adventurous.

Please add AttachPoint to the top of the list though, as well as changing it to use the right unit ID. AttachPoint is centrally located above the base bone, and IMO is the best target choice. Will help with things targeting it in adverse terrain as well.

speed2CZ:
T1 and T3 factory doesn't seem to have this issue. Or at least it wasnt reported. So if this fix help, I would rather not mess with the hitboxes

aeoncleanse:
Up to you to mess with the box or not. _Please add the additional bone to the list though!_

speed2CZ:
I might but I would still prefer if someone tried if this improves it and then we can mess with it even further

aeoncleanse:
Oh for goodness sake...

PR #2818, opened 2019-07-14
Fix Hives build effects

### LINKED ISSUE #2715: Hive build beams attached to wrong bone
2 beams that cotrols the flying bots are attached to the middle of the unit.

https://github.com/FAForever/fa/blob/develop/lua/EffectUtilities.lua#L470-L495
This code kinda expects to have same number of build bones as build bots, but the hive has 3 bots and only a single bone.

forgedalliance_2019-02-02_12-40-41

Prolly broken for upgraded hives as well, someone can check that...

speed2CZ:
Since build bots are calculated from the build power, hives don't have
1/2/3 bots anymore but 3/4/6
One per 15 build power
Fixes #2715

### DIFF
--- a/units/XRB0104/XRB0104_unit.bp
+++ b/units/XRB0104/XRB0104_unit.bp
@@ -161,6 +161,8 @@ UnitBlueprint {
             AimBone = 0,
             BuildEffectBones = {
                 'Attachpoint01',
+                'Attachpoint01',
+                'Attachpoint01',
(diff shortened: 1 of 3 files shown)


speed2CZ:
@aeoncleane do you know the reason why the number of build bots is calculated now insted of set?
Its making hives more laggy by spawning more bots than before

PR #2821, opened 2019-07-14
Fix Salem floating when killed during landing

### LINKED ISSUE #2638: cyb destroyer wreck levitation
image
when it dies during the leg deployment animation

speed2CZ:
Fixes #2638 

Test all possible combinations please, from water to land, from land to water, from water to orbit, you know...


speed2CZ:
It this tested that its getting merged already? :D if you end up with salems shooting into th sky...

PR #2823, opened 2019-07-14
Play unit death animations only for completed units

### LINKED ISSUE #2399: sinking mega wreck bug
image
image
[game.uid.7566744.log](https://github.com/FAForever/fa/files/1774669/game.uid.7566744.log)
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=7566744
if you ctrl k mega after 50% completion. death animation execute, mega falls but it is already on the ground. Hence it fells underground.

speed2CZ:
Fixes #2399
This bug isnt affecting only Mega, but all experimental units. depending on the animation, the unit could behave weirdly. Mega the most, since it has to first stand on the legs when completed and the death animation counts with this.

Anyway, others are affected as well, so I think the animation should be played only for units that are actually completed.

As usually test properly please.

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1246,7 +1246,7 @@ Unit = Class(moho.unit_methods) {
             self:PlayUnitSound('Killed')
         end
 
-        if self.PlayDeathAnimation and self:GetFractionComplete() > 0.5 then
+        if self.PlayDeathAnimation and self:GetFractionComplete() == 1 then
             self:ForkThread(self.PlayAnimationThread, 'AnimationDeath')
             self.DisallowCollisions = true
         end


aeoncleanse:
I suspect this won't be enough. If it dies the tick or two after completion, it's still getting up. We may have to wait until the stand-up animation completes to "turn on" the death animation?

speed2CZ:
Can someone test how those units behave in vanilla FA conpared to this PR, so we know if we're going the right direction

shalkya:
could you apply this only on ML and mega ?
it looks weird on GC, ythota and SR.
image
also it recreates the ghost atlantis bug and ghost tempest bug.
image
image
those are dead

PR #2824, opened 2019-07-14
Fix ML exhaust effects playing during construction

### LINKED ISSUE #2396: ML does its "engine smoke" while under construction
image
i don't think this is on purpose it should be able to use its engine only when finished.

speed2CZ:
Fixes #2396

### DIFF
--- a/units/URL0402/URL0402_script.lua
+++ b/units/URL0402/URL0402_script.lua
@@ -42,6 +42,7 @@ URL0402 = Class(CWalkingLandUnit) {
 
     OnStopBeingBuilt = function(self, builder, layer)
         CWalkingLandUnit.OnStopBeingBuilt(self, builder, layer)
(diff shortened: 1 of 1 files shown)


aeoncleanse:
This fix doesn't seem right... Surely we should need to track down, and remove, the original call which creates the effect in the first place?

speed2CZ:
It happens on layer change. You can see it in the script. Different effext for land and water layer.

Problem is that its called right away when you start building it. Since it goea from no layer to land for example.

aeoncleanse:
So is the correct solution not to put a guard in place around the call to create effects, inside layerchange, to check completion, thereby fixing this bug in _all_ units at once?

speed2CZ:
What all units? It's a ML specific effect inside the ML script, where else would you want the check to be?

aeoncleanse:
Oh yes, I see now

PR #2825, opened 2019-07-14
Stop factory building beams when paused

### LINKED ISSUE #2387: remove the build beam when factory is paused
when factory is paused, it still displays the build beam as if you were still building. it should be turn off, so opponent can see your factory is paused.
image
image
image
image

[game.log](https://github.com/FAForever/fa/files/1754413/game.log)
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=7538250

speed2CZ:
Fixes #2387

Please test properly, pausing unpausing all kinds of factories when building smth or when theey are being upgraded. Cancelint the unit in production or upgrade.
Even mobile factories, so nothing get broken by this.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -587,7 +587,7 @@ StructureUnit = Class(Unit) {
     end,
     
     DoTakeDamage = function(self, instigator, amount, vector, damageType)
-	    -- Handle incoming OC damage
+        -- Handle incoming OC damage
         if damageType == 'Overcharge' then
             local wep = instigator:GetWeaponByLabel('OverCharge')
             amount = wep:GetBlueprint().Overcharge.structureDamage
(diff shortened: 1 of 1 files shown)


aeoncleanse:
Can tell you right now that any carriers, Fatboy etc won't benefit from this as they don't inherit from these classes at all.

speed2CZ:
They dont need to benefit from it as they dont have any build effect since they use building the units inside them.

Not sure what you want with the second one? I basically took this from engineers

aeoncleanse:
Just below the new flag is an old self.BuildingUnit = true. There doesn't seem to be much need for this any more, since it can probably be replaced everywhere it's used by the new one.

speed2CZ:
I see, I'll check that

speed2CZ:
This goes actually much deeper than I wanted :D I did the tests only on UEF factories and they work correctly, but other factions are using different approach and this PR breaks them. So I either figure that shit out or I close it.

aeoncleanse:
Or if the other factions already work properly, move this UEF-specific fix into the UEF class file where they belong

shalkya:
other faction don't work properly, as you can see in the associated issue.
keep us updated on the progress you been doing here.

PR #2835, opened 2019-07-24
Use LOD1 models

speed2CZ:
Adds LOD1 model to be used for some units which have the model and textures and look decent enough (like other units LOD1 models that exist)

### DIFF
--- a/units/UAL0401/UAL0401_unit.bp
+++ b/units/UAL0401/UAL0401_unit.bp
@@ -108,9 +108,15 @@ UnitBlueprint {
             IconFadeInZoom = 130,
             LODs = {
                 {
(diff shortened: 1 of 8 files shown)


shalkya:
what's the goal of adding the lod1 in ? to make the transition smoother ? does it help for the GPU ?
It just make unit darker at first. I guess it help making 2 transitions instead of 1 for the lod 2 (guess it's lod 2).
for aeon battleship, it's smoother when you zoom in, but then zooming off it's less smooth. If you mean by smooth : that when you start zooming in, it rapidly change to lod 2.

speed2CZ:
Low poly models = less load = more FPS

PR #2840, opened 2019-07-30
Curated maps button

PhilipJFryFAF:
this PR adds a button to the skirmish and Multiplayer LAN Menu which opens the site http://forums.faforever.com/viewtopic.php?f=2&t=17820 which is currently a placeholder where the m&m councilor will add curated maps to encourage users to play different maps than they usually do

first time i add something to the loc file so not sure if i break any conventions here - if i do then pls tell me how to do it properly


loc/US/strings_db.lua
```diff
@@ -3585,6 +3585,7 @@ _Off="Off"
 _Ok="Ok"
 _On="On"
 _Options="Options"
+_Curated_Maps="Curated Maps"
```
speed2CZ: `_tag` is used for some general ones, better would be to find some `lobui_XXXX` that is free, so all the lobby stuff is in one place

PhilipJFryFAF: changed as requested in a661f60

loc/US/strings_db.lua
```diff
@@ -7065,6 +7064,7 @@ lobui_0429 = "Move Player to Observer"
 lobui_0430="Country not found. Please contact the development team on the forums, or ask a moderator for help"
 lobui_0431="Close - spawn mex"
 lobui_0432="Closed - spawn mex"
+lobui_0433="Cruated Maps"
```
speed2CZ: So is it Cruated or Curated? :P

lua/ui/lobby/lobby.lua
```diff
@@ -2821,7 +2821,7 @@ function CreateUI(maxPlayers)
     end
 	
 	-- curated Maps
-	GUI.curatedmapsButton = UIUtil.CreateButtonWithDropshadow(GUI.panel, '/Button/medium/', "<LOC _Curated_Maps>")
+	GUI.curatedmapsButton = UIUtil.CreateButtonWithDropshadow(GUI.panel, '/Button/medium/', "<LOC lobui_0433>")
```
speed2CZ: You can make the string `"<LOC lobui_0433>Curated Maps"` the code grabs the localized version and the one in here is a fallback. But it also makes it slightly more readable, so you don't have to search for what lobui_0433 is.

loc/US/strings_db.lua
PhilipJFryFAF: fixed spelling in 3d5ba19

lua/ui/lobby/lobby.lua
PhilipJFryFAF: added the fallback in 3d5ba19

BlackYps:
The button has the same tooltip like the Team Count button for the autofill feature. ("How many teams should there be?")

shalkya:
@BlackYps should be fixed, thanks for the report

PR #2842, opened 2019-08-03
AutoFill lobby slots

KionX:
Adds the AutoFill feature slots lobby AI and their division into nth number of teams


PhilipJFryFAF:
seems to work just fine - even supports AIs that are loaded via mods (eg. Uveso AI or Dilli)

shalkya:
would it be possible to add a message on hover especially for the number of team. It isn't obvious at first glance.

shalkya:
```
 `Teams Count`

How many teams should there be? (Players and AI will be split evenly between, will not work if the number of teams specified can't be divided into this number)
```
could you use this tooltip ? Is it too long ?
also can you make so when you can't divide (eg 8/3) then you it doesn't change or give a warning. I guess you can keep it like you did, but need to have a notification about that. Because putting 3 and getting 4 teams is weird.

PhilipJFryFAF:
@KionX i'd like to merge your PR but i'd prefer having the suggested changes from @shalkya in if possible

PR #2845, opened 2019-08-14
Add new naval platoon templates for the campaign AI

speed2CZ:
Used in campaign missions.


shalkya:
what do we need to test here ?
is it still on going ?

speed2CZ:
You can't test much here. Those are extra platoons that can be used in the campaign. Currently used only in the mission Im was making.

It's done for now.

PR #2847, opened 2019-08-19
Fix Reactive AI

speed2CZ:
It's used in the missions to react to players exps and game enders.

It needs testing since all FA missions use this but since it was mostly broken... so hard to say how good or bad it will now be.


shalkya:
sandbox : give AI a para and tons of BP, start an exp or a gameender, get it scouted and look for AI reaction ?
that's how to test it ?

speed2CZ:
This is only in the campaign, when such AI is set up. But right now it would require a lot of changes to all of the FA missions. Since the fixed version behaves quite differently.

PR #2855, opened 2019-09-28
sera ACU upgrade reorganisation

shalkya:
move regen aura from the right to the left. Allowing the combo gun + regen aura.
It removes the combo T2 + regen aura, that was useful for navy. But nano can do the survivability job.
Moving RAS from back to right, in order to allow several choice on the right arm.


shalkya:
duplicate of #2881

PR #2863, opened 2019-10-03
Add new optimal / mirrored spawn

Crotalus:
New optimal, mirrored option

For example, if you're ranked `#1` in your team you will be against the guy ranked `#1`  in the opposite spot.

Won't work on maps where spots aren't in logical order, I'm too lazy to start calculate mirrors by looking at left / right / top bottom option  + the x,y positions on the map

Needs testing


shalkya:
i'm not sure how to test this apart from merging in fafdevelop and setting games up there @PhilipJFryFAF

shalkya:
image
doesn't display as expected.

Crotalus:
> doesn't display as expected. i guess you want to name it "optimal mirrored balance"

Updated the PR with a fix now

shalkya:
Need to get tested in developed.

shalkya:
image
would it be better to change to "mirrored balance" only ?

PR #2865, opened 2019-10-04
Add allies resources balance to scoreboard

KionX:
Present three mode display resources.
Switch button located below energy icons.
I - Income (default), B - Balance, S - Storage

Transferring mass, energy, units on click with Shift and
request him with Ctrl.

Add display map size and replay ID.
Fixed formatted negative values displayed resources.

### DIFF
--- a/lua/sim/score.lua
+++ b/lua/sim/score.lua
@@ -128,6 +128,8 @@ function ScoreThread()
                     total = 0,
                     rate = 0                
                 },
+                StoredMass = 0,
+                StoredEnergy = 0,
                 MassReclaimRate = 0,
                 EnergyReclaimRate = 0
             }
@@ -176,6 +178,9 @@ function ScoreThread()
             ArmyScore[index].resources.energyout.total = brain:GetArmyStat("Economy_TotalConsumed_Energy", 0.0).Value
             ArmyScore[index].resources.energyout.rate = brain:GetArmyStat("Economy_Output_Energy", 0.0).Value
             ArmyScore[index].resources.energyover.total = brain:GetArmyStat("Economy_AccumExcess_Energy", 0.0).Value
+            
+            ArmyScore[index].resources.StoredMass = brain:GetEconomyStored('MASS')
+            ArmyScore[index].resources.StoredEnergy = brain:GetEconomyStored('ENERGY')
 
             for unitId, stats in brain.UnitStats do
                 if ArmyScore[index].blueprints[unitId] == nil then
--- a/lua/sim/score.lua
+++ b/lua/sim/score.lua
@@ -122,14 +122,18 @@ function ScoreThread()
                 },
                 massover = {
                     total = 0,
-                    rate = 0                
+                    rate = 0
                 },
                 energyover = {
                     total = 0,
-                    rate = 0                
+                    rate = 0
                 },
-                StoredMass = 0,
-                StoredEnergy = 0,
+                storage = {
+                    storedMass = 0,
+                    storedEnergy = 0,
+                    maxMass = 0,
+                    maxEnergy = 0,
+                },
                 MassReclaimRate = 0,
                 EnergyReclaimRate = 0
             }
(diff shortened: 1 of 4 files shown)


shalkya:
wouldn't it be better to have the total production of the player instead ?
possibly it would still be nice to have a way to display the instant eco, but i think total production is more important

KionX:
The current solution allows you to analyze the activities of allies.

shalkya:
@HUSSAR-mtrela do you think you will be able to integrate this into your supreme score board ?

shalkya:
doesn't work with supreme score board (obviously) and with ecomanger.
Tell us when you are done on your changes @KionX

Nojoke185:
Not sure this is working in latest commit?
see screenshot:
image
it is only displaying local player income/capacity (not allies)
Was ok yesterday before the addition of allied warehouse levels (using commit 1a109c57d0dad8f2fb3bee32cf4a00a086ae1ea8)

shalkya:
> 
> 
> Not sure this is working in latest commit?
> see screenshot:
> image
> it is only displaying local player income/capacity (not allies)
> Was ok yesterday before the addition of allied warehouse levels (using commit [1a109c5](https://github.com/FAForever/fa/commit/1a109c57d0dad8f2fb3bee32cf4a00a086ae1ea8))

did you disable your ui mods (especially the ecomanager one) ?

HUSSAR-mtrela:
@shalkya yes I will be able to integrate this into your supreme score board

Nojoke185:
> > Not sure this is working in latest commit?
> > see screenshot:
> > image
> > it is only displaying local player income/capacity (not allies)
> > Was ok yesterday before the addition of allied warehouse levels (using commit [1a109c5](https://github.com/FAForever/fa/commit/1a109c57d0dad8f2fb3bee32cf4a00a086ae1ea8))
> 
> I think those 2 weren't allies ;)
> (works fine on my end)

Ah yes, its always the small details that catch you out! :)
forgot to ally them in the lobby before launch, all was not lost however, at least it proved enemies cant see your resource usage!

Nojoke185:
Could this be made to have a toggle switch to show either current usage or current storage (and remember the last one used?)

KionX:
It's final version.

shalkya:
I couldn't figure out how to give mass and energy (using shift didn't work)

Nojoke185:
just thinking about Phantom game mode, would it need to be able to hide allied income etc for this mode else everyone knows who is the phantom! Not sure how that would be achieved

Nojoke185:
Tested latest commit, generally looks ok

Tested:
- allies can see each others : income, balance, storage
- figures reported look correct
- Enemy cannot see your eco
- Observers can see all players eco
- No log errors

Issues/Questions:
1
A player who dies moves to observer view and can see all players eco
This is fine if observers is enabled (expected people can be trusted to not share eco info with players)
If observers is disabled player list still shows.
Should we not share eco info when observers is disabled?

2
I can see active players list moves around a bit (however an observers players list stays static)
in below example:

Nojoke is host (allied to Nojoke1)
Nojoke2 is enemy

Player list sometimes shows order as:

- Nojoke
- Nojoke1
- Nojoke2

image

then it randomly switches for a short while to:

- Nojoke
- Nojoke2
- Nojoke1

image

then switches back and repeats

Perhaps there some sorting by eco going on (both Nojoke1 and 2 had same eco)?

3
Dead players eco is reported as 0
Should player list show 0 or remove thier eco info when they are dead?
example - Nojoke1 is dead in below:
image

4
Finally was there anything to restrict eco viewing for Phantom game mode in latest commit?
if yes not sure how to test that locally

Thanks

Nojoke185:
Hi

The latest commit allows a player to move viewpoint and control different teams units/structures via clicking other players name in the player list box?
Player can also view observers mode by clicking new observers button

This happens whether observers mode is enabled or disabled
This isn't intended behaviour I assume? (otherwise anyone could take over an opponents units and give orders) or view the whole map

Thanks

Nojoke185:
Ah that explains it, ok i will check it with cheats off also
Thanks

Nojoke185:
just an update, tested the latest code with no issues noticed

The cheat mode change you put in is quite useful actually now I understand it :)
one additional suggestion : can you calculate and display mass and energy overflow per player?

KionX:
I think I S modes enough to get all the necessary information. Mode B should be removed.

shalkya:
so i'm for merging this one. Tested it several time, don't any issue.
People with ecomanager won't benefit from this one, but that's ok.
If someone wanna review it can nice also.

KionX:
I think it's time to accept. This has already been verified more than 10 times.

shalkya:
@KionX while doing this video : https://www.youtube.com/watch?v=RN_3deEmk_k
i wasn't able to give energy from seconde 10 to seconde 16. And ally storage are low. I think it might be bugged.

KionX:
This code defines maxEnergy:

local EnergyRatio = brain:GetEconomyStoredRatio('ENERGY')
if EnergyRatio ~= 0 then
ArmyScore[index].resources.storage.maxEnergy = ArmyScore[index].resources.storage.storedEnergy / EnergyRatio
else
ArmyScore[index].resources.storage.maxEnergy = 0
end

This is to check the possibility of transferring:

local SentValue = scoreData.resources.storage['max'..resType] - scoreData.resources.storage['stored'..resType]
if SentValue <= 0 then return end

I have not found a better way.
The problem is potential division by 0.
If the target player has 0 in the storage, then resources cannot be transferred.
I think a good solution to this problem will adding a sim function to the engine that returns the capacity of the storage.

PR #2866, opened 2019-10-04
scu raw and ras rework

shalkya:
nerf BP : 56 to 20
buff raw BT : 14400 to 8400 (10400 for sera)
nerf ras BT : 8400 to 25000
buff sera mass/e prod from 2/200 to 3/300
nerf sera mass cost from 2050 to 2200
remove wreck for scu


Ithilis:
Remove wreck make sacu very dangerous. Are you sure about that? Compare to weaker T3 land its far better to build sacus, and tanks will be later in-game disappearing from the battlefield.

shalkya:
i understand your worries although we aim to make a preset of scu called preset "support" that will help making T3 armies still useful later in the game (chrono scus, regen field scus, bubble shields scus, stealth field scus).
Removing wreck will also make scu units that can be used in a more aggressive way. Balancing of their weapons is going to be required, especially since the T3 nerf patch. Finally scu will still not be as spammable (even if we want to make them more spammable than before to some extend) as T3 land. Goal is to have a late T3 stage made up of T3 armies in combo with support scu, and some rambo scu that can held front with important micro. Those rambo scu would be able to do some suicide mission to some extend (if the occasion appear) and not leave insane amount of reclaim behind.

Ithilis:
I remember a time when was reclaim removed from sacus, it was like 5 years in past, it took about one year to add it back, while late T3 change purely into Rambo sacu spam, especially UEF Rambo sacu spam what was horribly unfair and not fun at all. They break everything, and even after that dont get nothing. 

Yes there is lot of place to balance them and make support sacus is great way how to do it. Also i fully agree with remove shield on sera and nano from aeon, chono sacu  should be there from start, bubble shield sacu already exist.. (and none ever use it) but still missing reclaim on them will brink back this very old crucial problem that highjack main game mechanic. It logically can be on some very special autodestruct wannabe mobile nuke upgrade, but overall remove relciam is same as remove it from experimental -> in way to support agresion. It will bring only frustration to this who try to go back into the game but will gain nothing, only lose some defense, while there will be no reason for attacker to throw mass into the effort. 

The reason why game start be far more defensive lategame need aim on source and not gain new problems that will surely need solution later, while history learn as that it already happend.

Uveso:
Nooo please don't remove the wreckage of SCUs.
The game is ugly enough, don't remove more graphic.
Just to remember what we will lose:
WeWillMissYou
Maybe it's possible to reduce the wreckage mass, but leave the grafic ?

MerryPanda:
> nerf BP : 56 to 20
> buff raw BT : 14400 to 8400 (10400 for sera)
> nerf ras BT : 8400 to 25000

hello, could you pls clarify, what is raw, ras, BT and BT. thanks

MerryPanda:
> 
> 
> > nerf BP : 56 to 20
> > buff raw BT : 14400 to 8400 (10400 for sera)
> > nerf ras BT : 8400 to 25000
> 
> hello, could you pls clarify, what is raw BT, ras BT and BP. thanks

ok, i got what raw BT and ras BT stand for, but why BP (BuildRate, according to code), shouldn't it be BR?

MerryPanda:
well, in general it is quite controversial... why to make SCU more accessible? It is frequently abused already.

I consider SCU to be an elite T4 universal builder/producer/soldier, so it should not be easily amassed.

I would suggest increasing resource production for Sera SCU from 3/300 to 4/400 and restore previous BuildRate = 56, BuildTime = 14400 for all SCUs.

:D

shalkya:
BP stand for build power.
raw scu are nearly never built (thus reducing their BT). Increasing raw mass/e prod even more would be way to OP. scu without BP upgrade have too much BP already, thus why it is nerfed.

MerryPanda:
> 
> 
> BP stand for build power.
> raw scu are nearly never built (thus reducing their BT). Increasing raw mass/e prod even more would be way to OP. scu without BP upgrade have too much BP already, thus why it is nerfed.

looks like a typo in

shalkya:
thanks i will fix that

MerryPanda:
> 
> 
> BP stand for build power.
> raw scu are nearly never built (thus reducing their BT). Increasing raw mass/e prod even more would be way to OP. scu without BP upgrade have too much BP already, thus why it is nerfed.

But why to reduce raw BP? Will upgraded SCU buildTime be reduced as well (logically yes), so it means all other types (except RAS) will be built faster... I can't get your logic. SCUs are abused already, why make them even more affordable?

shalkya:
this is only a PR among a lot more reworking the upgrades completely. It cannot be look at on its own. I only pointed it to you, because it leads to a buff of sera scu resource production, which should fix the issue you opened.

PR #2873, opened 2019-10-08
fix air facto rebuild on wreck

### LINKED ISSUE #2386: air unit are being re-built from wreck
they get the 50% time decrease for construction, thx to the build from the wreck.
knowing that the wreck appear after 50% completion of the unit being killed.

this is for all air factory but : aeon T2 air HQ, aeon T3 air HQ, aeon T3 assist factory ; those one leave the wreck like land/naval facto.
[game.log](https://github.com/FAForever/fa/files/1754408/game.log)
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=7538250

shalkya:
if a wreck was created (unit killed after 50% completion) air facto would start construction of the same unit with 50% done already.

fix #2386

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2404,19 +2404,21 @@ Unit = Class(moho.unit_methods) {
         local wreckage = {}
         local bpid = unit:GetUnitId()
 
-        for _, p in props do
-            local pos = p.CachePosition
-            if p.IsWreckage and p.AssociatedBP == bpid and upos[1] == pos[1] and upos[3] == pos[3] then
-                local progress = p:GetFractionComplete() * 0.5
-                -- Set health according to how much is left of the wreck
-                unit:SetHealth(self, unit:GetMaxHealth() * progress)
+        if EntityCategoryContains(categories.ENGINEER, self) then
+            for _, p in props do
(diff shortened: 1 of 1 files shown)


Nojoke185:
Checked the PR
Best I can tell, this looks to be working as designed

Test ran:

1 - build following air factories:
- T1
- T2HQ
- T2Support
- T3HQ
- T3Support
2 - at each air factory build an air unit to above 50% completed
3 - before unit completed destroy it using enemy T1 bombers. A wreck is left
4 - rebuild same type of air unit at air factory
5 - ensure that the rebuild starts from 0% and not 50%
6 - let rebuild complete, verify unit is operational
7 - build another unit of same type
8 - ensure that the rebuild starts from 0% and not 50%
9 - let rebuild complete, verify unit is operational
10 - build other types of air unit, verify unit is operational
11 - verify no log errors


Test Results:
All units tested started at 0% for the rebuild
All units can be built to completion and function
Factory can cancel order or give new order and pause construction
Wreck is left and can be reclaimed
No log errors visible

Variations:
Test UEF, CYBRAN, SERA, AEON
Test a mix of different air units

Note:
Side effect of fix is that the part constructed unit leaves a wreck on air factory

PR #2874, opened 2019-10-09
Snipemode nerf

Strogoo:
Well, I'm still against this change :) But at the same time I agree that snipemode+shif-g comdo is too op. Removing collisions between ACU and t1 spam could also help (and seems pretty logical if you look at ACU and t1 size), but we need another engine hack from Kyryl for that, so yeah...

Anyway, for now only experimentals are allowed to use full snipemode/ACU priority from UI mod, others units use cut version without ACU at the top of the list. New limiter also adds ` -categories.Command` to every given priority, so you can't trick the system by combining allowed categories to achieve smth similar to `categories.Command` (or even simply pick "Engie" preset, because most of the time enemy ACU is the only "engineer" in range, especially in t1 stage)


shalkya:
nice i will test that. I just thought that it would still be nice to snipe unit to have the acu priority registered too.
so for example gunship, so you can dodge flak shot at the same time, but also T1 lab, so you put the priority before loading and then it focus the acu (T1 and T2 transport)

Strogoo:
Added `SNIPEMODE` category to unit blueprints (exps, acus, gunships and labs).

PR #2875, opened 2019-10-09
repair discount

shalkya:
make repair cost 75% of the price of the unit

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1003,6 +1003,7 @@ Unit = Class(moho.unit_methods) {
             local energy = 0
             local targetData
             local baseData
+            local repairRatio = 0.75
 
             if focus then -- Always inherit work status of focus
                 self:InheritWork(focus)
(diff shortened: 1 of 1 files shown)


Strogoo:
You have to check if unit is finished, otherwise it's possible to build stuff for 0.75 cost. There is `unit:GetFractionComplete()` call to engine, if it returns `1` then unit is finished. But it's better to just add some flag in `OnStopBeingBuilt` (less engine calls means better performance :)

Moving your code to  `-- Building/repairing something` section will save some cpu cycles as well.

https://github.com/Strogoo/fa/commit/c47d6d6449c2e4c6eef129db565301c4cf955ff0 smth like that ;)

Strogoo:
Still :)

Ithilis:
Also, check if not abusable while acu/sacu/shield/mex.. upgrading, while this unit is finished and still use build power that comes from repair. I was trying to make this work many time and all-time fail on this.

shalkya:
i don't get what you mean. But yes we will watch closely if people abuse it. For now we wanna make it more relevant (ie : allow to repair exp etc) but if we see people abusing by repairing pd in com drop or whatever.

Ithilis:
The abuse gameplay mechanic is okish, what i mean is to check _if it does not help upgrade units,_ that are already complete but build something on themself as ACU upgrade, of shields with 75% discount. 

On my opinion repair should be mass free as is air staging facilities and at least double fast as building, so even normal units repair will be rentable, while factory have lot of BP that enginers/mantis is not able to afford.

shalkya:
what you are describing could be abused by hives repairing exp/units in fights.

PR #2876, opened 2019-10-11
Pause facto effect fix

shalkya:
remove the building effect when the factory is paused

### DIFF
--- a/lua/EffectUtilities.lua
+++ b/lua/EffectUtilities.lua
@@ -557,35 +557,6 @@ function CreateAeonCommanderBuildingEffects(builder, unitBeingBuilt, BuildEffect
     end
 end
 
-function CreateAeonFactoryBuildingEffectsOnPause(builder, unitBeingBuilt, BuildBone, EffectsBag)
-    local bp = unitBeingBuilt:GetBlueprint()
-    local x, y, z = unpack(builder:GetPosition(BuildBone))
-    local mul = 1
-    local sx = bp.Physics.MeshExtentsX or bp.Footprint.SizeX * mul
-    local sz = bp.Physics.MeshExtentsZ or bp.Footprint.SizeZ * mul
-    local sy = bp.Physics.MeshExtentsY or sx + sz
-
-    local slice = nil
-
-    -- Create a pool mercury that slow draws into the build unit
-    local fraction = unitBeingBuilt:GetFractionComplete()
-    local scale
-    scale = 1 - math.pow(fraction, 2)
-    local BuildBaseEffect = unitBeingBuilt:CreateProjectile('/effects/entities/AeonBuildEffect/AeonBuildEffect01_proj.bp', 0, 0, 1, nil, nil, nil)
-    BuildBaseEffect:SetScale(sx * scale, 1.5 * sy * scale, sz * scale)
-    Warp(BuildBaseEffect, Vector(x, y - 0.05, z))
-    unitBeingBuilt.Trash:Add(BuildBaseEffect)
-    EffectsBag:Add(BuildBaseEffect)
-
-
-    local slider = CreateSlider(unitBeingBuilt, 0)
-    unitBeingBuilt.Trash:Add(slider)
-    EffectsBag:Add(slider)
-    slider:SetWorldUnits(true)
-    slider:SetSpeed(0)
-    slider:SetGoal(0, 0.5 * (fraction * sy - sy), 0)
-end
-
 function CreateAeonFactoryBuildingEffects(builder, unitBeingBuilt, BuildEffectBones, BuildBone, EffectsBag)
     local bp = unitBeingBuilt:GetBlueprint()
     local army = builder:GetArmy()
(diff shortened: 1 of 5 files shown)


shalkya:
forgot speed did the same work here : #2875 
ran into the same issue, will also look a bit into it.

shalkya:
Ok the last commit should fix the issues of fx for sera/aeon facto.
So this PR should be finished. Please test thoroughly.

shalkya:
Report written by Nojoke

> Hi this PR is fine, see below:

Tested all below mix of permutations

Factions:
Cybran
EUF
Sera
Aeon

Factory Types:
Land
Air
Navy
Gateway - no changes

Tech Levels:
T1
T2 HQ
T2 Support
T3 HQ
T3 Support



Verified that can pause the factory and the unit build animation stops
Verified that can unpause the factory and the unit build aninmation resumes
Verified that the pause/unpause animation works with a multiple selection of factories at different tech levels and Types (air/land/naval) and factions
Verified stop order stops the animation and clears the unit from factory
Verified right click on unit in factory production queue list, stops the build and stops the animation and clears the unit from factory
Verified right click on unit in factory build list, stops the build and stops the animation and clears the unit from factory
Verified that Aeon factory pause leaves the mercury pool visible
Verified no impact to gateways
Verified no log errors visible

Query:
Sera - black hole build effect goes away when factory is paused?
should it stay similar to Aeon mercury pool effect?
The part built unit stays visible so that part looks good.
I dont think this is necessarily wrong, it is just a query at this time

Other than that all working great

shalkya:
i think the black hole effect can be linked to the building effect, thus it doesn't need to be displayed on pause.
I could easily put it back in though. It's up for debate.
still need a review before merging.

speed2CZ:
This looks like a lot of code duplication. Im sure it could be simplified a lot.

With sera I would keep the black hole effect and just remove the beam from the factory. The same happens wheen building anything with engies and pausing them. The black hole stays, the beams goes away.

When the sera unit in the factory is almost finished and you pause/unpause the factory, the unit jumps a bit up and down.

shalkya:
i've also made a version (with much less code) that remove the fx bug sera (little teleportation when paused for big units) but then the unit keep on going down when paused (as it is right now, but without the beam).
Making the unit being built bone stop at good height when paused is way too difficult with the mechanism used by initial coders. I can only approximate it. (can post detail if necessary)

lua/EffectUtilities.lua
```diff
@@ -599,45 +570,62 @@ function CreateAeonFactoryBuildingEffects(builder, unitBeingBuilt, BuildEffectBo
...
+    end
     Warp(BuildBaseEffect, Vector(x, y - 0.05, z))
     unitBeingBuilt.Trash:Add(BuildBaseEffect)
     EffectsBag:Add(BuildBaseEffect)
 
-    CreateEmitterOnEntity(BuildBaseEffect, builder:GetArmy(), '/effects/emitters/aeon_being_built_ambient_02_emit.bp')
-    :SetEmitterCurveParam('X_POSITION_CURVE', 0, sx * 1.5)
-    :SetEmitterCurveParam('Z_POSITION_CURVE', 0, sz * 1.5)
+    if not builder:IsPaused() then
+        CreateEmitterOnEntity(BuildBaseEffect, builder:GetArmy(), '/effects/emitters/aeon_being_built_ambient_02_emit.bp')
```
speed2CZ: you have `army` in the variable on line 562, you don't need to call a function to get it again

lua/EffectUtilities.lua
```diff
@@ -599,45 +570,62 @@ function CreateAeonFactoryBuildingEffects(builder, unitBeingBuilt, BuildEffectBo
...
+        :SetEmitterCurveParam('X_POSITION_CURVE', 0, sx * 1.5)
+        :SetEmitterCurveParam('Z_POSITION_CURVE', 0, sz * 1.5)
 
-    CreateEmitterOnEntity(BuildBaseEffect, builder:GetArmy(), '/effects/emitters/aeon_being_built_ambient_03_emit.bp')
-    :ScaleEmitter((sx + sz) * 0.3)
+        CreateEmitterOnEntity(BuildBaseEffect, builder:GetArmy(), '/effects/emitters/aeon_being_built_ambient_03_emit.bp')
```
speed2CZ: same here with the army

lua/EffectUtilities.lua
```diff
@@ -751,62 +712,70 @@ function CreateSeraphimFactoryBuildingEffects(builder, unitBeingBuilt, BuildEffe
...
-
+    
     for _, vBone in BuildEffectBones do
         EffectsBag:Add(CreateAttachedEmitter(builder, vBone, army, '/effects/emitters/seraphim_build_01_emit.bp'))
         for _, vBeam in EffectTemplate.SeraphimBuildBeams01 do
-            EffectsBag:Add(AttachBeamEntityToEntity(builder, vBone, unitBeingBuilt, -1, army, vBeam))
+            if not builder:IsPaused() then
+                EffectsBag:Add(AttachBeamEntityToEntity(builder, vBone, unitBeingBuilt, -1, army, vBeam))
+            end
             EffectsBag:Add(CreateAttachedEmitter(unitBeingBuilt, -1, builder:GetArmy(), '/effects/emitters/seraphim_being_built_ambient_02_emit.bp'))
```
speed2CZ: This and next 3 lines, army is saved in local variable on line  709, so reuse that

speed2CZ:
I've noticed the mistake of not using the created `army` variable was there before. So it might be present in other functions in that file you haven't changed. It would be worth going though it and fixing them all. It's a small speed improvement, but every bit helps

shalkya:
tested and reviewed.

PR #2878, opened 2019-10-12
Refactor updateBuildRestrictions

Exotic-Retard:
now its cleaner, faster, and has better mod support.


shalkya:
tested works, need review before merging

PR #2881, opened 2019-10-12
Sera ACU regen field rework

Petricpwnz:
As per call changed the slots of regen field and ras around (I know there is an open PR for it but it seems to be missing a few lines of upgrade VFX code.)
Rebalanced to have a regen ceiling based on a tech level as opposed to a shared one.
Regen field ceilings:
* T1 - 10
* T2 - 15
* T3 - 25
* T4 - 40

Advanced regen field ceilings:
* T1 - 20
* T2 - 50
* T3 - 120
* T4 - 240

### DIFF
--- a/lua/sim/Buff.lua
+++ b/lua/sim/Buff.lua
@@ -367,6 +367,14 @@ function BuffCalculate(unit, buffName, affectType, initialVal, initialBool)
     local multsTotal = 0 -- Used only for regen buffs
     local bool = initialBool or false
     local floor = 0
+    local ceil = 0
+    -- Dynamic ceilings with fallback values for sera regen field
+    local ceilings = {
+        TECH1 = 10,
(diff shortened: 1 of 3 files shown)


Ithilis:
It sounds good and solve one problem, but think about switch it with T2/T3 position, so you can get army support with gun upgrade. 

Main problem of regen field is that you cant have gun, and you don't want to run acu in battlefield whiteout proper range.

Petricpwnz:
> It sounds good and solve one problem, but think about switch it with T2/T3 position, so you can get army support with gun upgrade.
> 
> Main problem of regen field is that you cant have gun, and you don't want to run acu in battlefield whiteout proper range.

Having gun and regen together is the entire point of this. But I'm not going to discuss it here, it was discussed in the balance call and im not the one to ultimately decide on it.

PR #2884, opened 2019-10-14
Supreme Props: Added missing fix

badbwoi:
I missed one fix in the Supreme Props integration. The 10 energy of the boulder/rock should be removed. The value should be 0 like all the other rocks.

See here: https://user-images.githubusercontent.com/28671151/57193683-2e7b5180-6f3e-11e9-9e24-cf2af24384e7.png


shalkya:
tested works.

Issue #2447, opened 2018-04-28
Observer abuse


Apofenas:
Hello

If observers are on you can switch your view to any army. This lets you spot pretty much everything for your team. Amount of resources, nuke location, teleport ect.

Though this will not prevent the abuse, but still make it harder by making observers unable to switch armies.

shalkya:
it is a useful tool for observer and caster, to be able to get into player's view and see their income/reclaim and other stuff. (especially in 1v1)
removing this would be more annoying than useful.
Not allowing observer at first is the best way to prevent the "cheating". 
Best solution, would be to make dead players something else than an observers, and thus making them unable to get into opponent team view.

PhilipJFryFAF:
maybe dead ppl should stay the same regardless of observers being enabled or not?

that way ppl who are actual observers would still be able to see stuff while dead players would not

Issue #2259, opened 2017-09-24
Observers can change armyview


PhilipJFryFAF:
Observers can use the "setarmyview" command to go from obs to an army view and then write messages and set map markers
that should not be possible since observers can see the entire map and thus can give vital information that might not be available otherwise

speed2CZ:
Each time when you send message or set marker in different army view the game says you're trying to cheat but doesn't stop it.

aeoncleanse:
Phil, could this be a side effect of sharedarmies exe hack?

PhilipJFryFAF:
not sure
speed2 doesn't have that exe but i do
gonna test again with the regular exe

PhilipJFryFAF:
tested again and same issue
doesn't seem to be related to the exe

aeoncleanse:
I can't find a command called setarmyview. Sure you don't mean SetFocusArmy?

speed2CZ:
yeah that one

shalkya:
just need to use the cheat panel and get into the player view. Then you can ping, and use ping (with writing).
it's not related to sharearmy.exe

Issue #2141, opened 2017-08-08
obs in game


shalkya:
is it possible to create a new way to treat observer ?
the issue we have is that if we enable observer, to let some person spectate a game, the players that will die, will also see the whole map.
it would be nice to separate the 2 features.
Make one that allow observer in game, but doesn't allow player to see whole map after death.
Make one that allow player to see the whole map after death.

MrNukealizer:
I don't think that will work. It would be possible to not automatically make someone an observer when they die, but if observers are enabled they could still type `SetFocusArmy -1` in the console to see everything.

CookieNoob:
duplicate of #1752

duk3luk3:
> but if observers are enabled they could still type `SetFocusArmy -1` in the console to see everything.

And we couldn't detect that and kick them out?

CookieNoob:
maybe we can just remove the console function `SetFocusArmy( )` for this lobby option?

MrNukealizer:
You can't really remove the console function because any UI mod could bring it back. Now that I think of it, it would be possible to add something to the sim function that's called when someone changes focus that would kick them if they started as a player but changed to -1.

CookieNoob:
we could just fork a thread from the acu "OnKilled" function which is attached to the AIBrain of the player and checks every second if GetFocusArmy is still -1

MrNukealizer:
There's no need to check the focus army every second. `/schook/lua/SimSync.lua` has the function `NoteFocusArmyChanged` which is called when it changes.

Issue #1752, opened 2016-12-28
When observers are enabled, dead players goes to observers.


Tokyto:
Making observer feature unsusable for 2v2+ Tournaments streams.

aeoncleanse:
This is the point of the option. I have no idea what your point is

Tokyto:
The observer option should only affect observers, not players. For example, a live-streamer joins a 2v2 game to stream it, but gets kicked becouse the option is broken and dead people see whole map when they die, giving free intel to his teamnate.

micheljung:
And the stream does not give free intel?

Tokyto:
the stream can have a delay

Tokyto:
This is mostly about caster+co-casters being in sync and people being able to use this feature in teamgames.

Exotic-Retard:
livereplays for example have a 5 minute delay to prevent ghosting. when streaming a tourney you can get a 40s delay or so but point is that there should be a difference from observers who were there from the lobby and those who became observers after being killed in the game.

aeoncleanse:
Unless I'm going insane, dying with Observers on sends you to a state where you can only see what your old team could.. You can act as an intel machine calling out stuff, but you don't get full-map vision... :/

Exotic-Retard:
you are going insane i guess?
thats with obervers off, pretty sure anyway.

aeoncleanse:
Right. Well, not sure what we can do about this... Or what people want even. New observer type???

Tokyto:
If player dies, he doesn't go into observers mode. That's it i think.

TheKeyBlue:
I think that we should add an option that says if dead players go to observers or if they stay in their own team view (default).
This way people can decide what they want.
Could be usefull for streamers like BRNK who maybe prefer to see the whole game after death instead of only his own team. (don't know what he prefers, but why not give him/others the possibility to chose?)

Tokyto:
And FFA games could use this feature (to go into obs after death).

duk3luk3:
The engine seems to be involved in this: The `SessionIsObservingAllowed` engine function can only be true or false (must be based on whether `gameInfo.GameOptions.AllowObservers` is set at init time?).

Then the ui module [`gameresult.lua`](https://github.com/FAForever/fa/blob/develop/lua/ui/game/gameresult.lua#L52-L54) switches users to observer when `SessionIsObservingAllowed` is true.

So since it's in the UI, it could be subverted by players.

[...]

PR #2885, opened 2019-10-15
Removed observing after death

KionX:
Issues: #2447, #2259, #2141, #1752

Used with "hook_ObserverFix.cpp" from https://github.com/FAETHER/FA-Binary-Patches/pull/8
Can not be accept separately.

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -44,30 +44,6 @@ local SimTriggers = import('/lua/scenariotriggers.lua')
 local SUtils = import('/lua/ai/sorianutilities.lua')
 local LetterArray = { ["Aeon"] = "ua", ["UEF"] = "ue", ["Cybran"] = "ur", ["Seraphim"] = "xs" }
 
-Callbacks.SetLocalCommandSource = function(data)
-    if (not ScenarioInfo.Options.AllowObservers) or SessionIsReplay() then return end
-    local humans = {}
-    local humanIndex = 0
-    for _, data in ScenarioInfo.ArmySetup do
-        if data.Human then
-            if IsAlly(GetFocusArmy(), data.ArmyIndex) then
-                if not ArmyIsOutOfGame(data.ArmyIndex) then
-                    return
-                end
-                table.insert(humans, humanIndex)
-            end
-            humanIndex = humanIndex + 1
-        end
-    end
-    
-    for _, index in humans do
-        for _, data in ScenarioInfo.ArmySetup do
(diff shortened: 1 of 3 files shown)


shalkya:
so this would allow people dying to only see the what allies can see, even if observers are allowed ?
If that's the case, i see benefit in tournament when we need to allow observers on to let streamer in.
But i think the drawback would be in FFA for example, when people want to see what's happening in the game when they died. (if i'm right would it be possible to make people see the whole map in case their team has been defeated ?).

shalkya:
didn't seems to work : 
I hosted a game with an ai ally vs an ai. Observers were on. When i ctrl k i can see whole map even if ally is still alive.

KionX:
This only works if the ally is human.

shalkya:
doesn't work with only human in the game

KionX:
When playing with players, I did not find any errors

shalkya:
image

KionX:
Sandbox mode should be disabled

KionX:
I can not reproduce the error, I need more details

shalkya:
it seems to be related to some ui mods. I will investigate it more in detail later.

shalkya:
ok so i figured out there is a conflict with supreme score board, which make allow people having the mod to see the whole map as usual when they die.

shalkya:
so i understand this is a bandage fix while waiting for engine fix.
issue is that this bandage/fix can be dodged by doing alt-f2 and switching view (so intentional decision) or by having ui mods like SSB, which is a quite common one (unintentional abuse).
This introduce inconsistency and can give advantages to a team (that know how to abuse or have the SSB mod) and thus be game breaking. 
I'm for closing it.

KionX:
I have already created a engine patch and will update PR soon.

KionX:
Updated to full-featured version.

KionX:
Finally tested version:)

shalkya:
from what i understood this change is completely different from the exe one, and the exe change doesn't require this one. Correct me if i'm wrong.
I tested once again, this PR gives an unfair advantage to player using SSB. I won't merge a PR that can lead to unfair (and important) advantages.

KionX:
The description says what is required https://github.com/FAETHER/FA-Binary-Patches/pull/8

shalkya:
image

KionX:
Modification exe allows you to control the behavior of observers from lua. Therefore, changes to lua are necessary.

shalkya:
tested by strogo.

PR #2887, opened 2019-10-17
lobby algorithm that finds the fairest teams

ageekhere:
Very new to this (also first time coding in lua), so this is a work in progress (also I have not set up the lobby environment so i cannot test the lua code)

This algorithm finds the fairest teams for the lobby

A button will need to be added to the lobby
also the lobby would need to be updated as well
ideley this script should be ran in a new thread

The idea for this script is to search all player combinations and store the fairest team with the highest computeQuality,

For large teams >11 players, the host should be able to stop the search once an acceptable Quality has been found.

Maybe running multiple instances in threads could speed up the search?

The ideal situation is in a 10 player map the host lets the room fill, once full the host presses the calculation button. Calculation finds the best Quality and resorts all the players in the lobby to the correct teams. Now the players can choose their team positions on the map to make it more fair.

### DIFF
--- a/lua/fairestTeams
+++ b/lua/fairestTeams
@@ -1,0 +1,55 @@
+--need a button for f_fairestTeam()
+--NOTE may need to run in a new thread
+local Trueskill = import('/lua/ui/lobby/trueskill.lua')
+function f_fairestTeam()
+	local playerArray = {} -- human players
+	local arrayStart = 0; --start position of the player array
+	local arrayEnd = 0; --the length of the player array
+	local fairestTeam = nil -- the teams with the highest computeQuality
+	local players = GetPlayerCount() + 1 -- number of players
+	local lastScore = 0 -- the last computeQuality
+		for i = 0, players do --adding players to the player array
+			if gameInfo.PlayerOptions[i].Human then --check if player is human
+				playerArray[i] = gameInfo.PlayerOptions[i] --add player options to playerArray
+			end --if
+		end --for
+	f_mutate(playerArray,arrayStart,i,lastScore)
+	
+	function f_mutate(playerArray,arrayStart,arrayEnd,lastScore)
+		local currentComputeQuality = 0 --the computeQuality for this calculation
+		if arrayStart == arrayEnd then 
+			local teams = nil --reset the teams
+			teams = Teams.create() -- create new teams	
+			local currentPosition = arrayEnd + 1 --update array position
+			for i = 0, currentPosition do
+				if i < (currentPosition * 0.5) then -- sort players into teams
+					teams:addPlayer(0, playerArray[i]) -- team 1
+				else
+					teams:addPlayer(1, playerArray[i]) -- team 2
+				end	--else
+			end --for
+			
+			local currentComputeQuality = Trueskill.computeQuality(teams) --get current Compute Quality from Trueskill
+			if currentComputeQuality > lastScore then -- check if the current Quality is better then the last Quality
+				lastScore = currentComputeQuality --update new Quality
+				fairestTeam = teams --set the fairest team
+			end --if
+			return
+		end --end if
+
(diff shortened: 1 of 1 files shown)


anihilnine:
Hey @ageekhere what do you think of this:
https://forums.faforever.com/viewtopic.php?f=42&t=18354&p=179697#p179697

ageekhere:
> Hey @ageekhere what do you think of this:
> https://forums.faforever.com/viewtopic.php?f=42&t=18354&p=179697#p179697

I think this method should be used when there is more than 8 or 10 players, because Brute force takes too long when there is more than 10 players.

However Brute force does find the perfect game quality.

So both I think are needed

anihilnine:
how about hybrid?
first assign all players > 8 via my method. 
then brute force the rest

ageekhere:
Not sure if that works, it is either one or the other

anihilnine:
how come? using mine for first few units removes half your permutations. it's not as perfect as using pure brute force, but it's more perfect than using pure-mine

ageekhere:
well brute force can be used with about 10 players. So if a hybrid method works after 10 players then cool.

shalkya:
@ageekhere you don't take into account the deviation do you ?

ageekhere:
Was using trueskill.computeQuality(team) to get the best score

ageekhere:
anihilnine's code needs to be added into the fairestTeams file so if there is more than 8 players it uses the fast, but less accurate method

TobiasJacob:
Hello, 

in fact, I think there is an algorithm that solves the problem faster in O(n log n) steps. The problem is related to the [Knapsack-Problem](https://en.wikipedia.org/wiki/Knapsack_problem) (which is NP-hard), however, there is an important difference: Each player always uses one slot ($w=1$).

So the algorithm to find the best team balance goes like this. Put all players in a list. Then pick the highest-rated player out of that list and put him into the team, which currently has a lower rating. Repeat this, until the list is empty.

Implementation:
Put all players in a sorted list O(n log n): [1200, 900, 900, 200]
Go through that list O(n):
1: Put the 1200 player into team A. Team A has 1200 points and Team B 0.
2: Put the 900 player into team B. Team A has 1200 points and Team B 900.
3: Put the 900 player into team B. Team A has 1200 points and Team B 1800.
4: Put the 200 player into team A. Team A has 1400 points and Team B 1800.
I guess this is the best split-up possible.

One more exception: If one team is full, add the remaining players to the other team. This can happen in extremely unbalanced cases, like [1200, 100, 100, 100, 100, 100]. The best solution is Team A: [1200], Team B: [100, 100, 100, 100, 100], but if there is the constrain that each team has to have three players, then it's Team A: [1200, 100, 100] and TeamB: [100, 100, 100].

ageekhere:
@TobiasJacob 
While much fastest it does not always find the fairest teams
Example
players
1252
1023
989
821
626
599
599
342

Your example
Team A
1252
821
599
342
Team B
1023
989
626
599
Difference 223

Where the fairest team is
Team A
1252
626
599
599
Team B
1023
989
821
342
Difference 99

However once you start having more than 10 players then the faster method has to be used.

TobiasJacob:
Hello @ageekhere,

you are right, sorry, my mistake. I was trying to think about examples where it won't work but was unable to come up with one. I think a hybrid approach will work well.

Garanas:
I like the ideas in this PR and the implementation that is already present, but I'm not sure whether it has aged at this point.

ageekhere:
I think it really needed someone more skilled in this area to take the ideas and impermeant something.

Garanas:
Thank you for your response.

I'm closing this PR in favor of https://github.com/FAForever/fa/pull/2903 . The code of this PR is in that PR too. I'll make sure to credit you accordingly for your work here when that PR is merged.

PR #2890, opened 2019-10-24
make selection of exp under construction act as building

shalkya:
this commit aim to fix mega under construction selection (doesn't achieve it)
it aim to use this code https://github.com/FAForever/fa/blob/develop/lua/ui/game/gamemain.lua#L453

by using the same trick as this one : https://github.com/FAForever/fa/blob/develop/units/XSL0101/XSL0101_script.lua#L37

@aeoncleanse you should be able to help me on this one.


shalkya:
#2942 fixes the issue

Issue #2238, opened 2017-09-02
differenciate the different idle unit from their faction


shalkya:
when you capture or a teammate gave you unit from another faction, would be great that you could differentiate engie/sacu/facto from the different faction in the idle list (top right)
![image](https://user-images.githubusercontent.com/11618427/29996783-6744455e-9005-11e7-8650-74b03891dc63.png)

Uveso:
Well, if we have an icon for every faction engineer, then we have 5x buttons for every tier and 1x sub commanders.
6*5 = 30 buttons for idle engineers ?
I guess this is not what you mean.

Can you explain how this should work exactly?

shalkya:
well first maybe don't make appear the one that are not needed. (ex : if you have 2 faction under your control, show only the icon for this 2 factions, or if there is only 2 faction in the game, show only the icon for this 2 factions)
then there is 4 tier of engie, 4 faction (no need to add nomads) makes 16 buttons for engie. So at max your can have 16 icon for engie. maybe we can have feature to hide some faction/tier manually.

Uveso:
There are plans to implement Nomads into FAF main game. So we maybe have this problem later.

And yes, normally you have only 1 additional faction and the menu will not be too big in case its adjusting its size and only displays factions you own. 

Sounds good.

Exotic-Retard:
it might make sense to add another "level" of expansion to the ui for selecting idle things: 
https://puu.sh/xqrjR/f8580dacfe.png
probably not copy exactly whats shown on here, and pretend that the second thing shows the engies of the different factions but you get the point i guess.

shalkya:
yeah that's what i was thinking of more or less

PR #2902, opened 2019-11-14
Separate strategic icon for SACUs

Strogoo:
Just a copy of ACU icon with different name. Don't want to spend any time on making icon that is good enough for main FAF patch, so let's leave this work for modders.


shalkya:
fixes #2238

shalkya:
image
i'm trouble seeing what it fixes.

Strogoo:
It doesn't fix anything, only adds additional sacu icon so it can be changed by mods like Advanced Strategic Icons later. Currently ACU and sacu use same icon (icon_commander_generic), so spotting enemy acu at lategame stage is a pain ;)

pwilkutowski:
> Just a copy of ACU icon with different name. Don't want to spend any time on making icon that is good enough for main FAF patch, so let's leave this work for modders.

for the love of good THANK YOU FOR THIS FIX 
Finally bad eye sight ppl will be able to snipe acu in map with lot's of sacu

PR #2907, opened 2019-11-17
Corrects text display if video doesn't play

KionX:
To reproduce the problem, delete the movies folder and launch the scenario.

As a consequence of this fix if video dialogue is broken, text dialogue is show.
It not adds desyncs and not eliminate them.
Broken dialogue video

History:
The game only plays unpacked videos.
Videos from gamedata are found but not played. The sound also does not play.
Problem in any SC1 scenario. For example Joust.
After switching to Win 10 and reinstalling FAF, the problem disappeared.

PR #2908, opened 2019-11-18
Now impossible destroyed ACU via Ctrl+K without cheats

KionX:
Also fixed displayed announcement without goalControl.

### DIFF
--- a/lua/ui/game/announcement.lua
+++ b/lua/ui/game/announcement.lua
@@ -1,0 +1,186 @@
+--*****************************************************************************
+--* File: lua/modules/ui/game/announcement.lua
+--* Author: Ted Snook
+--* Summary: Announcement UI for sending general messages to the user
+--*
+--* Copyright © 2007 Gas Powered Games, Inc.  All rights reserved.
+--*****************************************************************************
+
+local UIUtil = import('/lua/ui/uiutil.lua')
+local LayoutHelpers = import('/lua/maui/layouthelpers.lua')
+local Group = import('/lua/maui/group.lua').Group
+local Bitmap = import('/lua/maui/bitmap.lua').Bitmap
+
+local bg = false
+
+function CreateAnnouncement(text, goalControl, secondaryText, onFinished)
+    local scoreDlg = import('/lua/ui/dialogs/score.lua')
+    if scoreDlg.dialog then
+        if onFinished then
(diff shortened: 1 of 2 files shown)


shalkya:
why would you implement this ?
Or did i miss-understand you ?

KionX:
Suicide ACU maybe use for foul play, or ACU maybe accidentally destroyed.
In a normal game, you do not need to destroy ACU.

speed2CZ:
You need in in a normal game. Sometimes is better to self destruct to kill your opponent when they are on low health since you would die anyway and you gotta do that before they are out of range.

KionX:
It depends on what is more important to you.

Also, please pay attention to other PR: #2907, [Common Army for Coop](https://github.com/FAForever/fa-coop/pull/88), [Fixed save Timed Expansion lobby option](https://github.com/FAForever/fa-coop/pull/87), [Delete duplicate code in PlayerDeath function](https://github.com/FAForever/faf-coop-maps/pull/344).

shalkya:
i would consider merging this pr, if you can change it in order to remove the ctrl k for ACU only when traitor game condition is on.
otherwise it's a no-go for me.

PhilipJFryFAF:
even with traitors on there can be situations where you want to take someone down via ctrl k i suppose
i'm strongly against this change

KionX:
I know that the only option when it is necessary to use Ctrl + K is when the enemy escapes in transport, in other cases it is enough to be near him.

shalkya:
been thinking of it, can't you make it so if someone ctrl k, it gives the army to the strongest opponent ?

KionX:
> if someone ctrl k, it gives the army to the strongest opponent

This will allow specially give army to enemy.

shalkya:
I've got another input from another moderator that doesn't support this change.
I will close this PR unless you have another idea on how to fix the issue without disabling ctrl k on ACU.

KionX:
Are there any penalties for Ctrl K on ACU?

shalkya:
what do you mean by penalties ?
It is completely allowed and even necessary to pull off to get a draw when an opponent is dropping a heavily upgraded acu next to your own acu or when you are getting blocked and you will quit the explosion range for a draw.

KionX:
I think suicide should be fined.

Issue #1328, opened 2016-05-30
  Salem's 'walk on land' button


shalkya:
http://forums.faforever.com/viewtopic.php?f=42&p=127646#p127646

Uveso:
i was trying it, but i failed.

Salem has the MotionType AmphibiousFloating. So the SupCom.exe is in control for the movements and restrictions.
Waiting for an layer change event is to late for any action, because the layer is already changed then.

To prevent the Salem to enter Land i tried to change the motiontypes Physics.MotionType = 'RULEUMT_AmphibiousFloating' and AltMotionType = 'RULEUMT_Water'

This works, but this is stored inside the blueprint and i cant change blueprints after the unit is build.

I tried this of course, but changing the blueprint don't has a affect to an already build unit.
I guess the motion type is stored inside the c-object of the unit.

Well, i dont see a way to do this.

Crotalus:
A really ugly solution would be to respawn the unit without AltMotionType set if you click on the "no land walking" button. The downsides to this are massive though, units would lose targeting of it etc :)

JJ173:
Probably a dead issue (2 years later almost). As it's still open, let me have a fiddle with this and see if I can produce any other ugly or smooth solutions for this. Since it's not a bug, it'll remain on low priority, but I'll give it a go at some point.

Uveso:
The problem is still, that the layerchange event is triggered from the c-engine.
If you get the event, then the salem is already on land.

You have to move it back to water.

Thats all possible but i can't manage the AI behavior in this situation.
You have to disband the single unit from the attackplatton, then move it back to water.
Now you have the problem, if you add the unit back to the attackplatoon it will move on land instantly.
And you can't just let the unit where it is, because the AI will not move it if it has no platoon.

Please don't make a player only solution and have the AI stuck in this situation!

PR #2909, opened 2019-11-19
Amphibious mode toggle for Salem

Strogoo:
Requires new exe (https://github.com/FAETHER/FA-Binary-Patches/pull/5)

Some technicals details is here -> `engine/Patches.lua`.

~~In the future all balance/script changes to `urs0201` should be also applied to `urs0204`. Only real difference between them is MotionType (RULEUMT_AmphibiousFloating and RULEUMT_Water)~~
bp copying is done via assembly now

image

#1328


KionX:
Is there a guarantee that the relative address of the blueprint will not change?

Better to make a separate function.
Add a selection of blueprint by name.

Strogoo:
Well, with current bp names the relative adress will always be the same. urs0201 + 3 blueprints (and 3 blueprints = 1200h). But I actually don't like the idea with dummy bp in lua, so I'll remake it in assembly I think. A simple memcpy that copy urs0201 to our new exxt section (thanks to Kyryll) and then changing MotionType there and switching unit's bp pointer to our new dummy blueprint.

As for new function, I have no idea how to implement this and don't want to spend time on it. Sim version of unit:GetStat already does everything that we need + it's not used anywhere in lua code and we are free to do anything we want with it. Have to mention that it's still 100% functional after the patch and you can still get unit:GetStat("KILLS") etc.

KionX:
Mismatch of name and functionality is bad practice.
I have experience adding a sim version of SetFocusArmy, it was not difficult.

units/URS0204/URS0204_script.lua
```diff
@@ -0,0 +1,192 @@
...
+            if self:GetCurrentLayer() ~= 'Land' then
+                self:GetStat("h1_SetSalemAmph", 0)
+            else
+                self:SetScriptBit('RULEUTC_WeaponToggle', true)
+            end  
+        end     
+    end,
+}
+
+TypeClass = URS0201
```
norraxx: Do we really need to copy all that code, or there is the way to add on existing definition of `URS0201` bit methods?

norraxx: Shouldn't here be different number `URS0201` -> `URS0204`? Like file is named different too.

Strogoo: yep. Anyway, I'll try to get rid of this dummy `urs0204` bp and do all stuff in assembly

shalkya:
ok i tested works pretty well. Some side cases where you toggle off in shallow water, will lead to destro still crossing on land. Apart from this i didn't notice anything wrong.
if someone wants to review/test a bit more, he is welcomed. I will merge it in few days.

shalkya:
tested by nojoke

PR #2910, opened 2019-11-22
buff tmd vision

shalkya:
allow tmd to shoot at their full range

could be done only for uef at least to make up for being slow


shalkya:
doesn't fix the issue it was aiming at.

PR #2921, opened 2019-12-14
In-game weapon tooltips fix

wodzu93:
Screenshot
Fixes broken in-game weapon tooltips for units by doing the following:
1) Removes merging entries for multiple weapon mounts of the same name, as multiplier indicating this was put in the wrong entry.
2) Makes beam weapons count all damage ticks correctly
3) Makes weapons that fire multiple projectiles per volley count them correctly
4) Added "Damage Radius" to all weapon tooltips. (DMG, Range, DPS, Range, Radius)
5) Added death weapons to weapon tooltips.

These changes have been implemented in Total Mayhem Lite mod as of writing and tested to be working and correct.

### DIFF
--- a/units/DEL0204/DEL0204_Script.lua
+++ b/units/DEL0204/DEL0204_Script.lua
@@ -1,0 +1,34 @@
+local TWalkingLandUnit = import('/lua/terranunits.lua').TWalkingLandUnit
+local TWeapons = import('/lua/terranweapons.lua')
+local TDFPlasmaCannonWeapon = TWeapons.TDFPlasmaCannonWeapon
+local TIFFragLauncherWeapon = TWeapons.TDFFragmentationGrenadeLauncherWeapon
+local EffectUtils = import('/lua/effectutilities.lua')
+local Effects = import('/lua/effecttemplates.lua')
+
+DEL0204 = Class(TWalkingLandUnit) 
+{
+    Weapons = {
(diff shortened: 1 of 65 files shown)


HUSSAR-mtrela:
Thanks for trying to fix n-game tooltips. However, this is not correct way to correct these weapon stats because the game does not use weapon.ProjectilesPerOnFire at all. Also, we should not change ProjectilesPerOnFire in blueprints. I developed much more accurate method for calculating weapon stats and I'm adding configuration options which were requested by FAF community.
http://forums.faforever.com/viewtopic.php?f=42&t=16716&p=167783#p167783

wodzu93:
@HUSSAR-mtrela - Wow, that looks really nice. It's been a long time though - that post is more than a year old, and latest edit is 5 months old. Is that still being worked on, and when we can expect that to be merged into FAF if it is?

Yes, I know "ProjectilesPerOnFire" line has no gameplay effect. But what's the harm in modifying this then? Since it is only used in the tooltips at the moment.

HUSSAR-mtrela:
yes, it is old forum post and I'm still working on it in my free time. I should be done with it in 1 month or so. 

Changing ProjectilesPerOnFire will introduce confusion and really we should be removing it instead of adding it to other blueprints.

wodzu93:
Okay, fair enough. I'll close the pull request then.

PR #2922, opened 2019-12-15
Fix acu projectiles

shalkya:
- add yellow dot for uef/sera acu projectiles when zoomed out
- make the aeon/cybran acu projectiles fx appear until 160 lod


speed2CZ:
When adding new files, 1 commit with original files and then other commits with changes, so its possible to see what got changed

shalkya:
ah yeah completely forgot about that, will redo it

shalkya:
tested by nojoke

PR #2928, opened 2019-12-18
Adding shied to czar

TurinturambarFAF:


KionX:
I don't like it.
Want the possibility of carrying Experimental and reclaiming.

shalkya:
not liking it is a bit little for a discussion. 
transport exp is in nomads, auto reclaim is a feature of mega ?
also you didn't have any input as of why we are doing this change, and this was already accepted, we just delayed it because we weren't sure if it would be proper for the pre-tourney patch.

PR #2932, opened 2019-12-24
Improved convenience when using the unit creation menu

KionX:
Units spawn in the cursor position after pressing the left mouse button, which allows you to select an arbitrary point of spawning on the map.

The menu can be closed with the "Esc" key.


shalkya:
tested works well.

speed2CZ:
works fine

PR #2934, opened 2019-12-27
Death weapon balance

shalkya:
improve DeathWeapon of experimental and rebalance them

### DIFF
--- a/units/URL0402/URL0402_unit.bp
+++ b/units/URL0402/URL0402_unit.bp
@@ -781,11 +781,11 @@ UnitBlueprint {
         {
             Damage = 4000,
             DamageFriendly = true,
-            DamageRadius = 7,
+            DamageRadius = 6,
             DamageType = 'DeathExplosion',
             DisplayName = 'Death Weapon',
             DummyWeapon = true,
-            Label = 'DeathWeapon',
+            Label = 'SpiderDeath',
             WeaponCategory = 'Death',
         },
     },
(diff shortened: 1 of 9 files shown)


KionX:
`position[3] = position[3]+5`

How is unit rotation taken into account here?

shalkya:
nicely spotted thx. Please test that to be sure it's working as intended.

KionX:
Checked the colossus. Works correctly.
Verification Method: DrawCircle (position, 3, '0000ff')

It's not needed
local current_yaw = math.floor (a * (180 / math.pi) + 0.5)
Enough
position [3] = position [3] + 5 * math.cos(a)
position [1] = position [1] + 5 * math.sin(a)

units/URL0402/URL0402_unit.bp
```diff
@@ -781,11 +781,11 @@ UnitBlueprint {
         {
             Damage = 4000,
             DamageFriendly = true,
-            DamageRadius = 7,
+            DamageRadius = 6,
             DamageType = 'DeathExplosion',
             DisplayName = 'Death Weapon',
             DummyWeapon = true,
-            Label = 'DeathWeapon',
+            Label = 'SpiderDeath',
```
Crotalus: Make sure this rename doesn't cause side effects, if I remember correctly there is some checks if label == 'DeathWeapon' in lua/sim/Unit.lua

shalkya: will check

shalkya: looked into it, it shouldn't cause any issue.

PR #2935, opened 2019-12-30
Mounting ui textures from mod directories

wodzu93:
Right now, mods cannot change some ui textures, as game is hardcoded to read them from "/textures/ui" directory. This particularily impacts ability to change or add strategic icons. The only way to do it is to pack them into an .scd and drop it into gamedata folder - which means installing mods from the mod vault requires additional manual step. This code change fixes that.

Reference to the forum post:
https://forums.faforever.com/viewtopic.php?f=45&t=18453#

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -150,8 +150,8 @@ mount_map_dir(SHGetFolderPath('PERSONAL') .. 'My Games\\Gas Powered Games\\Supre
 mount_map_dir(InitFileDir .. '\\..\\user\\My Games\\Gas Powered Games\\Supreme Commander Forged Alliance\\maps\\', '**', '/maps')
 
 -- Begin mod mounting section
--- This section mounts sounds from the mods directory to allow mods to add custom sounds to the game
-function mount_mod_sounds(MODFOLDER)
+-- This section mounts sounds and ui textures from the mods directory to allow mods to add custom sounds and textures to the game
+function mount_mod_content(MODFOLDER)
     -- searching for mods inside the modfolder
     for _,mod in io.dir(MODFOLDER..'\\*.*') do
         -- do we have a true directory ?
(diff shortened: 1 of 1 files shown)


init.lua
```diff
@@ -190,8 +179,8 @@ local function mount_strategic_icons(MODFOLDER)
         end
     end
 end
-mount_strategic_icons(SHGetFolderPath('PERSONAL') .. 'My Games\\Gas Powered Games\\Supreme Commander Forged Alliance\\mods')
-mount_strategic_icons(InitFileDir .. '\\..\\user\\My Games\\Gas Powered Games\\Supreme Commander Forged Alliance\\mods')
+mount_mod_sounds(SHGetFolderPath('PERSONAL') .. 'My Games\\Gas Powered Games\\Supreme Commander Forged Alliance\\mods')
```
speed2CZ: Please rename the function, since it doesnt load only sounds anymore. `mod_mod_countent` or anything that describes it better

init.lua
```diff
@@ -164,19 +164,8 @@ function mount_mod_sounds(MODFOLDER)
                     mount_dir(MODFOLDER..'\\'..mod..'\\sounds', '/sounds')
                     break
```
speed2CZ: If the functions finds `sounds` folder it will break the loop and never load the textures. So at this point you want to go through all the folders and check them

init.lua
```diff
@@ -164,19 +164,8 @@ function mount_mod_sounds(MODFOLDER)
...
-
--- This code mounts ui textures from mods as root, so things like strategic icons can be modified directly from the mod folders.
--- Workaround for dropping an .scd into gamedata folder, so mod installation can be fully automatic.
-local function mount_strategic_icons(MODFOLDER)
-    for _,mod in io.dir( MODFOLDER..'\\*.*') do
-        if mod != '.' and mod != '..' then
-            for _,folder in io.dir(MODFOLDER..'\\'..mod..'\\*.*') do
+                -- This code mounts ui textures from mods as root, so things like strategic icons can be modified directly from the mod folders.
+                -- Workaround for dropping an .scd into gamedata folder, so mod installation can be fully automatic.
                 if folder == 'textures' then
```
speed2CZ: So this can be changed to elseif and joined to the `if` condition above. And also removing `break`

init.lua
shalkya: i think you meant ```mount_mod_content```

init.lua
```diff
@@ -162,25 +162,21 @@ function mount_mod_sounds(MODFOLDER)
...
-                -- This code mounts ui textures from mods as root, so things like strategic icons can be modified directly from the mod folders.
-                -- Workaround for dropping an .scd into gamedata folder, so mod installation can be fully automatic.
-                if folder == 'textures' then
+                -- mount ui textures if there are any
+                elseif folder == 'textures' then
                     for _,folder in io.dir(MODFOLDER..'\\'..mod..'\\textures\\*.*') do
                         if folder == 'ui' then
                           LOG('Found mod icons in: '..mod)
                           mount_dir(MODFOLDER..'\\'..mod..'\\textures\\ui', '/textures/ui')
-                          break
```
speed2CZ: Actually my bad, this break can stay, as it only stops the loop that is searching for the `ui` folder inside `textures` folder

wodzu93: Is this pull request good now, I should I add that break back in? This only runs on startup, so difference in speed is miniscule.

speed2CZ: The rest it good from my side. But please add it back in. It doesn't matter it runs only once, that's not an excuse to not have it done properly.
In the end it will be squashed into a single commit, so the git history won't get messy.

PR #2939, opened 2019-12-30
remove vet bar for unvettable units

shalkya:
this change apply to cybran land scouts, T1/T3 air scouts, stealth boat, mobile stealth, mobile shields, shield boat, tmd, smd, T1 transports, aeon T3 sonar, drones, mercy, beetle

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1453,8 +1453,19 @@ Unit = Class(moho.unit_methods) {
         end
 
         -- Find a weapon which is not a DeathWeapon / Suicide / DeathImpact / AntiMissile (AEON/SERA TMD) / Turret01 (UEF/CYBRAN TMD) / MissileRack (CYBRAN/SERA/AEON SMD) / AntiNuke (UEF SMD) / AntiTorpedo / AntiTorpedo01 (stealth boat)
+        local No_vet_label = {
+        ['DeathWeapon'] = true, 
+        ['Suicide'] = true, 
+        ['DeathImpact'] = true, 
+        ['AntiMissile'] = true, 
+        ['Turret01'] = true, 
+        ['MissileRack'] = true, 
+        ['AntiNuke'] = true, 
+        ['AntiTorpedo'] = true, 
+        ['AntiTorpedo01'] = true
+        }
         for index, wep in weps do
-            if wep.Label ~= 'DeathWeapon' and wep.Label ~= 'Suicide' and wep.Label ~= 'DeathImpact' and wep.Label ~= 'AntiMissile' and wep.Label ~= 'Turret01' and wep.Label ~= 'MissileRack' and wep.Label ~= 'AntiNuke' and wep.Damage ~= 0 and wep.Label ~= 'AntiTorpedo' and wep.Label ~= 'AntiTorpedo01' then
+            if No_vet_label[wep.Label] ~= true and wep.Damage ~= 0 then
                 return true
             end
         end
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1452,14 +1452,14 @@ Unit = Class(moho.unit_methods) {
             return false
         end
 
-        -- Find a weapon which is not a DeathWeapon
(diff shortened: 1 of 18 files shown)


lua/sim/Unit.lua
```diff
@@ -1452,14 +1452,14 @@ Unit = Class(moho.unit_methods) {
             return false
         end
 
-        -- Find a weapon which is not a DeathWeapon
+        -- Find a weapon which is not a DeathWeapon / Suicide / DeathImpact / AntiMissile (AEON/SERA TMD) / Turret01 (UEF/CYBRAN TMD) / MissileRack (CYBRAN/SERA/AEON SMD) / AntiNuke (UEF SMD) / AntiTorpedo / AntiTorpedo01 (stealth boat)
         for index, wep in weps do
-            if wep.Label ~= 'DeathWeapon' then
+            if wep.Label ~= 'DeathWeapon' and wep.Label ~= 'Suicide' and wep.Label ~= 'DeathImpact' and wep.Label ~= 'AntiMissile' and wep.Label ~= 'Turret01' and wep.Label ~= 'MissileRack' and wep.Label ~= 'AntiNuke' and wep.Damage ~= 0 and wep.Label ~= 'AntiTorpedo' and wep.Label ~= 'AntiTorpedo01' then
```
speed2CZ: This is a horrible way of doing it, make a table with entries ['DeathWeapon'] = true, and so on and then simple check if the Label is in that table.

lua/sim/Unit.lua
```diff
@@ -1453,8 +1453,19 @@ Unit = Class(moho.unit_methods) {
...
+        ['AntiMissile'] = true, 
+        ['Turret01'] = true, 
+        ['MissileRack'] = true, 
+        ['AntiNuke'] = true, 
+        ['AntiTorpedo'] = true, 
+        ['AntiTorpedo01'] = true
+        }
         for index, wep in weps do
-            if wep.Label ~= 'DeathWeapon' and wep.Label ~= 'Suicide' and wep.Label ~= 'DeathImpact' and wep.Label ~= 'AntiMissile' and wep.Label ~= 'Turret01' and wep.Label ~= 'MissileRack' and wep.Label ~= 'AntiNuke' and wep.Damage ~= 0 and wep.Label ~= 'AntiTorpedo' and wep.Label ~= 'AntiTorpedo01' then
+            if No_vet_label[wep.Label] ~= true and wep.Damage ~= 0 then
```
speed2CZ: I find `if not No_vet_label[wep.Label]` a bit more readable
And now you can change to comment to be only on those lines that actually need them

units/UAB4201/UAB4201_unit.bp
```diff
@@ -180,6 +180,7 @@ UnitBlueprint {
     StrategicIconSortPriority = 200,
     Weapon = {
         {
+            ExcludeFromVeterancy = true,
```
speed2CZ: You don't exclude weapons from veterancy, but units, so it doesn't make sense to have this line inside a weapon table. Cuz in this case it leads to having the line in both weapons. Better place would be either on the sam level as Weapon table is or inside General table

lua/sim/Unit.lua
```diff
@@ -1446,6 +1446,7 @@ Unit = Class(moho.unit_methods) {
     -- Returns true if a unit can gain veterancy (Has a weapon)
     ShouldUseVetSystem = function(self)
         local weps = self:GetBlueprint().Weapon
+        local gen = self:GetBlueprint().General
```
speed2CZ: Get the bp only once with
```LUA
```
and then change the conditions to
```LUA
if not bp.Weapon[1] then
```
and so on

lua/sim/Unit.lua
```diff
@@ -1458,8 +1457,8 @@ Unit = Class(moho.unit_methods) {
         ['DeathWeapon'] = true,
         ['DeathImpact'] = true,
         }
-        if not gen.ExcludeFromVeterancy then
-            for index, wep in weps do
+        if not bp.General.ExcludeFromVeterancy then
```
speed2CZ: I would do this check in the same step with checking for the weapons, since it leads to the same result. So:
```LUA
if not bp.Weapon[1] or bp.General.ExcludeFromVeterancy then
    return false
end
```
And only after that go through the weaspons

KionX:
Whats the general idea behind this? Whats the advantage?

speed2CZ:
It doesnt show the veterancy bar in the UI for the units that can't get veterancy

PR #2942, opened 2020-01-02
make selection of mobile exp/sonar under construction act as building

shalkya:
adding a cat that have been introduce with an exe update to reduce the priority of selection of mobile exp/sonars when being built.
i added the category only to mobile exp/sonar since T4 arties, para and novax doesn't have the selection issue.
This issue was brought when introducing CQUEMOV with exe patch to allow the selection of mobile exp/sonars under construction (in order to give rally point order for example) and to give attack order to pds (when their construction isn't yet finished)

worth noting : unfinished T4 building are selected over unfinished mobile exp/sonars, apart paragon that is selected last.

### DIFF
--- a/units/UAA0310/UAA0310_unit.bp
+++ b/units/UAA0310/UAA0310_unit.bp
@@ -116,6 +116,7 @@ UnitBlueprint {
         'OVERLAYMISC',
         'CANNOTUSEAIRSTAGING',
         'SNIPEMODE',
+        'LOWSELECTPRIO',
     },
     CollisionOffsetY = -1.5,
     Defense = {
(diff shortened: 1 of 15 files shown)


Strogoo:
I leave this pseudo C code that describes exe patch here:
```
Before patch:
    selectionPriority = unitBlueprint.General.SelectionPriority

After:
    if (unit::IsBeingBuilt() && unit::IsInCatgory("LOWSELECTPRIO")) {
        selectionPriority = 6
    } else {
        selectionPriority = unitBP.General.SelectionPriority
    }
```
So by default all unfinished units have their normal selection priority from blueprint. Adding `LOWSELECTPRIO` cat changes this value to 6.

shalkya:
tested by nojoke

PR #2944, opened 2020-01-03
fixes airscout

TurinturambarFAF:
fixes airscout being unable on its own bugg introduced by lower turnspeed.


shalkya:
taken care of by #2947

PR #2946, opened 2020-01-04
Improved message about banning ctrl+k in coop

KionX:
From this:
ForgedAlliance_2020-02-02_21-57-38
To this:
ForgedAlliance_2020-02-02_22-01-22

### DIFF
--- a/lua/ui/game/announcement.lua
+++ b/lua/ui/game/announcement.lua
@@ -1,0 +1,181 @@
+--*****************************************************************************
+--* File: lua/modules/ui/game/announcement.lua
+--* Author: Ted Snook
+--* Summary: Announcement UI for sending general messages to the user
+--*
+--* Copyright © 2007 Gas Powered Games, Inc.  All rights reserved.
+--*****************************************************************************
+
+local UIUtil = import('/lua/ui/uiutil.lua')
+local LayoutHelpers = import('/lua/maui/layouthelpers.lua')
+local Group = import('/lua/maui/group.lua').Group
+local Bitmap = import('/lua/maui/bitmap.lua').Bitmap
+
+local bg = false
+
+function CreateAnnouncement(text, goalControl, secondaryText, onFinished)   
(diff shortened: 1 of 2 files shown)


shalkya:
look here how to make proper commits when adding and changing new files.
https://github.com/FAForever/fa/pull/2922

what you can do : 
git checkout Improved-message-about-banning-Ctrl+K-in-coop
git reset --soft head~4
git stash
add the raw files back in and commit the addition of raw files
git stash apply
commit the change
git push -f

(if that didn't work for some reason)
git checkout Improved-message-about-banning-Ctrl+K-in-coop
save the files you are changing somewhere else
git reset --hard head~4 (be careful this delete your files/commits, make sure to save your files before)
add the raw files back in and commit the addition of raw files
replace raw files by the files you saved
commit the changes
git push -f

KionX:
Up!

speed2CZ:
First, since there's no description, don't be surprised some people didnt bother to figure it out.
Second, it's not improving the messege, it's changing it's behaviour.

From this:
ForgedAlliance_2020-02-02_21-57-38
To this:
ForgedAlliance_2020-02-02_22-01-22

So its just an announcement instead of a pop up you have to click away.

Personally I don't see how it's better, it's just different. Maybe easier to miss since it can overlap with some other announcement (objective added, completed, etc).
So I would like to know the reason for this change.

Lastly the ability to create an announcement without the `goalControl` is fine. Ideally I would imagine it showing up from the top centre of the screen instad of on it's target position, since all other announcements start from the edge of the screen. But it's not a condition.

But what I would like you to change is to split the commit into two. One that makes the announcement work without the `goalControl` and second that changes the behaviour of the self destruct message from a pop up to the announcement.

KionX:
The fact that this PR is done this way makes sense.
The pop-up window that you have to click blocks the control. Annoy.
The announcement works without goalControl and a live demonstration of this is included here.
I think you're picking on little things.

PR #2953, opened 2020-01-08
t1 scout attackmove range hotfix

Petricpwnz:
Set a GuardScanRadius for t1 scouts with weapons so they don't run into enemy units but shoot from proper range.

Might be worth it to look into the code that auto-sets this to see why they aren't affected and if there are other units that might suffer from the same problem. I _might_ do it later.

### DIFF
--- a/units/UAL0101/UAL0101_unit.bp
+++ b/units/UAL0101/UAL0101_unit.bp
@@ -1,4 +1,7 @@
 UnitBlueprint {
+    AI = {
+        GuardScanRadius = 50, --RadarRadius
+    },
     Audio = {
         AmbientMove = Sound {
             Bank = 'UAL',
--- a/units/UAL0101/UAL0101_unit.bp
+++ b/units/UAL0101/UAL0101_unit.bp
@@ -263,7 +263,7 @@ UnitBlueprint {
             },
             FiringTolerance = 2,
             Label = 'LaserTurret',
-            MaxRadius = 25,
+            MaxRadius = 30,
             MuzzleSalvoDelay = 0,
             MuzzleSalvoSize = 1,
             MuzzleVelocity = 25,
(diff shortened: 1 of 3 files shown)


shalkya:
they should stop to the range of T1 tank no ?

Petricpwnz:
I don't see why, that would mean they also would usually stop at the range of where enemy tanks can shoot them while they hold the range advantage.

shalkya:
well you want them in the army not being 10 range or more in the back. if they stop max range then you don't see what's beyond.

Petricpwnz:
You see the front line with vision and the rest with radar, doesn't really make sense to me that they should stop at an arbitrary range of a partner tank. Also since they have higher speed than said tanks they will always rush in front of them when attack moved and if they stop at the tank range enemy units have a decent chance of killing them anyway, which is what this is supposed to prevent. If you want, maybe let jagged decide here.

shalkya:
you can put 2 more range than the T1 tank, should be enough.
typical situation : attack move, scout stop at radar range when they see the first unit of the army, then aurora arrives, you have no idea how big the army is. You could get rolled over. Especially with the slow updating of radar blips, that could even fuck up the firing cycle of aurora when they kill the last that was in radar range (even if there is more in coming behind it)

Petricpwnz:
No, no, you misunderstood. They won't stop at radar range, they will stop at their own shooting range (a bit closer technically because of deceleration time). You will still see the front line with vision and the rest with radar just fine. Radar range will be the range they aggro at enemy units when on patrol and such.
For aurora-spirit it's already like you wan't because spirit range is even smaller than aurora (might actually want to change that)
Selen has smaller range than the tank so it stops even closer than the tank.
The only case left is snoop and 2 more range than the tank could work too but its not even too different from what I have here and again, doesn't make too much sense to me to have that slight difference which would ruin the logic and intuitiveness of one would expect attack move to work.

PR #2957, opened 2020-01-11
For compatibility 'Removed Observing After Death' with the new EXE

KionX:


shalkya:
i won't use this untested exe anyway. Also for further testing it would be nice that you explain why you needed to remove "the observer after death fix". So we understand what conflict there was.

KionX:
This is just a compatibility patch. You will need it with the new exe.

As I understand it "the observer after death fix" refers to this https://github.com/FAETHER/FA-Binary-Patches/pull/8
This patch disables checking the OutOfGame flag in SetFocusArmy. So the death of a player does not make him an observer automatically.

PR #2959, opened 2020-01-12
In cheat menu added the ability to spawn with veterans and improved usability.

KionX:
Incorrect entries in the "count" and "veterancy" fields are excluded.
Added immediate creation of a unit on the right mouse button.

Added default values if the above fields are empty.
image

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -119,6 +119,20 @@ Callbacks.CapMex = function(data, units)
     IssueGuard(units, builder)
 end
 
+Callbacks.SpawnAndSetVeterancyUnit = function(data)
+    if not CheatsEnabled() then return end
+    for bpId in data.bpId do
+        for i = 1, data.count do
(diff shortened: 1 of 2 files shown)


Nojoke185:
Hi, tested this with no real problems seen
couple of queries:

1 - units with vet do not spawn with full health. They have to level up to full health. Should they spawn at max health for the vet level?

2 - in the cheat menu, if you double click the 0 default value for veterancy you get the below (the 0 is highlighted):

image

if you then type a vet level to replace the default highlighted text (0) - for example to type 1, then the UI does not allow overwrite of the default 0 value (in normal windows highlighted text is replaced by the new input)
you first have to backspace then type the vet value
It would be nice if when highlighted the new input replaced the default without needing to backspace. If this is possible in the code?

KionX:
1 - will do.
2 - will do.

KionX:
Fixed.

PhilipJFryFAF:
tested and works fine
very nice addition

KionX:
Removed ForkThread leaks.
Added immediate creation of a unit on the right mouse button.

PR #2960, opened 2020-01-13
Seraphim shader patch

BlackYps:
I made another update for the seraphim shader. Pictures can be found here: https://forums.faforever.com/viewtopic.php?f=2&t=17676&p=181285#p181285
I can provide more pictures if you want, but it is best to try it out.


shalkya:
I made a video, i will get it post on the faf youtube and gather feedback !

BlackYps:
Nice! I don't know if it is really needed, because the change is not so drastic, but now that you have the video already go for it!

PR #2961, opened 2020-01-16
Black plane

BlackYps:
I deleted the background plane in the world border meshes. This removes the well known black plane bug where this plane intersects with maps with high elevations.
Tested on different maps and found no side effects.


shalkya:
Please direct future changes to deploy/fafdevelop. Thanks.

PR #2963, opened 2020-01-18
Refactoring

speed2CZ:
WIP

Saving the unit's army like we save EntityId

Afaik this should run faster than calling the engine function GetArmy()

Maybe someone like @Uveso would know how to measure it in his 300h long AI games?

And if it runs faster, it might be worth saving the army in all Entities, so things like Projectiles can also benefit from this, as there are more GetArmy calls in their scripts.

### DIFF
--- a/effects/Entities/InainoEffectController01/InainoEffectController01_script.lua
+++ b/effects/Entities/InainoEffectController01/InainoEffectController01_script.lua
@@ -15,12 +15,11 @@ local SIFInainoStrategicMissileEffect04 = '/effects/Entities/SIFInainoStrategicM
 
 InainoEffectController01 = Class(NullShell) {
     EffectThread = function(self, Data)
-        local army = self:GetArmy()
-        self:ForkThread(self.CreateInitialHit, army)
-        self:ForkThread(self.CreateInitialBuildup, army)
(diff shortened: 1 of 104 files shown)


Uveso:
Using a direct value instead of an function call always save CPU runtime.
I don't know if i can do a test and see how much the difference is, but ther is a difference for sure.

Could you also please add the unit BlueprintID to the unitclass ?

Uveso:
Sadly found a bug in `DefaultProjectiles.lua` line 423
`OCProjectiles[army] = OCProjectiles[army] - 1`
should be
`OCProjectiles[self.Army] = OCProjectiles[self.Army] - 1`

speed2CZ:
@Uveso I picked UnitId instead of BlueprintID. So it's same as the function

Uveso:
Shield.lua needs a change.
Add in line 48:
`self.EntityId = self:GetEntityId()`

Line 137 and 165.
Change
`amount = amount * (1.0 - ArmyGetHandicap(self.Army))`
back to
`amount = amount * (1.0 - ArmyGetHandicap(self:GetArmy()))`

Found no other errors.

Uveso:
@speed2CZ
@shalkya 
Looks very good now.
Can't find any additional errors.

Ready for merge.

PR #2969, opened 2020-01-25
Speed2 cz's Refactoring Update

Uveso:
Updated PR #2963


Uveso:
As usual i did something wrong :/

PR #2970, opened 2020-01-26
Added 4 new functions to the Engine documentation.

KionX:


PhilipJFryFAF:
please add a proper description so people are able to understand what this PR is supposed to do
otherwise testing is impossible for anyone who is not a developer

engine/Sim.lua
```diff
@@ -382,6 +382,10 @@ end
 function GetCurrentCommandSource()
 end
 
+-- Return list of deposits
+function GetDepositsAroundPoint(X, Z, Radius, Type)
```
speed2CZ: what deposits? mass or energy or both?
and in what format?

KionX:
I think it's self documented PR.

engine/Sim.lua
KionX: There's nothing special about it.
Everything is standard for FAF.
I didn't even think anyone was going to check it out.

speed2CZ: It doesn't seem standard to me. FA code uses 'MASS' or 'ENERGY' as parameters when calling similar functions conserning resources. I've never seen numbers. So that is already different.

Then I would still like the know in what format are the resources returned. I figured it's gonna be a table, but the structure of this table?

That's why we bother writing the documentation down, so you can read it up instead of having to test it every time.

Exotic-Retard: yeah that is actually pretty confusing, normally you expect a text input for that sort of stuff, especially in fa code. would be good to make it clear that this function is different from the rest.

KionX: A. Maybe I was wrong. So the type is presented in the engine.
I forgot to write a decryption of this type. Thanks for reminding me.

PhilipJFryFAF:
I'll still have to insist on a proper explanation of this PR @KionX 
what is this PR supposed to fix or change?
what was the situation before that made it necessary to make this PR?
how can a user test if this PR is doing what it is supposed to do and what may be potential issues with this PR that we may have to look out for?

speed2CZ:
This isn't fixing anything. Everything inside engine folder are functions available from the game engine. Basically this: https://supcom.fandom.com/wiki/LUADOC_1.5.3599 just split into parts, so it can be documented.

This PR adds new functions that got added with the new exe. And would be nice to have them documented properly from the people who made them.

KionX:
Hopefully now functions documented properly.

shalkya:
Fyi you don't need to close and reopen the PR in order to force push your branch. The PR will update automatically.

PhilipJFryFAF:
can someone link the PR that added the functions that are documented by this change?

KionX:
> you don't need to close and reopen the PR

Github closes them under certain circumstances.

> can someone link the PR that added the functions that are documented by this change?

The link to the repo is in README.md
There you can search by names.
With PRs, everything is confusing. So I didn't specify them.

Uveso:
Please add the special case for the salem:
https://github.com/FAForever/fa/pull/2965/files#diff-d189570bd1518639caf6d140f6e9ba78R174

PR #2976, opened 2020-01-31
Fixed position autofill window when resize the gamewindow

KionX:


PhilipJFryFAF:
tested and works

in the current game patch the autofill element will get moved out of its supposed location upon resizing the window

does not happen with this PR any more

PR #2978, opened 2020-02-01
T3 mobile artilery nerf

TurinturambarFAF:
broad idea for a cyb mobile arty nerf, directed at reducing their effectiveness against (moving) units, while keeping their effectiveness vs firebases unchanged


projectiles/CIFArtilleryProton01/CIFArtilleryProton01_script.lua
```diff
@@ -9,20 +9,14 @@ CIFArtilleryProton01 = Class(CArtilleryProtonProjectile) {
     OnImpact = function(self, targetType, targetEntity)
         CArtilleryProtonProjectile.OnImpact(self, targetType, targetEntity)
         local army = self:GetArmy()
```
speed2CZ: Use self.Army instead

Strogoo:
@speed2CZ That was my advice to copy-paste models and textures with renaming. I know you can use one model/texture with different entities but I don't remember what lines in blueprint are responsible for that

So I was like: "Hey, java boys always say memory is cheap, and disk space is even cheaper!" :D

speed2CZ:
If the only advice you have is bad and you know it's bad, maybe don't give such advice? There's a channel for game development where you can get help...

shalkya:
would be nice if turin follow the change asked by speed. Eg change army by self.Army
Now i'm having resolve conflicts.

projectiles/CIFArtilleryProton01/CIFArtilleryProton01_script.lua
```diff
@@ -8,21 +8,15 @@ CIFArtilleryProton01 = Class(CArtilleryProtonProjectile) {
...
-        end
-        ForkThread(self.ForceThread, self, self:GetPosition())
-        self:ShakeCamera( 20, 3, 0, 1 )        
-    end,
-
-    ForceThread = function(self, pos)
-        DamageArea(self, pos, 10, 1, 'Force', true)
-        WaitTicks(2)
-        DamageArea(self, pos, 10, 1, 'Force', true)
-        DamageRing(self, pos, 10, 15, 1, 'Fire', true)
```
shalkya, line 25: why did you remove the forcethread ? iirc it is used to destroy trees when the artillery impact

PR #2980, opened 2020-02-09
Fix UEF Navy UI things

speed2CZ:
- [x] Improves the positions of the Cruiser and Shield boat health bar.
- [x] Improves the selection thickness of the Shield boat.
Before:
ForgedAlliance_2020-02-09_09-41-26
After:
ForgedAlliance_2020-02-09_09-40-04

- [x] Fixes the position of the shield boat emitters
Before:
ForgedAlliance_2020-02-09_11-24-03
After:
ForgedAlliance_2020-02-09_11-23-14

- [x] Increased Shield boat idle effect (the white just around the ship)
Before:
ForgedAlliance_2020-02-09_11-57-15
After:
ForgedAlliance_2020-02-09_11-57-46

### DIFF
--- a/units/UES0202/UES0202_unit.bp
+++ b/units/UES0202/UES0202_unit.bp
@@ -213,8 +213,8 @@ UnitBlueprint {
         HelpText = '<LOC ues0202_help>Cruiser',
     },
     LifeBarHeight = 0.075,
-    LifeBarOffset = 3.45,
-    LifeBarSize = 1.75,
+    LifeBarOffset = 2.45,
+    LifeBarSize = 1.5,
     Physics = {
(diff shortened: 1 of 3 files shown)


shalkya:
can't see the difference between the 2 last pic.
otherwise cool stuff, i will test that soon.

speed2CZ:
The last one is very small change, its the white foam that is right next to the edges of the ship. Mostly noticable in this area:
firefox_2020-02-09_15-24-23

PhilipJFryFAF:
tested and works
as always good job on the fix!
also, the shield boat stuff really bothers me now - was that issue introduced when increasing the unit size?

PR #2983, opened 2020-02-09
Improve Frigate sink animation rate

speed2CZ:
Aeon Frigate sinks faster now
Cybran Frigate slower

Overall its +-similar for all faction. Each of the 3 animations has
different duration so it can still differ.


PhilipJFryFAF:
tested and works
there are still minor differences but it's a lot closer now
👍

PR #2987, opened 2020-02-10
Fix targetting of structures for beam weapons

speed2CZ:
Fixes some units from the list #2988

Adjusting the target bones of structures to not be at the ground level. Since any small height elevation causes beam weapons to hit the ground instead. e.g. sera destro not being able to hit anything.


speed2CZ:
This PR is done, the rest of the units will require adjusting bones in the model, so once that's done I'll make a new PR

speed2CZ:
Some pictures
Before:
ForgedAlliance_2020-02-15_16-54-52
After:
ForgedAlliance_2020-02-15_16-58-55
And same effect fot other all other structures changed by this PR

PhilipJFryFAF:
tested and works

PR #2990, opened 2020-02-14
Update units consumption values only if building something

speed2CZ:
Untested, might have side effects of breaking up something else. Up to the testers to find out all possible combinations of building and assisting.


speed2CZ:
Syntax errors pop out right away, what we're after is no semantic errors! :D

shalkya:
Tested, works and no error.

PR #2994, opened 2020-02-17
Ui scaling

speed2CZ:
Please test

What's changed:
* Added a lobby option to set the scale multiplier: 1.0, 1.25, 1.5, 1.75 and 2.0. Game restart is requires as it's the easiest way to refresh all the UI elments.
* Basic UI classes like Bitmap got scaling implemented in them. When a new texture is set, function that resets the dimensions scales the texture according to the scale multiplier
* For placing UI elements relative to some others, correct function from LayoutHelpers has to be used, since they scale the offset.
* Setting Width and Height also goes through helper function that scales the number.
* AlphaHitTest has to be disabled on the scaled buttons. Else they flicter when moused over and some buttons crash the game.


tatsujb:
https://forums.faforever.com/viewtopic.php?f=2&p=182124#p182094

this should probably be merged.

Uveso:
speed2CZ , that's amazing!
I tested it and i can see you will finally make it!
Can't wait to test the final version :D

shalkya:
well well, i can't test much of it since i don't have a big enough screen :'(
gonna look for testers able to test that one.
Although i've a question, if someone put 200%, then start the game, realize he has a little screen and want to change back but he can't reach the panel for modifying it back to normal, how can he solves he change back to previous UI scale ? (apart from reinstalling the whole game)

speed2CZ:
Well you can test yourself even with the small screen since nothing should break if you don't use the scaling, as all the offsets are just multiplied by 1.

And if someone manages to fuck up so bad, it's just modifying the single value in the game.prefs but I dont think it could be that bad

shalkya:
Well i kinda wanted to see how it would go with scaled, but i can't even launch the game.

Maybe it would be preferable to make a better warning telling the user that he might fuck himself by increasing the % (especially that you can't expect them to be able to change the game.prefs). Or you can slide in a reset button that would be located toward the middle of the screen, but that wouldn't be super aesthetic.

speed2CZ:
"can't even launch the game" isnt very descriptive. Anyway the game starts even with 800x600 and 200% scaling, you won't be able to get to the menu anymore tho.

I dont know how you imagine this reset button to work and where would it be?

shalkya:
""can't even launch the game" isnt very descriptive."
that's what i mean by not being able to launch a game. I can't access the button.
image

i guess you could split the change profile button in 2. Keep the "change profile" on the left and put a "reset UI" on the right.
image

speed2CZ:
There's no split function, that would require making new textures to split the button

Uveso:
I also had the problem with unreachable buttons.

Maybe we could add a command line parameter like /nosound to the game.

/resetui or /resetuiscaling could reset the ui scaling in case something is messed up.

speed2CZ:
That sounds the same diffuculty as editing the game prefs file .... For the user

KionX:
x1
x2
x3
I didn't find any other problems.

speed2CZ:
I dont know how to fix the options, it uses some extra logic that doesnt want to scale properly.

Cheat menu is a minor problem, it's common with all checkboxes (but visible only here), I'll look into that.

Fuzzy text looks normal to me, might be poor font/font size choise, else all texts are scaled the same way and this one is no exception.

Alt F11 - is there a Lua file for that? I havent found it, so can't fix it.

KionX:
> Alt F11 - is there a Lua file for that? I havent found it, so can't fix it.

Steam FA \lua\debug

shalkya:
Maybe you can put at the top of the menu (in place of the "campain" button) the "option" button. So that when people switch to 200% and realize it's not the good scale, then they can switch back to another scaling. Need to test if while opening the "option" button, they can still access the "UI scale" option

speed2CZ:
I dont feel like making this foolproof, if someone has small monitor and feel like 200% scaling is the only way to go, then they can go ahead and edit the prefs file or delete it and start over. It's not game for iditots, so the number of people doing this should be very low.

speed2CZ:
Cheat menu, checkboxes and engine stats are now scaled properly. Layout in the options is still resisting...

speed2CZ:
Alright boys and girls, options UI has been defeated and that should be all the UI elements scaled!

speed2CZ:
Score works fine for me, how to reproduce it?

KionX:
In the last test, everything is correct.

shalkya:
ok i'm testing by reducing the ui size instead of increasing it.
image
isn't the window suppose to scale down : so for example the bottom right of the slider and the border of the window, get to the bottom right the uef drone ?
image
position of the + isn't good. (same for notification management)
image
the yes/no answer are miss placed (same for endgame and exit)
image
same here, also this pop up shouldn't happen (didn't modify anything)

So this won't work with a lot of mods, but that was to be expected. At least it doesn't break them at normal scale.

shalkya:
it actually break stuff on normal scale with mods : 
image
doing a comparison with normal game : 
image
there is this weird control group bar on the right that does nothing, and the multi function bar disappeared. And the unit number is fucked up for ssb.
also it doesn't actually end the game when you click "yes", just switch your faction color to observer and don't let you perform order anymore (in sandbox)

So it actually conflict a lot with mods

shalkya:
image
that's when you only use eco manager.

To get back to normal scale, i edited the game.pref to put a 1. (tell me if that's wrong)
I won't look into more detail (knowing what mod conflict with what)
normal size need to not change the displayed lay out at all.
Here (as i see it) you are breaking the mods for all users, while improving the ui display for like 2-3% of the user. 
So this mean that you might need to have the scale function work only when value is different to 1.

speed2CZ:
Scaling down doesnt work. You can't go under 1, as you found out, everything will break.

speed2CZ:
All my 12 UI mods works unscaled just like before (including SSB). With scaling I had to disable 4.

It should not really break anything with scale set to 1. As all the numbers are just multiplied by 1, note that majority of the functions that scale the offset when placing the UI elements already existed and the scale way set to 1 in these files.

So everything works for me with scale 1:
ForgedAlliance_2020-03-15_07-07-10

I suspect some error at your end, can you send me eco manager so I can test it out?

speed2CZ:
@shalkya score/SSB might was broken because of this change wasnt in: https://github.com/FAForever/fa/commit/4ee4dd00804022722ad50ded871cd22392956b11
I've rebased the PR on the current deploy/fafdevelop

PhilipJFryFAF:
cannot reproduce your issue keyser
i'm using the following ui mods:
Additional Camera Stuff
Advanced target priorities
chat_mute
common mod tools
decals
ecomanager
noshake
pause replays at time
selection cost ui
spread move
ssb
zep_minmapzoom

shalkya:
Ok the rebase seems to have fix most issues for mods at scale 1.

Since scaling down doesn't work, i did some test with scale 1.1.
image
If you change your scale to a non-standard one in game.pref, each time you go in option - interface (and don't change anything) the pop up comes out. Not so dramatic if you ask me.
image
i think this shouldn't happen (issues with button)

speed2CZ:
Fixed

shalkya:
image
since your last commit (i think), this bug was introduced (using 125% or 110%)

tatsujb:
tech markers and Icons in the build menu are still not fixed since this was implemented :

Screenshot from 2020-06-05 19-45-16

this is as a result of UI-scaling right?

shalkya:
it's fixed in https://github.com/FAForever/fa/pull/3065

lee-duck:
Hi. I have recently started using the UI scale feature having just upgraded to a 4K monitor.

After some research I have managed to fix a couple of issues e.g. resource panel, ECO manager interface.

But I have one remaining issue: some of the overlay elements (e.g. nuke missile timer, displayed on top of the nuke missile building), seems to have its offset value set incorrectly, so it displays far away from the nuke building. Any way to fix?

(I also have an issue with ECO manager adornment overlays doing the same thing - I can live without that and realise it's a mod but if anyone knows a solution I'd be very grateful).

Thanks

speed2CZ:
@lee-duck my guess is its gonna work the similar way as the wreck reclaim markers, so to show the correct position, it has to calculate it with the scale factor
You can check the change in `lua/ui/game/worldview.lua`

PR #2995, opened 2020-02-18
Fix Kennel lag and animation

Uveso:
Old Issue; if you build more than 30 kennels and upgrade them at the same time, it was freezing the game for 2-3 seconds when new drones spawns.
The lag came from a table.deepcopy function that was making a deepcopy of the drone data.
The deepcopy was not only making a copy of the unit, it also did a copy of platoondata, Brain and all other tables that where only present by a pointer.

This was also triggering a error after several hours of playtime:
```
WARNING: Error running OnStopBuild script in Entity xeb0104 at 7c42e708: ...:\programdata\faforever\repo\fa\lua\system\utils.lua(91): stack overflow
         stack traceback:
         	...:\programdata\faforever\repo\fa\lua\system\utils.lua(91): in function `deepcopy'
         	...:\programdata\faforever\repo\fa\lua\system\utils.lua(102): in function `deepcopy'
```

Also fixed a minor issue with the door animation. They only opens if a real buildcommand was issued.
Selecting the drone and just let them start was not opening the doors.
Now doors stay open if the first drone is detached and will close if all drones return.

And last but not least i readded the door open and close sound.

Animation note:
The dooranimation is opening and closing the door in the same animation.
It takes 0.5 seconds to open the doors and 0.5 seconds to close the doors.
If we want the doors to stay open, we need to start the animation and stop it after 0.5 seconds.
That's why it needs a forked thread on animation open.

### DIFF
--- a/lua/terranunits.lua
+++ b/lua/terranunits.lua
@@ -471,7 +471,7 @@ TPodTowerUnit = Class(TStructureUnit) {
 
     OnTransportAttach = function(self, bone, attachee)
         attachee:SetDoNotTarget(true)
-        self:PlayUnitSound('Load')
+        self:PlayUnitSound('Close')
         self:RequestRefreshUI()
         local PodPresent = 0
         for _, v in self.PodData or {} do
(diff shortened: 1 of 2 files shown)


shalkya:
couldn't replicate the initial freeze issue.
Tested, doesn't trigger any errors.

PR #2998, opened 2020-02-19
Czar movementspeed adjustment

TurinturambarFAF:
changes czar movespeed back to 8


speed2CZ:
CZAR isnt broken, so changing the move speed can't be really considered as a fix. You should use that as the title of the commit, so something like: Adjust CZAR movement speed or even better: Decrease CZAR movement speed

Else the commit history will be impossible to read. The commit names should be always as descriptive as possible.

PR #3000, opened 2020-02-20
OC changes and tweaks

Petricpwnz:
oc kills mobile shield gen if it depleets it (only one, overspilled ones live)
vet counted, no reclaim left

DMG to e ratio changed to 1/3. Exponential increase in e cost for dmg made sense with very high dmg cap but then we had to change it to a 15k dmg cap due to oc scus or smh so im changing it, might be rebalanced to 3.5-4/1. 3/1 is almost the same in numbers as it is rn but easier to predict and much easier to edit. Min dmg changed to 1666 to keep the min e drain of 5k per oc.


lua/sim/DefaultProjectiles.lua
```diff
@@ -397,6 +398,7 @@ OverchargeProjectile = Class() {
                         idealDamage = data.minDamage
                     else
                         idealDamage = targetEntity.MyShield:GetMaxHealth()
+                        killShieldGenOnImpact = true
```
Strogoo: This also includes fatboys and buble sacus. Add new category to units blueprints which should die to OC and then check for it here.

Also since we do some other category checks in this code, you better put 
`targetCats = targetEntity:GetBlueprint().CategoriesHash` 
at line 394 and then use

```
if targetCats.CATNAME then
        killShieldGenOnImpact = true
```

Also line 397 `if EntityCategoryContains(categories.STRUCTURE, targetEntity)` to `if targetCats.STRUCTURE`
and line  408 `if EntityCategoryContains(categories.COMMAND, targetEntity)` to `if targetCats.COMMAND`

PR #3002, opened 2020-02-23
remove dragbuild from sensors/e storage

shalkya:
- all sonars
- all radars
- all stealth field structures
- all e storage

philip seems to be against the e storage. I still feels it promotes better gameplay and also make life a little bit easier for good players. I'll let you argue against or for it.

### DIFF
--- a/units/UAB1105/UAB1105_unit.bp
+++ b/units/UAB1105/UAB1105_unit.bp
@@ -35,7 +35,6 @@ UnitBlueprint {
         'ECONOMIC',
         'TECH1',
         'ENERGYSTORAGE',
-        'DRAGBUILD',
         'SIZE4',
         'VISIBLETORECON',
         'RECLAIMABLE',
--- a/units/UAB1105/UAB1105_unit.bp
+++ b/units/UAB1105/UAB1105_unit.bp
@@ -35,6 +35,7 @@ UnitBlueprint {
         'ECONOMIC',
         'TECH1',
         'ENERGYSTORAGE',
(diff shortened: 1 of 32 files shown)


speed2CZ:
Im also against storage

Uveso:
Removing dragbuild makes sense on buildings like radar, sonar etc.
2 Radar on the same spot are useless. I agree on that.

But this is not the case for storage buildings.
Sometimes you need drag build in case you want to maximise adjacency bonus.

shalkya:
" Sometimes you need drag build in case you want to maximise adjacency bonus."
well that's bad practice nearly all the time though. You shouldn't build e storage for adjacency bonus. Only thing is saw before is air player building some line of e storage in their back base for having some storage for aircraft production. So for this case i think i won't remove the drag build for e storage. although i think removing it would add annoyance to lower the number of noob building storages near their hydro and T3 pgen.

Btw didn't say that before i also added the soothsayer to that list.

JaggedAppliance:
I'm also against removing it from e storage.

Uveso:
@shalkya 
maybe its a bad behavior but i am using dragbuild in every game.
So please don't remove this features from the game while i am using it.

Have in mind i play mostly against AIs.
That means i am mostly turteling and my e-storage is perfectly safe
even when i drag build them through my whole base :)

shalkya:
yes i'm going to update the PR when i find some time

speed2CZ:
And I do it to show dominance

Issue #94, opened 2014-10-27
Range overlay reduces FPS by up to 90%


Crotalus:
Range overlays are a huge performance bottleneck, especially "Selection" and it is enabled for all users as default.

The toll is taken on the GPU so people with fast cards notice less impact but the problem is that if someone is running a low-end gfxcard (like me for the moment), the result is stall / stutter in a network game. The sim and render side isn't decoupled enough :/

For someone wanting to test the impact you can create say, 200 asf and then select those with selection range overlay enabled. As long as you have them selected on screen it will cause a huge fps drop, It is probably the "merge overlay circles" code which is crap.

My suggestion is to remove Selection-overlay as an option in the UI completely and make sure it starts disabled. The only time it should be on is if you select a single unit (which is already the case). It is useless anyway for the multi-unit select case.

I will put up a pull request with some other performance improvements at the same time.

aeoncleanse:
I've had my uncle take a look at this as it appears in game, not the code. He's an expert programmer who specialises in geometric functions for CAD operations, and his jaw dropped when he saw the impact these rings had on FPS. I think it's essential we find out how they are calculated, how the intersect is calculated, and why the hell this FPS drop is happening. On my machine, even selecting 3 ASFs drops my FPS from 60 to 25.

Crucial: Zoom level affects the FPS, so it does appear to be the display aspect rather than the calculation which is causing it.

aeoncleanse:
Since my last post on this, I've upgraded my laptop. The old one had a 560m GPU and an i7 CPU. The new one has an i7 which is slightly faster, but not by much. However, the GPU is a 980m, a SIGNIFICANTLY faster GPU. For me, this issue has completely vanished with the new machine, confirming this is definitely a rendering issue not a CPU side issue.

Crotalus:
Probably fixed by Eximius here:

3cb462d2e7b7b08e57e617b5a3f49d81b3251b9f

ChrisKitching:
Definitely not fixed by Eximius there, but made very slightly less awful.

You really should lurk in Slack more, Crotalus, it's where the low-latency stuff happens.

aeoncleanse:
This is not solved... But is it as far as we can go? Can it be closed?

Sheeo:
The bug remains.

HUSSAR-mtrela:
[...]

PR #3004, opened 2020-02-23
SingleRangeOverlay cleanup

Strogoo:
This code was added to fix issue #94 or at least to make it less harmful. But code logic is broken and doesn't work as intended.

Also idea is bad overall, cause if you have 'global' rings disabled and this code works as intended, then you can't see any rings for multiple selected units at all.

And it's 2020 after all, gpus are good enough to handle range rings in most cases + we did some optimisations in engine.

-----------------Testing-------------------
As I said, the code logic is broken and removing this code won't affect anything. Some basic tests: select units with `Selection rings on` then with `Selection rings off` and things like that.

### DIFF
--- a/lua/keymap/keyactions.lua
+++ b/lua/keymap/keyactions.lua
@@ -339,8 +339,6 @@ keyActions = {
         category = 'selection', order = 22,},
     ['toggle_selectedinfo'] = {action = 'UI_Lua import("/lua/keymap/selectedinfo.lua").ToggleOn()',
         category = 'ui', order = 23,},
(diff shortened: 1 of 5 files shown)


KionX:
In selectedinfo.lua is full of garbage related to this.
Can I clean this too?

Strogoo:
done

KionX:
Tested. Works.

Uveso:
Viewed the code.

Looks like the variable `SelectedInfoOn` also needs a clean  up.
As far as i can see, its compleatly useless.
`SelectedInfoOn` only exist inside `selectedinfo.lua`

shalkya:
tested a bit, didn't see any errors

PR #3005, opened 2020-02-24
Prevented unnecessary focus changes when cheats are enabled

KionX:
When any player dies, the focus army switches to the observer. Prevented.

### DIFF
--- a/lua/ui/game/gameresult.lua
+++ b/lua/ui/game/gameresult.lua
@@ -7,6 +7,8 @@
 
 local UIUtil = import('/lua/ui/uiutil.lua')
 
+local sessionInfo = SessionGetScenarioInfo()
+
 local OtherArmyResultStrings = {
     victory = '<LOC usersync_0001>%s wins!',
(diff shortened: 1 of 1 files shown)


lua/ui/game/gameresult.lua
```diff
@@ -42,7 +44,7 @@ function DoGameResult(armyIndex, result)
     local armies = GetArmiesTable().armiesTable
     announced[armyIndex] = true
 
-    if not SessionIsReplay() then
+    if (not SessionIsReplay()) and (sessionInfo.Options.CheatsEnabled ~= 'true') then
```
speed2CZ: I assume this option to be always either 'true' or 'false' so the condition will be more readable to test `sessionInfo.Options.CheatsEnabled == 'false'`

KionX: I don't trust FAF. I think it's safer that way.

shalkya:
works.
On a side note, SSB was already preventing this smh. (so if you wanted to reproduce the bug you need to start it without SSB)

Issue #3001, opened 2020-02-22
Torpedo bombers don't reset their attack orders properly


Strogoo:
Take torpedo bomber and some low hp units that will die in 1 pass. Then give order to attack them using shift and see how long it takes for bomber to reset its attack order after target is dead (air-to-ground bombers works fine btw). This huge delay leads to a useless approach towards dead unit and it's very crucial especially for Solace.

*I'm already working on a fix

added: happens mostly with t3 and t2 aeon torp bombers, others seems fine

added2: also happens with other torps if target moves in opposite direction

![1](https://user-images.githubusercontent.com/36897892/75625957-2dee2e80-5bd4-11ea-9e2f-482172dfbd56.png)

PR #3007, opened 2020-02-25
Fix torp bombers attack order + Solace buf

Strogoo:
#3001 
~Only affects t2 and t3 aeon torps, other factions works fine.~ t2 and t3 aeon + t2 sera torps. Description is in commit message

Solace buff will be added later, I just need this fix for upcoming retargeting system.

### DIFF
--- a/units/UAA0204/UAA0204_script.lua
+++ b/units/UAA0204/UAA0204_script.lua
@@ -20,6 +20,7 @@ UAA0204 = Class(AAirUnit) {
                         self:ChangeMaxRadius(100)
                         WaitTicks(1)
                         self:ChangeMaxRadius(self:GetBlueprint().MaxRadius)
+                        self:ResetTarget()
                     end)
                 end
                 AANDepthChargeBombWeapon.OnLostTarget(self)
--- a/units/UAA0204/UAA0204_script.lua
+++ b/units/UAA0204/UAA0204_script.lua
@@ -13,7 +13,18 @@ local AANDepthChargeBombWeapon = import('/lua/aeonweapons.lua').AANDepthChargeBo
 
 UAA0204 = Class(AAirUnit) {
     Weapons = {
-        Bomb = Class(AANDepthChargeBombWeapon) {},
+        Bomb = Class(AANDepthChargeBombWeapon) {
+            OnLostTarget = function(self)
+                if self.unit:GetTargetEntity().Dead then
+                    self:ForkThread(function()
+                        self:ChangeMaxRadius(100)
+                        WaitTicks(1)
+                        self:ChangeMaxRadius(self:GetBlueprint().MaxRadius)
+                    end)
+                end
+                AANDepthChargeBombWeapon.OnLostTarget(self)
+            end,
+        },
     },
 }
 
--- a/units/UAA0204/UAA0204_script.lua
+++ b/units/UAA0204/UAA0204_script.lua
@@ -1,0 +1,20 @@
+--****************************************************************************
+--**
+--**  File     :  /cdimage/units/UAA0204/UAA0204_script.lua
+--**  Author(s):  John Comes, David Tomandl, Jessica St. Croix
+--**
+--**  Summary  :  Aeon Torpedo Bomber Script
+--**
+--**  Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
+--****************************************************************************
+
+local AAirUnit = import('/lua/aeonunits.lua').AAirUnit
+local AANDepthChargeBombWeapon = import('/lua/aeonweapons.lua').AANDepthChargeBombWeapon
+
+UAA0204 = Class(AAirUnit) {
+    Weapons = {
+        Bomb = Class(AANDepthChargeBombWeapon) {},
+    },
+}
+
+TypeClass = UAA0204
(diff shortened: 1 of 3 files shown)


lua/sim/defaultweapons.lua
```diff
@@ -429,6 +429,16 @@ DefaultProjectileWeapon = Class(Weapon) {
         end
     end,
 
+    -- A workaround to fix a long delay in resetting attack order when target is dead
+    -- Happens to some torpedo bombers
+    ResetAttackOrder = function(self)
```
Uveso: PLease don't add a funtion to ALL projectiles if you only need this for 2.
It's not worth the memory it will cost to have this on every projectile.

Also please don't create a parent function with the only purpose to create a forked child function.
Please Fork the childfunction directly. (faster simspeed and better for debugging)

units/UAA0204/UAA0204_script.lua
```diff
@@ -13,7 +13,14 @@ local AANDepthChargeBombWeapon = import('/lua/aeonweapons.lua').AANDepthChargeBo
 
 UAA0204 = Class(AAirUnit) {
     Weapons = {
-        Bomb = Class(AANDepthChargeBombWeapon) {},
+        Bomb = Class(AANDepthChargeBombWeapon) {
+            OnLostTarget = function(self)
+                if self.unit:GetTargetEntity().Dead then
+                    self:ResetAttackOrder()
```
Uveso: Fork the thread directly here. Please don't call the forkthread with self:
self: will add the thread to the trashbag of the projectile.
In case the thread ist started and has changed the MaxRadius to 100 and the projectile will be destroyed a the same
time, then the forkedthread will be terminated and can't reset the MaxRadius to default.
So pleasee add this thread to the unit's trashbag.
```
                if self.unit:GetTargetEntity().Dead then
                    self.unit:ForkThread(function()
                        self:ChangeMaxRadius(100)
                        WaitTicks(1)
                        self:ChangeMaxRadius(self:GetBlueprint().MaxRadius)
                    end)
                end
```

lua/sim/defaultweapons.lua
Strogoo: Yeah, the initial plan was to add this one to all torp bombers (5 units). In this case you kinda want to have it in one place for easier editing. But now it can go to script file, I agree

units/UAA0204/UAA0204_script.lua
Strogoo: This one is from weapon class, not projectile. 
https://github.com/FAForever/fa/blob/develop/lua/sim/weapon.lua#L436
So it already goes to unit trashbag

lua/sim/defaultweapons.lua
shalkya: in the end it's not resquired for all torp bombers, because the others ones work fine ?

Strogoo: Hmm actually it might be an issue for other torps too. All my tests were vs idle ships, but what if they are moving in opposite direction. Like that:
image

it's just that aeon t3 torps (and t2 sometimes) have this bug even vs static targets. As for moving ones, it may appears for other factions too. I think I'll add this fix to other torps too

units/UAA0204/UAA0204_script.lua
Uveso: Oh yes you are right, i was looking at Projectile.lua for the ForkedThread function.

Uveso: Does the function `UnitWeapon:ResetTarget()` not work for this case?
Then we would not need a forked thread.

Strogoo: Nah, I tried it. As you can see we are already in Weapon:OnLostTarget() which is called by engine when weapon lose its target, so UnitWeapon:ResetTarget() won't change anything. Weapon already lost its target but unit still has attack order.

Strogoo:
Ok, Cyb and uef torps works fine even vs moving targets because of their fast projectiles. So only t2/t3 aeon and t2 sera torps need this fix.

Now it's time to start working on new retargeting system for Solace :)

Uveso:
After investigating this for some hours i am not satisfied with this solution.

The PR is working, but changing max weapon radius to 100 will lead to get targets outside the maximum weaponrange of 32. (happened several times while testing)

In case the main problem is not the weapon itself, but the sinking animation, (i did not test that)
Then patching all affected weapons is only a workaround, not a fix.

A fix would be to clone the (naval) unit on death, destroy the unit and let the clone sink.

Strogoo:
Well, I was thinking about that too, but 'proper' fix with cloning naval unit to fix attack order for 3 torp bombers sounds like an overkill for me. Also I think the bug you mentioned can be fixed with `Weapon:ResetTarget()` after weapon range is reverted to default.

Uveso:
I was testing the function with Weapon:ResetTarget() yesterday and got a stack overflow because
it can recall OnLostTarget in a loop.

Fixing a bug can't be a overkill ;)
Please hold this PR for a week.
I want at least one attempt to fix this a the base

Strogoo:
Yeah, calling :ResetTarget() without any condition inside :OnLostTarget() will cause stack overflow, but here we have:
```
if  self.unit:GetTargetEntity().Dead -- this will never be true after we call our thread once
    maxRange(100)
    1 tick
    maxRAnge(default)
    :ResetTarget() -- just in case weapon got target outside its range
```

Anyway, feel free to try something else, this PR won't be merged in near future anyway, cause I need to make new retargeting system for Solace first :)

Uveso:
@Strogoo 
bad news. My idea to create first the wreck and then make a sinkanimation does not work. :(

MrRowey:
@Tagada14 could you take a look at this PR for Torps and Solace

Tagada14:
Well, I see it but not much I can do here

Garanas:
I am closing this pull request. I understand that it holds value - someone can re-open it and pick it up in the future.

PR #3008, opened 2020-02-25
Fixed on/off of mods dependencies in ModsManager

KionX:
**Repeatedly** disabling dependencies did not disable dependent mods.
Fixed Esc when open from **main menu**. Not lobby.

### DIFF
--- a/lua/ui/lobby/ModsManager.lua
+++ b/lua/ui/lobby/ModsManager.lua
@@ -692,17 +692,12 @@ end
 -- Activates the mod with the given uid
 -- @param isRecursing Indicates this is a recursive call (usually pulling in dependencies), so should
 --                    not prompt the user for input.
--- @param visited The set of mods visited during this recursive set of calls (used to break cycles)
-function ActivateMod(uid, isRecursing, visited)
+function ActivateMod(uid, isRecursing)
     if mods.activated[uid] then
         return
     end
 
-    visited = visited or {}
-    if visited[uid] then
-        return
-    end
-    visited[uid] = true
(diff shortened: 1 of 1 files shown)


shalkya:
I need more explanation here.
I don't see any improvement compare to what we have right now. When you uncheck common mod tools, it already deactivate the dependent mods. When you checked a dependent mod and having common mod tools off, it would check it automatically.
When you are choosing your mods, you can already press esc to come back to lobby.

KionX:
Corrected the description.

shalkya:
"Fixed Esc when open from main menu. Not lobby." that i was able to test without issue.
"Repeatedly disabling dependencies did not disable dependent mods." that i still don't get, @Uveso maybe you can explain that to me ? maybe it's not for ui mods, i've tested only with common mod tools.

KionX:
Here are a few scenarios.
1. EcoManager on, Common Mod Tools off.
2. EcoManager off, EcoManager on, Common Mod Tools off.
3. Common Mod Tools off.

Uveso:
@shalkya 
`When you uncheck common mod tools, it already deactivate the dependent mods.`
Exactly this was NOT always the case.


Lets say you have `common mods tools` and `UI Party` on.
If you disable `common mods tools`, then also `UI Party` is off. (like expected)

Same again, `common mods tools` and `UI Party` are on.
First disable and enable `UI Party`.
Now, if you disable `common mods tools`, `UI Party` is still on.

shalkya:
alright i see now.
Tested works well.

PR #3009, opened 2020-02-25
Moved Mod Manager to the top menu. Reduced sound noise.

KionX:


lua/ui/menus/main.lua
```diff
@@ -124,9 +124,10 @@ function CreateUI()
             color = menuFontColorAlt,
         },
         {
-            name = '<LOC tooltipui0355>Extras',
-            tooltip = 'mainmenu_extras',
-            action = function() ButtonExtras() end,
+            name = '<LOC _Mod_Manager>',
```
speed2CZ: There is no reason for this change, most of the time player will get to the mod manager through the lobby anyway. + you kill the option to check credits and eula. So not this way

KionX: Who is interested in check credits and eula?
They are available in vanilla FA.

shalkya:
removing the sound is good.
To find a middle ground would it be possible to keep "extra" and to just add "mod manager" under ? With some repositioning it should fit.

Uveso:
Making the menu bigger does not fit into the mainscreen:
Mainmenu

Uveso:
Working now!
Menu fits to the screen and is working.

PR #3012, opened 2020-02-28
Misc Performance Enhancements

Stratege:
This PR contains 2 key changes
- Removing the asserts that happen on state changes (this happens ~8 times or something like that per weapon firing cycle)
- Caching the projectile data in the weapon (and correctly invalidating it) instead of recreating it on every shot

Both changes have been tested in my Simspeed++ mod for a few months now and have not caused any issue.

Getting good data on how much it helps is hard, for comparison for the sake of this PR I ran replay 11042968 at the max speed with and without those changes in it.
Without them it took 29min 30sec
With them it took 27min 30sec
Giving a roughly 7% speedup in a real game (a fairly short and casual ~1.2k setons)

### DIFF
--- a/lua/sim/weapon.lua
+++ b/lua/sim/weapon.lua
@@ -315,7 +315,8 @@ Weapon = Class(moho.weapon_methods) {
 
     end,
 
-    GetDamageTable = function(self)
+
+    GetDamageTableInternal = function(self)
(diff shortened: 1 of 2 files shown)


shalkya:
no errors.
I'm wondering if removing assert like this can let some bugs slip through, but i trust uveso.
good job.

PR #3017, opened 2020-03-01
Add support for custom map Props

Exotic-Retard:
This requires a change to the init file. running this without the init
file should result in no changes - nothing breaks but no new features.

This adds support for custom props, and more to maps, and allows passing
arbitrary data to blueprints.lua, which is insanely useful and has been
a limitation for a long time.

Currently this is only used to load the custom content in the current
map, but mods and maps could do loads of stuff with it. Such as:
1. adding mod settings to turn off/on changes
2. adding map settings for optional features, and props.
3.mods can edit the logic by hooking modblueprints, and so can have
total control over which parts of themselves get loaded.

a note on security - this might sound like it creates a hole but it
doesnt let you break anything that couldnt be broken before:
game preferences is arbitrarily readable from UI already,
no writing functionality that wasnt there before has been added.
any non-deterministic change in blueprints.lua results in desyncs, just
like before

### DIFF
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -627,6 +627,7 @@ end
 --- Sim -> LoadBlueprints() - no arguments, no changes!
 --- UI  -> LoadBlueprints('*_unit.bp', {'/units'}, mods, true, true, true, taskNotifier)  used in ModsManager.lua
 --- UI  -> LoadBlueprints('*_unit.bp', {'/units'}, mods, false, true, true, taskNotifier) used in UnitsAnalyzer.lua
+
 function LoadBlueprints(pattern, directories, mods, skipGameFiles, skipExtraction, skipRegistration, taskNotifier)
 
     local task = 'Blueprints Loading... '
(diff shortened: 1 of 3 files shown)


lua/system/Blueprints.lua
```diff
@@ -668,6 +669,25 @@ function LoadBlueprints(pattern, directories, mods, skipGameFiles, skipExtractio
...
+        task = 'Blueprints Loading: Blueprints from current map'
+        files = DiskFindFiles(PreGameData.CurrentMapDir, pattern)
+        for k,file in files do
+            BlueprintLoaderUpdateProgress()
+            -- update UnitManager UI via taskNotifier only if it exists
+            if taskNotifier then
+               taskNotifier:Update(task, total, k)
+            end
+            
+            safecall(task .. ': ' .. file, doscript, file)
```
speed2CZ: This allows to load unit BPs as well, so some smart individual could sneak in a unit bp change with the map, I dont think we should allow this.

shalkya:
didn't test it with custom props.
didn't trigger any error/warning (once i modified the init).

PR #3020, opened 2020-03-03
Massive localization fixes

KionX:
Updated RU localization
(translated Lobby Background, Notify Management, Keybindings dialogs, misc).
Updated US localization(added missing text)
When loading missing lines appends from US.
Fixed misc LOC-related errors.
Added Tooltips autosize(up to 250px) when width missing.
Fixed https://github.com/FAForever/fa/issues/3026


shalkya:
I'm not really convinced about this change, i will let other post their pov, here is an example of the change : 
image
instead of this : 
image
It was way more concise before, i feel like it was better.

speed2CZ:
From the original code: `if not width then width = 150 end` and that's what you see with the most of th tooltips. As that's a width that works just fine. So it was never missing, it was intended to be like this. I also prefer the way it is right now. It's compact and shows exactly where you expect it. So don't change that.

KionX:
250px ought to be enough for anybody©

speed2CZ:
Where is 150 not enough?

KionX:
Localization is not limited to the US language.

speed2CZ:
Im asking about the width of the tooltips. Afaik it just makes more lines if needed.

KionX:
There are formatted tooltips. For them, 150px is usually not enough.
US

RU

Another problem:
Hotkey didn't fit in 150px.

KionX:
This is the final... I hope.

shalkya:
seems fine for me, i'll merge in the next days unless someone has something to say against.

speed2CZ:
I do, please split it into two commits, one fixing the loc, other adjusting the tooltip code, else it's good.

PR #3023, opened 2020-03-07
Adding more restrictions to sim orders

Strogoo:
- dissalow sim attack-move for engies. Used by attack-move button, so not a big deal as most of the players use alt+RMB anyway.
- 10 sec delay between shift-g move orders.
- No delay for attack orders. I feel like it also can be abused in some way, so mb we should add 5 sec delay for attack too?

If you have better idea how to break ui mods such as auto attack-move engies, auto flanking, auto dodging etc., feel free to make another PR with it. (removing shift-g feature is not an option :)

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -258,11 +258,23 @@ function IsInvalidAssist(unit, target)
     end
 end
 
+-- This one is used for attack move button which was added in FAF.
+-- I think 90% of players use alt+RMB instead? So for now we exclude engies,
+-- but if some auto abuse will appear again, then better remove it completely.
 Callbacks.AttackMove = function(data, units)
-    if data.Clear then
-        IssueClearCommands(units)
+    units = SecureUnits(units)
+    local validatedUnits = {}
+
+    for k, unit in units do
+        if not unit:GetBlueprint().CategoriesHash.ENGINEER then
+            table.insert(validatedUnits, unit)
+        end
     end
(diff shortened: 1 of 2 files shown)


shalkya:
can you add a little message telling the user to wait until the end of the delay when he tries to spam shift g ?
can you describe how this would apply ? can you use shift g, 3 sec after the first shift g, when using other units ?
also i feel like adding delay on attack orders would lead to a lot of issues, especially to slow fire units. if you miss click the attack order, you can still rapidly switch to another unit. Also 5 sec can be game deciding, so it feels like that it would bring more issues than benefice.

Strogoo:
Yeah, attack is fine without delay I think. Also shift-g move delay is one for all units. I can make a separate delay for every unit np, idk if it will be better and more userfriendly. Like if you select 10 tanks, then shift-g them, then drag select same 10 tanks + 5 more and shift-g them to another place, and these 5 new will get shift-g orders and for old selection it will be aborted. Kinda confusing

shalkya:
yeah it's true but the idea of doing this restriction on shift g, is that you usally don't want to give 2 shift g order on the same unit in a 10 sec interval. But you can give a shift g on some hover on the south beach of seton, then switch to the top ocean and order a shift g for your frigates and be in that time frame.
I don't use shift g, i will ask if what you did is annoying, or if usually it won't annoy the player.

5cript:
Sounds like this breaks ACU blocking with units.

PR #3028, opened 2020-03-17
[WIP] TMM: Add startspot command line option to autolobby

Askaholic:
Merging in some of the changes from the galactic war mod found here: 
https://github.com/FAForever/fa-gw/blob/6d49c53f8e5214fc6713cea3abbc52567d584a5d/mods/galacticWar/hook/lua/ui/lobby/autolobby.lua

Adds the optional `/startspot` command line option for setting a player's `StartSpot`. If this is not provided it will fall back to using the old logic of sequential numbering.

I'm not sure what the `GpgNetSend('PlayerOption'...)` calls are designed to do. Are they needed for other clients to get the right parameters? I see that on the galactic war mod they are removed completely.


Askaholic:
@shalkya should this be rebased on `deploy/fafdevelop`?

shalkya:
yes you need to change the target aswell

PR #3030, opened 2020-03-18
rebalaces sera scu

TurinturambarFAF:
sera scu rework. rebalnces upgrade costs, removes shiedupgrade, splits sensor upgrade into a seperate sensor and a gunrange upgrade and adds a regenfield support upgrade.


shalkya:
in the end i didn't touch the upgrades in general. Since it's better to split commit, and leave this one as a merge conflict. You need to redo the preset though.

PR #3032, opened 2020-03-23
Issue/#3031 Don't send results after initiating a `SafeQuit`.

### LINKED ISSUE #3031: 'Exit to FAF' causes incorrect game results (and therefore rating issues)
While investigating the rating issues I noticed a strange pattern. Games which were not rated correctly seemed to always have a player who randomly quit before their ACU died. 

As it turns out exiting the game via the `Exit to FAF` causes the game to send a `defeat` result for every player in the game except yourself. This really messes with the rating code!

## To reproduce
1. Load up a game with at least 2 players
2. Build some buildings/unit (It seems like if you don't build anything then it does not send incorrect results)
3. One player hits `Exit to FAF`
Now if that player checks their game.log they will see that they did not report any results for themselves, but reported `defeat` for the other players in the game.

I've been trying to debug this and I haven't found the root cause of the issue. At first I though it was caused by the checks in `victory.lua` somehow determining that all players had lost due to faulty logic. However, when I add some debug log statements, it seems that this code is not what is triggering the game results. So it seems to me like `SessionEndGame()` must be setting the results of other players to `defeat` internally which is then being picked up by UserSync.lua.

Then again I don't have any experience with the lua code or the game internals so I imagine someone else will be able to debug this quicker than me.

Askaholic:
Fixes #3031.

Works like a charm. If our instance of FA closes via `Exit to FAF` then that will be picked up by the other players in the game and our `defeat` result will be reported appropriately, but we won't be spamming bogus `defeat` results for the other remaining players.

### DIFF
--- a/schook/lua/UserSync.lua
+++ b/schook/lua/UserSync.lua
@@ -96,7 +96,7 @@ OnSync = function()
     end
 
     if not import("/lua/ui/dialogs/eschandler.lua").isExiting then
-        for k,gameResult in Sync.GameResult do
+        for _, gameResult in Sync.GameResult do
             local armyIndex, result = unpack(gameResult)
             LOG(string.format('Sending game result: %i %s', armyIndex, result))
             GpgNetSend('GameResult', armyIndex, result)
--- a/schook/lua/UserSync.lua
+++ b/schook/lua/UserSync.lua
@@ -95,11 +95,13 @@ OnSync = function()
         end
     end
 
-    for k,gameResult in Sync.GameResult do
-        local armyIndex, result = unpack(gameResult)
-        LOG(string.format('Sending game result: %i %s', armyIndex, result))
-        GpgNetSend('GameResult', armyIndex, result)
-        import('/lua/ui/game/gameresult.lua').DoGameResult(armyIndex, result)
+    if not import("/lua/ui/dialogs/eschandler.lua").isExiting then
+        for k,gameResult in Sync.GameResult do
+            local armyIndex, result = unpack(gameResult)
+            LOG(string.format('Sending game result: %i %s', armyIndex, result))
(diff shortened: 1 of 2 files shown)


schook/lua/UserSync.lua
```diff
@@ -95,11 +95,13 @@ OnSync = function()
         end
     end
 
-    for k,gameResult in Sync.GameResult do
-        local armyIndex, result = unpack(gameResult)
-        LOG(string.format('Sending game result: %i %s', armyIndex, result))
-        GpgNetSend('GameResult', armyIndex, result)
-        import('/lua/ui/game/gameresult.lua').DoGameResult(armyIndex, result)
+    if not import("/lua/ui/dialogs/eschandler.lua").exiting then
```
speed2CZ: exiting != isExiting

Askaholic: Oh yea. Thanks

schook/lua/UserSync.lua
```diff
@@ -95,11 +95,13 @@ OnSync = function()
         end
     end
 
-    for k,gameResult in Sync.GameResult do
-        local armyIndex, result = unpack(gameResult)
-        LOG(string.format('Sending game result: %i %s', armyIndex, result))
-        GpgNetSend('GameResult', armyIndex, result)
-        import('/lua/ui/game/gameresult.lua').DoGameResult(armyIndex, result)
+    if not import("/lua/ui/dialogs/eschandler.lua").isExiting then
+        for k,gameResult in Sync.GameResult do
```
norraxx: `for k, gameResult in ...` :see_no_evil:

speed2CZ: `for _, gameResult in ...` 🙈

PR #3039, opened 2020-03-30
Deleting garbage

KionX:


speed2CZ:
Dude, it's not garbage, I've added those files with the UI scaling as they needed adjustments to scale properly. Look it up in their history.

KionX:
This is not used by the game - garbage.

speed2CZ:
There are 3 UI game layouts, left, middle, right, cycle through them using Atl + Left/Right or Up Down

PR #3047, opened 2020-04-04
Improved usability Cheat menu

KionX:
When opened it centered on the mouse cursor.
Tuning window size and spawn units timings.
Fixed bug when spawning from an observer.


shalkya:
didn't find the bug with spawning unit as obs
anyway it works otherwise, no errors

KionX:
First you need to switch to the observer. Then open Alt+F2 and create a unit without selecting an army. A very specific bug.

PR #3049, opened 2020-04-05
Removed redundant voice messages "group found" in lobby

KionX:
Now voice message play only when the last open slot is occupied.


JaggedAppliance:
It's nice to know that people are joining the game and when it's full you can see it clearly in the taskbar right?

PR #3054, opened 2020-04-08
percy target ACU over T1/T2

shalkya:


speed2CZ:
Other faction T3 units have this as well?

shalkya:
well percy is the worst with its firing cycle. Recently i had trouble with ~10-15 percies against a shield acu and a spam of T1/T2 and few T3

shalkya:
included into #3072

PR #3064, opened 2020-04-09
Window Resize Fix

KionX:
To reproduce the error, place the minimap on the edge of the window and reduce,
and then increase window size.
The right side of the map sticks to the window and stretches with it.
The chat window behaves the same way.

### DIFF
--- a/lua/maui/window.lua
+++ b/lua/maui/window.lua
@@ -72,19 +72,19 @@ styles = {}
 Window = Class(Group) {
     __init = function(self, parent, title, icon, pin, config, lockSize, lockPosition, prefID, defaultPosition, textureTable)
         Group.__init(self, parent, 'window')
-        
+
         self:DisableHitTest()
-        
+
         self._resizeGroup = Group(parent, 'window resize group')
         LayoutHelpers.FillParent(self._resizeGroup, self)
         self._resizeGroup.Depth:Set(function() return self.Depth() + 100 end)
         self._resizeGroup:DisableHitTest()
         self._pref = prefID
-        
+
         self._windowGroup = Group(self, 'window texture group')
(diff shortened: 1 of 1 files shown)


speed2CZ:
Split it please into two commits, one with a cleanup and one with the fix, so in the history you can see right aways what got changed.

shalkya:
i don't see the bug.
Maybe you are talking about increasing the size of the windows on the right, that scale the map while the bottom of the windows don't scale, so you end up zooming. Although that's not fixed with your pr.

KionX:
The right side of the map sticks to the window and stretches with it.
The chat window behaves the same way.
Do not merge it. I want to try to make this fix in a different way.

shalkya:
So one thing can be an issue : 
If you kind of fucked up the minimap previous the fix, and then switch to the fix, you get that : 
image
minimap is on the left, i can resize it, but part of it is outside the windows, and i can't move it around because of that. On top of that it is obviously broken since it's the minimap (you can move around the map click inside) but it doesn't display the minimap (actually show the game behind).

KionX:
```If you kind of fucked up the minimap previous the fix```
Correcting extreme cases was not the goal of this PR.
I managed to fix this by stretching the minimap to the edge of the window then shrinking the window.

shalkya:
ok the bug i linked isn't caused by your PR, it's something due to my game being launched in 2 different resolution depending if i use the exe for test or not. (Pretty weird one if you ask me, that i should fix at some point, maybe just changing the exe of testing) Resulting into test game keeping the size of the other resolution and ending up with some part of the mini map out of the game.
PR seems fine to me.

KionX:
The minimap might break if the resolution was changed multiple times. With this fix, this should stop.

PR #3065, opened 2020-04-10
Fix bigger strat icons

speed2CZ:
The whole code for the bigger strategic icons is a ugly hook from some mod instead of a proper implementation, so I'm not ashed with this solution for now.


shalkya:
tested : works well and no errors

PR #3067, opened 2020-04-10
Sera SCU changes

TurinturambarFAF:
part of SCU rework - rebalances Sera SCU.


Apofenas:
How about RAS upgrade for SCU for once during these years people were asking for it?

shalkya:
you can see changes about ras here : https://github.com/FAForever/fa/pull/2866

PR #3073, opened 2020-04-13
Tele delay fix

Strogoo:
broken by this https://github.com/FAForever/fa/commit/2bc7e86fdb581a48937a397fe8b00a02b511ddd9

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2321,16 +2321,17 @@ CommandUnit = Class(WalkingLandUnit) {
         self:PlayUnitSound('TeleportStart')
         self:PlayUnitAmbientSound('TeleportLoop')
         
-        local bp = self:GetBlueprint().Economy
-        local energyCost, time
-        if bp then
-            local mass = (bp.TeleportMassCost or bp.BuildCostMass or 1) * (bp.TeleportMassMod or 0.01)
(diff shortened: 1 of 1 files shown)


lua/defaultunits.lua
```diff
@@ -2330,7 +2330,7 @@ CommandUnit = Class(WalkingLandUnit) {
             time = energyCost * (bp.TeleportTimeMod or 0.01)
         end
 
-        local teleDelay = bp.General.TeleportDelay
```
speed2CZ, line 2333: There's already GetBlueprint() call on line 2324, combine them please

Strogoo: Yeah I saw it, just didn't want to touch `mass=` and `energy=` lines and left them short and clear. Also it's pretty cold code which is called 20 times per game at max.

Strogoo: nvm, I have bigbrain idea:
bp = self:GetBlueprint()
bpeco = bp.Economy

:+1:

speed2CZ: Yeah that's better. But never think "it's just few times" since it adds up with all the rest of the code. All the things Ive changed in the last patch were small things, on its own basically no improvement, but together it was several % sim speed boost.

Strogoo: yeah, true

PR #3074, opened 2020-04-13
alternative hotbuild

shalkya:
How it works : select the factory, press the hotkey to cycle through the template, and press middle mouse click to select the template. If there is only 1 template available, you don't need to press the middle click.

I need help for this one. It's nearly functional.

Issue is : 
```
function resetCycle(commandMode, modeData)
    -- Commandmode = false is when a building is built (left click with mouse)
    -- modeData.isCancel = false is when building is aborted by a right click... whyever
    -- modeData.isCancel = true when "canceling" by releasing shift
    if commandMode == false or modeData.isCancel == false then
        cyclePos = 0
    -- Set to 0, first one is 1 but it will be incremented!
    end
end
```
when you select a building, this function is spammed every tick or so. So cyclePos is reseted to 0. This cyclePos is used to select the templates. If i can't increment it, i can't really select other template than the 1st one.

can you guys help me on this one ? thanks.


lua/keymap/hotbuild.lua
```diff
@@ -266,6 +267,131 @@ function buildActionBuilding(name, modifier)
...
+        -- if currentCommandQueue then
+            -- for index = table.getn(currentCommandQueue), 1, -1 do
+                -- local count = currentCommandQueue[index].count
+                -- if index == 1 then
+                    -- count = count - 1
+                -- end
+                -- DecreaseBuildCountInQueue(index, count)
+            -- end
+        -- end
+    -- end
```
norraxx, line 285: why do you need it? if you don't need, don't leave this code here

lua/keymap/hotbuild.lua
```diff
@@ -266,6 +267,131 @@ function buildActionBuilding(name, modifier)
...
+            local fadeTime = options.hotbuild_cycle_reset_time / 2000.0
+            WaitSeconds(stayTime)
+            if not cycleMap:IsHidden() then
+                Effect.FadeOut(cycleMap, fadeTime, 0.6, 0.1)
+            end
+            WaitSeconds(fadeTime)
+            cyclePos = 0
+        end)
+    else
+        cycleThread = ForkThread(function()
```
norraxx, line 359: will you ever end this threads?

lua/keymap/unitkeygroups.lua
```diff
@@ -91,6 +91,9 @@ unitkeygroups = {
 
         "T2_Cruiser",
     },
+    ["TemplateFactory"] = {
+        "_factory_templates", -- Special :)
```
norraxx, line 95: what does it mean "Special :)" ?

lua/keymap/hotbuild.lua
```diff
@@ -266,6 +267,131 @@ function buildActionBuilding(name, modifier)
...
+                        local count = units.count
+                        IssueBlueprintCommand("UNITCOMMAND_BuildFactory", v, count)
+                    end
+                end
+            end
+        end
+    end
+    
+
+    -- WaitSeconds(5)
```
norraxx, line 391: why do you need that code? if this is dead code, delete it.

lua/keymap/hotbuild.lua
```diff
@@ -266,6 +267,131 @@ function buildActionBuilding(name, modifier)
...
+                button:SetAlpha(1, true)
+            else
+                button:SetAlpha(0.4, true)
+            end
+        end
+        cycleMap:Show()
+        -- Start the fading thread
+        cycleThread = ForkThread(function()
+            local stayTime = options.hotbuild_cycle_reset_time / 2000.0
+            local fadeTime = options.hotbuild_cycle_reset_time / 2000.0
```
norraxx, line 350: no need to compute same twice, reuse variable

lua/keymap/hotbuild.lua
```diff
@@ -266,6 +267,131 @@ function buildActionBuilding(name, modifier)
     CommandMode.StartCommandMode("build", {name = cmd})
 end
 
+function buildActionFactoryTemplate(modifier)
```
norraxx, line 270: what is that `modifier` used for? Imho it's unused at all, don't use this variable if it's not needed

lua/keymap/hotbuild.lua
```diff
@@ -266,6 +267,131 @@ function buildActionBuilding(name, modifier)
...
+    local effectiveTemplates = {}
+    local effectiveIcons = {}
+    local allFactoryTemplates = FactoryTemplates.GetTemplates()
+
+    if (not allFactoryTemplates) or table.getsize(allFactoryTemplates) == 0 then
+        return
+    end
+
+    local selection = GetSelectedUnits()
+    local availableOrders,    availableToggles, buildableCategories = GetUnitCommandData(selection)
```
norraxx, line 300: is this `space formatting` part of the LUA language? Or just it's a lot of spaces?

lua/keymap/hotbuild.lua
norraxx: I'd wish, you'll leave here comment, what your function does.

lua/keymap/hotbuild.lua
```diff
@@ -266,6 +267,131 @@ function buildActionBuilding(name, modifier)
...
+        -- end
+    -- end
+
+    -- Reset everything that could be fading or running
+    hideCycleMap()
+
+    -- Find all avaiable templates
+    local effectiveTemplates = {}
+    local effectiveIcons = {}
+    local allFactoryTemplates = FactoryTemplates.GetTemplates()
```
norraxx, line 293: how can this function return null or empty dict?

lua/keymap/hotbuild.lua
norraxx: if not, what will happen with that function? If yes, which i don't see, why do you need this as variable?

lua/keymap/hotbuild.lua
shalkya: i might use it afterward, when i fix the main issue, it's still work in progress

lua/keymap/hotbuild.lua
shalkya: it is used when you want to use 'shift' or 'alt' in combinaison with the hotkey
for example you can use 'alt' in the commented code afterward, to remove the whole unit queue in factory, apart from the unit currently building, and add on top the unit template
for 'shift' it can be used to increase by 5 the number of unit queue with the template for example (i don't think it's useful in this case)

lua/keymap/hotbuild.lua
shalkya: it can return null or empty dict, if you didn't create any factory template. When you have created some, they are saved in game.prefs. 
FactoryTemplates.GetTemplates() is used to grab the template from the game.prefs

lua/keymap/hotbuild.lua
shalkya: idk that's just some copy pasta from another part of the code. I think it's not needed, and will remove that

lua/keymap/hotbuild.lua
shalkya: same i copy pasta another part of the code. I guess it is made this way so you can tweak them if needed afterward.

lua/keymap/hotbuild.lua
shalkya: yes it is code i used for testing, i will remove that

lua/keymap/unitkeygroups.lua
shalkya: i kept it the same as https://github.com/FAForever/fa/pull/3074/files#diff-23fd3b3e2c4fb907846b53ea8f1915daR84
i guess what it means is that these 2 used specific function from the lua/keymap/hotbuild.lua

lua/keymap/hotbuild.lua
shalkya: i think i'll have to do a refactoring anyway, since this function use a lot of code from another function.

lua/keymap/hotbuild.lua
shalkya: hideCycleMap() at the very beginning of the function. in case the function is called again, the fading is interrupted.

function hideCycleMap()
    if (cycleThread) then
        KillThread(cycleThread)
    end
    cycleMap:SetNeedsFrameUpdate(false)
    cycleMap.OnFrame = function(self, elapsedTime) end
    cycleMap:Hide()
    cycleMap:SetAlpha(1, true)
end

shalkya:
```
function resetCycle() 
    local selection = GetSelectedUnits()
    local worldview = import('/lua/ui/game/worldview.lua').viewLeft
    local oldHandleEvent = worldview.HandleEvent
    worldview.HandleEvent = function(self, event)
        if event.Type == 'ButtonPress' then
            if selection[1]:IsInCategory('FACTORY') then
                if event.Modifiers.Right or event.Modifiers.Shift then
                    return
                else
                    cyclePos = 0
                end
            else
                if event.Modifiers.Middle or event.Modifiers.Shift then
                    return
                else
                    cyclePos = 0
                end
            end
        end
    end
end
```
tried this, but issue is that resetCycle is called in init() from gamemain : 
```
function init()
    unitkeygroups = getUnitKeyGroups()
    initCycleMap()
    local selection = GetSelectedUnits()
    -- CommandMode.AddEndBehavior(resetCycle)
    resetCycle()
end
```
which means this is called all the time, even if we don't have units selected. Either i need to find a way to not call the init(); or you have a better idea. (also when you have a unit selected, GetSelectedUnits() returns nil half the time, i guess that's another issue with being called in init() )

norraxx:
I'm offering to focus on code quality. That means not to format code well, but know exact what each line does. Especially, how threads works.

shalkya:
ok this is in functional state. It is a bit hacky `CommandMode.StartCommandMode("build", {name = ''})` means that the facto act like it's going to build like an engineer. 
This is required to allow the factory to cycle through the different template. Modifying the commandmod.lua is way more complicated and would lead to more issue (since function are called all the time).
And it doesn't seems to have any drawback (no change visually or errors).

shalkya:
known bug : after pressing the hotkey, you can't deselect the factory by pressing 'left click'

shalkya:
posting that here for the lazy people : https://youtu.be/deupInNKc28

shalkya:
last commit is controversial, but i think it's beneficial for the community as a whole , and the hotkey can be relearnt by the veteran.
Benefit are : 
- simplification for newcomers : 
    - less hotkey required : with 3 hotkeys (mass, pgen, factory), you can access the most build structures and units (60% of the units actually used in game). Add on top the AA hotkey and you reach 80% easily. The overall number of hotkey is reduced.

    - grouping of unit makes more sense : AA are grouped together, stealth are grouped together, shields are grouped together, missile units are grouped together, scouts/radar are grouped together, torpedo units are grouped together, raiding/kiting units are grouped together, main army units are grouped together.

- the hotkey for factories use the same logic as for engies. You cycle through them and select the one you want to build.


Known bug : 
- MAJOR : when using factory hotkey, you aren't able to build anything with engies afterward. `CommandMode.StartCommandMode("build", {name = ''})` which is the trick for allowing the cycle to work, also causes this bug. It is required to find a way to cancel this command after units are ordered.
- f1 doesn't detect the renaming of hotkeys for hotbuild in the keydescriptions.lua
- using hotkey with Megalith to build units, lead to Megalith building directly underneath it (without having to position the units to build). It's not so much of an issue, but the mega takes some time to rebuild a unit after 1st one is done.

shalkya:
i vaguely remember people being against e storage being in hotkey (because they would build it by mistake) thus why we have a mod for it. I think this is a non-issue, and that not everyone is awared of this kind of mod.

shalkya:
It's because they prevent from updating the hotkey description that are in keydescriptions.lua
The `LOC(KeyDescriptions[action])` in the following function from keymapper.lua returned old description. Although `KeyDescriptions[action]` had the good description.
```
function GetActionName(action)
    local name = ''
    if KeyDescriptions[action] then
        name = LOC(KeyDescriptions[action])
    else
        name = LOC("<LOC key_binding_0001>No description text") .. ' for action ' .. action
    end
    -- check if action is meant to be mapped with a key modifier, e.g. attack vs shift_attack action
    if string.find(action, 'shift_') == 1 then
        name = name .. ' - SHIFT version'
    end
    return name
end
```

lua/keymap/hotbuild.lua
```diff
@@ -167,6 +169,88 @@ function init()
...
+        cycleThread = ForkThread(function()
+            WaitSeconds(options.hotbuild_cycle_reset_time / 1000.0)
+            cyclePos = 0
+        end)
+    end
+end
+
+function cycleUnits(maxPos, name, effectiveIcons, selection)
+    -- Check if the selection/key has changed
+    if cycleLastName == name and cycleLastMaxPos == maxPos and old_selection == selection[1] then
```
norraxx, line 208: Don't you want to check for value in `old_selection` then compare `selection[1]` against that value?
Otherwise, i suppose, you're comparing against `nil`, which is not "good" practice, you're hiding that fact, that variable can be undefined (`nil`).

lua/keymap/hotbuild.lua
```diff
@@ -23,6 +24,7 @@ local cycleThread = false
 local cycleLastName
 local cycleLastMaxPos
 local cycleButtons = {}
+local old_selection
```
norraxx, line 27: From what i've seen, lua is camelCase language, so snake_case let be for python :-)

lua/keymap/hotbuild.lua
```diff
@@ -167,6 +169,88 @@ function init()
...
+    local options = Prefs.GetFromCurrentProfile('options')
+    if options.hotbuild_cycle_preview == 1 then
+        -- Highlight the active button
+        for i, button in cycleButtons do
+            if i == cyclePos then
+                button:SetAlpha(1, true)
+            else
+                button:SetAlpha(0.4, true)
+            end
+        end
```
norraxx, line 183: Can we optimize cycle over buttons?
Maybe we could acces directly? You know, `cyclePos`, I suppose you could know the previous selected button, and I suppose `cycleButtons` is the map. So what i'm offering is to swap alpha only at changed elements.

BTW: could it be, that previous cyclePos is the previous one item?

lua/keymap/hotbuild.lua
```diff
@@ -167,6 +169,88 @@ function init()
...
+
+        cycleMap:Show()
+        -- Start the fading thread
+        cycleThread = ForkThread(function()
+            local stayTime = options.hotbuild_cycle_reset_time / 2000.0
+            local fadeTime = options.hotbuild_cycle_reset_time / 2000.0
```
norraxx, line 189: Imho, `stayTime` and `fadeTime` could be one variable.

lua/keymap/hotbuild.lua
```diff
@@ -167,6 +169,88 @@ function init()
...
+            
+            WaitSeconds(stayTime)
+            if not cycleMap:IsHidden() then
+                Effect.FadeOut(cycleMap, fadeTime, 0.6, 0.1)
```
norraxx, line 193: I'm thinking, do we need that fading in and out? Like usualy it's slow and heavy operation.

lua/keymap/hotbuild.lua
```diff
@@ -179,7 +263,7 @@ function buildAction(name)
     local selection = GetSelectedUnits()
     if selection then
         -- If current selection is engineer or commander
-        if table.getsize(EntityCategoryFilterDown(categories.ENGINEER - categories.STRUCTURE, selection)) > 0 then
+        if table.getsize(EntityCategoryFilterDown(categories.ENGINEER - categories.STRUCTURE or categories.ENGINEER - categories.STRUCTURE, selection)) > 0 then
```
norraxx, line 266: `categories.ENGINEER - categories.STRUCTURE or categories.ENGINEER - categories.STRUCTURE` == `condition a or condition a` O_o

lua/keymap/hotbuild.lua
```diff
@@ -217,55 +301,73 @@ function buildActionBuilding(name, modifier)
...
+    
+    if maxPos == 1 then
+        for _, units in selectedTemplate do
+            local v = units.id
+            local count = units.count
+            IssueBlueprintCommand("UNITCOMMAND_BuildFactory", v, count)
+        end
+    else
+        CommandMode.StartCommandMode("build", {name = ''})
+        local worldview = import('/lua/ui/game/worldview.lua').viewLeft
```
norraxx, line 353: I'm offering to import this file forward. at the beginning

lua/keymap/hotbuild.lua
```diff
@@ -437,9 +486,16 @@ function hideCycleMap()
     cycleMap:SetAlpha(1, true)
 end
 
-function buildActionUnit(name, modifier)
+function buildActionUnit(name, modifier)
```
norraxx, line 489: `    ` at the end of the line

lua/keymap/hotbuild.lua
```diff
@@ -454,27 +510,58 @@ function buildActionUnit(name, modifier)
...
+                if event.Modifiers.Middle then
+                    IssueBlueprintCommand("UNITCOMMAND_BuildFactory", unit, count)
+                    CommandMode.EndCommandMode(true)
+                elseif event.Modifiers.Left then
+                    CommandMode.EndCommandMode(true)
+                end
+            end
+        end
+    end 
 end
```
norraxx, line 565: Why do I see copy-paste kung-fu? O_o. What's the difference between `buildActionUnit` and `buildActionFactoryTemplate`? Why do we need code copies?

lua/keymap/keydescriptions.lua
```diff
@@ -5,426 +5,423 @@
...
+    ['t3_submarine_hunter'] = 'build T3 Submarine Hunter',
+    ['t3_mobile_aa'] = 'build T3 Mobile Anti Air',
+    ['t2_support_factory'] = 'build T2 Support Factory',
+    ['t2_support_land_factory'] = 'build T2 Support Land Factory',
+    ['t2_support_air_factory'] = 'build T2 Support Air Factory',
+    ['t2_support_naval_factory'] = 'build T2 Support Naval Factory',
+    ['t3_support_factory'] = 'build T3 Support Factory',
+    ['t3_support_land_factory'] = 'build T3 Support Land Factory',
+    ['t3_support_air_factory'] = 'build T3 Support Air Factory',
+    ['t3_support_naval_factory'] = 'build T3 Support Naval Factory',
```
norraxx, line 426: `<LOC_key_desc_` R U SURE? They're here for some purpose, i suppose `<LOC` mean I18n strings

lua/keymap/unitkeygroups.lua
```diff
@@ -8,174 +8,182 @@ unitkeygroups = {
         "Air_Factory",
         "Naval_Factory",
         "Quantum_Gateway",
+        "xrl0002", -- T3 engineer for Megalith
```
norraxx, line 11: this naming is wierd

lua/keymap/unitkeygroups.lua
```diff
@@ -8,174 +8,182 @@ unitkeygroups = {
...
     ["Pgen"] = {
         "Power_Generator",
         "Hydrocarbon_Power_Plant",
-
-        "T2_Heavy_Tank",
-
-        "T2_Torpedo_Bomber",
-
-        "T2_Destroyer",
+		"Energy_Storage",
```
norraxx, line 101: space

lua/keymap/unitkeygroups.lua
```diff
@@ -8,174 +8,182 @@ unitkeygroups = {
...
-        "url0303", -- Loyalist
-        "uel0303", -- Titan
-
-        "T3_Spy_Plane",
-
-        "T3_Strategic_Missile_Submarine",
-        "T3_Submarine_Hunter",
+        
+        "T3_Mobile_AA",
+        "drlk005", -- T3 maa for Megalith
```
norraxx, line 136: maa? O_o

lua/keymap/hotbuild.lua
shalkya: I initialize `old_selection` at the beginning of the file, so it can't be `nil`
If it was `nil`, i would get an error

lua/keymap/hotbuild.lua
shalkya: ah yes my bad, i'm gonna change that. Thanks.

lua/keymap/hotbuild.lua
shalkya: don't know, it always been like that. Never had issue with that.

lua/keymap/hotbuild.lua
shalkya: pretty much the same as there is a `buildActionTemplate `and a `buildActionBuilding`
I've extracted some function from there to be called by each. But merging 2 of them, would lead to a lot of if condition, and wouldn't be more easy to read imo

lua/keymap/keydescriptions.lua
shalkya: well i didn't find out the use for now. That was the only solution i found to fix the issue, of them not updating.

lua/keymap/unitkeygroups.lua
shalkya: what naming ? the unit name or the description ?

KionX:
Make the correct localization. New lines and changes to add in the loc/US/strings_db.lua

shalkya:
working state, please test.

shalkya:
bind alternate hotbuild in the f1 menu
how it works : select a factory, press the hotkey, cycle through the unit, press left click to build the unit, each time you click left it will build 1 more unit, if you pressed "shift" you will build unit 5 by 5. Press right to allow left click to act normally (doesn't build unit anymore).
For build teamplate, it's pretty much the same, although left click will work normally after clicking once to build the template (no need to right click).

shalkya:
@Uveso 
Since i created alternative hotbuild, there is now several hotkey used to build the same units. Those hotkey relationship (unit name <-> hotkey name) are stored in otherRelations. Then you transfer that relationship to helpIdRelations to get (unit code <-> hotkey name). If there is several hotkey for 1 units, they will overright themself here : helpIdRelations[id] = otherRelations[group].lower(otherRelations[group])

Edit : These issue was already existing with hotbuild special hotkey, that were rewritten, thus the label weren't appearing in game.

shalkya:
This PR should be done, unless you spot bugs.

About optimisation i am not sure if using `pairs` over a table is good like this : 
`for id, group in pairs(helpIdRelations) do`

or if it's better to use something along those lines : 
```function awesome_next(t, k)
    k, t = next(t, k)
    if not t then return end
    return k, table.unpack(t)
end```

`for id, group in awesome_next, helpIdRelations do`

KionX:
Return LOCs keyCategories in lua/keymap/keycategories.lua

KionX:
```Press right to allow left click to act normally (doesn't build unit anymore).```
This also changes the factory's waypoint.

Maybe this is pointless but properkeyname_0166 is missing in strings_db.lua

shalkya:
> Press right to allow left click to act normally (doesn't build unit anymore).
This also changes the factory's waypoint.

Yes, it kept the previous behaviour. I think it's not an issue, since you can use middle click to quit the mode without changing the factory waypoint.

shalkya:
alternative_hotbuild

PR #3077, opened 2020-04-15
Update defaultunits.lua

The-Balthazar:
It is entirely possible that a unit doesn't have self.UnitId defined, and it costs nothing to have a fallback.

This is specifically to fix a compatibility issues with BrewLAN walls, but it's entirely possible that other mod units will try to use those scripts and cause an error by not having self.UnitId defined. There's no reason to not fallback to self:GetUnitId when self.UnitId is nil.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1986,7 +1986,7 @@ ConstructionUnit = Class(MobileUnit) {
         self.UnitBeingBuilt = unitBeingBuilt
         self.UnitBuildOrder = order
         self.BuildingUnit = true
-        if (unitBeingBuilt.UnitId or unitBeingBuilt:GetUnitId) == self:GetBlueprint().General.UpgradesTo and order == 'Upgrade' then
+        if (unitBeingBuilt.UnitId or unitBeingBuilt:GetUnitId()) == self:GetBlueprint().General.UpgradesTo and order == 'Upgrade' then
             self.Upgrading = true
             self.BuildingUnit = false
         end
@@ -2384,7 +2384,7 @@ ACUUnit = Class(CommandUnit) {
 
     OnStopBeingBuilt = function(self, builder, layer)
         CommandUnit.OnStopBeingBuilt(self, builder, layer)
-        ArmyBrains[self.Army]:SetUnitStat(self.UnitId or self:GetUnitId, "lowest_health", self:GetHealth())
+        ArmyBrains[self.Army]:SetUnitStat(self.UnitId or self:GetUnitId(), "lowest_health", self:GetHealth())
         self.WeaponEnabled = {}
     end,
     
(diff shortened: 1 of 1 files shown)


lua/defaultunits.lua
```diff
@@ -1986,7 +1986,7 @@ ConstructionUnit = Class(MobileUnit) {
         self.UnitBeingBuilt = unitBeingBuilt
         self.UnitBuildOrder = order
         self.BuildingUnit = true
-        if unitBeingBuilt.UnitId == self:GetBlueprint().General.UpgradesTo and order == 'Upgrade' then
+        if (unitBeingBuilt.UnitId or unitBeingBuilt:GetUnitId) == self:GetBlueprint().General.UpgradesTo and order == 'Upgrade' then
```
shalkya, line 1989: luac: ./lua/defaultunits.lua:1989: function arguments expected near ')'
GetUnitId => GetUnitId()

lua/defaultunits.lua
```diff
@@ -2384,7 +2384,7 @@ ACUUnit = Class(CommandUnit) {
 
     OnStopBeingBuilt = function(self, builder, layer)
         CommandUnit.OnStopBeingBuilt(self, builder, layer)
-        ArmyBrains[self.Army]:SetUnitStat(self.UnitId, "lowest_health", self:GetHealth())
+        ArmyBrains[self.Army]:SetUnitStat(self.UnitId or self:GetUnitId, "lowest_health", self:GetHealth())
```
shalkya, line 2387: same

lua/defaultunits.lua
```diff
@@ -2401,8 +2401,8 @@ ACUUnit = Class(CommandUnit) {
             aiBrain:OnPlayCommanderUnderAttackVO()
         end
 
-        if self:GetHealth() < ArmyBrains[self.Army]:GetUnitStat(self.UnitId, "lowest_health") then
-            ArmyBrains[self.Army]:SetUnitStat(self.UnitId, "lowest_health", self:GetHealth())
+        if self:GetHealth() < ArmyBrains[self.Army]:GetUnitStat(self.UnitId or self:GetUnitId, "lowest_health") then
+            ArmyBrains[self.Army]:SetUnitStat(self.UnitId or self:GetUnitId, "lowest_health", self:GetHealth())
```
shalkya, line 2405: same

speed2CZ:
What units would that be? It's saved in OnCreate, so this might be the case only with destructive hooks. And that's a mod problem and not FAF problem.

The-Balthazar:
> What units would that be? It's saved in OnCreate, so this might be the case only with destructive hooks. And that's a mod problem and not FAF problem.

It's possible to call the function before that value is defined even with nondestructive hooks.

speed2CZ:
Can you show me some examples?

The-Balthazar:
https://github.com/The-Balthazar/BrewLAN/blob/master/mods/BrewLAN/lua/walls.lua#L46

speed2CZ:
Well it shouldn't be in OnCreate

speed2CZ:
And if you dont want to move it, you can actually save the value in your hooked function, so it's a mod thing and doesnt really require faf code to change, so Im closing it.

The-Balthazar:
It adds a timing dependancy to the function which doesn't need to be there, and costs nothing to remove by just having this fallback.

PR #3083, opened 2020-04-23
AI Update V8

Uveso:
(Changelog updated 18.May)

Finally this patch will fix the game slowdown on long AI games.
Reason for the slowdown was the engineermanger while ping pong forking functions.
This was causing in the end a engine slowdown a finally a stack buffer overrun.

This patch also stores almost every forked thread to the corresponding unit/weapon,
so it can be destroyed on unit dead to prevent parentless threads.

The transport function is now using surfaceghight for the transportation target
for faster unit drops.

Also fixed an issue where the ACU is not moving on gamestart.
This was caused by the `BlockCommandQueue` from the ACU WarpInAnimation

Changelog:

------------------------------------------------------------------------------------

aiarchetype-managerloader.lua

function ExecutePlan()
-Opt: Storing Forked Thread inside the aiBrain Trashbag for better performance

------------------------------------------------------------------------------------

AIBehaviors.lua

function CommanderThread()
-Opt: Added check for 'BlockCommandQueue' to prevent stucked ACU after WarpInAnimation

function CommanderThreadImproved()
-Opt: Added check for 'BlockCommandQueue' to prevent stucked ACU after WarpInAnimation

function GetClosestShieldProtectingTarget()
-Opt: Added safeguard for dead attacker or target

function CommanderThreadSorian()
-Opt: Added check for 'BlockCommandQueue' to prevent stucked ACU after WarpInAnimation

------------------------------------------------------------------------------------

SorianStrategyBuilders.lua

BuilderGroup 'Sorian Excess Mass Strategy'
-Opt: Storing Forked Thread inside the aiBrain Trashbag for better performance

------------------------------------------------------------------------------------

AttackManager.lua

function FormAttackPlatoon()
-Opt: Storing Forked Thread inside the aiBrain Trashbag for better performance

------------------------------------------------------------------------------------

aiutilities.lua

function UseTransports()
-Opt: Changed command WaitTicks() to coroutine.yield() for better performance
-Opt: Added surfacehight to trasnport location for faster unit unloading

function EngineerTryReclaimCaptureArea()
-Fix: Function now only returns true if we have valid unit for capture

function UseTransports
-Fix: Removed unused local variable
-Fix: Added IssueClearCommands for transports to prevent flightloops

------------------------------------------------------------------------------------

BaseManagerPlatoonThreads.lua

function ConditionalBuildSuccessful()
-Opt: Storing Forked Thread inside the aiBrain Trashbag for better performance

function AMUnlockBuildTimer()
-Opt: Storing Forked Thread inside the Platoon Trashbag for better performance

function AMUnlockRatioTimer()
-Opt: Storing Forked Thread inside the Platoon Trashbag for better performance

------------------------------------------------------------------------------------

sorianutilities.lua

function Nuke()
-Opt: Storing Forked Thread inside the aiBrain Trashbag for better performance
-Opt: Removed debug log lines

------------------------------------------------------------------------------------

aibrain.lua

function AddBuilderManagers()
-Opt: Added a baseType to every expansion for faster expansion count

function GetManagerCount()
-Opt: To count expansions we are now searching for a baseType not 8 x name search

function InitialAIThread
-Fix: Delaying the AI so it can't reclaim the start area before it's cleared from the ACU landing blast.

------------------------------------------------------------------------------------

cybranunits.lua

Class CConstructionEggUnit()
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance

------------------------------------------------------------------------------------

aiattackutilities.lua

function GetBestThreatTarget()
-Fix: changed missplaced variables in GetThreatAtPosition call

------------------------------------------------------------------------------------
defaultunits.lua

function OnPrepareArmToBuild() (ConstructionUnit Class)
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance

function OnPrepareArmToBuild() (CommandUnit Class)
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance

------------------------------------------------------------------------------------

platoon.lua

several functions()
-Opt: Changed command WaitTicks() to coroutine.yield() for better performance

function AssistBody()
-Opt: Moved safeguard for nonexisting platoons at the start of the function

function EngineerBuildAI()
-Opt: Removed old sorianfunction for delaying platoonbuilder
-Fix: Excluded StationAssitPods from building tasks
-Opt: Using now .Location instead of the functioncall :GetLocationCoords()
-Fix: The safeguard at the end of the function is now working

function WatchForNotBuilding()
-Opt: Removed unused unitstate checks

function ProcessBuildCommand()
-Opt: replaced BuildToNormalLocation and NormalToBuildLocation and used numbers directly
-Fix: Units now get capture/reclaim and buildcommands queued instead of looping the ProcessBuildCommand
-Fix: Prevent disbanding a platoon while on a transport

function EngineerBuildAISorian
-Opt: Removed old sorianfunction for delaying platoonbuilder
-Opt: Using now .Location instead of the functioncall :GetLocationCoords()

------------------------------------------------------------------------------------

CollisionBeam.lua

function DoDamage()
-Opt: Storing Forked Thread inside the Beam Trashbag for better performance

------------------------------------------------------------------------------------

ScenarioUtilities.lua

function CommanderWarpDelay()
-Opt: Removed BlockCommandQueue for AI brains

------------------------------------------------------------------------------------

EngineerManager.lua

several functions()
-Opt: Changed command WaitTicks() to coroutine.yield() for better performance

function AddUnit()
-Fix: Excluded StationAssitPods from building tasks
-Fix: Changed the logic of forking threads to prevent gameslowdown at +1 hour

function GetNumCategoryBeingBuilt()
-Opt: returning a value directly without defining it local

function TaskFinished()
-Fix: Changed the logic of forking threads to prevent gameslowdown at +1 hour

Removed functions:
EngineerWaiting, InitialWait, DelayAssign, DelayAssignBody
-Fix: Changed the logic of forking threads to prevent gameslowdown at +1 hour

Replaced removed functiosn with new functions:
ForkEngineerTask, DelayAssign, Wait, EngineerWaiting
-Fix: Changed the logic of forking threads to prevent gameslowdown at +1 hour

------------------------------------------------------------------------------------

NukeDamage.lua

function NukeAOE()
-Opt: Storing Forked Thread inside the aiBrain Trashbag for better performance

------------------------------------------------------------------------------------

Projectile.lua

function DoDamage() (Projectile Class)
-Opt: Storing Forked Thread inside the Projectile Trashbag for better performance

------------------------------------------------------------------------------------

Unit.lua

function DeathThread()
-Opt: Destroy trashbags only if not already done

function DestroyAllTrashBags()
-New: This function clears all trashbags and unit effects

function OnDestroy()
-New: Destroy trashbags only if not already done inside DeathThread()

function StopRocking()
-Opt: Only kill a thread when it actual exist.

------------------------------------------------------------------------------------

Blueprints.lua

function HandleUnitWithBuildPresets()
-New: Adding SACU presets as unit category for AI builders

------------------------------------------------------------------------------------

CIFArtilleryProton01_script.lua

function OnImpact()
-Opt: Storing Forked Thread inside the Projectile Trashbag for better performance

------------------------------------------------------------------------------------

DSLK004_script.lua

function DoDamage()
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance

------------------------------------------------------------------------------------

UEL0001_script.lua

function CreateEnhancement()
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance

------------------------------------------------------------------------------------

UEL0301_script.lua

function CreateEnhancement()
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance

------------------------------------------------------------------------------------

XRB2308_script.lua

function OnStopBeingBuilt()
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance

function StartSinking()
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance

------------------------------------------------------------------------------------

XRB2309_script.lua

function StartSinking()
-Opt: Storing Forked Thread inside the Unit Trashbag for better performance


KionX:
Looks like a bug.

Before:
```ForkThread(self.WaitForBuildAnimation, self, true)```

After:
```self:ForkThread(self.WaitForBuildAnimation, self, true)```
The second parameter 'self' is unnecessary.

In self:ForkThread written:
```local thread = ForkThread(fn, self, unpack(arg))```
```self.Trash:Add(thread)```

```aiBrain:ForkThread(EconWatch, aiBrain)``` and similar ones, too.

**Comment by Uveso** [comment]
Use this call:
`ForkThread(EconWatch, aiBrain)`
will give these args to the called function:
EconWatch(`aiBrain`)

If i now call `aiBrain:ForkThread(EconWatch, aiBrain)`
it will produce:
EconWatch(aiBrain, `aiBrain`)
The first aiBrain comes from the call with `aiBrain:` and the second from the argument `, aiBrain)`

So you can use both call methods without breaking something.
Of course this will be removed after i checked compatibility to all functions
to remove the overload.

And thanks for finding the `self:ForkThread(self.WaitForBuildAnimation, self, true)`
Thats obviously wrong. I will fix that and recheck all other calls.

**Comment by Uveso** [comment]
@KionX 
Thanks for your time!

I was confused a bit at the start but i guess i found all issues.
I also removed every call where i was unsure how to overload `self` without
changing the targetfunction arguments.

File: lua/AI/aiutilities.lua [hunk 1]
```diff
@@ -1524,7 +1524,7 @@ function UseTransports(units, transports, location, transportPlatoon)
 
     local attached = true
     repeat
-        WaitSeconds(2)
+        coroutine.yield(20)
```
KionX, line 1527: I think WaitTicks will be faster coroutine.yield

lua/AI/aiutilities.lua
```diff
@@ -1572,11 +1572,20 @@ function UseTransports(units, transports, location, transportPlatoon)
...
         -- If no location then we have loaded transports then return true
         if location then
+            -- Adding Surface Height, so the transporter get not confused, because the target is under the map (reduces unload time)
+            location = {location[1], GetSurfaceHeight(location[1],location[3]), location[3]}
             local safePath = AIAttackUtils.PlatoonGenerateSafePathTo(aiBrain, 'Air', transports[1]:GetPosition(), location, 200)
             if safePath then
+                local LastPos
                 for _, p in safePath do
                     IssueMove(transports, p)
+                    LastPos = p
```
KionX, line 1582: LastPos not used

lua/platoon.lua
```diff
@@ -3555,34 +3534,18 @@ Platoon = Class(moho.platoon_methods) {
...
-            --  LOG('*AI DEBUG: Commander waiting for building: return home.')
-            --end
-            --if eng.UnitBeingBuiltBehavior then
-            --  LOG('*AI DEBUG: Commander waiting for building: unit being built.')
-            --end
+        while not eng.Dead
+            and eng.GoingHome
+            or eng.UnitBeingBuiltBehavior
+            or eng.ProcessBuild != nil
+            or not eng:IsIdleState()
```
KionX, line 3544: To avoid confusion it is better to place the brackets

lua/platoon.lua
```diff
@@ -3631,50 +3586,42 @@ Platoon = Class(moho.platoon_methods) {
...
-                end
-
-                local engpos = eng:GetPosition()
-                while not eng.Dead and eng:IsUnitState("Moving") and VDist2(engpos[1], engpos[3], buildLocation[1], buildLocation[3]) > 15 do
-                    WaitSeconds(2)
-                end
-
                 -- check to see if we need to reclaim or capture...
-                if not AIUtils.EngineerTryReclaimCaptureArea(aiBrain, eng, buildLocation) then
+                AIUtils.EngineerTryReclaimCaptureArea(aiBrain, eng, buildLocation)
```
KionX, line 3618: It seems that logic is broken here. Also bad for performance.

KionX:
self:ForkThread(callback, tempPlatoon) can be write it like this
local thread = ForkThread(callback, tempPlatoon)
self.Trash:Add(thread)

KionX:
lua/defaultunits.lua 2035
self:ForkThread(self.WaitForBuildAnimation, self, true) can be write as
self:ForkThread(self.WaitForBuildAnimation, true)

lua/platoon.lua
```diff
@@ -3631,50 +3586,42 @@ Platoon = Class(moho.platoon_methods) {
...
                     -- check to see if we can repair
-                    if not AIUtils.EngineerTryRepair(aiBrain, eng, whatToBuild, buildLocation) then
+                AIUtils.EngineerTryRepair(aiBrain, eng, whatToBuild, buildLocation)
                         -- otherwise, go ahead and build the next structure there
-                        aiBrain:BuildStructure(eng, whatToBuild, NormalToBuildLocation(buildLocation), buildRelative)
-                        if not eng.NotBuildingThread then
-                            eng.NotBuildingThread = eng:ForkThread(eng.PlatoonHandle.WatchForNotBuilding)
-                        end
-                    end
+                aiBrain:BuildStructure(eng, whatToBuild, {buildLocation[1], buildLocation[3], 0}, buildRelative)
```
KionX, line 3622: It seems that logic is broken here. Also bad for performance.

Uveso: Yes the logic is changed.

Before the change we first checked reclaimed CaptureArea, then left the ProcessBuildCommand and started the forked WatchForNotBuilding thread.
Then we called the ProcessBuildCommand again and checked for EngineerTryRepair.
Then again, we started the WatchForNotBuilding thread, called the ProcessBuildCommand the 3rd time to finaly issue the buildcommand.

Now we only call ProcessBuildCommand once, check for reclaim and capture and set the buildcommand. This change also fixed an issue where multiple engineers try to cap the same mass spot when its occupied with rocks/reclaim

lua/AI/aiutilities.lua
Uveso: Sprouto tested this and is using it inside his LOUD AI.
Both commands will be executed in the same time, WaitSeconds is only an alias for coroutine.yield.
But the LUA interpreter will find the command coroutine.yield earlier in the command list and this makes it a tiny bit faster.

lua/AI/aiutilities.lua
Uveso: Ah, yes a debug leftover.

lua/platoon.lua
Uveso: Oh yes, this structure was left over from debugging, i will change it  back.

Uveso:
Nice review, thanks a lot @KionX

KionX:
lua/AI/OpAI/BaseManagerPlatoonThreads.lua
lua/sim/CollisionBeam.lua
lua/sim/Projectile.lua
 units/dslk004/DSLK004_script.lua
ForkThread no longer needs to be placed in TrashBag?

KionX:
I don't know for sure but I feel that the logic in EngineerTryReclaimCaptureArea and EngineerTryRepair is broken.

lua/platoon.lua
```diff
@@ -3555,34 +3534,13 @@ Platoon = Class(moho.platoon_methods) {
...
-            --  LOG('*AI DEBUG: Commander waiting for building.')
-            --  eng:PrintCommandQueue()
-            --end
-            --if eng.GoingHome then
-            --  LOG('*AI DEBUG: Commander waiting for building: return home.')
-            --end
-            --if eng.UnitBeingBuiltBehavior then
-            --  LOG('*AI DEBUG: Commander waiting for building: unit being built.')
-            --end
+        while not eng.Dead and eng.GoingHome or eng.UnitBeingBuiltBehavior or eng.ProcessBuild != nil or not eng:IsIdleState() do
```
KionX, line 3540: So the loop can work even if the unit is dead.
I guess that's right
while not eng.Dead and (eng.GoingHome or eng.UnitBeingBuiltBehavior or eng.ProcessBuild != nil or not eng:IsIdleState()) do

Uveso: You are right, the loop should not run if the engineer is dead.
`if not true and true or true or true then  = true` (wrong)
`if not true and (true or true or true) then  = false` (right)

Uveso:
@KionX 
`ForkThread no longer needs to be placed in TrashBag?`
They are still need to be there. I will fix this in the next update since your idea to move them
manually to the trashbag allowes me to add some more threads to the trashbag.

`Air cannot be captured or reclaimed.`
fixed

`The target may be outside the construction radius.`
range is 10, constuction radius is 15. This will work.

`Can't repair allies or unnecessary condition.`
Right, this function is only to clear the buildplace from enemy units or wrecks.

`It is enough to check > 0.`
That's right !

`The function returns false instead of true.`
It only returns true if reclaim are found.

`Does Issue put orders in a queue or should the function be returned immediately?`
The issue command adds the command to the unit command queue directly,
so we can do all reclaim commands in a single loop.

`Tries to repair it regardless of the result of the previous function.`
Right, the previous function only clears the buildplace from enemy units and wrecks.

`Tries to build regardless of the result of the previous function?`
Yes, all commands will be added to the engineer command queue,
so it will first reclaim, then repair, then build.
All in a singe command queue.

KionX:
```range is 10, constuction radius is 15. This will work.```
In engineers blueprints MaxBuildDistance = 5
I meant that you should probably use the unit build radius.

Uveso:
Ah i see what you mean, the max distance of the engineer construction beam.
Yes, the engineer will maybe move while reclaiming enemy units on the buildplace.

We need the reclaim range of 10 because some buildingtemplates (multiple buildings)
need a constuction radius of max 15 map units.
(10 is already a reduced radius, but it will work for most templates)

KionX:
There is a problem that if the target is outside the radius of the construction, the engineer move closely to target. It looks like this problem comes from this function. This is inconvenient for players.

Uveso:
This is an AI only function and does not change the behavior of engineers for human players.
All changes in this patch are for the GPG & Sorian AI only.

Also the change for queueing first capture then reclaim and then buildcommand
is only for the GPG and Sorian AI.
Maybe thats why we get confused earlier^^

Human players are only affected by the Thread and Trashbag changes in this AI Update.

shalkya:
you need to do "git rebase deploy/fafdevelop" then you will get 13 conflicts, i let you solve them since you know better what to keep or not (it's related to AI mostly from what i saw)

Uveso:
i only had 1 conflict last time i rebased it.
Well then wreck this PR.

PR #3091, opened 2020-05-05
Speedup of score sync

Crotalus:
**Experiment: DO NOT MERGE, untested**

* Do not calculate score / history for civilian
* Do not calculate stats for dead players
* Only store delta changes in history table
* Do not sync everything in ArmyScore (unit bp stats etc), only store this in history
* Reduce number of tables by flattening structure
* Reduce number of score threads running

@KionX Take a look at this and see if this speeds up game / replays

### DIFF
--- a/lua/sim/score.lua
+++ b/lua/sim/score.lua
@@ -3,9 +3,13 @@ scoreInterval = 1
 resourcesInterval = 0.1
 alliesScore = true
 
-local scoreData = {current={}, historical={}}
-local scoreOption = ScenarioInfo.Options.Score or "no"
-local ArmyScore = {}
+local Victory = import('/lua/victory.lua')
+local ScoreHistory = {}
+local BpStatsToCollect = {}
+local ScoreEnabled = (ScenarioInfo.Options.Score or 'no') == 'no'
+local SentOnVictory = false
+local IsObserver = false
+
 
 -- Some of these values pre-existed and are used in other places, that's why their naming is not consistent
 local categoriesToCollect = {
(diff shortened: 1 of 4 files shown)


KionX:
Tested replay with a current of scoreboard then disable the collection of statistics completely.
There is no speed difference. This PR was not considered and looks over complicated.

Crotalus:
Did you test disable all score/history threads in score.lua so it doesn't send anything via Sync to rule out if that is the cause of the recent stuttering?

I've seen immense slowdown when too much data is sent over the Sync-table sim <-> UI. It won't affect the sim-speed speed-meter though but instead cause stutter, especially at higher sim speeds.
It got worse after the stuff with achievements was added some years ago. 

The scoreData.historical in current FAF version grows to enormous sizes in long games due to all statistics that's store there. Try put on a 8+ player AI game with a periodical SPEW(repr(scoreData.historical)) and you'll see what I mean. 

The bottom white dots/line shown in `ren_showframetimes` shows how much time is spent syncing data.

KionX:
It was about slowing down the replay due to collecting statistics.
I used 16 player AI game. There were no stutters.
Stuttering I had only on SIM speed 15+ even without AI.

KionX:
I think I've found the root of the evil:
https://github.com/FAForever/fa/blob/develop/lua/sim/score.lua#L253
It is performed for each sync, although this is not necessary at all.

KionX:
Tested this PR.
It solves the performance problem but it is overcomplicated.
Also found errors:

KionX:
Propose to close this PR. It is not functional. Must be completely redone.
I am optimizing statistics collection. How urgent is this?

shalkya:
removing this line : https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/score.lua#L253
is enough to fix the issue, so i guess we can start by that. But if you have better performance improvement, it's always welcomed.

Crotalus:
Yes, I'll close this for now. I was rather sure it would break something on the UI side anyway. 

Maybe could do a new PR with only the dead + civilian check to reduce the calculation somewhat in games.

Issue #3096, opened 2020-05-07
[BUG] Ennemy ACU randomly revealed to me at some point


Auriko:
Replay is https://content.faforever.com/faf/vault/replay_vault/replay.php?id=11676244
It happens around 30-32 min, at the end of the game. 

His commander was stealthed and underwater. Previous attempts to find him with spy planes failed, as expected. I had built a Rianne Eye which should be useless, and which i didn't use properly on his location anyway ... 

So around min 32 , the icon of his commander was revealed on my mini map and i was able to snipe him. He walked near the edge of the map but i don't see how that could be relevant. 

So I think it's just a weird bug. Quite impactfull in this particular game ...

Auriko:
Watching it back, i think it's a Rhianne Eye bug, I did order the thing to reveal the position few minutes earlier, but then ran off of power. I have the [e] please UI mod, so the Eye was turned off, and after few seconds i did turn it off manually. 
So when he walked to the previously-revealed area, that's when he appeared on my map. But the Eye was off, and no radar ... so even if the Eye spotted him, he shouldn't have stayed visible on my map ... it shouldn't have happened.

shalkya:
so i looked into that : 
Eye of Rhianne doesn't remove its detection ability on the seabed/underwater level, after being turn off. So it will keep showing amphibious and sub units on the last spot it was used.

And as a tip for you, T2/T3 aeon torpedo bomber use depth charges that provide water vision, so if you know there is a stealth ACU to kill in the area, you can ground fire to reveal the ACU and target him afterward.

Auriko:
Much interesting, thx !

PR #3098, opened 2020-05-09
Quantum optics facility fix

KionX:
Fixed a bug that allowed the building to detect underwater units.
This issue has also been fixed: #3096

### DIFF
--- a/lua/RemoteViewing.lua
+++ b/lua/RemoteViewing.lua
@@ -87,7 +87,7 @@ function RemoteViewing(SuperClass)
                         Omni = false,
                         Radar = false,
                         Vision = true,
-                        WaterVision = false,
+                        WaterVision = true,
                         Army = self:GetAIBrain():GetArmyIndex(),
                     }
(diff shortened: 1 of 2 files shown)


lua/RemoteViewing.lua
```diff
@@ -84,6 +87,7 @@ function RemoteViewing(SuperClass)
                         Omni = false,
                         Radar = false,
                         Vision = true,
+                        WaterVision = false,
```
shalkya, line 90: could you switch it to true ? So it keep the current balance.

lua/RemoteViewing.lua
```diff
@@ -87,7 +87,7 @@ function RemoteViewing(SuperClass)
                         Omni = false,
                         Radar = false,
                         Vision = true,
-                        WaterVision = false,
+                        WaterVision = true,
                         Army = self:GetAIBrain():GetArmyIndex(),
```
speed2CZ, line 91: I think this could be self.Army

PR #3099, opened 2020-05-09
Fix tarmac spawning

BlackYps:
This one fixes a minor visual bug.
Destroying a structure spawns a new tarmac with a limited lifetime. When this new tarmac is created when the unit is destroyed it will spawn over the explosion decal. Calling it together with OnKilled places it under the explosion decal, so it looks like the explosion decal appeared over the tarmac (like it should be).

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -412,16 +412,6 @@ StructureUnit = Class(Unit) {
 
     end,
 
-    -- Adding into OnDestroy the ability to destroy the tarmac but put a new one down that looks exactly like it but
-    -- will time out over the time spec'd or 300 seconds.
-    OnDestroy = function(self)
-        Unit.OnDestroy(self)
-        local orient = self.TarmacBag.Orientation
-        local currentBP = self.TarmacBag.CurrentBP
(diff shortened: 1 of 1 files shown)


shalkya:
hello i'm trying to figure out what this achieve.
Only difference i noticed, is when i spawn a unit (they don't have any mark on the ground). Without the fix, the mark on the ground appears at the end of the explosion sequence. With the fix, the mark on the ground appears at the start of the explosion sequence.
Is it what it is supposed to fix ?

BlackYps:
No that is just a side effect. It is primarily for structures. On small structures the wreck maybe covers the explosions decal, but you can reclaim the wreck to see it. With factories and T2 Pgens you can probably see it under the wreck already. It is more noticable with RK's explosions, because he uses bigger decals.
Just try it with a factory. If you still can't see it, hit me up again and I will make some screenshots to show what I mean.

shalkya:
yeah i meant structure, my bad.
Well i tested a bit can't see anything because it's always hidden by the wreck.
So that's an improvement for RK explosion mostly right ?

PR #3108, opened 2020-05-16
Lobby enhancements + refactoring

KionX:
This functionality is now available in the singleplayer.
Unlocked in SP

Disabled voices in lobby without FAF client.

Now when you press the "Fill Slots" button, the "Auto Teams" used.


shalkya:
you can't close/occupy an open slot
you can't Kick player (remove an AI) or Move an AI to another slot

Issue #2991, opened 2020-02-14
pls dear god save us from dual gap replays


PhilipJFryFAF:
someone pls revert the teamkill report feature
it's drowning the moderation team in cancer reports - half of which are from games against the AI/survival games or other meme games that aren't worth watching.
The other half are 50+ mins dual gap/astro crater games that are making my eyes bleed.
If someone wants to report another user for teamkilling they can do so manually.
thx in advance @speed2CZ

speed2CZ:
Last time I was promised 1TB SSD, what do you offer now?

MrRowey:
@Garanas i would say this could be closed if this is more to do with the server side

PR #3113, opened 2020-05-23
Prevent TeamKill auto report

KionX:
Prevent 'TeamkillHappened' https://github.com/FAForever/fa/issues/2991

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2452,18 +2452,8 @@ ACUUnit = Class(CommandUnit) {
             Sync.EnforceRating = true
             WARN('ACU kill detected. Rating for ranked games is now enforced.')
 
-            -- If we are teamkilled, filter out death explostions of allied units that were not coused by player's self destruct order
-            -- Damage types:
-            --     'DeathExplosion' - when normal unit is killed
-            --     'Nuke' - when Paragon is killed
-            --     'Deathnuke' - when ACU is killed
-            if IsAlly(self.Army, instigator.Army) and not ((type == 'DeathExplosion' or type == 'Nuke' or type == 'Deathnuke') and not instigator.SelfDestructed) then
-                WARN('Teamkill detected')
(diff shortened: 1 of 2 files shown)


shalkya:
image
i still get the message

KionX:
This is not an auto report. Previously, he sent without the participation of the player.

shalkya:
weird that we have an auto-report on top of the report.
@PhilipJFryFAF what do you want removed ? auto-report, the pop-up i screen, or both ?

lua/defaultunits.lua
```diff
@@ -2452,18 +2452,8 @@ ACUUnit = Class(CommandUnit) {
             Sync.EnforceRating = true
             WARN('ACU kill detected. Rating for ranked games is now enforced.')
 
-            -- If we are teamkilled, filter out death explostions of allied units that were not coused by player's self destruct order
-            -- Damage types:
-            --     'DeathExplosion' - when normal unit is killed
-            --     'Nuke' - when Paragon is killed
-            --     'Deathnuke' - when ACU is killed
-            if IsAlly(self.Army, instigator.Army) and not ((type == 'DeathExplosion' or type == 'Nuke' or type == 'Deathnuke') and not instigator.SelfDestructed) then
```
shalkya, line 2460: can you reintroduce those exception ? it's annoying to get the teamkill warning when these situation happens. Especially that you are making the report be useless (because mod can't deal with them).

KionX: These conditions blocked the teamkill by nuke.
I don't see a problem.

shalkya: because it's often that this kind of teamkill happen, and most of the time they aren't on purpose.

shalkya: from what gieb told me, auto-report is only a way to not send report to garbage. When removing auto-report, and people report manually, that report won't be treated. So basically we could even remove the whole report feature.

KionX: The server processes these messages independently. I'll leave this PR it is.

shalkya:
i think that should be closed, when you look at the comment of askaholic.
image

PR #3122, opened 2020-06-04
Added Aeon & Seraphim engineering stations

The-Balthazar:
FAO JaggedAppliance

A set of three sequentially upgrading Seraphim stations, and a single 
Aeon station.


MrRowey:
I would be 100% Against this from being implemeneted in to the basse game of faf these type of buildings should only been in mods and not the base version of faf this defets the purpose of faction diversity

Azraeel:
> I would be 100% Against this from being implemeneted in to the base game of faf these type of buildings should only been in mods and not the base version of faf this defets the purpose of faction diversity

Then remove cybran and uef engineering stations if not then they all should get engineering stations

JaggedAppliance:
The stations don't create and build effect beams when constructiong things and the upgrading of the seraphim station creates new stations on top of itself instead of actually upgrading.

image

FemtoZetta:
Can we get some pics of how they look when they are not stacked? Because Jagged's screenshots makes it look like from some low effort mod.
Also, I don't think Sera or Aeon need this at all...

And if I understood it correctly Aeon has a single station with no upgrades:
grafik
And it is way worse than UEF (which already has a pretty bad efficiency but at least it can go anywhere on the map)? 740 mass for only 25bp. What else can this station do that makes it so expensive? (being aquatic doesn't qualify imo)

shalkya:
image
image

speed2CZ:
Im not a fan of adding this

Garanas:
I'm not a fan of this either, to put it mildly.

What is the problem that this is trying to solve?

And what is next? Will all factions have some special t3 intel structure in the future too, like the Cybran and Aeon factions have? Or a t3 torpedo bomber? This is what faction diversity is for and bringing these in will open the gates to other changes that further diminish faction diversity.

FemtoZetta:
Aeon one looks OK imo, like an engineer merged into a tower lol.
The Sera one is really ugly and should be redone, it doesn't even fit the typical Sera style.

And like the others said, I think this hurts faction diversity and it's not even needed, you can just spam engineers, and that's actually more mass efficient... Maybe if the new stations were mechanically more interesting, and not just a copy of Hives it could work, but not like this.

EldoreiJK:
Maybe eng stations are not that of a bad idea as to fix engine issues with hundreds of units in production area. But although I'm a big sera fan I don't like that change either.
@FemtoZetta I've got an interesting idea for a good eng station mechanic, but I'm definitely not gonna script it. A good eng station would be a buff source, as if assisting an engineer, but actually "doubling" that's engineer power. Now that I think of it maybe it wouldn't be such a big pain to script it anyway, just set the engineering power to copy is target power if the target is an eng. If sacu/acu/building, take a smaller portition of it's build power. To make it most powerful assisting T3 Eng. If you ask for the interesting part, a good mechanic would be that such a station wouldn't be a bad choice itself, but would be much more proficient to actually assist an eng. You loose the eng, you loose some of station power until you replace eng and reset assist.

apriori:
I don't really consider it a valid argument about "diversity" when two factions are not able to scale their build power in any decent way in late game, while two others completely stomp with this. 'Diversity' is about equivalent differences  - and sera has no equivalent filling the gap of lack of build power growth. Nor does Aeon.

aeoncleanse:
Errrrrm, I _strongly_ recommend this patch not happen. It's a very lazy way to address the buildpower difference and stomps all over faction diversity. This... This is not a good place to break the "No new units unless extremely needed" guideline.

zhuclam:
Why don't give aeon and seraphim t3 engies a toggle with the ability to fly? That way we have faction diversity, no new units, and the bests engies for them with the trade-off that they are vulnerable to anti-air.

speed2CZ:
Cause it would look very dumb

apriori:
> Errrrrm, I _strongly_ recommend this patch not happen. It's a very lazy way to address the buildpower difference and stomps all over faction diversity. This... This is not a good place to break the "No new units unless extremely needed" guideline.

I don't see laziness in a solution that takes into account technical limitations of the game. There is no way to fix BP issues, by e.g. increasing BP of ground engineers, given how shit the games path finding is. It should be obvious to anyone.

PR #3125, opened 2020-06-23
Replaced product menu with dynamic mod filters.

The-Balthazar:
This replaces the Product filter group on the spawn menu with a Source group that has a default first option of 'Core Files' (any unit with a bp.Source value starting with /units/), and auto generates a button for each active mod with units included. Comes with a script to shorten mod names to fit in a way that should keep them distinguishable, and also with some additions for button wraping when more than 6 options are in a single filter.

This also includes commented out code if people actually use the Product tab and would rather have that and have this new group hide itself if no mods are active.


Uveso:
F2Balta
Window has only 5 columns for buttons.
Windows size does not change if Source has more then one line.

The-Balthazar:
I never noticed the FAF menu was narrower than the original. I'll update for these.

The-Balthazar:
Fixed those issues, and another if you had exactly n*columns source buttons: https://github.com/FAForever/fa/pull/3125/commits/f3133cfa2ec30fa1d35b63d3b6e7c3fd50b8579e

lua/ui/dialogs/createunit.lua
```diff
@@ -25,10 +25,318 @@ local defaultEditField = false
...
+                end
+            end
+            if givetab then
+                local key = string.gsub(string.lower(mod.name),"%s+", "_")
+                local titleFit = function(name)
+                    local l = 12
+                    if string.len(name) <= l then return name end --If it's short, just gief
+
+                    name = string.gsub(name, "%([^()]*%)", "") --Remove any brackets
+                    name = string.gsub(name, "[ \s]+$", "") --Remove trailing spaces, because I can't be arsed to work out how to do both in one regex
```
shalkya, line 56: can you make a regex that doesn't trigger travis ? I didn't look in detail but something along the lines of %s or maybe //s should do the trick

shalkya:
You have grouped everything that isn't in a mod in the category "core game". This mean we can't look for only "SC" or "SC-FA" unit. This a regression.
Can you also solve the conflict ? (they appeared because i had to target this PR against deploy/fafdevelop instead of develop)
Please keep the scaling changes during conflict handling.

The-Balthazar:
Added specific core filters, and changed the regex so things don't complain: https://github.com/FAForever/fa/pull/3125/commits/95082b128bd52f4b1002533e7afb4614b4469c3d

PR #3131, opened 2020-07-04
add topological sorting for mods

AugSphere:
The existing implementation of mod sorting seems to have several issues:

* `ModComp` comparison does not check the before and after lists of the
second argument

* sorting relies on `table.keys` from `lua/system/utils.lua` and through
it on builtin `table.sort`, which assumes that an order between any pair
of elements can be established directly, which is not the case of mods
(for example if `mod2` depends on `mod1` and `mod3` depends on `mod2`,
`ModComp` looking at before/after lists of `mod1` and `mod3` will not be
able to establish a relationship between them correctly)

* there appears to be no functionality to log a warning if mods declare
an inconsistent ordering, contrary to the specification on top of the
`lua/MODS.LUA` file

This change introduces a new function `ModTopSort`, which implements a
simple topological sorting algorithm by a depth-first traversal of
dependencies. A warning is logged in case a cycle is encountered. The
old `ModComp` function is no longer used anywhere and is removed.
`GetActiveModsFiltered`, which previously relied on `ModComp`, retains
the same API, returning mods in an arbitrary order when before/after
lists are inconsistent, but should sort things more reliably otherwise.

### DIFF
--- a/lua/MODS.LUA
+++ b/lua/MODS.LUA
@@ -321,27 +321,66 @@ function AllSelectableMods()
     return r
 end
 
--- sorts mods according to their before/after lists, otherwise alphabetically
--- returns true when first less than second
-local function ModComp(first, second)
+-- topological sort the mods using before/after lists
+-- log a warning and return nil in case of a cycle
+local function ModTopSort(mods_to_sort)
+    local required_for = {}
+    for uid,mod in mods_to_sort do
+        required_for[uid] = {}
+    end
 
-    if _mod_cache[first].before then
-        for i,uid in _mod_cache[first].before do
-            if _mod_cache[second].uid == uid then
-                return true
+    for uid,mod in mods_to_sort do
+        if _mod_cache[uid].before then
+            for i,before_uid in _mod_cache[uid].before do
+                if mods_to_sort[before_uid] ~= nil then
+                    table.insert(required_for[uid], before_uid)
+                end
+            end
+        end
+
+        if _mod_cache[uid].after then
+            for i,after_uid in _mod_cache[uid].after do
+                if mods_to_sort[after_uid] ~= nil then
+                    table.insert(required_for[after_uid], uid)
+                end
             end
         end
     end
 
-    if _mod_cache[first].after then
-        for i,uid in _mod_cache[first].after do
-            if _mod_cache[second].uid == uid then
-                return false
-            end
+    local result = {}
+    local processing = {}
+    local done_with = {}
+
+    local function visit(uid)
+        if done_with[uid] then
+            return 0
+        end
+        if processing[uid] then
+            return 1
+        end
+
+        processing[uid] = true
+        for i,child_uid in required_for[uid] do
+            local r = visit(child_uid)
+            if r == 1 then return 1 end
+        end
+
+        processing[uid] = false
+        done_with[uid] = true
+        table.insert(result, mods_to_sort[uid])
+        return 0
+    end
+
+    local r = 0
+    for uid,mod in mods_to_sort do
+        r = visit(uid)
+        if r == 1 then
+            WARN('Inconsistent mod order lists, load order will be arbitrary')
+            return
         end
     end
 
-    return first < second
+    return table.reverse(result)
 end
 
 -- Get a list of ingame UI mods that should be active, based on preferences.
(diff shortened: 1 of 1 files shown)


AugSphere:
I'd really appreciate it if someone gave it a thorough review, since I don't have the time to set up proper automated tests to make sure it correctly (and an acceptable time) sorts various configurations of mods. Manual tests I did are likely inadequate.

shalkya:
I'm not sure what this commit aim to change in the game.
```
 Mod ordering

   It often matters in what order mods take effect. For example, perhaps two mods
   both adjust the tuning of some units, and they have a few in common. The mod
   that is applied last will take precedence.

   By default, mods are applied in simple alphabetical order. However, you can
   get more control by listing other mods in the 'before' and 'after' sections.

   If two mods specify an inconsistent ordering (e.g. they both ask to be before
   the other), SC will choose an arbitrary ordering and log a warning, but both
   mods will still be active.
```
Are you trying to improve the way the game handle conflict between 2 mods that are changing the same stuff (eg : stat of a unit). In this case, i don't remember a visual effect indicating which mod will be prioritized over the one; and the only solution for me to test that is to compare unit stat in game.

AugSphere:
> Are you trying to improve the way the game handle conflict between 2 mods that are changing the same stuff (eg : stat of a unit). In this case, i don't remember a visual effect indicating which mod will be prioritized over the one; and the only solution for me to test that is to compare unit stat in game.

I'm trying to make non-conflicting mods that depend on each other be correctly ordered, which is not the case right now. The current implementation of mod comparison does not sort mods correctly.

AugSphere:
> the only solution for me to test that is to compare unit stat in game.

The way I examined the effect is by adding logging statements inside `GetActiveModsFiltered`, which is called when a lobby is created. That way one can see which order the mods are loaded in directly. Then I created dummy mods with various before/after relationships and examined whether they get loaded in correct order or not.

shalkya:
using the protocol described previously, i was able to test the changes.

relent0r:
Hey AugSphere,
Is there a way a modder can intentionally get his mod to be the last one loaded? 
I have an AI mod that hooks alot of functions. But there are other AI mods that have the same hooks without conditional if statement so they will overwrite some of mine if loaded afterwards, which means I can't run the AI's together. With the existing sorting I just need to make sure my mods UUID is the highest value in order to make sure its loaded last.

AugSphere:
> Is there a way a modder can intentionally get his mod to be the last one loaded?
> I have an AI mod that hooks alot of functions. But there are other AI mods that have the same hooks without conditional if statement so they will overwrite some of mine if loaded afterwards, which means I can't run the AI's together. With the existing sorting I just need to make sure my mods UUID is the highest value in order to make sure its loaded last.

If you want to load your mod last, specifying UUIDs of all the mods it should be after in the `after` field of your mod's `mod_info.lua` should do the trick, as described [on the wiki](https://github.com/FAForever/fa/wiki/mod_info.lua-documentation). Do the mods you're working with specify an inconsistent ordering between each other (A saying it should go before B, but B saying it should go before A and so on)?

relent0r:
Thanks AugSphere,
AI mods don't usually have an ordering information. I didn't even know it existed :) Uveso does a bunch of work/testing for the default FAF AI's so he has to hook default functions so if his AI loads after mine it will replace mine, if mine loads after his it has conditional if statements which stops it overriding others so I've always tried to make mine load last. I'll look at achieving it with the after field, I imagine it means I'll need to update every time someone else updates an AI mod that causes this since their UUIDs will change.

AugSphere:
> 
> 
> Thanks AugSphere,
> AI mods don't usually have an ordering information. I didn't even know it existed :) Uveso does a bunch of work/testing for the default FAF AI's so he has to hook default functions so if his AI loads after mine it will replace mine, if mine loads after his it has conditional if statements which stops it overriding others so I've always tried to make mine load last. I'll look at achieving it with the after field, I imagine it means I'll need to update every time someone else updates an AI mod that causes this since their UUIDs will change.

Yup, you'll need to update it every time: that unfortunately inconvenient behaviour is by design. Tbf, you'd have to do the same to your UUID to make sure it remains the biggest one, unless your UUID was some specific big value instead of randomly generated and thus risking a collision with someone else trying to do the same thing.

relent0r:
It looks like you can achieve it just by specifying the mod names rather than UUIDs which will make it nice and easy and actually better than my current method, so thanks I learnt something today :)
I used to do it by just generating a UUID that starts with 'f' since its unlikely others will start with that.

PR #3134, opened 2020-07-14
Platoon movement

Garanas:
Introducing an alternative approach to move platoons around. The current approach forces the platoon to orient to the south at the end of every move order.  The new approach either aligns them to the line between the current and next node (...ByFuture) or between the current and past node (...ByPast).

To showcase, see the following video:
https://www.youtube.com/watch?v=e4LH9oaKwTQ

The video becomes especially interesting about 30 seconds in, when the platoons are moving to the north.

From the outer to the inner line:
 - (new) platoon:MovePathOrientedByFuture(...)
 - (new) platoon:MovePathOrientedByPast(...)
 - (old) platoon:MoveToLocation(...)
 - (old) platoon:Patrol(...)

Especially platoon:MovePathOrientedByPast(...) (2nd from the outer one) is promising: it appears to be very smooth and is able to keep up with one of the older approaches that has a lower total distance. They finish at the same time.

I intend to make similar functions for:
 - IssueFormAggressiveMove(...)
 - IssueFormAttack(...)
 - IssueFormPatrol(...)

Waiting for feedback on the coding style, naming and / or the taken approach before I continue.

### DIFF
--- a/lua/ScenarioPlatoonAI.lua
+++ b/lua/ScenarioPlatoonAI.lua
@@ -2,7 +2,7 @@
 -- File     :  /lua/ai/ScenarioPlatoonAI.lua
 -- Author(s):  Drew Staltman
 -- Summary  :  Houses a number of AI threads that are used in operations
--- Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
+-- Copyright ï¿½ 2005 Gas Powered Games, Inc.  All rights reserved.
 ------------------------------------------------------------------------
 
 local Utilities = import('/lua/utilities.lua')
--- a/lua/ScenarioPlatoonAI.lua
+++ b/lua/ScenarioPlatoonAI.lua
@@ -2,11 +2,7 @@
 -- File     :  /lua/ai/ScenarioPlatoonAI.lua
 -- Author(s):  Drew Staltman
 -- Summary  :  Houses a number of AI threads that are used in operations
-<<<<<<< HEAD
--- Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
-=======
 -- Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
->>>>>>> 2acb9e2fde7072a42d457403b24efa36077f4f64
 ------------------------------------------------------------------------
 
 local Utilities = import('/lua/utilities.lua')
--- a/lua/platoon.lua
+++ b/lua/platoon.lua
@@ -6620,26 +6620,29 @@ Platoon = Class(moho.platoon_methods) {
     IssuePatrolAlongRoute = function(self, path)
 
         -- check for optional / default values
-        local first = 1 
-        local last = table.getn(path)
         local formation = self.PlatoonData.UseFormation or 'NoFormation'
+
+        -- check if the parameters are correct
+        if not path or not (type(path) == 'table') then
+            error("IssuePatrolAlongRoute: The path is not a table. For paths with only one node, use { node } as the path.")
+            return { }
+        end
+
+        if table.empty(path) then
+            error("IssuePatrolAlongRoute: The path is empty.");
+            return { } 
+        end
 
         -- keep track of all the commands we issued
         local commands = { }
-
-        -- check if we have a path, if no path we can return immediately
-        if first > last then
-            return commands
-        end
 
         -- we have no formation, further computations are not required. We use this
         -- shortcut because calling IssueFormPatrol() with no formation causes them
         -- to not move at all.
         if formation == 'NoFormation' then 
             local units = self:GetPlatoonUnits()
-            for k = first, last do 
-                local point = path[k]
-                local command = IssuePatrol(units, point)
+            for k, node in path do
+                local command = IssuePatrol(units, node)
                 table.insert(commands, command)
             end
 
@@ -6649,7 +6652,7 @@ Platoon = Class(moho.platoon_methods) {
         -- check if we have a path of tables, instead to a path of vectors. A lot of the functionality provided by
         -- this library generates lists of tables instead of lists of vectors. Functionality in this file requires
         -- a list of vectors. Convert it if neccesary.
-        if not path[first].x then 
+        if not path[1].x then 
             local oldPath = path
             path = {}
             for k, node in oldPath do
@@ -6658,19 +6661,20 @@ Platoon = Class(moho.platoon_methods) {
         end
 
         -- store locally for better performance
+        local count = table.getn(path)
         local GetAngleCCW = Utilities.GetAngleCCW
         local GetDirectionVector = Utilities.GetDirectionVector
 
         -- pre-compute the angles
         local angles = { }
-        for k = first, last do 
+        for k = 1, count do 
     
             local curr = path[k - 1]
             local next = path[k]
     
-            -- if we're trying to look before the first node of the path, use the platoons current position instead
-            if k - 1 < first then 
-                curr = path[last]
+            -- if we're trying to look before the first node of the path, use the last node instead
+            if k - 1 < 1 then 
+                curr = path[count]
             end
 
             -- base orientation when the angle is 0 for the function IssueFormMove
@@ -6682,7 +6686,8 @@ Platoon = Class(moho.platoon_methods) {
 
         -- move over the path in formation
         local units = self:GetPlatoonUnits()
-        for k = first, last do 
+
+        for k = 1, count do 
             local point = path[k]
             local angle = angles[k]
             local command = IssueFormPatrol(units, point, formation, angle)
@@ -6699,17 +6704,21 @@ Platoon = Class(moho.platoon_methods) {
     -- @return Table of commands.
     IssueAggressiveMoveAlongRoute = function(self, path)
         -- check for optional / default values
-        local first = 1 
-        local last = table.getn(path)
         local formation = self.PlatoonData.UseFormation or 'NoFormation'
+
+        -- check if the parameters are correct
+        if not path or not (type(path) == 'table') then
+            error("IssueAggressiveMoveAlongRoute: The path is not a table. For paths with only one node, use { node } as the path.")
+            return { }
+        end
+
+        if table.empty(path) then
+            error("IssueAggressiveMoveAlongRoute: The path is empty.");
+            return { } 
+        end
 
         -- keep track of all the commands we issued
         local commands = { }
-
-        -- check if we have a path, if there is no path we can return immediately
-        if first > last then
-            return commands
-        end
 
         -- we have no formation, further computations are not required. We use this
         -- shortcut because calling IssueFormAggressiveMove() with no formation causes 
@@ -6717,9 +6726,9 @@ Platoon = Class(moho.platoon_methods) {
         if formation == 'NoFormation' then 
             -- store the commands / orders
             local units = self:GetPlatoonUnits()
-            for k = first, last do 
-                local point = path[k]
-                local command = IssueAggressiveMove(units, point)
+
+            for k, node in path do
+                local command = IssueAggressiveMove(units, node)
                 table.insert(commands, command)
             end
 
@@ -6729,7 +6738,7 @@ Platoon = Class(moho.platoon_methods) {
         -- check if we have a path of tables, instead of a path of vectors. A lot of the functionality provided by
         -- this library generates lists of tables instead of lists of vectors. Functionality in this file requires
         -- a list of vectors. Convert it if neccesary.
-        if not path[first].x then 
+        if not path[1].x then 
             local oldPath = path
             path = {}
             for k, node in oldPath do
@@ -6738,18 +6747,19 @@ Platoon = Class(moho.platoon_methods) {
         end
 
         -- store locally for better performance
+        local count = table.getn(path)
         local GetAngleCCW = Utilities.GetAngleCCW
         local GetDirectionVector = Utilities.GetDirectionVector
 
         -- pre-compute the angles
         local angles = { }
-        for k = first, last do 
+        for k = 1, count do 
     
             local curr = path[k - 1]
             local next = path[k]
     
             -- if we're trying to look before the first node of the path, use the platoons current position instead
-            if k - 1 < first then 
+            if k - 1 < 1 then 
                 local pos = self:GetPlatoonPosition()
                 curr = Vector(pos[1], pos[2], pos[3])
             end
@@ -6763,7 +6773,8 @@ Platoon = Class(moho.platoon_methods) {
 
         -- move over the path, store the commands
         local units = self:GetPlatoonUnits()
-        for k = first, last do 
+
+        for k = fi1rst, count do 
             local point = path[k]
             local angle = angles[k]
             local command = IssueFormAggressiveMove(units, point, formation, angle)
@@ -6780,17 +6791,21 @@ Platoon = Class(moho.platoon_methods) {
     -- @return Table of commands.
     IssueMoveAlongRoute = function(self, path)
         -- check for optional / default values
-        local first = 1 
-        local last = table.getn(path)
         local formation = self.PlatoonData.UseFormation or 'NoFormation'
+        
+        -- check if the parameters are correct
+        if not path or not (type(path) == 'table') then
+            error("IssueMoveAlongRoute: The path is not a table. For paths with only one node, use { node } as the path.")
+            return { }
+        end
+
+        if table.empty(path) then
+            error("IssueMoveAlongRoute: The path is empty.");
+            return { } 
+        end
 
         -- keep track of all the commands we issued
         local commands = { }
-        
-        -- check if we have a path, if there is no path we can return immediately
-        if first > last then
-            return { }
-        end
 
         -- we have no formation, further computations are not required. We use this
         -- shortcut because calling IssueFormMove() with no formation causes them 
@@ -6798,9 +6813,8 @@ Platoon = Class(moho.platoon_methods) {
         if formation == 'NoFormation' then 
             -- store the commands / orders
             local units = self:GetPlatoonUnits()
-            for k = first, last do 
-                local point = path[k]
-                local command = IssueMove(units, point)
+            for k, node in path do
+                local command = IssueMove(units, node)
                 table.insert(commands, command)
             end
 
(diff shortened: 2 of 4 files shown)


shalkya:
hello,
I've trouble replicating the issue of unit pointing to the south.
when you talk about platoon, it's something related about AI movement ?

Garanas:
It happens when you tell a platoon to move through scripting. For platoons moved by players this behavior does not happen.  For example, using:
 - platoon:MoveToLocation(...)
 - platoon:Patrol(...) 
 - IssueFormMove(...)

This is the example map from the video:
 - https://we.tl/t-9OfUFbCX2d

If you start the map, you should be able to observe the same behavior. Take note that the inner circles use the original behavior, the outer circles use the new behavior. 

When you open up the campaign, and start the first mission on normal you'll see that the Percivals that run around your base show the same behavior: on each way point they first orientate themselves south before they continue the patrol.

Garanas:
Renamed the functions slightly to match convention of other functions. Implemented the patrol and aggressive move.

Another showcase:
 - https://www.youtube.com/watch?v=HqxOKFYlxGE&feature=youtu.be

From the outer to the inner platoon:
 - (new) platoon:IssueMovePath(...)
 - (old) platoon:MoveToLocation(...)
 - (new) platoon:IssuePatrolPath(...)
 - (old) platoon:Patrol(...)

The corresponding test map from the video:
 - https://we.tl/t-u8ycHM6E9E

I am going to do a read through of the comments tomorrow, remove semicolons that I introduced again by habit and look for other quirky things.

Garanas:
I cannot see anything to change or add further, suggestions are welcome. I've also checked whether functions such as CPlatoon:IsMoving(squad) and function CPlatoon:IsPatrolling(squad) correctly pick up that the platoon is moving / patrolling with the new approach. They do.

lua/platoon.lua
```diff
@@ -6611,4 +6611,162 @@ Platoon = Class(moho.platoon_methods) {
...
+
+        -- defaults
+        first = first or 1 
+        last = last or table.getn(path)
+        local formation = self.PlatoonData.UseFormation or 'NoFormation'
+
+        -- check if we have a formation, IssueFormMove doesn't work if the formation argument is 'NoFormation'.
+        if formation == 'NoFormation' then 
+            WARN('MovePathOrientedByPast: No platoon formation provided, defaulting to GrowthFormation.')
+            formation = 'GrowthFormation'
```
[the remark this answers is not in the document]
Garanas: The functions IssueFormMove, IssueFormAggressiveMove, etc, do not work without a formation. I can add an if statement at the bottom to use IssueMove instead of IssueFormMove when no formation is provided.

lua/platoon.lua
```diff
@@ -6611,4 +6611,162 @@ Platoon = Class(moho.platoon_methods) {
...
+        end
+
+        -- check if we have a path, if no path we can return immediately
+        if first == last then
```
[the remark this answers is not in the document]
Garanas: Will fix that in the next commit!

lua/platoon.lua
```diff
@@ -6611,4 +6611,162 @@ Platoon = Class(moho.platoon_methods) {
...
+            local base = Vector( 0, 0, 1 )
+            local direction = Utilities.GetDirectionVector(next, curr)
+            local angle = Utilities.GetAngleCCW(base, direction)
+            angles[k] = angle
+        end
+
+        -- move over the path, store the commands
+        local commands = { }
+        local units = self:GetPlatoonUnits()
+        for k = first, last, 1 do
```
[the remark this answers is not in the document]
Garanas: Check, wasn't aware of the shortcut. I'll change it in the next commit.

Garanas:
That is a good question, if I would make these changes more 'global' in that sense: where would I start doing this? I haven't looked a lot into the code surrounding the AI's.

lua/platoon.lua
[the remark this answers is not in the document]
Garanas: Looking into it, how do you retrieve the desired formation of a platoon? I see the following line throughout platoon.lua:
`local formation = self.PlatoonData.UseFormation or 'NoFormation`

But that appears to be always nil. Calling LOG(repr(self)), which is the platoon in this context, doesn't show me a lot of information either.

Uveso:
If you want to see whats inside a table, you can't simply use repr() because inside classes are too many sub tables.
Use this function to only display the main entries inside a table:
```LUA
function DebugArray(Table)
    for Index, Array in Table do
        if type(Array) == 'thread' or type(Array) == 'userdata' then
            LOG('Index['..Index..'] is type('..type(Array)..'). I won\'t print that!')
        elseif type(Array) == 'table' then
            LOG('Index['..Index..'] is type('..type(Array)..'). I won\'t print that!')
        else
            LOG('Index['..Index..'] is type('..type(Array)..'). "', repr(Array),'".')
        end
    end
end
```
then call it with `DebugArray(self)` or `DebugArray(self.PlatoonData)`

When we get this to work, then you don't need to implement this into the normal AI.
I will do it in my next AI game patch or at least share the information to impement this into
our custom AIs mods for testing.

Garanas:
With your function, it shows:
```
NFO: Special debug:
INFO: Index[_c_object] is type(userdata). I won't print that!
INFO: Index[Trash] is type(table). I won't print that!
INFO: Index[PartOfAttackForce] is type(boolean). "\000false\000".
INFO: Index[PlatoonData] is type(table). I won't print that!
INFO: Index[CreationTime] is type(number). "\0000\000".
INFO: Index[EventCallbacks] is type(table). I won't print that!
```

With the LOG(repr()) it shows:
```
INFO: { <metatable=table: 1B285938>
INFO:   CreationTime=0,
INFO:   EventCallbacks={ OnDestroyed={ } },
INFO:   PartOfAttackForce=false,
INFO:   PlatoonData={ },
INFO:   Trash={ },
INFO:   _c_object=userdata: CScriptObject* at 1C350C68 = [CScriptObject at 0x1B2DE180]
INFO: }
```

Regardless of the information, the formation information is not in there. 

The platoons (of the video) are made with:

```
local u1 = ScenarioUtils.CreateArmyGroup(army, "p4", false);
local p1 = brain:MakePlatoon('', '');
brain:AssignUnitsToPlatoon(p1, u1, 'Attack', 'GrowthFormation')
```

The formation is applied / given to a squad. Looking through the definitions of the platoon.lua and cplatoonlua files, I cannot seem to be able to retrieve a squad (and then, in turn, its formation).

I did find:
self:GetSquadUnits('Scout')
function CPlatoon:SetPlatoonFormationOverride(formation) (...) end

But that doesn't allow me to retrieve the formation.

Given this information: my code doesn't look for squads of any kind - should I add that in?
And perhaps add a formation parameter?

Uveso:
```NFO: Special debug:
INFO: Index[_c_object] is type(userdata). I won't print that!
INFO: Index[Trash] is type(table). I won't print that!
INFO: Index[PartOfAttackForce] is type(boolean). "\000false\000".
INFO: Index[PlatoonData] is type(table). I won't print that!
INFO: Index[CreationTime] is type(number). "\0000\000".
INFO: Index[EventCallbacks] is type(table). I won't print that!```

My Debugarray only shows you the main or root values, as you can see there is a table called PlatoonData.
Now you need look into the PlatoonData table:
`DebugArray(self.PlatoonData)`
or
LOG( repr(self.PlatoonData) )

**Comment by Garanas** [comment]
Writing the following:

```
LOG("debug: platoon")
DebugArray(self);
LOG("debug: platoon data")
DebugArray(self.PlatoonData)

LOG("debug: end of debugging")
```

Produces in the console:

```
INFO: debug: platoon
INFO: Index[_c_object] is type(userdata). I won't print that!
INFO: Index[Trash] is type(table). I won't print that!
INFO: Index[PartOfAttackForce] is type(boolean). "\000false\000".
INFO: Index[PlatoonData] is type(table). I won't print that!
INFO: Index[CreationTime] is type(number). "\0000\000".
INFO: Index[EventCallbacks] is type(table). I won't print that!
INFO: debug: platoon data
INFO: debug: end of debugging
```

The table is empty. That is why I was looking at C functions  in the faf repository, perhaps the formation data is stored somewhere else.

Looking into it further shows the function:

```
-- inside platoon.lua
SetPlatoonData = function(self, dataTable)
    self.PlatoonData = table.deepcopy(dataTable)
end,
```

Given that the code from the repository attempts to use:

```
local PlatoonFormation = self.PlatoonData.UseFormation or 'NoFormation'
self:SetPlatoonFormationOverride(PlatoonFormation)
```

I started looking for where 'UseFormation' is defined, which is in the files:
 - fa\lua\AI\AIBuilders\AILandAttackBuilders.lua (18 occurences)
 - fa\lua\AI\AIBuilders\AISeaAttackBuilders.lua (5 occurences)
 - fa\lua\AI\AIBuilders\SorianLandAttackBuilders.lua (18 occurences)
 - fa\lua\AI\AIBuilders\SorianSeaAttackBuilders.lua (10 occurences)

These are 'Buildergroups', of which I am know they are related to platoons but unsure whether the data is being transferred to the platoon data table. If they are being transferred, then all the squads typically lose their original formation because they get overridden. 

In the GPG editor you can also define, or at least view, the platoon data manually. I am not sure whether that is being used given that one drop down mentions that it is being generated dynamically. 

That being said, this would make the function worthless for map-related code unless you specifically define the platoon data on your platoon, to pass the formation. Which would be quite hidden. Therefore I propose the following:

```
--- Moves the platoon along the path, orientating at each node to match the line from the previous node to the current node.
-- @param self The platoon itself.
-- @param path A table of positions, preferably of type Vector. Converted otherwise.
-- @param opts A table of optional values, containing:
--        opts.First the index of the first node the path
--        opts.Last the index of the last node of the path
--        opts.UseFormation the formation to use for the platoon, overrides the formation of the squads
-- @return Table of commands.
IssueMovePath = function(self, path, opts)
    -- check for optional / default values
    local first = opts.First or 1 
    local last = opts.Last or table.getn(path)
    local formation = opts.UseFormation or self.PlatoonData.UseFormation or 'NoFormation'

    -- This is done elsewhere too after checking for self.PlatoonData.UseFormation. Should we do this here?
    -- self:SetPlatoonFormationOverride(formation) 

   -- (the rest of the function)
end
```

If you manually define a formation, that formation is used. This is useful when the function is called through a map-based script. If that information is not provided, the typical place for a formation is checked. This is useful for AI-related calls. And if there is no formation there either, then no formation is used at all. Down below, the function IssueMove is called instead of IssueFormMove when no formation is provided.

In the gpg-code the call for overriding the squad formations always follows after a manual formation is set. Should we do this here too, when we override the platoons formation as a whole?

**Comment by Garanas** [comment]
@Uveso : I'd like to know what to add / change on the current code, or whether I can consider this to be done 😄  .

**Comment by Uveso** [comment]
Hey Garanas :)
I already talked to keyser about this PR.
My girl is visiting me at this weekend so i will do the check on next Tuesday (11.August)

File: lua/platoon.lua [hunk 4]
```diff
@@ -6611,4 +6611,246 @@ Platoon = Class(moho.platoon_methods) {
...
+    --        opts.UseFormation the formation to use for the platoon, overrides the formation of the squads
+    -- @return Table of commands.
+    IssuePatrolAlongRoute = function(self, path, opts)
+        -- check for optional / default values
+        local first = opts.First or 1 
+        local last = opts.Last or table.getn(path)
+        local formation = opts.UseFormation or self.PlatoonData.UseFormation or 'NoFormation'
+
+        -- This is done elsewhere too after checking for self.PlatoonData.UseFormation. Should we do this here?
+        -- self:SetPlatoonFormationOverride(formation)
```
Uveso, line 6630: We don't need this, we are using IssueFormPatrol to set the formation.

lua/platoon.lua
```diff
@@ -6611,4 +6611,246 @@ Platoon = Class(moho.platoon_methods) {
...
+            local next = path[k]
+    
+            -- if we're trying to look before the first node of the path, use the platoons current position instead
+            if k - 1 < first then 
+                curr = path[last]
+            end
+
+            -- base orientation when the angle is 0 for the function IssueFormMove
+            local base = Vector( 0, 0, 1 )
+            local direction = Utilities.GetDirectionVector(next, curr)
```
Uveso, line 6677: Those function are called in a loop, please make them local for better performance
`local GetDirectionVector = Utilities.GetDirectionVector`

lua/platoon.lua
Garanas: Agreed!

lua/platoon.lua
```diff
@@ -6611,4 +6611,246 @@ Platoon = Class(moho.platoon_methods) {
...
+    
+    --- Patrols the platoon along the path, orientating at each node to match the line from the previous node to the current node.
+    -- @param self The platoon itself.
+    -- @param path A table of positions, preferably of type Vector. Converted otherwise.
+    -- @param opts A table of optional values, containing:
+    --        opts.First the index of the first node the path
+    --        opts.Last the index of the last node of the path
+    --        opts.UseFormation the formation to use for the platoon, overrides the formation of the squads
+    -- @return Table of commands.
+    IssuePatrolAlongRoute = function(self, path, opts)
```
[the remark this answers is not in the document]
Garanas: The idea is that typically the first and last node are nodes that are important for path finding, but they aren't necessarily desirable to move towards to. For example, the first node may be behind the platoon and the last node past the target.

This can be solved outside this function however, hence I am removing the parameters as a whole. 

Also removing the formation parameter - I noticed that other platoon functions (such as in ScenarioPlatoonAI.lua) extensively use the PlatoonData variable for this, hence it is more 'common' that I first anticipated it was! I'll update the comments too to fit in line with the rest of the AI functions.

lua/platoon.lua
Garanas: Done!

Do you have an article as to why this is faster? I tried to find one myself, but wasn't able to

Garanas:
Link for the latest version of the map:
 - https://we.tl/t-wW3F7ZKP5R

Went through all the changes Uvseo recommended.

lua/platoon.lua
```diff
@@ -6611,4 +6611,263 @@ Platoon = Class(moho.platoon_methods) {
             end
         end
     end,
+    
+    --- Patrols the platoon along the path, orientating at each node to match the line from the previous node to the current node.
+    -- @param self The platoon itself.
+    -- @param path A table of positions, preferably of type Vector. Converted otherwise.
+    -- @param self.PlatoonData.UseFormation The formation to apply, such as GrowthFormation, AttackFormation or NoFormation.
+    -- @return Table of commands.
+    IssuePatrolAlongRoute = function(self, path)
```
Uveso, line 6620: Please add the formation as argument to overwrite the `PlatoonData.UseFormation`.
`IssuePatrolAlongRoute = function(self, path, formation)`
so we can call this with 
`platoon:IssuePatrolAlongRoute(path3, 'AttackFormation');`
I need this to overwrite the formation settings in close combat.
(units that need to move fast away from the enemy should not use the formation saved in `PlatoonData.UseFormation`)

lua/platoon.lua
```diff
@@ -6611,4 +6611,263 @@ Platoon = Class(moho.platoon_methods) {
...
+
+        -- check for optional / default values
+        local formation = self.PlatoonData.UseFormation or 'NoFormation'
```
Uveso, line 6623: To check the extra argument change it to
`local formation = formation or  self.PlatoonData.UseFormation or "NoFormation"`

lua/platoon.lua
```diff
@@ -6611,4 +6611,263 @@ Platoon = Class(moho.platoon_methods) {
...
+        -- shortcut because calling IssueFormPatrol() with no formation causes them
+        -- to not move at all.
+        if formation == 'NoFormation' then 
+            local units = self:GetPlatoonUnits()
+            for k, node in path do
+                local command = IssuePatrol(units, node)
+                table.insert(commands, command)
+            end
+
+            return commands
```
Uveso, line 6649: Why it's needed to return the table with movecommands ? 
They are already issued, and we could save some memory garbage here.
(the LUA garbage collector eats an insane amount of CPU and RAM speed.)

lua/platoon.lua
```diff
@@ -6611,4 +6611,263 @@ Platoon = Class(moho.platoon_methods) {
...
+        local commands = { }
+
+        -- we have no formation, further computations are not required. We use this
+        -- shortcut because calling IssueFormPatrol() with no formation causes them
+        -- to not move at all.
+        if formation == 'NoFormation' then 
+            local units = self:GetPlatoonUnits()
+            for k, node in path do
+                local command = IssuePatrol(units, node)
+                table.insert(commands, command)
```
Uveso, line 6646: If we don't return the table then we don't need to fill it here
(see `Why its needed comment` below)

Uveso (changes requested):
The map is no longer available as download, if someone want to test this.
Overall everything is running.
Please change the functions to accept a formation override argument.
Also if possible, can you rebase the PR to the latest game version please ?

Garanas:
I'll do the changes again, I think its best if we sit down some day this week to finish it up with direct and quick communication!

lua/platoon.lua
Garanas: I'll change it back :)

lua/platoon.lua
Garanas: I'll change it back :)

lua/ScenarioPlatoonAI.lua
```diff
@@ -2,7 +2,7 @@
 -- File     :  /lua/ai/ScenarioPlatoonAI.lua
 -- Author(s):  Drew Staltman
 -- Summary  :  Houses a number of AI threads that are used in operations
--- Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
+-- Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
```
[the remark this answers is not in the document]
Garanas: No, it wasn't. I'll revert the change. I was looking at how other functions related to platoons formulate their top-level function comments.

lua/platoon.lua
Garanas: Because one may want to check something when a command is done, I intend to use it in my survival to re-check its target at each node of the path.

Uveso:
Yes sure, but its already finished.
The platoon-override is for mission scripting and for Azraels, Relentors and my AI.
I already tested the implementation to the normal AI, but i will patch this to the default AI in one of the next AI patches.
It's a bit more work, since not all platoons uses PlatoonData.UseFormation.

Issue #3101, opened 2020-05-10
Wreck hitbox


shalkya:
![image](https://user-images.githubusercontent.com/11618427/81501771-5da76a00-92da-11ea-9b2b-d2f8975e25d6.png)
Once of the most obvious issue is with the GC. The hitbox doesn't change its position after the GC fall down.

Garanas:
I'd like to investigate this further but cannot find the console command to show those bounding boxes. What console commander did you use for it?

shalkya:
if you use the default preset for hotkey it's ctrl-shift-alt-C (Toggle collision debug info).

Garanas:
Well, I would've never found that. Thank you for sharing!

I think I solved it:
[image]<a href="https://ibb.co/kqxNDs9"><img src="https://i.ibb.co/HGV5Pfx/collision-fix-screenshot.png" alt="collision-fix-screenshot" border="0"></a>[!image]

In the death thread of the unit the collision box needs to be adjusted! I'll make a pull request, any other units that needs these kind of fixes? I can think of the Seraphin t2 Assault Bot from the top of my head.

shalkya:
nicely done, you changed it in the unit script ?
i think it would be better to change the main death thread, so it applies for all units. So we don't change 1 by 1 the script for all relevant units.

Garanas:
I changed this in the unit script, see also: https://github.com/FAForever/fa/pull/3139

That being said, I don't know how to dynamically compute the bounding box of a unit. I tweaked this bounding box manually until it fit right in, can we access the mesh (vertex) information somehow?

Lets continue the discussion in the pull request :).

PR #3139, opened 2020-07-27
Fix for various hit boxes for wreckages

Garanas:
Solves the wreck bounding box issue described in #3101.

Will add in more units over the time of the pull request, if you find any unit that should have their wreck bounding box fixed but hasn't been included yet, please do comment!

Current list of fixed wreck bounding boxes:
 - Galactic Colossus


shalkya:
maybe you can access the position of bones. You can see them in game doing alt-V and alt-B (to get the name)

Garanas:
Where do you find all these shortcuts? I'll look into it!

Garanas:
Given a GC, this would produce a reasonable boundingbox:
<a href="https://ibb.co/kH9C737">bones-check</a>

It would be larger, however, allowing for it to take damage on places where there is 'no GC'.

That being said, it doesn't work for all units:
<a href="https://ibb.co/30QM56M">bones-check-2</a>

The blue surrounded set of bones is a T3 Cybran Static Artillery (Disruptor?), which would have almost no collision box if we'd compute it in a general manner. Let alone the wall segments: they only have one bone. Their collision box would be a point.

The red boxes (produces by alt-V + alt-B) appear to be useful, but when the GC falls the red box is unchanged. See also the first picture.

I think doing this manually isn't too bad.

Garanas:
Using the function Entity:GetCollisionExtents() doesn't help us either: it is based off the collision shape that you set by Entity:SetCollisionShape(...).

units/UAL0401/UAL0401_Script.lua
```diff
@@ -72,7 +72,17 @@ UAL0401 = Class(AWalkingLandUnit) {
         end
 
         self:DestroyAllDamageEffects()
-        self:CreateWreckage(overkillRatio)
```
speed2CZ, line 75: No, you cant hardcode this like that. You have to find a solution to cover all the units and so it doesnt break if the hitbox of the unit is changed

shalkya:
`Where do you find all these shortcuts? I'll look into it!`
just press f1 in game to access the shortcut menu

units/UAL0401/UAL0401_Script.lua
Garanas: I understand that this cannot be done with this approach if all 300 units need an edit like this. That would be a nightmare to do and a nightmare to maintain.

In practice I cannot think of that many units that really have noticeable problems like the Colossus does. Perhaps the Seraphim T2 Assault Bot too, but due to its size it is a lot less noticeable that it's indeed off. 

Also, I don't think the wreckage hitbox should have some relation to the non-wreckage hitbox. They can be completely different, as for the Seraphim T2 Assault bot: if you want it to be accurate, the hitbox would stretch which is hard to determine dynamically and for the colossus it would rotate.

To meet your suggestion and to understand the requirements:
 - The wreckage hitbox is a function that takes the original hitbox and returns a new hitbox based on the original hitbox, but then (for example) rotated 90 degrees or scaled over some axis.
 - The transformation must be dynamically decided and computed for each unit.

Without access to the mesh I don't think this can be done. And with access, it would be expensive to compute. Bones are very unreliable (not all root bones start at the ground. Not all units have very predictable / useful bones. Bones can also get really messed up. We'd need to know the state of the bones before the unit does its death animation to determine the required orientation / scale / translation).

Perhaps we can add information to the blueprints, of which I am unsure is possible / accepted, allowing it to be more easily maintained. We can then define the wreck hitbox. If this new blueprint data exists, we use that to define the hitbox for the wreckage and otherwise we use the default hitbox for the wreckage. This fallback prevents us from adding the information to all 300 units.

Garanas:
Thank you for the tip on the F1 key in Supreme Commander!

Garanas:
Looking into the suggestion I made in response to the request of change by Speed2, I've tried:
 - Defining SizeX, SizeY, SizeZ and CollisionOffsetX, CollisionOffsetY, CollisionOffzetZ in the wreckage bit of the blueprint
 - Adjusting the CreateWreckage function in wreckage.lua:

Inside UAL0401_unit.bp (blueprint of the colossus):
``` lua
    Wreckage = {
        Blueprint = '/props/DefaultWreckage/DefaultWreckage_prop.bp',
        EnergyMult = 0,
        HealthMult = 0.9,
        MassMult = 0.9,
        ReclaimTimeMultiplier = 1,
        CollisionOffsetX = 0,  -- added
        CollisionOffsetY = 0,  -- added
        CollisionOffsetZ = 3,  -- added
        SizeX = 2.5,           -- added
        SizeY = 1.5,           -- added
        SizeZ = 6.5,           -- added
        WreckageLayers = {
            Air = false,
            Land = true,
            Seabed = true,
            Sub = true,
            Water = true,
        },
    },

```

Inside wreckage.lua:
Old:
``` lua
    prop:SetScale(bp.Display.UniformScale)
    prop:SetPropCollision('Box', bp.CollisionOffsetX, bp.CollisionOffsetY, bp.CollisionOffsetZ, bp.SizeX * 0.5, bp.SizeY * 0.5, bp.SizeZ * 0.5)

    prop:SetMaxHealth(bp.Defense.Health)
```
New:
``` lua
    prop:SetScale(bp.Display.UniformScale)

    -- try to use collision hitbox data defined in the wreck, otherwise use from the blueprint
    cx = wreck.CollisionOffsetX or bp.CollisionOffsetX
    cy = wreck.CollisionOffsetY or bp.CollisionOffsetY
    cz = wreck.CollisionOffsetZ or bp.CollisionOffsetZ
    sx = wreck.SizeX or bp.SizeX
    sy = wreck.SizeY or bp.SizeY
    sz = wreck.SizeZ or bp.SizeZ  
    prop:SetPropCollision('Box', cx, cy, cz, sx * 0.5, sy * 0.5, sz * 0.5)

    prop:SetMaxHealth(bp.Defense.Health)
```
Which reverses the problem when the GC dies without its animation:

<a href="https://ibb.co/9vrpRhN">collision-fix-bad</a>

The animation is the cause of all the troubles: I do not know how to take that into account at such a low / generic level in the code.

shalkya:
honestly the death weapon not firing when gc get reclaim is just silly. I think that was fixed for the ythota to allow the lighting to spawn. The only case we want to keep it is with volatile structure, in this case you can denie the explosion, which now is part of the gameplay.

Garanas:
The death animation does trigger when you reclaim the Colossus. I just tested it. The wreckage you are seeing there is made through code:

``` lua
  local gc = CreateUnitHPR('UAL0401', 'ARMY_1', 512, 0, 512, 0, 0, 0)
  local wreckage = gc:CreateWreckage( 0 )
```

Which is also how it is done when the you start the map with wreckage on it:
``` lua
function CreateWreckage(unit, needToRotate)
    prop = unit:CreateWreckageProp(0) -- calls Wreckage.CreateWreckage in turn
    (...)
    unit:Destroy()
end
``

**Comment by shalkya** [comment]
in game : image
don't ctrl k unit, but them killed instead.
Wreck at start of the map looks like they have been killed : https://github.com/FAForever/fa/pull/2544

**Comment by Garanas** [comment]
Ah, when it dies the usual way. I didn't know there was a difference with CTRL + K and killing a unit. Didn't know about that fix either, pretty cool to know! Learned a lot myself during this pull request.

Tomorrow I'll make a map with all the units of a typical game and compare their wrecks with their bounding boxes. Just to see how many of them actually have this type of issue.

**Comment by Garanas** [comment]
I've taken a look at it, these are the units that would have an issue:
 - Monkeylord (bone-animation)
 - Colossus (bone-animation)
 - aeon sniperbot (bone-animation)
 - percival (bone-animation)
 - Megalith (bone-animation)
 - Ythotha (bone-animation)
 - Seraphim sniperbot (bone-animation)

The script I used to spawn all the units:
``` lua
function CreateUnit(blueprint, x, z)
    return pcall (CreateUnitHPR, blueprint, 'ARMY_1', x, 0, z, 0, 0, 0)
 end

function TestWrecksList(set, z)
    local x = 10
    for k = 0, 4 do 
        for l = 0, 15 do 
            local bp = set .. string.format('%d', k) .. string.format('%02d', l)

            -- skip commanders
            if not string.find(bp, '0001') then 
                local ok1, result1 = CreateUnit(bp, x, z)
                local ok2, result2 = CreateUnit(bp, x, z + 13)
                if ok2 then 
                    result2:Kill()
                end

                if ok1 and ok2 then 
                    x = x + 10
                end
            end
        end
    end
end

function FactionSpawn(version, faction)
    TestWrecksList(version .. faction .. 'a0', 10)
    TestWrecksList(version .. faction .. 'b0', 35)

    TestWrecksList(version .. faction .. 'b1', 60)
    TestWrecksList(version .. faction .. 'b2', 85)

    TestWrecksList(version .. faction .. 'b3', 110)
    TestWrecksList(version .. faction .. 'b4', 135)

    TestWrecksList(version .. faction .. 'b5', 160)
    TestWrecksList(version .. faction .. 'c0', 185)

    TestWrecksList(version .. faction .. 'l0', 210)
    TestWrecksList(version .. faction .. 's0', 235)
end

function TestWrecks()

    -- a = aeon
    -- e = uef
    -- r = cybran (replicators!)
    -- s = seraphim
    local uefcybranaeonserpahim = 's'

    -- u = old units
    -- x = new units
    -- z = hq factories
    local version = 'x'

    FactionSpawn(version, uefcybranaeonserpahim)
end
```

I've also checked, sea and air units do move their wrecks (they fall / sink) but they orientate properly. 
``` lua
    -- unit.lua
    -- this codes make navy units sink, which is an animation
    PlayAnimationThread = function(self, anim, rate)
        local bp = self:GetBlueprint().Display[anim]
        if bp then
            local animBlock = self:ChooseAnimBlock(bp)
            if animBlock.Mesh then
                self:SetMesh(animBlock.Mesh)
            end
            if animBlock.Animation and (self:ShallSink() or not EntityCategoryContains(categories.NAVAL, self)) then
                local sinkAnim = CreateAnimator(self)
                self:StopRocking()
                self.DeathAnimManip = sinkAnim
                sinkAnim:PlayAnim(animBlock.Animation)
                rate = rate or 1
                if animBlock.AnimationRateMax and animBlock.AnimationRateMin then
                    rate = Random(animBlock.AnimationRateMin * 10, animBlock.AnimationRateMax * 10) / 10
                end
                sinkAnim:SetRate(rate)
                self.Trash:Add(sinkAnim)
                WaitFor(sinkAnim)
                self.StopSink = true
            end
        end
    end,
```
This piece of code then moves the wreck to the terrain height, throughout the animation the box collider is still at the top and it doesn't move at all.
```
        -- unit.lua, CreateWreckageProp(...)
        if not halfBuilt and (layer == 'Air' or EntityCategoryContains(categories.NAVAL - categories.STRUCTURE, self)) then
            pos[2] = GetTerrainHeight(pos[1], pos[3]) + GetTerrainTypeOffset(pos[1], pos[3])
        end
```
For air units the same code is used to position the wreckage, but I cannot find where the air units do their 'aeriel' movements when they fall from the skies. I know that isn't some animation because they can bounce off shields and their hit box is rotated, where as all animated deaths do not change the hit box at all.
``` lua
            -- defaultunits.lua
            -- Create a projectile we'll use to interact with Shields
            local proj = self:CreateProjectileAtBone('/projectiles/ShieldCollider/ShieldCollider_proj.bp', 0)
            self.colliderProj = proj
            proj:Start(self, 0)
            self.Trash:Add(proj)
```
I cannot find where this projectile is actually being used / linked for the described behavior. I suppose this is done by the engine and not in LUA because it takes part in the physics simulation.

TLDR: animations are not taken into account for the collision box. And that is why the given list of units are a problem, they animate away from their original position.

Further analysis:
The DeathThread function of units.lua calls:
``` lua
    -- units.lua
    --- Called at the end of the destruction thread: create the wreckage and Destroy this unit.
    DestroyUnit = function(self, overkillRatio)
        self:CreateWreckage(overkillRatio or self.overkillRatio)

        -- wait at least 1 tick before destroying unit
        WaitSeconds(math.max(0.1, self.DeathThreadDestructionWaitTime))

        self:Destroy()
    end,
```
Which in turn creates the wreck. For the list of units at the top, the bold ones do not call the original DeathThread function:
 - **Monkeylord**
 - **Colossus**
 - aeon sniperbot
 - percival
 - **Megalith**
 - **Ythotha**
 - Seraphim sniperbot

Hence, solving this all the way up wouldn't solve the original problem for the colossus.

My proposal:
We add in the following properties into the blueprints of the corresponding units, at the same location where the original values are defined:
``` lua
    CollisionOffsetWreckAnimatedX
    CollisionOffsetWreckAnimatedY
    CollisionOffsetWreckAnimatedZ
    SizeWreckAnimatedX = 0.45,
    SizeWreckAnimatedY = 1.55,
    SizeWreckAnimatedZ = 0.9,
```
We can then use this data along with a toggle: if the unit died animated, we use these values. Otherwise we use the original values.

One approach for the toggle is to send it down (OnKilled (this is where we know we are animating) -> DeathThread -> DestroyUnit -> CreateWreckage -> CreateWreckageProp -> Wreckage.lua.Wreckage -> Wreckage.lua.WreckageProp) as a parameter with a default value of false so that we do not break any code.

Another approach is to apply this check inside Wreckage.lua.Wreckage:
``` lua
        local bpDeathAnim = bp.Display.AnimationDeath
        if bpDeathAnim and table.getn(bpDeathAnim) > 0 then
            self.PlayDeathAnimation = true
        end
```
But then we hold the same situation: if the unit is did not use its animation when it died for some reason, the collision box is off again. Which can happen when you reclaim the Colossus right before it dies.

For the experimentals that overrides the DeathThread, calling the new function in turn is an easy fix. And for any future units with long animations when they get killed can use the new blue print values.

I am going to implement the former later today, just to see how it looks.

shalkya:
```
--- Create a wreckage prop.
function CreateWreckage(bp, position, orientation, mass, energy, time)
    local bpWreck = bp.Wreckage.Blueprint

    local prop = CreateProp(position, bpWreck)
    prop:SetOrientation(orientation, true)

    prop:SetScale(bp.Display.UniformScale)
    prop:SetPropCollision('Box', bp.CollisionOffsetX, bp.CollisionOffsetY, bp.CollisionOffsetZ, bp.SizeX * 0.5, bp.SizeY * 0.5, bp.SizeZ * 0.5)

    prop:SetMaxHealth(bp.Defense.Health)
    prop:SetHealth(nil, bp.Defense.Health * (bp.Wreckage.HealthMult or 1))
    prop:SetMaxReclaimValues(time, mass, energy)

    --FIXME: SetVizToNeurals('Intel') is correct here, so you can't see enemy wreckage appearing
    -- under the fog. However the engine has a bug with prop intel that makes the wreckage
    -- never appear at all, even when you drive up to it, so this is disabled for now.
    --prop:SetVizToNeutrals('Intel')
    if not bp.Wreckage.UseCustomMesh then
        prop:SetMesh(bp.Display.MeshBlueprintWrecked)
    end

    -- This field cannot be renamed or the magical native code that detects rebuild bonuses breaks.
    prop.AssociatedBP = bp.Wreckage.IdHook or bp.BlueprintId

    return prop
end
```
can't we just change this function in wreckage.lua, it should also apply to exp that use/modify it in their unit.script ?

`But then we hold the same situation: if the unit is did not use its animation when it died for some reason, the collision box is off again. Which can happen when you reclaim the Colossus right before it dies.`
I think we shouldn't take as excuse the fact that other bug are laying around.

speed2CZ:
Well it is intended behaviour, that reclaiming doesnt trigger explosions and stuff.

You could set some flag on the wreckage like DeathAnimationPlayed = true

Since you know if it played and depending on that, adjust the hitbox

Garanas:
Is this the more generic approach? For any unit that:
 - Did do a death animation
 - Has some animation offset in its wreckage blueprint
Will use the new values for the hit box defined in the wreckage, otherwise we keep the old ones.

However:
<a href="https://ibb.co/8sqqXsk">collision-huh</a>

When the unit is reclaimed, the hit box doesn't change. But when the unit dies by hostile fire during reclaiming, the hit box does change. The animation is supposed to run, therefore the flag is set, but cancelled somewhere else.

I cannot find where it is being cancelled.

shalkya:
hello there was a fix for reclaiming preventing death animation : https://github.com/FAForever/fa/pull/2545/commits/b39a7e84a50d69ca4a0800514c57b315ae050e24
but this apply to all type of unit. If you can make it not apply to volatile structure (apart the paragon, like currently), then i will happily merge this PR.

Garanas:
Appears to work properly, I've tested the GC with:
 - Just killing it (works)
 - Reclaiming it while its being shot (works)
 - Reclaiming it (works)
 - CTRL + K (works)
 - Just creating the wreckage prop (works)

The correct hit box is used time.

I've changed the wreckage hit box of various units depending on the animation they play. I've made a screenshot for each hit box / animation for you to verify: https://drive.google.com/file/d/1160IRiKQk5zdRD42lbER1FtMsCvxfItT/view?usp=sharing

PR #3140, opened 2020-07-29
Shaders Needed for SCTA Mod

Dragun123:
TA Shaders coded by Armored Fish for SCTA Mod for build animations.

### DIFF
--- a/effects/mesh.fx
+++ b/effects/mesh.fx
@@ -8139,6 +8139,489 @@ technique NomadsPhaseShield2_LowFidelity
     }
 }
 
+/////////////////////
+// TA Build Shader //
+/////////////////////
+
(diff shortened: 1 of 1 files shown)


speed2CZ:
This doesnt belong here, it needs to be part of the mod. Same as nomads.

Dragun123:
> This doesnt belong here, it needs to be part of the mod. Same as nomads.

Except a mod doesn't have the ability to modify a mesh.fx file without being a featured mod/having its own init file. If I put this file/modification in my mod it wouldn't actually affect or change anything.

shalkya:
I saw that the mesh.fx was modified by nomads already, so I guess we can do the same for scta ? If someone is against this speak up.

PR #3146, opened 2020-08-05
Increase nuke and exp nuke hp 1000 times. Also increase SMD damage.

Kazbek:
Fix nuclear missle can be killed by satellite (and others).


JaggedAppliance:
Thanks for making this pr, we will discuss it in the balance team very soon.

JaggedAppliance:
We decided we want to implement this. Can you point this pr at deploy/fafbeta and also split the seraphim experimental nuke into two commits; one that adds the original file as it is in the base game and a second that makes the changes to the file.

Kazbek:
@JaggedAppliance I guess its done #3155

JaggedAppliance:
merged #3155

PR #3151, opened 2020-08-07
update UEF SCU rebalance

TurinturambarFAF:
updates SCU upgrade and preset descriptions and naming


Dragun123:
Currently UEF SACU is it meant to be able to build two Drones or just one? If two, the SACU Engi Com only starts with one of the two drones built and is unable to build the second

PR #3158, opened 2020-08-23
death explosion t4 nuke launcher

shalkya:
make distortion field visible (will also impact aeon nukes)
add visual effect to T4 nuke launcher death explosion
20k damage inner ring of 15
5k damage outer ring of 20


FemtoZetta:
Why? :(

shalkya:
Because it's a nuke, and all nuke deal damage when killed.

myEmperorr:
Does it have any balance benefits? 
Why make an almost useless unit even less desirable?

PR #3160, opened 2020-08-26
remove move shift g

shalkya:
do you think this is enough to prevent cheating ?

### DIFF
--- a/lua/spreadattack.lua
+++ b/lua/spreadattack.lua
@@ -405,7 +405,7 @@ function GiveOrders(Data)
     if not IssueOrderFunctions then
         IssueOrderFunctions = {
          ["Attack"]             = IssueAttack,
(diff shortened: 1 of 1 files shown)


zlo-faf:
Some players are concerned if you will still be able to queue spread attack after move orders. E.G. queue to move strats around aa and then press shift+g. will move orders dissapear or what?

RobustnessFAF:
I'm against

Azraeel:
I'm against theres gotta be a better way then removing a extremely used feature of the game.

InspektorKotFAF:
im against. This is better function for a spam formation. Without it units will be splitted most time and fight like a flock of sheep

Techmind:
There are Issue* commands anyway, the only stuff spread attacks does is "batching" of those commands, so they are a bit more efficient than issuing them directly.

PR #3163, opened 2020-09-08
Slight increase of Percival health to "Brick+100" to maintain UEF faction theme

venomdragoon:
This is to restore the UEF Percival to have slightly higher health than the brick. UEF design theme is to always have slightly higher health than their counterparts. This slight increase should not affect balance, but restores the theme of the units to their original standing prior to T3 rebalance (9300 hp percival vs 9000 hp brick).


JaggedAppliance:
400 extra HP will definitely impact balance. If Percivals are to have more HP then other things will also have to change because right now they are well balanced.

venomdragoon:
Yeah, good point. 
What about if a proportional (5%) cost increase went with it as well?

PR #3173, opened 2020-09-22
V1 Sorian Update [Support Factories]

Azraeel:
Fixed Sorian, so He'll be able to make support factories.


shalkya:
this change were pushed to faf by mistake and actually break the AI see : https://forum.faforever.com/topic/668/new-patch-vs-sorian-ai
I will revert that in an hotfix. Could you fix them on your side ?

Dragun123:
It doesn’t look Azreal fixed this commit and it should be unmerged. It broke the AI last time.

shalkya:
well as you can i did already revert those changes in here : https://github.com/FAForever/fa/commit/626180f88340c2563be33eae0bec65da3da5c7e3

PR #3174, opened 2020-09-23
Quantum optics facility fix

### LINKED ISSUE #3172: Vision markers are broken
This commit broke it https://github.com/FAForever/fa/commit/ca16b123abb89655ff0f38a63ed1c1607d623654

Especially this part: `if self.WaterVision != false then` and  `if self.WaterVision then` are not the same thing (well all those ifs that got changed.)
The original code created all the visions unless it was specifid not to. New code creates only visions if they are specified.

This breaks all visions markers in all coop missions and I dare to say even some more, since it's significant change of how the vision markers work.

Please revert this change and just keep the optimization of how the Eye works. Thanks in advance!

KionX:
#3172


shalkya:
also fix T1 uef arty vision shell and T1 sera scout vision after death.

PR #3175, opened 2020-09-23
Weather in Supreme Commander

Garanas:
Weather is available in Supreme Commander but not commonly used. A reason is that people don't know about it, it is not documented properly and then there is the issue with vision: if the markers are not in vision then you won't see any cloud at all.

This pull request refactors the weather lua files, they were a bit of a mess. There is a clear opt-out section, clear data transformations functions and overall it feels more solid. By default weather is now visible through the fog which is different behavior than the original behavior.

For more information on weather:
 - https://forum.faforever.com/topic/277/about-weather-generating-clouds

For maps with weather:
 - Rainmakers
 - Hardshield Oasis
 - Adaptive Pandemonium Pass
 - Perrote Glacier

Edit: because these files are new to the repo you cannot see the original files. Essentially I refactored the entire file and there is not much left of the original file: it was a bit of a mess and I cleaned it all up, made the transformations clear and added in the feature to allow clouds to spawn in the fog.

Edit2: With thanks to Speed2 the first commit are the original files, the second is the improved weather system. Yes, the commit name / description is poor. I've received feedback on that too 👍

### DIFF
--- a/lua/weather.lua
+++ b/lua/weather.lua
@@ -1,0 +1,268 @@
+local WeatherDefinition = import('/lua/weatherdefinitions.lua')
+local MapStyleList = WeatherDefinition.MapStyleList
+local MapWeatherList = WeatherDefinition.MapWeatherList
+local ScenarioUtils = import('/lua/sim/ScenarioUtilities.lua')
+local util = import('/lua/utilities.lua')
+local Entity = import('/lua/sim/Entity.lua').Entity
+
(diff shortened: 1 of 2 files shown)


shalkya:
To see the difference you should try playing on "Mezmerizing Paradise betav1" (the map Jip gave already add the fix in their map script). This way you can see the clouds appear when using the fix from this PR.

PR #3177, opened 2020-10-01
Allow for custom vault path

1-alex98:
Variable can be set in faf_path.lua

### DIFF
--- a/init_faf.lua
+++ b/init_faf.lua
@@ -171,7 +171,7 @@ end
 
 load_vault(InitFileDir .. '\\..\\user\\My Games\\Gas Powered Games\\Supreme Commander Forged Alliance')
 load_vault(SHGetFolderPath('PERSONAL') .. 'My Games\\Gas Powered Games\\Supreme Commander Forged Alliance')
-if custom_vault_path != nil then
+if custom_vault_path ~= nil then
 	LOG('Loading custom vault path' .. custom_vault_path)
 	load_vault(custom_vault_path)
 end
--- a/init_faf.lua
+++ b/init_faf.lua
@@ -161,13 +161,6 @@ function mount_mod_sounds(MODFOLDER)
     end
 end
 
-load_vault(InitFileDir .. '\\..\\user\\My Games\\Gas Powered Games\\Supreme Commander Forged Alliance')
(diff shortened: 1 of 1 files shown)


init_faf.lua
```diff
@@ -176,6 +169,14 @@ local function load_vault(vault_path)
 	mount_contents(vault_path .. '\\maps', '/maps')
 end
 
+load_vault(InitFileDir .. '\\..\\user\\My Games\\Gas Powered Games\\Supreme Commander Forged Alliance')
+load_vault(SHGetFolderPath('PERSONAL') .. 'My Games\\Gas Powered Games\\Supreme Commander Forged Alliance')
+if custom_vault_path != nil then
```
speed2CZ, line 174: ~= in Lua

speed2CZ: Just trying to maintain the same format across all files we work with

axel1200:
If we should merge that this change would need to manually deployed to the server. Also what is the difference between init.lua and init_faf.lua . Which other init.luas should I change

speed2CZ:
Idk, you tell me :D you're the client guy. Anyway each featured mod has it's own init file.

axel1200:
@speed2CZ they are all based on faf one tho... or at least they should be

axel1200:
I like this... whether it is used or not it will not break stuff right?

shalkya:
Ok so coming back on this one : 
init upload have been fixed. So it's relevant to look into these PR again.

"Also what is the difference between init.lua and init_faf.lua."
init_faf.lua is used in the server to create the init.lua that is used with the game (main mod FAF).
other init like init_fafdevelop.lua are used similarly (to deploy the init.lua) for other featured mods.

Where is the faf_path.lua, you are talking about to set the variable ?

1-alex98:
faf_path.lua is a file written by the client

1-alex98:
So I guess init.lua and init_fafdevelop.lua would need to be changed as well

1-alex98:
In faf_path.lua the client writes the path to FA by setting a variable... The same thing could be done with the vault patht... that was the idea behind this

1-alex98:
https://github.com/FAForever/downlords-faf-client/blob/f845162a3db52563a0c30019b1042c6e39f4a4dd/src/main/java/com/faforever/client/patch/GameUpdaterImpl.java#L117

Garanas:
Closed in favor of the new PR that can be tested.

PR #3181, opened 2020-10-11
Excessive energy consumption of fabricators

GodFuper:
Excessive energy consumption of fabricators. When unbuilt fab pause and unpause.
It is easy to notice when there is a lack of energy and the EM mod is enabled with the Throttle energy = On parameter.


speed2CZ:
Adding new files and refactoring them need to be 2 separate commits.

PR #3182, opened 2020-10-14
Excessive energy consumption of fabricators 

GodFuper:
Excessive energy consumption of fabricators. When unbuilt fab pause and unpause.
It is easy to notice when there is a lack of energy and the EM mod is enabled with the Throttle energy = On parameter. 
Recreate for @speed2CZ


speed2CZ:
Whats the difference to have it in OnCreate?

GodFuper:
Example:
`warning: Error running OnScriptBitSet script in Entity urb1104 at 2dc20b08: ...\gamedata\units.scd\units\urb1104\urb1104_script.lua(30): attempt to call method SetTargetSpeed (a nil value)
         stack traceback:
             ...\gamedata\units.scd\units\urb1104\urb1104_script.lua(30): in function OnProductionPaused
             ...gramdata\faforever\gamedata\lua.nx2\lua\sim\unit.lua(487): in function <...gramdata\faforever\gamedata\lua.nx2\lua\sim\unit.lua:462>`

It tries to change the speed of the Rotator when it is not present.
Previously, Rotator was created when the building was ready. So to avoid an error about trying to change the speed, I just created a Rotator in OnCreate.

shalkya:
For some reason e throttle wasn't even working for me with this fix on
Anyway, speed told me it introduce a warning about stopping a rotator that doesn't exist (there was already one warning about trying to pause/unpause unfinished mexx before the fix). This warning won't happen in for player that aren't using EM since you can't access/select unfinished mass fab. For some reason UI can access unfinished mass fab...
My call : This should be fixed in the mod itself, to not access unfinished mass fab.

speed2CZ:
It doesnt introduce it, it fixes the warning.

Garanas:
It appears this PR has been merged in some other fashion, but I'm not sure. Is this PR still relevant?

PR #3184, opened 2020-10-21
increase performance of table.empty and math.round

Divran:
I discovered that many functions in this file are are implemented in a very inefficient way. These two were just the easiest to fix, which is why I took a few seconds of my time to do so.

The new `table.empty` no longer needs to iterate over the whole table to know whether it's empty. It's therefore O(1) as opposed to O(n).
The new `math.round` no longer converts the value to a string and back again. It also supports using negative values to round to the nearest digit on the left side of the decimal point, as opposed to only allowing positive values to round to sub-decimal values.

Quick performance comparison of math.round
```
old round 5,000,000 iterations:	4.74 sec
new round 5,000,000 iterations:	0.628 sec
```

### DIFF
--- a/lua/system/utils.lua
+++ b/lua/system/utils.lua
@@ -405,7 +405,7 @@ end
 
 --- table.empty(t) returns true iff t has no keys/values.
 function table.empty(t)
-    return table.getsize(t) == 0
+    return next(t) == nil
 end
 
 --- table.shuffle(t) returns a shuffled table
@@ -623,12 +623,13 @@ function Sort(itemA, itemB)
 end
 
 -- Rounds a number to specified double precision
-function math.round(num, idp)
+function math.round(num,idp)
     if not idp then
         return math.floor(num+.5)
-    else
-        return tonumber(string.format("%." .. (idp or 0) .. "f", num))
     end
+
+    idp = 10^idp
+    return math.floor(num*idp+.5)/idp
 end
 
 --- Clamps numeric value to specified Min and Max range
--- a/lua/system/utils.lua
+++ b/lua/system/utils.lua
@@ -628,7 +628,7 @@ function math.round(num,idp)
         return math.floor(num+.5)
     end
 
-    idp = 10^idp
+    idp = math.pow(10,idp)
     return math.floor(num*idp+.5)/idp
 end
 


lua/system/utils.lua
```diff
@@ -623,12 +623,13 @@ function Sort(itemA, itemB)
...
-function math.round(num, idp)
+function math.round(num,idp)
     if not idp then
         return math.floor(num+.5)
-    else
-        return tonumber(string.format("%." .. (idp or 0) .. "f", num))
     end
+
+    idp = 10^idp
+    return math.floor(num*idp+.5)/idp
```
HUSSAR-mtrela, line 632: your new math.round() returns numbers with incorrect precision:
```
0.09090909361839 << math.round_NEW(0.1, 1) -- instead of:
0.10000000149012 << math.round_ORG(0.1, 1) 
```
your new math.round() returns numbers with incorrect numbers:
```
0.12500000000000 << math.round_NEW(0.1, 2) -- instead of:
0.10000000149012 << math.round_ORG(0.1, 2) 
```

HUSSAR-mtrela (changes requested):
@Divran the new implementation of math.round() does not work as expected!

lua/system/utils.lua
Divran: I'm unable to reproduce that. It's possible FAF uses a super old and modified version of lua which has this problem. In my tests, performed outside of FAF, the returned values are identical.

Divran: Tests
```lua
function math.round_new(num,idp)
    if not idp then
        return math.floor(num+.5)
    end

    idp = 10^idp
    return math.floor(num*idp+.5)/idp
end
function math.round_old(num, idp)
    if not idp then
        return math.floor(num+.5)
    else
        return tonumber(string.format("%." .. (idp or 0) .. "f", num))
    end
end

for i=1,10 do
	local num = 1/(10^i)
	local idp = i
	print("math.round_new("..num..","..idp..")",string.format("%18.18f",math.round_new(num,idp)))
	print("math.round_old("..num..","..idp..")",string.format("%18.18f",math.round_old(num,idp)))
	print("---")
end
```
```
math.round_new(0.1,1)	0.100000000000000006
math.round_old(0.1,1)	0.100000000000000006
---
math.round_new(0.01,2)	0.010000000000000000
math.round_old(0.01,2)	0.010000000000000000
---
math.round_new(0.001,3)	0.001000000000000000
math.round_old(0.001,3)	0.001000000000000000
---
math.round_new(0.0001,4)	0.000100000000000000
math.round_old(0.0001,4)	0.000100000000000000
---
math.round_new(1e-005,5)	0.000010000000000000
math.round_old(1e-005,5)	0.000010000000000000
---
math.round_new(1e-006,6)	0.000001000000000000
math.round_old(1e-006,6)	0.000001000000000000
---
math.round_new(1e-007,7)	0.000000100000000000
math.round_old(1e-007,7)	0.000000100000000000
---
math.round_new(1e-008,8)	0.000000010000000000
math.round_old(1e-008,8)	0.000000010000000000
---
math.round_new(1e-009,9)	0.000000001000000000
math.round_old(1e-009,9)	0.000000001000000000
---
math.round_new(1e-010,10)	0.000000000100000000
math.round_old(1e-010,10)	0.000000000100000000
---
```

Divran: @HUSSAR-mtrela perhaps FAF's lua has modified the math.pow operator `^` in some strange way. If you have a good way to test this in the FAF implementation of lua already set up and ready to go, could you please try modifying the code I wrote from this `idp = 10^idp` to this `idp = math.pow(10,idp)` and see if that fixes the issue? Thanks

Divran: I would also like to point out that for some strange reason, FAF has multiple implementations of round. All the other ones already look half like the old inefficient code, and half like my replacement here. See links below

<https://github.com/FAForever/fa/blob/develop/lua/ui/notify/notify.lua#L220>
<https://github.com/FAForever/fa/blob/develop/lua/ui/notify/notifyoverlay.lua#L91>
<https://github.com/FAForever/fa/blob/develop/lua/ui/lobby/trueskill.lua#L422>

HUSSAR-mtrela: @Divran FA game is using old version Lua (5.0.1) and I think we cannot upgrade to newer version of Lua because this version is baked in in the engine of the game. Please always check your code by actually running the FAF game instead of running your code outside of the game.  

That being said, I changed your code to use the ```math.pow()``` instead of ^ operator and now the new ```math.round()``` function works as the original version. So can you correct your PR to use ```math.pow()```?
```
 0.10000000149012 <= math.round_ORG(0.1, 2) 
 0.10000000149012 <= math.round_NEW(0.1, 2) 
 0.019999999552965 <= math.round_ORG(0.02, 2) 
 0.019999999552965 <= math.round_NEW(0.02, 2) 
```

It is really good that you found 3 other implementations of ```round()``` and they are used only on UI side so they are not causing noticeable issue because they do not affect the SIM side of the game where they probably cause mayhem with the ^ operator.  Anyway, those 3 implementation of ```round()``` should be removed and the code that used them should be replaced with ```math.round()``` function. Can you make this change in this PR as well? This way, we have this conversation as a reason why they were removed. Also, it would be great if you could check if we are using ^ operator in rest of the code.

HUSSAR-mtrela: I found 4th implementation of the round function and it should also be removed
https://github.com/FAForever/fa/blob/develop/lua/AI/sorianutilities.lua#L1356

Divran: > Can you make this change in this PR as well? This way, we have this conversation as a reason why they were removed. Also, it would be great if you could check if we are using ^ operator in rest of the code.

I'm unsure if I can do this. I'm unfamiliar with the way FAF handles lua. Will all these other files have direct access to `math.round` when it is defined in this single file? I'm wondering if these other locations will throw undefined errors if we're not careful. In other Lua projects I have worked with it would be possible, but since I've never touched FAF before I can't know for sure.

HUSSAR-mtrela: They SIM and UI side of the FA game has access to functions defined in utils.lua file. However, since you have never run FAF game from source of this repo then we can fix the other implementation of round() and ^ operator in some another PR. 
@speed2CZ can you point him to a guide on how to set up running FA game against this repo?

HUSSAR-mtrela (approved):

lua/system/utils.lua
HUSSAR-mtrela: @Divran, I found the guide with info on setting up DEV init file and running the game against your changes:

https://github.com/FAForever/fa#running-the-game-with-your-changes
https://github.com/FAForever/fa#setting-up-a-development-init-file

speed2CZ:
Does not?

HUSSAR-mtrela:
> Does not?
@speed2CZ what are you asking about?

speed2CZ:
Your last comment...
> 
> 
> @Divran the new implementation of math.round() **does not** work as expected!

HUSSAR-mtrela:
@speed2CZ my conversation with Divran was hidden because I marked the request for changes as `resolved` but now it should show up above. Apparently, the `^` operator is not the same as ```math.pow()``` function in Lua (v5.0.1) that FA is using. As you can follow it, he fixed all issues and now `math.round()` works as expected and it runs faster than the original implementaion.

speed2CZ:
fair enough

PR #3187, opened 2020-10-25
Add option to hide resource markers on preview

### LINKED ISSUE #3186: Remove resource markers in map preview in lobby
Currently in the FA lobby all maps have the resource markers and spawn markers populated onto the map. 

It would be nice if there was an option in the scenario file of maps that would indicate whether these markers should be generated.

Sheikah45:
Fixes #3186 

This change adds an option to not spawn the resource and spawn markers in the spawn preview. It introduces a boolean variable hidePreviewMarkers in the scenarioInfo of a map that is used to determine whether or not the preview should include markers.

This allows the determination to be made on a map by map basis.

### DIFF
--- a/lua/ui/controls/resmappreview.lua
+++ b/lua/ui/controls/resmappreview.lua
@@ -214,7 +214,7 @@ ResourceMapPreview = Class(Group) {
             local slot = inSlot
 
             -- Create an ACUButton for each start position.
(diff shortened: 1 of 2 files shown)


speed2CZ (approved):
Good solution

HUSSAR-mtrela (approved):

PR #3190, opened 2020-11-01
UI Scaling fixes and improvements

speed2CZ:
Some stuff that was looked over or forgotten in the first run.


shalkya:
i still get that issue on the option : 
image
could you fix that too ?

shalkya:
went from that for reclaim 
image
to that : 
image
I find the brown one, less easy to spot ?

speed2CZ:
what scale do you use, it works fine for me

shalkya:
i've 125% scaling

PR #3209, opened 2020-11-17
better unit distribution for spread attack and some fixes

Maaach:
Spread attack randomly changes order queue for each unit, which can result in uneven distribution of units between different orders (more units attack one target than another, ex. 10 bombers split between 3 targets can cause 7 bombers to attack first, 2 to attack second and 1 to attack third target)

This changes it so that units are distributed as evenly as possible between their first attack orders (same number of units attacking each target, ex 10 bombers split between 3 targets will have 4 bombers attacking first, 3 second and 3 third target every time), the rest of order queue is randomized as before for each unit.


lua/spreadattack.lua
```diff
@@ -347,26 +347,19 @@ function SpreadAttack()
...
-            local stableTargetNum = math.floor(numOrders / unitCount) + ((math.mod(numOrders, unitCount) >= index) and 1 or 0)
-            for i = 0, stableTargetNum - 1 do
-                -- For if the targets outnumber the units targeting them.
-                local targetBlock = i * unitCount
-                unitOrders[i + beginAction],unitOrders[targetBlock + beginAction + index - 1]
-                        = unitOrders[targetBlock + beginAction + index - 1],unitOrders[i + beginAction]
+            --For each unit use its index to select which order from order queue should be that unit's first order, this evens out units between their first orders
+            local indexorder = index
+            --If there are more units than orders, skip back to start of order queue once last order in order queue is given to a unit and select orders from start of order queue for next unit
+            while (indexorder > endAction)==true do
```
speed2CZ, line 353: `while indexorder > endAction do` does the same thing

lua/spreadattack.lua
```diff
@@ -347,26 +347,19 @@ function SpreadAttack()
...
+                indexorder = indexorder - endAction
             end
-            beginAction = beginAction + stableTargetNum
-
-            -- Randomize the remaining mixable orders.
-            for i = beginAction,endAction do
-                local randomorder = math.random(beginAction,endAction)
+            if indexorder ~= beginAction then
+                unitOrders[indexorder],unitOrders[beginAction] = unitOrders[beginAction],unitOrders[indexorder]
```
speed2CZ, line 357: we write space after ,

speed2CZ (changes requested):

lua/spreadattack.lua
speed2CZ: after every , please

speed2CZ (approved):

Maaach:
Also if attack orders are queued after any number of queued move orders, the units will consider their last move order they have, individually, as position to look for closest attack order from, instead of their current position.

Maaach:
Also forgot to mention, if units have other orders (move) queued before last group of attack orders in queue (which is what spread attack effects), they will use position of _last order they have before first attack order_ for finding closest attack order to be given as first, instead of their current position (which may be completely different at the time), individually. This means if they are queued to move somewhere before attack order queue, including completely different move orders from each other, spread attack will consider the last position they are queued to move to each, to find closest attack order for each to give as first, instead of whatever position they are currently in (which it uses if they dont have other orders queued before attack orders).

shalkya:
This PR introduce a bug : 
```WARNING: Error running lua command: ...pierre-yves\documents\github\fa\lua\spreadattack.lua(322): access to nonexistent global variable "action2"
         stack traceback:
         	[C]: in function `error'
         	...alliance\gamedata\mohodata.scd\lua\system\config.lua(53): in function <...alliance\gamedata\mohodata.scd\lua\system\config.lua:52>
         	...pierre-yves\documents\github\fa\lua\spreadattack.lua(322): in function `SpreadAttack'
         	[string "import("/lua/spreadattack.lua").SpreadAttac..."](1): in main chunk
```
step to reproduce : 
- have 2 groups of bombers, give a move order to one, then select both group and attack 2 targets, then split attack.

Issue is that 1 group of bomber have a move order while the other one, directly has the attack order. Then the current PR can't handle this situation, while it was working before.

Maaach:
> step to reproduce :
> 
> * have 2 groups of bombers, give a move order to one, then select both group and attack 2 targets, then split attack.
> 

I cant reproduce this, for me it doesnt give errors with latest version:
https://imgur.com/a/sBqATsN

shalkya:
giving the spread attack order before the end of the move order : 
image
after the end of the move order : 
image

I think i was able to replicate what you did only once. But those order don't make sense anyway : 
image

KionX:
https://github.com/FAForever/fa/blob/cb3deaf57c97410c151a6a5200ca36f5a591d3c3/lua/spreadattack.lua#L321
I guess the 321 line is missing local

shalkya:
also i just checked the code, you need to replace all those tabulations by spaces

Maaach:
ok I was able to replicate it by waiting until units are about to reach move order before pressing spread attack, seems not to happen if local is added at line 321 like KionX said, also replacing tab with 4 spaces then? (rest of code seems to use that many)

KionX:
Follows wherever you can add local.

Maaach:
bug*, not but, sorry idk how to edit

Garanas:
What is the status of this PR?

Maaach:
> What is the status of this PR?

sorry for late reply, but this was finished with that last commit unless there are any further changes needed

Maaach:
> What is the status of this PR?

@Garanas sorry for late reply, but this was finished with that last commit unless there are any further changes needed

Garanas:
In that case - I'll re-open it and review it.

Garanas:
I've given this a look at the time and it appears there is at least one issue: when you have eight units shift g eight other units, sometimes it ends up attacking 8 -> 7, leaving one alone.

Meanwhile another pull requests got through that already guarantees that there is an even distribution for the first X units, which is already live for a long time and working fine. I am closing this pull request.

Maaach:
@Garanas I fixed that problem now (even fixed disperse move mod with same solution), can't put a commit with fix here it seems sorry idk how github works exactly, also this isn't just about even distribution of units between first orders, what it also does:

- for any selected group of units that have queued attack orders, once you press the hotkey, it distributes the selected units between these orders so that _closest unit to order gets that order as its first_ instead of random order, _while keeping the number of units that each order gets as its first as even as possible_, the rest of attack orders (after first one for each unit) get randomized as before
- any _move orders in queue_ of each unit prior to last group of attack orders in that unit's queue _are not effected at all_, in fact this lets you queue even different move orders and different number of move orders for different units before selecting them all and giving attack orders and then use spread attack _which will only effect the last group of attack orders_, after which you can queue move orders again (also for any of units and any differing amount of move orders) and attack orders after them and press the hotkey again, and it will only effect the last group of attack orders instead of all of them, in a way, move orders act as a barrier separating groups of attack orders where spread attack only effects the last group, so the units themselves can have different queues of move orders (and even attack orders before them as long as move orders are separating them before the last group of attack orders) and it works
- if the unit has move orders queued prior to attack orders when hotkey is pressed, spread attack will consider _position of its last queued move order_ as the unit position for checking _which attack order is closest to it_ so when the unit does get there, it goes to nearest attack order from there first instead of to closest to its current position (which may be completely different at the time)

So to tldr it, it makes the initial (first order) splitting more predictable and sensible (closest instead of random, and evenly distributed instead of random) and possible to use after queued move orders without effecting them and even chaining spread attacks by alternating move order queues with attack order queues (move orders in queue acting as barriers for spread attack). It is more than just even distribution on first orders. I could create a new pull request since this isn't _just_ even distribution like it was originally going to be.

Also what other pull request are you referring to that also fixes unit distribution on first orders? As far as I can tell spread attack behaves the same in latest patch as it did when I started this.

PR #3210, opened 2020-11-17
Add link to the patchnotes to the changelog popup

speed2CZ:
Reaction to: https://forum.faforever.com/topic/537/read-only-sections-on-the-forum/23

Looks like this:
ForgedAlliance_2020-11-17_20-50-06


shalkya:
would have been better to say "balance patchnotes". So people don't start looking for game patchnote.

speed2CZ:
Button size is limited, maybe in a tooltip?

speed2CZ:
Well but the webpage says its for balance, so... that should be enough?

shalkya:
hmm well i guess. Will see we get any feedback from users on that afterward.

speed2CZ:
My guess is that almost no one will click on it :D

PR #3211, opened 2020-11-20
V2 Sorian AI-Update

Azraeel:
Cleaned Up All The Old # in the AI LUA Code for Sorian and did a lot of adjustments and more code cleaning.

Will clean up default AI too in the future.


Azraeel:
whoops did it wrong.

Azraeel:
There we go B)

shalkya:
you need to rebase it on deploy/fafdevelop

KionX:
I think this is a waste of energy. Sorian has poor performance and does not work on all maps. Better to use UvesoAI.

Azraeel:
> 
> 
> I think this is a waste of energy. Sorian has poor performance and does not work on all maps. Better to use UvesoAI.

Well there are multiple AI, AI-Swarm, AI-Uveso, and RNGAI; and yes people do play these but people still play Sorian. I am just making sure Sorian is "Up To Date", am just cleaning the code and fixing things, i am not adding or implementing new stuff at all.

PR #3214, opened 2020-11-21
Fix cybran acu sonar

Tagada14:
This should fix cybran ACU having Sonar even without the Torpedo upgrade. It also should fix the intel rings making it look like Cybran acu would have radar.


speed2CZ (changes requested):
This isnt very nice way of doing it. There's a field in the bp that says smth like ActiveIntel. Remove sonar from it and you can enable/disable it.

Since the code that can enable disable intels ignores active ones.

Tagada14:
I tried moving the Sonar = true, to the intel instead of Active intel and then the game just broke, like all the upgrades for the acu, it spawned as t2 instead of t1 etc.

Tagada14:
New PR: #3215

PR #3215, opened 2020-11-22
Fix Cybran Acu Sonar Bug

Tagada14:
This fixes the bug where cybran ACU has sonar even without the torpedo upgrade.


Tagada14:
I want to also fix the intel rings for Vanilla ACU, currently it shows the ring for Sonar with the range of 60 (SonarRadius = 60).
I tried to set the SonarRadius = 0 and create new value of NewSonarRadius that is used when you upgrade the ACU with the Torpedo upgrade but it caused some bugs.

shalkya:
been testing a bit, i don't understand what second commit is aiming for. My cyb acu never display any sonar ranga/radius overlay (before or after this PR)

shalkya:
image
ok back to work. That's what i got using your PR (here raw cybran acu) : it still does display the sonar range even if ACU didn't get upgraded.

Tagada14:
Yes, it has a sonar range of 26, the same as OMNI so it doesn't matter. All ACU's would spot that sub.

shalkya:
tested

PR #3220, opened 2020-11-26
Create adjacency effects only once

speed2CZ:
`CreateAdjacentEffect` is called twice, since there are always 2 building adjacent to each other. But the check if the adjacency beams are already created doesnt work, since the function CreateAdjacencyBeams is threaded and runs twice at the same time. Since there is no waiting inside it, it's safe to not thread it, which fixes the check for duplicates.

In the end you get only one set of effects per 2 adjacent units instead of two (they overlapped most of the time, so they were just brighter)

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -559,7 +559,7 @@ StructureUnit = Class(Unit) {
                 return
             end
         end
-        self:ForkThread(EffectUtil.CreateAdjacencyBeams, adjacentUnit, self.AdjacencyBeamsBag)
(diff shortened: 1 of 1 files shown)


shalkya:
i thought the middle adjancy effect that is bright would look like the other ones : 

I don't see any changes before and after the change

speed2CZ:
Picture is broken, there isnt much of a visual change as most of the beams were overlapping, but it runs half the code to get the same result. And it runs it as it was intended, create the beams only if they dont exist yet.

shalkya:
there must be other duplicated beams, thus why we get brighter beams.
This PR does fix some duplication, you could get 2 separate beams linking the same buildings.

PR #3226, opened 2020-11-29
Fixed getting game/replay id from /savereplay cmd argument

timocov:
This is the port of fixes https://github.com/FA-mods/SupremeScoreBoard/pull/20.

Previously `/savereplay` cmd arg had a format of `gpgnet://localhost/uid/username.SCFAreplay`, but in https://github.com/FAForever/downlords-faf-client/commit/1ae6c7e936305f0336ecb9c324e9fc0560b843c3#diff-3d47503b30b47f5a259049a2e45e1eb555b502bd1f9cc68c936394a0d8fc6f1fR192 it has been changed to `gpgnet://localIp:localReplayPort/uid/username.SCFAreplay`, what leads a bug when in the `ID` text in score board instead of replayid there is local replay port ("magic number" `20` in `string.find(url, "/", 20)` and `string.sub(url, 20, lastpos-1)` is for length of `gpgnet://localhost/`) ¯\_(ツ)_/¯

This PR fixes this bug.

### DIFF
--- a/lua/ui/uiutil.lua
+++ b/lua/ui/uiutil.lua
@@ -1090,9 +1090,12 @@ function GetReplayId()
     if HasCommandLineArg("/syncreplay") and HasCommandLineArg("/gpgnet") and GetFrontEndData('syncreplayid') ~= nil and GetFrontEndData('syncreplayid') ~= 0 then
         id = GetFrontEndData('syncreplayid')
     elseif HasCommandLineArg("/savereplay") then
+        -- /savereplay format is gpgnet://local_ip:port/replay_id/USERNAME.SCFAreplay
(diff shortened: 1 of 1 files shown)


shalkya:
merging it to see how it works directly in deploy/fafdevelop

shalkya:
alright i'm not completely sure it fixes : 
that's what i'm getting ingame : 
image
this is what i'm getting when reading the local replay : 
image

this behaviour is exactly the same than in current game

timocov:
@shalkya Isn't SSB on your screenshots instead of built-in score board?

shalkya:
Alright i thought it was a fix with SSB too. All good.

timocov:
I meant, if you'd like to test the fix for SSB, I guess you need to apply patch from SSB repo instead of this one.

Frankly I didn't test the fix with built-in score board, I just found a copy-paste from SSB and apply the fix here (I hope it was copy-paste, not sure what's the source actually).

I guess it should fix the following field and show a replay id instead of local port:

image

PR #3231, opened 2020-12-03
Field engineering

Dragun123:
Adds Field Engineer Icons for Aeon/Sera/Cybran so if they capture or receive a field engineer, they have idle icons to indicate that when looking at idle engineers. Icons themselves provided by Balthazar.


lua/factions.lua
```diff
@@ -179,6 +179,7 @@ function OrgFactions()
         IdleEngTextures = {
             T1 = '/icons/units/ual0105_icon.dds',
             T2 = '/icons/units/ual0208_icon.dds',
+            T2F = '/icons/units/sal0209_icon.dds',
```
speed2CZ, line 182: Im pretty there's no unit with ID sal0209

Dragun123: Update to use Sparky Icons

PR #3232, opened 2020-12-06
Fix wait on NukeAI to enable Aeon SML time to complete firing animation.

relent0r:
Currently the Aeon Nuke launcher will never fire for an AI that is using this platoon function. Have corrected for the default AI and the Sorian AI NukeAI platoon function.
The default WaitSeconds time of 10 does not allow the Aeon SML enough time to complete its animation to fire a missile. This results in the default AI issuing a clear commands request and canceling the launch sequence. Increased to 12 seconds to allow time to launch missile. 

To test, spawn an Aeon T3 SML for the AI to pick up in its main base and validate missile is fired when completed.

### DIFF
--- a/lua/platoon.lua
+++ b/lua/platoon.lua
@@ -448,7 +448,7 @@ Platoon = Class(moho.platoon_methods) {
                 nukePos = import('/lua/ai/aibehaviors.lua').GetHighestThreatClusterLocation(aiBrain, unit)
                 if nukePos then
                    IssueNuke({unit}, nukePos)
-                   WaitSeconds(10)
+                   WaitSeconds(12)
                    IssueClearCommands({unit})
                 end
                 WaitSeconds(1)
@@ -3785,7 +3785,7 @@ Platoon = Class(moho.platoon_methods) {
                 nukePos = import('/lua/ai/aibehaviors.lua').GetHighestThreatClusterLocation(aiBrain, unit)
                 if nukePos then
                     IssueNuke({unit}, nukePos)
-                    WaitSeconds(10)
+                    WaitSeconds(12)
                     IssueClearCommands({unit})
                 end
                 WaitSeconds(1)


speed2CZ:
This dosnt look like a safe fix, better to save the command in loval variable and wait until its done, that will survive any changes

relent0r:
I was just making the minimum changes required so that it would be easy to validate and approve :)
I haven't seen anyone else take that approach before as its a sim function rather than platoon so couldn't use IsCommandsActive to validate its completed. I haven't used the sim function IsCommandDone before either which looks like it would be the candidate for that approach.

speed2CZ:
Yeah, IsCommandDone should do the trick here

relent0r:
ok thats cool, I'll test it using that tonight/tomorrow and see if its still happy. I've been running my own AI with the 12 second wait for the last 6 months so was confident about it. I had a quick chat to Sprouto about this as neither him nor Uveso issue a clear commands function to nuke launchers post missile fire command so its unknown why the original code does it this way.

But I'll do some testing and see if I can use the IsCommandDone function, if so I'll put it into testing for a while.
Am I right to assume your expectation would be something like
cmd = IssueNuke({unit}, targetPos) 
then a 
while IsCommandDone(cmd) loop afterwards?

speed2CZ:
yes, thats how it should work
I dont know either why its done like this. You'd have to dig deeper into the nuking code.

relent0r:
I did some testing of this, but it looks like the IssueNuke doesn't create a command object like other other commands when I saved it to a variable and tested it. I also tried emulating how the RolloffBody function uses it with the following.
`                if nukePos then
                    LOG('Set nuke to cmd variable')
                    unit.NukeCommand = IssueNuke({unit}, nukePos)
                    local loop = 0
                    LOG('Waiting for command to be done')
                    while not IsCommandDone(unit.NukeCommand) and aiBrain:PlatoonExists(self) do
                        loop = loop + 1
                        LOG('Waiting for command to be done, current '..loop)
                        WaitTicks(30)
                    end
                    LOG('Firing loop completed, clearing commands')
                    unit.NukeCommand = nil
                    IssueClearCommands({unit})
                end`

But the IsCommandDone is expecting an object which the IssueNuke function isn't providing. So you get the following.

`WARNING: Error running lua script: g:\code\fa\lua\platoon.lua(454): Expected a game object. (Did you call with '.' instead of ':'?)
         stack traceback:
         	[C]: in function `IsCommandDone'
         	g:\code\fa\lua\platoon.lua(454): in function <g:\code\fa\lua\platoon.lua:424>`

This could explain why the nuke function works the way it does (yay gpg).

PR #3234, opened 2020-12-09
loading tips adjustment because of overcharge change

FemtoZetta:
-adjusted the tip about how many energy storages you need to one-shot units to make it accurate again

-fixed a typo in a comment


shalkya:
just so it is write down somewhere, as of now : 
you can fire an OC with 5000 e, which is 1 ACU and 5 field engineers

you need : 
1 e storage to kill all T2 units part from ilshy/obsidian
2 e storage to kill Ilshy/obsidian/loyalist
3 e storage to kill titan
4 e storage to kill othuum/harbi
6 e storage to kill percy/bricks

Issue #3202, opened 2020-11-15
Score data does not record income rate without reclaim rate


HUSSAR-mtrela:
Current Behavior:

1. In replay session, the mass income rate fluctuate wildly in score data because FAF does not exclude reclaim rate from income rate like it did [before](https://github.com/FAForever/fa/commit/dd0a1fae63e86584b7db1442ac9d244e70d60bc1?branch=dd0a1fae63e86584b7db1442ac9d244e70d60bc1&diff=split#diff-b0fdd58b5f454c1e0e9f0bd5f12208ba3ccb2803cf78c94297d259a54e0e8418L230). This is a regression issue because previous behavior was implemented on many requests from FA casters (including Gyle). As result of this issue, it is difficult to compare players eco while watching replays because income stats fluctuate wildly while some players are reclaiming and other do not.

2. Also, now storing reclaim values is happening on [separate thread](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/score.lua#L47) at 1 second interval while rest of eco stats are recorded at [different interval](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/score.lua#L155). As result, it is difficult to even create a UI mod (e.g. SupremeScoreBoard) that could adjust income rate by subtracting reclaim rate.

3. Lastly, the Economy UI (left side) shows small difference between total income (60) and sum of reclaim rate (50) and mass in rate (11) which is 61 and it displayed in score UI (right side).

Expected Behavior:
1. Exclude reclaim rate from income rate in score data OR add a new field in score data that will have reclaim rate excluded from income rate
2. Store reclaim rate and income rate at the same interval 
3. Correct total income in Economy UI by rounding up the sum of reclaim rate (50) and mass income rate (11) 

![image](https://user-images.githubusercontent.com/11665978/99198153-bf063500-2764-11eb-9fc8-e771cb0267dc.png)

Sheikah45:
This can be closed I believe as #3242 was merged

Garanas:
Closed on demand, good spot.

PR #3242, opened 2020-12-30
Remove reclaim rate from score income

Sheikah45:
This removes the reclaim rate from the score income rate as mentioned in #3202 1-2

This has the trade off of measuring all rates as the estimated change over time based off the change in totals rather than using the instantaneous values returned from the GetArmyStats() function. This was done to try to preserve some of the sim performance increases.

The alternative would be to calculate the rates every tick which is ultimately a trade off of resource usage for instantaneous accuracy.

### DIFF
--- a/lua/sim/score.lua
+++ b/lua/sim/score.lua
@@ -44,24 +44,6 @@ function CalculateBrainScore(brain)
     return math.floor(resourceProduction + battleResults + (commanderKills * 5000))
 end
 
-local function ScoreResourcesThread()
-    while not victory.gameOver do
-        WaitSeconds(1)
-        for index, brain in ArmyBrains do
-            if ArmyIsCivilian(index) then continue end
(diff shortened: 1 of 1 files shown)


KionX:
ScoreResourcesThread doesn't have much impact on performance.
Note that the accuracy of the measurement will vary depending on the number of live players and sync frequency.
Update reclaim rate each tick gives the worst accuracy.

Sheikah45:
Moved resources into their own thread to have independent accuracy interval

shalkya:
image
Hello it seems to work well, the only issue i've is that transferred resources count as income (as you can see on the pic). Well i guess it's another issue.

Sheikah45:
Are these resources that are being transferred as overflow or through the explicit give resources. Actually it is probably both

shalkya:
yeah transferred resources and overflow both induce that issue on mass income. It's even worse for total mass, sending it, increase your ally total mass, if he overflow that mass, you inflate your total mass resources. In game you won't have more mass, but it fuck up the statistic.

PR #3243, opened 2020-12-30
Set rebuild progress based on ratio of mass to max possible wreckage mass

### LINKED ISSUE #3095: wreck half built unit
after 50% completion, a building will leave a wreck if you ctrl k it. If it get killed while being built, the unit will automatically rebuild it.

But there is several issues :
1st : if there is no one constructing the building, then when it get kill it won't leave any wreck.
2nd : if a unit rebuild on top of a wreck of an unfinished building, whatever the amount of mass left in the wreck you will rebuild it at 50% (if you didn't reclaim part of the wreck)

Expected behaviour : 
1st : it should spawn a wreck
2nd : the new building, should be at 50% of the completion state of the old building. If the wreck took damage, the new building completion state should be lowered proportionally, the same way it works when you rebuild on a half reclaimed wreck.

Sheikah45:
Fixes #3095 

This sets the rebuild progress from a wreck proportionally to the amount of mass left in the wreck compared to the buildCost*wreckageMult.

This has to retrieve the unit blueprint first to calculate the max wreckage mass for the unit. This could be avoided if overkill damaged the wreck rather than just changing the initial mass value.

Additionally this sets the maximum rebuild progress after ctrl king a building to 45% from 50% due to the difference in intial reclaim left vs the buildCost*wreckageMult.

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2500,8 +2500,8 @@ Unit = Class(moho.unit_methods) {
             for _, p in props do
                 local pos = p.CachePosition
                 if p.IsWreckage and p.AssociatedBP == bpid and upos[1] == pos[1] and upos[3] == pos[3] then
-                    local bp = unit:GetBlueprint();
-                    local MaxMassReclaim = bp.Economy.BuildCostMass * (bp.Wreckage.MassMult or 0);
+                    local bp = unit:GetBlueprint()
+                    local MaxMassReclaim = bp.Economy.BuildCostMass * (bp.Wreckage.MassMult or 0)
                     local progress = (p.ReclaimLeft * p.MaxMassReclaim) / MaxMassReclaim * 0.5
                     -- Set health according to how much is left of the wreck
                     unit:SetHealth(self, unit:GetMaxHealth() * progress)
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2501,10 +2501,12 @@ Unit = Class(moho.unit_methods) {
                 local pos = p.CachePosition
                 if p.IsWreckage and p.AssociatedBP == bpid and upos[1] == pos[1] and upos[3] == pos[3] then
                     local bp = unit:GetBlueprint()
-                    local MaxMassReclaim = bp.Economy.BuildCostMass * (bp.Wreckage.MassMult or 0)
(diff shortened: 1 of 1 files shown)


lua/sim/Unit.lua
```diff
@@ -2500,7 +2500,9 @@ Unit = Class(moho.unit_methods) {
             for _, p in props do
                 local pos = p.CachePosition
                 if p.IsWreckage and p.AssociatedBP == bpid and upos[1] == pos[1] and upos[3] == pos[3] then
-                    local progress = p:GetFractionComplete() * 0.5
+                    local bp = unit:GetBlueprint();
+                    local MaxMassReclaim = bp.Economy.BuildCostMass * (bp.Wreckage.MassMult or 0);
```
shalkya, line 2504: there is no ; in lua

shalkya (changes requested):

lua/sim/Unit.lua
Sheikah45: whoops, habit

lua/sim/Unit.lua
```diff
@@ -2500,7 +2500,9 @@ Unit = Class(moho.unit_methods) {
             for _, p in props do
                 local pos = p.CachePosition
                 if p.IsWreckage and p.AssociatedBP == bpid and upos[1] == pos[1] and upos[3] == pos[3] then
-                    local progress = p:GetFractionComplete() * 0.5
+                    local bp = unit:GetBlueprint()
+                    local MaxMassReclaim = bp.Economy.BuildCostMass * (bp.Wreckage.MassMult or 0)
+                    local progress = (p.ReclaimLeft * p.MaxMassReclaim) / MaxMassReclaim * 0.5
```
Crotalus, line 2505: MaxMassReclaim could be 0 so need to add check for division by zero here,

Sheikah45: Added a check

Sheikah45:
As a note this also would change the rate for structures in water since they have halved mass anyway although this could be accounted for if desired

shalkya:
alright so i tested, this fix the main issue of rebuilding at 50% is fixed. It doesn't fix the issue of requiring to have a unit building the construction while it get killed in order to get the wreck.
As for water building, i would be up to take it with balance team, to not lower the mass if they get killed.

Sheikah45:
I will look into it not leaving a wreck if not being built

PR #3245, opened 2021-01-07
Fix problem with Sorian game result logic.

relent0r:
When SorianAI was migrated to faf an error was made in the if statements related to the game status Result. The original code used `if not v:IsDefeated then`, when it was migrated to faf the following was used.
`if not v.Result == "Defeat" then`
This never returns true.
I have corrected it to 
`if v.Result ~= "Defeat" then`
across the instances within the sorian builder/strategy conditions and one of its platoon function.

p.s not sure why it wanted to recommit my previous change again. maybe because I modified the platoon.lua again.

If you want to validate the above you can simply test it in lua with the following, only the second if will return true.
```
v = {}
v.Result = nil

if not v.Result == "Defeat" then
    print('not result is true')
end

if v.Result ~= "Defeat" then
    print('~= result is true')
end
```

### DIFF
--- a/lua/AI/AIBuilders/SorianStrategyBuilders.lua
+++ b/lua/AI/AIBuilders/SorianStrategyBuilders.lua
@@ -867,9 +867,9 @@ BuilderGroup {
             local enemies = 0
             local allies = 0
             for k,v in ArmyBrains do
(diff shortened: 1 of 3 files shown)


relent0r:
Closing to to another PR being included. Will resubmit with only this change included.

PR #3247, opened 2021-01-08
Replaced getsize & getn to empty

KionX:
Replaced ~300 calls.
Tests show a ~ +1 increase in Sim speed, ~ +2 in a game with AI.
+In the new exe, these functions are optimized.


shalkya:
i can't evaluate/review/test all those changes. It will require real game testing to spot if any issue appear. At least it doesn't crash the game from the start.

Garanas:
How was this tested?

And did you check if it does not interfere with this definition: https://github.com/FAForever/fa/pull/3184

KionX:
Executable use the same one that is shipped with the client
This uses #3184

Garanas:
And how did you test the performance gain?

KionX:
By eye by sim speed. And I don't remember the details anymore.

PR #3248, opened 2021-01-13
Fix Tooltip for Experimental Transporter

Uveso:
Some mods provide Experimental Trasnporter and the ability function only checks for Tech1-3 units.
Experimental Transporters will trigger an error.

Also the description for transportcapacity only has estimated values so we need to add a helper variable.

---

Added Class1Capacity to all transporter blueprints to set the real transportcapacity

Unit ura0104 (Air Transport) Added missing transport capacity variable (Class1Capacity = 10,)
Unit xsa0104 (Air Transport) Added missing transport capacity variable (Class1Capacity = 16,)
Unit xsa0107 (Light Air Transport) Added missing transport capacity variable (Class1Capacity = 8,)
Unit ura0107 (Light Air Transport) Added missing transport capacity variable (Class1Capacity = 6,)
Unit xea0306 (Heavy Air Transport) Added missing transport capacity variable (Class1Capacity = 28,)
Unit uea0104 (Air Transport) Added missing transport capacity variable (Class1Capacity = 14,)
Unit uaa0104 (Air Transport) Added missing transport capacity variable (Class1Capacity = 12,)
Unit uaa0107 (Light Air Transport) Added missing transport capacity variable (Class1Capacity = 6,)
Unit uea0107 (Light Air Transport) Added missing transport capacity variable (Class1Capacity = 6,)
Unit uea0203 (Gunship) Added missing transport capacity variable (Class1Capacity = 1,)

### DIFF
--- a/lua/ui/game/unitviewDetail.lua
+++ b/lua/ui/game/unitviewDetail.lua
@@ -366,16 +366,11 @@ GetAbilityDesc = {
     end,]]
     ability_transport = function(bp)
         local text = LOC('<LOC uvd_Capacity>')
-        if bp.CategoriesHash.TECH1 then
-            return text..'≈6'
-        end
-        if bp.CategoriesHash.TECH2 then
-            return text..'≈12'
-        end
-        if bp.CategoriesHash.TECH3 then
-            return text..'≈28'
-        end
-        return ''
+        return bp.Transport and bp.Transport.Class1Capacity and text..bp.Transport.Class1Capacity
+            or bp.CategoriesHash.TECH1 and text..'≈6'
+            or bp.CategoriesHash.TECH2 and text..'≈12'
+            or bp.CategoriesHash.TECH3 and text..'≈28'
+            or ''
     end,
     ability_airstaging = function(bp)
         return LOCF('<LOC uvd_RepairRate>', bp.Transport.RepairRate)..', '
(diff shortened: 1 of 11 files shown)


lua/ui/game/unitviewDetail.lua
```diff
@@ -375,6 +375,9 @@ GetAbilityDesc = {
         if bp.CategoriesHash.TECH3 then
             return text..'≈28'
         end
+        if bp.CategoriesHash.EXPERIMENTAL then
+            return text..'≈28'
+        end
```
KionX, line 380: I think this should be done as single line
return text..'--'

Uveso: Yes good idea.
I checked some Experimental Transporter and they can somsometimes only transprot 20 units.
And maybe we have units without a techlevel category.

The-Balthazar:
An alternative fix which opens options for better UI information would be to change the function to:
```lua
ability_transport = function(bp)
    local text = LOC('<LOC uvd_Capacity>')
    return bp.Transport and bp.Transport.Class1Capacity and text..bp.Transport.Class1Capacity or bp.CategoriesHash.TECH1 and text..'≈6' or bp.CategoriesHash.TECH2 and text..'≈12' or bp.CategoriesHash.TECH3 and text..'≈28' or ''
end
```
This will allow the vanilla transports to have the values defined in the blueprints, and will also allow mod makers to specify the capacity value for custom units and also maintains the undefined fix for undefined transports.

The estimated values could also be removed if/when/in-anticipation-of all the vanilla transports have(ng) that field defined.

Uveso:
@The-Balthazar 

Thanks for posting this here.
I did some test and its working.
Will change the PR in a second ;)

shalkya:
tested

PR #3249, opened 2021-01-14
spelling fix != and ~= in menucommon.lua

Eugene8571:


shalkya:
what is smart_engi.lua for ? 
`print( 'lol' )` doesn't look like to be a very serious PR ?

Eugene8571:
smart_engi.lua is a mod i working on. Should have switched branches. sorry.

PR #3258, opened 2021-02-28
Add and Update Loading Tooltips

MrRowey:


loc/US/strings_db.lua
```diff
@@ -7731,6 +7731,7 @@ loadingtip_0040="Stealth only hides units from radar, cloak only from vision"
 loadingtip_0041="Veterancy is based on the mass value of the killed unit and how much damage a unit dealt to it"
 loadingtip_0042="To kill with a single overcharge, you need 1 energy storage for T1 units, up to 2 for T2, and up to 6 for T3"
 loadingtip_0043="To see stealthed underwater units you need water vision or omni"
+loadingtip_0044="Overcharge will drain a maximum of 90% of stored energy"
```
FemtoZetta, line 7734: It won't when you only have 5k, then it will drain everything.

loc/US/strings_db.lua
```diff
@@ -7722,7 +7722,7 @@ loadingtip_0031="Patrol and Attack-Move have the same functionality, but Patrol
 loadingtip_0032="Use T1 Mobile Artillery to kill T1 Point Defenses"
 loadingtip_0033="Power Generators adjacent to Static Artilleries will increase their rate of fire"
 loadingtip_0034="Engineers are more efficient for build power than Engineering Stations"
-loadingtip_0035="Overcharge can deal up to 15,000 damage to a unit, requiring around 49k energy capacity"
+loadingtip_0035="Overcharge can deal up to 15,000 damage to a unit, requiring 60k energy "
```
FemtoZetta, line 7725: I think it's nice to keep the info about how much storage you need here, instead of just how much energy it will drain. Because storage is the more important information. I would keep the same wording as before and just change the amount to 67k.

FemtoZetta (changes requested):
You also need to add the changes to the "loadingtips.lua" file.

loc/US/strings_db.lua
MrRowey: Not sure i follow as this is not about storage it about how much energy it cost to deal up to 15k damage. as well as the Patchnote for the cange stats 60k E and would be more confusing with it saying 67k

MrRowey:
#3259 added to loadingtips.lua

shalkya:
i already merged the first part in #3273
i would prefer if you combine the changes in #3259 in here also, to keep it in 1 PR.
I also asked the balance team, if the specific behaviour at 5k e is normal or a bug.

Uveso:
oops
i viewed and commented Update loadingtips.lua #3259 right now. 

Well, yes make 1 PR please :)

PR #3259, opened 2021-03-02
Update loadingtips.lua

MrRowey:
Update LoadingTip_35 and added a LoadingTip_44

### DIFF
--- a/lua/ui/help/loadingtips.lua
+++ b/lua/ui/help/loadingtips.lua
@@ -34,7 +34,7 @@ Tips = {
     "<LOC loadingtip_0032>Use T1 Mobile Artillery to kill T1 Point Defenses",
     "<LOC loadingtip_0033>Power Generators adjacent to Static Artilleries will increase their rate of fire",
     "<LOC loadingtip_0034>Engineers are more efficient for build power than Engineering Stations",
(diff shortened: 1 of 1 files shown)


Uveso:
loadingtip_0035 still displays: "Overcharge can deal up to 15,000 damage to a unit, requiring around 49k energy capacity"
must be also changed here:
https://github.com/FAForever/fa/blob/deploy/fafdevelop/loc/US/strings_db.lua#L7726

Also all other language files are not changed for loadingtip_0035
(If you just change a text then this will never be translated because no one will find this.)

In this case you only need to change the numbers.
If you can't translate something then better remove the translation tag completely from the language file.
Missing tags can be found easily.

loadingtip_0044 has no translation text at all.

Can you fix it or do you need help ?

FemtoZetta:
I believe [this PR](https://github.com/FAForever/fa/pull/3273/files) already fixed the issue from this one here.

Uveso:
@FemtoZetta 

Yes i read another PR after this one here: #3258
Looks like this is only one part of 3 PRs. 

I let this open until its fixed on the other PR.

Garanas:
Can this PR close? I'm not sure whether there is a duplicate PR going on here.

PR #3268, opened 2021-03-21
Improve readability/fix style consistency by adding spaces after titles

Melanol:


speed2CZ:
Those are not titles but localization tags (everything inside <>) so when the code encounters a string with this tag, it will try to use the string of the currently selected language from strings.lua file in LOC folder. If the string isnt in that file it will use the text from this file as a fallback text. Starting right after the > symbol. So there should not be a space after it.

Melanol:
If there should not be a space, how come 95% of the lines have it?

speed2CZ:
Dont know, anyway its gonna use the strings from this file (if you have the game in english) https://github.com/FAForever/fa/blob/deploy/fafdevelop/loc/US/strings_db.lua

KionX:
Spaces are not needed.

shalkya:
It doesn't impact the game, as the other said.
But i guess we can merge it in for "code" readability/consistency.

Uveso:
Stings from this file are changed with the stings from the localisation file.

As example; Unit_Description_0004 is (also for english) translated here:
https://github.com/FAForever/fa/blob/deploy/fafdevelop/loc/US/strings_db.lua#L1429

The text here with space at start will be replaced with the text from US/strings_db.lua line 1429
without the leading space.

So you will never see any text change from this PR, this file here stores only the backup text in case the localisation is missing.

PR #3269, opened 2021-03-21
Fix Aeon ACU description

Melanol:
There is no energy consumption. I guess it was removed at some point.


shalkya:
we are talking here about the chrono dampner.

PR #3271, opened 2021-03-29
Fix Ingame CPU Benchmark

Uveso:
Benchmark is now testing RAM speed, not CPU calculating speed.

Benchmarks: (lower = better)
old= old CPU benchmark
new= new RAM benchmark
             
old: 144 new: 98 - netjaxx ( i7-6700K DDR4 )
old: 156 new: 144 - Uveso ( i7-4790K DDR3 )
old: 184 new: 115 - Moby ( Ryzen 5 DDR4 )
old: 327 new: 325 - Lionheart ( Q8400 DDR2 )
Bench1

Ingame gamespeed multiplayer: (higher = better)
+7 netjaxx ( i7-6700K DDR4 ) CPU is slower than i7-4790 but RAM is DDR4
+6 Moby ( Ryzen 5 DDR4 ) slightly faster than i7-4790K Also because of DDR4
+6 Uveso ( i7-4790K DDR3 )
+3 Lionheart ( Q8400 DDR2 ) Old Intel Core 2 Quad
Bench2

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -5228,20 +5228,15 @@ local BenchTime
 --  CPU Benchmarking Functions
 --------------------------------------------------
 function CPUBenchmark()
+    -- Uveso: Actual we are checking more or less RAM speed not CPU speed
     --This function gives the CPU some busy work to do.
     --CPU score is determined by how quickly the work is completed.
     local totalTime = 0
     local lastTime
     local currTime
     local countTime = 0
-    --Make everything a local variable
-    --This is necessary because we don't want LUA searching through the globals as part of the benchmark
-    local h
-    local i
-    local j
-    local k
-    local l
-    local m
+    local TABLE1 = 'ABC123.-/'
+    local TABLE2 = {}
     for h = 1, 48, 1 do
         -- If the need for the benchmark no longer exists, abort it now.
         if not lobbyComm then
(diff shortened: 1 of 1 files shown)


KionX:
I don't see any proof that speed is limited by memory.
+3 Lionheart ( Q8400 DDR2 ) - slow cpu and slow memory.
+7 netjaxx ( i7-6700K DDR4 ) - fast cpu and fast memory.

Uveso:
Yes, a benchmark with CPU and RAM would be more accurate. (i had both benchmarks running while testing)
If we also bench with CPU then we would have these results:

CPU+RAM: 121 netjaxx ( i7-6700K DDR4 )
CPU+RAM: 149 Moby ( Ryzen 5 DDR4 )
CPU+RAM: 150 Uveso ( i7-4790K DDR3 )
CPU+RAM: 326 Lionheart ( Q8400 DDR2 )

This is reflecting the real ingame situation better than only a RAM test.
But because we had many trouble to benchmark CPUs in idle state i decided to make a benchmark without CPU load.
(Running both benchmarks cost the same time, both marks are only calculating half of the normal loops. So we could add it without any drawbacks - beside the idle problem)

The proof or idea why RAM is the bottleneck is my own testing with different PCs, CPU and RAM clocks.
I also posted some benchmark results inside the old forum with different RAM speeds on the same PC.
The conclusion was that the CPU speed does not affect the SIM speed on fast CPUs like i7 or Ryzen 5.
But RAM speed does. Especially latency (not throughput) is affecting the sim speed.

I have no idea why the Ryzen with DDR4 has the same speed than the 4790K with DDR3 RAM
Its a new PC and i have not played around with it (no time)

Uveso:
I readded the CPU Benchmark, now we have a better result, but it can be manipulated like before.

New results:

CPU+RAM: 123 netjaxx ( i7-6700K DDR4 ) Gamespeed +0,5 compared to i7-4790K
CPU+RAM: 152 Uveso ( i7-4790K DDR3 )
CPU+RAM: 155 Moby ( Ryzen 5 DDR4 ) Gamespeed +-0 compared to i7-4790K
CPU+RAM: 333 Lionheart ( Q8400 DDR2 )  Gamespeed -4 compared to i7-4790K

So the benchmark result reflect the ingame situation very good.

GodFuper:
Maybe need to make the "Host" the ability to request to make a reTest cpu for any player. For example, I created a game and an unknown player came to me. I don't want to ask him, so I personally remotely run this reCPU and wait for the result.

One of the disadvantages is that this can be abused.

shalkya:
tested
ref : https://forums.faforever.com/viewtopic.php?f=45&t=15908#p160934
https://forums.faforever.com/viewtopic.php?f=2&t=18816#p182378

PR #3273, opened 2021-04-09
Updated loading tips to reflect the OC changes correctly

FemtoZetta:
See title, just updating the tips to correct information.


shalkya:
just tested, it cost me 60k e to oc at 15000 damage

FemtoZetta:
Yes, but you should need 67k in your storage to be able to use 60k energy for OC, according to the formula. I guess it could be reworded to make that more clear...

shalkya:
maybe "Overcharge can deal up to 15,000 damage to a unit, costing 60k energy, but it requires to have 67k energy capacity"
tell me if you prefer that, either way i will merge it

FemtoZetta:
grafik
grafik

I found something weird... It only used like 44k energy to deal 15k damage.

FemtoZetta:
I reworded it a bit, now it should be clearer.

PR #3286, opened 2021-05-15
Fix shield ApplyDamage() function for ACU overcharge

Uveso:
Function was checking for EntityId not UnitId.
This caused an error in executing instigator:GetWeaponByLabel()
while checking shields for a overcharge weapon.

This happens when UnitId is false and EntityId is true.
In this case we got overspill damage from a shield unit near the ACU that was impacted
by the splat damage of an ACU overcharge weapon.

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -178,7 +178,8 @@ Shield = Class(moho.shield_methods, Entity) {
     end,
 
     ApplyDamage = function(self, instigator, amount, vector, dmgType, doOverspill)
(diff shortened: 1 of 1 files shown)


Petricpwnz:
Is this related to https://github.com/FAForever/fa/issues/3287? Can reproduce the error but can't reproduce the crash. Need to know if we need a hotfix and if it's related at all.

Uveso:
@Petricpwnz 
This error here causes only a LUA "NIL" error while trying to execute a non existent LUA function.
Its not crashing the game.

PR #3289, opened 2021-05-17
Nomadicons

Dragun123:


shalkya:
At first, I thought this would be not required anymore with the fix of the init upload. But thinking back on it, since we won't have to redownload everything (especially what is unchanged) in the next patch (this is a fix that will come alongside with the init upload). So user will need to download that once, and those icon can become very handy for modders.

KionX:
``
we won't have to redownload everything (especially what is unchanged) in the next patch
``
I assume this applies to featured mods as well. In this case, this PR is not needed.

Dragun123:
If a featured mod has unique strategic icons they are not avaliable in FAF Vanilla game mode.

KionX:
init file don't solve this problem?

shalkya:
@KionX will init fix, allow modders to include dds into their mods now. They weren't able to do that before.
But i guess we can add dds that are commonly used by modders, since as I said, players would need to download them only once. I guess the question is to know how much modders need that.

Uveso:
@Dragun123 

please write a propper comment to your PR next time. Thanks !

Since this problem can be solved with the ini.lua file i am closing this.

PR #3290, opened 2021-05-23
Remove curated maps from lobby

TheKeyBlue:
Curated maps button currently points to the archived forum. This is really dumb.
Talked with biass about it and he suggested to maybe use the new forum link for it:
https://forum.faforever.com/topic/347/m-m-team-top-picks-map-submissions
But he mentioned that it was also already outdated and not been worked on in months.

Replacing the old archived link with another half-dead link seems like a bad idea, especially if you need a game patch to update it.
So in my opinion, we better just remove it, unless somebody can provide a good link for it.
Before
image

After
image


shalkya:
image
biass about that change. So If I understand well we can close this PR ?

Uveso:
button will stay with the new link; see PR #3313

PR #3291, opened 2021-05-23
Fix bugged acu button textoverlay labels when opening large map preview for second time

TheKeyBlue:
When you open the large map preview in the lobby you can right click on the acu buttons to close spots.
When you open it the first time, you can right click and X appears to show that it slot is closed.
When you open it the second time or more, you can still remove X by right clicking it, but no more X will appear if you want to close something.

Weirdly when you hide and reopen the large map preview, the overlays all appear correctly.
But once you edit the correct X aren't shown.

This PR fixes this, so the X will always appear when you right click even on second or third try.

image

### DIFF
--- a/lua/ui/controls/acubutton.lua
+++ b/lua/ui/controls/acubutton.lua
@@ -99,13 +99,16 @@ ACUButton = Class(Group) {
 
         self.indicator = indicator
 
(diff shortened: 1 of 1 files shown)


shalkya:
tested.

PR #3292, opened 2021-05-24
Fast ordering of extractors

KionX:
Just select the builder, hover the cursor over the resource and click.


Crotalus:
Do I understand correctly that every time the mouse cursor is moved  this codeblock is executed:
https://github.com/FAForever/fa/pull/3292/files#diff-c868bcf16513e477142439864f64f550326e64645698ac535cebdd61fc5a1490R181

I'm not sure that's a good idea.

Maybe add some flag that is enabled when engineers are selected, maybe in the same logic that shows / hide the build UI when you only have builders selected. 

This is probably something that should be an UI mod or at least an interface option so people can choose to enable / disable it.

KionX:
This should not have a noticeable impact on the performance of the game.

This feature is as unobtrusive as possible. I know of one very specific case where it can be a little inconvenient.
If you need to put an engineer on deposit.

Uveso:
@KionX ! 
What an awesome enhancement!

I want it!
Can we have an option for it like for the mass storage ?

(OPTIONS_0273="Assist Mex to Build Mass Storages")

KionX:
Is this really necessary? I'd rather not add it.

Uveso:
I am sorry Kionx, i know how annoying Options and localization stuff is.
And i am always for reducing options, not to adding endless more.

But in this case we are changing a main key function (left mouse click) without any modifier
(modifier = a second key like we need to use hotbuid)

Without a modifier you don't have any choice if you want the new build behavior or not.
So an option is necessary in this case.

---

btw, default setting for this feature should be "on" (same as Assist Mex to Build Mass Storages)

KionX:
To convince me to add an option, you need at least one example from practice when a new enhancement interferes.

Uveso:
@KionX 

1. like i already wrote i want the same handling with this enhancement like for the mex cap
(at least a reason for me and a red line how i handle those enhancements in the future)
2. If you use Area command, while drawing the area box you could leftclick a mex by mistake.
3. Playing the map `Big Game hunters TeamPlay (10x10)` if you try to reclaim the big rocks that are over the mex point you can't rightclick on the rock when a mex point is under the cursor (and you can't see it)
4. Last but not least, its an personal favor, because if you don't do it, i have to make the options.

I bet you're already convinced you only enjoy it when I write walls of text, right? :D

KionX:
I have a feeling that you haven't tested this PR.
I'm still against new meaningless option.

The use of Area command has not changed in any way.
You can also mistakenly leftclick on any reclaim.

Situations like on Big Game hunters TeamPlay (10x10) are vanishingly rare and it does not bring significant inconveniences.
Just need another rightclick.

If it's so important to you, then make it yourself. Sorry.

Uveso:
Are you joking ?
Since i have this enhancement i don't play without it.

Well, it looks like you are not against an option you are just hmm lazy ? :D
Perfect, then i will add a option asap.

Uveso:
Added Option, default is "ON"

Some strings need to be translated, but for not this PR is ready to merge.

KionX:
Lazy? Also a factor.
Since I don't like this option, I'm glad that I didn't have to do it)

Uveso:
No KionX, you are not lazy. Just joking. ;)

KionX:
Want to note that it is not necessary to add text to all localizations.
The missing text will be taken from the English localization.

speed2CZ:
But how would the people working on the translations know its not translated, if its not in their file?

Uveso:
in case the localization is missing, then we don't use the english translation, we use the fallbacktext behind the translation tag:
`title = "<LOC OPTIONS_0285>This text will be used if translation is missing",`

Sadly some people have only changed the english translation and forgot to also change the fallback text in the past.
So i set the correct english text to the other translation files to prevent wrong translations.

KionX:
The missing text will be taken from the English localization. My PR added this.
https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/system/Localization.lua#L47
The fallback text is no longer used.

```But how would the people working on the translations know its not translated, if its not in their file?```
In the game. It won't be worse than it is. Or missing texts can be determined by comparing tables in LUA.
It is advisable to make a warning in the developer version about the missing text.

Uveso:
Nice to know, newer saw this change.

shalkya:
I got several point to make on this one.

first i don't see the new option ?
image

2nd, i think it would be preferable to make all mexx T1 (even if the unit can build T2 or T3). But that's not a major concern.

3rd, with a T3 engie (for example) if you position your mouse over a mexx, then press hotkey to build a T1 or T2 mexx, then move your mouse by 1 pixel, it goes back to T3 mexx... Would happen rarely i guess, because you usually press hotkey without being on top of the mexx, but if it happens, it can be quite annoying.

KionX:
``i think it would be preferable to make all mexx T1 (even if the unit can build T2 or T3).``
I will add it to the options.

``mouse over a mexx, then press hotkey to build a T1 or T2 mexx, then move your mouse by 1 pixel, it goes back to T3 mexx...``
I think this can be fixed.

shalkya:
I gotta learn to fetch before testing pr, thus why option wasn't working.
Anyway it's tested and it works well.

PR #3293, opened 2021-05-24
just make unit cap 1500

FtXCommando:
another deeply nuanced ftx addition to the faf repo


Uveso:
I am missing the adaption for the AI to manage 1500 units.

With this PR the AI will build units until it reaches 1500 unitcap and then it will start to destroy the units.

There are some Platoonbuilders that keep track on the unitcap and try to attack before the unitcap is reached.
These builders needs to be patched so that platoons will be formed way earlier than max unit cap to prevent
stucked units inside the AI base. Its also a waste of attack power when units just stay inside the base.

Main problem here is the fact that there is a unitwatch thread. (watching unitcap)
In case the platoonbuilder is not able to create enough platoons to attack, then the unitwatch thread will start
to destroy units to free the unit cap.

The second problem is, that the AI waits for a decent unitcap before it will use some attack platoons.
So you will see that the AI is waiting longer until it starts an attack in midgame.

So you can't just let an AI run into the higher unitcap without loosing (destoying) units on AI side
and without weaken the attack logic.

You can simply test this with an skirmish AI game. The playerlobby already has the posibility to set a unitcap to 1500.
Run a game agains an AI, cover your base with stealth and cloak units (Total Mayham mod has cloak field units), so
the AI can't see any targets, and wait.
After 50-70 minutes the AI will reach the unitcap and start to destroy units standing inside the base.

We had a similar discussion on the forum related to unlimited unitcap:
https://forum.faforever.com/topic/1623/an-unlimited-unit-cap-option

When you patch the AI functions please have also in mind that all functions must be able to handle also a low
unitcap like 250 units. Or the AI will build only mobile units and is no longer able to expand the base.

Beside the functions it will also be needed to check the platoonbuilders for the changed unitcap.
There are 233 evidence of UnitCapCheckLess inside the platoonbilders that need to be balanced for attacking
and 14 builders for emergency attack on maxunitcap need to be fixed.

speed2CZ:
autolobby is for the ladder/matchmaker
Is AI involved in that?

Uveso:
Yes. We had some diskussions in the forum about including AIs in ladder/matchmaker.
Setting the high unitcap will not crash the game, but some AI functions will not work like they should.

I am patching my AI for a higher unit cap right now.
When i have finished this i will take a look at GPG and Sorian AI.
Most work will be to check/fix the 233 platoonbuilders.

ThomasHiatt1:
Makes no sense to delay changing the unit cap on ladder because someone's AI doesn't work with a higher unit cap. Will be years before there are AI on ladder, if ever.

Garanas:
I'll need to look into the exact impact of the change outside of the AI - but I'm all for merging this as part of the patch on Monday if there is no other impact.

BlackYps:
This would impact all queues. I don't know if 1500 unit cap would be a problem for higher player numbers. If it is, we could either not merge this and define the higher unit cap for 1v1 using the new game options functionality or merge this one and define a lower unit cap for specific queues.

Garanas:
That would be a better solution - using https://github.com/FAForever/fa/pull/3385 you can set the unit cap dynamically through the server.

Sheikah45:
Yes with the game options change allowing it to be set dynamically this is now basically obsolete and can be closed really.

Garanas:
I'll close it when it is being set in practice.

Askaholic:
The default should be set to what will be used most frequently. It makes no sense to keep the default as 1000 and then override it with 1500 everywhere. You also have to consider that setting game options from the server does cost some network bandwidth, so it would again be best to set overrides in the game types that are least common. IMO we should merge this as it has an effect for 1v1 but probably makes no difference for larger games because they will either be fullshare (total unit cap will higher by a factor of num_players / 2) or in share until death games, you could run into the exact same problem if all but 2 players die. Although I expect at that point the game would end pretty fast. I don't really see a downside to merging this.

Garanas:
Fine - then I'll merge this change tomorrow morning. I'm fine with it either way, as long as it happens.

Garanas:
With regards to the post of Uveso [on the forum](https://forum.faforever.com/topic/183/increase-ladder-unit-cap/7) I am closing this PR.

We can set the higher unit cap through https://github.com/FAForever/fa/pull/3385 and this may cost slightly more bandwidth, but unless I am mistaken then we are talking about bytes per player per match.

PR #3295, opened 2021-05-27
Props ancient-earth

Garanas:
Adds in the props from Madmax, Thacon and Biass. 

Checklist for Biass:
 - [x] Checked ancient-earth props
 - [x] Checked Thacon props
 - [x] Change texture of the gate

Props to add:
 - [x] crystals

~~Note: do _not_ merge this (yet)!~~
Ready for review!


Garanas:
Note from speed2: you can use the old models by having a relative path to them in the .bp file

Garanas:
This is a map with all the props on it:
 - https://we.tl/t-9HGAjjzjXc

Garanas:
Only thing remaining is the gate being pretty messed up pathing wise (CTRL + ALT + O).

Garanas:
@shalkya I think it is ready for review

shalkya:
Yeah so I looked into it, mostly it's fine. I just have some doubt on the reclaim amount of the props worth 300 reclaim. It's seems a bit high, in comparison to their look.

Garanas:
I agree - according to biass this is what they are in vanilla too. I can't ping him here, but I'll message him

Garanas:
@shalkya it should be all good now! I've checked it with the default props

shalkya:
tested again, approved

PR #3297, opened 2021-05-29
Add dock hotkey with translations

TheKeyBlue:
I couldn't find a hotkey for dock. So I made it.
Uses the translations for tooltipui0425
image


shalkya:
image
are you sure about this translation ? Couldn't we find a word for that order ?

PR #3298, opened 2021-05-30
Fix GetRandomInt & GetRandomOffset2

KionX:
https://github.com/FAForever/faf-coop-maps/issues/365


Uveso (approved):
Functions are working without any issues.

PR #3301, opened 2021-06-05
fix BaseMonitorPlatoonDistress function from creating infinite table …

relent0r:
This is a fix to the default AI function BaseMonitorPlatoonDistress.
It corrects an edge case crash that can happen when platoons are using the 'PlatoonCallForHelpAI' addon plan.

By default if 2 or more platoons are under distress the function will only check if one is under distress and then add the next even if it is already within the table. Should 2 or more platoon remain under distress for a prolonged period this causes the table size to exponentially increase in size until the game eventually starts lagging and crashes due to the table size. In logging I saw the table size go into the tens of thousands within a 3 minute period once the issue started happening.

### DIFF
--- a/lua/aibrain.lua
+++ b/lua/aibrain.lua
@@ -2828,7 +2828,7 @@ AIBrain = Class(moho.aibrain_methods) {
         local found = false
         for k, v in self.BaseMonitor.PlatoonDistressTable do
             -- If already calling for help, don't add another distress call
-            if table.equal(v.Platoon, platoon) then
+            if v.Platoon == platoon then
                 found = true
                 break
             end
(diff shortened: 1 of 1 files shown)


relent0r:
Uveso pointed out that there is another function related to this base monitors distress table that is also broken by default so this can't be validated. I'll fix that up and commit to this fix.

relent0r:
have added the extra fixes and tested to validate no obvious errors.

Uveso (approved):
Tested and working flawless

shalkya:
tested, no crash / no warning. Trusting Uveso on this.

PR #3313, opened 2021-06-22
New link for curated map button

Uveso:
Old link was leading to the old forum.
New link goes: https://forum.faforever.com/topic/347/m-m-team-top-picks-map-submissions


shalkya:
tested

PR #3314, opened 2021-06-28
fix ArmyBrain result not equal to defeat bug in Sorian AI logic

relent0r:
resubmission of previous PR that had some unwanted extra code.

When SorianAI was migrated to faf an error was made in the if statements related to the game status Result. The original code used if not v:IsDefeated then, when it was migrated to faf the following was used.
if not v.Result == "Defeat" then
This never returns true.
I have corrected it to
if v.Result ~= "Defeat" then
across the instances within the sorian builder/strategy conditions and one of its platoon function.

Impacted AI Logic.
Platoon Functions. 
AvoidBasesSorian. Impacts AI's ability to detect if they are avoiding bases correctly.

BuildConditions.
AIOutnumbered. Calculating players vs AI strength.
EnemyToAllyRatioLessOrEqual. Calculating enemies vs allies.
ClosestEnemyLessThan. Allows the sorian rush AI to use the ACU in early game.
EnemyInT3ArtilleryRange. Allows the AI to build T3 arty when it knows the enemy is in range of its mainbase.

Strategy.
SorianSmallMapRush. Don't personally know the impact on these or if the sorian strategy manager is enabled. But it wouldn't trigger without this fix.

If you want to validate the above you can simply test it in lua with the following, only the second if will return true.

```
v = {}
v.Result = nil

if not v.Result == "Defeat" then
    print('not result is true')
end

if v.Result ~= "Defeat" then
    print('~= result is true')
end
```

### DIFF
--- a/lua/AI/AIBuilders/SorianStrategyBuilders.lua
+++ b/lua/AI/AIBuilders/SorianStrategyBuilders.lua
@@ -867,9 +867,9 @@ BuilderGroup {
             local enemies = 0
             local allies = 0
             for k,v in ArmyBrains do
-                if not v.Result == "defeat" and not ArmyIsCivilian(v:GetArmyIndex()) and IsEnemy(v:GetArmyIndex(), aiBrain:GetArmyIndex()) then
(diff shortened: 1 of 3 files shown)


Uveso (approved):
Works perfect.
Thanks a lot.

Issue #3191, opened 2020-11-03
Death weapon being fired for unfinished T4


satoraresatohi:
https://youtu.be/rnP5oKz-Xo4
If you good at hotkeys, you can place 3-4 experementals near a enemy commander land or underwater doesn't matter, then ctrl+k them to cause fast damage, each unit deal 4k dmg, Thats means you can made tele saku and just snipe any unit/build for few seconds w/o spent any resource
(Tested on GC, ML,  Megalith, Ythotha all of them deal wreck damage)

MrRowey:
@Garanas can be closed ? with pr merged

satoraresatohi:
I guess yes, if you already fixed it. Thanks! @Garanas

PR #2891, opened 2019-10-26
fix half built ML/mega wreck falling under the ground

shalkya:
Fixes #2399 
correct the pr #2823

PR #3316, opened 2021-07-04
Death damage fix

Garanas:
Fixes the following bug:
 - https://youtu.be/7JlFxlfGQOc

Currently the damage animations (e.g., the explosions) are still happening. Maybe those shouldn't happen at all when the unit dies under construction? It would simplify the check as a whole: do not run the death thread at all if the unit isn't finished building.

### DIFF
--- a/lua/maui/button.lua
+++ b/lua/maui/button.lua
@@ -81,8 +81,8 @@ Button = Class(Bitmap) {
                 if self.mMouseOver then
                     self:SetTexture(self.mHighlight)
                     self:OnRolloverEvent('exit')
-                    self:Play()
                     self:OnClick(event.Modifiers)
+                    self:Play()
                 end
             end
             dragger.OnCancel = function(dragger)
--- a/units/UAL0401/UAL0401_Script.lua
+++ b/units/UAL0401/UAL0401_Script.lua
@@ -58,16 +58,19 @@ UAL0401 = Class(AWalkingLandUnit) {
             WaitFor(self.DeathAnimManip)
         end
 
-        local bp = self:GetBlueprint()
-        local position = self:GetPosition()
-        local qx, qy, qz, qw = unpack(self:GetOrientation())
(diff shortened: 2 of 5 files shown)


shalkya:
since death animation are triggered at 50% of completion, maybe we should apply the death damage at that point ?
Should also look into the air experimental, to be sure they don't deal death damage when they are killed during construction. (I think they only do crash damage, but this should be checked)

Garanas:
I recall checking them (trying to stand underneath them as much as I could) and was not able to damage to anything. I wasnt aware the death animations were applied at some point, I can use that as a tipping point.

Garanas:
Noticed that the animation is not applied to the Monkeylord and the Megalith. That has to do with https://github.com/FAForever/fa/pull/2891. Should the damage be done in the case of a Monkeylord / Megalith? 

We could also gradually move the unit upwards to solve the problem if it ending up underground.

shalkya:
alright i tested this a bit more. I understand your concern about ML/mega, which differs from how the other experimental works (inconsistency). Although since they are built on the ground, and their animation is triggered at 100% (to get up on their feet), I think it's ok to leave this as it is.

I tested CZAR, it does induce 1k damage upon dying. It should behave like the other air exp, don't you think ? Apart from that, I'll merge the PR.

Garanas:
Can't do fall damage if you're on the ground. I'll look at the CZAR and others later today

Garanas:
This PR also fixes #3191

Garanas:
I'm going to fix the other fall death damage experimentals this weekend and then I'll merge this PR next Monday morning if there are no objections on the merge.

Techmind:
Such a cool bug, should be a feature instead)

Garanas:
Haha, it isn't finished yet though: for the future I'll add the actual death animation to those experimentals that do damage at some point. That should make it look a lot better 😄

PR #3317, opened 2021-07-04
Fix shield warning

### LINKED ISSUE #3315: warning shields
you can get this warning : 
image

way to reproduce, put 2 commanders on auto oc, put some mobile shields to support them, and run them 1 against eachother. (maybe it requires shields from both players to cover eachothers)

Garanas:
Fixes #3315 

Also fixes the following error:
```
WARNING: Error running lua script: ...rs\jip\documents\supreme commander\fa\lua\shield.lua(188): attempt to call method `GetWeaponByLabel' (a nil value)
         stack traceback:
         	...rs\jip\documents\supreme commander\fa\lua\shield.lua(188): in function `ApplyDamage'
         	...jip\documents\supreme commander\fa\lua\overspill.lua(90): in function <...jip\documents\supreme commander\fa\lua\overspill.lua:83>
```
         	
Which is caused by the overspill function passing in the wrong table to the instigator parameter (source instead of instigator). Not sure if this has impact on other areas of the code - in my own testing I couldn't find any issues. I tried shooting with a commander that has enough storage for full overcharge at a few shields, behavior appears to be normal and no errors.

### DIFF
--- a/lua/overspill.lua
+++ b/lua/overspill.lua
@@ -87,7 +87,7 @@ function DoOverspill(source, instigator, amount, dmgType, dmgMod)
                 local targetId = v.EntityId
                 if v:IsUp() and not DidTakeDamageAlready(targetId, instigatorId, amount) then
                     local direction = Util.GetDirectionVector(source.Owner:GetCachePosition(), v.Owner:GetCachePosition())
-                    v:ApplyDamage(source, (amount * dmgMod), direction, dmgType, false)
+                    v:ApplyDamage(instigator, (amount * dmgMod), direction, dmgType, false)
                     RegisterDamage(targetId, instigatorId, amount)
                 end
             end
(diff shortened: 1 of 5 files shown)


Uveso:
@Garanas 

always a pleasure to read something from you!
Thanks for the fix! (will test it soon)

shalkya:
hello I tested this PR, I still got this errors : 
image

I would like also to say that you should checkout deploy/fafdevelop branch, pull, before creating a branch that will be targeted against deploy/fafdevelop. To reduce conflict that might happen, with already merged in PRs.

Garanas:
I'll do the pull in the future.

image

We're going deeper down the rabbit hole with this one. I ended up with an error in defaultprojectiles.lua, just to end up with another error in unit.lua. The last one is of the screenshot attached.

shalkya:
tested, warning disappeared.

PR #3318, opened 2021-07-04
Rotate towards enemy logic redo

Garanas:
The original function first gathers all the units in radius, sorts them and then chooses the unit with the most threat. That is O(n) space and O(n * lg(n)) computation time, where _n_ is the number of units found. Since we're only interested in the unit with the most threat we can store it in O(1) space and run it in O(n) complexity.

It doesn't happen often that this function runs on a large amount of units - but when building artillery pieces it can happily sort all hostile units on the map and that may cause a stutter.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -52,40 +52,70 @@ StructureUnit = Class(Unit) {
     end,
 
     RotateTowardsEnemy = function(self)
+
+        --- constructs a piece of data that we can compare
+        local function MakeTarget(location, distance, threat)
+            local data = { }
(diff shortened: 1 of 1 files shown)


speed2CZ (approved):
Looking good

Garanas:
Refactored the code to prevent table allocations (and de-allocations) for each threat update.

Garanas:
As far as I can see the code is stable and functional. See also:
image

I am merging it on Monday morning.

Garanas:
Tested one last time and all is good.

PR #3319, opened 2021-07-05
Optimize hqs

Garanas:
The original engi mod implementation was inefficient. In its core it looped over every engineer and factory, just to loop over every engineer and factory again for each engineer and factory. Yes - that is a long sentences and it is a lot of work for the computer too. The complexity was O(n^2), where n is the number of engineers + factories of a player. Each engineer / factory was updated individually.

Note that this was called when:
 - An engineer was built somewhere
 - A factory was built somewhere
 - A factory HQ is built or lost somehow

And since it is the meta to make a lot of factories and engineers - this expensive function was called all over the place all the time.

The new implementation is O(n):
 - The brain has a table that keeps track of the HQs per faction, layer and tech. Nomads should be supported - I didn't break that as far as I am aware. 
 - Each HQ factory has a callback when they are built, gifted, reclaimed, captured and killed. These callbacks update the internal state of the brain. This is O(1). As an example: an Aeon Land T3 HQ is built, therefore we subtract one from the Aeon Land T2 HQ table if we're upgraded (and not spawn in via script / cheat menu) and add one to the Aeon Land T3 HQ. Then the overall brain restrictions are updated to allow or prevent support factories for the brain as a whole. This is O(1). Then each unit that should be affected individually (other factories) are updated too. This is O(n), where n is the number of factories.

I've tested the following scenario's:
 - Random creation / destroying / reclaiming / capturing of factories between two armies via the cheat menu
 - Gifting between me and another me in another army (two game instances, multiplayer)
 - Enhancements of the commander and creation / destroying / reclaiming / capturing of factories
 - Random creation of air factories and see how it affects land factories (for engineers)

I've also removed a bug: experimentals that can build T3 units from the get go no longer lose this capability if an HQ is lost. Experimentals are now excluded from the HQ concept- they can always build whatever they can build. I think it would even be interesting to count them as a HQ, but I'll leave that to the balance team to consider. I'm just here to speed things up.


Garanas:
In case you want to test how often the function to search for HQs was called: add a log statement and run a game with 4 AIs in it. The logs get quickly very, very cluttered.

Garanas:
This PR removes the old function (updateBuildRestrictions). This may break mods that assume that function exists. Perhaps we should have a function in there with a warning so that mods can adjust? For compatibility reasons?

shalkya:
Ok I tested, it seems to work well.
several comments here : "experimentals that can build T3 units from the get go no longer lose this capability if an HQ is lost" I have never experienced that, and i tried to replicate, but couldn't.
why would the check of HQ get called when an engie is built ? HQ doesn't have any impact on engies.

What kind of mods would be impacted ?

Garanas:
_Why would the check of HQ get called when an engie is built ? HQ doesn't have any impact on engies._
Engineers can build support factories. Those can come and go as HQs come and go.

_"experimentals that can build T3 units from the get go no longer lose this capability if an HQ is lost" I have never experienced that, and i tried to replicate, but couldn't._
You're right, my implementation caused them to lose it. But it is fixed now, so all is good.

_What kind of mods would be impacted ?_
Any mod that did something with the engy mod directly, like calling the original check for their own engineer type. Or, as an real example, the RNGai hard-hooked some function and re-did some of the logic, including the engy mod check. That would break if we do not add in the dummy function.

speed2CZ:
> 
> 
> This PR removes the old function (updateBuildRestrictions). This may break mods that assume that function exists. Perhaps we should have a function in there with a warning so that mods can adjust? For compatibility reasons?

It was kinda decided long time ago it's not good to rename stuff like this, but if you really need to, you should keep the old function in for at least 6 months and add WARN() there informing to update it to a new function and just call the new function.

Garanas:
I'll make a backup.

Garanas:
I'll merge the pull request on Monday morning. As far as I can see it works as it should be working.

PR #3322, opened 2021-07-10
 Save player stats to replay during the game.

Kazbek:
First prototype to save player statistic to replay file.

### DIFF
--- a/lua/ui/game/score.lua
+++ b/lua/ui/game/score.lua
@@ -50,6 +50,177 @@ function armyGroupHeight()
     return height
 end
 
+local function SetData(saved, target, data, index, codifier)
+	tmp = data[index]
+	if tmp ~= nil and saved[codifier] ~= tmp then
+		saved[codifier] = tmp
+		target[codifier] = tmp
+	end
+end
+
+--stackoverflow says it make it more perfomance
+local next = next 
+
+local function SaveFork()
+	gameTick = GameTick()
+	WaitSeconds(2)
+	svd = {["A"] = -2}
+    unitsInvokeDelay = 0
+
+	while true do
+		tmpCurrentScores = currentScores
+		if tmpCurrentScores == false then
+            -- for work with mods wich destroy OnBeat function (like SSB)
+			tmpCurrentScores = ScoresCache
+		end
+
+		newGameTick = GameTick()
+		if newGameTick ~= gameTick and tmpCurrentScores then
+			gameTick = newGameTick
+			for index, scoreData in tmpCurrentScores do
+				focusArmy = GetFocusArmy()
+				if index == focusArmy then
+					if scoreData["Defeated"] then
+						ta = { ["A"] = index, ["Defeated"] = scoreData["Defeated"] }
+						if scoreData["general"]["score"] then
+							ta["S"] = scoreData["general"]["score"]
+						end
+						SimCallback({ Func = "SaveData", Args = ta})
+						return
+					end
+				
+					--A - army, R - resources, U - units
+					if index ~= svd["A"] then svd = {["A"] = index, ["R"] = {}, ["U"] = {}} end
+					args = {["A"] = index}
+
+					resources = scoreData["resources"]
+					if resources ~= nil then
+						tmpRersources = {}
+						--using single codes since it will be most stored date
+                        --T - total, TR - total reclaimed, RR - reclaim rate, R - rate
+						SetData(svd["R"], tmpRersources, resources["massin"], "total", "MIT")
+						SetData(svd["R"], tmpRersources, resources["massin"], "reclaimed", "MITR")
+						SetData(svd["R"], tmpRersources, resources["massin"], "reclaimRate", "MIRR")
+						SetData(svd["R"], tmpRersources, resources["massin"], "rate", "MIR")
+						
+						SetData(svd["R"], tmpRersources, resources["massout"], "total", "MOT")
+						SetData(svd["R"], tmpRersources, resources["massout"], "rate", "MOR")
+						SetData(svd["R"], tmpRersources, resources["massout"], "excess", "MOE")
+						
+						SetData(svd["R"], tmpRersources, resources["energyin"], "total", "EIT")
+						SetData(svd["R"], tmpRersources, resources["energyin"], "reclaimed", "EITR")
+						SetData(svd["R"], tmpRersources, resources["energyin"], "reclaimRate", "EIRR")
+						SetData(svd["R"], tmpRersources, resources["energyin"], "rate", "EIR")
+						
+						SetData(svd["R"], tmpRersources, resources["energyout"], "total", "EOT")
+						SetData(svd["R"], tmpRersources, resources["energyout"], "rate", "EOR")
+						SetData(svd["R"], tmpRersources, resources["energyout"], "excess", "EOE")
+						
+						SetData(svd["R"], tmpRersources, resources["storage"], "storedEnergy", "SE")
+						SetData(svd["R"], tmpRersources, resources["storage"], "maxEnergy", "ME")
+						SetData(svd["R"], tmpRersources, resources["storage"], "storedMass", "SM")
+						SetData(svd["R"], tmpRersources, resources["storage"], "maxMass", "MM")
+						
+						if next(tmpRersources) ~= nil then
+						   args["R"] = tmpRersources
+						end
+					end
+					
+                    unitsInvokeDelay = unitsInvokeDelay - 1
+					units = scoreData["units"]
+					if unitsInvokeDelay <= 0 and units ~= nil then
+						unitsInvokeDelay = 4
+						tempUnits = {}
+						SetData(svd["U"], tempUnits, units["air"], "lost", "AL")
+						SetData(svd["U"], tempUnits, units["air"], "kills", "AK")
+						SetData(svd["U"], tempUnits, units["air"], "built", "AB")
+						
+						SetData(svd["U"], tempUnits, units["tech3"], "lost",  "3L")
+						SetData(svd["U"], tempUnits, units["tech3"], "kills", "3K")
+						SetData(svd["U"], tempUnits, units["tech3"], "built", "3B")
+						
+						SetData(svd["U"], tempUnits, units["tech2"], "lost",  "2L")
+						SetData(svd["U"], tempUnits, units["tech2"], "kills", "2K")
+						SetData(svd["U"], tempUnits, units["tech2"], "built", "2B")
+						
+						SetData(svd["U"], tempUnits, units["tech1"], "lost",  "1L")
+						SetData(svd["U"], tempUnits, units["tech1"], "kills", "1K")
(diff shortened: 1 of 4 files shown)


Crotalus:
IMHO this information does not belong embedded into the replay command stream but should instead be sent to the server by GpgNetSend. 
(was discussed here in 2016 https://forums.faforever.com/viewtopic.php?f=41&t=11979)

One problem with encoding it into the replay is that the replays overall gets larger but the bigger problem is that you increase the network traffic between the players (scaling with the no. players) with information that has no use for the current game session.

We already send information about the game to the server when a player leaves the game, used by the achievement system. So that function could be extended with also sending the economic stats. (see: https://github.com/FAForever/fa/blob/deploy/fafdevelop/schook/lua/UserSync.lua#L104)

The game stores the eco stats in a score table for all players (https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/score.lua#L89).

(I had a patch that compacted that information by just storing delta-values between two entries but it messed up hotstats and was deemed too complicated (PR: https://github.com/FAForever/fa/pull/3091))

When the server then receives that information it could then be stored on disk next to each replay so you could get it from the api the same way as replays.

Kazbek:
This is not belong to command stream, its a trade between having info online benefits and network bandwidth.

Why I dont want to send data after game to the server? If we will have it in a command stream, we (or I) can make some mobile/other type apps which you can "watch" (at least heat map, commands and (with this update) graphs, power rankings) online without lauching game. I have prototype for live mobile app already (not working since API is closed).

About camera (in mentioned article), if we will have store camera position in command stream we can get player POV in live replay (usual live replay). No other way to make it without write to command stream.

Why not saving data in json replay header? In total there will be a lot data and json header is uncompressed in replay file so it will consume a lot space but in command stream it will be good compressed.

Dealing with bandwith. Im already send only changing data from score,  and not sending blueprints data at all. Units data send only in 1 of 5 calls. Also tried to minimize technical size ovehead by minimizing (for reasonable limit) names of parameters.

P.S. In vault present some eco mods which try to send mass to friendly players (with same mod) every tick (if no mass send needed they also send send 0 mass command). For my opininon it creates much more load to the network.

<details>
  <summary>Parsed command stream when player have this mod</summary>
All this icons represent GiveResourcesToPlayer command.

image

</details>

Uveso:
Checked the PR in multiplayer with 16 AI's.

It has no effect on the traffic between clients. (~1.4 kb/sec.)
But this was expected since this PR does only effect the replay stream.

So far, i found no errors.
(Code formatting could be better on some lines)

I am just not convinced, that we need this without seeing the result how its used.

Kazbek:
@Uveso thanks for testing. This PR only affects when someone play for some army.
`if index == focusArmy then`
This line checking which data to send (every client must send its own data to prevent duplicates).
So at this moment we need to decide how often need to send data. This can be done in this places:
`WaitSeconds(1)` in the end and
`if(tmpScore > 0 and tmpScore ~= svd["S"]) then` for more less important data.
Im expecting may be near 5-7% traffic increase per every player. But it was tested before sending optimizations.

About usage. I show few examples, one for sure was in zulip, it wass mass in storages and income.

lua/ui/game/score.lua
```diff
@@ -50,6 +50,180 @@ function armyGroupHeight()
...
+					end
+				
+					--A - army, R - resources, U - units
+					if index ~= svd["A"] then 
+                        svd = {["A"] = index, ["R"] = {}, ["U"] = {}}
+                    end
+					args = {}
+
+					resources = scoreData["resources"]
+					if resources ~= nil then
```
speed2CZ, line 99: We're using `if resources then` it is the same, as any value other than nil or false will result in true. Also makes it more readable than double negative :)

Kazbek: May be I had some problems in this case. But may be it was related for something else.

Uveso:
yes, i saw the `if index == focusArmy then` change.

focussing different armies on every PC (i run the test on 4 PC in LAN multiplayer) was part of my test.
(when you start a game with 16 AIs as observer you need to focus an AI army or you can't change the game speed)

Would be nice to see/test what the "app" displays when i focus different armies. ;)

Kazbek:
@Uveso give me replay file and I will create you some screenshots. But there just will be no points for army for time when no one is focus it. So as all lines (in faf analytics) it will connect two closest (even with a big gap) points.

lua/ui/game/score.lua
Kazbek: Edit by maintainers is enabled, feel free to edit also stored variable names. Im tried to make them as small as possible without losing some readability but you can change it as you want.

Kazbek:
Charts examples:
1) Stored mass
image-1
2) Total mass
unknown-5

lua/ui/game/score.lua
speed2CZ: It's not much about what I want but more about the standard that is used in this repo :)
Can you also change the tabs to 4 spaces?

Kazbek: Changed if statements and tabs to 4 spaces.

Garanas:
I've discussed this and correct me if I am wrong:
 - This PR doesn't introduce anything that wasn't possible before. In other words: there is no (new) data send from the sim to the UI. There is data send from the UI to the sim to make sure it is available in the replay file somehow.
 - This PR does introduce information to the sim, but that was send to the UI to begin with and is therefore already present.
 - One counter argument for this PR is that it may inflate the replay file. If I recall a maximum increase of 20% after compression is considered reasonable.
 - Another counter argument is that this information doesn't belong in the replay file. It could be send to the server instead - but that would not allow Kazbeks tool to retrieve the information live.

Then my own two coins:
 - I really like the idea of having this information live, but it adds a lot of data to a lot of games and is likely not going to be used for 99.9% of them. Perhaps it would be more interesting if we'd (only) use this for tournaments somehow?
 - I'd like to clean up the code quite a bit for both maintainability and readability. 

It this summary accurate?

Crotalus:
> Kabek wrote:
> Why I dont want to send data after game to the server? If we will have it in a command stream, we (or I) can make some mobile/other type apps which you can "watch" (at least heat map, commands and (with this update) graphs, power rankings) online without lauching game. I have prototype for live mobile app already (not working since API is closed).

Sounds cool!

> Kabek wrote:
> Why not saving data in json replay header? In total there will be a lot data and json header is uncompressed in replay file so it > will consume a lot space but in command stream it will be good compressed.

The idea would be to write down the stats to something like <replayId>-stats.json.gz and not in the regular replay json header. If the new replay format was a container (i.e. tar/zip) we could have included metadata into that

> Kabek wrote:
> P.S. In vault present some eco mods which try to send mass to friendly players (with same mod) every tick (if no mass send needed they also send send 0 mass command). For my opininon it creates much more load to the network.

Yeah the younger me made that mod, I'm actually about to release a new version where I'll pull out the autoshare part. It's useless now that RAS was nerfed + now possible to quickly share in UI. Sounds weird that it sends 0 mass though, maybe a rounding error.

---

> Garanas wrote:
> I've discussed this and correct me if I am wrong:
> 
> * This PR does introduce information to the sim, but that was send to the UI to begin with and is therefore already present.
> * One counter argument for this PR is that it may inflate the replay file. If I recall a maximum increase of 20% after compression is considered reasonable.

(20% increase of replay file also means ~20% increase of network traffic to keep the sim in sync, assuming both compress about the same). 

I would also imagine that after this been merged someone comes up with a brilliant idea to also show count units / type of units + zillion other stats (camera POV?) they want to be able to extract directly from the replay file and suddenly we have 100%+ increase.

> Garanas wrote:
> * Another counter argument is that this information doesn't belong in the replay file. It could be send to the server instead - but that would not allow Kazbeks tool to retrieve the information live.

Not sure what live information live means here but it would be possible to download the game stats json file with a regular api call assuming the game has ended. A way to get live data without encoding it into replay is to GpgNetSend calls in game and then the server publish them to a RabbitMQ server for anyone wanting to listen to game stats data. Would be less impact on network

Another compromise could be to add the stats in a final command at the end of the game - at the same time as we send achievements to the server, that way we wouldn't get any network increase during the game. Could be that command payload has a max size though.

An idea could be to make this a sim mod as a start to be used in 16 player games, tournaments etc and see how it impacts stuff, merging it into the game without any serious testing sounds a bit risky]**

Garanas:
_An idea could be to make this a sim mod as a start to be tested in 8+ player games, tournaments etc and see how it impacts stuff, merging it into the game without any serious testing sounds a bit risky_

We just did this during a voice chat. Looking forward to the results.

Sheikah45:
I think the concept of being able to get the stats live is a good one as it is something that some people have expressed interest in having and could provide some useful insights.

However I do have reservations about the increased traffic and replay size. As even if we limit it to 20% that is fairly large and as Crotalus mentioned this number can't be allowed to grow ad infinitum when the next person wants to add data and we add another 20%. So if we decide to add it to the replay stream it should be the final form or at least have an absolute upper limit but that is hard to set as replay sizes already vary so much.

And have to just consider that we are adding this extra cost for statistics that I am not sure how many people will actively use. But will be a cost on everyone's games. So it is a bit of a hard choice.

All that being said it definitely deserves some pretty extensive testing to ensure that it really does stay below 20% even in the 8v8 hour long games and what it's impact is in the wild per say.

I do think the idea of using the gpgsendnet command and having either a separate stream or service collect and serve the data is one that is worthy of some investigation as that would allow it to be decoupled from the actual gameplay in a similar way that recording of the replay in the first place is now. Granted that will take more integration and wouldn't allow for any mods that use this info while viewing the replay but might be a cleaner way to do it.

I don't know if the only reason to do this being so a mobile app can consume it is a good enough reason on its own. But do understand there could be other uses as well.

Ultimately if we do this right and set a hard limit of saying this is all that will be added and don't allow more creep in the future then this could be okay. But does just need to be thoroughly tested.

Kazbek:
So mod has been created by Jip and uploaded by me. Now we need some volounters.
Mods name: Replay stats #11350

Garanas:
I'm going to use this mod with some seton matches. Will report back with replay IDs.

Avilzron:
That a pretty good idea :O

Avilzron:
That a pretty good idea :O

Kazbek:
3 months celebration party 🎉🎉🎉

Kazbek:
12 Players
https://replay.faforever.com/15627917

Kazbek:
https://replay.faforever.com/15673744

Garanas:
I am closing this pull request. After doing more and more performance analysis there is one rule that sticks out: be careful with how you allocate / de-allocate memory. The amount of additional memory management this pull request introduces is not something we have spare of. The feature we get in return is not worth the potential cost to performance - and the latter has my priority at this point.

Kazbek:
We cant delete all slicing logic and just put stats from sim.

Garanas:
It is about the syncing to the sim that worries me. It introduces entire tables (that can be quite large) for the mere purpose of storing it in the replay, instead of actually being used. In turn, this PR introduces a ton of allocations for the 'sole purpose' of de-allocating them again.

PR #3323, opened 2021-07-11
Use FAF lua instead of lua5.2

Askaholic:
Verify syntax using a custom modification of lua. Eventually I want to be able to run actual unit tests this way.

Also fixes a bug where files with spaces in the name would not be checked.

FAF lua code is here: https://github.com/FAForever/lua-lang

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -1,5 +1,6 @@
 name: Build
-on: [push]
+
+on: [push, pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


Garanas:
I've looked at the PR and all appears to be good. I've asked one question about why the testing phase is so short (3s) to make sure it checks what it is supposed to check.

Askaholic:
Hmm. It takes about 12s to run on my local machine so that does seem a little weird.

Askaholic:
Well its checking the same number of files so... seems ok.

PR #3324, opened 2021-07-13
Category miss match

Garanas:
A few units from FA that has strange categories. I suspect from features that they didn't end up having.


shalkya:
in the case of the battlecruiser, we can then remove the battleship tag i guess, no ?

speed2CZ:
I would be careful with removing it, as the AI uses it for sure, some of the missions as well, formation code surely uses it to place it correctly and maybe some more things.

speed2CZ:
Same goes for removing it from the missile ship. Adding a new tag that is not used anywhere is not useless, but removing current ones will break things @Garanas ^^

Garanas:
> Same goes for removing it from the missile ship. Adding a new tag that is not used anywhere is not useless, but removing current ones will break things @Garanas ^^

I've been asked by an AI dev to make it more distinguishable.

Garanas:
> in the case of the battlecruiser, we can then remove the battleship tag i guess, no ?

I didn't do that for the reason Speed2 writes out - for example selection mods will assume that it is a battleship of some kind.

Azraeel:
Categories are extremely valuable. These are very helpful changes ^^

Uveso:
NUKESUB in unit XSS0304 is used in formations:

`local SubNaval = [...] + (categories.NUKESUB * categories.ANTINAVY - categories.NUKE)`

So NUKESUB without NUKE inside the blueprint has a meaning here.

Garanas:
I've reverted the changes to the missile ship and battlecruiser there are bigger fish to catch then fixing the formations. I did fix the formation for the t3 seraphim non-nuke sub.

units/XES0307/XES0307_unit.bp
```diff
@@ -57,6 +57,7 @@ UnitBlueprint {
         'DIRECTFIRE',
         'ANTIMISSILE',
         'VISIBLETORECON',
+        'T3SUBMARINE',
```
speed2CZ, line 60: XES0307 is battlecruiser, isnt it?

units/XSS0304/XSS0304_unit.bp
```diff
@@ -80,7 +80,6 @@ UnitBlueprint {
         'SUBMERSIBLE',
         'VISIBLETORECON',
         'RECLAIMABLE',
-        'NUKESUB',
```
speed2CZ, line 83: Maybe add T3SUBMARINE here?

lua/formations.lua
```diff
@@ -441,7 +441,7 @@ local GrowthChevronBlock = {
 
 local LightAttackNaval = categories.LIGHTBOAT
 local FrigateNaval = categories.FRIGATE
-local SubNaval = categories.T1SUBMARINE + categories.T2SUBMARINE + (categories.NUKESUB * categories.ANTINAVY - categories.NUKE)
+local SubNaval = categories.T1SUBMARINE + categories.T2SUBMARINE + categories.T3SUBMARINE +(categories.NUKESUB * categories.ANTINAVY - categories.NUKE)
```
speed2CZ, line 444: This looks like it should be all subs - nuke subs right? since on line 449 there's NukeSubNaval
If there is new T3SUBMARINE category, then you should be able to simply substract NUKESUB from it.

speed2CZ (changes requested):

lua/formations.lua
Dragun123: If we are changing formations.lua for the land code where the deceiver is listed can we add/change to stealthfield instead?

Garanas:
Closing this PR - there are discussions among the AI developers to do a better category overhaul.

PR #3325, opened 2021-07-18
Fixed Failed To Build in EngineerFailedToBuild Callbacks

Azraeel:
Within the Engine Code, this callback is completely busted there is no way to fix this.
Instead Duncan decided in his wisdom to change this... to where the build would not be repeated, which 99% of the time fixed the issue.
When True we would not attempt to repeat the job, When False would attempt to repeat the job again in case it had accidentally failed.
This as stated fixes the Engine Issue.

I logged all processes and tested it.
ProcessbuildCommand Function
EngineerBuildAI Function
EngineerSetupCallbacks
WatchforNotBuilding

Attempted to fix with several fixes (help from Sprouto) none fixed said issue. Concluding that this relates back too callbacks and when logging callback Failed to Build, you will see it fail to build.
When changing said callback to false for it to try/attempt build again. 2nd Attempt works. 
Since how callbacks works are from engine we can not fix this and this is just a work around in reality but it does fix the issue in all my test when i was working on my AI Swarm.

### DIFF
--- a/lua/platoon.lua
+++ b/lua/platoon.lua
@@ -3513,17 +3513,19 @@ Platoon = Class(moho.platoon_methods) {
     EngineerFailedToBuild = function(unit, params)
         if not unit.PlatoonHandle then return end
         if not unit.PlatoonHandle.PlanName == 'EngineerBuildAI' then return end
+        if unit.UnitBeingBuiltBehavior then
+            if unit.ProcessBuild then
+                KillThread(unit.ProcessBuild)
+                unit.ProcessBuild = nil
+            end
+            return
+        end
         if unit.ProcessBuildDone and unit.ProcessBuild then
             KillThread(unit.ProcessBuild)
             unit.ProcessBuild = nil
         end
         if not unit.ProcessBuild then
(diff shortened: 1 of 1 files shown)


Uveso:
ProcessBuildCommand has to be called with true only if the build was finished.
All other cases need false so we can resume or try again.

Duncan changed this to fix the issue with the commanders behavior function.

Root cause is the CDROverCharge function. It stops the ACU platoon every second and try to resume the build with a repair command.
If we call the ProcessBuildCommand before the CDROverCharge had time to resume the build we will build a new building and don't resume the old one.

I will added a fix for the CDROverCharge function. 
In case the CDROverCharge function is running and we are building someting, then UnitBeingBuiltBehavior will be true.
Then we don't need to fork the ProcessBuildCommand because the CommanderBehaviorThread will do it for us.

Azraeel:
Tested and Ready to Go :)

Garanas:
@Uveso You mention you still wanted to make changes to CDROvercharge, are those for this PR? And are they critical before the PR is merged? If not, the morning after your message.

Garanas:
Discussed it with @Azraeel and merging it.

PR #3327, opened 2021-07-26
Reduce the Mole's Dummy weapon and GuardScanRadius range

Tagada14:
Reduce the range of the GuardScanRadius as well as the Dummy Gun

### DIFF
--- a/units/URL0101/URL0101_unit.bp
+++ b/units/URL0101/URL0101_unit.bp
@@ -1,8 +1,5 @@
 UnitBlueprint {
     AI = {
-        GuardScanRadius = 19, --MaxRadius of the dummy gun instead of 
-        -- the radar range because no point to aggro on units outside the
-        -- shooting range as the gun isn't real
     },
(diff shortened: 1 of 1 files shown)


Uveso:
```
MaxRadius = 19, -- Tweaked so it stops a bit behind mantis using GuardScanRadius
```
Thats not true, the unit stops because of the maxRadius not GuardScanRadius.
It will only stop on GuardScanRadius when its lower than the weapon MaxRadius.

```
GuardScanRadius = 19,
    -- MaxRadius of the dummy gun instead of 
    -- the radar range because no point to aggro on units outside the
    -- shooting range as the gun isn't real
```
GuardScanRadius and GuardReturnRadius (that is missing here) is used to move a unit to the target
until it reaches the max weapon range.
Since this has no function for a unit with a 0 damage weapon and this value will be generated automatically
it can be removed from the blueprint entirely.

Tagada14:
Merged with another PR

PR #3328, opened 2021-08-02
T1 Land Scout Rework

Tagada14:
Remove the GuardScanRadius to prevent the scouts from agro'ing onto units and buildings from very far away, now they will trigger based on their weapon's maxRadius which is still much larger then their tank counterparts which should prevent them from suiciding too often when attack moved while removing the great annoyance of your scouts getting stuck if they are attack moved or assisted.

### DIFF
--- a/units/UAL0101/UAL0101_unit.bp
+++ b/units/UAL0101/UAL0101_unit.bp
@@ -1,6 +1,4 @@
 UnitBlueprint {
-    AI = {
-    },
     Audio = {
         AmbientMove = Sound {
             Bank = 'UAL',
--- a/units/UAL0101/UAL0101_unit.bp
+++ b/units/UAL0101/UAL0101_unit.bp
@@ -1,6 +1,5 @@
 UnitBlueprint {
     AI = {
-        GuardScanRadius = 50, --RadarRadius
     },
     Audio = {
         AmbientMove = Sound {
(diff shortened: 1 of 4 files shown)


units/UAL0101/UAL0101_unit.bp
```diff
@@ -1,6 +1,5 @@
 UnitBlueprint {
     AI = {
```
speed2CZ, line 2: You should remove the whole AI table if it has no entry

speed2CZ (changes requested):

speed2CZ (approved):

Tagada14:
Needs more testing and adjustment to find sweet balance between scouts suiciding and agro'ing on random stuff when they shouldn't.

Tagada14:
UEF and Cybran :
Weapon range -> 26
GuardScanRadius -> 26
They shouldn't suicide into T1 tanks that are not Aurora
Aeon :
Weapon Range ->33
Guard Scan Radius -> 33
The Spirit shouldn't suicide into any T1 tanks including Aurora (Sometimes it depends on the refresh rate of the radar but I didn't want to increase it's weapon range too much)
Selen :
Adjusted the GuardScanRadius -> 26 same as UEF & Cybran for similar behaviour.
Selen will suicide into all T1 tanks because it's weapon range is only 18 (you need 25 at least) and because of gameplay reasons it can't be adjusted.
Doubled Aeon, UEF, Cybran Break Speed so they stop a bit faster and you can shave of 1 range from their weapon + Guard Scan Radius for better behaviour.

Issue #3326, opened 2021-07-20
Riptides unable to hit aeon/cybran t1/t2 sonar


Farmsletje:
https://user-images.githubusercontent.com/26172055/126376756-49e3887d-b871-49a9-9e48-382dcec54918.mp4

See video. The same happens with aeon t2 & cybran t1/t2 sonars.
Sounds like a similar issue as https://github.com/FAForever/fa/issues/new

Garanas:
I've made changes to prevent this from happening. See also the mention. Thank you for reporting the issue!

PR #3331, opened 2021-08-07
Changed collision size of sonars to make them easier to hit

Garanas:
Especially relevant when firing with low-arc units such as the riptide and the aurora. Fixes issue #3326. Tested on all sonars, also tested submarines being able to hit them or not. All should be good.


Garanas:
This change was approved by Tagada through Discord. I am merging it on Monday morning.

PR #3332, opened 2021-08-08
Optimize Trashbag

Garanas:
It is now optimized according to the loops benchmark - instead of using table.getn we keep a separate counter. The trashbag is not
used often (primarily during initialization) but it does have a significant difference overall. There is commented blocks of code that you can uncomment to trace where invalid calls to the class are made - makes them easier to track down and clean up.

### DIFF
--- a/lua/system/trashbag.lua
+++ b/lua/system/trashbag.lua
@@ -1,0 +1,39 @@
+local getn = table.getn
+
+#
+# TrashBag is a class to help manage objects that need destruction. You add objects to it with Add().
+# When TrashBag:Destroy() is called, it calls Destroy() in turn on all its contained objects.
+#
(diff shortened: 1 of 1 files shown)


KionX:
Count contains the index of the next element instead of count.
If you swap lines 52 and 53 in places, then Count will be correct.

Garanas:
@KionX That is on purpose. See also these benchmark results:

image

The first version (using getn) is how the original approach was. The bag wouldn't have 100 elements at any point, but it is still significantly slower. The reason the 2nd approach is faster is because the CPU doesn't have to wait on the result of the addition in order to perform the table operation. Inside the CPU there is a (superscalar) pipeline that is quite long - this works best when code is independent of each other, as dependencies (such as reversing the two) can cause the pipeline to stall.

Garanas:
Tested it a few times with AIs again and all seems good.

PR #3333, opened 2021-08-08
Benchmarks

Garanas:
Adds the benchmarks to the repository. The profiler (soon released too) can interact with them, allowing them to be re-run for example. You can manually make a call to /lua/benchmarks/controller.lua -> RunAllBenchmarks(true) to log all the information to the moholog.

Todo before merged:
 - [x] Add timing information at the top of each value
 - [x] Add descriptions / explanations as to why there is no / a difference


Garanas:
This PR with some of the original benchmarks is coming to a conclusion. My question to everyone reading this is: are there more benchmarks that may be interesting to compare / test ?

Garanas:
Merging this for the sake of making references easier. There will be more benchmarks in the future including an improved system to allow to measure variance.

PR #3306, opened 2021-06-13
Displaying teams in the loading

GodFuper:
Without further ado, just a picture
<a href="https://ibb.co/7gkc2N0">image</a>

Optionally, can make display the overall rating, balance percentage, and so on.
Ideally, so that someone would add localization.

PR #3334, opened 2021-08-09
Prefetch loading in lobby to lower time spent in loading screen

Garanas:
Allows prefetching of all assets (including mod blueprints) up to the point where 'BeginSession' is called. No mod or map code is actually being run during prefetching. Lowers the loading time significantly, especially for slower computers.

Inspired by LOUD.

Tested with and passed with flying colors:
 - AI-Uveso
 - All Blackops mods
 - Dilli Dalli AI
 - SSB

To test it, make sure to have the Moholog open while in the lobby. You can do so by having /showlog as a program argument. You can see that when you change mods / maps, the loading already happens. Where as previously, that only happened during loading.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -2148,6 +2148,11 @@ local function UpdateGame()
             GUI.mapView:SetScenario(scenarioInfo)
             ShowMapPositions(GUI.mapView, scenarioInfo)
             ConfigureMapListeners(GUI.mapView, scenarioInfo)
+
+            -- prefetch the session to make loading screen shorter
+            local mods = Mods.GetGameMods(gameInfo.GameMods)
+            PrefetchSession(scenarioInfo.map, mods, true)
+
         else
             AlertHostMapMissing()
             GUI.mapView:Clear()


Garanas:
I noticed that it may make #3306 obsolete as the loading screen is barely shown. The only situation when it is shown is when there is a lot of code being run in BeginSession(), for example by the marker generator of Uveso.

Crotalus:
Make sure total memory usage of the process doesn't increase with the prefetching. One of FA's big problems is the memory usage, it being 32-bit etc. 

I played around with prefetch in 2014, some discussion here:
https://github.com/FAForever/fa/pull/146

The idea I had back then is to let the client make a disk scan in the FA directories at startup, that way the caching of files ends up in memory outside of the FA process

Garanas:
LOUD is notorious with regards to RAM - if this did cost RAM they would not have it. I will do a few tests tomorrow to be sure, but I'm not expecting any increase in RAM. 

I've looked at your PR but in the changes all I can see are '#' -> '--' comment changes. Where exactly is the code that would perform the disk scan, or is that part of the client repository?

Crotalus:
Hmm true, seems like I forced pushed away that change and then the rest was merged.

Anyway, that prefetch attempt seemed to have been including game textures so that would explain the memory increase I saw. 

In 2014 many people still were using HDD so every random disk access was really a toll on the game. A slow disk would lag the game for all, especially at start when stuff isn't read into OS filecache yet.

Garanas:
I've looked into it.

```
151mb -> 521mb (no prefetching)
469mb -> 522mb (prefetching, starting immediately)
700mb -> 608mb (prefetching, after scrolling)
700mb -> 670mb (prefetching, immediate rematch during same process)
488mb -> 524mb (prefetching, after scrolling)
490mb -> 530mb  (prefetching, after scrolling)
```

The results are quite ... diverse. 

In general, if you have the right map right away then the amount of MBs is not that much different. If you scroll through the maps (and pick them) then sometimes the MBs are reset, other times they're not reset completely. I'm not sure why they would or would not get reset. I suspect the game holds on to the resources, expecting them to get used and may release them later on.

Garanas:
We're going to have to make a choice here whether we think these changes are better or not. I'd like to hear from other contributors what their opinion is - is it worth having essentially no loading time over the fact that the game may keep a few more resources in memory than it has to?

Crotalus:
IMHO give it a shot and see what happens, it's easy to reverse if it causes any trouble.

PR #3337, opened 2021-08-09
Remove rocking behavior calls

Garanas:
All calls to starting / stopping the rocking behavior are removed. If you're wondering what it is - it wasn't used in practice. It requires blueprint values set that aren't set by any blueprint, both of the base game files (including those that are not part of the repository) and common mods. 

Removing the call on its own is good - but preventing _a lot_ of threads from being made is better. To give an idea, each time StartRocking is called a thread is made. Each time StopRocking is called, if a thread was made, a new thread is made to stop the previous one. In a game of 4 AIx fighting these functions were called 2030 times to start rocking and introduce a thread and 2580 times to kill the start thread and start a new thread to stop rocking. This was over a timespan of 13 minutes, which means (2030 + 2580) / 780 = 5.9 threads a second. If you take into account that almost no thread spawns the first 5 minutes because there are no units on the water then we get (2030 + 2580) / 480 = 9.6 threads a second. That is almost one thread a tick, for just four AIs in a game that is 13 minutes long.

And all of that - for nothing!

There is one mod that calls the rocking behaviors: Jaggeds_Infrastructure_Pack. I suspect it is because the original files (where the author looked at) made similar calls. There are no mods that define the right values in their blueprint files, hence there would be no actual rocking even though a call was made.

I've looked at the following mods, highlighted some that I think are important:
#2x Resources,Storage,BuildRate,BuildRange
**#Marlos mods compilation**
additionalCameraStuff
Advanced target priorities
AdvancedShieldsV4Fixed
AI-Swarm
AI-Uveso
AllFactionsFixed
AllFactions_FAF_BO_Nomads
All_Faction_Quantum_Gate
Big boom
**BlackOpsFAF-ACUs
BlackOpsFAF-EXUnits
BlackOpsFAF-Unleashed
BrewLAN
BrewLAN_Units**
CinematicsMod
Commanders_x4 BlackOps
Compat_GilbotCore+BlackOpsACU
Cybran UnBalanced
dear-windowing
DilliDalli
Hard FFA (ACUS spawn in T2 transport in center)
Improved Close-up Camera
Jaggeds_Infrastructure_Pack
kazbek-replay-ui
Low Commander Upgrades Costs Tweaked
Low Commander Upgrades Costs Tweaked 2
Music_PhilCrofts
No Friendly Fire
NuclearRepulsorShields
Oof
Personal Phantomz Blackops Balance Patch
PxModAlphaV6
rks_explosions
RNGAI
SACUAI
simspeed++
Sorian Edit
StorageRich_x10Fixed
SupremeScoreBoard
T1 UNIT HEALTH X2
Unholy Alliance - BlackOps ACUs
x2 Build Range

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -381,11 +381,10 @@ StructureUnit = Class(Unit) {
             Unit.OnFailedToBuild(self)
             self:EnableDefaultToggleCaps()
 
-            if self.AnimatorUpgradeManip then self.AnimatorUpgradeManip:Destroy() end
-
-            if self:GetCurrentLayer() == 'Water' then
-                self:StartRocking()
+            if self.AnimatorUpgradeManip then 
+                self.AnimatorUpgradeManip:Destroy() 
             end
+            
             self:PlayUnitSound('UpgradeFailed')
             self:PlayActiveAnimation()
             self:CreateTarmac(true, true, true, self.TarmacBag.Orientation, self.TarmacBag.CurrentBP)
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -4227,39 +4227,49 @@ Unit = Class(moho.unit_methods) {
     --- Allows the unit to rock from side to side. Useful when the unit is on water. Is not used
     -- in practice, nor by this repository or by any of the commonly played mod packs.
     StartRocking = function(self)
-        KillThread(self.StopRockThread)
-        self.StartRockThread = self:ForkThread(self.RockingThread)
+        local bp = self:GetBlueprint().Display
+        local speed = bp.MaxRockSpeed
+        if (not self.RockManip) and (not self.Dead) and speed and speed > 0 then 
+
+            -- clear it so that GC can take it, if it exists
+            if self.StopRockThread then 
+                KillThread(self.StopRockThread)
+                self.StopRockThread = nil 
+            end
+
+            self.StartRockThread = self:ForkThread(self.RockingThread, speed)
+        end
     end,
 
     --- Stops the unit to rock from side to side. Useful when the unit is on water. Is not used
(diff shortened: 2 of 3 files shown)


KionX:
Maybe should add the default settings instead of deleting rocking?

Garanas:
What would the default settings be?

KionX:
For example, like this:
bp.Display.MaxRockSpeed = 4

speed2CZ:
There actually is a default value (1.5), it's just on a wrong line
https://github.com/FAForever/fa/blob/6da131f33004f5f447eb398234770d63e4ac4023/lua/sim/Unit.lua#L4248-L4249

Garanas:
I don't understand - why do we want to introduce something that has never run (not even in the steam version)?

speed2CZ:
Well it's an option, small detail. We dont know why it was removed (if it ever was a thing). I know it grinds your gears since it goes against everything you're doing here now :D
Anyway if someone decides to enable it, without adding the values into the bps.
At least this line would have to change https://github.com/FAForever/fa/blob/6da131f33004f5f447eb398234770d63e4ac4023/lua/sim/Unit.lua#L4258
to
```LUA
self.RockManip:SetTargetSpeed(- (bp.MaxRockSpeed or 1.5)) 
```
But saving the value direcly in the unit would be better.

Garanas:
I've changed the code slightly so that if it does gets called, at least the threads are not made _before_ the checks whether it makes sense to make them. Made a small benchmark and creating threads is expensive. With that said - I'm not adding in the calls that do the rocking. It is not used by the game and it is not used by any of the common mods. I do not see why we need it, or need to re-establish them as a whole.

image

Garanas:
I'm going to merge this next Monday morning unless someone has an objection to that.

Garanas:
Received no further objection - merging.

We could look at this again in the future if we have spare ms / tick to spent at something.

PR #3338, opened 2021-08-09
Rework Factory Builder Manager

Azraeel:
This updates FactoryBuilderManager Builder Types
Why is this Important? 

This clears up lots of factory searches, what I mean by this is that Factories originally would searching every tech tier so a T1 Land Factory would be looking at T2 and T3 Build Options which of course makes no sense and is tons of Build Conditions that just did not need to be checked and lots of builders that need to be searched. 
This allows the FBM to breathe very much on the factories generally and it should allow them to be much quicker in their decision making and process generally. 
I tested this and it all works including AIs not using the new FBM Builder Types because I left the original ones.


Azraeel:
@Uveso This should not interfere with any other AIs 😄 
Did around 10 Test with LOGs and such. 

Tested with: 
RNGAI
AI-Swarm
AI-Uveso 
Sorian AI
Vanilla AI

lua/sim/FactoryBuilderManager.lua
```diff
@@ -167,15 +175,47 @@ FactoryBuilderManager = Class(BuilderManager) {
         if not self:FactoryAlreadyExists(unit) then
             table.insert(self.FactoryList, unit)
             unit.DesiresAssist = true
+           
             if EntityCategoryContains(categories.LAND, unit) then
                 self:SetupNewFactory(unit, 'Land')
+
+            elseif EntityCategoryContains(categories.LAND * categories.TECH1, unit) then
```
Garanas, line 182: It is better to cache the computations of categories in an upvalue. As an example, this benchmark:
 - https://github.com/FAForever/fa/pull/3333/commits/3f85b343ecabebcdd63243ba33fbce082f28fa49#diff-0f3c3fd31bdc75364cafbeef1c0de7864afd617a6c46366a31114781d2528941

And the results of that benchmark:
image

Caching category computations is always good, especially for longer sequences.

Garanas:
How did you do the testing, as in, where did you add the log statements?

lua/sim/FactoryBuilderManager.lua
speed2CZ: Since this is all one big elseif, non of these other elseifs can be reached if the first one is true.
So in case of the land factory, the very first IF is gonna be true
```
if EntityCategoryContains(categories.LAND, unit) then
```
and checks for the tech category will never run.

Same will happen with air and naval factories as well

Azraeel: @speed2CZ I'll re look into but I was sure it worked. The elseif statements are not grand and I probably should just redo such anyhow

Azraeel:
> How did you do the testing, as in, where did you add the log statements?

I put logs in between each statement that's why theres spaces in between I probably should of left them in, I'll do such when I'm home and I'll fix such things.

Azraeel:
I originally planned on just doing FBM builder types but preferably would like to also optimize this file and the actual buildermanager & PFM eventually as these are expensive AI files on the sim.

Azraeel:
A little bit of an update on this.
This is a bit more complex then originally planned
Although balthazar is helping me with something's so might take a little longer

Garanas:
Take your time - there is no rush. Better to have it good.

Azraeel:
@Garanas any optimization can be done 
100% works 
works with all ai.

lua/sim/FactoryBuilderManager.lua
```diff
@@ -163,19 +172,29 @@ FactoryBuilderManager = Class(BuilderManager) {
...
-                self:SetupNewFactory(unit, 'Land')
-            elseif EntityCategoryContains(categories.AIR, unit) then
-                self:SetupNewFactory(unit, 'Air')
-            elseif EntityCategoryContains(categories.NAVAL, unit) then
-                self:SetupNewFactory(unit, 'Sea')
-            else
-                self:SetupNewFactory(unit, 'Gate')
-            end
+            
+            local ECC = EntityCategoryContains
```
Garanas, line 180: Instead of localizing it each call, turn it into an upvalue. This is performed once when the file is imported for the first time, instead of during each call of this function. The performance is always better or equal to having a local value. See also: https://github.com/FAForever/fa/pull/3333/files#diff-a4e7d18409bfc7d5ce8d08eecf4d20e3de70edb782d87ae9bcac4133863de2aaR12

lua/sim/FactoryBuilderManager.lua
```diff
@@ -163,19 +172,29 @@ FactoryBuilderManager = Class(BuilderManager) {
...
+            
+            local layer = ECC(categories.LAND, unit) and 'Land'
+            or ECC(categories.AIR, unit) and 'Air'
```
Garanas, line 183: You can cache the table calls in an upvalue too, see also: https://github.com/FAForever/fa/pull/3333/files#diff-0f3c3fd31bdc75364cafbeef1c0de7864afd617a6c46366a31114781d2528941R2

Note that the category is larger in the benchmark than you have here therefore the difference wouldn't be as large - but it would still matter to skip the table operation.

lua/sim/FactoryBuilderManager.lua
```diff
@@ -163,19 +172,29 @@ FactoryBuilderManager = Class(BuilderManager) {
...
+            self:SetupNewFactory(
+                unit,
+                self.Brain.HasLayerTBuilders and layer and (
+                    layer..(
+                        ECC(categories.TECH1, unit) and 'T1'
+                        or ECC(categories.TECH2, unit) and 'T2'
+                        or ECC(categories.TECH3, unit) and 'T3'
+                        or ''
+                    )
+                ) or layer or 'Gate'
```
Garanas, line 195: I'm quite confident that this can never return 'Gate' . In what situation would layer not have a value set?

lua/sim/FactoryBuilderManager.lua
```diff
@@ -163,19 +172,29 @@ FactoryBuilderManager = Class(BuilderManager) {
         return retUnits
     end,
 
-    AddFactory = function(self,unit)
+    AddFactory = function(self, unit)
         if not self:FactoryAlreadyExists(unit) then
             table.insert(self.FactoryList, unit)
```
Garanas, line 177: Caching the table.insert in an upvalue is faster. Currently you are pulling a global each time this function is called. Suggestion for the name: TableInsert instead of table.insert. I think it matters to keep the naming convention conveniently the same over all the files for a specific upvalue.

Garanas: I do not know how often this factory list is used and in what other functions it can get an additional value inserted - but keeping track of a counter would be most efficient here. See also: https://github.com/FAForever/fa/pull/3333/files#diff-8e110de264df1c95495186a55be0c54d9acd1e2f57f5e5879372084274f8f6cdR136

Note the additional **0** in comparison to using table.insert.

Note that this is difficult to maintain, especially if the table in question is changed throughout various files.

Garanas (commented):
I have not looked into whether it works as intended - just from a performance perspective. I'll test it later and perhaps @speed2CZ can share his opinion on this too.

lua/sim/FactoryBuilderManager.lua
Azraeel: I'll do the table.inserts throughout the entire FBM File then. :)

lua/sim/FactoryBuilderManager.lua
Azraeel: local EntityCategoryCount = EntityCategoryCount 

for example?

lua/sim/FactoryBuilderManager.lua
```diff
@@ -86,7 +90,7 @@ FactoryBuilderManager = Class(BuilderManager) {
                     end
                 end
             end
-            WaitSeconds(300)
+            coroutine.yield(300)
```
Garanas, line 93: This is not a direct synonym. See also the original definition in my workspace repository:
 - https://gitlab.com/supreme-commander-forged-alliance/other/workspace/-/blob/master/mohodata/lua/simInit.lua#L37

If I'm correct it should be multiplied by 10.

Garanas: And on another note - you are replacing one global call for another (WaitSeconds -> coroutine) and introducing a table operation (the '.'). It is better to cache it in an upvalue (say CoroutineYield = coroutine.yield) and use that instead.

lua/sim/FactoryBuilderManager.lua
```diff
@@ -12,6 +12,10 @@ local AIUtils = import('/lua/ai/aiutilities.lua')
 local Builder = import('/lua/sim/Builder.lua')
 local AIBuildUnits = import('/lua/ai/aibuildunits.lua')
 
+local FAFInsert = table.insert
```
Garanas, line 15: Please do not use the 'FAF' acronym in variable names. It hides the intent and it doesn't add anything. Instead, I'd propose using TableInsert (using [Pascal case](https://www.theserverside.com/answer/Pascal-case-vs-camel-case-Whats-the-difference) notation) for each upvalue.

lua/sim/FactoryBuilderManager.lua
```diff
@@ -12,6 +12,10 @@ local AIUtils = import('/lua/ai/aiutilities.lua')
...
+
+local EntityCategoryCount = EntityCategoryCount
```
Garanas, line 17: This also applies to the FAFInsert / TableInsert variable above - please add a comment as to why these values are added as an upvalue. An example is [this file](https://github.com/Garanas/fa/blob/optimize-effects/lua/sim/Projectile.lua) where I make a distinction between uplifting globals, moho functions and others.

Garanas (commented):
I didn't look whether it works - I'll do that once the review is done.

lua/sim/FactoryBuilderManager.lua
```diff
@@ -423,7 +406,7 @@ FactoryBuilderManager = Class(BuilderManager) {
     end,
 
     DelayRallyPoint = function(self, factory)
-        WaitSeconds(1)
+        WaitTicks(1)
```
Sheikah45, line 409: This change is not equivalent anymore is it? Is this intended?

Azraeel: yes

Dragun123:
It would be better instead of checking TLayerBuilder exists instead to do in AddBuilder do something like
  ```
    AddBuilder = function(self, builderData, locationType)
        local newBuilder = Builder.CreateFactoryBuilder(self.Brain, builderData, locationType)
        if newBuilder:GetBuilderType() == 'All' then
            for k,v in self.BuilderData do
                self:AddInstancedBuilder(newBuilder, k)
            end
        elseif newBuilder:GetBuilderType() == 'Land' then
            --for __,v in self.BuilderData do
                self:AddInstancedBuilder(newBuilder, 'LandT1')
                self:AddInstancedBuilder(newBuilder, 'LandT2')
                self:AddInstancedBuilder(newBuilder, 'LandT3')
            --end
        elseif newBuilder:GetBuilderType() == 'Sea' then
            --for __,v in self.BuilderData do
                self:AddInstancedBuilder(newBuilder, 'SeaT1')
                self:AddInstancedBuilder(newBuilder, 'SeaT2')
                self:AddInstancedBuilder(newBuilder, 'SeaT3')
            --end
        elseif newBuilder:GetBuilderType() == 'Air' then
            --for __,v in self.BuilderData do
                self:AddInstancedBuilder(newBuilder, 'AirT1')
                self:AddInstancedBuilder(newBuilder, 'AirT2')
                self:AddInstancedBuilder(newBuilder, 'AirT3')
        else
            self:AddInstancedBuilder(newBuilder)
        end
        return newBuilder
    end,
    ```
    No longer having to check for TLayerBuilder in the AddFactory function is the main thing here

lua/sim/FactoryBuilderManager.lua
Azraeel: Gate does work, logged it and such.

lua/sim/FactoryBuilderManager.lua
```diff
@@ -178,20 +156,21 @@ FactoryBuilderManager = Class(BuilderManager) {
...
             
+            --LOG('The Layer is ' .. repr(layer) .. ' and I am ' .. self.Brain.Nickname)
             self:SetupNewFactory(
                 unit,
                 self.Brain.HasLayerTBuilders and layer and (
                     layer..(
-                        EntityCategoryCount(categories.TECH1, unit) and 'T1'
-                        or EntityCategoryCount(categories.TECH2, unit) and 'T2'
-                        or EntityCategoryCount(categories.TECH3, unit) and 'T3'
+                        (EntityCategoryContains(categories.TECH1, unit) and 'T1')
```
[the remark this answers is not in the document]
Azraeel: Im not a huge fan of the if.. elseif 
especially pertaining to this, i'd rather through the effort of just commenting on it if people cant read it in all honestly.
I already left 2 important logs for ai devs that can tell them exactly how it works, anyhow.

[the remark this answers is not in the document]
Sheikah45: If the if elseif structure increases readability then that is a large reason to use it. Otherwise if you aren't around to explain it then people have to put in a lot of effort to figure out what is going on.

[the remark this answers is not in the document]
Azraeel: ill have to restructure it if were having a riot about it and really want it to be if elseif because its not built for that rn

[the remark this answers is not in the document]
Garanas: And from another perspective: comments are typically unmaintained and un-executed pieces of code that can also be ambiguous. Meaning that nobody likes to write / update comments, comments are not checked and what I think I get from the comment can be different to what you intended.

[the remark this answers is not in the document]
Azraeel: well hopefully not taken the wrong way,
but I'd have to redo the entire addfactory function basically so :/

Garanas (commented):
A lot more readable, even if it costs 4x the number of lines. I'll inspect it later - great work.

Dragun123:
To reiterate my concerns: 
The way it is handling backwards compatibility by having an arbitrary variable defined in the AIbrain. In this case TLayer. There is no reason for that. At all. What so ever. Zero. Zip. Nada. 

All in this scenerio, btype = ‘Land’ builder actually means is that it can be built by all land factories. While newer the buildertypes added by Azreal is limiting what a factory can build based on that factory tech level.

He also adds a backwards compatibility check so that if this variable is not set to something do this other thing. But let us rewind here. Under ‘AddBuilder’ that function is guess what:

LITERALLY MEANT TO SAY ALLOW you to say one buildertype = many types. So instead of adding or having an utterly pointless check. In this case ‘self.TLayer’ you could instead under addbuilder. Do sometningn like this 

if btype = ‘land’ then 
AddInstanceBuilder(‘LandT1’)
AddInstanceBuilder(‘LandT2’)
AddInstanceBuilder(‘LandT3’)
end

just have to make sure its not in the for loop.

Azraeel:
> To reiterate my concerns:
> The way it is handling backwards compatibility by having an arbitrary variable defined in the AIbrain. In this case TLayer. There is no reason for that. At all. What so ever. Zero. Zip. Nada.
> 
> All in this scenerio, btype = ‘Land’ builder actually means is that it can be built by all land factories. While newer the buildertypes added by Azreal is limiting what a factory can build based on that factory tech level.
> 
> He also adds a backwards compatibility check so that if this variable is not set to something do this other thing. But let us rewind here. Under ‘AddBuilder’ that function is guess what:
> 
> LITERALLY MEANT TO SAY ALLOW you to say one buildertype = many types. So instead of adding or having an utterly pointless check. In this case ‘self.TLayer’ you could instead under addbuilder. Do sometningn like this
> 
> if btype = ‘land’ then
> AddInstanceBuilder(‘LandT1’)
> AddInstanceBuilder(‘LandT2’)
> AddInstanceBuilder(‘LandT3’)
> end
> 
> just have to make sure its not in the for loop.

I disagree, its actually a really nice way for ai devs to opt in and opt out of the new tlayers with just a line of code change. Lemme just like tell you that the way you want it is very forceful for no reason. The way it currently goes is the nice way and you dont want forceful code just adding shit, and factories have no say. It's really a hack job I wont be putting into the main base code. Sure it might work, but we ain't looking for stuff that will just work dragun. It has to work and work right and work the best.

Azraeel:
I disagree, its actually a really nice way for ai devs to opt in and opt out of the new tlayers with just a line of code change. Lemme just like tell you that the way you want it is very forceful for no reason. The way it currently goes is the nice way and you dont want forceful code just adding shit, and factories have no say. It's really a hack job I wont be putting into the main base code. Sure it might work, but we ain't looking for stuff that will just work dragun. It has to work and work right and work the best.

Dragun123:
Yeah have absolutely no idea how this code works do you. What your code does is 3 things:

It seperates the builder jobs into 3 types one a tech level so 9 builders in total.

Then it defines 9 style of builders each corresponding to a job. 

It also allows you to set a multiole buildertype yo a single builder job for example.


If this code you took from LOUD works properly it’ll allow you define or give a builder multiple builder types. Then creates an instance for each builder type defined in the builder data.
 
 If we go as I describe for the old 3 Btype. When the manager loops over at game start a buildertype = ‘land’ for example. What it will do is read that as buildertype = ‘LandT1’, ‘LandT2’, and ‘LandT3’. Its not forceful nor is inellegent it removes a pointless check that literally adds nothing.

 In fact if we wanted we go literally go through tje older ai’s and do this manually doing BuilderType = ‘LandT1, LandT2, LandT3’ if the code you are porting from LOUD Works correctly it be no different than if BuilderType = ‘Land’ under the current setup.

Garanas:
Calm down please - remember that we're all here to make the game better. I'll ask Relentor what his opinion is on the matter.

Azraeel:
That if you go back and recode the bulders to have LandT1, LandT2 or LandT3, you now have the ability to target builders at specific types of factories something you cannot currently achieve with just 'Land'. This is entirely the same as it is with Engineer Builders, some jobs are ideal for T1 or T2 engineers, but not T3 or SACU and vice versa.   The old 'Any' buildertype ignored all that and forced every engineer to review every job in order to figure out which ones he could do.   That's a lot of work if you have 200 different Engineer builders and to summarize, it's no different for factories.   Why should a t1 factory have to wade thru hundreds of jobs to figure out which ones it can do ? By creating this separation, each factory builder can be perfectly tuned to see only the jobs it can technically perform.  No guessing, no wasted effort on tests that will always fail.

lua/sim/FactoryBuilderManager.lua
Garanas: Note that coroutine.yield(1) doesn't wait at all at times. Use coroutine.yield(2) instead.

Dragun123:
I’ll be waiting to continue this discussion further after Uveso review.

lua/sim/FactoryBuilderManager.lua
```diff
@@ -163,18 +163,50 @@ FactoryBuilderManager = Class(BuilderManager) {
...
-                self:SetupNewFactory(unit, 'Sea')
-            else
-                self:SetupNewFactory(unit, 'Gate')
+            
+            local layer = EntityCategoryContains(categories.LAND, unit) and 'Land'
+            or EntityCategoryContains(categories.AIR, unit) and 'Air'
+            or EntityCategoryContains(categories.NAVAL, unit) and 'Sea'   
+            
+            --LOG('The Layer is ' .. repr(layer) .. ' and I am ' .. self.Brain.Nickname)
+            if self.Brain.HasLayerTBuilders and layer then
```
Uveso, line 176:
```
    if self.Brain.HasLayerTBuilders and layer then 
    end 
    if layer and not self.Brain.HasLayerTBuilders then
    end
```
should be changed to 
```
    if layer then
        if self.Brain.HasLayerTBuilders then
        else
        end
    end
```

Uveso (changes requested):

Garanas:
After discussing it with Azreal, I am closing this pull request due to it being aged. It can be re-opened in the future.

PR #2986, opened 2020-02-09
More refactoring

speed2CZ:
Changed
- :IsDead() to .Dead
- :GetArmy() to .Army
- :GetUnitId() to .UnitId
- optimized :GetBlueprint() calls
- optimized local imports
- Save BuildEffectBones for construction units for later use
- refactored Hives code

PR #3340, opened 2021-08-10
Cache engine calls

Garanas:
This is a large PR.

First of all it adds in all the remaining files from the base game. This include the remaining projectule bp / lua files, remaining unit bp / lua files and remaining moho / lua lua script files. I can not think of a reason to not have them in the repository anymore, especially since we do not want to skip any files with these changes. As an example - a while ago the IsDead getter was removed (#2986 ), yet some files still use it simply because it was not in the repository. 

Secondly, All '#' type of comments are replaced with '--' type of comments - this makes it easier for Visual Studio Code to properly recognize the comments. I've made sure to not replace '#' tokens that are part of a string or not inside a bp / lua file.

Thirdly, the following list of refactoring is applied:
 - Moves self.Army = self:GetArmy() from Entity to the 2nd base classes (weapons, projectiles, units, buffs).
 - Adds self.Blueprint as a value for all 2nd base classes.
 - Adds self.Launcher as a value for all projectile classes.
 - Replaces all var:GetBlueprint() calls with var.Blueprint.
 - Replaces all var:GetArmy() calls with var.Army
 - Replaces all var:GetLauncher() calls with var.Launcher.
 - Make sure the base version of OnCreate is called before doing any logic.

The reasoning is performance. The benchmark shows that making engine calls are quite expensive. Therefore making them once in OnCreate(), storing it in self and re-using the results throughout the code is an improvement overall. There are only a few real entities in the game, which are weapons, projectiles, units and buffs. All other entities are somehow one of these through inheritance. If I missed one, please point it out.

image

Note that there are no changes to UI files. These work different and I have not been able to cache the blueprint reliably there. I am also thinking that it may not matter as much.

Note that none of these changes break with any mods, as the original calls (:GetBlueprint, as an example) is still available.

I've tested with a few AI games and there were no errors. It is hard to say whether things got (significantly) faster because each game is completely different from the next. Soon with the benchmark from Marlo we will be able to see changes more reliably.


Garanas:
I've spoken with Softles and I'll use his auto-ai-matchup tool to quickly test dozens of games. I'll report back here.

KionX:
Possible problems Shift-F6 and
https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Patches.lua

I believe that cache engine calls are dangerous optimizations and do not achieve the goal.

Garanas:
_Possible problems Shift-F6_

Changing Economy -> Storage Energy: immediate impact
Changing Defense -> (Max) Health: impact on the next unit made (is set during OnCreate)
Changing Display -> HideLifeBars: immediate impact
Changing Physics -> Motion Type: immediate impact
Changing Physics -> Wobble speed / factor (needs hover motion type to work):  immediate impact, they wobble now :)
Changing Weapons -> Max Radius: immediate impact

All that GetBlueprint() does is retrieve a table that is already part of __BLUEPRINTS (with BlueprintId as its key), which is part of the global state (_G). All we hold is a reference, whether we use GetBlueprint() all over the code or storing its reference to prevent future calls. If anything in the table changes, we will follow along as we would previously. If you change the Blueprint on the fly we will follow with the next unit, as was the case previously too for the majority of the changes.

_https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Patches.lua_

I don't see how this is an argument against this PR. For two reasons:
 - What would someone possible change to these values that we cache that matters?
 - And, less relevant but still relevant, do we actually have someone who works on the engine side of the code?

For the latter, as far as I know we had someone do that type of work about four years ago. For the former, a blueprint remains a blueprint and therefore a reference to some existing table. An army remains a number. And a launcher remains a weapon / unit, therefore a reference to some existing table. For the former and the latter, it doesn't matter as any change to the table underneath is reflected. For the one in the middle: we were already doing this, I just pushed it out over all the files.

_I believe that caching engine calls based on the assumption that the engine will not change anything is a dangerous optimization and does not achieve the goal._

The calls that are changed in this PR are static.
 - Blueprint information does not change over the course of a game.
 - Army of an unit does not change over the course of a game (note: when a unit is switched between armies a new unit is made, with the new army cached).
  - Launchers do not change for projectiles - it is set in stone the moment the projectile is launched.

I don't understand why you'd call them dangerous. LOUD has been doing these type of improvements for years and is as, or more stable than FAF in a typical game.

I also don't understand why you're saying it doesn't achieve the goal. The (theoretical) benchmark and (practical) LOUD clearly show that these small changes matter. Otherwise they wouldn't be doing this over the years.

_I also believe that the main reason for slowing down the game has not yet been discovered._

There is no 'main reason' or 'golden bullet'. There is a lot of inefficient code in the repository, and we need to fix that bit by bit. This PR is a large bit, but certainly one that is worth it.

_According to my observations, the game slows down over time and the speed does not recover even after completely clearing the map of units._

If you have a replay where you took everything out and it doesn't get back to ~ 1 / 2 ms a tick, send it over and we can have a look at it. I did not manage to observer such behavior, even after creating and destroying thousands of units in a single game the last week for certain checks. 

It would indicate that the Trashbag is not always used as intended, causing (likely) Entities to remain and impact the sim.

Long live the Wobble Factor / Speed:

image

KionX:
Here is an experiment that shows that __blueprints != GetBlueprint() != Shift-F6

Shift-F6 AverageDensity = 0.9
SimLua __blueprints['url0001'].AverageDensity = 1
SimLua WARN(__blueprints['url0001'].AverageDensity)
SimLua WARN(SelectedUnit():GetBlueprint().AverageDensity)

Garanas:
You are right that the table in __blueprints is not the table that you get from unit:GetBlueprint(), by default.

Lets dive into the specifics.

Say you have this code in your map:

```
function OnStart(scenario)

    local sx = ScenarioInfo.size[1]
    local sz = ScenarioInfo.size[2]
    local unit = CreateUnitHPR("xsl0001", 1, sx / 2, 0, sz / 2, 0, 0, 0)

    ForkThread(
        function()
            while true do 
                WaitSeconds(1.0)
                LOG(".Blueprint: " .. tostring(unit.Blueprint))
                LOG("GetBlueprint: " .. tostring(unit:GetBlueprint()))
                LOG("__blueprints: " .. tostring(__blueprints[unit.Blueprint.BlueprintId]))

                LOG(".Blueprint (avg. Density): " .. tostring(unit.Blueprint.AverageDensity))
                LOG("GetBlueprint (avg. Density): " .. tostring(unit:GetBlueprint().AverageDensity ))
                LOG("__blueprints (avg. Density): " .. tostring(__blueprints[unit.Blueprint.BlueprintId].AverageDensity ))

                LOG(".Blueprint (Lifebars): " .. tostring(unit.Blueprint.Display.HideLifebars))
                LOG("GetBlueprint (Lifebars): " .. tostring(unit:GetBlueprint().Display.HideLifebars ))
                LOG("__blueprints (Lifebars): " .. tostring(__blueprints[unit.Blueprint.BlueprintId].Display.HideLifebars ))
            end 
        end
    )

end
```

We retrieve every possible blueprint we can find. This is running on the code of this PR.

image

What we see here are the table references. Two of them are the same - each iteration. One is different, but the same overtime. The two that are the same is from the .Blueprint (a cached :GetBlueprint() call) and the :GetBlueprint() call. The one that is different is from _G.__blueprints, which means :GetBlueprint() doesn't retrieve a copy from that table as I originally thought it did.

However, it does retrieve the same table each time we do a :GetBlueprint() call. This means that we can just store the reference, because it remains the same table throughout the game.

image

Here is your Shift-F6 experiment. Regardless of our approach, the changes made in Shift-F6 are applied in-game (health bar is hidden) but not in any of the blueprint tables that we have. This means that your experiment would not work, regardless of the method or table that you used. The changes I've made:
 - Set AverageDensity to 5.0
 - Set Display.HideLifebars to true

The last change is applied in-game - the health bar is hidden. I'm not sure where Shift-F6 applies it changes, as I can't find them. Note that this PR caches the GetBlueprint() call, and is therefore referencing the right table.

But what happens when we change the .bp file, causing the game to notice the change?

image

The reference in the __blueprints table and the one retrieved from :GetBlueprint() matches, where as the (cached) reference in the unit no longer matches. The latter is expected, as it isn't updated. The former is surprising and backs my original claim: these are the same tables (now).

Note that the latter doesn't matter, as in practice:
 - You do not change BPs during runtime.
 - When you change BPs, you typically spawn a new unit anyhow.

It may even be interesting to make a reference to __blueprints instead of :GetBlueprint, causing less unique tables to be allocated :) .

Also the values we changed in the entity inspector are now applied. Not sure why it took a reload for them to apply.

_https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Patches.lua
This is an example of the fact that the engine can change._

Re-iterating the same argument while I already proposed counter arguments does not make it valid again. My arguments still stand (the variables are static, and what could an engine patch possible change to these functions that would matter).

_These are just my observations. It looks like 'main reason' is somewhere in the engine._
Sure, the engine can be better. But there is a large hiccup in the Lua code and this PR fixes one of those. LOUD didn't do any engine patches, yet it can run thousands of units with no problems in contradiction to us, where the game stutters and slows down after 10 - 20 minutes. They didn't do engine patches.

KionX:
The found features confirm the potential danger and a small impact on performance.
I think we've discussed everything. The choice is yours.

Garanas:
_The found features confirm the potential danger_
Can you elaborate?

_small impact on performance._
Can you elaborate?

KionX:
_small impact on performance._
"GetBlueprint() doesn't retrieve a copy"
This is a time measurement for 60 seconds with 2380 units and 8 AI.
Time: 0.094666, Cnt: 155568, What: C
Name: moho.entity_methods.GetBlueprint

Time: 0.023010, Cnt: 19872, What: C
Name: moho.weapon_methods.GetBlueprint

_The found features confirm the potential danger_
"__blueprints is not the table that you get from unit:GetBlueprint()"
"The reference in the __blueprints table and the one retrieved from :GetBlueprint() matches, where as the (cached) reference in the unit no longer matches."
The engine obviously does not know what it wants)

Garanas:
Can you describe how you made the time measurement?

KionX:
_Can you describe how you made the time measurement?_
https://github.com/FAForever/FAFProfiler
Map: Alpha 7
I was an observer.
At the twentieth minute, I started function profiling.

Why do you call that dangerous?
Only because if you miss something, it will be difficult to find the errors associated with it.
And I think 0.15% is not worth the risk.

Garanas:
Could you explain to me how your profiler generated the time information? Does it call `GetSystemTimeSecondsOnlyForProfileUse` before and after each usage?

KionX:
This is a difficult question)
Everything is very difficult there.
I wrote this 6 months ago and have already forgotten the details.
It measures the execution time of code segments between function calls.

Garanas:
That would be the better approach than what I described, but if you can't explain to me how you took into account the threading and waiting and all the other hassles that it introduces by computing the time between segments then I can not rely on your profiling results as a counterargument to my benchmark. An an example, how does it deal with a call that happens only once per thread?

KionX:
Threads don't matter. They all execute code segments. I just sum them up through all the threads.

Waiting is not part of the thread's work. They are discarded.
There are waits between the end of a function and the beginning of another one.
And Yield(). These are not code segments. Yield() is a function counted as a function.

_I can not rely on your profiling results_
Don't rely on my profiling. We need to find out which one is more accurate.

Garanas:
_Don't rely on my profiling. We need to find out which one is more accurate._

That is a statement I can work with!

I'm not sure how you discard the waiting, I'd need to deep-dive into your code.

KionX:
I'm not on Discord, don't speak english. It will be difficult to discuss this. I use Zulip.

Garanas:
I'd like to add to the arguments here that LOUD has been doing quite _exactly_ what my profiler / benchmarking is showing:
 - https://github.com/LOUD-Project/Git-LOUD/commit/d2356c206ddcfbcc1eed910a01a1259d0f14514d

And now they have breached the record of having 1000 units, with 2 AIs running at +10 sim on a 20x20 map. Hence, I still think that 0.15% improvement is an understatement and we should look into what the benchmarks are showing.

KionX:
There is a suspicion that your tests get a huge boost to speed from the processor cache.
Your test is very simple and easily cached.

How can I measure the execution time(in the wild) of GetBlueprint in your profiler? I didn't find any documentation.

KionX:
LOUD-Project/Git-LOUD@d2356c2
__blueprints[v.BlueprintID] This is not exactly how this PR does it.

Garanas:
_There is a suspicion that your tests get a huge boost to speed from the processor cache._
That is likely the case, but so do the tests that they are compared with. And as long as that is the case and the comparison holds up then it doesn't matter. It is about the difference between the tests / benchmarks that are related, not their absolute values.

_How can I measure the execution time(in the wild) of GetBlueprint in your profiler? I didn't find any documentation._ 
It can't do that. I tried getting a reliable time difference between smaller functions and failed, several times. I am still a firm believer that you can not reliably profile a single function call of this low complexity. Even with your alternative approach - I do not trust it is reliable as the thing it is measuring 

___blueprints[v.BlueprintID] This is not exactly how this PR does it.(Just noted the difference.)_
You're correct. This PR is the evolution of how Sprouto would do it, completely evading the table operation too. I've discussed it with him and the conclusion was that this is better than what he has right now. The reason I was referencing the PR is me saying 'look, these small things matter and LOUD is the practical proof' and not necessarily on what he does (but note that a lot of the other parts of the PR (caching categories in upvalues, using the alt count loop, localizing often used globals - etc are inspired by the benchmarks and apparently have quite the impact on their sim - lets make it impact our sim too :) )

KionX:
_are inspired by the benchmarks and apparently have quite the impact on their sim - lets make it impact our sim too_
Every argument you make convinces me otherwise.

https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Patches.lua
Imagine that this function will not accept 0 or 1, but a reference to a table.
Can assign any blueprint to the unit. This optimization will break.

Maybe someone else has something to say?

KionX:
```
function Test()
    local unit = SelectedUnit()
    local t = GetSystemTimeSecondsOnlyForProfileUse()
    for k = 1, 155568 do
        unit:GetBlueprint()
    end
    WARN(GetSystemTimeSecondsOnlyForProfileUse() - t)
end
```
Test:
WARNING: 0.034255

My profiler:
warning: Time: 0.079239, Cnt: 155568, What: C
warning: Name: moho.entity_methods.GetBlueprint

warning: Time: 0.041336, Cnt: 1, YieldTicks: 0, Yield: 0, Line: 887, What: Lua
warning: Name: Test
warning: @c:\programdata\faforever\gamedata\lua.nx5\lua\siminit.lua

In wild(previous):
Time: 0.094666 = 0.15%, Cnt: 155568, What: C
Name: moho.entity_methods.GetBlueprint

One call GetBlueprint is 2.201930e-7
In my opinion, the accuracy is excellent for such a small function.

Garanas:
I am done discussing this with you.

I'd like to hear what other people are thinking.

env/Seraphim II/Props/Emitters/SnowSeraphim1_script.lua
```diff
@@ -1,9 +1,12 @@
...
-BlowingSnow01 = Class(import('/lua/sim/Prop.lua').Prop) {
+--
+-- LavaSteam01
+--
+
+local Prop = import('/lua/sim/Prop.lua').Prop
+BlowingSnow01 = Class(Prop) {
 
     OnCreate = function(self)
+        Prop.OnCreate(self)
```
speed2CZ, line 9: What does this change?

lua/AI/AIBaseTemplates/ChallengeExpansion.lua
```diff
@@ -0,0 +1,77 @@
...
+            SCU = 1,
+        },
+        MassToFactoryValues = {
+            T1Value = 6,
+            T2Value = 15,
+            T3Value = 22.5
+        },
+    },
+    ExpansionFunction = function(aiBrain, location, markerType)
+        if markerType != 'Expansion Area' then
```
speed2CZ, line 66: Can you also replace all != to ~= ?

env/Seraphim II/Props/Emitters/SnowSeraphim1_script.lua
Garanas: It guarantees that the OnCreate is called since the blueprint and other values are cached during OnCreate. If something doesn't propagate to the top it either has to set those values itself, or add the original function.

lua/AI/AIBaseTemplates/RushMainBalanced.lua
```diff
@@ -203,14 +203,14 @@ BaseBuilderTemplate {
         end
 
         if true then
```
speed2CZ, line 205: Iguess this IF can be removed completely

lua/AI/AIBaseTemplates/ChallengeExpansion.lua
Garanas: Sure I can.

env/Seraphim II/Props/Emitters/SnowSeraphim1_script.lua
speed2CZ: Isn't this and others just effects? So the only thing it needs to do is to create the emitters?

Garanas:
I thought I excluded those, let me fix that :)

Crotalus:
According to profiling, I'm not sure that either of your benchmarks can show how much of a gain this is, it's not all about CPU cycles.

Imagine late game when the garbage collector is saturated with work, then I guess that not having to create a copy of a table every GetBlueprint() could be very well worth it. 

What about running a long setons replay as benchmark, if we're lucky this change won't affect the sim state so you can run same replay before / after the change.

Garanas:
That is why Marlo is making a benchmark map that would allow us to make reasonable comparisons. It would run the same way each time, allowing us to A check behavior, and B, check whether it runs faster.

env/Seraphim II/Props/Emitters/SnowSeraphim1_script.lua
Garanas: Yes, in this case. I suppose this one could get removed. I did it for consistency - in case it gets changes in the future. I suppose I can revert these changes along with other effect props.

Crotalus:
IMHO the biggest issue with FAF is the late game lag so any effort that can help fix this is commendable.

In the PR you linked example I tried to reduce table allocation & transfer between Sim.Sync -> UI side, I know that's expensive and it runs every tick. 

There's low hanging fruits in SimSync.lua where it allocates empty tables every tick which seems unnecessary. Better do check if a table exist on the UI side (CameraRequests, Sounds, Voice, AIChat etc)

Garanas:
And those are exactly the things that the benchmarking is showing too.

On another note, how do I run FAF replays?

image

Garanas:
A few test results where the left side is the adjusted branch based on deploy/faf and the right side is the deploy/faf branch.

image

image

image

image

image


And a few disclaimers:
 - The right side started about 0.5 seconds earlier than the left side.
 - Both FA instances have their own unique CPU cores on real time priority.
 - Contains only the blueprint change (caching it in self.blueprint) for the files that are in the deploy/faf repository. This is by far not all the files that it should reach.
 - This does not contain the launcher / army / (remaining) IsDead changes included in this PR.
 - The left side desyncs at minute 10, but the output is identical so far. This causes the console to be filled with warnings - I disabled seeing warnings (but errors are still visible if they are any) for both windows and cleared the moholog every for both instances ~minute (real time, not game time).

With that said, the left side doesn't just catch up it gets ahead by quite a bit and this isn't touching all the code or doing all the changes.

Crotalus:
Well that shows us that these improvements can actually make a real improvement for the game speed. Maybe keep adding "non deterministic breaking" changes into a branch and use this replay as a benchmark then.

Garanas:
I'm going to split this PR into five PRs tomorrow:
 - One that adds all game files (in combination with the schook changes).
 - One that removes all remaining .IsDead() calls.
 - One that removes all remaining :GetArmy() calls (except during OnCreate).
 - One that removes all :GetLauncher() calls (except during OnCreate) (we can all agree that this does not change).
 - And last, one that removes all :GetBlueprint() calls (except during OnCreate).

The latter stirs the soup the most, then we can continue the discussion about that in that specific PR.

Garanas:
I'm closing this PR in favor for the smaller PRs.

PR #3341, opened 2021-08-11
Update T3 Mobile AA File Directory Names

MrRowey:
The Unit Id on the Unit database should take you to the Blueprint for the unit. This is the quickest way to do it but with the T3 Mobile AA the Unit Folders are all in LowerCase so the link the the Unit DB is pulling seems to be Case Sensitive as it throwing a 404 page as it is after the unit folder in Capitals like all the other units

So the Unit Folders need to be as following
DRLK001
DELK002
DALK002
DSLK004


units/DELK002/DELK002_unit.bp
```diff
@@ -177,7 +177,7 @@ UnitBlueprint {
             LAYER_Orbit = false,
             LAYER_Seabed = false,
             LAYER_Sub = false,
-            LAYER_Water = false,
+            LAYER_Water = true,
```
Garanas, line 180: Is this an intended change?

units/DELK002/DELK002_unit.bp
```diff
@@ -177,7 +177,7 @@ UnitBlueprint {
             LAYER_Orbit = false,
             LAYER_Seabed = false,
             LAYER_Sub = false,
-            LAYER_Water = false,
```
speed2CZ, line 180: Why is this changed?

Garanas: You beat me to it 📦

MrRowey:
Ive Not Changed any files inside the folder i literlay downloed FA-Develop Renamed the folders in CAPS and deleted the folders in lower case thats all i changed. Im not aware of a quicker and simpler way of renaming a Folder Directory on Github.

Garanas:
I'll revert the changes - after that I see no reason why we should not accept this.

Garanas:
@Tagada14 is the last change of Rowey all good? It means that the reload bar is visible, similar to summits.

Garanas:
And @MrRowey could you make a separate PR for the reload bars?

MrRowey:
@Garanas Yeh sorry forgot to make a seperate branch on my fork so i pull the first change here xD

Garanas:
Reverted the changes, merging this PR on Monday morning unless there are objections. Great work Rowey!

PR #3342, opened 2021-08-11
Strategic bombers rework

Tagada14:
Increase the mass killed required for a vet from 1.25x mass cost -> 2x mass cost (4.2k mass)  for all Strategic Bombers

All strats: 
Reduce the GuardScanRadius and MaxRadius to 65 (radar range), this reduces the trigger range for the strats which should limit the frustration of bombers randomly triggering on units far far away.

Cybran Strat :
AoE Radius 7 -> 6.5
EMaintananceCost 60 -> 150

UEF Strat :
Damage 3000 -> 3100
AoE Radius 6 -> 5.5

Sera Strat
Damage 3250 -> 3000
AoE Radius 5 -> 6

Aeon Strat
Damage 3450 -> 3000
AoE Radius 4 -> 5.25
Added 3s stun to its bomb
Added a Chrono-like effect to the Aeon Bomb
Note: The Stun radius is set to 5.80 which is the closest approximation I could get after many tests since before even with the same value as the weapon AoE it had a lot less range allowing the bomb to damage units without stunning them. With a 5.80 range, it's both possible to stun without damaging and damage without stunning but in 95%+ of cases, the bomb should both stun and damage the target.

To do :
Edit the emitter effects to better reflect the new AoE ranges. - Not needed, it's pretty close already
Add a Chrono-like effect to the Aeon Bomb. - Done

### DIFF
--- a/effects/emitters/quark_bomb_chrono_effect_01_emit.bp
+++ b/effects/emitters/quark_bomb_chrono_effect_01_emit.bp
@@ -1,0 +1,155 @@
+EmitterBlueprint {
+	BlueprintId = 'aeon_chrono_dampener_02',
+	Lifetime = 1.00,
+	Repeattime = 1.00,
+	TextureFramecount = 1.00,
+	Blendmode = 1.00,
+	LocalVelocity = false,
+	LocalAcceleration = false,
+	Gravity = false,
+	AlignRotation = false,
+	AlignToBone = false,
+	Flat = true,
+	LODCutoff = 100.00,
+	EmitIfVisible = true,
+	CatchupEmit = true,
+	CreateIfVisible = false,
+	SnapToWaterline = false,
+	OnlyEmitOnWater = false,
+	ParticleResistance = false,
+	InterpolateEmission = true,
+	TextureStripcount = 1.00,
+	SortOrder = 0.00,
+	LowFidelity = true,
+	MedFidelity = true,
+	HighFidelity = true,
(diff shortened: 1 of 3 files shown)


units/URA0304/URA0304_Script.lua
```diff
@@ -20,7 +20,9 @@ URA0304 = Class(CAirUnit) {
     
     OnStopBeingBuilt = function(self,builder,layer)
         CAirUnit.OnStopBeingBuilt(self,builder,layer)
+        Turns Stealth off when unit is built
```
speed2CZ, line 23: This should be a comment

units/URA0304/URA0304_Script.lua
```diff
@@ -20,7 +20,9 @@ URA0304 = Class(CAirUnit) {
...
         self:SetScriptBit('RULEUTC_StealthToggle', true)
+
```
speed2CZ, line 25: No need for extra line here

speed2CZ (changes requested):

units/URA0304/URA0304_Script.lua
Tagada14: Fixed.

units/URA0304/URA0304_Script.lua
Tagada14: Fixed

speed2CZ (approved):

PR #3343, opened 2021-08-11
Adjust T4 Arty Ranges

Tagada14:
Give both Salvation and Scathis more range to make them equal to Mavor's range.


Tagada14:
Increased Scathis Range to 4k (same as Mavor)
Salvation's range can't be increased sadly.

Issue #3344, opened 2021-08-11
wld_RunWithTheWind


Garanas:
This console command _should_ run the sim at the highest speed possible. However, instead it runs at a lower fps / sim rate than just running the game at +10.

I've been going back as far as as this commit (2017):
 - https://github.com/FAForever/fa/commit/08a23255fef9103bb3fdcef6a50caa30d682c927

Yet, the issue is still there. My question is whether someone knows when it broke.

Garanas:
Commits where it works:
 - https://github.com/FAForever/fa/commit/08a23255fef9103bb3fdcef6a50caa30d682c927

Garanas:
Closing this issue as this has been fixed with game version 3736.

PR #3345, opened 2021-08-11
Refactor no singleplayer net lag

Garanas:
Instead of having it on with no choice it is now an option in the option menu. It allows for AI-only games to run at maximum-maximum speed. Useful for the AI-match tool that Softles (Softly) made to test larger PRs more effectively. The game needs to be restarted in order to make use of the adjusted option. 

Solves issue #3344 .


Crotalus:
Hmmm, that's a weird side effect of setting net_lag 0. I just tried setting net_lag to 5000 while running wld_RunWithTheWind and a singleplayer game runs even faster than net_lag 500.

So for someone trying to maximize replay speed they should run net_lag with a high value and then wld_runwiththewind. 

Maybe a quick command with some obscure CTRL+SHIFT+<key> combo to toggle this would be better than this option.

Crotalus:
Hmmm, I seem to remember a way to change the interval of which a checkpoint should be run.

It would be actually interesting to raise that value alot, we don't care about if the game tells about a desync every 5 second instead of every 0.5 second.

Crotalus:
Console command: `sim_ChecksumPeriod`

The default is 50 ticks, so it calculates the checksum of the game every 5 seconds as default. Not as bad as I thought then.

Crotalus:
Another idea would be:
 `if SessionIsReplay() then ConExecute('net_lag 5000') end`

But maybe doesn't work if the replay needs to be restarted before you see an effect? In a local game net_lag change seems to work instantly,

Garanas:
I like those ideas. Wouldn't it be interesting to lower the checksum interval in general? As the majority of games do not desync and when it does it doesn't matter how often the checksum runs anyhow.

I am personally not a fan of making it a hidden toggle through hotkeys. Primarily because it does impact mods too, such as the run at max sim speed mod. And all that mod does is enable wld_RunWithTheWind.

The replay needs to be restarted because I do not set the net_lag value back to its original value. I could do that too.

Crotalus:
I changed sim_ChecksumPeriod to 5000 after a couple of seconds in game:
image
image

Only 8 checksums in 1011 ticks so it seems to be possible to change in game

Garanas:
I understand it doesn't persist, we could set it on launch like the net_lag is set right now. Or, we could set it before launch (say, in the lobby when you click the launch button) if it needs to happen before the sim starts. I wonder if the checksum has a (relative) large impact on the performance of the game.

Crotalus:
Seems like you could set it while game session is already running (have to be tested with multiple players - my test was in a singleplayer game).

I can't say I notice any big difference with a less frequent checksum though :/

lua/ui/game/gamemain.lua
```diff
@@ -232,9 +232,12 @@ function CreateUI(isReplay)
         AddBeatFunction(UiBeat)
     else
         local clients = GetSessionClients()
-        if table.getsize(clients) <= 1 then
-            -- No need for unnecessary lag when playing alone
-            ConExecute('net_lag 0')
+        local option = Prefs.GetFromCurrentProfile('options')['no_singleplayer_netlag']
```
KionX, line 235: It is better to write it like this(faster):
local option = Prefs.GetFromCurrentProfile('options.no_singleplayer_netlag')

lua/ui/game/gamemain.lua
```diff
@@ -232,9 +232,12 @@ function CreateUI(isReplay)
...
+        if option then 
+            if table.getsize(clients) <= 1 then
```
KionX, line 237: if table.empty(clients) then

lua/ui/game/gamemain.lua
Garanas: Yes - and that works in your PR. But it didn't work in this one yet :) !

lua/ui/game/gamemain.lua
Garanas: No - when it is empty there are no clients at all. This triggers also when there is one client - when someone plays alone.

lua/ui/game/gamemain.lua
KionX: Perhaps this is due to _

Crotalus:
sim_ChecksumPeriod requires cheating on so it's not possible to change the default without exe-modification

Garanas:
Did you try calling that through code, or by manually entering it in the console?

A lot of console commands require cheating on when done by an individual, but doesn't trigger it afaik if it is run by everyone through code.

Crotalus:
Tried doing ConExecute('sim_ChecksumPeriod 100') in gamemain but then unless cheating was on it complained about trying to cheat.

When I put it into lobby code instead the setting wasn't changed, even with cheating on. Network settings is possible to set without cheating though so I'll make a PR with a lobby option to raise the network limits, defaulting to standard.

Garanas:
Running it like this causes the same issue:

image

Question is - does it get applied regardless? I don't know how to check that.

Crotalus:
WIth cheating off the only way to check it would be to parse the replay and count the number of checksum-entries.

My guess would be that the setting hasn't been changed then, I can verify later tonight

Garanas:
Closed in favor of #3653 , I'm not sure why but something is wrong with my local setup.

Issue #2527, opened 2018-07-17
add load time bar


shalkya:
that could be limited to percy, bombers, neptune, destro, bs...
this would behave like T2 mobile arty, mml or cruiser
http://forums.faforever.com/viewtopic.php?f=42&t=16440

MrRowey:
Ive managed to get the Percy working with the load time bar for the reload time of the shot but not sure how to go about the other ones EG

Air is allready using the bar to display the fuel of the aircraft
Neptue,Destroyers & BS Most have more than 1 weapon that tend to not be synced so having the load bar would jsut make things more confusing

Garanas:
The summit already has a reload bar and is not synced either (with regard to the various turrets). I think it helps a lot, as typically they all fire in sequence anyhow.

MrRowey:
Oh intresting, Is the summit the only BS that has the reload bar

MrRowey:
See #3347 for changes

Garanas:
As far as I know - yes.

Garanas:
Closed as the #3347 has been merged. Note that the percy was denied a loading bar by the balance team.

PR #3347, opened 2021-08-12
Add render fire clock for Battleships & Battlecruiser

MrRowey:
Added the Render Fire Clock to the Percy & the other 3 faction BS as Summit allready has it as well as the Neptune. base of Issuse #2527


Garanas:
I think this is also a balance PR. @Tagada14 , can you look into this?

Garanas:
I thought I'd change base to the balance branch, but after seeing the explosion of additional commits I regretted my decision quite immediately.

Tagada14:
For battleships it's fine but Percies are a no go.

Garanas:
@MrRowey if you can make the changes before Sunday then I can test it and then it will be merged on Monday morning.

MrRowey:
@Tagada14 So the BS and the BC are fine just remove Percy Change ? just want to fully clarify this before I make the changes

Garanas:
This PR has been in FAF Develop for a bit and people got back to me about the usefulness of the reload bar for the Cybran and Seraphim battleships. Their reload times are rather short, making it look strange when turrets fire while it didn't 'reload' yet.

Also - what about the Tempest?

PR #3348, opened 2021-08-12
T3 Sonar rebalance

Tagada14:
Changes:
Add Jamming to UEF t3 Sonar
Maintenance Cost 500 -> 550


speed2CZ:
Has anyone tested this? You disable jamming and you get free sonar.

Petricpwnz:
I briefly tested but wasn't paying attention to the drain, good catch. Notified tagada, otherwise will fix later together with some aeon sonar changes perhaps.

PR #3349, opened 2021-08-13
Update README.md

KionX:


Garanas:
I've been working on profiler for about a month now and recently it has a repository that you can find here:
 - https://gitlab.com/supreme-commander-forged-alliance/other/profiler

image

It prevents using the logger to show information as the logger itself inherently slows down everything. It includes benchmarking to allow you to experiment with various snippets of code. It also includes an UI (if you have the Dear Windowing mod enabled from the vault) that allows you to interact with everything. And last, it allows for counting how often functions are called. Keeping track of time has proven to be unreliable (whether because the function logic is too small to reliably capture or because of wait statements).

In the near future it will also support timers (that you have to place yourself) and stamps differences (allowing you to see what functions and how often they are called between two lines of code).

With that said, I do not understand why you claim your profiler to be the 'Actual FAF profiler' while the repository existed for about four hours at the time of writing. If you want it to be the one everyone uses - I don't particularly mind. I'll just use my own. But I think a discussion before making a PR like this is appropriate.

The same applies to the other change (the binary patches): your repository is a fork. Is it the one we use in practice?

KionX:
I didn't know about your profiler. And no one has documented this before me. I suggest adding a link to yours too.
The fork of binary patches is agreed with the author. Now it's official.
If you are interested in the details, start here:
https://faforever.zulipchat.com/#narrow/stream/203495-game-development/topic/exe/near/202249212

Garanas:
That is important too, but not what I was referring to. I was referring to whether FAF actually uses your branch. And if that is the case, then I'm all good having it there as the actual patch.

KionX:
FAF is currently using the pre-fork version. Maybe should write "official"?

Crotalus:
Logger slowdown can be prevented by using SPEW('logmessage') instead of LOG('logmessage')

KionX:
I don't think this is a problem. The logger is needed.

Crotalus:
Isn't the result of SPEW('blabla') and LOG('blabla') the same? Text ending up in console and in log file. SPEW is much faster though.

Very noticable if something spams LOG message due to crashing UI etc. But I guess in a normal game the number of log messages per second is quite low so less gain there

KionX:
I think that SPEW and LOG are just different prefixes(debug and info).
But I didn't measure their performance.

Crotalus:
I just tested and you seem to be right, they both seem to be slow. 

```
for i = 0, 10000 do
        SPEW("Log")
    end
```

Locks up the game for ~10 s, I wonder why it's so fucking slow

Garanas:
The reason SPEW tends to be faster is because it is not checked to view by default in the moholog. If you uncheck LOG / WARNING / ERROR when it gets polluted the game is suddenly faster too (it still has an impact - though!)

I'm not sure about the details, but I suspect some string concatenation is going on and that is why it is slow when visible. The moholog is part of the engine if I am correct, so we can't see what is really happening.

KionX:
I'm of the same opinion.

KionX:
Maybe it's time to solve something with this PR?

Garanas:
As long as FAF doesn't use your fork in practice then I personally do not think it should be referenced as 'Actual' - maybe as 'Future'.

With regard to the profiler - I'll leave that to Keyser.

PR #3351, opened 2021-08-13
Imgui

Garanas:
_Do **not** merge this branch yet_

This PR will add a UI library to the game that is easy to use, instead of the complicated and often convoluted code that working with the UI requires. It is based on the principles of the [Dear Imgui](https://github.com/ocornut/imgui) C library. The aim is to be able to write UI in a declarative manner. 

As an example, the Profiler UI that is invisible in this screenshot is [this code](https://gitlab.com/supreme-commander-forged-alliance/other/profiler/-/blob/main/mods/profiler/modules/views/table.lua):
image

Another example, the UI visible in this screenshot is [this code](https://gitlab.com/supreme-commander-forged-alliance/other/profiler/-/blob/main/mods/profiler/modules/views/main.lua#L62):
image

The library allows for easy debug interfacing, instead of endlessly using the log for everything. The UI can be changed on the fly, allowing you to make changes in your code and immediately seeing the changes in-game (assuming diskwatch is enabled). It is great for making. It can also be used for mods to make their UI, an example is the Battle Royale or Cinematics mod that is in the vault (make sure to have the vault version enabled and not this branch, as they expect the vault version).

 - [ ] Create examples to show how you can work with the code.
 - [ ] Fix the list not working for some elements.


Crotalus:
Looks super nice, a revamped UI for the game would be very nice. Especially if we could get rid of all the "update ui by looping every x ticks" and instead make it more reactive.

(the lazyvar system is underused in the UI code)

Garanas:
How do lazyvars work actually? I never understood what made them different, except for generating hard to understand error messages.

Crotalus:
Yeah I don't like them either, hard to debug, but you can enable a debug flag (`ExtendedErrorMessages`) on them in /lua/lazyvar.lua so they show more information about the "lazy backtrace". I don't think lazyvar.lua is checked into FAF though.

IIRC when they change value they trigger all other lazyvar that depend on them

It's something like:
```
element.bgCol = LazyVar.Create()
bgCol.OnDirty = function(var)
   -- This will run when value is changed
    self:SetNewColor(var)
end

-- If you set the variable, the OnDirty method will run
bgCol:Set('#ff0000')

-- Also you can have other lazyvar depend on each other by passing in a function
element.color = LazyVar:Create()

element.color:Set(function()
-- set the same color as background, very useful
return bgCol
end)




```

Garanas:
Hence, the lazy part - only computed when it matters.

Sheikah45:
These seem similar to observable values that I have worked with. I wonder what the computation cost of them is here in lua? But observables like this are very handy in reducing check time for the UI.

Garanas:
Means the imgui library got some more work to do, just what I was looking for 😄

Crotalus:
You're already using lazyvar when doing stuff like this:
image

What's missing in current FA is using lazyvar or similar for the data updates in text labels etc, not only the UI sizing stuff.

Garanas:
Could you give an example as to how that would look like?

Crotalus:
```lua
Mass = LazyVar.Create(0)
Energy = LazyVar.Create(0)

-- Function that makes the necessary changes to the UI when mass / energy changes
function UpdateEconomyUI()
  if Mass() > 0 then
    LOG("You have more than 0 mass!")
  else
    LOG("You are broke")
  end

  if Energy() > 0 then
    LOG("You have more than 0 energy!")
  else
    LOG("You are broke")
  end
end


-- First run with initial values
UpdateEconomyUI()

-- In another file we decide to do all state updates so there we do:

Mass:Set(500)
Energy:Set(1000)

-- Now all functions that have used those lazy vars will be run, so in this case the log output is:
-- INFO: You have more than 0 mass!
-- INFO: You have more than 0 energy!
```

There's a variant of this observable pattern here `lua\ui\lobby\data\watchedvalue` which ckitching made for the lobby, I know that one supports array & tables

Garanas:
Ah, I did it again. Should I have force pushed after the rebase?

Sheikah45:
Yeah after rebasing you have to force push because the history has changed

Garanas:
Move backwards to milestone 3727 because of time pressure.

Garanas:
I am closing this pull request. The last week I've been working with the UI system in place and that is fully functional and reasonable - the error messages are just quite poor and incomprehensible. In my opinion any future effort on improving the UI system would be to introduce better error messages.

PR #3352, opened 2021-08-13
Schook removal

### LINKED ISSUE #3335: Schook
The schook.scd file from the base game is also schooking in the repository. This means that we're hooking, while we have access to all the code. There is no need to do that. Instead, the code in schooking should be embedded in the actual files and the schook.scd should be removed from the init_faf.lua whitelist.
 - [ ] Embed code in repository/schook in their respective files
 - [ ] Add schook.scd to blacklist of init file
 
With regard to #3334 - the only thing that is taking loading time at the moment is this unnecessary hooking that can't happen until the game starts.

Garanas:
This PR will close #3335 . The aim is to completely remove the schook folder, as it adds (maintainability) complexity that I'm not entirely sure of why we would want that.

As soon as I got the AI matches working I can firmly test these changes.


Garanas:
Note that when testing, you need to remove the schook.scd and schook.nx2 from your init_dev file.

KionX:
In SC2, schook is also removed.

Garanas:
After discussing it with Sheeo with regard to https://github.com/FAForever/fa/issues/201 it appears this route was chosen specifically for trying to adhere to the GPLv3 license.

KionX:
My thoughts:
1. Schook exists because the FA developers used it.
2. They changed their mind in SC2.
3. https://github.com/FAForever/fa/issues/201
  From what I can see, no one has ever adhered to this.
4. If we need this, then we should transfer all the FAF files to Schook.

Garanas:
I understand your perspective, but just because we didn't adhere to in the past doesn't mean we shouldn't start thinking about doing it. Regardless - I am on your side on this discussion and I'd prefer having all the base game files in the repository.

It may be good to have a discussion about this with the association. Not being able to have all base game files makes development more difficult than it has to be.

Crotalus:
The idea of schook everything and never add base game files may be an ideal but will never happen. So I'm all for adding the rest of the blueprints / lua.

I understand the other side concerning the license but I'm not sure it matters much if we have 30% or 100% of the lua files in this repo. 

Also, dont quote me on this but I think adding all files to the repo also would give a slight performance gain. Due to the layered filesystem the game will check each hooked location when it imports a file. First all the maps, mods directories, then the faf files in nx2 and then the original base game files. In this game it will be found without checking base dir.

Garanas:
I'll discuss it with Sheeo - I won't quote you on the performance part :)

KionX:
If the copyright holder wants to press the FAF, he will find a reason.

Garanas:
I'd like this to be reviewed beyond just testing with bots - if you're available Speed?

Garanas:
I still want this to happen - but this pull request has aged.

PR #3354, opened 2021-08-15
Reduce table creation with EventCallbacks

Crotalus:
Instead of creating 24 tables for every unit we create these when required

Could break some mod if they are manipulating / calling EventCallbacks table on a unit instead of using the AddUnitCallback etc

### DIFF
--- a/lua/sim/DefaultProjectiles.lua
+++ b/lua/sim/DefaultProjectiles.lua
@@ -142,7 +142,7 @@ NukeProjectile = Class(NullShell) {
 
     LauncherCallbacks = function(self)
         local launcher = self:GetLauncher()
-        if launcher and not launcher.Dead and launcher.EventCallbacks.ProjectileDamaged then
+        if launcher and not launcher.Dead and launcher.EventCallbacks and launcher.EventCallbacks.ProjectileDamaged then
             self.ProjectileDamaged = {}
             for k,v in launcher.EventCallbacks.ProjectileDamaged do
                 table.insert(self.ProjectileDamaged, v)
--- a/lua/sim/Prop.lua
+++ b/lua/sim/Prop.lua
@@ -19,10 +19,12 @@ Prop = Class(moho.prop_methods, Entity) {
     end,
 
     OnCreate = function(self)
+        --[[
         self.EventCallbacks = {
             OnKilled = {},
             OnReclaimed = {},
         }
+        ]]
         Entity.OnCreate(self)
         self.Trash = TrashBag()
         local bp = self:GetBlueprint()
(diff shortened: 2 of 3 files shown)


Garanas:
This sounds like a good deal. This type of improvement is applicable everywhere:

In [https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L127](unit.lua) we could not pre-allocate all callback tables, but make them as we go. In PvP these tables are essentially not used. Then we also have [these thrashbags] that do not apply to all units. For example, an Aurora will never use the BuildEffectsBag. This could be moved down the hierarchy.

We have a similar situation in [props.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Prop.lua#L22). Note that a map can easily contain thousands of props - all having no callbacks at all in PvP.

Then we have inside [weapons.lua] the weapon priority tables that can easily be cached. For each unit made, each weapon is initialized and a new table is made just to use it once and then throw it away. Especially when spam is involved this may easily be hundreds of tables a second. I managed to capture this [in this branch](https://github.com/Garanas/fa/blob/optimize-effects/lua/sim/weapon.lua#L427) where we give each weapon a unique id (unit blueprint + weapon index) and use that as a key for caching. If the unit - weapon combination doesn't exist yet, it is computed and stored. Otherwise the same table is retrieved as the previous unit used during initialization. This prevents re-allocation and it prevents deallocation.

Then we have these allocations in [projectile.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Projectile.lua#L86) that we could either remove completely, or initialize to false. The same would apply to [defaultprojectiles.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/DefaultProjectiles.lua#L38) and all the faction-projectiles such as [inside aeonprojectiles.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/aeonprojectiles.lua#L69). It would require checking for the variable during a for loop - which is more efficient than running the for loop on an empty table. However, that requires quite some refactoring. That is [quite](https://github.com/Garanas/fa/commit/2cc658ed293097ac09a3af5a0a8250e63b268d5f) [a](https://github.com/Garanas/fa/commit/913f935cbec6b8044d94b2cc30aeaa2b48caf322) [lot](https://github.com/Garanas/fa/commit/6e1e20bd777e9310e783edb4a559e7970419821b) [of](https://github.com/Garanas/fa/commit/03dfd914a3b2ad2537018c1186aa4cfbb5397c0e) work when you combine it with other optimalisations, but - you do get less stutters and 400 zthuee can run almost 30% - 40% faster.



In

speed2CZ:
Im all up for improvement but it would be best to avoid breaking mods

Garanas:
Speed2 is right about that one - this PR makes mods tricky, especially when unmaintained.

Garanas:
After thinking about it again - @speed2CZ how does this PR break mods exactly? If they took the regular approach to making the callback then all should be good.

speed2CZ:
If the mod is adding new entry into `self.EventCallbacks` this would break it. So at least that table should stay.

PR #3356, opened 2021-08-15
Factory builders dragun

Dragun123:
This is just an example of how I would do the builder rework by Az/NotSure how to  easily edit the branch he is working on to showcase/Would require for Jip added his improvements

### DIFF
--- a/lua/aibrain.lua
+++ b/lua/aibrain.lua
@@ -1081,9 +1081,18 @@ AIBrain = Class(moho.aibrain_methods) {
     ForceManagerSort = function(self)
         for _, v in self.BuilderManagers do
             v.EngineerManager:SortBuilderList('Any')
-            v.FactoryManager:SortBuilderList('Land')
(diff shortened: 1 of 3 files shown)


Dragun123:
Just to explain how this works. 1) Its backward compatible and while similar in some sense to SCTA code it isn't.
What this does in comparison to Azreal, you can assign a builder so that only a factory of a specific tech level can build it. And unlike Azreal whose code is basically "Is I have TLayerBuilders or I don't". This code allows you to set a Builder LAND so if you want T1 arty to be spammed throughout game at any factory. You can do so by well basically doing as you would do now. 

SCTA Code doesn't do tech level distinctions just looks for categories distinctions (Kbot, Vehicle, Hover). But that isn't the case where I seperated and allow you have only a proper tech level factories build a certain type of unit. Or you can set it so every tech level of factory can attempt to build it.

PR #3358, opened 2021-08-17
Layer optimizations

Garanas:
This caches the GetCurrentLayer() outcome each time when OnLayerChange() is called by the engine. In turn, we save on engine calls while at the same time not breaking compatibility with mods and / or unchanged code.


units/XSB1102/XSB1102_script.lua
```diff
@@ -5,7 +5,7 @@
 --#**
 --#**  Summary  :  Seraphim Hydrocarbon Power Plant Script
 --#**
---#**  Copyright � 2007 Gas Powered Games, Inc.  All rights reserved.
+--#**  Copyright � 2007 Gas Powered Games, Inc.  All rights reserved.
```
Sheikah45, line 8: Ha seems like there might be an unintentional encoding change

Garanas: Ah, you are correct.

engine/Sim/Unit.lua
```diff
@@ -130,7 +130,8 @@ end
 function Unit:GetConsumptionPerSecondMass()
 end
 
---- Return the name of the layer the unit is currently in.
+--- Return the name of the layer the unit is currently in. This value is cached inside
+-- unit.Layer each time a layer changes (when OnLayerChanged is called).
```
KionX, line 134: OnLayerChange

KionX:
There is no inherited call OnLayerChange in defaultunits.lua 2145...

Garanas:
I think you are referring to https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L2225 . You are right, missed that one.

Garanas:
Closed in favor of https://github.com/FAForever/fa/pull/3366 - which is essentially the same but because of my git setup the previous one was on FAForever and this one is coming from a fork. Also easily fixes the encoding problem Sheikah pointed out.

PR #3361, opened 2021-08-17
Experimental network settings in lobby

Crotalus:
Two new settings:

**NetworkLimits**:
`Standard` - same as now
`Experimental` - Doubles the maximum send rates and removes the ack / send delays. Also reduce the resend delays in case a packet is lost.

Reasoning:
Latency is the biggest problem for the game. By shaving off the ack, send, resend delays we gain 25ms in both directions to actually ship a package before the game stutters (500ms with standard net_lag). The ack/send delays is there to be able to ack multiple packages at a time so this tradeoff potentially results in higher send rates. The average bandwidth of people have increased alot since 2007 so hopefully that's not a problem. 

**NetLag**:
`500 ms` - same as now
`Adaptive` - adaptive netlag depending on lobby ping to players

Reasoning:
This setting is more controversial. Reducing net_lag increases the responsiveness of the game while risking to increase the stuttering if someone has a fluctuating ping. Measuring ping in the lobby is far from optimal and currently this setting only measure ping between host -> players. I plan on adding all players -> all players measuring but this is a first test. Lobby ping is never reliable though, ping fluctuations often are the result of bandwidth limits that doesn't happen until game starts.

I had an earlier PR of adaptive netlag(#573) that was 100% adaptive by changing the net_lag value while in game but a concern of desync was raised (haven't been able to desync the game by having diffing net_lag values between players when I test though). By setting constant net_lag at game start we won't have any risk of desync though.

The real gains would be in, say, a 1v1 / 2v2 game where all people have <300ms ping to each other. Then net_lag in this case would be set to something like <ping + 100>ms instead of 500 ms and if the pings are stable the gameplay will feel snappier. This could be incorporated into the autolobby (1v1 ladder etc).

I also raised the maximum adaptive net_lag to 600ms, could help in a game were people have ~500ms ping to eachother
(less stutter but 100ms sluggier command response).

A suggestion could be to let this live in the deploy/fafdevelop branch so we could get some 4v4+ test games going and see if it has any positive impact. A recommendation for anyone testing this is to test one setting at a time, maybe I should even split it up and remove the net_lag bit for now.


Garanas:
This is an interesting PR - quite an alternative approach. It will require quite a bit of testing from players to see whether they enjoy having 600 ms delay one game and 150 ms the next. Overall I am positive and applaud the PR in general.

_A suggestion could be to let this live in the deploy/fafdevelop branch so we could get some 4v4+ test games going and see if it has any positive impact. A recommendation for anyone testing this is to test one setting at a time, maybe I should even split it up and remove the net_lag bit for now._

In early september there will be a patch - I discussed this earlier today with keyser. This means it will take a while before this PR gets merged as I agree that it should first be thoroughly tested on the develop branch before putting it into practice.

How would you test this on your own?

Crotalus:
I tested it locally by using a couple of FA instances and emulate shitty connection with Clumsy (http://jagt.github.io/clumsy/)

Could try to make a sim mod and test some games that way, can't mod the lobby though but could instead add the network changes into gamemain.lua and then use some setting in lobby as the flag.

Crotalus:
I made a sim mod now "netboost" that start with the same network settings as experimental here. net_lag is 500 at start though.

With that mod it's also possible to change network settings  and propagate them to all players.

Example:
In console you do "net_lag 200" and press Ctrl-Shift Backspace, the setting will then sync between all players.

Sheikah45:
I think the idea of adaptive net lag is interesting in terms of making the game feel snappy. However I do wonder as Jip mentioned how much players have unconsciously adapted to the 500ms lag and if having varying lags in different games would throw people off.

Garanas:
I'm all up for testing and seeing how people react. I'll make sure to run a Setons later today / tomorrow and have this mod on. Does everyone need the mod, or can I influence the entire game by myself?

Crotalus:
Everyone will have the mod activated thanks to it being a sim-mod and downloaded when you join the game lobby.

While in game, in case all players have like 200 ping you could try setting `net_lag 250` in console and press Ctrl-Shift Backspace to sync that to everyone and then see what happens

JaggedAppliance:
This sounds amazing, if you guys throw it onto fafdevelop let me know and i'll try it out.

Garanas:
It appears you do not need fafdevelop for this - Crotalus and I did a test session yesterday. He'll write more about it today.

Tagada14:
This is really interesting and exiting. Having a +-100 ms difference won't have a noticeable effect for 99% of the players in my opinion but I can see this having a hug impact on competitive 1vs1 if we would go eg. from 500 -> ~150 allowing for much more interesting and skillful micro battles. I will make sure to test this.

Garanas:
By rebounding it again - no idea about any other approach.

Garanas:
I am closing this pull request for now. Members of the balance team fear that a significant lower delay would cause quite a shift in balance. And I think they're correct on this one.

Issue #2258, opened 2017-09-23
Missing texture on a tree


speed2CZ:
```INFO: Can't find texture "/env/lava/props/trees/xDead01_s1_normalsTS.dds" -- trying fallback```
These trees are on Haven's Reef (X1MP_004)

The game is complaining but I don't see anything bad with them on the first look.

badbwoi:
that dds is indeed missing but the file is linked in the blueprint.
Each lava tree should have one "..._s1_normalsTS.dds" file, but only one is present.

Dead01_s1_normalsTS.dds (present)
LV_Dead01_s1_normalsTS.dds (missing)
XDead01_s1_normalsTS.dds (missing, but tree only differs in color from "Dead01_s1" and not in shape)

I replaced the missing friles with the "Dead01_s1_normalsTS.dds", but no visible change could be detected.

MrRowey:
![image](https://user-images.githubusercontent.com/20344151/130026191-d0c3e436-19b3-432f-97b0-c55828b859a0.png)
Still and Issue With Original Version and FAF Version of the Map

PR #3363, opened 2021-08-19
Fix missing normals on lava/trees

Garanas:
With the missing normals:
image

Without the missing normals:
image

This PR closes https://github.com/FAForever/fa/issues/2258 - as badbowi suggested there is no visual difference except for missing out on a few warnings in the log.


Garanas:
I am merging this next Monday morning.

PR #3366, opened 2021-08-20
Layer optimizations 

Garanas:
Optimizes the access to the current layer of a unit. Instead of performing an engine call the layer is stored when it is changed inside the unit. All current code in the repository will reference this value instead of performing the engine call.

This does not break compatibility because mods will still perform the engine call.


Garanas:
After testing it a few more times I found no errors and the code is completely backwards compatible.

PR #3375, opened 2021-08-24
Reworking AIUtils, AIAttackUtils, and [Etc]

Azraeel:


lua/AI/aiutilities.lua
```diff
@@ -298,31 +287,39 @@ end
 
 
 function AIGetMarkerLocations(aiBrain, markerType)
+    if ScenarioInfo.Env.Scenario.MasterChain[markerType] then
```
Garanas, line 290: This isn't completely error proof. Imagine playing on the crazy rush map where new extractors are introduced over time.

In another branch I solved it like this, see also ScenarioUtils and the history of the branch.
 - https://github.com/Garanas/fa/commits/marker-refactor

You can then hook the global function that creates deposits to add to the correct cache type when it is called to create a certain resource.

lua/AI/aiutilities.lua
```diff
@@ -298,31 +287,39 @@ end
...
         for k, v in tempMarkers do
             if string.sub(v.Name, 1, 5) == 'ARMY_' then
-                TableInsert(markerList, {Position = v.Position, Name = v.Name})
+                counter = counter + 1
+                markerList[counter] = {Position = v.Position, Name = v.Name}
             end
         end
     else
-        local markers = ScenarioUtils.GetMarkers()
+        local markers = Scenario.MasterChain._MASTERCHAIN_.Markers
```
Garanas, line 305: Best to cache this in an up value to prevent all the table operations.

lua/AI/aiutilities.lua
```diff
@@ -389,33 +386,39 @@ end
 function AIGetMarkerPositionsAroundLocation(aiBrain, markerType, pos, radius, threatMin, threatMax, threatRings, threatType)
     local markers = AIGetMarkersAroundLocation(aiBrain, markerType, pos, radius, threatMin, threatMax, threatRings, threatType)
     local retMarkers = {}
+    local counter = 0
     for _, v in markers do
-        TableInsert(markers, v.Position)
+        counter = counter + 1
```
Garanas, line 391: Better to use 'next' over 'counter' idea - add one after you inserted the value. The reason is that there is now a direct dependency of the table operation over the counter, e.g., the table operation can not be queued in the CPU until the counter has gone through the entire pipeline. See also:
 - http://www.edwardbosworth.com/CPSC5155/PH4/Ch04/Ch04_Piplelining.pdf

And, as practical proof:
 - https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/benchmarking/benchmarks/table-insert.lua#L8

Where the alternative version is still 20% - 30% faster!

lua/AI/aiutilities.lua
```diff
@@ -1531,7 +1534,7 @@ function UseTransports(units, transports, location, transportPlatoon)
 
     local monitorUnits = {}
     for num, data in transportTable do
-        if not table.empty(data.Units) then
+        if not TableEmpty(data.Units) then
```
Garanas, line 1537: Best not to use table.empty or table.getn at all when you just want to know if it is empty, see also this benchmark:
 - https://gitlab.com/supreme-commander-forged-alliance/other/profiler/-/blob/main/mods/profiler/modules/benchmarking/benchmarks/table-empty.lua#L8

Just use if table[1] then ... .

Garanas (changes requested):
Good work overall!

Note that removing a function is dangerous - as maybe it is not being used in the code bases that are available to you.

I think caching the markers in a local cache (similar to the PR I linked you) is better in general. It makes it more clear that this is intended for AI use. 

Regardless, great work and keep it up.

edit: and please keep the logs in as comments, it makes it easier to test for me (and Uveso).

lua/AI/aiutilities.lua
```diff
@@ -159,7 +157,7 @@ function SortLocationsClosestToDefensivePoints(aiBrain, points)
     defPoints = AISortMarkersFromLastPos(aiBrain, defPoints, 5, nil, nil, nil, nil, points[1])
 
     local sortedList = {}
-    for i = 1, table.getn(points) do
+    for i = 1, TableGetn(points) do
```
Garanas, line 160: This function is performing bubble sort. That has an O(n^2) complexity while sorting can be solved in O(n) for integral numbers and O(n*lg2(n)) for real numbers.

Garanas (changes requested):
I'm not sure how often this function is called - but even if it is only a few times a seconds, best to refactor it to use a proper sorting algorithm.

lua/AI/aiutilities.lua
Garanas: (couldn't add a comment where you didn't make a change)

lua/AI/aiutilities.lua
KionX: This is madness.

Garanas: You're right - and in this case they are.

Garanas:
After discussing it with Azreal, I am closing this pull request due to it being aged. It can be re-opened in the future.

PR #3378, opened 2021-08-25
Adjust LAB hitboxes

Tagada14:
Adjust LAB hitboxes so that they can be hit more easily when un micro'ed by T1 tanks and other LABs.

### DIFF
--- a/units/URL0106/URL0106_unit.bp
+++ b/units/URL0106/URL0106_unit.bp
@@ -180,9 +180,9 @@ UnitBlueprint {
     SelectionSizeX = 0.3,
     SelectionSizeZ = 0.3,
     SelectionThickness = 1.2,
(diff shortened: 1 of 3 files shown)


units/URL0106/URL0106_unit.bp
```diff
@@ -180,9 +180,9 @@ UnitBlueprint {
     SelectionSizeX = 0.3,
     SelectionSizeZ = 0.3,
     SelectionThickness = 1.2,
-    SizeX = 0.6,
-    SizeY = 0.6,
-    SizeZ = 0.6,
+    SizeX = 0.7, #from 0.6
+    SizeY = 0.7, #from 0.6
```
speed2CZ, line 184: You shouldnt be leaving comments like this. That's what the git history is for.

PR #3380, opened 2021-08-26
Jester fix

Tagada14:
Remove the AoE, give the projectile slight tracking.
This allows Jester to hit units properly.


Petricpwnz:
Agree on the soft homing proposal, the aoe stuff was nice in that it uncovered some interesting game innerworkings :D but it only made sense in its practically nonexistent radius state where it didn't impact the balance much

shalkya:
just to add on top of that, jester can actually do friendlyfire with the aoe on

projectiles/CDFLaserHeavy01/CDFLaserHeavy01_proj.bp
```diff
@@ -0,0 +1,54 @@
...
+    },
+    Interface = {
+        HelpText = 0,
+    },
+    Physics = {
+        DestroyOnWater = true,
+        InitialSpeed = 22,
+        Lifetime = 2.5,
+        UseGravity = true,
+        TrackTarget = true,
```
Garanas, line 49: For future file additions - add the file, stage and commit it, then make further changes. It makes it easier to see what you changed in comparison to the base file.

PR #3385, opened 2021-08-28
Check for game options on the command line

Askaholic:
This is not really my ideal solution but it would work if I don't come up with anything better. With these changes you can set the fullshare option with the args `/Share FullShare`. 

Ideally it would work like this:
`/GameOption Share FullShare /GameOption Victory Demoralization /GameOption ...`
But that might be hard / impossible to implement given the way `GetCommandLineArg` works


Garanas:
I would suggest using key-value pairs as arguments. As an example: `/gameoptions share:fullshare victory:demoralization`.

The function `GetCommandLineArg` returns the number of parameters of an argument until it runs out of parameters, then it returns false. You can write a loop that stops when it hits false, uses the previous result and returns all the parameters (key-value pairs). A bit of parsing needs to happen to get the values. This would make it quite flexible, as it will even allow you to set options of mods, maps and other random things the server deems important to let the game now in the future. 

The map options are added right before launch, see also https://github.com/FAForever/fa/pull/3138 .

BlackYps:
I don't understand what we would input as num in `GetCommandLineArg(arg, num)`. I found no definition, probably because it is an engine function. It seems that num specifies how many parameters it reads and returns, but what do we do, if they are unknown?

Garanas:
I think this could work:

```lua
local next = 1
while GetCommandLineArg("gameoptions", next) do 
  next = next + 1
end

local args = GetCommandLineArg("gameoptions", next - 1)
```

At some point GetCommandLineArg runs out of arguments (next is too high) and then we take the previous entry. This would make it dynamic for any number of arguments and any key / value pair we wish to inject into the game options.

BlackYps:
I haven't tested this yet, but I think the idea becomes clear:
```lua
local next = 1
while GetCommandLineArg("gameoptions", next) do 
    next = next + 1
end

local args = GetCommandLineArg("gameoptions", next - 1)

for arg in args do
    local option, local value = string:match("([^:]+):([^:]+)")
    if option == nil or value == nil then
        LOG("Malformed gameoption. ignoring...")
    else
        gameInfo.GameOptions[option] = value
    end
end
```
I don't really know what string options are available with the supcom lua version.

Garanas:
Something like that would work - I'd need to check the string options myself too.

Askaholic:
It would make sense to me to put that into a helper function so it can be used elsewhere. What file would that go into?

Garanas:
I honestly don't know - anyone else has suggestions? I suspect making a new utils file for this type of thing doesn't particularly harm.

lua/ui/lobby/autolobby.lua
```diff
@@ -364,6 +364,21 @@ function HostGame(gameName, scenarioFileName, singlePlayer)
...
 
+    local next = 1
+    while GetCommandLineArg("gameoptions", next) do
+        next = next + 1
+    end
+
+    local args = GetCommandLineArg("gameoptions", next - 1)
+
+    for arg in args do
+        local option, value = string.match(arg, "([^:]+):([^:]+)")
```
Askaholic, line 375: Problem is, I don't think `string.match` exists in GPG's version of lua

Askaholic:
Now another problem we have is that if someone who doesn't have an updated client becomes the host and doesn't pass the game options to the game, then the game will be played using wrong game settings. Is there some way we can have each of the joining players verify the game options as well and refuse to launch the game if the game options don't agree to what they're expecting?

lua/ui/lobby/autolobby.lua
```diff
@@ -364,6 +365,24 @@ function HostGame(gameName, scenarioFileName, singlePlayer)
...
 
+    local next = 1
+    local args, nextArgs = nil, nil
+    repeat
+        nextArgs, args = GetCommandLineArg("/gameoptions", next), nextArgs
+        next = next + 1
+    until not nextArgs
+
+    if args then
+        for _, arg in args do
```
1-alex98, line 376:
```suggestion
        for _, arg in nextArgs do
```

1-alex98 (changes requested):

BlackYps:
I don't know if that is a good solution, but we could set the required client version to be that version, so everybody with an older client is forced to update to be able to log in.

Askaholic:
We should definitely do that too, but that’s not even going to get everyone since it won’t  hit py client users and those random people who are still on the tmm beta test client or other weird ass builds of the client. I’d rather that games for those people become completely unplayable so that they are forced to update.

KionX:
_I’d rather that games for those people become completely unplayable so that they are forced to update._
Pass the client version to the game and check it.
Or compare the settings of the players:
https://github.com/FAForever/fa/blob/7542596ea1c16be28615fdc40cc9bf9c8bb5481a/lua/ui/lobby/autolobby.lua#L269

BlackYps:
I tested this some more toady and the code works as expected. I.e. gameoptions from the db get applied, a ladder queue without any custom options launches with default option, and old that don't relay the options still make the game launch with default options if they are the host.
The tests were overshadowed by a faction selection bug, but that is because the client on the develop branch interprets the server commands wrong and that needs to be fixed either in the client or the server.

BlackYps:
Regarding the client version check: I'd rather have the game compare the player settings than the client version. The settings are the actual important part that needs to match and later we could have different client versions that all support relaying the options correctly.

Askaholic:
@KionX thanks, I think I mostly want to compare the game options so I guess I’d do that here:

https://github.com/FAForever/fa/blob/7542596ea1c16be28615fdc40cc9bf9c8bb5481a/lua/ui/lobby/autolobby.lua#L284

lua/system/utils.lua
```diff
@@ -538,7 +538,7 @@ end
 function StringSplit(str, sep)
     local sep, fields = sep or ":", {}
     local pattern = string.format("([^%s]+)", sep)
-    str:gsub(pattern, function(c) fields[table.getn(fields)+1] = c end)
+    string.gsub(str, pattern, function(c) fields[table.getn(fields)+1] = c end)
```
Askaholic, line 541: Need to check if this change is actually needed, because I'm pretty sure its not.

Garanas:
Should this be included in the hotfix too?

Askaholic:
Needs to be tested first, but once it’s ready it can go out whenever.

lua/system/utils.lua
BlackYps: Testing showed that this is not needed.

lua/ui/lobby/autolobby.lua
```diff
@@ -398,3 +448,40 @@ function DisconnectFromPeer(uid)
...
+            LOG("Malformed gameoption. ignoring...")
+        end
+    end
+end
+
+-- Return a table parsed from key:value pairs passed on the command line
+-- Example:
+--  command line args: /arg key1:value1 key2:value2
+--  GetCommandLineArgTable("/arg") -> {key1="value1", key2="value2"}
+function GetCommandLineArgTable(arg)
```
Askaholic, line 467: Should this move to utils?

lua/ui/lobby/autolobby.lua
```diff
@@ -173,24 +206,35 @@ local function CheckForLaunch()
...
-end
 
+    ForkThread(function()
+        if WaitLaunchAccepted() then
+            lobbyComm:LaunchGame(gameInfo)
+            return
+        end
 
+        LOG("Some players rejected the launch! " .. repr(peerLaunchStatuses))
+        SetDialog(Group(parent, "controlGroup"), Strings.LaunchRejected, "<LOC _Exit>", CleanupAndExit)
```
Askaholic, line 228: The UI doesn't seem to come up here. It just spams 'LazyVar' errors in the log

lua/ui/lobby/autolobby.lua
```diff
@@ -173,106 +206,97 @@ local function CheckForLaunch()
...
-
-    ---------------------------------------------------------------------------
-    -- set up map panel
-    ---------------------------------------------------------------------------
-    local controlGroup = Group(parent, "controlGroup")
-    LayoutHelpers.AtCenterIn(controlGroup, parent)
-    LayoutHelpers.SetDimensions(controlGroup, 970, 670)
 
-    UIUtil.ShowInfoDialog(controlGroup, "<LOC lobui_0201>Setting up automatch...", "<LOC _Cancel>", ExitApplication)
+    SetDialog(parent, "<LOC lobui_0201>Setting up automatch...", "<LOC _Cancel>", ExitApplication)
```
Askaholic, line 245: What are the implications of attaching stuff to the parent directly instead of using this `controlGroup`? As far as I can see the result looks to be the same, however because the code in `InitLobbyComm` didn't call the LayoutHelpers as was done here, it was throwing a bunch of errors that prevented the dialog from being rendered properly. Instead only the message text would be added at the top right of the screen.

Garanas: All UI elements need to have certain values set or it doesn't know where to render it. For example: you can set Left, Top, Width and Height. If you'd skip Top in that list then it'd try to determine the height using Bottom, which isn't defined, then it tries with Top, which isn't defined - etc. That is the cyclic dependency warning - and it is annoying. This is most relevant for images - text appears to allows spawn in the top left (right?) of the screen.

lua/ui/lobby/autolobby.lua
Garanas: That sounds good, if you do so make sure to make the split value (':') a parameter of the function.

lua/ui/lobby/autolobby.lua
Askaholic: So what does the `controlGroup` do?

lua/ui/lobby/autolobby.lua
Askaholic: I kindof intentionally made it not a parameter so that all usages of it would be consistent. It doesn't sound like a great idea to me to use i.e. `key:value` for `/gameoptions` but then use `key|value` for `/somethingelse`. That's just going to be confusing and annoying to deal with in the client.

lua/ui/lobby/autolobby.lua
Garanas: In this case not much - it is just a container of UI elements such that any UI element attached to it receive the same signals are the control group does. As an example: if you destroy the controlGroup then all children are destroyed too. Or, if you hide / show the controlGroup then all children are hidden / shown too.

Garanas: A more practical example is [this](https://github.com/FAForever/fa/pull/3351/files#diff-d26022466a15cb8b0e34ecd6f22bf7902fb8050ee17bfedae18ebc206ad422bdR14) or [this](https://github.com/FAForever/fa/pull/3351/files#diff-45e96a5a8a17a1e521d6f8f75df6ca5e26b25957347230680d3f57f7174a5327R28), where we hide / show the entire UI element (button / tab bar) and all its sub elements.

lua/ui/lobby/autolobby.lua
```diff
@@ -173,106 +206,93 @@ local function CheckForLaunch()
...
     end
 
     lobbyComm.Ejected = function(self, reason)
         LOG("EJECTED " .. reason)
-        if connectingDialog then
-            connectingDialog:Destroy()
-        end
-
-        local failedDlg = UIUtil.ShowInfoDialog(controlGroup, Strings.Ejected, CleanupAndExit)
+        SetDialog(parent, Strings.Ejected, "<LOC _OK>", CleanupAndExit)
```
Askaholic, line 274: Also just want to make a note of this since it's not obvious from the git diff. I added the "OK" button here since all of the other dialogs had it.

lua/ui/lobby/autolobby.lua
```diff
@@ -281,7 +301,23 @@ local function InitLobbyComm(protocol, localPort, desiredPlayerName, localPlayer
             # Non-Host Messages
             if data.Type == 'Launch' then
                 LOG(repr(data.GameInfo))
-                lobbyComm:LaunchGame(data.GameInfo)
+                hostOptions = table.copy(data.GameInfo.GameOptions)
```
Askaholic, line 304: Need to make this local... Oops, python habits :P

lua/ui/lobby/autolobby.lua
```diff
@@ -398,3 +435,14 @@ function DisconnectFromPeer(uid)
     GpgNetSend('Disconnected', string.format("%d", uid))
     lobbyComm:DisconnectFromPeer(uid)
 end
+
+
+function SetGameOptionsFromCommandLine()
+    for name, value in utils.GetCommandLineArgTable("/gameoptions") do
```
Askaholic, line 441: Have not tested this yet since refactoring GetCommandLineArgTable to utils

lua/ui/lobby/autolobby.lua
Askaholic: This was happening because the control group didn't have a size and location set.

Garanas (changes requested):
Overall I think the PR is good.

However, due to the nature autolobby.lua, we'll likely not touch them again for a long time. And by then, new people may be looking at these files and we may not be available anymore to help them out. Therefore I suggest adding comments to autolobby.lua:
 - [x] Add a description to each function, describing its intent and the parameters.
 - [ ] Add a general description to this file as to how the client and the server communicate with the game in general. This doesn't have to be extensive: the aim is to give future maintainers a direction to look into to understand the context of this file. As an example: how are the parameters chosen, if we'd need to edit that where do we do that?

Last but not least: is using the word 'client' in one of the error messages ambiguous? I am afraid people may interpret as their game not being up to date, instead of the Downlord FAF client.

PR #3387, opened 2021-08-28
Optimize marker usage

Garanas:
Specifically it allows AI developers / map scripters to work with markers without having to worrying about underlying inefficiences. As an example, retreiving the mass markers on a map is a common operation for AIs. If done through the base game code it would re-allocate a new table of markers each time an AI condition manager starts checking the state of the game. That is quite wasteful. This file keeps track of previous calls, caching the result.

Supports crazyrush-like maps.

Things to add:
 - [ ] Support for underwater mexes


Garanas:
Once this PR is merged all that remains is slowly replacing the AIs in the base game and in mods to use the new functionality. Please do mention if any useful functionality is missing - I'll add it accordingly.

Garanas:
Example of the pull request in action:
 - https://www.youtube.com/watch?v=yBwMmbYpIBs

PR #3389, opened 2021-08-29
decals friendlyfire re-fix

shalkya:
removed the if statement


Garanas (approved):
Tested in game - allied shields no longer receive damage by the DamageArea.

PR #3392, opened 2021-08-29
Optimize projectiles

Garanas:
This PR is an example of the performance / stability increase we can get when we optimize projectiles according to the benchmarks. Up to this point the following units / projectiles are properly optimized:
 - Zthuee (T1 Seraphim Mobile Artillery)

When having 400 Zthuee idling we spent 3 ms / tick on both branches. When we let them ground fire, the results are:
 - Current default FAF branch: 11 / 13 ms -> 8 / 10 ms for firing
 - This branch: 8  / 9 ms -> 5 / 6 ms for firing

Disclaimer: _make sure you are _completely_ zoomed out when checking the performance of the sim. The amount of effects Zthuee spawn do effect the sim and the results if they get rendered._

This shows that there is a significant gain to be had for applying the benchmark optimizations to projectiles in general. This include:
 - No empty and / or unnecessary table allocations
 - [Upvalue moho functions](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/benchmarking/benchmarks/function-origin.lua)
 - [Pre-allocating table elements](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/benchmarking/benchmarks/table-hash.lua)
 - [Inlining of functions](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/benchmarking/benchmarks/function-inlining.lua)

I intent this PR to also form a basis for a discussion as to whether creating a second, optimized hierarchy can allow us to increase the performance of the game without breaking mod compatibility. The structure makes a copy of each file part of the hierarchy with 'Opti'  attached. As an example for the `SIFThunthoArtilleryShell01_script.lua` projectile script:
 - `SIFThunthoArtilleryShell01_script`: inherits `SeraphimProjectilesOpti.SThunthoArtilleryShell`
 - `SThunthoArtilleryShell`: inherits `DefaultProjectilesOpti.MultiPolyTrailProjectile`
 - `MultiPolyTrailProjectile`: inherits `DefaultProjectilesOpti.EmitterProjectile`
 - `EmitterProjectile`: inherits `ProjectileOpti.Projectile`

The original hierarchy is still completely intact and untouched. The file just inherits from a different, more optimized base class. And the same applies for that parent, and the ancestor, and the grand ancestors - etc. Essentially creating a new, optimized hierarchy next to the existing hierarchy.

The question is: if we'd apply this to weapons and units too (creating a 2nd, optimized hierarchy) - up to what extent would this break mods?
 - Mods that create new units?
 - Mods that adjust existing units through soft-hooking functions?
 - Mods that adjusts existing units through blueprint merging?

Note that we're not creating a 2nd unit / weapon / projectile: these change the classes they inherit from instead. This is to ensure the unit name / blueprint value does not change and so that we do not get duplicated units.

One situation where I can expect something to break is when a mod tries to soft-hook an existing unit. But I'm not sure how common this is - any idea of mods that would do this?

### DIFF
--- a/lua/SeraphimProjectilesOpti.lua
+++ b/lua/SeraphimProjectilesOpti.lua
@@ -1,0 +1,52 @@
+------------------------------------------------------------
+--
+--  File     :  /cdimage/lua/seraphimprojectiles.lua
+--  Author(s):  Gordon Duclos, Greg Kohne, Matt Vainio, Aaron Lundquist
+--
+--  Summary  : Seraphim projectile base class definitions
+--
+--  Copyright � 2007 Gas Powered Games, Inc.  All rights reserved.
+------------------------------------------------------------
+
+--------------------------------------------------------------------------
+--  SERAPHIM PROJECTILES SCRIPTS
+--------------------------------------------------------------------------
+
+local EffectTemplate = import('/lua/EffectTemplates.lua')
+local DefaultProjectileFile = import('/lua/sim/defaultprojectiles.lua')
+local MultiPolyTrailProjectile = DefaultProjectileFile.MultiPolyTrailProjectileOpti
+
+SThunthoArtilleryShell = Class(MultiPolyTrailProjectile) {
+
+    FxImpactTrajectoryAligned = false,
+
+    FxImpactLand = EffectTemplate.SThunderStormCannonHit,
+    FxImpactNone = EffectTemplate.SThunderStormCannonHit,
+    FxImpactProjectile = false,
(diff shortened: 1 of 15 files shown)


Garanas:
Closing this PR, conclusion: creating a separate hierarchy next to the original will cause a lot more issues than it solves:
 - Mods that hook into the original hierarchy somehow fail silently.
 - Mods that hook into units and expect the original hierarchy fail.

It is better to just update the current hierarchy instead, taking into account that it will break mods and prevent that as much as we can.

Issue #3374, opened 2021-08-24
Add custom vault path support


Garanas:
This issue is a consequence of #3373 .

The PR changes the init_faf.lua to allow for a custom vault path, but some of the featured mods and the develop init file are not updated. Those require changes too in order for the client feature to be stable.

 - [ ] init_fafdevelop.lua
 - [ ] init_nomads.lua
 - [ ] init_nonxt.lua
 - [ ] init_phantomx.lua
 - [ ] init_tutorials.lua

Sadly, 4 / 5 of these are not part of this repository.

Garanas:
Closing this issue as this is supported.

PR #3396, opened 2021-08-30
Synchronization init files

KionX:
Also modules.nx was removed because
https://github.com/FAForever/fa/search?q=modules&type=commits

We should do something about the copy-paste.
I suggest putting the general part in init_common.lua
The problem is only with patches...


Garanas:
Similar issue to: https://github.com/FAForever/fa/issues/3374

And yes - we do not have direct access to the other init files.

Garanas:
I'm looking into this PR this week - I think we can sneeze it into the patch.

PR #3397, opened 2021-08-31
Fixed Seraphim Rambo SACU buildmode keys

slinkingant:
When using the Seraphim SACU in buildmode, the hotkeys are not working.

The unit was already present in buildmodedata.lua, but a (most likely) copy/paste error in the unit-name prevented it from working.


Garanas:
I'll test it later this week - I think we can squeeze this one into the patch. Thank you for making the PR!

Garanas:
Tested and works as described.

PR #3398, opened 2021-08-31
Improves the performance of aeon build animations and fixes a bug

Garanas:
Improves the performance by applying the following optimizations: 
 - [x] Upvalue global functions
 - [x] Prevent allocation of tables
 - [x] Prevent allocation of entities
 - [x] Use dummy projectiles instead of full-fledged projectiles

Also fixes a critical bug in the game where the aim bones of buildings are underground when a building is being made. This is at the expense of losing the build shader progressing properly. Note that the pool and the effects accompanying it are still there.


Garanas:
There's an interesting bug that I had with this PR: when pausing / unpausing the game while in `cam_Free` mode it could hard-crash. It doesn't always happen, and with the latest version I can no longer reproduce it. But I don't know why 😅  .

Therefore I am asking for some other contributors to:
 - (1) Pull in this branch (gh pr checkout 3398).
 - (2) Start a FAF game in the client to make sure you have the release executable.
 - (3) Launch the game with this PR.
 - (4) Pause / unpause a lot for a minute, then turn on `cam_Free` and try it again.
 - (5) Close the game, start a FAFDevelop game in the client to make sure you have the develop executable.
 - (6) Pause / unpause a lot for a minute, then turn on `cam_Free` and try it again.

Report back whether the game crashes for you. In all cases if it hard-crashed for me, it hard-crashed with this error:

image

I've also experienced the same _a few times_ on the FAFDevelop branch of the client, I was not able to reproduce it on the FAF branch of the client.

KionX:
local EntityDestroy = moho.entity_methods.Destroy
Is it definitely faster than self.Destroy?

Garanas:
Is that the cause of the error?

And yes - it has a 8% - 10% increase in performance.

KionX:
Couldn't reproduce.
I want to take a look at yours exception message or  .dmp

KionX:
Crash here:
https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/controls/worldview.lua#L188

Garanas:
So it has nothing to do with the changes in this PR?

baurse:
Couldn't reproduce it either, regardless of executable.

KionX:
I need more .dmp

Garanas:
Let me find a few more for you.

KionX:
I managed to reproduce. This is not your PR. The problem is the invalid world mouse position.

Garanas:
Thank you for looking into it

lua/EffectUtilitiesOpti.lua
```diff
@@ -251,4 +273,284 @@ function CreateCybranEngineerBuildEffects(builder, buildBones, buildBots, total)
...
+    local entity = Entity()
+    TrashBagAdd(effectsBag, entity)
+
+    Warp(entity, o)
+    EntitySetOrientation(entity, orientation, true)
+    EntitySetScale(entity, display.UniformScale)
+    EntitySetMesh(entity, display.BuildMeshBlueprint, true)
+
+    -- make sure enemies don't get to see them
+    EntitySetVizToAllies(entity, 'Intel')
```
Crotalus, line 326: Can't remember 100% how EntitySetVizAllies works but should this be set to 'Always' instead of 'Intel'?

lua/EffectUtilitiesOpti.lua
```diff
@@ -251,4 +273,284 @@ function CreateCybranEngineerBuildEffects(builder, buildBones, buildBots, total)
...
+    vc[2] = oy - (2 * sy)
+    vc[3] = oz 
+    Warp(entity, vc)
+    EntitySetOrientation(entity, orientation, true)
+    EntitySetScale(entity, display.UniformScale)
+    EntitySetMesh(entity, display.BuildMeshBlueprint, true)
+
+    -- make sure enemies don't get to see them
+    EntitySetVizToEnemies(entity, 'Intel')
+    EntitySetVizToAllies(entity, 'Intel')
```
Crotalus, line 474: Can't remember 100% how EntitySetVizAllies works but should this be set to 'Always' instead of 'Intel'?

lua/EffectUtilitiesOpti.lua
```diff
@@ -251,4 +273,284 @@ function CreateCybranEngineerBuildEffects(builder, buildBones, buildBots, total)
...
+                vc[3] = oz
+                Warp(entity, vc)
+            end
+
+            -- wait a tick
+            WaitTicks(2)
+        end
+
+        -- show the actual unit
+        UnitShowBone(unitBeingBuilt, 0, true)
```
Crotalus, line 554: Do we need to check if unit is dead here? I.e:

if not unitBeingBuilt.Dead then
    UnitShowBone(unitBeingBuilt, 0, true)
end

Crotalus (approved):
Ran some testing of the PR and couldn't provoke any error. Build effects seems to be working fine also

lua/EffectUtilitiesOpti.lua
Garanas: Yes, I think you're right.

lua/EffectUtilitiesOpti.lua
Garanas: I think you're right here too. However, the entity has no army set. Therefore I do not know who its allies would be. I tried setting it but I did not succeed in doing so.

Garanas:
Last fix is for the orientation of the pool for the air factory.

MrRowey:
Dose this fix finaly make use of the GC build animation  where its down on one knee  and stands up after construction?

Garanas:
Yes it does!

image

Including one made by Marlo (on the left).

MrRowey:
nice i always wondered why the GC never really had a build animation that would make it more in line with like the mega/ML have like a start-up sequence when the unit it completed

PR #3403, opened 2021-09-05
Optimize default explosions

Garanas:
Optimizes how explosions when units die work in general. For each explosion it adjusts:
 - [x] Prevent unnecessary function call sequences that add little value.
 - [x] Prevent unnecessary table creations that add little value.
 - [x] Prevent unnecessary thread being made.
 - [x] Prevent unnecessary entity creation that was quite literally destroyed in the same tick.

And it optimizes the debris of units. For each debris element:
 - [x] Inherits from a dummy projectile, instead of the standard projectile. This does not alter its functionality (for the purpose of being debris).
 - [x] Prevent dozens of tables entries and / or tables from being made because of it.
 - [x] Prevents a lot of logic that make no sense for debris to run, like checking whether it should do damage.

Next to that, it adds:
 - [x] Comments on how the functions (do not) work in `/lua/defaultexplosions.lua` 

And meanwhile I found this issue:
image

Those functions return `bp.SizeX * 0.5` or `bp.SizeX * 0.333` instead of doing the computation with fallbacks. This means that _all_ explosions used solely the x parameter to determine the amount of debris or the size of effects, instead of the radius as a whole. The irony 😄 .

Aesthetically the explosions / debris should not have changed.

What still needs to be done:
 - [ ] With the corrected logic, all the magic values should be adjusted accordingly.


Garanas:
Overall these changes will help out the garbage collector a lot during fights, as dozens of units die and tons of debris is being made - all easier on memory usage now.

KionX:
_((bp.SizeX or 0 + bp.SizeY or 0 + bp.SizeZ or 0) * 0.333)_
"or 0" in my opinion, they are not needed at all.

Garanas:
They are - as the value may not be set. But the issue is that the first 'or' takes the entire expression on the right (including the additions!) as it second argument. Therefore the entire sequence returns just bp.SizeX * 0.5 or bp.SizeX * 0.333.

I've left these unchanged (but did change it accordingly in the inlined code) because mods may depend on this particular behavior.

KionX:
_They are - as the value may not be set._
This means it is necessary to correct the blueprints.
Replacing "bp. sizeX or 0" with "bp.sizeX" everywhere will be a good micro-optimization.

The-Balthazar:
I posted this on github, but I'll also post it here,

Default size values are 1 not 0. In a situation where a unit doesn't have SizeX defined it will fallback to the other size values if they are defined but x isn't, but with a sequential line of pointless and inaccurate 0+'s preceding them for each that wasn't defined. Also if they weren't missing the brackets around each value, and they weren't using the wrong fallback value, it would still be outputting the diameter, not the radius.

The size values are used as the diameters, not radii, when set by the engine. The function that lets you change collision boxes is bugged and uses the input values as radius instead, which is why all instances of that have the size * 0.5. (Except the Salem, which for whatever reason was using the size scale as the multiplier, which is wrong but no one noticed because the scale of the unit was 0.48). This is probably where that off by *0.5 confusion came from.

In short, the functions are more wrong than you thought.

They should be:
```lua
return (bp.SizeX or 1) + (bp.SizeZ or 1) * 0.25
```
and
```lua
return (bp.SizeX or 1) + (bp.SizeZ or 1) + (bp.SizeY or 1) * 0.166
```
If you actually want the bounding *radius* averages.

Garanas:
@KionX You're right, I made a benchmark for it:

image

Having no `or` at all is almost twice as good - but the impact of the or in general is extremely small. When running ten million times (10.000.000) it takes ~70 ms, where as other tests run on average a hundred thousand times (100.000) and take about the same amount of time. Considering changing this may suddenly break mods I think we should leave it out in this case.

@The-Balthazar It is even more messed up then. I'm not sure if we should change it, I'll at least make a comment about how broken it is based on your post.

Garanas:
The heading of a unit was not taken into account. For non-square units this caused debris to spawn out of thin air.

image

A quick benchmark when comparing with the FAF branch:
 - 400 t1 land factories: spikes sim from ~150ms -> ~50ms.
 - 400 titans: spikes sim from ~85ms -> ~35 ms.

Overall the game stutters a lot less when the explosions happen - I suspect this is because of all the garbage that is prevented by using the dummy projectile.

Garanas:
The last change prevents debris from participating in the sim. In turn, the debris takes up almost nothing. In comparison with the FAF branch:
 - 400 t1 land factories: spikes sim from ~150ms -> ~23ms.
 - 400 titans: spikes sim from ~85ms -> ~7 ms.

The time consumption that remains is purely from the effects that spawn. What are people's opinion about it, that debris can no longer collide with entities (units / props / shields)?

Crotalus:
I rather take a faster sim speed than watching some debris bounce off a unit / shield. When it matters (late game) no one sees 99.9% of the debris anyway.

Garanas:
I think this PR has more an impact than I realized. I just tested it on my laptop and with 400 titans it deep dives to 200ms for a single tick, where as with this PR it stays steadily at about 10 - 12 ms.

Garanas:
From a functional perspective this PR is ready for a review.

It still needs some tweaking with the size of the explosions, but that is just changing magic numbers.

lua/sim/DefaultProjectiles.lua
```diff
@@ -312,6 +333,133 @@ BaseGenericDebris = Class(EmitterProjectile){
...
+        -- Possible targetType values are:
+        --  'Unit', 'Terrain', 'Water', 'Air', 'Prop'
+        --  'Shield', 'UnitAir', 'UnderWater', 'UnitUnderwater'
+        --  'Projectile', 'ProjectileUnderWater
+
+        local impactSnd = false
+        local impactEffects = false
+        local impactEffectscale = 1
+
+        if targetType == 'Water' then
```
Crotalus, line 355: Some micro-optimization could be to reorder these if-statements to the most likely targetType, I would guess on most maps this would be Terrain instead of Water

Also, maybe initialize the values to the most common one
```
impactSnd = "Impact"
impactEffects = false
impactEffectscale = 1.0
```

and then just change the variables for the ones that differ

Garanas: Great suggestion - a lot of the if statement can essentially be cut out.

PR #3410, opened 2021-09-10
Removed duplicate test for sender

### LINKED ISSUE #3294: SendTaunt
https://github.com/FAForever/fa/blob/0c63ff40aee2a0f4fc19a1112877c423232d6c86/lua/ui/game/taunt.lua#L160-L164
SendTaunt has its own condion for sender in it,no check needed

Benzi-Junior:
fixes #3294 
A check for the sender variable being performed before calling the "sendTaunt" function which itself then checks for the variable.


Garanas:
In what scenario would the sender not be set?

Benzi-Junior:
That is a good question, I'm guessing this is to handle a case where an additional mod or a custom map sends taunts

Garanas:
Yes, I can see that happening. I'll see if I can test it through a map script later this morning.

Garanas:
The sender is not set by default, but used by Sorian AI to determine what AI send the taunt. All is good!

PR #3413, opened 2021-09-10
Added show MuzzleBones

KionX:
https://faforever.zulipchat.com/#narrow/stream/203495-game-development/topic/Atlantis/near/252656810


Garanas:
@KionX Am I correct that this change effects all units?
@Tagada14 Are the calculations correct? I suspect this is related to #3412. I have no experience with these type of things.

KionX:
_Am I correct that this change effects all units?_
Yes

PR #3414, opened 2021-09-10
Attempt to optimizing ASF

Garanas:
Attempts to optimizing the ASF in-game.

### DIFF
--- a/lua/system/trashbag.lua
+++ b/lua/system/trashbag.lua
@@ -69,5 +69,8 @@ TrashBag = Class {
                 self[k] = nil
             end
         end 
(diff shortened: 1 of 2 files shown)


Garanas:
This branch has an issue at the moment.

I am trying to optimize the Corona (T3 Aeon ASF) but the effects / sounds just stop working halfway. Except when I have another ASF with similar effects, such as a Wasp (T3 UEF ASF). When there is another ASF then the effects work partially: they jump in and out of view. 

My question is whether someone can reproduce it and whether someone knows what is causing it.

PR #3415, opened 2021-09-10
Adjust hitboxes and target bones for T1 PDs

Tagada14:
Adjust the hitboxes of T1 PDs - make it the same size for all 4 PDs.
This will make the biggest difference for Aeon PD's.
All of the PD's will now be similar in effectiveness to old Cybran/UEF ones.


Tagada14:
I will wait for MadMax to adjust the Aeon T1 PD model and bones, then it will be all good.

PR #3417, opened 2021-09-13
Add unit tests for string utils

Askaholic:
I found a single file test framework that was pretty easy to convert to lua 5.0 syntax. With the help of the `repr` function it gives fairly useful test output on failures too. I had to change all of the `str:foo` calls to `string.foo(str)` because by default in lua 5.0 you must invoke them like that. But, I see the `str:foo` all over the codebase so I feel like GPG must have added it to SupCom (which means I'll need to add it to faflua and then probably revert all those changes in this PR).

Here's an example of what the console output looks like:
image

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -5,7 +5,7 @@ on: [push, pull_request]
 jobs:
   test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


Garanas:
I'll be reviewing this during the weekend and then likely merging on Monday.

PR #3360, opened 2021-08-17
Disable ping buttons for defeated players

KionX:
Fixed https://github.com/FAForever/fa/issues/3359

PR #3418, opened 2021-09-13
Refactor ping-preventing code

Garanas:
This new solution allows you to ping as long as:
 - You are not an observer.
 - Your current focus army is an ally to you.
 - You are not in a replay.

It also ensures that the name of the original author is displayed under the ping. This can be evaded however - in the end we have no full control over the UI code and someone can write a UI mod to prevent this basic check.

 - [x] Disable ping buttons intuitively or communicate that you can not ping when your focus army is not allied to you.


lua/ui/game/gamemain.lua
```diff
@@ -136,6 +136,11 @@ function OnFirstUpdate()
 end
 
 function CreateUI(isReplay)
+
+    -- TODO: is this the best place? I tried that userInit.lua but because of the preloading that starts loading during the lobby and it crashes.
```
Garanas, line 140: Please make suggestions on this todo.

KionX:
It is better to make a revert for #3360 and #3240

Garanas:
As in, revert through Git?

I have no experience with that @shalkya could you take care of that?

Garanas:
Woops - did something wrong.

Garanas:
Ready to be squashed 👍

KionX:
Tested it. Works.

I have an idea.
After death, do not automatically switch to the observer.
Prohibit ping if you were an observer.

PR #3420, opened 2021-09-15
Unit fixes

### LINKED ISSUE #3400: Aeon Temple civilian structure missing model for zoom level
The Aeon Temple is missing a model for the low-detail zoom level.
High deatil:
image
Low detail:
image

Garanas:
Contains a few small fixes for various units.

### Cybran Tech 1 Artillery

Only the barrel moves up / down instead of the entire arm.

t1arty_og
_old_

t1arty_new
_new_

### Aeon temple

Has a proper LOD so that it doesn't just flat out disappears. Closes #3400 

image

With thanks to Madmax for his work.


InspektorKotFAF:
Honestly, old animation looks better. On the new it looks almost without recoil

Garanas:
Look closely at the old animation - it looks strange / off from up close. We can tweak the new one slightly to make it look like it has more recoil.

InspektorKotFAF:
> Look closely at the old animation - it looks strange / off from up close. It even intersects its own mesh near the hinge. We can tweak the new one slightly to make it look like it has more recoil.

Ye, that will be cool. Just I woudnt like to lose this view of THIS BIG PEW . Big pew looks epics and dangerous, shows the power of shot that is mixed good with the splash :)
(Btw, i'm not sure that anybody else noticed this problem xd)
Thanks for your work on my favorite faction <3

Garanas:
@InspektorKotFAF 
t1arty_new

How about this? We'll see if we can add a tiny bit of smoke come out at the bottom of the barrel.

Garanas:
image

Zapper now remains inside the sphere - closes #3434

PR #3423, opened 2021-09-18
Add textures for map generator

Garanas:
Adds a few textures based on the base-game textures and / or creations of Thacon / Emperor_Penguin.


Garanas:
The textures in question:

image

PR #3432, opened 2021-09-20
Fix bug with targetCats not defined

### LINKED ISSUE #3431: A crash with the overcharge projectile
image

I just know it is the overcharge projectile at [this line](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/DefaultProjectiles.lua#L495). It happens very rarely.

Crotalus:
Fixes #3431


Garanas:
That was quick - I'll look into it tomorrow!

Crotalus:
Just realized targetEntity probably is null in this case which will be bad. I'll push another fix until tomorrow

Crotalus:
Used a flag now instead of checking targetCats again in the ForkThread scope

lua/sim/DefaultProjectiles.lua
```diff
@@ -369,7 +369,7 @@ BaseGenericDebris = Class(DummyProjectile){
 
                 -- create emitter and scale accordingly
                 emit = CreateEmitterAtBone(self, -2, army, v)
-                if effectScale != 1 then
+                if effectScale ~= 1 then
```
Garanas, line 372: 📦

lua/sim/DefaultProjectiles.lua
```diff
@@ -429,59 +423,51 @@ OverchargeProjectile = Class() {
...
+            end
+            local energyLimitDamage = self:EnergyAsDamage(energyLimit)
+            -- Find max available damage
+            damage = math.min(data.maxDamage, energyLimitDamage)
+            -- How much damage do we actually need to kill the unit?
+            local idealDamage = targetEntity:GetHealth()
+            local maxHP = self:UnitsDetection(targetType, targetEntity)
+            idealDamage = maxHP or data.minDamage
+            
+            local targetCats = targetEntity:GetBlueprint().CategoriesHash
```
Garanas, line 440: I believe you mentioned it in your comments, can this also be nil?

Garanas: ah - you're looking into that already.

lua/sim/DefaultProjectiles.lua
```diff
@@ -429,59 +423,51 @@ OverchargeProjectile = Class() {
...
-
-                -- prevents radars blinks if there is less than 5k e in storage when OC hits the target
-                if energyAvailable < 5000 then
-                    damage = energyLimitDamage
-                end
-
+                --MaxHealth instead of GetHealth because with getHealth OC won't kill bubble shield which is in AoE range but has more hp than targetEntity.MyShield.
+                --good against group of mobile shields
+            end
+                  ------ ACU -------
```
Garanas, line 456: fix tabs / layout

Garanas:
I do use Visual Studio Code - never touched any formatter sadly. I always format it myself (yes - it hurts at times) because it allows me to think things through.

Garanas:
I was also referring to whether the targetCats value can be nil somehow, are those hashes always set?

Garanas:
@Crotalus is this PR finished and ready for review?

Issue #2045, opened 2017-05-29
Uveso's long-term bugtracing


Uveso:
I am using this to track/link errors over longer time periods

EXCEPTION_ACCESS_VIOLATION (0xc0000005) at address
0x0213d267 #1803
0x023fde58 #1804
0x0047a820 #348
0x00598eec #1900
0x005f39f7 #1574 (Fixed.)
0x0067aa5f #318
0x006cbc63 #1375 (xrb0104)
0x008b94f0 #1559
0x00938ac8 #1117
0x0095854f #1445
0x009d7b75 #1467

GitHub gamebuild stability tests:
11.05.2017 (3676) part I #1969 result=perfect
12.05.2017 FAForever/deploy/faf (3677) #1971 result=stable
15.05.2017 FAForever/deploy/beta (3678) #1991 result=DEADLOOP
18.05.2017 master (3682) #2001 result=stable
29.05.2017 (3676) part II #2044 result=stable
05.06.2017  (3684) + PR AdvancedLeadTarget #2060 result=stable
06.06.2017 deploy/fafdevelop (3684) #2061 result=stable

Unknown/rare errors without error messages:

ERROR1 - count:4 last:06.Jun.2017
Freeze after transition screen while zooming to CDR.
(Transition screen is half on the way to fade out. ~50% alpha)

ERROR2 - count:3 last:13.May.2018
Crash to desktop while initiating AI Archetype
Last logline: info: *AI DEBUG: ARMY \0004\000: Initiating Archetype using SorianMainTurtle
Last logline: info: *AI DEBUG: ARMY \0001\000: Initiating Archetype using UvesoMainAdaptive
(Transition screen is half on the way to fade out. ~50% alpha)

ERROR3 - count:2 last:19.May.2018
Game Frozen after ACU kill.
Last logline: warning: ACU kill detected. Rating for ranked games is now enforced.

ERROR4 - count:1 last:05.Jun.2017
Game frozen in midgame, sound and sim still running, but UI/gfx is frozen

Last Update 30.Dec.2019
(no new errors since 06 / 2017)

Garanas:
@KionX That is a good find - great work. Here's a few more for this issue:

-------------------

Post: https://forum.faforever.com/topic/2488/constant-game-crashes-need-help-figuring-it-out/17
Replay: 15229180
Branch: FAF
Log: [15229180.txt](https://github.com/FAForever/fa/files/7089985/15229180.txt)

Error provided by game:
```
[...]

PR #3436, opened 2021-09-20
GetReclaimCosts() - Grab blueprints after we know it's a unit

Crotalus:
As a safeguard only get blueprints after we're sure it's a unit and actually need to use blueprint values.

See comment here about GetBlueprint() on a shield bubble entity crashes the game.
https://github.com/FAForever/fa/issues/2045#issuecomment-922841853

### DIFF
--- a/changelog-3724.md
+++ b/changelog-3724.md
@@ -8,7 +8,9 @@ Patch 3721 (19 September, 2021)
 ### Bugs
  - (#3442) Fix scathis packing animation time
  - (#3439) Fix Cybran drone visibility for other players than the owner
-
+ 
+### Stability
+ - (#3436) Prevent fetching blueprints for potential entities with no blueprints
 ### Performances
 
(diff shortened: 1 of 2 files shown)


Garanas:
Is this also called on props?

Crotalus:
Yeah, GetBlueprint() on props are safe. So in this case it probably never will cause a problem but we save a call to GetBlueprint() at least.

Garanas (approved):

Garanas:
I did something wrong, why are the other commits in here? Should I have force-pushed?

Crotalus:
Seems like the branch got merged into itself somehow. I can try re-push the branch to correct it, one sec

Crotalus:
Now it's based on the latest deploy/fafdevelop so "Rebase and merge" should work now

Garanas:
Great 😄 , merging this too

PR #3438, opened 2021-09-20
Destroyer Rebalance

Tagada14:
Main Turret Yaw Range and Turn Speed Adjustments :
UEF and Sera -> significant buffs, they will behave much better  when micro'ed and thus it will be more rewarding
Cybran -> minor buff
Aeon -> small nerf to make them less oppressive

Torpedo changes :
UEF -> adjust the firing cycle, the dps is the same (30) but the destroyer will now deal a lot more damage to t2 subs (vs aeon ~50% increase; vs cybran ~60% increase)
Sera -> adjust firing cycle, now it will have 120 DPS compared to old one ~95 DPS
Aeon -> decrease the dmg and range of the Depth charge. 106,(06) DPS -> 87.5 DPS, 45 range -> 35 range. The Aeon Destroyer had insane DPS vs subs (181,(06) DPS) with the next destroyer (cybran) having 100 DPS. The Aeon destro will still be extremely good vs Subs but it leaves the enemy with some room to micro their subs before they get absolutely annihilated (subs have 45 range, the destro's Torped and Depth charge also had 45, now the torpedo retained 45 range but the Depth charge has 35)
Cybran -> replaced the DoT (Some nice random stuff that was there since original SC it seems) with just upfront damage (there is absolutely no reason for it to be DoT). The DPS stayed the same (100 DPS).


Garanas:
Quite the changes - that may make the UEF destroyer finally worth making 😄

PR #3441, opened 2021-09-20
Experiment with dummy / insignificant units

Garanas:
This is a bit of an experimental branch and it requires some hefty testing.

I am experimenting with having dummy units: units that have essentially nothing in them. A lot of the behavior of a unit is defined through the blueprint - therefore a lot of the Lua is actually not required at all.

In this branch I am toying with making the cybran drones definite dummy units - they only have the functionality they need for their purpose. The question is: is this safe?

~~Because of it, there is another performance increase: instead of costing about 2 ms (difference) it is now 1 ms (difference).~~
See the 2nd post about performance. 

If a dummy unit is stable, it would open the gate for https://github.com/FAForever/fa/pull/3398 to have a proper team color throughout the animation.


Garanas:
More information on the performance side (for 50 hives):
 - With these changes: baseline = 0.6ms, peak (when zoomed out) = 1.5 ms
 - Typical develop branch: baseline = 0.6 ms, peak (when zoomed out) = 1.8 ms

therefore the difference is smaller than I anticipated it would be. But regardless - we can use this to test if it is stable or not.

PR #3447, opened 2021-09-22
Skip AI-related functionality in game with no AI

Garanas:
The base code has AI-related functionality scattered through some files. This PR aims to find that and disable it if we're playing a game with no AI. If you know of more code that is solely run for an AI then please change the PR and / or add a comment and I'll change it accordingly.


speed2CZ:
I think this function was introduced a long time ago to fix some bug with the consumption that was in the base game. But I might be wrong (if it is the case, it was very long time ago),

PR #3200, opened 2020-11-15
add decals/tree falling to more projectiles

shalkya:
also contains a rework of billy fx

PR #3449, opened 2021-09-24
Fix hard-crashes in /lua/DefaultCollisionBeams.lua

Garanas:
This PR undo's some work done by #3200. According to KionX [in this topic](https://github.com/FAForever/fa/issues/2045#issuecomment-924858053) the cause of exceptions with the address `0x0050dfd8` originate from this file, in specific the functions that this PR comments out.

The issue with these hard-crashes is that they do happen, but they're hard to reproduce. I have not been able to reliably reproduce them myself as of yet, and therefore I can not test if this will solve the actual problem.

### DIFF
--- a/changelog-3724.md
+++ b/changelog-3724.md
@@ -9,6 +9,9 @@ Patch 3721 (19 September, 2021)
  - (#3442) Fix scathis packing animation time
  - (#3439) Fix Cybran drone visibility for other players than the owner
 
+### Stability
+ - (#3449) Fix significant hard-crash potential that patch 3721 introduced
+
 ### Performances
 
 ### AI
@@ -16,4 +19,4 @@ Patch 3721 (19 September, 2021)
 ### Other
 
 ### Contributors
- - Jip (#3442, #3439)
+ - Jip (#3442, #3439, #3449)
--- a/lua/defaultcollisionbeams.lua
+++ b/lua/defaultcollisionbeams.lua
@@ -5,7 +5,7 @@
 --
 --  Summary  :  Default definitions collision beams
 --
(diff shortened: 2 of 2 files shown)


Garanas:
This PR needs to be reviewed as quickly as possible - aim is to have it in develop by tonight.

KionX:
Self is CollisionBeamEntity without blueprint.
Perhaps the stream was not deleted in time.

Garanas:
Not sure about the thread leak, but I suspect the thread lived on _at least the same or one tick_ after the collision beam died. Maybe it depends on the order of threads - not sure. But these change should circumvent the issue all together until we have a good idea of what really happens. As currently the game is quite unstable.

Have you looked over the changes? Could you comment or approve them?

KionX (approved):
I do not know why these DamageArea were added.
If you remove them, it will prevent crashes.

Crotalus:
So this is another example of GetBlueprint() crashing when it's called on something else than unit / project? Same problem as I had with mobile shield bubble. 

@KionX 
Would it be possible to check if the caller (I assume first arg?) of the engine function GetBlueprint() is null and then just return nil result? Then we would get a lua error instead of hard game crash?

PR seems good to me, these force damage aren't essential so lets see if this reduces this kind of crash.

Crotalus (approved):

Garanas:
Thank you both for reviewing it.

_I do not know why these DamageArea were added._
To make the simulation more dynamic, as part of #3200. Typically trees do not budge until some form of artillery is involved.

KionX:
_Would it be possible to check if the caller (I assume first arg?) of the engine function GetBlueprint() is null and then just return nil result?_
It's possible. But I would prefer to look for the reason for the null blueprints first.

PR #3450, opened 2021-09-26
Fix icon loading

Garanas:
This branch will try to fix the icon loading procedure that was added during the 3721 patch. It is very broken from many perspectives, including:
 - It loads in textures regardless of whether the mod is active.
 - It loads in _all_ ui textures, regardless of whether they're icons.

These two combined can result in strange behavior, such as inactive mods interfering with the base game. An example as reported by FtGr is the following:
image
image

These are probably defined in some mod on his system. They're override the default icons on the UI.

### DIFF
--- a/init_faf.lua
+++ b/init_faf.lua
@@ -219,6 +219,9 @@ mount_dir_with_whitelist(InitFileDir .. '\\..\\gamedata\\', '*.nx2', '/')
 -- load in any .nxt that matches the whitelist / blacklist in FA gamedata
 mount_dir_with_whitelist(fa_path .. '\\gamedata\\', '*.scd', '/')
 
+--load preferences into the game as well, letting us have much more control over their contents. This also includes cache and similar.
+mount_dir(SHGetFolderPath('LOCAL_APPDATA') .. 'Gas Powered Games\\Supreme Commander Forged Alliance', '/preferences')
+
 -- TODO: ?
 mount_dir(fa_path, '/')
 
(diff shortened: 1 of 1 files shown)


Garanas:
[Penguins Icon Mod.zip](https://github.com/FAForever/fa/files/7231127/Penguins.Icon.Mod.zip)

The pipeline works in terms of receiving the data at the right moment. But - no icon is rendered. Thoughts on what could be going wrong? I suspect it is the path.

Garanas:
I think I got it working reliably now, here you have a mod to test with:
- [Penguins Icon Mod.zip](https://github.com/FAForever/fa/files/7231900/Penguins.Icon.Mod.zip)

I can't check if it causes a desync in multiplayer - when testing locally both instances try to read the same prefs file and it causes one to lock out the other - so the other crashes.

I need someone else to help test that.

Garanas:
Keyser and I tested this PR tonight - it works! 🪂

Garanas:
And a few examples on how to work with this new approach to loading icons:
 - https://github.com/Garanas/strategic-icon-mod-examples

Garanas:
A few notes on this PR:
 - Prefetching is disabled again. We can not send data to blueprints.lua reliably when prefetching is enabled.
 - This PR may cause the game to crash when you have multiple instances of the game on the same computer in the same lobby, all starting at once. As the game writes the preference file on launch, blocking it for other instances.

Dragun123:
Do I need change my icons file structure now?

Garanas:
Not sure - if it is for a sim mod then I don't think so. If it is for a UI mod then yes, 100% certain.

changelog-3724.md
```diff
@@ -29,3 +30,4 @@ Patch 3721 (19 September, 2021)
  - KionX (#3449)
  - Crotalus (#3436)
  - Balthazar (#3450)
+ - Speed2 (#3453)
```
speed2CZ, line 33: small "s" please

Issue #3445, opened 2021-09-22
Decals being made on ocean floor


Garanas:
![image](https://user-images.githubusercontent.com/15778155/134324836-4aa078fe-9a73-40dc-8365-69673487cd3d.png)

When the Cybran frigate was firing upon the UEF frigate decals were made on the ocean floor. That shouldn't be happening. I suspect a decal is made when it hits a unit, regardless of the unit being on water or not.

Garanas:
Fixed with the 3732 release.

PR #3454, opened 2021-09-27
fix projectile issue for weapon without aoe

shalkya:
aim to fix this issue : https://forum.faforever.com/topic/353/blackopsfaf-unleashed-only-for-faf-v19/18

original code works with current faf unit, but it can induce bug with modded unit (or if balance team remove the aoe of an unit).

This work is in progress. 1st commit was made without thinking it through and through (and might induce other unexpected side effect). More commits are needed to fix the other factions.

### DIFF
--- a/lua/cybranprojectiles.lua
+++ b/lua/cybranprojectiles.lua
@@ -68,40 +68,43 @@ CIFProtonBombProjectile = Class(NullShell) { -- T3 strategic bomber
     FxImpactLand = EffectTemplate.CProtonBombHit01,
 
     OnImpact = function(self, targetType, targetEntity)
-        local army = self.Army
-        local pos = self:GetPosition()
         local radius = self.DamageData.DamageRadius
-        local FriendlyFire = self.DamageData.DamageFriendly and radius ~=0
         
-        CreateLightParticle(self, -1, army, 12, 28, 'glow_03', 'ramp_proton_flash_02')
-        CreateLightParticle(self, -1, army, 8, 22, 'glow_03', 'ramp_antimatter_02')
+        if radius > 0 then
+            local FriendlyFire = self.DamageData.DamageFriendly
+            local army = self.Army
+            local pos = self:GetPosition()
         
-        DamageArea( self, pos, radius, 1, 'Force', FriendlyFire )
-        DamageArea( self, pos, radius, 1, 'Force', FriendlyFire )
-        
-        self.DamageData.DamageAmount = self.DamageData.DamageAmount - 2
-        
-        if targetType ~= 'Shield' and targetType ~= 'Water' and targetType ~= 'Air' and targetType ~= 'UnitAir' and targetType ~= 'Projectile' then
-            local rotation = RandomFloat(0.0,6.28)
+            CreateLightParticle(self, -1, army, 12, 28, 'glow_03', 'ramp_proton_flash_02')
+            CreateLightParticle(self, -1, army, 8, 22, 'glow_03', 'ramp_antimatter_02')
             
-            DamageRing(self, pos, 0.1, radius, 10, 'Fire', FriendlyFire, false)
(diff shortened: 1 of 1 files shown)


Garanas:
Adding an if statement that will be guessed wrong each time by the CPU is not idea. In general, I'm not a huge fan of this addition - I don't think people will actually see it because they weren't even aware what debris was that is now faster due to https://github.com/FAForever/fa/pull/3430. If anything, the commander can have this and nothing else.

Regardless - what if we just add 0.5 to the radius for the argument inside the DamageArea function? It would allow you to fix it with a well-versed regular expression.

Garanas:
And there is also this: https://github.com/FAForever/fa/issues/3445

Garanas:
I think we should change the entire logic.  

```lua
-- old logic
local radius = self.DamageData.DamageRadius
local FriendlyFire = self.DamageData.DamageFriendly and radius ~=0

DamageArea( self, pos, radius, 1, 'Force', FriendlyFire )
DamageArea( self, pos, radius, 1, 'Force', FriendlyFire )

self.DamageData.DamageAmount = self.DamageData.DamageAmount - 2

if targetType ~= 'Shield' and targetType ~= 'Water' and targetType ~= 'Air' and targetType ~= 'UnitAir' and targetType ~= 'Projectile' then
     local rotation = RandomFloat(0,2*math.pi)
     local army = self.Army

     CreateDecal(pos, rotation, 'scorch_001_albedo', '', 'Albedo', radius+2, radius+2, 300, 70, army)
end
```

```lua
-- potential new logic
if targetType == "Terrain"then 
    local radius = self.DamageData.DamageRadius
    local FriendlyFire = self.DamageData.DamageFriendly and radius ~=0

    DamageArea( self, pos, radius, 1, 'Force', FriendlyFire ) -- add minimum radius
    DamageArea( self, pos, radius, 1, 'Force', FriendlyFire ) -- add minimum radius

    self.DamageData.DamageAmount = self.DamageData.DamageAmount - 2

    local rotation = RandomFloat(0,2*math.pi)
    local army = self.Army

    CreateDecal(pos, rotation, 'scorch_001_albedo', '', 'Albedo', radius+2, radius+2, 300, 70, army)
end
```

This is significantly more fool proof for a lot of situations, especially when mods are involved. As mods pick and choose what they need regardless of the context. We do not want to create decals (or do DamageArea) when hitting a unit for a gun that turned air-to-air instead of something-to-ground as it is used in the base game. Or, as #3445 shows, when we've hit another boat and then a decal spawns underneath it, on the bottom of the ocean.

And while you're at it, please also optimize the code accordingly:

```lua
-- start of outer scope (file scope)

-- upvalue moho functions for performance 
local EntityGetPosition = _G.moho.entity_methods.GetPosition

-- upvalue math functions for performance
local MathPi = math.pi
local MathPi2 = 2 * math.pi

-- upvalue globals for performance
local Random = Random

-- end of outer scope (file scope)

-- potential new logic inside OnImpact
if targetType == "Terrain"then 

    -- required for tree falling and decal creation
    local pos = EntityGetPosition(self)

    -- keep all interactions with data together
    local data = self.DamageData
    local radius = data .DamageRadius
    local friendlyFire = data .DamageFriendly and radius ~=0
    data.DamageAmount = data.DamageAmount - 2

    -- guard in case there is no radius
    if radius <= 0 then 
        radius = 0.5 
    end

    -- make trees fall
    DamageArea( self, pos, radius, 1, 'Force', friendlyFire )
    DamageArea( self, pos, radius, 1, 'Force', friendlyFire )

    -- create a decal
    local army = self.Army
    local rotation = MathPi2 * Random()
    CreateDecal(pos, rotation, 'scorch_001_albedo', '', 'Albedo', radius+2, radius+2, 300, 70, army)
end
```

As `OnImpact` is called (tens) of thousands of times during a game.

Garanas:
Closing this pull request in favor of one that approaches this problem from another perspective.

PR #3457, opened 2021-09-29
Drone issues

Garanas:
Fixes a few drone issues:
 - Add INSIGNIFICANTUNIT category as that prevents all kinds of interaction.
 - Add UnitId to the drone as that is a value scripts expect to be cached.
 - Removes drones from UpdateAssistersConsumption function.

An example of interaction that is not desirable: being able to change the target the drone is 'building' as seen in [this cast](https://youtu.be/n_ka_woZ3Q4?t=249). 

Has no impact on the performance of the drones.

With thanks to Softles and Archsimkat.

### DIFF
--- a/changelog-3724.md
+++ b/changelog-3724.md
@@ -8,6 +8,7 @@ Patch 3721 (19 September, 2021)
 ### Bugs
  - (#3442) Fix scathis packing animation time
  - (#3439) Fix Cybran drone visibility for other players than the owner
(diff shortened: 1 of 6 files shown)


Garanas:
Just tested it again and reviewed the code changes - all appears to be good. The issues described by Archsimkat & Softles appear to have been solved as I can no longer re-produce the issue of Softles AI.

PR #3458, opened 2021-09-29
Fix explosion of UEF Triad and UEF Destroyer

Garanas:
Fixes the (accidental) change in [this commit](https://github.com/FAForever/fa/commit/b80bf286ee0ddd79f3376b77e275568ae637dea9). 

Destroyer finally got its 'wow' factor, just to see it taken away again.

### DIFF
--- a/changelog-3724.md
+++ b/changelog-3724.md
@@ -4,6 +4,7 @@ Patch 3721 (19 September, 2021)
 ### Lobby
 
 ### Gameplay
+ - (#3458) Fix UEF Triad and UEF Destroyer projectile on impact animation
 
 ### Bugs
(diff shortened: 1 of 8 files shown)


Garanas:
This PR got a bit weird at some point - but it is all good now:
 - T1 Frigate is as it was
 - T2 Triad is as it was
 - T2 Destroyer is slightly bigger than T1 Frig / T2 Triad (does more damage)
 - T3 Summit is as it was
 - T4 Fatboy is slightly smaller than Summit (0.9 multiplier).

Garanas:
Just checked it again on a different computer - all appears good.

PR #3460, opened 2021-09-30
Fix potential nil on transfer

Garanas:
Both `GetReclaimablesInRect` and `GetUnitsInRect` return nil if nothing is found, instead of an empty table. This can happen, as it did during [this replay](https://replay.faforever.com/15419462) at around 35:00 to LilMiss.

image

### DIFF
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -379,16 +379,22 @@ function TransferUnfinishedUnitsAfterDeath(units, armies)
                 failedToTransfer[ID].Orientation = unit:GetOrientation()
 
 
-                for _, reclaim in GetReclaimablesInRect(unit:GetSkirtRect()) do --wrecks can prevent drone from starting construction
-                    if reclaim.IsWreckage then
-                        reclaim:SetCollisionShape('None') --so we set collision shape 'None'
-                        table.insert(modifiedWrecks, reclaim) --and save wrecks to revert our changes later
+                local wrecks = GetReclaimablesInRect(unit:GetSkirtRect())
+                if wrecks then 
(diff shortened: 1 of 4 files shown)


lua/SimUtils.lua
```diff
@@ -379,16 +379,22 @@ function TransferUnfinishedUnitsAfterDeath(units, armies)
                 failedToTransfer[ID].Orientation = unit:GetOrientation()
 
 
-                for _, reclaim in GetReclaimablesInRect(unit:GetSkirtRect()) do --wrecks can prevent drone from starting construction
-                    if reclaim.IsWreckage then
-                        reclaim:SetCollisionShape('None') --so we set collision shape 'None'
-                        table.insert(modifiedWrecks, reclaim) --and save wrecks to revert our changes later
+                local wrecks = GetReclaimablesInRect(unit:GetSkirtRect())
```
speed2CZ, line 382: for _, reclaim in GetReclaimablesInRect(unit:GetSkirtRect()) or {} do

Garanas: I find the current solution to be more readable.

units/UES0103/UES0103_script.lua
```diff
@@ -5,7 +5,7 @@
 --#**
 --#**  Summary  :  UEF Frigate Script
 --#**
---#**  Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
+--#**  Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
```
Sheikah45, line 8: Unintended encoding change

Garanas: Yep, my laptop has issues with that 😞

PR #3461, opened 2021-10-02
Removing game quality computation for games with more than 2 teams

Benzi-Junior:


Garanas:
I'll review this after the hotfix of next Monday.

Tagada14:
Afaik the FFA shouldn't be ranked and I would treat games with more than 2 teams in a similar way since they inherit the same set of issues and imbalances as the FFA, these game modes just don't work well with our rating and game scenario where each player has only 1 life.

Benzi-Junior:
So, with the wiki out of date/not addressing this issue  and everybody I've talked to giving noncommittal answers (to be fair this isn't really at the forefront of current devolpments for FAF) I decided to dig up the piece of [code](https://github.com/FAForever/server/blob/e2dec4b5531f438a41bbd9a89df361ee59ca6ab2/server/games/game.py#L627)  that decides what games are ranked.

The check gets made server side and the rules are:

1. The game only uses ranked mods
2. There are no AI players
3. The game has no more than 2 teams 
4. The game is not FFA
5. AIReplacement is off
6. Fog of War is set to "explored"
7. Cheats are disabled
8. Prebuild is off 
9. No rush is off
10. There are 0 restricted units
11. Teams are locked
12. All teams have the same number of players
13. There are at least 2 players
14. Victory condition is Assassination

points 3. and 4. are really what matters for this pull request and shows that it is in alignment with what happens on the server side.

Garanas:
I'm willing to merge this, but we need a good way of communicating to people that the game quality can not be computed (reliably) because some conditions are not met (e.g., having more than two teams).

As an alternative: We could add to the UI the number of games and / or rating of each team. That way they can 'balance' it themselves a little easier.

Garanas:
I am merging this. We can come back to this PR when we do a larger overhaul of the lobby.

PR #3275, opened 2021-04-11
Fixed Draw Bug

KionX:
Reverting changes of https://github.com/FAForever/fa/pull/3166 which were the cause of the Draw Bug.
Reason: You can just remove it because that fix didn’t work at all.
https://faforever.zulipchat.com/#narrow/stream/203478-general/topic/DrawBug/near/230181105

PR #3152, opened 2020-08-08
Use presence of Sync.Score to detect bogus results

Askaholic:
After many test games (thanks for helping me @ThomasHiatt) I noticed that results are always accompanied by a score update in the sync table. For example:
```
info: Sync: {
info:   AIChat={ },
info:   CameraRequests={ },
info:   GameResult={ { 1, "defeat -10" } },
info:   PlayerQueries={ },
info:   QueryResults={ },
info:   Reclaim={ },
info:   ReleaseIds={ },
info:   Score={
info:     {
info:       blueprints={ xsl0001={ built=1, lost=1, lowest_health=9000 } },
info:       faction=4,
info:       general={
info:         built={ count=1, energy=0, mass=0 },
info:         currentcap={ count=1000 },
info:         currentunits={ count=1 },
info:         energy=4164,
info:         kills={ count=0, energy=0, mass=0 },
info:         lastReclaimedEnergy=0,
info:         lastReclaimedMass=0,
info:         lost={ count=1, energy=5000000, mass=2000 },
info:         mass=658.19799804688,
info:         score=16
info:       },
info:       name="Blodir",
info:       resources={
info:         EnergyReclaimRate=0,
<snip>
info:   },
info:   ScoreAccum={ },
info:   Sounds={ },
info:   UnitData={ },
info:   Voice={ }
info: }
```

However, the bogus results never have that field set:
```
info: Sync: {
info:   AIChat={ },
info:   CameraRequests={ },
info:   GameResult={ { 2, "defeat -10" } },
info:   PlayerQueries={ },
info:   QueryResults={ },
info:   Reclaim={ },
info:   ReleaseIds={ },
info:   Score={ },
info:   ScoreAccum={ },
info:   Sounds={ },
info:   UnitData={ },
info:   Voice={ }
info: }
```
(That result should not be sent, army 1 ctrl-k'd and left the game, so it should not be reporting that army 2 was defeated)

I am pretty confident that this behavior is consistent. We tested this for many ladder games (maybe 6 or 7) and also tested it for some FFA games with 3 people (only had 3 people available for testing so it's the best we could do).

PR #3115, opened 2020-05-25
Issue/#3087 Result reporting on 'Exit to Windows` fix

Askaholic:
New idea, use existing `RequestingExit` property of the Sync table to selectively ignore results. I found this property after tracing what happens when a user clicks the "Score" button, as this turns out to be another way of causing the false results. I suspect it can happen whenever the simulation is stopped early, which probably causes the engine to set all players to "defeated" for a brief moment before exiting. However, I am also hoping that this same code will also be setting the `RequestingExit` property at the same time, and that we can use this to ignore results. 

Still needs to be tested.

Hopefully fixes #3087.

PR #3468, opened 2021-10-04
Revert the changes made to how game results are send

Garanas:
And here we are - this PR is a direct result of https://github.com/FAForever/fa/pull/3275 where changes were made to how games are reported to the server. And it intents to close https://github.com/FAForever/fa/issues/3459 .

Before that there were several PRs that changed the reporting (#3032, #3115, #3152, #3166) as a result of a [discussion on Zulip](https://faforever.zulipchat.com/#narrow/stream/203478-general/topic/DrawBug) and those got into the previous patch (https://github.com/FAForever/fa/releases/tag/3715) and that was considered stable with the exception of one particular draw bug.

I've been trying to re-produce the errors that people [have been reporting](https://forum.faforever.com/topic/2656/game-was-not-reted-reason-unknow-result) [on the forums](https://forum.faforever.com/topic/2596/rating-system-is-not-rating-certain-games-although-it-should/20) and sadly, out of maybe 30 local tests I've been able to re-produce it once and I'm not even certain anymore if I wasn't dreaming.

Therefore I'm reverting the only changes that were made to how the score was send over - hoping that we get back to the same state as before.

### DIFF
--- a/changelog-3724.md
+++ b/changelog-3724.md
@@ -1,8 +1,5 @@
-Patch 3721 (19 September, 2021)
+Patch 3724 (04th October, 2021)
 ============================
-
-### Lobby
-
 ### Gameplay
  - (#3450) An alternative approach to loading in custom strategic icons
  - (#3458) Fix UEF Triad and UEF Destroyer projectile on impact animation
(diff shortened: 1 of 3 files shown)


BlackYps (approved):
This reverts #3275 
effectively going back to the behaviour we had from gamepatch 3717 to 3720.

KionX:
_In that case you can just remove it because that fix didn’t work at all._
This means that (#3032, #3115, #3152, #3166) never worked. Why return them?
@Askaholic

Garanas:
If they never worked - then why do we suddenly have dozens of games with conflicting results after the patch? As far as I could see the only change directly related to this is what this PR reverts, and if I'm not mistaken then that is what BlackYps confirmed.

BlackYps:
From my understanding this is a hotfix to revert to the state we had before where the draw bug existed, but at least the games were rated. A solution to fix the whole game result reporting for good is still missing but out of the scope of this release.
Why are you so sure that the other for PRs you linked are useless? From my impression they alleviated the problem.

Garanas:
Your understanding is correct BlackYps - this is by no means the last PR on this subject. But for now, this appeared to have worked better than we have right now.

Askaholic:
Sounds good. The code I wrote to try and patch over the issue is pretty shit, but I think it happens to catch enough cases that the majority of games are rated.

I had some method that I was able to reproduce it with pretty reliably. I’m sure I wrote it on an issue or PR somewhere already, but I think it went something like:

1. join 2 players into a game
2. One player ctrl-k’s and then immediately exits the game

PR #3470, opened 2021-10-04
Fix overcharge indicator

### LINKED ISSUE #3469: Color indication for overcharge cursor incorrectly indicates units will die
When you mouse over a unit the overcharge cursor is supposed to be green if it will kill a unit, orange if it will not, and grey if it is unknown. Currently, it sometimes shows green when there is not quite enough energy to kill the unit. I suspect this is because the indicator may be using the old energy to damage ratio before it was changed to 3:1. The easiest way to see this is to make a single energy storage and place the overcharge cursor over an ilshavoh. It will turn from orange to green as the storage becomes full, but the overcharge leaves the unit alive with a small amount of health.

Garanas:
Closes #3469 .

Essentially what happened is that the original authors of the change decided to duplicate code (copy / paste). As a result, when the overcharge damage was adjusted the authors of those changes were not aware of this hidden coupling of functionality. This PR introduces a shared space between the sim and ui where functions such as these are defined instead.


Garanas:
I just made it worse.

PR #3471, opened 2021-10-04
Fix overcharge indicator and add shared code

Garanas:
The latter is to fix issues like the former in the future. The original authors copied the code over, and then when the change was made the coupling was not clear and therefore it got left with the old damage calculations.


shalkya:
tested works well

PR #3476, opened 2021-10-06
Fix consumption errors when pausing / unpausing during construction

Garanas:
This fixes errors that the eco manager creates when it tries to pause or un pause fabricators that are not finished yet. The rotators are typically defined during the OnStopBeingBuilt event - which is triggered when the fabricator is finished. 

In particular we're talking about error messages such as these:

```
WARNING: Error running OnScriptBitSet script in Entity xsb1303 at 1878bb08: 
  ...\gamedata\units.scd\units\xsb1303\xsb1303_script.lua(27): attempt to call method SetSpinDown' (a nil value) stack traceback: 
  ...\gamedata\units.scd\units\xsb1303\xsb1303_script.lua(27): in function OnProductionPaused'
  ...gramdata\faforever\gamedata\lua.nx2\lua\sim\unit.lua(443): in function 
  <...gramdata\faforever\gamedata\lua.nx2\lua\sim\unit.lua:418>
```

Performance-wise I'm not a huge fan of this fix as we're adding an engine call to a function that will be called quite often  by eco manager mods. Open for suggestions - perhaps a field set by the OnStopBeingBuilt function and we check that instead?


GodFuper:
The T3 radar also consumers energy when it is not yet completed. If it was paused at the time of construction.

Garanas:
This PR is more involved than I thought it would be and I'm moving it to milestone 3727.

Garanas:
I think this is one type of error that we're going to have to live with. There is no 'catch' that we can apply, instead we'd need to add this individual check to all units that could be paused by the UI. In my opinion the mod would need to be rewritten.

To describe what happens:
 - When a structure is being built it can be paused and unpaused by a UI mod
 - Any logic initialized during `OnStopBeingBuilt` is not run yet, because the structure is being built
 - Therefore any logic depending on that when pausing and unpausing will crash
 
This can happen quite a bit with manipulators such as sliders and rotators. 

We have two solutions:
 - (1) We introduce checks for every unit of the base game for the pause / unpause functionality
 - (2) Eco manager changes its approach to how it pauses units, by performing a completion check

Where (1) is not complete - a mod will hook the pause / unpause callbacks to introduce their own rotators for a structure. We'd never be able to catch those with this approach. In my opinion we should look at (2) instead.

PR #3480, opened 2021-10-10
Change the visual effects of UEF Gauss projectiles

Tagada14:
Change the visual effects on the impact of gauss projectiles of these units:
UEF T2 Destroyer 
UEF T2 PD

To do:
UEF Cruiser and Frigate
UEF Striker and Pillar (maybe)


lua/EffectTemplates.lua
```diff
@@ -1944,7 +1956,18 @@ TShipGaussCannonHit02 = {
     EmtBpPath .. 'shipgauss_cannon_hit_09_emit.bp',
 }
 
-TLandGaussCannonHit01 = {
```
Garanas, line 1947: This broke a few mods 😄 

One can not remove entries due to backwards compatibility.

Issue #3483, opened 2021-10-14
The function of automatically building mass storage around the miners.


Sutrinc:
This is a very handy feature.
But there is a question. Why is an assisting order used when appointing a team of engineers?
As shown in the picture.

![1](https://user-images.githubusercontent.com/92538938/137361444-3f614129-ed91-4922-a469-bfb6c105d7eb.png)

The disadvantage of this is that after the work has been done, the assist is not removed. All engineers remain attached to one. This is not very convenient.

The game, on the other hand, allows you to build buildings by a group of engineers without an assist order.
Maybe there is an opportunity to change this. Or add an option to the settings.

![2](https://user-images.githubusercontent.com/92538938/137361472-b3585569-1b2e-4f53-bb3b-29a7f46da0ae.png)

PR #3484, opened 2021-10-15
Adjust extractor capping

Garanas:
This PR is based on #3483 which is based on [this forum topic](https://forum.faforever.com/topic/2714/the-function-of-automatically-building-mass-storage-around-the-miners/20?_=1634282620577).

image

The current implementation is best described by looking at the changes of this PR, but for clarities sake:
 - Assuming all engineers can build the first viable storage will queue that storage to build.
 - Anything that can build but not that storage will assist one of the engineers.

Sadly, the function `IssueBuildMobile` has some unexpected behavior: instead of giving all the units the same order (as you'd expect when doing a similar action as a player) it gives the orders to the nearest engineer to each structure. E.g., if you have all your engineers to the north of an extractor then only one gets to build all the storages. If you have them surrounding the extractor then four engineers start building each of the storages individually. Both are not acceptable.

Instead - we now queue the same order for all valid engineers. In turn, they all build the same storage. But they do not share the same build queue, e.g., manipulating the build queue (by removing a storage from the queue) is not possible when having multiple engineers selected, and it doesn't do the job when you only remove it from the queue of one engineer. Long story short: you'd need to give all the engineers a new order to cancel the storages.

@KionX Can you find out what the table (last argument) of `IssueBuildMobile` is supposed to do? It needs to be either `{ }` or `{ {x, y} }` where `x` and `y` are some number - but it doesn't appear to be doing anything.

### DIFF
--- a/lua/ui/game/commandmode.lua
+++ b/lua/ui/game/commandmode.lua
@@ -356,7 +356,7 @@ function OnCommandIssued(command)
                 BlueprintID = command.Blueprint,
                 TextureName = '/meshes/game/flag02d_albedo.dds',
                 ShaderName = 'CommandFeedback',
-                UniformScale = 5,
+                UniformScale = 1,
             }, 
             0.7
         )


lua/SimCallbacks.lua
```diff
@@ -79,44 +87,68 @@ Callbacks.ClearCommands = function(data, units)
...
         left = Vector(pos.x - 2, pos.y, pos.z),
         right = Vector(pos.x + 2, pos.y, pos.z),
     }
 
+    -- order them to build things
     for key, location in locations do
         if CanBuildInSpot(mex, msid, location) then
-            IssueBuildMobile({builder}, location, msid, {})
+            for _, builder in builders do 
+                IssueBuildMobile({builder}, location, msid, {})
```
speed2CZ, line 145: why not `IssueBuildMobile(builders, location, msid, {})` ?

Garanas:
Thank you for looking into it - that means that it wouldn't fix the issue described in my post (instead of all engineers doing the same order, the nearest are chosen and the rest is ignored).

A few more ideas after WhenDayBreaks approached me:
 - 4 clicks to mass fab cap a T3 extractor
 - 2 clicks to mass storage cap a t3 fabricator
 - 2 clicks to t1 power gen cap a radar
 - 2 clicks to t1 power gen cap a t2 artillery
 
These are not as volatile as capping a pgen with energy storages and they're regularly used, but then through templates.

lua/SimCallbacks.lua
speed2CZ: Interesting, thanks

Sutrinc:
Will it be possible to disable the assist order somehow?
This was the original disadvantage.

Sutrinc:
I think no consensus is possible. There will be endless arguments.
If you can do that separete option, it'll be fine for everyone.
It will be great. awsome

I would immediately build a cannon in the construction menu with capping. for default.
And did not make additional options. and radar or t2 artillery with pgen.
because it is cheap.

Sutrinc:
The task was to remove the order for assisting.
It turns out that you could do nothing at all.  )))))

Sutrinc:
I don't know why to tie a commander to an engineer?
On the contrary, I can understand.

you play 2v2. your teammate is dead. you are building repositories. again you receive an order to assist. engineers move in a crowd across the map. just that was inconvenient.

you wrote

 When there are engineers of multiple factions in a selection the faction with the most engineers is chosen as the 'main builders', engineers of other factions (or units that can't build storages, poor ACU) assist one of the chosen engineers.
We have three situations:

    When all engineers are of the same faction, they can all build the same storage. No assisting happening.
    When you have engineers of two factions, one must assist the other as they can't build the same storages.
    When you have engineers of one faction and units that can't build the storage (kennel drones, ACU) then they must assist an engineer as they can't build the storages themselves.


I wrote.

in all three situations, there is no order to assist.
Builders who cannot build buildings do not receive orders at all.
we select a group with a large number of types of builders, and build buildings.

I asked to add an option. 
you disagreed.

I could not convince you to cancel the order or add an option. or settings file. asist=0/1

Sorry. There is nothing more I can do. 
Sorry.

PR #3490, opened 2021-10-18
Prevents loading maps and mods twice through various vault locations

Garanas:
When a mod is loaded twice it can cause collisions. This is can happen when someone decides to turn on the backup vault location and the original mod was already present in the default vault location. A common source of issues for SCTA.


Garanas:
Just need to take into account ASI so that I do not break that setup. How does that work at this point?

Garanas:
Added support for ASI and tested if it works.

Garanas:
Added the init_fafbeta.lua file as it was not part of the repository yet. This should overwrite the existing init_fafbeta.lua.

PR #3492, opened 2021-10-23
Remove grey rating for smurf detection

### LINKED ISSUE #3487: Remove rating color interpolation
Better known as grey rating. The comment in the code suggests that this was done for smurf prevention. It has devolved into a hurdle for new players, because being grey excludes you from many games and it can take dozens of games to get a white rating.
 
I am aware that hosts would probably switch to requiring a certain game count, but that would still be an improvement, because it is clearer how to achieve that, than how to get a white rating.
We (Morax, Askaholic, Tagada, Sheikah and me) recently had a call where we talked about how we want to proceed with the matchmaker queues and we agreed that we also want to display the total amount of games in custom lobbies and get rid of grey rating to make it possible for new players to also play custom games after they gained experience through the matchmaker.

https://github.com/FAForever/fa/blob/cca5e78068964969a5f13a9f44d8d7678692abf6/lua/ui/lobby/lobby.lua#L844-L866

Garanas:
Closes #3487.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -841,30 +841,6 @@ function GetPlayerDisplayName(playerInfo)
     end
 end
 
---- Players with a higher deviation have their rating colour tarnished, to make smurfs easier to
--- detect.
-function GetRatingColour(deviation)
-    if deviation < 100 then
-        return "ffffffff"
-    end
-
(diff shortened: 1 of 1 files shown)


Garanas:
Nothing that this won't fix it for Kyro's lobby (the alternative lobby).

PR #3511, opened 2021-10-28
Cache global C funcs

KionX:
The impact on performance is unknown.
- [x] Cache global C funcs
- [x] Cache C funcs of classes
- [x] Cache LUA funcs of modules
- [x] Cache math, string, table, ...
- [x] Cache imports
- [x] Cache in all files


KionX:
I have a lot of questions:
- Which of us will support optimizations?
- Do need to optimize ui files?
- Do need to optimize files missing from FAF?
- I came across a limit of 32 upvalues. At the expense of what to reduce them?
- There is an idea to optimize the search in the table instead of upvalues.
- Do need to format and how aggressively?

Garanas:
_Which of us will develop and support optimization?_
I can't help with an automatic approach, but I am working on this manually by preventing memory allocations and rewriting particular parts of the scripts.

_Do need to optimize ui files?_
When trying this myself they proved to be difficult. I do not think the UI scripts are causing issues at the moment. I suspect they can def. benefit from better code in general (e.g., less garbage creation and more usage of the 'lazy' evaluation of the UI). But besides that - I don't think they require uplifting and all the other things.

_Do need to optimize units, projectiles, props, schook folders?_
Absolutely - initial in-game tests show a 25% increase when using the optimized projectiles and optimized base files. Similar performance can be gained by taking care of the entire hierarchy. Schook folders should be gone soon enough, #3352 .

_Do need to optimize files missing from FAF?_
That is an on going discussion with Sheeo - as long as the association doesn't have the rights transferred we can not suddenly add in all the base game files. We'll have to do those in a separate pass - which is a bit sad.

_I came across a limit of 32 upvalues. At the expense of what to reduce them?_
Remove the ones that are used least, e.g., keep the ones that are used more often.

_There is an idea to optimize the search in the table instead of upvalues._
Can you elaborate?

_Do need to format and how aggressively?_
I am not a huge fan of repository-wide formatting if we're not using the same formatter afterwards. E.g., it would deviate anyhow in the long haul.

KionX:
Which of us will develop and support optimization?
I can give this case away @ChessBerry. If he's interested in it. And I'll do other things.

I came across a limit of 32 upvalues. At the expense of what to reduce them?
I was hoping for something more specific.

There is an idea to optimize the search in the table instead of upvalues.
In the engine it is a linear search through the list. There is an idea to do a search in constant time.

ChessBerry:
> _Which of us will develop and support optimization?_  
> I can give this case away @ChessBerry. If he's interested in it. And I'll do other things.

@KionX If I can save you time by taking over the automated grunt work optimization for the repo, we should probably do that, as you can be much productive than me everywhere else

> _I came across a limit of 32 upvalues. At the expense of what to reduce them?_ 
> I was hoping for something more specific.

You came across an upvalue limit? Where (and why) does such a limit exist?

KionX:
Uploading latest update. There are mistakes.
Pay attention to https://github.com/KionX/fa/blob/deploy/fafdevelop/lua/AI/AIBehaviors.lua

KionX:
_There is an idea to optimize the search in the table instead of upvalues._
I looked at the code again. It has already been optimized.

KionX:
Uploaded new version. This is a working version. Optimized all FA files except UI. The optimization effect is negligible.

Garanas:
A few globals that I think should be excluded:
 - Anything from `debug`.
 - Any logging solution, e.g. LOG, error, WARN.

KionX:
The more optimizations I add, the less I like the way the code looks.
So I'm suspending work on this in favor of LuaJIT integration.
There is no need for perversions with a cache of references to functions.
At the moment I was able to achieve loading init_faf.lua

Garanas:
I'm closing this PR as KionX stated he's suspending work on it. Can't wait to see LuaJIT become a reality - that would solve all our performance issues in one go.

Issue #3508, opened 2021-10-27
Take into account when AI can take over


Garanas:
Has to do with #3447 where AI-related functionality can be skipped. It should take into account the lobby option of an AI taking over from a player.

PR #3512, opened 2021-10-31
Fixes ai detection and threat addition

Garanas:
This closes https://github.com/FAForever/fa/issues/3508 and removes threat additions when a mobile unit dies. This has been discussed with the AI devs and they are in favor.

 - [x] Toggle 'GameHasAIs' when we're in a campaign-like mission
 - [x] Allow a map to toggle 'GameHasAIs' to turn on functionality

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1545,52 +1545,7 @@ MobileUnit = Class(Unit) {
     end,
 
     OnKilled = function(self, instigator, type, overkillRatio)
-
-        -- only assign threat values when we have AIs that will make use of them.
-        if GameHasAIs then 
-            -- Add unit's threat to our influence map
-            local threat = 5
-            local decay = 0.1
-            local currentLayer = self.Layer
-            if instigator then
-                local unit = false
-                if IsUnit(instigator) then
-                    unit = instigator
-                elseif IsProjectile(instigator) or IsCollisionBeam(instigator) then
-                    unit = instigator.unit
-                end
-
-                if unit then
-                    local unitPos = unit:GetCachePosition()
-                    if EntityCategoryContains(categories.STRUCTURE, unit) then
-                        decay = 0.01
-                    end
-
-                    if unitPos then
-                        if currentLayer == 'Sub' then
-                            threat = self:GetAIBrain():GetThreatAtPosition(unitPos, 0, true, 'AntiSub')
-                        elseif currentLayer == 'Air' then
-                            threat = self:GetAIBrain():GetThreatAtPosition(unitPos, 0, true, 'AntiAir')
-                        else
-                            threat = self:GetAIBrain():GetThreatAtPosition(unitPos, 0, true, 'AntiSurface')
-                        end
-                        threat = threat / 2
-                    end
-                end
-            end
-
-            if currentLayer == 'Sub' then
-                self:GetAIBrain():AssignThreatAtPosition(self:GetPosition(), threat, decay * 10, 'AntiSub')
-            elseif currentLayer == 'Air' then
-                self:GetAIBrain():AssignThreatAtPosition(self:GetPosition(), threat, decay, 'AntiAir')
-            elseif currentLayer == 'Water' then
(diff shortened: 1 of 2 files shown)


relent0r:
Happy to validate the changes, will set aside some time this weekend for testing.

relent0r (approved):
Threat Types leveraged by this function.

### **AntiSub**

Calculates AntiSub threat at position and used as a builder condition for naval attack platoons.

**Ramifications**

BuilderConditions
Will no longer trigger when threat is below the threashold.e.g T3 attack required 180 threat. Blueprint for a single t3 battleship is 225 surface threat.
Note. Due to errors in the blueprints certain units don't produce antisub threat so would be invisible without threat assignment. e.g the Seraphim T3 submarine hunter produces zero sub threat. A t2 sub hunter produces 9 sub threat. 
_It is recommended that blueprints for naval units including defences be validated so that this change does not create invisible threat units. With the current implementation they are invisible until they start killing other naval units._

AttackForceAI
Platoon function leveraged by default and sorian AI for various unit type.

**Ramifications**

Builder conditions that have high threat requirements for formations may not trigger without justification. Platoons leveraging this may not path to areas where large battles have recently taken place since the threat will be lower.

### **AntiAir**

GetBestThreatTarget

**Ramifications**

Will provide a more accurate assessment of threats at a position.

**BuilderConditions**
LessThanThreatAtEnemyBase
HaveLessThreatThanNearby
Alot of the other sorian specific builder conditions leveraging threat have been disabled in FAF. Either by Sorian himself or a FAF dev likely due to issues related to threat bloating (or performance).

**AttackForceAI**
Same as AntiSub threat type.

### **AntiSurface**

This as the biggest impact because it is leveraged the most.
Used by multiple platoon functions including AttackForceAI, GuardEngineer, platooncallforhelp, distressresponse, ACU behaviors. Certain build condition functions such as AISortMarkersFromLastPos (mass marker checks)
Certain changes were found in the code that try to minimiz impact by the threat bloating, for example.
This logic artificially inflates a platoons threat to make it call for help later. With the changes this is less likely to trigger and may need to be adjusted.
```
if threat and threat > (myThreat * 1.5) then
    --LOG('*AI DEBUG: Platoon Calling for help')
    aiBrain:BaseMonitorPlatoonDistress(self, threat)
    self.DistressCall = true
end
```

Certain expansion functions that the AI leverage may be negatively impacted due to the decrease in threat at expansions combined with the lack of intel.

**Base monitor functions**
There is impact to this, more positive than negative. One of the observed issues with threat bloat was that if an artillery or other long range weapon was firing at an AI's units close to its base it would trigger the threat response. This results in calling more units to the location, which resulted in them dying and bloating the threat numbers even higher, due to the decay on threat this created a threat loop that made the AI constantly try to attack a threat that didn't exist at its base.


The general consensus is that the native impacts for this change are if the AI has weak scouting/intel behaviors, then this change can have negative consequences and result in them sending engineers to their deaths or not responding correctly to threat conditions around the map.
For AI's that have strong scouting/intel behaviors this change will benefit them as they will get more accurate threat readings.
The biggest global AI impact this change has is on pathing functions. AI unit pathing is less likely to get in a 'locked' situation where due to threat bloat the AI platoons couldn't find a path to a location.
From this developers perspective this change has a positive outcome and allows more accurate decision making by the AI since the threat numbers are no longer a guess.

It is worth pointing out that most modern AI do not use IMAP threat at all for decision making. It is also worth noting that no blueprint threat changes have been made to take into account balance changes over the lifetime of faf so there will be situations where the threat being generated while accurate may not represent the danger a specific unit signifies.

Examples of AI that leverage IMAP for multiple AI behaviors.
Default, Sorian, RNGAI, SorianEdit.
Examples of AI that use IMAP in a limited capacity (e.g pathing or certain threat checks)
Uveso-AI, Swarm
Exampes of AI that do not use IMAP at all.
DilliDalli, M27.

Garanas:
Great work on the evaluation - I am merging this.

PR #3516, opened 2021-11-01
Upvalue units and projectiles

ChessBerry:
Hello,

during our boi's @Garanas optimization crusade a while back, we had the idea to write a lua parser to be able to automatically do lots of small code optimizations. I.e. automate all of the grunt work that doesn't change the gameplay, but still speeds up the game.

One such grunt work project was the upvalue-ing of all moho functions that get regularly accessed, which is what this pull request is about. I mananged to upvalue almost all important moho functions, but some of them I couldn't make work. For more details, see here:
https://github.com/ChessBerry/faf_lua_editor/issues/2

Also, there are less functional changes to the game files than the line count might suggest, much of it is just formatting changes that are a result of the parser turning all code into an AST and then using it's own formatting to spit it back out. So even if the lua code is exactly the same, stuff like indentations might be different.

To separate the two, I made a second pull request here:
https://github.com/FAForever/fa/pull/3517
that is only about the format changes.

I tested the upvalued game files in an AI test game from T1 to T4 and didn't get any warnings that didn't appear in the base fafdevelop branch. That's far from bullet proof testing, but it's a start.


Garanas:
The problem with this PR is that it is too big to test for issues. I'll get a replay and run that on this branch to see if there are gains.

For the future, please do sections of the code first to test things out:
 - Only apply it to the unit files.
 - Only apply it to the projectile files.
 - Apply it to the rest of the code base, excluding UI files.

Note that UI files may appear to have shared functionality with sim files, but typically that is not the case.

baurse:
Yeah splitting the PR is no problem. How many line changes do we want per PR?

Also like half the changes are just formatting that I split into the other PR. Looking over that and merging it first would already reduce this PR's changes by a lot.

Garanas:
For now the list I provided is fine - that way we can run a game with all the units and see what happens. Then run a game with all the projectiles, see again what happens. And when they both succeed then we do not need to 'hit every line' of the rest of the changes.

baurse:
Should I change this PR or just make two new ones?

Garanas:
Better to make new ones - I'll use this one for a performance check or two!

ChessBerry:
Btw. most bugs that result from an upvalue can probably traced back to this moho function reference file:
https://github.com/ChessBerry/faf_lua_editor/blob/main/faf_lua_editor/moho_reference/moho_sim_reference_removed_unsafe.json

The editor goes over all of them and then makes a list of moho functions it _thinks_ can be unambiguously upvalued (i.e. no duplicate names). A few of those upvalues are probably wrong but don't produce an error in normal gameplay. 
E.g. my last commit fed8a3a198b52bb9cc6a30d0fd4fd17e15e12c64  reverted 7 erroneous upvalues of the IAniManipulatorMethods.Disable function, only 1(!) of which actually threw a warning in my test game. The other's either failed silently, or didn't get called at all.

ChessBerry:
Fixed the dropped string arguments in the files:
units/URL0402/URL0402_script.lua
units/XRL0403/XRL0403_script.lua

Unbelievable that this didn't throw an error/warning during my test games, as the lines were converted to utter nonsense. I guess this just demonstrates that "running a test game for half an hour and see if any warnings are thrown" is far from a foolproof method of catching bugs :P

ChessBerry:
@Garanas Could you please run the workflow again to see if something else is broken?

KionX:
Why are global functions renamed? Like GlobalMethodsDamageArea.
I preferred to use such renaming: DamageArea, projectile_methodsSetMaxSpeed, TableCopy.
There are also plans to transfer the maximum of functions to local(Those that are not referenced from other modules).

How was the type proj found?
https://github.com/FAForever/fa/blob/d4065e9483a0a6f4c8c91f895bfbafd5389b3b0c/projectiles/AIFFragmentationSensorShell01/AIFFragmentationSensorShell01_script.lua#L48-L51

Garanas:
I agree with KionX that global functions that are merely upvalued should not be renamed. Besides that, I am myself a fan of EntityGetPosition over entity_methodsGetPosition or ProjectileSetMaxSpeed over projectile_methodsSetMaxSpeed.

ChessBerry:
My Goal was to make it as transparent as possible if and where to a lua function was upvalued.
Imo, "GlobalMethodsDamageArea" does a better job at this than just "DamageArea". The latter could just be some random local function defined somewhere else in the file, while the former makes it clear where it comes from.

Especially considering that we may want to make many more functions local, I think we should keep the GlobalMethods prefix.

That being said the names are easily adjusted if we decide we want to use some other naming scheme.

> How was the type proj found?

What exactly do you mean by that? In the lines you linked, the editor first converted the colon-invoke to a normal function call and then upvalued it. As such the steps were:
```
1. proj:SetVelocity(velocity) 
2. proj.SetVelocity(proj, velocity) 
3. ProjectileMethodsSetVelocity(proj, velocity)
```

KionX:
I am a proponent of minimal renaming.
Do not obfuscate the code with newly added names.

How the optimizer determined that proj is projectile_methods?

ChessBerry:
It didn't it compared the function name "SetVelocity" to this moho reference:
https://github.com/ChessBerry/faf_lua_editor/blob/main/faf_lua_editor/moho_reference/moho_sim_reference_removed_unsafe.json
and saw that there was a function with the exact same name in there, so it upvalued it. 

The optimizer has zero understanding of what function the lua code actually calls, because I didn't know how to give it that understanding. 
All upvalues are essentially guesses.

ChessBerry:
>I am a supporter of minimal renaming.
Do not obfuscate the code with newly added names.

I'd argue it's the exact opposite of obfuscation, it's verbosity. It tells you where the moho function is from, instead of not doing that.

But if you two are in agreement regarding the naming of global moho functions, I'll change it

ChessBerry:
As I was working on this stuff right now I stumbled upon the line
```
DamageArea( self, pos, radius, 1, 'Force', FriendlyFire )
```
in SBOOhwalliStrategicBomb01_script.lua.

This function can be upvalued to a moho function _also_ called DamageArea (or in my old naming scheme GlobalMethodsDamageArea). 
If I now upvalue this function DamageArea to a new function also called DamageArea, which of the two functions will actually be called? The faster upvalued one or the slower non-upvalued one?

ChessBerry:
Tested with a couple replays and the last couple commits seemed to have fixed all the desyncs

Garanas:
Closing with similar reasons to https://github.com/FAForever/fa/pull/3517 .

PR #3517, opened 2021-11-01
Reformat all unit and projectile files.

ChessBerry:
This is the second PR mentioned here:
https://github.com/FAForever/fa/pull/3516

Please read the first PR for details.


ChessBerry:
Tested with a couple replays and the last couple commits seemed to have fixed all the desyncs

Garanas:
But what caused the desync exactly? Removing parenthesis when applicable shouldn't cause any issues

ChessBerry:
Well it previously also removed not superfluous parenthesis. It changed e.g. this
```
a * (b + c)
```
to this
```
a * b + c
```
which is very much not what we want. 

Weirdly this only seemed to matter in the megalith lua file due to a long calculation that happens when it dies, hence I didn't find it immediately, but fixing it for the general case is still important ofc.

Garanas:
Closed in favor of another approach - one in which we start smaller and only refactor the values described by the Lua book on performance: table functions, math functions, etc.

PR #3518, opened 2021-11-01
extremely unimportant loading tips typo

CaptainKlutz:
it has been bothering me


Garanas:
And it has been fixed.

MrRowey:
If your going to fix this typo make sure that the one in Loc/Us/strings_db.lua is also updated as well to keep both in sync

Garanas:
Good point - missed that. @CaptainKlutz could you look into that?

PR #3520, opened 2021-11-03
Add typo fix to localization files

Garanas:
Completes the typo fix of Captain in #3518


Garanas:
Good reminder that when squashing it uses the first commit as its title and not the title of the PR 😄

Issue #3506, opened 2021-10-27
Toggle Moholog settings


Garanas:
@KionX Would it be possible to have all the moholog settings switched off by default:

![image](https://user-images.githubusercontent.com/15778155/139046874-09cbede3-7d8c-4373-9dac-29be7f5b9f5a.png)

It speeds up incidental logging, which happens a lot when AIs are in play (to inform the developer). And it can cause stutters to the game. A dev can then turn it on during their own sessions.

Garanas:
Closed by #3523

PR #3523, opened 2021-11-05
moholog settings switched off by default

KionX:
#3506

### DIFF
--- a/lua/user/prefs.lua
+++ b/lua/user/prefs.lua
@@ -8,15 +8,14 @@
 
 local optionsLogic = import('/lua/options/optionsLogic.lua')
 
-if table.empty(GetPreference('Options.Log')) then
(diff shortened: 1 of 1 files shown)


Garanas:
I was not aware that this was part of the preferences too. What if we just set it to false during each time the game starts? There is no reason for typical people to have all that going on in the background.

KionX:
I propose a compromise.
We will disable LOG for everyone in the next update.
Then we will disable only for new profiles.

Garanas:
And how would that value be set? Through a command line argument?

KionX:
The readme FAF says:
To start several processes of the game you need to add a line
debug = { enable_debug_facilities = true }
to %userprofile%\AppData\Local\Gas Powered Games\Supreme Commander Forged Alliance\Game.prefs

Issue #3463, opened 2021-10-02
Fix Seraphim Sonar being considered a t3 structure by unit restrictions


Garanas:

MrRowey:
fa/lua/ui/lobby/UnitsRestrictions.lua >

Line 117 for T3 Structure restriction has Mobile Sonar Tag which the T2 Sonar has  
  
    T3_BASE_SPAM  = "(TECH3 * STRUCTURE - FACTORY - MASSEXTRACTION - MASSSTORAGE - MASSFABRICATION - ENERGYPRODUCTION - ENERGYSTORAGE + MOBILESONAR)",

PR #3528, opened 2021-11-07
Unit Restriction fix for T2 Sera Sonar

MrRowey:
This fixes the unit restriction for the T3 Spam button disabling the sonar due to the MobileSonar Tag it has which is normally for t3 sonar. #3463


Garanas:
I'll review this tomorrow evening.

Garanas:
@MrRowey thoughts?

MrRowey:
@Garanas the issues was that on t3 base Spam the MoblieSonar Pulls in the T2 Seraphim sonar as the Mobile Sonar tag only relay used for T3 Sonar with the exceptions for the Seraphim which was given the ability to move by FAF

Garanas:
Yes - that happens now. Your solution leaves the t2 sonar in when other factions lost their t2 sonar.

MrRowey:
I can explain more when I'm home

PR #3530, opened 2021-11-09
Comment out Lobby Presets Button 

MrRowey:
@Garanas this one seems hard to test on FAF Dev Setup as Presets don't show up there seems to be faf specific.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -3475,12 +3475,12 @@ function CreateUI(maxPlayers)
 
     if singlePlayer then
         GUI.restrictedUnitsOrPresetsBtn:Hide()
-    elseif isHost then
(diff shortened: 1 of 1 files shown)


KionX:
You can comment everything out --[[ ]]

Garanas:
To me it feels better to just hide the button for now:
 - https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/lobby/lobby.lua#L3478
 
 By commenting that elseif. If suddenly people were using it we can easily get it back and / or if nobody was using it we remove the legacy code for patch 3727.

MrRowey:
> 
> 
> You can comment everything out --[[ ]]

I wondered if there was an easy way to do that just using the Comment out code shortcut in VS Code

MrRowey:
Okay something in the part i commented out Prevent game from launching when lunch button is clicked need to investigate further

Issue #3244, opened 2021-01-02
Matchmaker & Unit Cap


FtXCommando:
It will be necessary to implement a few things involving the unit cap for the matchmaker to function a bit better.

The primary one is that we will need to have shared unit cap enabled so that full share can actually, well, function. This will be more and more important as the team sizes increase on the matchmaker. Can't imagine a sentons without it. 

The second one is that we should just push the unit cap max per player to 1500 rather than 1000. It's mainly for the same reason, no real reason not to do it and I'm getting people whining about it every week. 

Also I'm supposed to link this to the TMM project but don't see that available anywhere here.

KionX:
I suggest making unit cap max per player unlimited.
And the income from t2, t3 mex cut in half.

MrRowey:
@Garanas change have been merged

PR #3533, opened 2021-11-09
Updated Autolobby file for TMM

### LINKED ISSUE #3524: Share unit cap with allies
Shared unit cap should be the default in the autolobby file. Especially for the 4v4 matchmakers this will be important, because at the moment unit cap isn't shared.
Duplicate of #3244 
Also, do we want 1500 unit cap on more than just 1v1?

MrRowey:
to fix #3524 & #3244 
- Updated the Unit cap to 1500
- Added Shared unit cap to Allies upon death


Garanas:
I recall https://github.com/FAForever/fa/pull/3293 and the issue is that this approach messes with the auto host functionality that Softles made.

BlackYps:
what auto host functionality?

Garanas:
He made a tool that could auto-start games between AIs. But he just mentioned this PR should have no impact on his tool. Just waiting for Uveso's input.

lua/ui/lobby/autolobby.lua
```diff
@@ -47,10 +47,11 @@ local gameInfo = {
         RevealCivilians = 'Yes',
         GameSpeed = 'normal',
         FogOfWar = 'explored',
-        UnitCap = '1000',
+        UnitCap = '1500',
```
BlackYps, line 50: The other thing is we haven't really talked about if we want to have the high unit cap for all team sizes including 4v4. Depending on what we decide it might make more sense to leave this on 1000

Garanas:
After discussing it with the AI devs this does not interfere with the tool in question. They also appear to be fine with it for now.

PR #3534, opened 2021-11-10
Fix and optimize props

Garanas:
There are a few fundamental issues to props that need a fix. An example is how tree groups use the first damage they receive to split and do not pass that damage over to the trees underneath. More examples are inefficient code with regard to burning trees (the burning happening at the start can easily take 2% - 5% of the budget, a bit sad when the same happens during a fight). Another example is the requirement to have 'DamageArea' twice in order to properly break and force tree groups down.


Garanas:
@MrRowey How's this for coloring / scaling of the labels?

image

MrRowey:
image

@Garanas So did a test with the mod on my end with my  Color Looks fine to me Could you do another screen grap with how the mid setons reclaim looks as on this image the 194 -> 389 reclaim label seems a lot smaller than the mod and I know the Salme wrecks on settons are massive .
Setons Relaim for a Comparison 
image

Garanas:
image

That looks like this. The exact scale / color can be tweaked and I may not have the right values right now.

MrRowey:
Okay Colour looks Pretty alright would make the Ones on the Salems more near red so it stands out a bit more from the others and as you said just some general tweaking the Salem tag seems about right , 389 could be a tad bigger or just make the other smaller tag a tad smaller

Garanas:
I'll tweak them when the pull request is finished. It is getting a bit out of hand 😄

Garanas:
Closing this - partially applied to #3714 . The other half (that improves the reclaim overview) still has a lot of merit, but we'll have to tackle that another time.

Issue #755, opened 2015-07-13
No Hover or Select effect for SCU Presets


CodingSquirrel:
There is no hover effect for the SCU presets:
- [x] no tooltips when hovering over icons in the Quantum Gateway
- [ ] no semi transparent overlay when hovering over icons while they are in a transport

Also, there is no select effect for the SCU presets:
- [ ] icon background does not change when selecting the SCU in the transport to be dropped individually

CodingSquirrel:
Added pull request #765 with a fix for this.

HUSSAR-mtrela:
cannot reproduce first part of this issue in the 3652 patch but the last two items are still present.

aeoncleanse:
Missing tooltips is this?

HUSSAR-mtrela:
yes, I think tooltips were missing for the first item and I added more info to description of this issue

CodingSquirrel:
I don't remember what I wrote initially but I believe it was about the hover effect, not the tooltip. Basically, there's no visual effect on the icon when hovering over the SCU presets. Other units, when you hover over their icons it creates a slight glow on the icon.

aeoncleanse:
Can we get a check on the status of this?

Nojoke185:
if I have understood the issues correctly then the below seem to still be problems:

1 - SACU presets in a factory, when you hover over the preset icons you dont get a glow when hovering over the icon

2 - same applies to SACU loaded in a transport - when hovering over there is no glow effect

3 - Load a few SACU in a transport, select a single SACU for drop from the transport, there is no selection change unit highlight to show which one is selected for drop (for engineers you can see which are selected in the same scenario)

MrRowey:
Icons for the presets are being pulled from the textures/ui/common/icons/units folder where there seems to be customer Icons made for the presets and I've tested and it's definitely pulling them from there but dose does not seem to be applying the switch between icons when I've added the btn_down/up/over.dds files but this is a start

Garanas:
Can you create the pull request with the work so far? It makes the discussion easier.

MrRowey:
@Garanas i think the best way to implement this would be to create a Clone of the SACU and rename them and add those to the Quantum gate which would allow for the effect to work. just to how this was first implemented you can only have a single image and not the switch between ones like for the normal sacu

MrRowey:
This Dose not affect the gameplay it just a Ui element that doesn't visually show. For now will close as Unplanned but if someone want to pick it up and finish it off feel free to reopen it

PR #3536, opened 2021-11-10
SACU Preset have no Hover/Up/Down Effect

MrRowey:
to fix issue #755 

Added 3 icons for Engineer present so I can make this pr as these are what I'm testing functionality with

made this a Draft as this is still ongoing as I've not actually changed any Lua yet still looking for where the Presets are called from to place in the quantum gate

Testing is done so far:
Removed the Engineer_icon image from texture/ui/common/icons/units this did in fact then apply the deault_icon image and also tried it with the button down image renamed and that displayed so it is pulling the Preset_Icon.dds file fine from the location. the issue it seems is that I cant see yet is it not calling/ cant see the other files that are in the folder with the correct naming.


Garanas:
I have not worked enough with images to help out - maybe @The-Balthazar has a pointer for you.

Garanas:
Closing this pull request - it should be fixed at some point, but it is not critical.

PR #3541, opened 2021-11-13
Add lazy var evaluation for reclaim labels

Garanas:
This is an experimental branch with regard to reclaim labels - it will be embedded into #3534 .

I noticed that these labels were not using lazy evaluation, and instead re-creating functions and closures each frame to update their positions. On this branch they use lazy evaluation. This works like a charm because now the labels are only updated during movement and not during each individual frame. This has a significant performance boost: just rendering the 1000 labels is not an issue. But, updating them (even through lazy evaluation) the memory usage just spikes (~10 mb per second) and then it stutters because the garbage collector kicks in. 

@Crotalus You mentioned using lazy evaluation as a way to speed up the performance of the UI. Do you have any idea what causes the massive memory usage?

Steps to reproduce:
 - Turn on `ShowStats` in the console
 - Open up `Heap` and `Frame` 
 - Start the map `Forest Nothing`
 - Zoom out after starting, then press CTRL + SHIFT to show reclaim
 - Watch framerate drop when moving, but stable when not moving and check logs to see how often things are updated

I have no idea where the memory is being used for. At least when merging with the other PR the performance will be steady when the camera is not moving at all.

And another question: how does the lazy evaluation work internally? I found the file in the base game, but it isn't quite as informative.

### DIFF
--- a/lua/lazyvar.lua
+++ b/lua/lazyvar.lua
@@ -2,8 +2,10 @@
 -- LazyVar module
 --
 
+-- upvalue for performance
+local pcall = pcall
 
-local EvalContext = nil
+local EvalContext = false
 
 local LazyVarMetaTable = { }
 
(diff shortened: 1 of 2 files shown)


Garanas:
The performance status at this point, with a base line fps of 90 when no labels are shown:
 - When showing 1000 labels without moving the camera: 15 fps -> 80 fps
 - When showing 1000 labels while moving the camera: 15 fps -> 35 fps
 
 There's also the question of grouping labels: is there interesting in this? I personally think that having a 1000 labels doesn't make anything easier. Perhaps 100 - 150 tops and combine them when the number of labels becomes too high.

Garanas:
Closed in favor of #3534 .

Issue #3540, opened 2021-11-13
Window class TitleGroup scale 


4z0t:
Scale Window TitleGroup on higher resolutions through LayoutHelpers (line 220):
self.TitleGroup.Height:Set(30)  --> LayoutHelpers.SetHeight(self.TitleGroup, 30)
[(https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/maui/window.lua)](url)

Garanas:
Could you make this into a pull request?

PR #3542, opened 2021-11-14
Window class TitleGroup scale

MrRowey:
Pr for #3540


Garanas:
Shouldn't the rest of the window scale accordingly too?

MrRowey:
Probably I'll fix accordingly later did decide to do this at like 1am my time do didn't realy read much into it 😆

MrRowey:
image
 @Garanas I assume you're talking about the other 3 about the line I changed ?

Garanas:
What LayoutHelpers does is make that part scale accordingly with the UI scaling implemented a while ago. What we have now is that the height of the window scales accordingly, but its size (top / left / right) does not.

MrRowey:
i can adjust this

speed2CZ:
Well have you tested it? How does it look before/after the change? The size top/left/right is not set by the number in this case. It's a reference to other element, so it changes to whatever the referenced element changes to. So it should be always correct.

Garanas:
Closing this pull request - the issue is there to pick up for someone else.

PR #3550, opened 2021-11-19
Fix Rhino missing shots

### LINKED ISSUE #3548: Rhino Shots Miss
Rhino's miss moving targets

Have units moving while the rhino is moving in a diagonal direction and firing at them

The image below shows a rhino missing. It misses 3 times in a row until it changes direction
image
The replay: https://replay.faforever.com/15713773
This only ever seems to happen in online games.

Garanas:
An attempt to fix the rhino from missing shots when the beam weapon got introduced. The test case I had going was no longer reproducible after changing this file. The test case was having a gatling bot (UEF, T2) walk around the Rhino at about half range while the gatling bot was aiming for the Rhino. This would cause the Rhino to miss, I assume because it tried to aim ahead of the entity. 

I'm not entirely confident that this solves the issue.

Closes #3548 .

### DIFF
--- a/units/URL0202/URL0202_unit.bp
+++ b/units/URL0202/URL0202_unit.bp
@@ -203,7 +203,7 @@ UnitBlueprint {
                 },
             },
             BallisticArc = 'RULEUBA_None',
-            BeamCollisionDelay = 0.1,
+            BeamCollisionDelay = 0,
             BeamLifetime = 0.1,
             CollideFriendly = false,
             Damage = 25,


Garanas:
More people have reported this. It appears that the collision box is updated 10x a frame, where as the aim of the collision beam is updated 60x a frame. In turn, fast moving units with unit boxes that are small in comparison to their velocity are prime candidates to cause beam weapons to miss. We'll need to look into alternatives as to how we can approach this.

With thanks to Balthazar as source.

PR #3554, opened 2021-11-22
Add functionality to swap players in lobby

4z0t:
Numbers of players in lobby get functionality to swap players


Garanas:
Works as advertised - great job!

PR #3557, opened 2021-11-22
Optimize seraphim build effects

Garanas:
Optimizes the seraphim build effects:
 - Logic of slider is simplified using cubic function to scale
 - No vector allocations
 - Tables cached as up value when possible
 - No projectiles created that didn't add anything
 - No calls to WaitSeconds
 - No accumulating threads being made when pausing / unpausing
 - Adds build extents to unit on creation to prevent additional blueprint lookup
 
And fixes a few small issues:
 - Fix bone orientation of t2 Seraphim engineer
 - Fix units 'jumping' on the pad when they finish


Garanas:
Tested extensively visually - have not found any issues yet.

PR #3562, opened 2021-11-25
Stealth Generator Projected area fixes

MrRowey:
**Stealth Gens Completed**

- [ ] Cybran
- [ ] UEF
- [ ] Aeon
- [ ] Seraphim

---
So for the baseline when you build in this case the cybran Stealth gen you can fit 27 wall sections on the left and right side of the generator & 28 walls on top and bottom which are inside the ring. this is with the Offset set to 0,0,4
image

here is what a radar shows by default with the stealth gen
image
you can see that No walls show at the top, 2 walls to the left, 2 from the bottom, 1 from the right.

after a few hours of testing, I've come to a 95% fix with the offset to 2,-2,2 only showing 1 wall section from the right which is already better than the default
image

Note the only thing that will need fixing that I don't know how is the stealth effect that shows on the building is now offset from it . https://youtu.be/D7ooI9GMKcM


Garanas:
@The-Balthazar Do you know how the offset parameter works that Rowey has been editing? It has three coordinates, where I'd expect it to have only two (stealth works on the plane, e.g., X and Z for supreme commander)

MrRowey:
Hmm @Garanas The Offset seems to be inconsistent just re set the offset back to 2, -2, 2 and the outcome has changed again so the offsets don't seem to be consistent.

Just check the UEF on that on is out as well but the Script file doesn't even contain the offset changes

Garanas:
Does that mean a cliff could reduce / increase the radius of the stealth field?

MrRowey:
I will be testing is more this coming week, had IRL thing todo and I'm off this weekend to make sure the changes I make are consistent and I will update this PR Accordingly

Garanas:
@MrRowey What is the progress on this, do you reckon it to be done this weekend?

MrRowey:
I'm working this weekend so I'm going to say no at this moment tin time

Garanas:
That is fine - pushing it back. Good luck with your work.

Garanas:
Closing this due to inactivity.

PR #3575, opened 2021-11-27
Separate t2 tanks icons

4z0t:
Add new icon path for t2 amphibian tanks (also hover), but textures are the same, so it opens possibilities for modded icons.


Garanas:
I'd say to rebase this to the FAF Beta branch but then it may take several months for it to emerge. I do think this is a balance related issue - I'm going to let @Petricpwnz / @Tagada14 decide on this.

4z0t:
Found better solution, PR not needed 🙂

Garanas:
Enlighten us - what other solution did you find?

4z0t:
> Enlighten us - what other solution did you find?

I don't think it's a good idea to do such a change in icon structure, because a lot of people using icon mods will have problems. 
I noticed that i can specify for unit what icon to use through **mod_icons.lua**

Garanas:
Sounds good - means the implementation Balthazar and I made is robust.

PR #3582, opened 2021-12-03
Fix aeon build animations rebased

Garanas:
Rebased version of #3398.


Garanas:
A list of changes. Top image is before, bottom image is after.

image
image

Fix the puddle orientation for units made in the land factory.

image
image
image

Introduce build animation for the Colossus.

image
image

Fix the bones being underground during the build animation - the bones (and therefore the aim bones) are now always where you'd expect them to be. No more units firing unto an unfinished Aeon PD :) .

image
image

Prevent jumping of hover units when they're finished. They have a special slider to ensure they're at hover height. Only applies to hover units.

image
image

The build animation now uses the same pixel shader as the regular shader. This makes them match in style, e.g., the jumping between the build shader and the regular shader is a lot smaller now.

Other changes with no direct reference:

image

Introduce build animation for the Tempest.

image

Introduce build animation for the CZAR.

image

Improve the build animation for the Paragon.

Issue #3578, opened 2021-11-29
Refactor `OnMotionHorzEventChange` of the Unit class


Garanas:
Description
-------------
This involves the refactoring of [OnMotionHorzEventChange](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L3143) as defined in the [Unit class](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua). The function is called extremely often and it is too generic for its purpose: a lot of the code that is continuously being run does not apply to half the units.

Course of action
------------------
Refactor the function to the base classes in [defaultunits.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua). In particular bits and pieces of code should move to:
 - [ ] [The structure class](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L43)
 - [ ] [The mobile unit class](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L1535)
 - [ ] [The walking unit class](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L1618)
 - [ ] [The sub unit class](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L1654)
 - [ ] [The air unit class](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L1666)
 - [ ] [The land unit class](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L2028)
 - [ ] [The sea unit class](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L2161)

Note that not all of these require the logic - you'll need to figure out what fits where best. You can test this by spawning in new units and moving them around. Make sure to add LOG statements and keep them in the PR until after it has been reviewed. 

Always feel free to ask questions and make remarks.

Test plan
------------------

Take out one bit at a time. Carefully examine what code should remain generic. Use one pull request where in each commit you move one part of the logic to the correct base class. Make sure to have a list of units where you're sure it calls the right code to help make the pull request more testable for both yourself and the reviewer.

Learning goals
------------------
Feel comfortable working on the repository. Learn about the class hierarchy that units (and other classes, like projectiles) have.

PR #3587, opened 2021-12-04
Reduce engine calls on common functions

Garanas:


lua/sim/Unit.lua
```diff
@@ -3645,7 +3663,7 @@ Unit = Class(moho.unit_methods) {
     end,
 
     GetWeaponClass = function(self, label)
-        return self.Weapons[label] or import('/lua/sim/Weapon.lua').Weapon
+        return self.WeaponAccess[label] or import('/lua/sim/Weapon.lua').Weapon
```
The-Balthazar, line 3666: This line needs reverting. self.Weapon is populated on the unit-script level with the actual weapon scripts. It's supposed to return the class, not the instanced weapon object.

Garanas: Spot on.

Garanas:
This PR is related to #3578. It tries to remove engine calls as those are proven to be expensive:
 - https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/benchmarking/benchmarks/function-origin.lua
 - https://gitlab.com/supreme-commander-forged-alliance/other/profiler/-/blob/main/mods/profiler/modules/benchmarking/benchmarks/blueprints.lua
 
By caching these individual values we're able to:
 - Prevent various engine calls
 - Prevent additional table lookups
 
Especially relevant as the number of engine calls to retrieve a blueprint could easily happen multiple times in a single stack trace:
 - OnLayerChange -> PlayUnitSound
 - OnLayerChange -> PlayUnitAmbientSound
 - OnLayerChange

And for example:
 - OnMotionHorzEventChange
 - OnMotionHorzEventChange -> PlayUnitSound
 - OnMotionHorzEventChange -> PlayUnitAmbientSounds
 - OnMotionHorzEventChange -> UpdateMovementEffectsOnMotionEventChange

That means we're retrieving the same blueprint - which is an expensive operation - multiple times just to read one value. No more 😄 .

PR #3591, opened 2021-12-08
Fix backwards compatibility of blueprints.lua

Garanas:
Introduces a guard in case a blueprint doesn't have categories.


Garanas:
Confirmed by Dragun that it works.

PR #3593, opened 2021-12-09
Fix window methods

4z0t:
Makes  `:SetSizeLock` and `:SetPositionLock` usable.

PR #3594, opened 2021-12-09
Minimap pin

4z0t:
Adds minimap pin that prevents resizing and moving.
Requires #3593


Garanas:
Once localization is added I'll merge this. Works as described.

AigizIskuzhin:
- game launched
- no factory under mouse
image
- on first hover on factory it being created and you can notice it
image
- now it is gone
image

Garanas:
I'm going to test it tomorrow 😄

PR #3624, opened 2021-12-21
possible fix of a bug with secondary display

4z0t:
- the template removal window pops up again
 - the unit highlights in transports work again
 - the factory queue on hover works too


Garanas:
The UI where the actual queue of a factory is no longer updates when queueing units. Only on roll off and when reselecting the factory.

PR #3627, opened 2021-12-22
Whitelist Kyro's lobby

Garanas:
Adds a separate exception for Kyro's lobby to the initialization files. In particular, a file named `kyros.nxt` is accepted.

I was not aware that many people were using this. Note that this doesn't mean that we officially support it - if the game breaks then there is nothing we can / will do about that.

### DIFF
--- a/init_fafbeta.lua
+++ b/init_fafbeta.lua
@@ -88,6 +88,7 @@ allowedAssetsScd = LowerHashTable(allowedAssetsScd)
 
 -- typical backwards compatible packages
 local allowedAssetsNxt = { }
+allowedAssetsNxt["kyros.nxt"] = true
 allowedAssetsNxt["texturepack.nxt"] = true
 allowedAssetsNxt["advanced strategic icons.nxt"] = true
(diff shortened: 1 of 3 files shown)


Kazbek:
Can you explain how I need to pack lobby mod to start working?

init_fafbeta.lua
```diff
@@ -88,6 +88,7 @@ allowedAssetsScd = LowerHashTable(allowedAssetsScd)
 
 -- typical backwards compatible packages
 local allowedAssetsNxt = { }
+allowedAssetsNxt["kyros.nxt"] = true
```
Garanas, line 91: @Kazbek You name it kyros.nxt, that should contain all the files accordingly. It is a zip file.

Askaholic:
Why not make it something more generic like `custom_lobby.nxt` or something?

Garanas:
I don't want to encourage people to write their own lobbies. Instead, I'd like them to work on this repository to improve the current lobby. Kyro is legacy, hence the name.

PR #3444, opened 2021-09-21
Optimize projectiles

Garanas:
This pull request will once completed completely optimize all the projectiles in the game. These are the definition of garbage creation as they are allocated just to explode again. Initial tests show that we can create a significant difference in sim time, as an example this [earlier pull request](https://github.com/FAForever/fa/pull/3392).

Work to be done: 
 - [x] /lua/sim/Projectile.lua
 - [x] /lua/sim/DefaultDamage.lua
 - [ ] /lua/sim/DefaultProjectiles.lua
 - [ ] /lua/aeonprojectiles.lua
 - [ ] /lua/cybranprojectiles.lua
 - [ ] /lua/terranprojectiles.lua
 - [ ] /lua/seraphimprojectiles.lua
 - [ ] All individual projectiles of the base game
 - [ ] Tactical missile manager to prevent threads from being made

PR #3630, opened 2021-12-23
Cache the blueprint via meta table

Garanas:


Garanas:
Closed in favor of a different approach, such as starting with projectiles first as in https://github.com/FAForever/fa/pull/3444 .

PR #3635, opened 2021-12-27
Experiment with torpedoes

Garanas:
This is likely a balance related PR than a develop related one. 

image
_old behavior_

I noticed that there are two issues with torpedo bombers:
 - Projectiles aim at their target while not being in the water
 - Projectiles essentially fly towards their target, unless that target is considerably deep (aka, a unit on the ocean floor)
 - Because the projectiles fly, torpedo defenses do not trigger

image
_new behavior_

In this PR I change this to:
 - Projectiles take the velocity of the bomber, but with a slightly increased y velocity
 - Projectiles do not aim until they are in the water
 - Projectiles slow down on impact with the water surface (essentially to a velocity of 0.1), this ensures that they do not immediately drop into the floor of a shallow pond
 - Torpedo defenses work properly, in fact the Aeon frigate is now quite rigid against the UEF torpedo bomber because the first torpedo is always intercepted.

I personally think the new behavior is how it should be. People complained that it would make it impossible to hit things in say the pond of Seton's. Tweaking the projectiles slightly allows you to still hit things (al though not as reliable as some torpedoes drop on land now, but that makes sense :) ). It also allows torpedo defenses to be relevant when torpedo bombers try and strike your fleet. For example, you may want to take out the UEF torpedo boats _before_ you aim for the cruisers.

Note that this PR now only applies to the UEF torpedo bomber.

### DIFF
--- a/projectiles/TANAnglerTorpedo06/TANAnglerTorpedo06_proj.bp
+++ b/projectiles/TANAnglerTorpedo06/TANAnglerTorpedo06_proj.bp
@@ -43,11 +43,10 @@ ProjectileBlueprint {
         DestroyOnWater = false,
         InitialSpeed = 14,
         Lifetime = 20,
-        UseGravity = false,
         MaxSpeed = 14,
         StayUnderwater = true,
-        TrackTarget = true,
+        TrackTarget = false,
         TurnRate = 230,
-        VelocityAlign = false,
+        VelocityAlign = true,
     },
 }
--- a/projectiles/TANAnglerTorpedo06/TANAnglerTorpedo06_proj.bp
+++ b/projectiles/TANAnglerTorpedo06/TANAnglerTorpedo06_proj.bp
@@ -43,10 +43,11 @@ ProjectileBlueprint {
         DestroyOnWater = false,
         InitialSpeed = 14,
         Lifetime = 20,
+        UseGravity = false,
         MaxSpeed = 14,
         StayUnderwater = true,
         TrackTarget = true,
         TurnRate = 230,
-        VelocityAlign = true,
+        VelocityAlign = false,
     },
 }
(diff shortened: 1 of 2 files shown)


RabidPope:
Hi. I thought torp defence does not work against torp bombers. They count as depth charges.

Garanas:
some of them are, like the Aeon torpedo bomber. But others (like Cybran and UEF) simply fly to their target in the sky. That is why it won't trigger.

EvildrewFAF:
The cost of torpedo bombers were already increased in the past because they were considered too strong. If you add this counter, it would be another weakness on top, then that would nerf them once more. In principle torpedoes should swim towards their target in the water, I would prefer this too. However I would think that a delay on firing torpedo defenses to enable close range torpedoes such as those of bombers to hit would make for maintaining the state of things and allow for torpedoes to swim to their target rather than fly. That would also strengthen subs that move into close range, maybe a desirable consequence or not, who knows. 
If we take gun shootings in movies like The Matrix and others as an example a bullet fired from close range is near impossible to dodge whereas dodging one fired from way down town seems like a cakewalk.
Also seraphim and aoen torpedo bombers fire 3 projectiles, cybran and uef only 2, if 1 out of 3 doesnt hit for 2 factions and 1 out of 2 for the other 2 factions that might be considered unbalanced by some people. 

What is your real motive, to have torpedo bombers drop them in the water before they make impact or some partial torpedo damage to be preventable? How would torpedo bombers do vs a stealth boat with its special kind of torpedo defense? Should T2 subs be able to counter them too? Many questions :)

ChessBerry:
From the discord and some testing: Only uef torpedos seem to actually trigger torpedo defenses, cybran torpedos dropped from torps for some reason don't. 
As a side effect, uef technically has the worst torps. For example: If you shoot uef torps on a moving t2 cybran sub, the sub reliably intercepts the first of the two torpedos, which it doesn't for the other three factions

Garanas:
Closing this - the balance team is well aware of the issue and they can take the inspiration of this branch as they see fit.

Issue #3631, opened 2021-12-25
Keep assisted ferry routes alive


Brutus5000:
One of the coolest but majorly underused features are ferry routes.

- You can set land factory waypoint to the route to send all units along
- you can set air factories waypoints to the route so new transports assist the ferry route

However: as soon as the initial transport gets shutdown, the ferry is removed completely. This makes investing time into complex routes a lost cause.

Suggestions: either
a) if the leading transport dies, give ownership of the ferry route to a random assisting transport
b) keep the ferry route even if all transports die so you can reuse it (you have to remove it manually)

Garanas:
I'm not sure where the beacon interaction is coded. The beacons do exist on the lua side as units, but I can not find where they are created and where they are destroyed. It appears all of this is done by the engine. @The-Balthazar do you have suggestions?

PR #3640, opened 2021-12-30
Feature/persist ferry

Crotalus:
Persistent ferry routes, #3631

I resurrected an old branch I had on persistent ferry routes. Half of the work (simcallbacks, helper unit) was already merged years ago. The dummy unit wasn't destroyed correctly then but I fixed that now.

Needs to be tested, especially with multiple routes in same ferry route. Unfortunately we cannot access command queue on the UI side

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -95,6 +95,7 @@ Callbacks.PersistFerry = function(data, units)
 
     -- function CreateUnit(blueprint, army, tx, ty, tz, qx, qy, qz, qw, [layer])
     local helper = CreateUnit('hel0001', units[1].Army, start[1], start[2], start[3], 1, 1, 1, 1, 'Air')
+
     TableInsert (units, helper)
     IssueClearCommands(units)
     for _, r in data.route do
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1515,6 +1515,7 @@ TransportBeaconUnit = Class(StructureUnit) {
         StructureUnit.OnCreate(self)
         self:SetCapturable(false)
         self:SetReclaimable(false)
+        self.CanBeKilled = false
     end,
 }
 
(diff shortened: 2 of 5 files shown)


Garanas:
Interesting - I'm not entirely sure how it works yet but I'll have a look 😄 .

Crotalus:
It creates a hidden dummy unit that is part of the ferry order so even when the original transport(s) are dead the dummy unit keeps it alive.

I was about to try the same with persistent building orders but never got to it :)

Garanas:
What building orders would you want to be persistent?

Crotalus:
Had an idea of making orders persistent so you can so if the engineers died you could just right click the order with some new engineers and they continue on the same stuff

But I've noticed that many orders creates a bit lag when you shift and it needs to show orders (when you queue up 100 t2 fabs etc)

Garanas:
Yes - the command queue is quite hefty on the simulation. I'll look into this PR next week.

Crotalus:
Probably needs to test that the dummy units doesn't count as real unit, maybe it should have INSIGNIFICANTUNIT or what the flag is called

Garanas:
I think it should have both INSIGNIFICANTUNIT and DUMMYUNIT. That way they're not picked up by the players or script in a meaningful way.

Crotalus:
I've tested this with annihilation / supremacy / assassination now.

victory.lua checks the flag CanBeKilled when looking for still alive units so I use this flag on the transport beacon and ferry helper to make them not count.

Unfortunately I cannot set INSIGNIFICANTUNIT on the helper, it seems to make the ferry beacon order invisible. I added DUMMYUNIT on the helper though.

Garanas:
There's some odd behavior:
 - (1) If you remove the transport beacon when transports are still attached they still try and transport - but there is no beacon anymore
 - (2) If you queue a transport sequence, then reselect the unit, and queue up the remainder of the transport sequence it creates a new queue instead
 
 I think we can fix (1) by issuing a stop order to all guarding units. Not sure how to approach (2).

Garanas:
Closing this issue - we'll try and fix it via assembly instead.

PR #3642, opened 2022-01-04
Change army color order

BlackYps:
We didn't really think about army colors in the team matchmakers, so they were pretty random. This new order creates team red vs.team blue for 2v2. For 4v4 we get a "warm" team against a "cold" team.
I ordered the rest of the colors in a way that somwhat similar colors are in order when picking them in the lobby. As I don't expect that we will make any 5v5+ matchmaker in the foreseeable future.

### DIFF
--- a/lua/GameColors.lua
+++ b/lua/GameColors.lua
@@ -7,18 +7,18 @@ GameColors = {
     "FFe80a0a",      -- Cybran red
     "FF2929e1",      -- UEF blue
     "ff901427",      -- dark red
-    "ff66ffcc",      -- aqua
+    "ff9161ff",      -- purple
     "FFFF873E",      -- Nomads orange
-    "ff40bf40",      -- mid green
+    "ff66ffcc",      -- aqua
     "fffafa00",      -- new yellow
     "ffff88ff",      -- pink
     "ffff32ff",      -- new fuschia
-    "ff9161ff",      -- purple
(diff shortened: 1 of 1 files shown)


Garanas:
I expected this to change the default lobby too, but I see no effect there.

image

Tried making a new profile in case that was messing with it. What makes you sure that the players will have these new colors in the match maker?

BlackYps:
You should be able to see the differences in a normal game lobby. I used texturepack.nxt to sneak this file into my game files, because I haven't set up a proper dev environment. I have used this successfully in the past in my games, so I know that it will show up in matchmaker games if it shows up in a custom game lobby or in replays.
I'm not sure why it isn't working for you.

Garanas:
Merging this on Monday - I'm not sure why it doesn't work but the changes are not game breaking anyhow.

BlackYps:
After some test games I wonder if it wouldn't be better to switch the green one with purple

Garanas:
I'll try and sneak in these changes like you did. I'd still like to see it work.

BlackYps:
I just tested the changes with a properly set up dev environment and I can see the changed colors when starting the game. I don't know why it is not working for you.

Garanas:
That is enough for me 😄

Garanas:
image

It works on the develop game type

PR #3646, opened 2022-01-11
Adjust map preview button

### LINKED ISSUE #3645: Map preview in lobby
The magnifying glass icon in the bottom right corner of the map preview that you need to click on to see the large preview of the map - could this icon be made much bigger? Maybe three times as large? There's no advantage to having it so small and I wouldn't be surprised if some players don't even know it's there.

Also, could we change it so the map preview can be clicked even after you have x-ed up to say that you are game ready? Basically stop the icon being disabled at all? I've never understood why you can't indicate you are ready and then stare at the large preview of the map while you wait for the game launch!

Garanas:
Closes #3645 . In particular:
 - Increases the size of the button.
 - Allows the button to remain enabled when the user is ready.
 
image

Issue is the resolution of the button. We'll need a new texture. I've attached the current textures: [zoom.zip](https://github.com/FAForever/fa/files/7850305/zoom.zip), we'll need icons at a resolution of 32x32 in a similar style / fashion and the same button presets.


FemboyJavi:
I will try to make the 32x32 button icon tomorrow.

Garanas:
image

Looking good in my opinion 😄

PR #3648, opened 2022-01-18
Adds map version to tooltip in scoreboard

Garanas:
As part of the [Small suggestions topic](https://forum.faforever.com/topic/3206/small-suggestions-topic) but was send to me privately.

Todo:
 - [x] Proper LOC and Russian translation


Garanas:
image

image

Garanas:
They're nil because it is skirmish (no replay ID) and a GPG map (no map version).

PR #3653, opened 2022-01-27
Fix single player delay

Garanas:
In favor of https://github.com/FAForever/fa/pull/3345 .

Instead of having it on with no choice it is now an option in the option menu. It allows for AI-only games to run at maximum-maximum speed. Useful for the AI-match tool that Softles (Softly) made to test larger PRs more effectively. The game needs to be restarted in order to make use of the adjusted option.

Closes issue #3344 .


Garanas:
I'm going to remove this as an option and add it to the command line instead. Only developers need this - any sane user wouldn't toggle that option.

PR #3657, opened 2022-02-02
Introduce an ETA on economy board

Garanas:
Based on [this forum topic](https://forum.faforever.com/topic/3336/mass-energy-storage-full-depleted-eta-timers).

image

image

image

And a [video on Youtube](https://www.youtube.com/watch?v=VJumnxGCkDY).

A few possible improvements:
 - Remove the word ETA and center the time. It is clear from the context
 - Compute over various samples of time to prevent reclaim from causing hiccups
 
 Other changes:
  - When an economy number is more than 1000 it is shortened. As an example: 1520 is 1.5K
  - When an economy number is more than 10000 it is shortened even further. As an example: 15200 is 15K
  
  This guarantees space for the ETA timer.
  
And a few todos:
 - Make a toggle for it 😄


bcdrme:
"ETA" could be changed to something more meaningfull : deplets in, or something like that.

Garanas:
I agree - that is already in. Let me do a commit and a screenshot

Garanas:
image

By all means, the positioning of the UI needs to be fixed still 😄

Nojoke185:
rather than "Depleted in" maybe the phase "Empty in" would be a better complement?
i.e. full vs empty are more usually opposite terms in English

either way nice addition

bcdrme:
We should ask a native speaker tbh

MadMaxFAF:
empties in

Garanas:
I think this still has value, but we'll have to look at it another time.

PR #3658, opened 2022-02-03
Refactoring connectivity windows

Garanas:
Refactors the connectivity windows to:
 - (1) Improve performance, no more GetSessionClients() each frame
 - (2) Improve readability
 - (3) Improve communication
 
(2) is done by appending the ping with a box plot: we sample the ping over X seconds and compute some statistical data over it. We visualize this with a small box plot. The tooltip will contain information about what is seen in the box plot. This should give you a better idea of how the connection between you and someone else performs.

(3) is done by adding a button that allows you to send your ping information to the game. As an example, if you suspect someone is lagging then you can click this button and the boxplot is translated to some string that represents the relevant information.


Garanas:
This PR is delayed until #3659 is merged.

Crotalus:
Will the new connection window show "jitter" i.e. min / max values of the ping? Fluctuating ping often indicates loss of packets / lack of bandwidth while a stable 400 ping could be completely fine.

Garanas:
Yes - the idea is this:

image

But then with a better UI. The box plot will be more informative about how the ping evolves over time. Left is 0 ping, right is 600 ping. The tooltip will contain a description of the boxplot so that people can make their own interpretations. It will also contain the fact that any (stable) ping < 500 is fine, including all those poor Australians 😄 .

Garanas:
Closing due to the introduction of the new UI build pattern, which this PR doesn't use.

PR #3660, opened 2022-02-06
Dynamic AI threat calculations

Garanas:
Introduces the dynamic computation of AI threat values. These have not been changed over the past few years after balance changes. With this pull request the threat values are computed on the fly during the blueprint modding process.


lua/system/blueprints-ai.lua
```diff
@@ -0,0 +1,249 @@
...
+local MathFloor = math.floor
+
+local StringFind = string.find
+
+
+local function TakeIntoAccountBuildrate(bp)
+    if not bp.Economy.BuildRate then
+        return false
+    end
+    if TableFind(bp.Categories, 'HEAVYWALL') then
```
The-Balthazar, line 18: This should be changed to 'WALL' instead of 'HEAVYWALL'.

It was orignally added to exclude the T3 walls from my mod from this calculation, but they have since been reworked and the 'HEAVYWALL' category was removed. Otherwise you get my T3 walls with like 800 economic threat.

The-Balthazar: It could also be merged with the previous check into
```lua
if not bp.Economy.BuildRate or arrayFind(bp.Categories, 'WALL') then return end
```
And assuming this happens *after* the categories are hashed, it could be updated to check that instead of looping through the whole array.

Azraeel:
Thanks Jip & Balthazar, This helps AI in so many ways its hard to describe 🕺

lua/system/blueprints-ai.lua
Garanas: These threat values are the very last computations on the blueprints before they are send to the engine

Garanas:
This is ready for merge, but there's this:

image
UEF commander

image
Aeon commander

image
Seraphim commander

image
Cybran commander

The differences between the commander is quite severe.

PR #3662, opened 2022-02-13
Dynamic LOD settings

Garanas:
Some of the LOD values are out of balance. This can cause sudden framerate drops. This pull request scales the LOD of entities according to their size. A small entity occupies less screen space, therefore it is less detailed and can do with a lower LOD. The same (but reversed) logic can be used for a large entity.

This pull request is a result of a discussion with Strogo about [this video](https://youtu.be/dAqlgrOjy2Q).

### DIFF
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -61,6 +61,8 @@ pcall = pcall
 doscript = doscript
 DiskFindFiles = DiskFindFiles
 
+doscript("/lua/system/blueprints-lod.lua")
+
 --- Load in the pre game data that is defined in the lobby through the preference file.
 local function LoadPreGameData()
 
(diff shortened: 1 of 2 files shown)


Strogoo:
For me, it's perfect now. Tested these changes in same replay (Vsync On and OFF):
https://youtu.be/KMksnLwFXaY

Also it makes a whole picture less messy, unlike in current patch when small props appears too early and you can't recognize them anyway

Garanas:
Noticed the quadratic nature of the values I had in the table - made it into a function so that we support any amount of LODs on an entity.

Strogoo:
Tested some maps, on Adaptive Pyramid for example, these small rocks appears too late as for my taste. But since we found out how LOD console commands works, we can add a LOD graphical settings that actually work :). Can even make a UI mod where you can set default LOD for every map and so on.

And default bp values should stay reasonably low for better performance.

Garanas:
We could also add a second scale argument: the (mass) value of the prop

PR #3667, opened 2022-02-20
Disable mesh rendering for minimap

Strogoo:
I'm currently working on ability to change LOD modifiers for different type of objects via console and one thing that bothered me is why mesh renderer code that makes LOD calculations for visible objects is called so often, as if there are thousands objects in view, even though I'm zoomed in and there is 1 unit and a few props on the screen.

The answer is simple - minimap. Why bother adding simplified renderer for minimap, if you can just copy-paste main camera implementation and apply fancy cartographic view on it :+1: And things become even more dramatic if you accidentally zoom in and minimap starts to  render all these props and units.

Anyway, with this engine patch https://github.com/FAForever/FA-Binary-Patches/pull/7 `cam_DefaultMiniLOD 0` command will disable minimap mesh renderer completely. As a result minimap becomes an actual minimap, stops eating your fps and shows only 2 things that you expect from it: the map itself and unit icons. The option is added to minimap settings and it's ON by default:
Screenshot_9

Exe for testing:
https://mega.nz/file/QYI3gCyJ#JVUH0BHUFEPG_arlgPE0PMyR9yvRuptxXk1_EbogAY8

### DIFF
--- a/lua/ui/help/tooltips.lua
+++ b/lua/ui/help/tooltips.lua
@@ -2043,6 +2043,9 @@ Tooltips = {
         title = "<LOC tooltipui0629>Save Template",
         description = "<LOC tooltipui0630>Creates construction template by saving units/structures and their position",
     },
+    minimap_mesh = {
+        description = "<LOC tooltipui0720>Disables mesh rendering for minimap. Greatly improves performance",
+    },
 
     -- **********************
     -- ** GazUI
--- a/lua/ui/help/tooltips.lua
+++ b/lua/ui/help/tooltips.lua
@@ -2044,7 +2044,7 @@ Tooltips = {
         description = "<LOC tooltipui0630>Creates construction template by saving units/structures and their position",
     },
     minimap_mesh = {
-        description = "<LOC tooltipui0720>Disables mesh rendering for minimap. Greatly improves performance",
+        description = "<LOC tooltipui0720>Disables the rendering of meshes of units and props on the minimap to reduce the burden on the rendering thread. Can potentially improve your framerate.",
     },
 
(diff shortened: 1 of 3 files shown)


lua/ui/help/tooltips.lua
```diff
@@ -2043,6 +2043,9 @@ Tooltips = {
         title = "<LOC tooltipui0629>Save Template",
         description = "<LOC tooltipui0630>Creates construction template by saving units/structures and their position",
     },
+    minimap_mesh = {
+        description = "<LOC tooltipui0720>Disables mesh rendering for minimap. Greatly improves performance",
```
Garanas, line 2047: Perhaps change it to: `Disables the rendering of meshes of units and props on the minimap to reduce the burden on the rendering thread. Can potentially improve your framerate.`

It appears to have no impact on my game, but my framerate appears to be capped at 100 even though VSync is off.

Garanas:
Tested it with the executable, works as intended. I'll leave the assembly analysis for Kionx.

lua/ui/help/tooltips.lua
Strogoo: Yep, I'll change the tooltip.

Also to remove fps cap use `sc_FrameTimeClamp 1` console command

The effect isn't that noticeable on empty map. But in real game with a lot of stuff on the screen, disabling meshes will make a ren thread life a lot easier :)

Garanas: Ah - yes. You're correct, as expected rendering less things improves the framerate.

Garanas:
The assembly changes have been merged

PR #3668, opened 2022-02-20
Improve weapon state when owner dies

Garanas:
Solves two open issues with weaponry:
 - The weapon manipulators were removed of a unit before the wreck was made. All weapons switched back to their default orientation when the unit dies. This looks bad. The wreck then copies the default location.
 - Trash of weapons were not removed properly. This could leave potential lingering beams when units die. This is no longer the case.


Garanas:
image

As an example, the weapon remains its orientation when the Spider Bot dies.

PR #3676, opened 2022-02-23
Add the Lua 5.1 functions string.match and string.gmatch

The-Balthazar:


Askaholic:
Add tests!

Garanas:
I agree - these are perfect candidates for a few unit tests.

KionX:
The 5.1 documentation says: "Function string.gfind was renamed string.gmatch"

The-Balthazar:
I missed that. I guess I'll just have gmatch redirect to that then.

PR #3681, opened 2022-03-01
Optimize shield interaction

Garanas:
A small change that limits the number of impact effects that a shield can produce. Impact effects consist of the allocation of an entity and various emitters. This can be quite expensive:
 - 100 mantis firing on the ground takes up to 5 ms / tick 
 - 100 mantis firing on a shield takes up to 10 ms / tick
 
With this pull request, firing at a shield with 100 mantis is just about as expensive as firing on the ground. This is particularly relevant when units produce multiple damage instances.


Garanas:
here-is-gif-9
This is the default visual when 20 janus targeting a shield. I had to reduce the resolution, but at the end it shows that the time taken per tick is about 10 - 12 ms

here-is-gif-10
This is with the changes of this pull request. Frame count is lower because of file constraints. The effects are still there at a nice maximum 3.5ms / tick.

Garanas:
To improve consistency we can add a damage factor. As an example, 'losing' the impact effect of a mantis doesn't look odd, but ' losing' the impact effect of a strategic bomber is most certainly odd.

Marc-Spector:
When you try to make out the text in a gif 😄 

image

Garanas:
I'll do some proper comparisons tonight, this pull request requires a good review to make sure the behavior of the overspill damage is the same. The performance gains are quite significant, on my laptop from 120ms / tick to ~ 20ms / tick.

lua/shield.lua
```diff
@@ -86,9 +107,115 @@ Shield = Class(moho.shield_methods, Entity) {
...
+                -- debugUnits = { }
+                -- for k = 1, head - 1 do 
+                --     debugUnits[k] = self.OverlappingShields[k]
+                -- end 
+
+                -- keep track of the number of adjacent shields
+                self.OverlappingShieldsc = head - 1
+            else 
+                -- no units found
+                self.OverlappingShieldsc = 0
```
Sheikah45, line 211: Might be better to be explicit and call it OverlappingShieldsCount

PR #3685, opened 2022-03-04
Improve collision interactions of projectiles

Garanas:
Improves the performance of checking whether projectiles should collide or not.

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -263,26 +263,6 @@ Shield = Class(moho.shield_methods, Entity) {
         return math.min(self:GetHealth(), amount)
     end,
 
-    OnCollisionCheckWeapon = function(self, firingWeapon)
-        local weaponBP = firingWeapon:GetBlueprint()
-        local collide = weaponBP.CollideFriendly
-        if collide == false then
-            if not (IsEnemy(self.Army, firingWeapon.unit.Army)) then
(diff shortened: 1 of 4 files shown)


Garanas:
image
Previous implementation, give or take 11.5 ms / tick with a base line of 2 ms (only 147 mantis remaining)

image
Current implementation, give or take 9.5 ms / tick with a base line of 1.8 ms (only 150 mantis remaining)

This one is tricky to benchmark because it also depends on the number of allied-intersections that are happening.

Garanas:
image
Previous implementation, roughly 14 - 15ms / tick with a base line of 1.9 ms (150 mantis and an ACU)

image
Current implementation, 10 - 11 ms / tick with a base line of 2.1

This is a better test, where all mantis are on top of one another. This means that each projectile intersects with all mantis. By all means, the test is not representative for an actual game. But regardless, a 20% to 40% performance boost is decent. This pull request is based on this wiki page from a few years back:
 - https://github.com/FAForever/fa/wiki/LUA-Blueprint-and-Category-lookup-speeds

Marc-Spector:
That's great! Good job!

Garanas:
Not compatible with #3684 , yet.

Garanas:
Supersedes #3684 .

PR #3686, opened 2022-03-04
Improve performance of unit sound interactions

Garanas:
According to commit https://github.com/FAForever/fa/commit/dfb363dcdc6685bdf4daf8f94178a7da2b06ae8f the sound of submarine units do not work. The solution introduced is applied to all units and allocates a handful of entities per unit. These allocations are expensive!

By default, units try to play the sound themselves. For submarine units we introduce one entity that plays all the sounds of the submarine. This appears to work fine - note that sound is important, but during the late accurate sound is by far less critical than having a high performing simulation.

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -263,26 +263,6 @@ Shield = Class(moho.shield_methods, Entity) {
         return math.min(self:GetHealth(), amount)
     end,
 
-    OnCollisionCheckWeapon = function(self, firingWeapon)
-        local weaponBP = firingWeapon:GetBlueprint()
-        local collide = weaponBP.CollideFriendly
-        if collide == false then
-            if not (IsEnemy(self.Army, firingWeapon.unit.Army)) then
-                return false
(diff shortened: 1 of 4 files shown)


lua/shield.lua
```diff
@@ -263,26 +266,6 @@ Shield = Class(moho.shield_methods, Entity) {
...
-            for _, v in pairs(weaponBP.DoNotCollideList) do
-                if EntityCategoryContains(ParseEntityCategory(v), self) then
-                    return false
-                end
-            end
-        end
-
-        return true
-    end,
-
```
Sheikah45, line 285: Was this meant to be added to this PR?

Garanas: no, only the last commit. But this pull request will require something #3685 introduces. I'll rebase it once that is merged.

Garanas: Conclusion: this was not a good approach. tons of conflicts now 😄

PR #3688, opened 2022-03-05
Optimize shield interactions

Garanas:
On each hit against a shield a thread is de-allocated and then re-allocated: the regeneration thread. When dealing with units like a Mantis or a Janus this happens many times over per second. That is a bit of a waste! 

With this pull request we create one regeneration thread when the shield starts existing and suspend it accordingly. The ability to suspend a thread is not widely used in the code base.

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -302,10 +302,10 @@ Shield = Class(moho.shield_methods, Entity) {
             end
         end
 
-        -- damage correction for overspill
+        -- damage correction for overspill, only apply to bubble shields
 
         local instigatorId = (instigator and instigator.EntityId) or false
-        if instigatorId then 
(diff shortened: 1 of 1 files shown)


Garanas:
image
Current situation on FAF Develop, 100 mantis against 1 shield: roughly 7.2 ms with a base line of 1

image
Current situation with this pull request, 100 mantis against 1 shield: roughly 6.5 ms with a base line of 1

By all means, this is not a gigantic boost for the simulation. Allocations are expensive, but do not immediately impact the simulation. De-allocations can be painful however, as the garbage collector kicking in can cause stutters in-game.

Garanas:
Also further reduces the 5 Janus and 30 hover shields: as low as 10 ms / tick with a baseline of 1 ms.

PR #3690, opened 2022-03-08
add build mode hotkeys for FAF default mods

unbring:
adds hotkeys for sparkies, engymod support factories, and t3maa included with FAF by default

### DIFF
--- a/lua/ui/game/buildmodedata.lua
+++ b/lua/ui/game/buildmodedata.lua
@@ -228,7 +228,8 @@ local UEFT1Air = {
 }
 
 local UEFT2Air = {
-    ['F'] = 'dea0202',
+    ['J'] = 'dea0202',
+    ['F'] = 'xel0209',
     ['E'] = 'uel0208',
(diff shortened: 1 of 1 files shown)


lua/ui/game/buildmodedata.lua
```diff
@@ -82,6 +82,7 @@ local AeonT3Land = {
     ['E'] = 'ual0309',
     ['O'] = 'ual0303',
     ['R'] = 'ual0304',
+	['N'] = 'dalk003',
```
Garanas, line 85: Indenting issue

Garanas:
This seems good -  I'll test it later this week.

lua/ui/game/buildmodedata.lua
Garanas: Interesting, that appears to be a bug of Github 😄

Garanas:
I'd like the support factories to work too, but I didn't manage to make it work. If you happen to do so in the future, please make an additional pull request.

PR #3695, opened 2022-03-11
Allow Colossus to retarget if the target is no longer viable

Garanas:
Pull request is based on a small suggestion from [KonjikiNoYami](https://forum.faforever.com/user/konjikinoyami) of [this post](https://forum.faforever.com/topic/3206/small-suggestions-topic/56?_=1647005342680).

The issue is that the colossus can try to engage targets that are behind it. This causes the beam to go all over the place. This pull request resets the target of each individual weapon of the colossus if the target is considered to be behind the colossus. This sounds like a good fix, but there's a catch: the weapon tries to target the units closest to its current target. And that one is right on the edge again, hence and repeat.

@The-Balthazar is there an option to force the GC to pick a target in front of him?


The-Balthazar:
I always find weapons fiddley and unreliable when you want something specific, but the Heading Arc bp values might give you what you're after.

PR #3704, opened 2022-03-16
Lazyvar imports

4z0t:
Remove multiple lazyvar imports in UIutils.lua


Garanas:
Note: the scope of these changes is only executed when the file is imported for the first time.

4z0t:
> Note: the scope of these changes is only executed when the file is imported for the first time.

@Garanas But still, its more readable.

Issue #2380, opened 2018-02-14
Player color: "order green" is broken


Geosearchef:
When another player picks order green in the lobby and you are therefore unable to select it, the next time this color is freed (the player that picked it switches), it does not get added to your color dorp down.

When a user switches to lime this also sometimes causes it to show up in your dropdown while that user still has the color.

PR #3705, opened 2022-03-16
Bugfix in-lobby color selection drop-down lists

Emperor-Penguin1:
Made it so the in-lobby color selection drop-down lists will be accurate.  They previously would be inaccurate in many instances (ie: showing unintended numbers of colors, which did not properly match the colors available).

Also removed some extra spacing.

While additional performance improvements could be made to the lobby's color-related functionality, that is largely beyond the scope of this PR, as the purpose of this PR is to fix some buggy behavior, and it runs fine.


Emperor-Penguin1:
Btw, I think this PR resolved issue #2380.

PR #3706, opened 2022-03-16
Add colors, reorder colors, add color order functionality, & fix lobby AI color bug

Emperor-Penguin1:
Adjusts which colors are available
Reorders available colors
Adds a ColorOrder local to lobby.lua that specifies the default color order for TMM/etc that gets used if color is not specified otherwise
Theoretically fixes TryAddPlayer's IsColorFree functionality, which theoretically used to give a false negative every time it was positive

Available Colors:
image

Default Color Order:
image

### DIFF
--- a/lua/GameColors.lua
+++ b/lua/GameColors.lua
@@ -15,7 +15,7 @@ GameColors = {
         "ff8e7cc3",      -- light purple
         "ff9161ff",      -- purple
         "FF5F01A7",      -- dark purple
-        "ff920092",      -- dark rich purple
+        "ff920092",      -- rich purple
         "ff901427",      -- dark red
         "ffff88ff",      -- pink
         "ffff32ff",      -- new fuschia
@@ -43,7 +43,7 @@ GameColors = {
         "ff8e7cc3",      -- light purple
         "ff9161ff",      -- purple
         "FF5F01A7",      -- dark purple
-        "ff920092",      -- dark rich purple
+        "ff920092",      -- rich purple
         "ff901427",      -- dark red
         "ffff88ff",      -- pink
         "ffff32ff",      -- new fuschia
--- a/lua/GameColors.lua
+++ b/lua/GameColors.lua
@@ -13,10 +13,12 @@ GameColors = {
         "ff1a9ba2",      -- dark cyan
         "ff6fa8dc",      -- sky blue
         "ff8e7cc3",      -- light purple
+        "ff9161ff",      -- purple
         "FF5F01A7",      -- dark purple
         "ff920092",      -- dark rich purple
         "ff901427",      -- dark red
         "ffff88ff",      -- pink
+        "ffff32ff",      -- new fuschia
         "FFe80a0a",      -- Cybran red
         "FFFF873E",      -- Nomads orange
         "ffb76518",      -- new brown
@@ -39,10 +41,12 @@ GameColors = {
         "ff1a9ba2",      -- dark cyan
         "ff6fa8dc",      -- sky blue
         "ff8e7cc3",      -- light purple
+        "ff9161ff",      -- purple
         "FF5F01A7",      -- dark purple
         "ff920092",      -- dark rich purple
         "ff901427",      -- dark red
         "ffff88ff",      -- pink
+        "ffff32ff",      -- new fuschia
         "FFe80a0a",      -- Cybran red
         "FFFF873E",      -- Nomads orange
         "ffb76518",      -- new brown
--- a/lua/GameColors.lua
+++ b/lua/GameColors.lua
@@ -1,49 +1,62 @@
+-- Default color order for TMM/etc is set in conjunction with these tables,
+-- and is specified in the local ColorOrder in lobby.lua.  So, if you change
+-- the colors/order in the ArnyColors and or PlayerColors tables, you should
+-- probably change the ColorOrder table in lobby.lua accordingly.
+
 GameColors = {
     CivilianArmyColor = "BurlyWood",
 
     -- Faction colours
     ArmyColors = {
-    "ff436eee",      -- new blue1
-    "FFe80a0a",      -- Cybran red
-    "FF2929e1",      -- UEF blue
-    "ff901427",      -- dark red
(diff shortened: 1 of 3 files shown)


Sheikah45:
@BlackYps had recently changed the colors so that TMM is warm vs cool so this should probably keep that intact

Emperor-Penguin1:
Yes, I looked over BlackYps' PR when I was deciding the order of the colors for the ColorOrder local.  The order I settled on is largely based on a combination of doing reddish vs blueish teams and making the first several colors particularly distinct,

BlackYps:
I find this new color choice for the matchmaker way less appealing visually. Also, I consciously excluded green as green is used by the idle engie mod, so it works less good if your icons are also green. What is bad about the original choice?

Emperor-Penguin1:
I am open to changes, but I did put thought into this:

The above color order includes a near-match for each of the 6 primary light/pigment colors in the first 8 slots, contains quite distinct colors, and has a reddish vs blueish theme.  I found that combination more desirable than the current order.

By comparison, the current order contains colors that are less distinct, contains fewer near-matches to primary colors, has cybran red and nomads orange, which are similar enough that it's arguably questionable to even keep both as color options.

IIRC, I actually used 4 of the current top 8 colors in the top 8, switched out two colors for near-matches to primary colors, and added two aesthetically desirable new colors that didn't exist in the previous list.

BlackYps:
Serious question: do you have some sort of color blindness? Cybran red and Nomads orange do not seem similar to me at all.

Emperor-Penguin1:
No (note that different people can have different qualities of monitors with different settings though); when I asked for feedback on the whole color list, asking if any colors were too similar to each other, some people expressed dissatisfaction with cybran red and nomads orange being too similar to each other, saying things like, "I've had the 2 oranges be an issue" (multiple people actually referred to them as the two oranges).  Given the level of dissatisfaction and their similarity, I would've probably removed or altered 1 of them if it was a new color.

The colors labelled cybran red and nomads orange have relatively similar hues, saturations, and values (brightness).  They also contain a very similar very high amount of red.  That combination of factors, especially the very bright very strong red similarity, is arguably overpowering compared to their main difference of a different amount of green and to a smaller extent blue.

image
image

BlackYps:
For reference these are the current colors, but in this order they are only available on faf develop. faf gamemode still uses the old order that didn't really distinguish the two 4v4 teams into coherent color schemes.
grafik

I think people meant these colors instead of cybran red and nomads orange when they were referring to the two oranges. I don't know why you are sure they meant red instead.
grafik

Emperor-Penguin1:
When I compared to the current order above, I was comparing to the current order on FAF develop, which is what you posted.  I also showed that order with others for comparison when asking for feedback.  I also showed in-game examples of how icons looked with the different colors when asking for feedback on if anything was too similar.  The significant in-game similarity between cybran red and nomads orange seemed quite clear.  I think it's significantly easier to distinguish that brown from nomads orange than cybran red from nomads orange.

4z0t:
as I know @Tagada14 is colorblind, ask him about this PR

BlackYps:
This starts to touch on a lot of different aspects and I would like to see them separated so we can discuss them individually.
So far I see:
1. adding new colors
2. changing existing colors
3. reordering them in the color picker in the lobby
4. changing which colors get used in matchmaker games
5. providing functionality to choose which color to use in matchmaker games

I'm in general fine with 1. and 3., 2. and 4. would need to be discussed more and I think 5. is a typical case where more choice for the individual leads to a worse experience for everyone collectively.

Emperor-Penguin1:
@BlackYps 
Regarding #2, I don't recall changing existing colors.  I did remove a small number of colors with too much similarity to other colors.  I could add them back if people want, but when I crowd-sourced feedback, people were in favor of them being removed.
Regarding #4, I already explained the proposed combination's merits compared to the current order.  I'm not opposed to it being further refined in the future.  (I could change it to the current order if necessary, but that seems like a step in the wrong direction.)
Are you sure #5 is actually included in this PR?
Autolobby.lua seems to source player color from playerInfo.PlayerColor = lobby.ColorOrder[slot], which refers to the preset color order set in the ColorOrder local.

BlackYps:
Removing a color and putting in a new one is effectively the same as changing one.
I am still convinced that you mistakenly thought that people were referring to red and orange when pointing out that the two oranges were too similar. Maybe your monitor color settings differ from mine?
I chose colors to form a warm vs. cold team because I think it is more important to quickly recognize if a unit is in your or the enemy team while you try to have maximum distance between the original colors. (Not sure why you dissect the color wheel into six segments and then need to queeze two more into them instead of directly dividing it into eight parts, but that is a detail.)
Obviously we won't convince the other one of our solution and in general we should get more input, so I propose we take this to the forum with both versions and see what people have to say about it.
No, 5. is not yet included in this PR but what I gathered from you on discord you were thinking about how to add this so we might as well now talk about if this is a good idea.

Garanas:
I agree with BlackYps: adding in new colors is fine. But changing existing colors (in particular for ladder) does not immediately sound beneficial to me. I also think that number 5 (that is not part of this pull request) is fixing something that is in my opinion not broken: it seems best to me to have a default set of colors each ladder game.

Emperor-Penguin1:
Alright, I figure that reordering and removals can be handled in future PR's.  So, I readded the removed colors.  So, this PR no longer removes existing colors from the list (nor does it change them), and I also changed the color order to match the current color order for the top 8 slots.  So, this PR no longer changes which colors gets used in TMM games with up to 8 players.

Garanas:
I'll review this tomorrow.

Garanas:
The `ColorOrder` is tightly coupled with the `ArmyColors` / `PlayerColors` tables - moved it in proximity to those, file-wise. Prevents importing the regular lobby in the auto lobby.

Garanas:
This is the default order with this pull request:
image

Garanas:
The drop down list to choose a color closes when another player joins / exits / chooses a color of their own. That isn't quite user friendly.

Emperor-Penguin1:
I don't think that problem is caused by this PR.

PR #3707, opened 2022-03-17
UI refactor

4z0t:
_Something's wrong i can feel it..._
* Use **not IsDestroyed(...)** for controls existing check instead null check
* LayoutHelpers preferable instead direct Set
* **self** in methods (**control** if **self** in outer scope)

### DIFF
--- a/lua/ui/maputil.lua
+++ b/lua/ui/maputil.lua
@@ -57,11 +57,7 @@ end
 
 -- given a scenario, determines if it can be played in skirmish mode
 function IsScenarioPlayable(scenario)
-    if not scenario.Configurations.standard.teams[1].armies then
-        return false
-    end
-
-    return true
+    return not not scenario.Configurations.standard.teams[1].armies
 end
 
 -- EnumerateScenarios returns an array of scenario names
--- a/lua/ui/menus/eula.lua
+++ b/lua/ui/menus/eula.lua
@@ -3,7 +3,7 @@
 --* Author: Ted Snook
 --* Summary: Front end display of the EULA
 --*
---* Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright ï¿½ 2005 Gas Powered Games, Inc.  All rights reserved.
 --*****************************************************************************
 
 local UIUtil = import('/lua/ui/uiutil.lua')
@@ -33,7 +33,7 @@ function CreateEULA(inParent, callback)
         end
     end
 
-    UIUtil.MakeInputModal(parent, function() exitButton.OnClick(exitButton) end, function() exitButton.OnClick(exitButton) end)
+    UIUtil.MakeInputModal(parent, function() exitButton:OnClick() end, function() exitButton:OnClick() end)
 
     local title = UIUtil.CreateText(bg, "<LOC EULA_TITLE>End User License Agreement", 20)
     LayoutHelpers.AtHorizontalCenterIn(title, bg)
--- a/lua/ui/notify/customiser.lua
+++ b/lua/ui/notify/customiser.lua
@@ -58,23 +58,21 @@ local function EditMessage(parent, data, line)
     helpText:SetText(clarityTable[data.source])
     helpText:SetCenteredHorizontally(true)
 
-    messageEntry = Bitmap(dialogContent)
+    local messageEntry = Bitmap(dialogContent)
     messageEntry:SetSolidColor('FF282828')
-    messageEntry.Left:Set(function() return dialogContent.Left() + 15 end)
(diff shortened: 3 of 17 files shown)


lua/ui/maputil.lua
```diff
@@ -57,11 +57,7 @@ end
 
 -- given a scenario, determines if it can be played in skirmish mode
 function IsScenarioPlayable(scenario)
-    if not scenario.Configurations.standard.teams[1].armies then
-        return false
-    end
-
-    return true
+    return not not scenario.Configurations.standard.teams[1].armies
```
HUSSAR-mtrela, line 60: this should be simplified to:
```return scenario.Configurations.standard.teams[1].armies```

File: lua/ui/notify/customiser.lua [hunk 2]
```diff
@@ -171,16 +155,15 @@ local function SelectLine(dataIndex)
...
-    if not bgSize then bgSize = 20 end
-    if not bgColor then bgColor = 'FF343232' end
-    if not txtColor then txtColor = UIUtil.categoryTextColor end
-    if not txtSize then txtSize = 18 end
-    if not txt then txt = '?' end
+    bgSize = bgSize or 20
+    bgColor = bgColor or 'FF343232'
+    txtColor = txtColor or UIUtil.categoryTextColor
+    txtSize = txtSize or 18
+    txt = txt or '?'
```
HUSSAR-mtrela, line 162: good job!

lua/ui/uiutil.lua
```diff
@@ -980,15 +979,16 @@ function CreateWorldCover(parent, colorOverride)
         local index = i
         if GetFrame(index) == parent:GetRootFrame() then
             worldCovers[index] = Bitmap(parent)
-            worldCovers[index].ID = index
-            worldCovers[index].OnDestroy = function(self)
+            local worldCover = worldCovers[index]
```
HUSSAR-mtrela, line 982: defining `worldCover` locally on line 982 and again `worldCover` on line 1001 might lead to bugs if someone refactors these code or try to understand it. I suggest to renaming one of these local variables:
```local worldCover = worldCovers[index]```
to something like this:
```local frameCover = worldCovers[index]```

**Review by HUSSAR-mtrela — CHANGES_REQUESTED** [review]

File: lua/ui/maputil.lua (hunk 1 above)
Comment (4z0t) [reply lua/ui/maputil.lua:60]: It must return bool, so, I cheated 😁

File: lua/ui/uiutil.lua (hunk 3 above)
Comment (Garanas) [reply lua/ui/uiutil.lua:982]: I agree.

**Comment by Garanas** [comment]
I'd like to note that I am in favor of several smaller pull requests instead of one larger pull request.

**Comment by 4z0t** [comment]
I don't think it's gonna be that large, currently just removing strange and useless stuff and changing code for better readability

File: lua/ui/maputil.lua (hunk 1 above)
Comment (ChrisKitching) [reply lua/ui/maputil.lua:60]: Maybe consider a function like `Bool()` to perform the conversion, then? Javascript-style implicit conversion to boolean may cause engineer killing sprees.

File: lua/ui/menus/eula.lua [hunk 4]
```diff
@@ -3,7 +3,7 @@
 --* Author: Ted Snook
 --* Summary: Front end display of the EULA
 --*
---* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
```
ChrisKitching, line 6: You saved this file with non-UTF-8 encoding, breaking this character. Reconfigure your editor appropriately ;)

4z0t: it is utf-8
image

4z0t:
I also know why would chat crash if you spamed messages a  lot

Rackover:
> I also know why would chat crash if you spamed messages a lot.
Because Mr [@Rackover](https://github.com/Rackover) decided to create for each message received a control...

it's Ms 😂   and do know that I receive this kind of ping by mail directly, so editing my name out doesn't change much...
but yes, i believe this is & was my only PR that ever dealt with Lua code , and supcom's Lua API is not one of my strong suits   Feel free to refactor and shuffle around in the way you feel is fit! ❤️

4z0t:
Yeah, I think I will make new pr for this one, because it requires significant changes

lua/ui/menus/eula.lua
ChrisKitching: Huh, so why'd this break? :D

4z0t:
Gonna separate on smaller PRs to prevent overgrowing

Garanas:
Sounds better 😄 .

PR #3709, opened 2022-03-20
Remove duplicates in files

4z0t:


lua/ui/game/nukelaunchping.lua
```diff
@@ -38,7 +38,7 @@ function DoNukePing(NukeLaunchData)
         end
 
         if pingOkFlag then
-            local data = {Owner = army, Type = pingType, Location = position, Type = pingType}
+            local data = {Owner = army, Type = pingType, Location = position}
```
Garanas, line 41: How is this duplicated information?

units/XRS0205/XRS0205_unit.bp
```diff
@@ -53,7 +53,7 @@ UnitBlueprint {
         'OVERLAYRADAR',
         'OVERLAYSONAR',
     },
-    CollisionOffsetY = -0.15,
+    -- CollisionOffsetY = -0.15,
```
Garanas, line 56: I recall we discussed this about whether you should comment the duplications or just remove them. Looking at the pull request, for clarities sake I think it is better to just remove them.

4z0t:
I haven't removed duplicates because some of them require a decision about which one to choose. And @Garanas there is a **Type=pingType** literally typed twice...

Issue #3638, opened 2021-12-28
Dialog to inform of base-giving in full share games


Sheeo:
Proxy-issue from Ninrai.

His description:

If a teammate dies in a full share game, create a popup info (similar to when a t4 finishes) to notify the player who received the dead ally's units / buildings about the additional base and units in order to avoid confusion/idle buildings and units.

Garanas:
We can do that - what shape should it have? A pop up (that you must interact with) seems undesirable. Would a simple text message / whisper suffice?

Sheeo:
Ninrai's comment

> I think in order to keep it simple, a simple pop up info like with the t4 for the one given the stuff would suffice. nothing with clicking. maybe it should look a little different from the t4, more popping. possibly with a different color (red?) to highlight the urgency for a reaction by the player

I'd add personally I think it should be configurable so experienced players can turn it off if they want :)

Garanas:
Always a toggle 😄

4z0t:
I would suggest to do a User Sync info about units been given at all, for mod purposes )

Garanas:
Can you clarify on that? To allow a mod to hook it?

4z0t:
There are no ways no determine whenever you were given a unit on UI level

Garanas:
Ah, I see. What type of UI interaction / callback would you want here?

4z0t:
an if statement in `UserSync.lua` mb

Garanas:
And then for each unit given, or for all units given at once?

4z0t:
A way which wont impact on performance, i think. 😁

Garanas:
And do you want the unit IDs too? Or just a content: X gave Y Z units

4z0t:
The fact that units were given is enough, i think, but i would  appreciate if its possible to provide with  ids too 😊

PR #3712, opened 2022-03-22
Transfer units improvements

Emperor-Penguin1:
Creates info text of "[name]'s units have been transferred to you" for the player who receives units when an army's units are transferred upon player death (such as from a dead teammate in a fullshare game)

Changes priority for unit transfer to be based on rating (and if rating does not exist (such as with regular AI's), by score, after players with positive rating), instead of just by score


Emperor-Penguin1:
PS: this resolves the core goal of issue #3638, among other things.  Additional improvements could be added later, especially if additional UI-Sim interaction is added (presumably involving UserSync), but I think the current functionality in this PR is already useful enough to add.

lua/aibrain.lua
```diff
@@ -576,16 +576,29 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                            end
                             WaitSeconds(1)
                         end
                     end
                 end
             end
 
-            -- Sort the destiniation armies by score
+            -- Sort the destiniation brains (armies/players) by rating (and if rating does not exist (such as with regular AI's), by score, after players with positive rating)
             local function TransferUnitsToHighestBrain(brains)
```
Garanas, line 589: This changes more than it appears. The original behavior uses score. Your implementation uses rating. As a consequence only the highest rated players will have the joy of managing two bases at once. What is the idea behind this change?

Emperor-Penguin1: Even with the current score system, in practice, it is often that 1 player on a team receives all the bases from dead teammates because getting additional bases helps to easily cement a score lead normally.  I think giving dead player's units to the highest rated teammate makes more sense than giving them to the teammate with the highest score (aka, the one who eco'ed the most).  I think players with higher rating are generally able to handle multiple bases better on average.  In practice, the result of the current score-based priority is that when a lower-rated teammate receives the base, that player often doesn't use it much or at all, gifts it to the higher-rated teammate, and or receives flak/complaints about not gifting it to the higher-rated teammate.  This decision of whether or not to transfer the base to the higher rated teammate is often the difference between winning and losing, and the frustration regarding this contributes to toxicity as well.  By comparison, if the base is gifted to the higher-rated player by default, that is less likely to happen, and the higher-rated player can still gift the base to a teammate if desired, and the higher-rated player is generally much more likely to know how to shift-click-and-drag for easier/faster gifting.

Garanas: I understand - thank you for elaborating it.

I am not against the chance - I'd like to hear the opinions of other people to make sure we're not missing anything.

lua/aibrain.lua
```diff
@@ -576,16 +576,29 @@ AIBrain = Class(moho.aibrain_methods) {
                         local units = self:GetListOfUnits(categories.ALLUNITS - categories.WALL - categories.COMMAND, false)
                         if units and not table.empty(units) then
                             TransferUnitsOwnership(units, brain.index)
+                            if brain:GetArmyIndex() == GetFocusArmy() then 
+                                PrintText(self.Nickname .. '\'s units have been transferred to you', 30, 'ffffffff', 12, 'center')
```
Garanas, line 580: Add a russian translation, @4z0t can you help with the translation itself?

lua/aibrain.lua
Garanas: The balance team appears to agree with you

Garanas:
I'll merge this once the translation is in.

4z0t:
Wouldn't it have a UserSync version for being displayed on UI side?

Garanas:
We could add that - but at the moment it just prints a message in the center of the screen.

4z0t:
@Garanas i mean, just make a sync for it, no need to make ui right now

Garanas:
I'll add it in. Can you translate that sentence for me?

4z0t:
Yeah, of course, but string format needed

Garanas:
here-is-gif-26

PR #3717, opened 2022-03-24
Move the water / land sounds into amphibious / hover class

### LINKED ISSUE #3578: Refactor `OnMotionHorzEventChange` of the Unit class

Garanas:
Closes #3578.

Still needs a bit of work - I'll leave that to someone else to finish up as an exercise 😄 

todo:
 - [ ] Add Youtube link to first part of this pull request


lua/defaultunits.lua
```diff
@@ -2104,7 +2123,24 @@ SlowHoverLandUnit = Class(HoverLandUnit) {
 }
 
 -- AMPHIBIOUS LAND UNITS
-AmphibiousLandUnit = Class(MobileUnit) {}
+AmphibiousLandUnit = Class(MobileUnit) {
+    OnMotionHorzEventChange = function(self, new, old)
+        MobileUnit.OnMotionHorzEventChange(self, new, old)
```
healther, line 2128: I experimented a bit with the logic here and I think calling the super function is not what we actually want. I _think_ we want to handle all special sounds explicitly (there are about 20 total) and have the generic function only execute generic sounds. The problem with calling the implementation we are shadowing is, that we then either start/stop _both_ sounds (maybe the engine prevents that and only one will actually be played, didn't test) or we still require the checking logic in the general implementation.

lua/sim/Unit.lua
```diff
@@ -3128,40 +3130,28 @@ Unit = Class(moho.unit_methods) {
...
-        end
 
-        if new == 'Stopped' or new == 'Stopping' then
-            -- Stop ambient sounds
             self:StopUnitAmbientSound('AmbientMove')
-            self:StopUnitAmbientSound('AmbientMoveWater')
             self:StopUnitAmbientSound('AmbientMoveSub')
-            self:StopUnitAmbientSound('AmbientMoveLand')
+
+            self:DoOnHorizontalStartMoveCallbacks()
```
healther, line 3148: This should be under `old == 'Stopped'`

healther (commented):
As mentioned in the inline comments: I think we don't want to execute specialized functions _in addition_ to the general one, but rather _instead of_. As is it would trigger either multiple `PlayUnitSound`s or require logic in the general function that we want to get rid of.

I'm not sure if all units implement the specialized sounds. In that case I would suggest duplicating the logic in the specialized functions as a backup.

I'll write up a PR implementing my current understanding some time next week (hopefully). Let me know if I'm misunderstanding something, or if this sounds good to you.

healther:
I have a fixup commit for this at https://github.com/healther/fa/commit/d128bca13eaf8f60f2099f161a833c2de5ad0e1e but I'm not sure if what I'm doing there is actually correct...

lua/sim/Unit.lua
```diff
@@ -3114,12 +3114,14 @@ Unit = Class(moho.unit_methods) {
...
     OnMotionHorzEventChange = function(self, new, old)
+
         if self.Dead then
             return
         end
+
         local layer = self.Layer
 
-        if old == 'Stopped' or (old == 'Stopping' and (new == 'Cruise' or new == 'TopSpeed')) then
+        if old == 'Stopped' then
```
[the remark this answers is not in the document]
Garanas: I agree, there is a functional change here. We could 'move it up' one level and change all the statements to 'Stopping' instead of 'Stopped'.

lua/defaultunits.lua
Garanas: We have to call the base class. We depend on every function calling the base class when we make changes to the base class. Or when mods hook into the base class. 

A unit can play only one sound, and one ambient sound. Therefore no sounds are played twice: one simply overrides the other. This is still a slight waste by all means. We could introduce a flag into the function that defaults to nil (not set) to tell the function whether a sound is already playing. But I'm not sure if it is worth the additional logic, and mods will break that functionality again when hooking with the old function signature.

PR #3718, opened 2022-03-24
Optimize Zthuee projectiles (amongst other projectile things)

Garanas:
Our benchmark is 400 Zthuee firing at once on the map White Fire. The map has no civilians. All functionality of the original projectile is still there, just faster. A base line is the same scenario, but without the Zthuee firing.

On the FAF branch: `12 - 14 ms` with a base line of `3 ms`
here-is-gif-15-faf

This pull request: `9 - 12 ms` with a base line of `3 ms`
here-is-gif-15-develop

This pull request, without decals: `9 - 10.5 ms` with a base line of `3 ms`
here-is-gif-15-develop-no-decals

If we'd reduce the base lines:
 - FAF branch: `9 - 11 ms` for firing
 - This pull request: `6 - 9 ms` for firing
 - This pull request, without decals: `6 - 7.5 ms` when firing
 
Therefore we can safely assume that there's at least a 20% performance improvement. Some of these changes apply to all projectiles.

### DIFF
--- a/engine/Sim/Projectile.lua
+++ b/engine/Sim/Projectile.lua
@@ -18,6 +18,7 @@ end
 
 ---
 --  Projectile:CreateChildProjectile(blueprint)
+-- Internally calls import to find the projectile
 function Projectile:CreateChildProjectile(blueprint)
 end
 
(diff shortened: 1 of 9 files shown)


Garanas:
Ran tests on several games, all appears to be good.

PR #3720, opened 2022-03-25
Fix LODs of effects

Garanas:
Description
-------------
Emitters can be quite expensive on the simulation. Up to 15% of the emitters in the game render up to extreme distance. These should be culled accordingly by introducing a sane LOD cutoff in their blueprint definition.

Course of action
------------------
Claim a few projectiles by making a comment in this pull request. This prevents people from doing the same projectiles. There are a few important factors to take into account:

### LOD settings

In order to be consistent, make sure that your LOD settings are on 'high':

image

### LOD of the emitter

Effects have an impact on the simulation speed. In particular the spawning of a particle has an impact. All effects have a LOD threshold set that determines at what distance from the camera it can generate particles. The **L**evel **o**f **D**etail threshold is used to cull entities when they contribute to no more than a few pixels. The value is called `LODCutoff` in the blueprint. When the LOD is not set the emitter always renders, regardless of the distance to the camera. 

here-is-gif-16
An example is the t1 Aeon artillery, as shown in the gif: their LOD cutoffs are set to -1 by default and the unit is widely used. Therefore it can easily take up 2% to 5% of the simulation speed (2ms to 5ms / tick) during a battle, even though you're too zoomed out to actually enjoy the graphics.

```lua
EmitterBlueprint {

       -- (...)

        -- values that we're interested in
	LODCutoff = -1,                 -- distance threshold to camera when rendering
	EmitIfVisible = true,           -- emits only if visible, this should typically be set to true
	CatchupEmit = true,           -- allows emitter to catch up if it becomes visible
	CreateIfVisible = false,       -- flag to only create if the emitter is visible upon creation. This should be set to false for effects with a long duration, such as a trail. Effects that happen often such as terrain effects can have this value set to true.

        -- (...)
    },
```

### Finding the unit of an emitter

We make the assumption that you are using Visual Studio Code, as described in the [setup of the development environment](https://github.com/FAForever/fa/blob/deploy/fafdevelop/setup/setup-english.md#development-environment). 

It could very well be that an emitter is not used in the base game.  In that case you can set it to a default value of 300. Mods can still use those emitters.
 - (1) Pick a set of emitters you intend to work on. Please post your selection in this pull request to prevent duplicated work. For this example, we'll stick to `cybran_cruise_missile_launch_01_emit`. 
 - (2) Assuming that you have the repository setup as your workspace, use `CTRL + SHIFT + F` to search through all the files. use the name of the emitter (without the path!) as a search argument. This will likely take you to an entry in `/lua/EffectTemplates.lua`. In this case, it is `CIFCruiseMissileLaunchSmoke`.
 - (3) Repeat the search of the previous step but the with the new search argument: `CIFCruiseMissileLaunchSmoke`. This takes us to `CIFMissileLoaWeapon` where the effect is used as a muzzle flash. 
 - (4) Repeat the search of the previous step but the with the new search argument: `CIFMissileLoaWeapon`. This takes us to the unit class `URS0304`, which is the Cybran Strategic Missile Submarine. 
 
here-is-gif-25
_The effect in question, on the water surface_
 
### Determine the settings

There are a few things that we need to take into account:
 - Is the effect part of an projectile, or of a unit / weapon?
 - What is the size of the effect as a whole?
 
If the effect is part of a unit / weapon then one can argue:
 - `LODCutoff`: should be no higher than the LOD of the unit
 - `EmitIfVisible`: should always be true
 - `CatchupEmit`: should likely be false if the effect has a short duration (and the unit has a reasonable rate of fire)
 - `CreateIfVisible`: should likely be true if the effect has a short duration (and the unit has a reasonable  rate of fire)

If the effect is part of a projectile then one can argue:
 - `LODCutoff`: Should be no higher than the LOD of the projectile, and if none is set then it depends on the size of the effect
 - `EmitIfVisible`: should always be true
 - `CatchupEmit`: if it is part of a trail then this should be true, if it is part of an explosion then it is questionable because those are quite short
 - `CreateIfVisible`: if it is part of a trail then this should be false, if it is part of an explosion then it is questionable because those are quite short

Test plan
------------------
Make a branch based on this branch. Work on at most 20 emitters per branch. Once you've reached the threshold, make a pull request to merge back into this branch. Note that the default merge branch is set to `deploy/fafdevelop`. You'll have to adjust that. 

Make sure to mention what units / weapons / projectiles use the emitters. Your changes will be easier to review with that information. If applicable a screenshot is appreciated too.

If you're unsure about anything you can discuss it in the #game-repository channel on the [FAF Discord](https://discord.gg/mXahVSKGVb) (after giving yourself the tester role in the #role-selection channel) or you can make a comment in this pull request.

Learning goals
------------------
Feel comfortable working on the repository and with Git in general.

Pull request progress
-----------------------

....


Garanas:
Another pull request (#3719) defaults the LODs when they are not set (e.g, `LODCutoff = -1.00`) to 160. This is wrong for some effects - this pull request tries to fix that over time.

Garanas:
I made a mistake and accidentally merged the branch of @MadMaxFAF 😞 .

Garanas:
I'll update the track record once I'm back in a few days. For anyone else interested in working on this, please ask @MadMaxFAF which ones are still free.

Garanas:
I've updated the original description with the remaining effects that have an infinite LOD cutoff - only 114 to go.

Garanas:
With thanks to @MadMaxFAF for all the hard work 😃 .

PR #3721, opened 2022-03-26
Display team ratings in the observer window of the lobby

Emperor-Penguin1:
This PR adds team ratings (sum totals) for each team to the observer window.  It shows the unrounded displayed rating (mean - dev * 3) and then shows the base rating (mean) +/- uncertainty (3 * dev) in parentheses.
If there are two or fewer teams, they are listed before observers.  If there are more than two teams, they are all listed after observers.

image

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -2099,6 +2099,35 @@ end
 local function refreshObserverList()
     GUI.observerList:DeleteAllItems()
 
+    -- create the table that will hold the data for displaying team rating information
+    local teamRatings = {}
+
+    -- cycle through each player
+    for i, player in gameInfo.PlayerOptions:pairs() do
+
+        -- get the team number (which is 1 higher on the backend)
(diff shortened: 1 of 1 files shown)


Garanas:
Looks like a clean implementation. I'll give it to you - initially I was against because it would re-use the observers window / box. But now it seems better to have it, instead of waiting for the lobby to magically create more space for us 😄 .

I'll review it again tomorrow.

Garanas:
In practice the numbers have a lot of decimal points:
image

@Emperor-Penguin1 Could you make a new pull request that rounds them to integers?

Garanas:
There's a few other issues:
 - When a team is assigned manually it only updates for the host
 - When a player chooses their spot on the map it doesn't update
 
And I suspect other cases where a team switch happens (like switching two players) doesn't cause the observers window to update either.

Emperor-Penguin1:
Okay, I'll look into fixing these and will make a new PR soon.

Issue #3580, opened 2021-12-02
Refactor `CreateVeterancyBuffs` of the Unit class


Garanas:
Description
-------------
The function in question performs a lot of table allocations that in turn are used only once. This is the definition of garbage creation and when it happens too often the game starts to stutter. Solving this issue won't empty the bucket, but it will take out a small cup of water.

Course of action
------------------
Refactor the following tables by localizing them at the top of the file:

 - [ ] `techLevels`
 - [ ] `regenBuffs`

Make sure that their names make sense, as an example: `CreateVeterancyBuffsTechLevels` instead of `techLevels`. This makes it clear from the context.

Refactor the return type such that it returns two values instead of a new table. Make sure to adjust the `SetVeteranLevel` function to match the new return type. Preferred course of action is to remove the for loop all together by unrolling it over the two values.
 
Test plan
------------------
Use a single pull request. This PR is relatively easy to test with various units. If it doesn't cause errors and the veterancy system still works as intended then you're good. The optimization is not directly noticeable.

Learning goals
------------------
Feel comfortable working on the repository.

PR #3722, opened 2022-03-26
Refactor CreateVeterancyBuffs

healther:
Closes: #3580

Moves the (repeated) allocation of two tables to the global scope. This prevents
the useless repeated reallocation and garbage collection of the data structures.
From the game perspective nothing should change, from a performance POV this should
have limited effect, as this function is only called once per unit type and vet
level.


Garanas:
Looks good. 

Can you also remove the [table allocation](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L1436) when the function returns? Lua supports returning multiple arguments over the stack.

PR #3725, opened 2022-03-26
Fix LazyVars behaviour with `false` value

4z0t:
Dealing with LazyVars I recently found out that setting value of lazyvar `false` causes troubles:
First of all:
`a  = Lazyvar.Create(false)`
`LOG(a())` - returns `0` that is not `false` at all (FYI 0 is a true value in Lua, so, 
`LOG(not 0)` is false)
Secondly:
`a:Set(false)` - ok fine, now it is false
`LOG(a())` - oh no, tried to call `nil` value?
Thirdly:
`a.OnDirty = function(self) LOG('i have a new value'..self()) end` - if we set a new value this would be called
`a:Set(false)` - no, we didn't...

### DIFF
--- a/lua/lazyvar.lua
+++ b/lua/lazyvar.lua
@@ -1,0 +1,115 @@
+--
+-- LazyVar module
+--
+
+
+local EvalContext = nil
+
+local LazyVarMetaTable = { }
+
+LazyVarMetaTable.__index = LazyVarMetaTable
+
+local WeakKeyMeta = { __mode = 'k' }
+
+-- Set this true to get tracebacks in error messages. It slows down lazyvars a lot,
+-- so don't use except when debugging.
+ExtendedErrorMessages = false
+
+function LazyVarMetaTable:__call()
+    if not self[1] then
+        if self.busy then
+            error("circular dependency in lazy evaluation for variable " .. (self.trace or ''), 2)
+        end
+        self.busy = true
+        local u
+        for u in self.uses do
+            u.used_by[self] = nil
+        end
+        self.uses = {}
+        local oldContext = EvalContext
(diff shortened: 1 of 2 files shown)


Garanas:
First import the base file in a separate commit and then make the changes. It makes it easier to review 😄 .

4z0t:
i closed it somehow xdddddddd

Garanas:
I think it closed automatically because there are no file changes 🤔

Garanas:
Interesting - any idea what this means: `result = {&1&4}`?

KionX:
_Interesting - any idea what this means: result = {&1&4}?_
In LuaPlus, this indicates to preallocate memory for the table.

4z0t:
@KionX is there a detailed doc about those numbers and how it calculates the final size?

KionX:
First & is hash table size rounded up to 2^(trunc(log2(nh))+1).
Second & is array linear size.

4z0t:
Isn't this line would break all other tests right after merging?

Garanas:
Yes - we can't merge this in its current state.

4z0t:
Is there some kind of filter for running tests? Just put this file into blacklist 🙂

Garanas:
We'd need to make changes to this file:
 - https://github.com/FAForever/fa/blob/deploy/fafdevelop/run_lua_tests.sh
 
I'm fine with an exception.

4z0t:
@Askaholic can you help us with it? 😀

Garanas:
Added the exception, @4z0t is this pull request ready?

4z0t:
@Garanas sure

PR #3740, opened 2022-03-27
Seraphim Effects Cleanup

brokerjoe:
This is a work in progress PR (Draft) (Feel free to add to it or move it out of draft mode to merge).

- fixing effect utilities loops
- seraphim stored progress in global context changed to local while loop
- When building large units (Seraphim bomber, for example) the LOD of the effect doesn't match the size of the unit. This causes inconsistencies. The LOD is now always set according to the LOD of the mesh of the unit.


lua/EffectUtilitiesSeraphim.lua
```diff
@@ -125,13 +125,15 @@ function CreateSeraphimFactoryBuildingEffects(builder, unitBeingBuilt, effectBon
         WaitFor(slider)
     end
 
+    -- localize for optimal access
+    local UnitGetFractionComplete = UnitGetFractionComplete
+    local completed = UnitGetFractionComplete(unitBeingBuilt)
     -- # Gradually move the unit to the plateau
-
-    while not unitBeingBuilt.Dead do
-        completed = UnitGetFractionComplete(unitBeingBuilt)
```
brokerjoe, line 131: @Garanas This variable was reading from the global context, so I changed the code to read it from a local variable, variable is defined outside for localized access.

lua/EffectUtilitiesSeraphim.lua
```diff
@@ -150,15 +150,12 @@ end
 function CreateSeraphimBuildThread(unitBeingBuilt, builder, effectsBag, scaleFactor)
 
     -- # initialize various info used throughout the function
-
-    local effect = false
```
brokerjoe, line 152: this variable was defined twice

lua/EffectUtilitiesSeraphim.lua
Garanas: I had to look at it twice - well spotted.

brokerjoe:
I think you are right, lets move that in - can make always a new PR.

Garanas:
Took the liberty to add a fix to this pull request. When building large units (Seraphim bomber, for example) the LOD of the effect doesn't match the size of the unit. This causes inconsistencies. The LOD is now always set according to the LOD of the mesh of the unit.

At the time I didn't knew this was possible 😃 .

lua/EffectUtilitiesSeraphim.lua
```diff
@@ -148,30 +150,39 @@ end
...
     local effect = false
 
     -- matches with number of effects being made, pre-allocates the table
     local emitters = { false, false, false, false, false }
     local emittersHead = 1
 
+    -- determine a sane LOD cutoff for the size of the unit
+    local lods = unitBeingBuilt.Blueprint.Display.Mesh.LODs
+    local count = TableGetn(lods)
+    local LODCutoff = 0.9 * lods[count].LODCutoff or (90 * MathMax(unitBeingBuilt.BuildExtentsX, unitBeingBuilt.BuildExtentsZ))
```
Garanas, line 169: What LOD value should we use here? At the moment it uses the value of the lowest LOD (typically there are only two, so the second). But maybe it is more sane to use the LOD value of the high quality mesh?

PR #3741, opened 2022-03-27
New lobby option to disable Manual unit share

Crotalus:
Make it possible to disable manual unit sharing, units are still given at death if full share is on

Options:
yes - manual share allowed (default)
no_engineers - no sharing of engineers/factories (filter out units with CONSTRUCTION or ENGINEER)
none - no manual share of units


lua/ui/lobby/lobbyOptions.lua
```diff
@@ -465,6 +465,29 @@ globalOpts = {
...
+        help = "<LOC lobui_0791>Are players allowed to manually give units?",
+        key = 'ManualUnitShare',
+        values = {
+            {
+                text = "<LOC _Yes>Yes",
+                help = "Manual unit sharing are allowed",
+                key = 'all',
+            },
+            {
+                text = "Yes except builders",
```
Garanas, line 480: Can you add this one too to the translation?

loc/US/strings_db.lua
```diff
@@ -7666,6 +7666,13 @@ lobui_0781="Teams will be balanced with up to 5%% tolerance of best setup to mak
 lobui_0782="Optimal balance (Mirrored)"
 lobui_0783="Teams will be optimally balanced, mirrored start locations"
 
+lobui_0790="Manual Unit Sharing"
+lobui_0791="Are players allowed to manually give units?"
+lobui_0792="Manual unit sharing are allowed"
```
Garanas, line 7671: should this be `is` instead of `are`?

loc/US/strings_db.lua
```diff
@@ -7666,6 +7666,13 @@ lobui_0781="Teams will be balanced with up to 5%% tolerance of best setup to mak
...
+lobui_0793="Yes except builders"
```
Garanas, line 7672: And I think this should be `Yes, except builders`

loc/US/strings_db.lua
```diff
@@ -7666,6 +7666,13 @@ lobui_0781="Teams will be balanced with up to 5%% tolerance of best setup to mak
...
+lobui_0794="No manual sharing of builder/factory"
```
Garanas, line 7673: and then specify them here as engineers and factories, anything that has a build rate

lua/SimUtils.lua
```diff
@@ -556,9 +556,16 @@ function TransferUnfinishedUnitsAfterDeath(units, armies)
...
+    if manualShare == 'none' then return end
+
     if units then
         local owner = units[1].Army
         if OkayToMessWithArmy(owner) and IsAlly(owner,data.To) then
+            if manualShare == 'no_builders' then
+                units = EntityCategoryFilterDown(categories.ALLUNITS - categories.CONSTRUCTION - categories.ENGINEER, units)
+            end
+            
             TransferUnitsOwnership(units, data.To)
```
Garanas, line 569: There needs to be some UI element telling you if you're trying to give engineers / factories but you can't

4z0t:
Also, what would happen to Exps like fatboy, megabot, atlantis, czar, tempest and novax?

lua/SimUtils.lua
Garanas: You can compare the length of units before / after if the filtering to confirm it has engineers in it

Garanas:
I've asked the balance team. For experimentals, we can add on a per-unit bases restrictions to prevent them from being able to make engineers.

Marc-Spector:
T_R_U_Putin is sad somewhere 😅

Emperor-Penguin1:
If you do add restrictions for building units with T4 (which makes sense), you should probably also make it so those restrictions get removed when the units are transferred via full share.

loc/US/strings_db.lua
Crotalus: Fixed

loc/US/strings_db.lua
Crotalus: Fixed

loc/US/strings_db.lua
Crotalus: Changed to engineers/factories (until we know exactly what will be banned)

lua/SimUtils.lua
Crotalus: Added a print message now when not all units could be given

Garanas:
If I am not mistaken then the consensus of the balance team is that this is not a balance related change. @Tagada14 can you confirm this?

Garanas:
I'm merging this - I think it can use some fine tuning before the release based on feedback. In particular, you can't share some experimental units at the moment due to their building capacities.

PR #3728, opened 2022-03-26
Lobby team ratings improvements

Emperor-Penguin1:
Rounds displayed team ratings
Refresh observer list more often
Fix weird bug involving the observer list's numTeams sometimes being calculated incorrectly

PR #3742, opened 2022-03-27
Lobby Team Ratings Update

Emperor-Penguin1:
Makes in-lobby team ratings being displayed in the observer window conditional upon spawns being fixed
Makes in-lobby displayed team ratings update for everyone (rather than just the host) more often
Moves refreshObserverList earlier so it can be called in SetSlotInfo and ClearSlotInfo
Calls refreshObserverList in SetSlotInfo and ClearSlotInfo and removes some now-superfluous calls to it


Emperor-Penguin1:
@Garanas I think this fixes the problems you mentioned in #3721 and #3728.

Garanas:
Good! One last addition for a better user experience:

image

When you're **not** the host, I think it is nice to add a header to the team statistics to tell people it are the team statistics (red arrow), to add an empty line between the observers and the team statistics (yellow I) and then to add a header to the observers list (blue arrow). And by all means the observers needs to be below the last mentioned header (purple arrow).

For the host the box is too small for this layout. But for the regular user the box is quite large.

Emperor-Penguin1:
Alright, I added a space between the team ratings and observers for non-host lobby users.  I think that is a questionable change tbh, but it's not a big deal.  I think adding 2 additional title lines to the window would be an inefficient use of space and would make things more clogged though.  Something like color-coding the team ratings might make more sense, but I think that would be a QoL improvement, not a necessary feature.

Garanas:
I think informing the user what it is looking at is a perfect use of space. Especially since we have the space when you're not a host. And that is the type of view that the majority of us get to see.

Emperor-Penguin1:
@Garanas Okay, I added them.

lua/ui/lobby/lobby.lua
```diff
@@ -857,6 +857,104 @@ function GetPlayerDisplayName(playerInfo)
...
+        end
+
+        for i, team in teamRatings do
+            numTeams = numTeams + 1
+        end
+
+        -- if there are 1 or 2 teams, list them before observers
+        if numTeams == 1 or numTeams == 2 then
+            if not lobbyComm:IsHost() then
+                GUI.observerList:AddItem('Team Ratings:')
```
Garanas, line 894: @4z0t can you add a russian translation here?

Garanas:
This is ready to merge once the translations are in. Great work.

Garanas:
@Emperor-Penguin1 subtle hint from @4z0t to use string.format (or LOCF) when you format a string (over direct concatenation such as `..`)

PR #3743, opened 2022-03-27
Issue 3727 replace table trashbags

jkstpierre:
Refactored several trashbags in Unit.lua to use the TrashBag class. Testing seems to show no errors with this new implementation.

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -37,6 +37,9 @@ local UpdateAssistersConsumptionCats = categories.REPAIR - categories.INSIGNIFIC
 -- upvalue for performance
 local rawget = rawget 
 local IsAlly = IsAlly
+local TrashAdd = TrashBag.Add
+local TrashDestroy = TrashBag.Destroy
+local TrashEmpty = TrashBag.Empty
 
 SyncMeta = {
     __index = function(t, key)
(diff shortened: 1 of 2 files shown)


lua/sim/Unit.lua
```diff
@@ -37,6 +37,9 @@ local UpdateAssistersConsumptionCats = categories.REPAIR - categories.INSIGNIFIC
 -- upvalue for performance
 local rawget = rawget 
 local IsAlly = IsAlly
+local TrashAdd = TrashBag.Add
```
Garanas, line 40: You can also add the TrashBag itself as an upvalue (the class instantiation is part of the global scope now)

lua/sim/Unit.lua
```diff
@@ -4031,18 +4026,17 @@ Unit = Class(moho.unit_methods) {
         self:DestroyIdleEffects()
         self:DestroyMovementEffects()
 
-        table.insert(self.TransportBeamEffectsBag, AttachBeamEntityToEntity(self, -1, transport, bone, self.Army, EffectTemplate.TTransportBeam01))
-        table.insert(self.TransportBeamEffectsBag, AttachBeamEntityToEntity(transport, bone, self, -1, self.Army, EffectTemplate.TTransportBeam02))
-        table.insert(self.TransportBeamEffectsBag, CreateEmitterAtBone(transport, bone, self.Army, EffectTemplate.TTransportGlow01))
+        TrashAdd(self.TransportBeamEffectsBag, AttachBeamEntityToEntity(self, -1, transport, bone, self.Army, EffectTemplate.TTransportBeam01))
```
Garanas, line 4029: And while you're at it - could you add in these emitter / beam engine calls as upvalues too?

Garanas:
Interesting find - the Trashbag can gradually grow over time. I 'optimized' the trashbag in https://github.com/FAForever/fa/pull/3332 but it turns out that the original implementation was actually quite solid on its own. 

What happens is that during the existence of the trashbag entries may be collected prematurely by the garbage collector. The 'optimized' solution didn't take this into account, allowing a single tank to create a trashbag that can easily be in the hundreds of elements. 

By all means my mistake and glad that we caught it.

Garanas:
image
image

900 ASF on the base branch, 9 to 10 ms

image
image
900 asf on this branch, 8 - 8.5 ms

It is difficult to replicate the test exactly because the number of times the trashbags are refreshed depends on the exact queued orders. I've tried several, they all have the same range, give or take. Values are taken after one minute to stabilize the patrol.

Tiny change with a disproportionally positive outcome - great work.

PR #3746, opened 2022-03-28
Aeon Effect Utilities

brokerjoe:
## Description

* Fix the most obvious wrong access patterns in the looping business

I keep it as a draft for now - since I haven't tested those changes yet.

### DIFF
--- a/lua/EffectUtilitiesAeon.lua
+++ b/lua/EffectUtilitiesAeon.lua
@@ -116,7 +116,7 @@ local function SharedBuildThread(pool, unitBeingBuilt, unitBeingBuiltTrash, unit
                     progress = 0 
                 end
 
-                scale = 1 - progress * progress
+                local scale = 1 - progress * progress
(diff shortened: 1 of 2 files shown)


Garanas:
Looking good. 

Could you look into fixing the LODs of the effects, similar to what I did in https://github.com/FAForever/fa/pull/3740 ? I'm not sure if we can tweak the LOD parameter of the puddle, since that is a projectile and not an emitter. Specifically the experimental build effects have LOD issues, such as the Tempest and CZAR due to their sheer size.

brokerjoe:
Actual, I just looked into that and added a scaler to the particle effect when building the experimental. If you build it with 20 commanders it looks pretty neat, alternative the particle effect can be reversed and size it down with progress. and remove when  inside the mesh. Same can be done on the larger objects. both commits can be reversed to - i think I was just to interested how that part works.

Garanas:
Closing as it has been months.

PR #3750, opened 2022-03-30
In-lobby auto balancing

Emperor-Penguin1:
New AutoBalancing function written from scratch - it works in-lobby
Added Autobalancing button and its 20 images (there are 20 because faction color-coding/button highlighting/etc)...
Moved lobby buttons to accommodate the new button

What it does: it balances all occupied slots into two teams with equal numbers of players. If half of the occupied slots are set to team 1 and half to team 2, then it balances the players while keeping the team-slot matches. If the teams are not set that way, they are changed automatically to be alternating team 1 and team 2. If there are an odd number of occupied slots, the last one is set to team - (no team) and the others are balanced without it.

How it balances: this function checks every possible balance combination for making the two teams (while keeping their player counts equal to half the number of occupied slots, rounded down, and not using the last player if there is an odd number of players). To do this, the function sums up all the relevant players' ratings (keeping mean and deviation separate - it balances teams to have similar total ratings, and also similar total uncertainties (grayness)), and then divides by two. That yields the goal values for each team. Any deviation from those values is calculated to help determine a team's imbalance value. Then, the various team combinations are tested, and the one with the lowest imbalance value is used.

Note that the process used by this function improves efficiency by minimizing the number of iterations, while still checking every possible balance combination.

Also note that the balance values used in this are tentative and may be adjusted in the future.

image


Sheikah45:
Summing the mean and the deviation independently is not the same as taking the sum of the ratings. This will likely lead to very oddly balanced teams.

Also the game already has an auto balancing function it uses for optimal balance why not just reuse that?

Garanas:
I think the button needs to be disabled when the teams of the game are not set to top vs down, odd vs even, left vs right as it doesn't support more than two teams.

image

Besides that, the current shape is ready for faf develop. I do highly encourage you to refactor the function such that it works with input / output pairs. Where the input is the players in the lobby and the output is how they should be moved around for an optimal game. That way we can test the functionality and even write tests for it.

Emperor-Penguin1:
@Garanas Since the AutoBalancer may set teams to manual, I don't think it should be disabled when teams are set to manual, as they can create extra hassle for the user (ie: when wanting to hit the button again after a player joins or quits).  So, I set it to be disabled if a player is on team 3 or greater instead, as I think that covers the situations you were most concerned about.

ageekhere:
Would making the AutoBalancing function run as a Coroutine help?
https://www.lua.org/pil/9.1.html

PR #3759, opened 2022-04-04
Improve matchmaking communication

Garanas:
Introduces a new UI class specifically designed to communicate the status of the autolobby. The auto lobby is used by the matchmaker.

At the moment it supports displaying:
 - The current number of players that are connected over the total number of players

Todo:
 - [x] Integrate it with `autolobby.lua`


lua/ui/lobby/autolobby-classes.lua
```diff
@@ -0,0 +1,68 @@
...
+        self.Background:SetAlpha(0.2)
+        LayoutHelpers.FillParent(self.Background, self, 0.01)
+
+        UIUtil.SurroundWithBorder(self, '/scx_menu/lan-game-lobby/frame/')
+
+        -- generic header
+        self.HeaderText = UIUtil.CreateText(
+            self, 
+            "Connection status", 
+            16 * LayoutHelpers.GetPixelScaleFactor(),
```
speed2CZ, line 28: Text going through `UIUtil.CreateText` is already scaled, this will scale it twice.

Garanas:
Question towards the reviewers: I'd like to do this correct the first time. In particular:
 - The UI class (ConnectionStatus) should not crash at any point. It can be tested separately by attaching a hotkey to the profiler window using this PR. By all means, that won't be part of the production code. I am hoping that 4z0t can have a good look at this.
 - The status of the window should always be correct. This is a bit tricky, as I'm not 100% confident of the flow of code. I am hoping that BlackYps can take a good look at this.
 
And please find any typo's in the `autolobby.lua` file - as those would be pretty silly to ship into production 😄 .

Garanas:
Oh, and @4z0t can you add in a russian translation? As promised: the it is already using string formatting.

BlackYps:
Why is there a checkbox in front of the text?

Garanas:
The idea was that there could be multiple criteria, such as a being connected to the host and / or being the host. If all the checkboxes are set, then the launch is about to happen. I can remove it again.

BlackYps:
A checkbox implies that the user can/should interact with it. Maybe you can find a tickmark image in files somewhere and use that instead. Or just remove it altogether. It's pretty straightforward that everybody needs to connect and once that happened the match will start and the dialogue will not be visible anymore anyway.

Garanas:
image

There we go 😃

4z0t:
Consider making "connection status" text larger and another colour

Garanas:
@4z0t using the title font as defined in uiutil

image

BlackYps:
This here seems odd
grafik

lua/ui/lobby/autolobby.lua
```diff
@@ -416,27 +430,48 @@ function JoinGame(address, asObserver, playerName, uid)
...
 
-function DisconnectFromPeer(uid)
+function DisconnectFromPeer(uid, doNotUpdateView)
     LOG("DisconnectFromPeer (uid=" .. uid ..")")
     if wasConnected(uid) then
         table.remove(connectedTo, uid)
     end
     GpgNetSend('Disconnected', string.format("%d", uid))
+
+    -- sometimes we disconnect immediately, but secretly connect through a proxy
```
BlackYps, line 470: If the only other realistic option for a disconnect is that someone closes his game, then this should be ok. If someone does that, the server will abort the match and all clients will close the game.
If there are other reasons for a disconnect this could potentially lead to the counter ticking up to full but the game not starting.

lua/ui/game/Profiler.lua
```diff
@@ -149,12 +149,37 @@ list = false
...
+            status:RemoveConnectedPlayer()
+            WaitSeconds(1.0)
+            status:AddConnectedPlayer()
+            WaitSeconds(1.0)
+        end
+    end)
+
+    if true then 
+        return 
+    end
```
BlackYps, line 172: isn't this equivalent to a simple return?

Garanas: You can't - return expects an `end` token right after it.

lua/multiplayer/onlineprovider.lua
```diff
@@ -0,0 +1,34 @@
...
+--* Author: Sam Demulling
+--* Summary: 3rd party integration options.  This will receive and send commands
+--* by hooking into stdin/out in the supremecommander app.
+--*
+--* Copyright � 2006 Gas Powered Games, Inc.  All rights reserved.
+--*****************************************************************************
+
+
+# Called from C++ code when we are launching from gpg.net.
+function CreateLobby(autolaunch, protocol, port, playerName, uid, natTraversalProvider, hasSupcom)
```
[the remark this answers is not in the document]
Garanas: Nothing - I am not even confident that it is being used. But for testing purposes it got added. I can remove it again.

lua/ui/lobby/autolobby-classes.lua
```diff
@@ -0,0 +1,102 @@
...
+    ConnectedPlayersCount = 1,
+
+    -- View elements
+
+    --- Updates the view of the model / view / controller of this UI element
+    UpdateView = function(self)
+        local headerText = LOC('<LOC AutoLobbyHeaderText>Connection status')
+        self.HeaderText:SetText(headerText)
+
+        local connectionsText = LOCF('<LOC AutoLobbyConnectionsText>%s / %s are connected', tostring(self.ConnectedPlayersCount), tostring(self.TotalPlayersCount))
```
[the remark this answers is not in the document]
Garanas: We can do that!

lua/ui/lobby/autolobby.lua
Garanas: I understand - but where would we detect that, if not here?

Garanas:
> This here seems odd grafik

Yes - I'm not sure why but it is because the texture used on the sides is 'compressed' vertically. It is barely visible on a dark background, such as the matchmakers.

lua/ui/lobby/autolobby.lua
```diff
@@ -416,27 +430,48 @@ function JoinGame(address, asObserver, playerName, uid)
...
+    local args = GetCommandLineArg("/players", 1)
+    if args then
+        requiredPlayers = tonumber(args[1])
+        LOG("requiredPlayers was set to: "..requiredPlayers)
+    end
+
     SetGameOptionsFromCommandLine()
 
+    -- update the connection status GUI
+    connectionStatusGUI:SetTotalPlayersCount(requiredPlayers or 1)
```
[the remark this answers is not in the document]
Garanas: I agree.

lua/ui/game/Profiler.lua
BlackYps: wow

lua/multiplayer/onlineprovider.lua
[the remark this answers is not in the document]
BlackYps: If nothing changed, why did you add it?Doesn't it use the files from the basegame normally?

BlackYps:
can you point me to that texture?

lua/ui/lobby/autolobby.lua
BlackYps: To answer that we would have to know if there is a plausible scenario that a once connected player disconnects again. I think for the moment it is fine to leave it as is, because I can't think of any. On the other hand I don't really know how the connection process works...

4z0t:
@BlackYps it is ninepatch

lua/multiplayer/onlineprovider.lua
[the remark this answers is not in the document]
Garanas: I added it to be able to add LOG statements. To understand how the autolobby is triggered over the regular lobby, see also the discussion on Zulip.

loc/US/strings_db.lua
```diff
@@ -4676,6 +4676,10 @@ lobui_0701="The last player is kicked because he does not have this mod: "
 lobui_0702="Team Ratings:"
 lobui_0703="Team %d: %d (%d+/-%d)"
 
+AutoLobbyHeaderText="Connection status"
+AutoLobbyConnectionsTextSingular="%s / %s is connected"
```
Askaholic, line 4680: I think in this context it is always correct to use 'are'. I would say '1/8 are connected' sounds more correct than '1/8 is connected'.

lua/ui/lobby/autolobby.lua
```diff
@@ -416,27 +430,48 @@ function JoinGame(address, asObserver, playerName, uid)
     LOG("Joingame (name=" .. playerName .. ", uid=" .. uid .. ", address=" .. address ..")")
     CreateUI()
 
+    -- TODO: I'm not sure if this argument is passed along when you are joining a lobby
```
Askaholic, line 433: You could error on the side of caution and just show a `?` or not show a total number at all if this isn’t passed in.

Garanas: According to BlackYps on Zulip it is passed along, if I understood him correct.

loc/US/strings_db.lua
Garanas: I'm fine with either. I'm not a native speaker.

lua/ui/lobby/autolobby.lua
Askaholic: But that doesn’t mean that no one will ever come along and write a client that doesn’t do that, or that there will never be a bug that causes it not to be passed along. IMO when it’s easy enough to handle such as in this case, you shouldn’t be relying too much on external software to behave in a certain way. It doesn’t hurt to play it safe and add a little error checking :)

loc/US/strings_db.lua
4z0t: https://thegrammarexchange.infopop.cc/topic/1-in-10-is-or-are#:~:text=%22One%20in%20ten%20people%20is,%22is%22%20would%20be%20correct. :)

lua/ui/lobby/autolobby.lua
Garanas: I understand - but it already defaults to 2. Therefore the program wouldn't crash. It would be bad for the user experience, but so would a `?` be. I am leaving it as is.

BlackYps: Just showing "X are connected" when the total number is unknown seems like the best solution

Askaholic: I agree. Showing 2 is just misleading.

loc/US/strings_db.lua
Askaholic: That makes sense to me too, which is why I would choose `are` here since it makes the code simpler.

BlackYps: fine with me

lua/ui/lobby/autolobby.lua
Garanas: I can work with the change of BlackYps - I'll add it in tomorrow.

Garanas:
image

Garanas:
@4z0t can you fix the translations?

lua/ui/lobby/autolobby-classes.lua
```diff
@@ -0,0 +1,124 @@
...
+    -- Initialisation
+
+    __init = function(self, parent)
+        Group.__init(self, parent)
+
+        -- set our dimensions
+        LayoutHelpers.SetDimensions(self, 200, 100)
+
+        -- put a border around ourselves
+        UIUtil.SurroundWithBorder(self, '/scx_menu/lan-game-lobby/frame/')
```
BlackYps, line 23: I can't find /frame in this path. How does this work?

Garanas: It ends up calling `UIFile` to retrieve the files, it appends the path depending on the skin used.

Garanas: These are part of the FA repository, apparently:

image

BlackYps: I was looking in the common folder instead of the faction specific one. That was the issue

BlackYps:
I checked the files and both the vertical and horizonal lines have a slight shadow to them. But for whatever reason we only really see the shadow on the vertical line. Maybe it's caused by the left and right images being compressed vertically, so the transparency stacks up. It might be worthwhile to just remove that shadow. I can do that if you want, but I don't think I can commit to this branch, so I would have to send them to you, so you can commit them.

Garanas:
I'd appreciate it - you can send them over Discord to me.

PR #3764, opened 2022-04-05
Changes to new Autobalance

Emperor-Penguin1:
Fix an edgecase (trying to autobalance with 0 players) and change rating-based sorting order to factor in deviation as well (this doesn't impact who gets whom as a teammate, but does impact intrateam slot order.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -3864,8 +3864,8 @@ function CreateUI(maxPlayers)
                 goalValue[2] = goalValue[2] + player.DEV
             end
 
-            -- if there is only one player, there is no need to balance
-            if playerCount == 1 then
+            -- if there are fewer than 2 players, there is no need to balance
+            if playerCount < 2 then
                 return
             end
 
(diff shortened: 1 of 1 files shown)


lua/ui/lobby/lobby.lua
```diff
@@ -3864,8 +3864,8 @@ function CreateUI(maxPlayers)
                 goalValue[2] = goalValue[2] + player.DEV
             end
 
-            -- if there is only one player, there is no need to balance
-            if playerCount == 1 then
+            -- if there are fewer than 2 players, there is no need to balance
+            if playerCount < 2 then
```
Garanas, line 3868: shouldn't this be fewer than three? As a 1 v 1 is always 'balanced'

Emperor-Penguin1: I considered that, but I figured that people might want to use the button as an easy way to set teams in that context (ie: going from 2 players on team 1 to 1 player on team 1 and 1 on team 2).
If you want me to change it anyway, lmk and I will; I just figured it was slightly better this way.

Garanas: Fair enough to me.

PR #3765, opened 2022-04-05
Better available color list update functionality

Emperor-Penguin1:
This makes the in-lobby color list able to update without "closing."

Fixes the problem of the available color list "closing" (being destroyed and recreated, with the new one set to hidden) when the available color list is updated.  Now, when a BitmapCombo is changed via ChangeBitmapArray, the new one is not hidden if the old one was not hidden.
Also simplifies where Check_Availaible_Color is called and makes it get called when a player switches to observer (the available color list was previously not triggered to update by this action).

### DIFF
--- a/lua/ui/controls/combo.lua
+++ b/lua/ui/controls/combo.lua
@@ -656,8 +656,10 @@ BitmapCombo = Class(Group) {
         -- prev will be last control here
         self.ddm.Bottom:Set(prev.Bottom)
 
-        self._dropdown:Hide()
-        self._ddhidden = true
+        if self._dropdown:IsHidden() then
+            self._dropdown:Hide()
+            self._ddhidden = true
+        end
     end,
 
     SetBitmap = function(self, bmp, name)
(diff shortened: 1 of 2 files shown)


Garanas:
Sounds like one of the issues described earlier - great job.

4z0t:
Why you check if it is hidden and hide again?

Emperor-Penguin1:
@4z0t The function destroys the previous version and recreates the new version.  If the previous version was hidden, the new version should be hidden.  If the previous version was not hidden, the new version should not be hidden.  Maybe there is some unnecessary hiding, but I wasn't trying to refactor everything here.  This PR didn't add new hiding; it just made some old hiding happen less often.

Garanas:
I think 4z0t is right - you're checking the same entity here. if it was hidden, then hiding it again wouldn't help.

4z0t:
@Garanas It is correct but a bit confusing, doing this he hides newly created children

Garanas:
Ah, it hides the children 😄 .

PR #3768, opened 2022-04-06
In-Lobby AutoBalance Improvements

Emperor-Penguin1:
Adds AutoBalance button functionality of automatically setting spawns to fixed if that's not already the case
Adds AutoBalance button functionality for odd vs even, top vs bottom, and left vs right AutoTeams configurations (rather than forcing manual and using the previously occupied slots) - slots are used in numerical order, only skipping closed slots and slots necessary to skip to keep team size equal (if any)

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -3992,7 +3992,7 @@ function CreateUI(maxPlayers)
             end
 
             -- if the teams were not set properly, set them properly
-            if table.getn(sortedTeam1Slots) < teamSize or table.getn(sortedTeam2Slots) < teamSize or (manualTeams and (numPlayersTeam1 != teamSize or numPlayersTeam2 != teamSize)) then
+            if (not manualTeams and (table.getn(sortedTeam1Slots) < teamSize or table.getn(sortedTeam2Slots) < teamSize)) or (manualTeams and (numPlayersTeam1 != teamSize or numPlayersTeam2 != teamSize)) then
                 -- set AutoTeams to none (so, they can be set by slot by this function)
                 gameInfo.GameOptions.AutoTeams = 'none'
                 local counter = 0
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -3848,12 +3848,15 @@ function CreateUI(maxPlayers)
         GUI.PenguinAutoBalance.OnClick = function()
 
             -- set spawns to fixed
-            gameInfo.GameOptions.TeamSpawn = 'fixed'
-            -- tell everyone else to set spawns to fixed
-            lobbyComm:BroadcastData {
-                Type = 'GameOptions',
-                Options = {['TeamSpawn'] = 'fixed'}
-            }
+            if gameInfo.GameOptions.TeamSpawn ~= 'fixed' then
+                gameInfo.GameOptions.TeamSpawn = 'fixed'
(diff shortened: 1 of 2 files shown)


Garanas:
I was wondering what this changed - I think it makes sense. But I also think that a dialogue to confirm the change would be good for the user experience. Otherwise people may not be aware it changed if they, for example, were just toying with the feature.

Emperor-Penguin1:
I think a confirmation popup that the user has to click every team he/she wants to autobalance teams would be more annoying than helpful.  So, I just added an in-chat text message that says "Enabled fixed spawn locations" instead.

Garanas:
It would only pop up if the settings would actually change, by all means. But a chat message works too.

Garanas:
Open an additional pull request for other improvements - otherwise I can keep re-reviewing it!

lua/ui/lobby/lobby.lua
```diff
@@ -3847,25 +3847,37 @@ function CreateUI(maxPlayers)
...
+                }
+                AddChatText(LOC("<LOC lobui_0446>Enabled fixed spawn locations"))
+            end
+
             -- a table of the target mean, target deviation, and the lowest logged imbalance value
             local goalValue = {0, 0, 99999}
 
             local playerCount = 0
-            local lastSlot = {0, 0}
+            local lastSlot = {0, 0, false}
```
Garanas, line 3865: Add a comment to describe the content of this table, like you did on line 3861.

lua/ui/lobby/lobby.lua
```diff
@@ -3920,8 +3939,60 @@ function CreateUI(maxPlayers)
...
+                            team1OrderNum = team1OrderNum + 1
+                            sortedTeam1Slots[team1OrderNum] = i
+                        else
+                            team2OrderNum = team2OrderNum + 1
+                            sortedTeam2Slots[team2OrderNum] = i
+                        end
+                    end
+                end
+            elseif gameInfo.GameOptions.AutoTeams == 'tvsb' then -- top vs bottom
+                local midLine = GUI.mapView.Top() + (GUI.mapView.Height() / 2)
```
Garanas, line 3965: I understand the idea here, is this how other code determines the slots being part of the top / bottom team too?

lua/ui/lobby/lobby.lua
```diff
@@ -3920,8 +3939,60 @@ function CreateUI(maxPlayers)
...
+                        end
+                    end
+                end
+            else
+                manualTeams = true
+            end
+
             -- if the teams were not set properly, set them properly
-            if numPlayersTeam1 != teamSize or numPlayersTeam2 != teamSize then
+            if (not manualTeams and (table.getn(sortedTeam1Slots) < teamSize or table.getn(sortedTeam2Slots) < teamSize)) or (manualTeams and (numPlayersTeam1 != teamSize or numPlayersTeam2 != teamSize)) then
```
Garanas, line 3995: Make this a multi line if statement (as an example: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/shield.lua#L2430 and add a comment for each part of the if statement, to improve readability.

lua/ui/lobby/lobby.lua
Emperor-Penguin1: yes

Garanas:
I'm having a hard time testing this - from a code perspective it looks good.

PR #3772, opened 2022-04-07
Add parties to lobby

Emperor-Penguin1:
Adds parties to lobby
Adds support for party functionality when in-lobby autobalancing
Makes parties disabled and set to - (no party) when spawns are not set to fixed (ie, for opti) - this can be changed later if opti/etc is changed to support parties
Some formatting changes were made as well, notably to the column positions and widths (so the party column could be added properly)
Note that if people are set to be in the same party, the only current effects are to display the party numbers for lobby members to see (which is useful) and to try to keep them together when in-lobby autobalancing

image


MrRowey:
why make the loby screen more complicated than it needs to be ? if you want a part thier tmm for that ?

MrRowey:
will jsut cause more issuse than is needed

Emperor-Penguin1:
A lot of people have wanted in-lobby party support for a long time.  It's useful for custom games.  There are many reasons many people prefer to play custom games a lot (or all) of the time.  It's easily worth the UI space.

lua/ui/lobby/lobby.lua
```diff
@@ -424,7 +424,7 @@ end
 
 --- Get the value of the LastFaction, sanitised in case it's an unsafe value.
 --
--- This means when some retarded mod (*cough*Nomads*cough*) writes a large number to LastFaction, we
+-- This means when some potentially problematic mod (*cough*Nomads*cough*) writes a large number to LastFaction, we
```
Garanas, line 427: heh

lua/ui/lobby/lobby.lua
```diff
@@ -4030,6 +4128,25 @@ function CreateUI(maxPlayers)
 
             testCombinations(1, 1)
 
+            -- check if bestTeam's player count is wrong
+            if table.getn(bestTeam) != teamSize then
```
Garanas, line 4132: When would this occur?

Garanas:
I'll be honest, it does look crowded.

Garanas:
I think having parties use letters instead of numbers would make it look less ambiguous with teams.

lua/ui/lobby/lobby.lua
Emperor-Penguin1: It would occur if people make a combination of parties that is impossible to satisfy.  It might also occur if something is broken.  I think it's a sensible check to make.
Example 1: a 4v4 with 1 party of 5 people does not work.
Example 2: a 3v3 with 3 parties of 2 people does not work.

Garanas: But can't you catch those before / during the balancing, instead of afterwards?

Emperor-Penguin1: Trying to catch it during the balancing sounds too inefficient as adding an additional something for this to every iteration that the balancing does does not sound like a good idea.  Trying to catch it beforehand is theoretically possible (and would require additional thinking/coding), but would presumably normally have the effect of slowing down the autobalancing process a little for everyone while only saving time when people have set up impossible party combinations (which should be a relatively small percent of the time).  I am not aware of any present case beyond impossible party combinations that would cause the bestTeam to not match the teamSize here though, but just having the single if-check is not costly and is a good safety net in case that gets messed up in some unexpected way in the future.

Emperor-Penguin1:
@Garanas Okay, I added them.

image

PS: if someone wants to make the letters look better, I'm not opposed to them being edited/replaced with more aesthetic designs.

Sheikah45:
Why not use the same font the game does? As it is now it is a little jarring

Emperor-Penguin1:
@Sheikah45 These are images... do you know a good way to easily get the game font for individual letters in the proper file format/size?

4z0t:
try other inhame fonts 
image

Sheikah45:
As 4zot showed just find the font name and or file, install it and type out as you need in word or something if that is what is easiest for you then use ps or gimp to convert it to an image and you got it.

Emperor-Penguin1:
I didn't even make these with a font... I made them in an image editor... someone else is welcome to go make ones that look perfect.  I personally don't want to spend hours of my time on this trying to get just the right font in just the right size with perfect aesthetics in the right file format.  I'm happy to switch them back to numbers if the letters are too jarring.

Eternal-ll:
label = UIUtil.CreateText(PARENT, "TEXT", FONTSIZE, "FONT-FAMILY")
You can use others in-game fonts

Garanas:
Why are we using images? There are combo boxes that support text.

Emperor-Penguin1:
@Garanas 
It was using images because it was based off of the team combo boxes, which use images.  Here is what it looks like using a text combo box with a game font.  As you can see, while it does look like the game's text this way, it does not stand out very well and it does not match the team numbers, which are images...  Personally, I prefer the boldness of the images, but if you want me to make a commit with this text combo box, let me know.
image

4z0t:
For bold use Arial Black

Emperor-Penguin1:
Arial Black actually looked too thick.  Arial Bold looks reasonable though.  So, I'll probably go with that.
image

Garanas:
I am not sure about this and I'm leaning towards closing this pull request.

There are a couple of disadvantages:
 - The lobby looks a lot more crowded
 - It seems to me it is easy to abuse
 - It doesn't work with the Opti game mode, people will definitely expect that (even if it is disabled) because of the 'how hard can it be if it is already here!' notion

And then there's one advantage:
 - using your auto balance algorithm, parties can now _potentially_ be taken into account, but it may fail
 
According to Phong on the forums, the latter doesn't take that much time in practice. At best it would be switching one or two roughly-equally rated players in the lobby, which is quite easy with a feature introduced by @4z0t that makes it a two click procedure. I can't find the pull request at the moment.
 
The real solution would be that people create parties in the client and join a custom game as a party. That, to me, would be a feature that we can work with. It would prevent the existence of a combo box all together and people could have some color identification to show the parties in question. Parties can not be 'abused' by people adding themselves to a party without the consent of said party. The host can then choose to ignore parties or not by manually switching them into the same team. Again by the same procedure where you switch two roughly-equally rated players in the lobby.

Emperor-Penguin1:
@Garanas 
* Having parties listed is useful for in-lobby autobalancing and manual balancing.  It helps to show people who is in a party with whom.  Hosts often have trouble keeping track of that properly when balancing manually.  This PR would help a lot with that problem.
* Given that the game doesn't actually start when the host presses the autobalance button, I think you are being overly concerned about abuse here.  We arguably have several other features that are more ripe for abuse that are already in-lobby.  That being said, if you really want, I'm open to adding some sort of toggle for party selection.
* There are many times where you can't just switch two roughly equally rated players across teams in the lobby.
* There are also many times where there are multiple groups of of 2-4 people in the lobby who want to play together, which can get particularly tricky to balance sometimes.  I've often seen hosts struggle with balancing multiple parties for several minutes straight, only to end up with a mediocre result or giving up.
* A lot of people have wanted in-lobby party support for a long time. It's useful for friends who want to play together on a team and for the hosts who want to accommodate them (people often leave lobbies if they aren't allowed to play on a team with their friends).   Making it easier to balance with friends who want to play on a team together is good for the user experience.. It's easily worth the UI space.

It really looks fine in the lobby now.

Here's how it looks for a regular player in the lobby with 2 people in a party:
image

Here's how it looks for a regular player in the lobby with no one in a party:
image


If I had to choose between in-lobby party selection and in-client party selection for custom games that acts similarly to how in-client party selection for TMM works, I would choose in-lobby party selection hands-down.  I'm not saying that they'd have to be mutually exclusive, but I simply find the in-client party selection system more rigid and inconvenient by comparison.  Additionally, that is not coded ,nor do I expect it to be coded any time soon, if ever.  By comparison, this is already coded and functional.  That is a very significant difference.

Garanas:
> Having parties listed is useful for in-lobby autobalancing and manual balancing. It helps to show people who is in a party with whom. Hosts often have trouble keeping track of that properly when balancing manually. This PR would help a lot with that problem.

I think stating that 'hosts often have trouble keeping track' is a bit of an overstatement. I can't recall the last time this was an issue in a lobby, besides the host just not caring about it all together. 

> Given that the game doesn't actually start when the host presses the autobalance button, I think you are being overly concerned about abuse here. We arguably have several other features that are more ripe for abuse that are already in-lobby. That being said, if you really want, I'm open to adding some sort of toggle for party selection.

The abuse is simple: there's no consent. As an example:
 - if someone can choose their own party in the lobby, then everyone teams up with Tagada, if Tagada is in a party himself
 - if only the host can determine parties in the lobby then the host teams up with Tagada, if Tagada is in a party himself
 
And then either the host starts ejecting people, people need to leave the game because the host is derpy or your auto balance algorithm doesn't really make it balanced anymore. Either case it is an interaction that is undesired. 

> There are many times where you can't just switch two roughly equally rated players across teams in the lobby.

Yep, which also applies to your auto balancer as to the algorithm they do appear equal. Not being aware that one has an optimized seton air build and the other only plays gap. With that, fixing the parties / teams afterwards sounds more appealing to me as you're going to have to do that either way.

> There are also many times where there are multiple groups of of 2-4 people in the lobby who want to play together, which can get particularly tricky to balance sometimes. I've often seen hosts struggle with balancing multiple parties for several minutes straight, only to end up with a mediocre result or giving up.

A party of four is always tricky to balance. Your algorithm doesn't make it less tricky - it just increases your awareness that there is no balanced game while keeping that party intact. I don't consider this to be a solution to the problem, as the problem itself can not be solved regardless what you do sometimes.

It may be useful if there are multiple parties of two, but that doesn't happen too often.

> It really looks fine in the lobby now.

To you, yes. But to the average or new user the lobby (especially when you're a host) is quite crowded. And this doesn't make it less crowded.

>Additionally, an in-client party system that works with custom games is not coded ,nor do I expect it to be coded any time soon, if ever. By comparison, this is already coded and functional. That is a very significant difference.

This logic doesn't add up - it can be applied to any feature that currently doesn't exist.

> I imagine that this stepping stone being accomplished plus the increased expectation might drive someone/some people to code opti party support at some point.

Hope is not a strategy - we'll have to assume that it doesn't happen.

I think we've made quite a jump towards the wishes of some community members by introducing the auto balancer in the lobby. To quote [Phong](https://forum.faforever.com/topic/3635/in-lobby-auto-balancing/19) on the forums: `I've been using it all day and I am very satisfied with it. It works great even as it is now if I close the open slots before clicking.` If this particular feature is as desired, then we'll open up this pull request again and look into it for the patch after the 22nd. When that will be is yet to be defined.

Emperor-Penguin1:
@Garanas I was in the process of getting screenshots to demonstrate the changes that I had written today when you closed this...

I added a 'Party' button that locks/unlocks setting parties.  It is the small button that says 'Party' on it at the top of the 'Party' column.  The button only shows for the host, and it looks discreet.  So, if a host doesn't want to deal with parties in his/her lobby, it's easy to avoid them.

Here's what it looks like for a host with no parties and parties locked:
image

Here's what it looks like for a regular player with no parties and parties locked:
image

Here's what it looks like for a regular player with 1 party of 2 and parties locked:
image

As you can see, this enables hosts to have a basically party-less UI if desired, with just 1 discreet button.  It also covers a lot of your concerns about abuse.

Regarding concerns about the host abusing this to set unfair teams; the host can already do that without this PR.  This PR doesn't add that possibility.  It already exists.  It already happens.  If a host wants to set unfair teams, this is not at all a prerequisite.  Also, people can also still look at the balance percent in the top right.  If anything, this PR would actually make current stacking more overt if the party system is used for it.  And since this PR also makes balancing parties easier, people are more likely to leave stacked games and go host themselves with this PR's functionality.  Thus, this PR could actually help reduce stacking.  It also saves time when there are parties to balance and takes advantage of the balancing ability of the autobalancer.

Emperor-Penguin1:
Btw, if the 'Party' lock/unlock button is too discreet, just let me know, and I can make it more overt.

phongu:
> According to Phong on the forums, the latter doesn't take that much time in practice when done manually. At best it would be switching one or two roughly-equally rated players in the lobby, which is quite easy with a feature introduced by @4z0t that makes it a two click procedure. I can't find the pull request at the moment.

This is a bit out of context. I was trying to highlight the upsides of the new autobalancer compared to old opti balance, that provides no accommodation for premades whatsoever, or simple manual balance where i had to guess what swaps made the % better.

Hosting for premades is only easy (for me, i host 10 games a day minimum, dunno how a new host feels) once the players communicate their intentions properly, and, surprisingly, this can be really tedious because of this stupid tradition players have of referring to people using numbers, specifically in this case, slot numbers.

The worst offenders just pepper the chat with messages like "1 > 6" "3 > 8" in an attempt to get with their teammate. Others say something like "me and 5 together". Of course, they don't mention this while the lobby is filling, because they've already picked out slots on the same team and are happy. They only speak up once i start balancing. And once i'm swapping people around i'm payin attention to the map, not the chat, so i could easily miss their pleas in the flood of gray system messages generated by the swaps. And as soon as a swap is made, the numbers don't match the people anymore and even if i didn't touch slots 1, 3, 6 or 8 i'm not sure what slots i touched, since i use minimap swap.

So i have to get them to type out "me + nickname_of_friend" before i know what i'm doing. But wait, the players in question are non-english-speaking russians and simply sending "please tell me the nicknames of your friends" doesn't cut it. i have to run that through google translate first. And then they get the impression i'm a fluent russian speaker and reply with some cyrillic hieroglyphs and i have to point my mobile phone running google translate at my PC screen to decipher their request for clarification because, of course, one can't simply copy text from the lobby chatlog and paste it into google translate.

TL DR emperor_penguin's contribution is a step in the right direction for solving a real and very common communication problem, at the very least.

MrRowey:
I don't understand why we're making g the lobby more complex for the host and the having to spend more time in the lobby sorting things out if you want to play as a team host your own game ?. To me I think as much as I like the idea cutting the lobby with yet another colum the can be a abused is just going to cause more issuse than it helps. Like what you going to do is a group of 5 join and completely imbalance the game are you going to split them up ...

Garanas:
> TL DR emperor_penguin's contribution is a step in the right direction for solving a real and very common communication problem, at the very least.

I would agree, if it would prevent someone from joining an existing party without the parties consent. But at the moment it does allow that to happen. That smells like trouble to me - as I described earlier.

Penguin5:
> I don't understand why we're making g the lobby more complex for the host and the having to spend more time in the lobby sorting things out if you want to play as a team host your own game ?. To me I think as much as I like the idea cutting the lobby with yet another colum the can be a abused is just going to cause more issuse than it helps. Like what you going to do is a group of 5 join and completely imbalance the game are you going to split them up ...

This saves time for the host (see Phong's post and my previous explanations).  Also, a host does not have to accommodate parties with or without this system (although many people will leave games if not allowed to play on a team with their friends).  Additionally, the way the party system in this PR is coded, the host can avoid having parties with 1 click of the party button.  It can even be disabled by default if desired.

TLDR; It can be very useful for hosts who want to accommodate parties, and it can be easily disabled for hosts who don't.

Penguin5:
Alright, @Garanas, I added functionality to prevent someone from joining an existing party without an extant party member's consent.  When someone attempts to join an existing party, members of that party will see a clickable add button for that player's party.  Others will see the party letter with a question mark for that player's party (while that player is requesting to join the party but is not added to it).

image

image

image

image

Garanas:
Interesting - how much testing have you done so far?

Penguin5:
@Garanas I've tested a decent bit and fixed the edgecases I've found.  I think it's ready to go on FAF Develop now.

Garanas:
I've been thinking about this and I think it looks crowded because it is at the wrong place. It has nothing to do with a team. It has to do with the players. And in particular with the player names. What if we put it on the left of the player name?

image

Code-wise, is this a big change @Penguin5 ?

Penguin5:
@Garanas Given that the lobby code does not make changes like that convenient (properly adjusting the positioning for almost all of the columns and selectors, and their tooltips and buttons), I'd really rather not pursue that.  Additionally, I think it makes more sense to have it next to the team column anyway.  When parties are used for balancing, people in the same party would be put in the same team if possible.  So, it has a very close relationship with teams in that regard.

Sheikah45:
I would agree with Jip that parties are more related to the players and I think it would be better to the left if we add them.

Penguin5:
The player chooses all of their options (color/faction/team) in consecutive columns to the right of the playername.  I think it makes a lot of sense to have the party column there; it is related to team and is something that is selected in the same manner as those consecutive columns are (btw, iirc, it was actually Jip's suggestion to put the party column where I have it now).  I think moving it to the left of player names would be worse, but if someone else wants to do that and properly reformat things accordingly, he/she is welcome to.  I really don't want to waste hours on trying to get things reformatted just right (as part of the moving process), only to have the result be imperfect, and then have to spend additional hours trying to fix it again.  

PS: I already added the party lock feature, which can basically disable the party stuff and make the lobby look very similar to how it looks now.

Here's what it looks like for a regular player with no parties and parties locked:
image

Garanas:
Yep, that was initially my suggestion when there was no preview yet. Things change, and I don't think it fits where it is right now.

4z0t:
nice one

PR #3702, opened 2022-03-15
Optimize UEF build effects

Garanas:
Improves the performance of the UEF build effects, similar to how the Aeon, Cybran and Seraphim build effects were improved in the past. Upvalues functions accordingly, prevents table allocations, re-uses tables where possible and 

As par a suggestion of Sprouto to cache the entities the inspiration got a bit out of control where the default build beams now use projectiles too. And those can move from point to point, instead of jump from point to point. Looks a lot more faction unique. 

here-is-gif-19
here-is-gif-20
Examples with the UEF land factory

here-is-gif-22
here-is-gif-23
Example with the UEF Naval factory

Part of the Game development sessions series on Youtube:
 - [Default build beams](https://www.youtube.com/watch?v=qxtgH7frdKE&list=PL0nxuIUIjpFvqJ5i1HfPwoA8FnBCtGLWn&index=1)
 - [Optimize UEF slice beams: initial analysis](https://www.youtube.com/watch?v=qxtgH7frdKE&list=PL0nxuIUIjpFvqJ5i1HfPwoA8FnBCtGLWn&index=3)
 - [Optimize UEF slice beams: low hanging fruit](https://www.youtube.com/watch?v=qxtgH7frdKE&list=PL0nxuIUIjpFvqJ5i1HfPwoA8FnBCtGLWn&index=4)
 - [Optimize UEF slice beams: tricky fruit](https://www.youtube.com/watch?v=qxtgH7frdKE&list=PL0nxuIUIjpFvqJ5i1HfPwoA8FnBCtGLWn&index=5)
 - [Optimize UEF slice beams: finishing up](https://www.youtube.com/watch?v=qxtgH7frdKE&list=PL0nxuIUIjpFvqJ5i1HfPwoA8FnBCtGLWn&index=6)

PR #3773, opened 2022-04-08
Global Lua optimization. Part 1

KionX:
Continued #3511
A total of ~1000 files are expected.

Number of optimizations 4585(globals, math, table, string, exclude ui).

### DIFF
--- a/lua/aibrain.lua
+++ b/lua/aibrain.lua
@@ -294,6 +294,13 @@ AIBrain = Class(moho.aibrain_methods) {
         -- they are capitalized to match category names
         local layers = { "LAND", "AIR", "NAVAL" }
         local techs = { "TECH2", "TECH3" }
+    
+        self.Jammers = { }
+        setmetatable(self.Jammers, { __mode = 'v' })
+
+        self.JammerResetTime = 15
+
+        ForkThread(self.JammingToggleThread, self)
 
         -- populate the possible HQs per faction, layer and tech
         self.HQs = { }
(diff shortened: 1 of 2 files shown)


Garanas:
I think we should start simple: only rescope common table and math operations. This is considered a good improvement according to [the Lua book in the chapter 'Basic facts'](https://www.lua.org/gems/sample.pdf). It is also easy to review and it won't cause any hidden issues.

I agree with your last statement in #3511 that doing this type of work on all the files just makes them a lot more difficult to read. Instead, I think our focus should be on better memory management and refactoring hot-code to improve efficiency. An example of the former is https://github.com/FAForever/fa/pull/3743 that has a 10% improvement in performance when you have 900 ASF because of recycling of tables. An example of the latter is https://github.com/FAForever/fa/pull/3702, where we drastically reduce the amount of Lua byte code generated and prevent a lot of table operations by using mathematics.

KionX:
At the moment, the total number of microoptimizations is 25699.
Maybe it's worth disabling some optimizations?

I don't plan any further work on the existing FAF code and am considering whole remake this.

Garanas:
Yes - I'd say we limit it to:
 - globals (e.g., `CreateEmitterAtBone`)
 - math functions (as an example, `math.abs`)
 - table functions (as an example, `table.insert`)
 
And leave other engine and lua functions out of it, as they make the code a lot less readable.

Garanas:
And what does a remake of FAF entail? 😄

KionX:
I suggest a couple more optimizations:
Extract imports to global, for k,v in t do -> for k,v in next, t, nil do

The main idea of the FAF remake is that everything will be written in LuaJIT x64.
If necessary, other functions are added via Lua dll modules.
Graphics - DX9 (this is the simplest)
I care about the engine. I hope others will do the rest.

Garanas:
`Extract imports to global, for k,v in t do -> for k,v in next, t, nil do` 
I was not aware of this one, I did a few benchmarks. I do not notice a significant difference, how much do you think this change is faster?

`Also after optimization, you can rename everything for better readability.`
That is not viable / maintainable in the long term, especially if you want to integrate the optimizer.

I'd be fine with this list:
 - globals (e.g., CreateEmitterAtBone)
 - math functions (as an example, math.abs)
 - table functions (as an example, table.insert)
 - k,v in t do -> for k,v in next, t, if it is faster
 
Anything else is not required for the average code and can lead to errors down the route. As an example, if you'd optimize the brain and use `GetUnitsAroundPoint` or `GetListOfUnits` directly then you'd skip the Lua versions of those functions. The same applies to Lua functions in general, as mods / maps can override those (without hooking the file, quite destructive) and we'd still have a reference to the old function.

`I care about the engine. Hope others will do the rest.`
I don't know what type of changes this would require but I'd be open to discuss it and help you with it.

KionX:
Number of optimizations 4585(globals, math, table, string, exclude ui).
This PR is completed.

Garanas:
This is looking good. Surprised to see that the number of optimizations is still so high. Good idea to include the string functions too.

I'll look through it and experiment with it. Great work.

KionX:
Think about using this https://github.com/FAForever/FAFLuaJit for the lua test.
Example: https://github.com/FAForever/LOCChecker

KionX:
What's with this PR?

Garanas:
I've been testing it on the benchmark map and did not notice a significant change. I'm not sure if it is worth it.

KionX:
If it does not spoil the convenience, then it will not be worse.

PR #3776, opened 2022-04-08
Add teams 7 and 8 and make in-lobby team numbers Arial Bold

Penguin5:
Enables setting team to 7 or 8 in-lobby... these really should've been added earlier...

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -2819,9 +2819,9 @@ function CreateSlotsUI(makeLabel)
         factionSelector.OnEvent = defaultHandler
 
         -- Team
-        local teamSelector = Combo(newSlot, 17, 9, nil, nil, "UI_Tab_Rollover_01", "UI_Tab_Click_01")
+        local teamSelector = Combo(newSlot, 14, 9, nil, nil, "UI_Tab_Rollover_01", "UI_Tab_Click_01")
         teamSelector:AddItems({' - ', ' 1', ' 2', ' 3', ' 4', ' 5', ' 6', ' 7', ' 8'})
-        teamSelector._text:SetFont('Arial Bold', 17)
+        teamSelector._text:SetFont('Arial Bold', 14)
         teamSelector._titleColor = 'White'
         newSlot.team = teamSelector
         newSlot:AddChild(teamSelector)


Garanas:
The font is slightly different. in #3772 you managed to get the combo boxes working with text. Could you use the same font here, to make the team numbers instead of having images?

Emperor-Penguin1:
I could replace the images with text, but that would make team numbers 1-6 look different from how they have looked for many years AFAIK.  So, I think that is less desirable.

Garanas:
I think that is better over having a seven and an eight that looks different.

Emperor-Penguin1:
I still think it would be better to stick with the images.  I could remake the 8 to be blocky like the 6 is if desired.  I'll go ask others which set they'd rather use.  Here's the straw poll: https://strawpoll.vote/polls/h205umbw/vote

image

image

Garanas:
Can you describe which image is what?

Emperor-Penguin1:
The first image is the current style plus the new 7 and 8.  The second image is the new style that uses a font to redo all 8 numbers.  I'm fine with either tbh.  They're both mildly flawed (the 7 and 8 don't perfectly match the the style in the first one, but the second set changes the aesthetic and lacks the number thickness in the dropdown menu - I didn't see an easy enough way to properly change that), but I figure this visual comparison is fine to just get some votes on to decide.

Garanas:
We don't need votes on this. Can you make a commit to use the font over the images? It is more consistent - looking at it, the five and six feel slightly different over the one, two, three and four too.

Emperor-Penguin1:
Sure, and that makes sense about the 5 and 6, since those were previously added by FAF AFAIK.

Emperor-Penguin1:
Alright, I increased the size compared to the earlier example.  So,  now it looks like this.
image

Garanas:
image

Arial, 14 - just like the rest of the fonts. That looks good to me. It is different than the original, but it doesn't have to be the exact same.

Emperor-Penguin1:
I think it's better to have bigger/bolder team numbers in-lobby for easier identification at a glance.

Garanas:
To me it screams attention while it is irrelevant in the average lobby that is set to top vs bottom, left vs right or odds vs even.

Emperor-Penguin1:
I still think it would be better as 17, since different maps have different slot-team configurations for different AutoTeams types, but I switched it to 14.

Garanas:
These changes are (accidentally) part of #3782 .

PR #3779, opened 2022-04-10
Update init files

Garanas:
This causes issues with the new Aeon build animation, with thanks to Speed2 for taking the time to investigate it on voice.


BlackYps:
Where do you even get the "official" texturepack.nxt?

Garanas:
Good question - I have no idea. Maybe it is part of some of the old game types that we got rid of.

Garanas:
Introduced a 'deprecated' list of mods, in this case simspeed++ that directly changes shield interactions. The novel shield interactions introduced by #3681 is significantly better.

PR #3786, opened 2022-04-16
Add youtube graphics

Garanas:
Adds in the Youtube thumbnails made by Phong - ensures everything is in one place. They are part of the Large File System (lfs) of Github.


Garanas:
How do I confirm if they are indeed tracked by LFS?

PR #3795, opened 2022-04-19
Fix window resizing bug

4z0t:


Garanas:
What was the bug exactly?

lua/maui/window.lua
```diff
@@ -485,8 +484,8 @@ Window = Class(Group) {
                 self.Top:Set(top)
 
                 -- we can scale these accordingly as we applied the inverse on saving
-                self.Width:Set(LayoutHelpers.ScaleNumber(width))
-                self.Height:Set(LayoutHelpers.ScaleNumber(height))
+                self.Right:Set(LayoutHelpers.ScaleNumber(width) + left)
```
speed2CZ, line 487: this looks like smth that would break with scaling

4z0t:
On current dev branch, reaize chat window and move it, you will see the bug

lua/maui/window.lua
4z0t: Wdym?

4z0t:
It also happens with other windows, like minimap

Garanas:
interesting, I'll review it 😄

4z0t:
Basically the problem was in using Height and Width as reference for window size, it means  Bottom and Right are dependent on Top and Height and Left and Width, but resizing of window changes Top, Left, Bottom and Right causing window having 6 components being independent of each other.

4z0t:
@Garanas will you merge this?

PR #3796, opened 2022-04-20
Fix projectile-related issues

Garanas:
image

Fixes the issue where collision beams can destroy projectiles other than tactical missiles and strategic missiles. Re-applies some table reductions removed by #3794 , but this time correct and with more documentation.

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -646,8 +646,8 @@ Shield = Class(moho.shield_methods, Entity) {
         return IsEnemy(self.Army, other.Army)
     end,
 
-    --- Called when a unit collides with a collision beam to check if the collision is valid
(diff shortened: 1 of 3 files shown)


lua/shield.lua
```diff
@@ -637,7 +637,7 @@ Shield = Class(moho.shield_methods, Entity) {
         end
 
         -- special behavior for projectiles that represent strategic missiles
-        local otherHashedCats = other.BlueprintCache.HashedCats
+        local otherHashedCats = other.Cache.HashedCats
```
Garanas, line 640: Allows strategic missiles to skip shields again 😃

Garanas:
image

Test setup, where 200 mantis are continuously firing on the ground.

PR #3804, opened 2022-04-23
Fix issues with extractors

Garanas:
Fixes an issue reported by AlfaViTe#4489 on Discord where extractors would still produce mass even when they are paused. Removes some old TA-like code that caused an extractor to loop over all the markers of the map when it is created. This can easily be more than thousands of markers on AI-related maps.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -1064,24 +1064,6 @@ MassCollectionUnit = Class(StructureUnit) {
         adjacentUnit:RequestRefreshUI()
     end,
 
-    OnCreate = function(self)
-        StructureUnit.OnCreate(self)
-        local markers = ScenarioUtils.GetMarkers()
-        local unitPosition = self:GetPosition()
-
-        for _, v in pairs(markers) do
-            if v.type == 'MASS' then
-                local massPosition = v.position
-                if (massPosition[1] < unitPosition[1] + 1) and (massPosition[1] > unitPosition[1] - 1) and
-                    (massPosition[2] < unitPosition[2] + 1) and (massPosition[2] > unitPosition[2] - 1) and
-                    (massPosition[3] < unitPosition[3] + 1) and (massPosition[3] > unitPosition[3] - 1) then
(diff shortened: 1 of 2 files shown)


Sheikah45:
Oh wait it actually did use the amount of the marker? Why when I changed it then did that never affect the production rates

Garanas:
In a `_save.lua` file a mass marker is typically defined as:
```lua
                ['Mass 38'] = {
                    ['size'] = FLOAT( 1.000000 ),
                    ['resource'] = BOOLEAN( true ),
                    ['amount'] = FLOAT( 100.000000 ),
                    ['color'] = STRING( 'ff808080' ),
                    ['editorIcon'] = STRING( '/textures/editor/marker_mass.bmp' ),
                    ['type'] = STRING( 'Mass' ), -- <-- camel case
                    ['prop'] = STRING( '/env/common/props/markers/M_Mass_prop.bp' ),
                    ['orientation'] = VECTOR3( 0, -0, 0 ),
                    ['position'] = VECTOR3( 337.5, 74.9258, 449.5 ),
                },
````

The code in question:

```lua
        for _, v in pairs(markers) do
            if v.type == 'MASS' then -- <-- all upper case
                local massPosition = v.position
                if (massPosition[1] < unitPosition[1] + 1) and (massPosition[1] > unitPosition[1] - 1) and
                    (massPosition[2] < unitPosition[2] + 1) and (massPosition[2] > unitPosition[2] - 1) and
                    (massPosition[3] < unitPosition[3] + 1) and (massPosition[3] > unitPosition[3] - 1) then
                    self:SetProductionPerSecondMass(self:GetProductionPerSecondMass() * 10)
                    break
                end
            end
        end
```

It would never find the correct marker type.

Garanas:
And with that said - especially with Uveso's marker generator in play - when you start building any extractor (whether upgrading or directly) it would happily go over all the markers of the map 😄

Sheikah45:
Lol the name mismatch

speed2CZ:
Fix the string and see the bug reports coming it to find out how many maps have this :D

Issue #3820, opened 2022-04-26
Wrong file name


speed2CZ:
```
WARNING: Failed to create emitter as you passed in an invalid blueprint name /effects/emitters/antiair_muzzle_fire_01_emit.bp.
WARNING: Error running lua script: ...media\documents\github\fa\lua\sim\defaultweapons.lua(212): attempt to call method `ScaleEmitter' (a nil value)
         stack traceback:
         	...media\documents\github\fa\lua\sim\defaultweapons.lua(212): in function `PlayFxMuzzleSequence'
         	...media\documents\github\fa\lua\sim\defaultweapons.lua(663): in function <...media\documents\github\fa\lua\sim\defaultweapons.lua:596>
```
This file https://github.com/FAForever/fa/blob/deploy/fafdevelop/effects/emitters/antiair_muzzle_fire_01_emit%20.bp
has a space before .bp, that's what's most likely whats causing this error.

Garanas:
Interesting, I've seen that error myself but couldn't determine the cause. I'll try it without the space.

Issue #3818, opened 2022-04-26
Air units playing death explosion sound under water


speed2CZ:
If air unit is killed over water, it doesn't play the explosion sound when it hits the water (thats when the flash happens) but when it turns into wreck underwater, which should already be without an explosion.

Garanas:
I'll look at it.

Garanas:
Fixed with 3736.

PR #3822, opened 2022-04-26
Fix issues with release

Garanas:
Closes https://github.com/FAForever/fa/issues/3820 and https://github.com/FAForever/fa/issues/3818 . I suspect the latter has been an issue for a long time 😄 .


Garanas:
Also fixes:

```
warning: Error running lua command: g:\code\fa\lua\ui\dialogs\aiutilitiesview.lua(220): access to nonexistent global variable "CreateDefaultWindow"
```

And an error when a UEF build animation is stopped right after starting it.

PR #3823, opened 2022-04-26
Autobalance team spawn and shuffling

Penguin5:
Adds Autobalance as a possible choice for the team spawn lobby option (when enabled, this option choice balances with the same functionality that the in-lobby autobalancing does, but it does not allow the host to change the result in-lobby, nor does it allow viewing who will be on which team or in which slot before the launch button is pressed)
Also adds player pair shuffling to the new autobalancing (this affects in-lobby autobalancing as well, and it does not change who is on which team, nor does it intend to change who is mirrored against whom) - this adds slot randomness and has the effect of allowing the repeated clicking of the in-lobby autobalance button to shuffle player pairs around randomly to different slots


Garanas:
Why would we want this?

Penguin5:
It does not remove the old opti.  It offers an alternative option.  It balances better than opti does in some cases.  For example, it is way more likely to balance noobs (grays) better than opti does.  Additionally, in conjunction with #3772, this would add the ability/option to have autobalancing with parties without the result being editable by the host or seen before the launch button is pressed.  I want to stress that that would be an option.  So, people can choose.  I certainly would rather use this autobalancing than opti's autobalancing in many cases, and I prefer to have the ability/option to have this autobalancing be automatic and not stackable by the host after pressing the button.

Garanas:
It has no value for users to choose between two opti algorithms - we should only have one. 

Apparently opti doesn't do anything smart, I thought it used some Trueskill implementation. And it does so to check the quality, but not to build up the teams 😞 .

Crotalus:
Better balancing of grey / noob players are a bit of a contradiction, they are grey / noob for a reason. The trueskill rating / game quality is jibberish anyway in those cases.

How does this new autobalance team spot handle shuffling of players after game start (so you cannot find out position just by pressing F11)

Penguin5:
I think you misunderstood, @Crotalus.  I wasn't saying that this autobalancer shuffles more secretly than opti does.  I was just noting that this PR adds shuffling to this autobalancer, and that is desirable.  Also, this autobalancer balances both teams' base ratings and uncertainties. Presently, I have it set to care about balancing 1 point of base rating about as much as 2.5 points of uncertainty (that ratio can be changed, but I believe it has performed quite well so far).  Balancing both base rating and uncertainty like that is an advantage of this system.  It makes this autobalancer way more likely to balance noobs (grays) against each other more fairly than if it just balanced based on displayed rating or base rating.  By comparison, the regular opti balancing often distributes players with more uncertainty more disproportionately, and people often resort to manual balancing to avoid that.

To give a conceptual balancing example:
Player 1: 1000 displayed rating = 1200 base rating - 200 uncertainty
Player 2: 1100 displayed rating = 2300 base rating - 1200 uncertainty
Player 3: 1100 displayed rating = 1700 base rating - 600 uncertainty
player 4: 1200 displayed rating = 1400 base rating - 200 uncertainty

Given a situation like that, the new autobalancer would put players 1 and 2 vs players 3 and 4.  I think the large majority of FAF players who properly understand the rating system would find that more balanced than putting players 1 and 4 vs players 2 and 3, even though the displayed ratings would be more even that way.

To give a different conceptual balancing example:
Player 1: 200 displayed rating = 1500 base rating - 1500 uncertainty
Player 2: 500 displayed rating = 1700 base rating - 1200 uncertainty
Player 3: 1300 displayed rating = 1600 base rating - 300 uncertainty
player 4: 1400 displayed rating = 1600 base rating - 200 uncertainty

Given a situation like that, the new autobalancer would put players 1 and 4 vs players 2 and 3.  I think the vast majority of FAF players would find that more balanced than putting players 1 and 2 vs players 3 and 4, even though doing so would yield a more even base rating.

Penguin5:
@Garanas 
> It has no value for users to choose between two opti algorithms - we should only have one.
> 
> Apparently opti doesn't do anything smart, I thought it used some Trueskill implementation. And it does so to check the quality, but not to build up the teams 😞 .

Yes, as I have said; opti is flawed.

However, given that I like player choice, and given that people formed such a strong reaction about removing things like blinking lights or one of the pinks, I didn't even want to consider thinking about removing opti...

Given that there are many cases when the new autobalancer balances better than opti does, I do think there should be a team spawn option for the new autobalancer.  Additionally, its support of party functionality (in conjunction with #3772) is also desirable.  So, I think it (perhaps under opti's name, or a new name) should be a team spawn option, regardless of whether it replaces opti or coexists with it.

Crotalus:
Does your algorithm use trueskill quality measure (%Quality) when it finds the best setup? There will be a problem if the game looks unbalanced according to visible rating (mu - sigma*3) or quality %. 

Can't remember exactly how I made opti but I think it's trying a couple of different algorithms and then pick the best one according to trueskill quality, it should find the highest quality in most cases. 

What I meant with hidden/revealed positions was to make sure that functionality is still there if this replaces the current opti options:

SetupSession()
https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/simInit.lua#L74
ShuffleStartPositions()
https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/simInit.lua#L40

Penguin5:
> Does your algorithm use trueskill quality measure (%Quality) when it finds the best setup? There will be a problem if the game looks unbalanced according to visible rating (mu - sigma*3) or quality %.

My algorithm uses its own imbalance value calculations using players' mean (base rating) and deviation  (uncertainty / 3) data.  You can read my previous explanation and conceptual examples above.  While I could make it use trueskill's compute quality function, I don't think that would be better to do (trueskill is not perfect).  The system I have for calculating the imbalance value is quite functional, and it checks every single possible balance combination for splitting an even number of players into two equally-sized teams.  I can still tweak the exact formula used in calculating the imbalance value, but so far, it has performed quite well.

> Can't remember exactly how I made opti but I think it's trying a couple of different algorithms and then pick the best one according to trueskill quality, it should find the highest quality in most cases.

You are welcome to test the algorithms in practice and see how they compare.  My algorithm is currently in-use with the new in-lobby autobalance button.

> What I meant with hidden/revealed positions was to make sure that functionality is still there if this replaces the current opti options:

Regarding features that my autobalancer does not presently have; functionality to balance more than 2 teams, functionality to balance teams of unequal player count, and additional team spawn options for different shuffling/balancing/reveal results.

While I could add all of those things to it, I'd rather not do that at present.  I could investigate making it so that the opti and opti mirrored spawn options use my algorithm when they are selected and there is an even number of players on 2 teams (I can add a random shuffling option in my algorithm - it currently yields mirrored results).  However,  I'd only want to spend time on doing that stuff if those changes would actually get merged.  By comparison, I have already made this PR, which creates the new autobalance team spawn option, which uses my algorithm, and which would also support having parties (in conjunction with #3772).  So, I'd rather this PR just get merged and they coexist for now.

Crotalus:
So your main argument (trueskill isn't perfect) is that you've invented a generic trueskill quality method which is more efficient in calculating balanced team of different sizes than the mathematicians at Microsoft? 

Information about the underlying math here:
https://www.moserware.com/assets/computing-your-skill/The%20Math%20Behind%20TrueSkill.pdf

You probably need to come up with some mathematical proof / raw data why this algorithm create more balanced team and I don't mean games with random 400+ sigma players with <25 games. That uncertainty is just that, no one knows that player's skill. 

Another thing, I'm not sure it's a good idea to use the term "optimal balance" when potentially only one team has pre-made teams. Everyone (except the algorithm) knows that the other team will be at a disadvantage even if the team looks balanced on paper. So in reality this create a stacked setup unless there is a party on both sides. 

As someone mentioned earlier, aren't pre-made teams encouraged to use the matchmaker which already has too few players using it?

Penguin5:
@Crotalus You keep misunderstanding me, and you keep asking questions I've already answered... trueskill being imperfect was not my main argument, nor should it be a controversial claim (there are papers discussing the pros and cons of trueskill compared to some of its alternatives), and its not like using my algorithm removes trueskill from FAF anyway...

There are a variety of reasons that many thousands of FAF users prefer to play custom games a very large portion of the time (just look at the number of custom games vs the number of matchmaker games in a given month).  There is no need to restrict autobalancing of parties to the matchmaker and manual balancing when there is a functional and coded alternative that balances well.

"Optimal balance" seems like a misnomer tbh, as the current opti does not balance optimally in many cases.  Even when there are no parties or equivalent, people often resort to balancing manually to avoid opti's flaws.

I had said that "I didn't even want to consider thinking about removing opti..."  Please stop acting like I am advocating its removal.  I am just trying to add an alternative that balances better in some cases and also supports party functionality (in conjunction with #3772).

Regarding the idea of calculating a team as having a higher rating if it contains a party, that is something that I am open to and can readily add to my algorithm if desired.

Penguin5:
The autobalance feature has now been used thousands of times in numerous situations without issue.  I believe its balancing ability has been more than sufficiently demonstrated, and it should be added as a team spawn option.  The shuffling functionality this PR adds is desirable as well.  So, can this PR be merged?

Garanas:
Merging this in to make testing easier.

PR #3824, opened 2022-04-26
Improve performance of entity allocations

Garanas:
Commit 590a09855c17ab922047a32d6f74bf4f21e535ea introduced the replacement of the `entity:GetArmy()` engine call with a cached `entity.Army` value. But it takes it slightly too far. It is relevant for units, weapons, projectiles, shields and the brain. But raw entities are always dummy things, often created with no associated army and no need for one.

There is also a strange series of imports that causes a UI script to import the Entity script:
```lua
-- - unitview.lua -> import of game.lua (ui file)
-- - game.lua -> import of BuffField.lua (sim file)
-- - BuffField.lua -> import of Entity (sim file)
```

I'd like to remove this by moving the tables that are being used into a file that is part of the shared folder, introduced last December. However, at the moment I am not sure if that is a save change with regards to UI mods - I can't be certain if they are  independent of `game.lua`.


Garanas:
After doing some investigations, it appears the `game.lua` file was originally a UI-related file, that is not in the Ui folder. We introduced the creation of buffs into it at FAF in some point.

Garanas:
@speed2CZ do you have an idea what `BuffFields` are? They do not appear in the base code and the initialisation appears to always block itself:

```
    InitBuffFields = function(self)
        -- Creates all buff fields
        local bp = self.Blueprint
        if self.BuffFields and bp.BuffFields then -- <-- expects self.BuffFields field
            for scriptName, field in self.BuffFields do
                -- Getting buff field blueprint

                local BuffFieldBp = BuffFieldBlueprints[bp.BuffFields[scriptName]]
                if not BuffFieldBp or type(BuffFieldBp) ~= 'table' then
                    WARN('BuffField: no blueprint data for buff field '..repr(scriptName))
                else
                    -- We need a different buff field instance for each unit. This takes care of that.
                    if not self.MyBuffFields then
                        self.MyBuffFields = {} -- <-- but this is the only place it is populated with any value
                    end
                    self.MyBuffFields[scriptName] = self:CreateBuffField(scriptName, BuffFieldBp)
                end
            end
        end
    end,
```

The only unit that has a BuffFields attribute in the blueprints is the Seraphim commander, who doesn't cause the OnCreate of BuffFields to be called either.

Garanas:
@speed2CZ do you have an idea what `BuffFields` are? They do not appear in the base code and the initialisation appears to always block itself:

```
    InitBuffFields = function(self)
        -- Creates all buff fields
        local bp = self.Blueprint
        if self.BuffFields and bp.BuffFields then -- <-- expects self.BuffFields field
            for scriptName, field in self.BuffFields do
                -- Getting buff field blueprint

                local BuffFieldBp = BuffFieldBlueprints[bp.BuffFields[scriptName]]
                if not BuffFieldBp or type(BuffFieldBp) ~= 'table' then
                    WARN('BuffField: no blueprint data for buff field '..repr(scriptName))
                else
                    -- We need a different buff field instance for each unit. This takes care of that.
                    if not self.MyBuffFields then
                        self.MyBuffFields = {} -- <-- but this is the only place it is populated with any value
                    end
                    self.MyBuffFields[scriptName] = self:CreateBuffField(scriptName, BuffFieldBp)
                end
            end
        end
    end,
```

The only unit that has a BuffFields attribute in the blueprints is the Seraphim commander, who doesn't cause the OnCreate of BuffFields to be called either.

speed2CZ:
The sera regen field sound like what it was for, since that had some special complicated logic to it, but other than that, no.

Garanas:
It appears to be unused legacy code, introduced in 2011 by 5d1b52cc8b6c86e383ec6a1813a940da7d140598 . I did not manage to trigger the buff fields class. No references are made to it. The initialization blocks itself and the only reference in the blueprints I found across the base game and all the mods in my mods folder is the Seraphim commander. Which works just fine when these files and their references are removed.

Removing the code to prevent confusion in the future.

Garanas:
Fixed the strange import sequence.

PR #3826, opened 2022-04-27
Introduce .git-blame-ignore-revs

Garanas:
Allows us to ignore various large commits when using the blame view.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -2,6 +2,7 @@
 
 # Changes all lines in lobby.lua
 b5ad96a044b34f12b1c95121c4b49297fc0d243d
+27d8e201037bbf0b78a631e3baa42ce22c007dd0
 
 # Changes all line endings to LF
 76f8ac36b322bae6c0145e9da7b56e3e5dabb1ca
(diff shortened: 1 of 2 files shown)


Garanas:
@Crotalus it doesn't appear to work in visual studio code - do you know if I need to include a plugin for it to work?

Crotalus:
Seems like the file is a github invention so it only works in the Github Blame UI

Maybe VSCode will adopt it as well, both have the same MSFT daddy :)

Garanas:
We'll find out - at least this will clean up the Github Blame UI 😄

Crotalus:
Actually there is a way to get it to work in vscode if using gitlens:
https://github.com/gitkraken/vscode-gitlens/issues/947

Garanas:
Is that the file I just introduced? The naming convention appears to be off, but it is correct according to:
 - https://www.theserverside.com/blog/Coffee-Talk-Java-News-Stories-and-Opinions/Where-system-global-and-local-Windows-Git-config-files-are-saved

Crotalus:
I would guess they mean the file .git/config

In command line you can add to that file by doing:
git config --local blame.ignoreRevsFile .git-blame-ignore-revs

Garanas:
That makes more sense. Still doesn't work with the in-line pop up:

image

Garanas:
My mistake, it works! The configuration file is local, how can I automatically share that with every other user of the repository?

Garanas:
This is great - I can now see commits from eleven years ago 😄 .

Garanas:
I'll be adding in more commits in the future via other pull requests.

PR #3827, opened 2022-04-27
Remove deprecation warnings

Garanas:
Removes the deprecation warnings as it confused mod and map makers. It will come back in an alternative format that makes them easier to toggle on / off for the game developers.

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -906,37 +906,36 @@ Shield = Class(moho.shield_methods, Entity) {
 
         Main = function(self)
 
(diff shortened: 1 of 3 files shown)


Uveso:
The only thing that confused me was the fact that the game without mods also gives these warnings.
And because i need to have most of my mods compatible with the old game version, i can't patch those functions anyway.

Garanas:
> The only thing that confused me was the fact that the game without mods also gives these warnings.

The reason is because half the files are not in the repository. As long as that is the case then we can never get rid of all of them.

PR #3831, opened 2022-04-28
Increase modularity and testability in PenguinAutoBalance

Penguin5:
Increases modularity/testability for the new autobalancer via the DetermineTeams function, which can generate testing data by being called  with (true, true)


Garanas:
This is great - is it decoupled enough that you could place it in another file, and import that accordingly? That would reduce the amount of code in the lobby.

Penguin5:
Theoretically, but when I tested that, it would not work in its current location (for the in-lobby autobalance button) because of an error for too many upvalues (limit of 32).

Garanas:
Ah, yes - that is a known issue 😞 .

Penguin5:
So, can this be merged?

PR #3833, opened 2022-04-30
Improve changelog dialog

Garanas:
Creates a new interface for the in-game changelog. It emphasis the number of patches by using a list instead of a combo box. The buttons link to:
 - Github: http://github.com/FAForever/fa/blob/develop/changelog.md
  - Bug report: http://github.com/FAForever/fa/wiki/report-a-bug
  - Balance notes: http://patchnotes.faforever.com/3725.html, where 3725 is replaced with the version number
  
Closes #3812 .

PR #3835, opened 2022-05-01
Layouter as alternative for LayoutHelpers

4z0t:
Examples:
```lua
local group = Group(parent)
    LayoutFor(group)
        :Height(500)
        :Width(500)
        :AtCenterIn(parent)

    group.popup = Popup(parent, group)
    LayoutFor(group):Over(group.popup, 10)

    group.TopLine = 1
    group.SizeLine = table.getsize(globalOptions)

    group.Title = UIUtil.CreateText(group, 'UI Mods Options', 16, UIUtil.titleFont, true)
    LayoutFor(group.Title)
        :AtHorizontalCenterIn(group)
        :AtTopIn(group, 5)

    group.scroll = UIUtil.CreateLobbyVertScrollbar(group, -20, 10, 25) -- scroller
    LayoutFor(group.scroll)
        :Over(group, 10)

    group.QuitButton = UIUtil.CreateButtonWithDropshadow(group, '/BUTTON/medium/', LOC("<LOC _Close>Close"))
    LayoutFor(group.QuitButton)
        :AtHorizontalCenterIn(group)
        :AtBottomIn(group, 5)
        :Over(group, 20)
```
```lua
local group = Group(parent)
    LayoutHelpers.SetDimensions(group, 500, 500)
    LayoutHelpers.AtCenterIn(group, parent)

    group.popup = Popup(parent, group)
    LayoutHelpers.DepthOverParent(group, group.popup, 10)

    group.TopLine = 1
    group.SizeLine = table.getsize(globalOptions)

    group.Title = UIUtil.CreateText(group, 'UI Mods Options', 16, UIUtil.titleFont, true)
    LayoutHelpers.AtHorizontalCenterIn(group.Title, group)
    LayoutHelpers.AtTopIn(group.Title, group, 5)

    group.scroll = UIUtil.CreateLobbyVertScrollbar(group, -20, 10, 25) -- scroller
    LayoutHelpers.DepthOverParent(group.scroll, group, 10)

    group.QuitButton = UIUtil.CreateButtonWithDropshadow(group, '/BUTTON/medium/', LOC("<LOC _Close>Close"))
    LayoutHelpers.AtHorizontalCenterIn(group.QuitButton, group, 0)
    LayoutHelpers.AtBottomIn(group.QuitButton, group, 5)
    LayoutHelpers.DepthOverParent(group.QuitButton, group, 50)
```
[Fluent interface](https://en.wikipedia.org/wiki/Fluent_interface)

### DIFF
--- a/lua/maui/layouthelpers.lua
+++ b/lua/maui/layouthelpers.lua
@@ -525,8 +525,8 @@ function LayouterMetaTable:HitTest(bool)
     return self
 end
 
-function LayouterMetaTable:NeedsFrameUpdate(state)
-    self.c:SetNeedsFrameUpdate(state)
+function LayouterMetaTable:NeedsFrameUpdate(bool)
+    self.c:SetNeedsFrameUpdate(bool)
     return self
 end
 
--- a/lua/maui/layouthelpers.lua
+++ b/lua/maui/layouthelpers.lua
@@ -476,6 +476,10 @@ function InvScaleNumber(number)
     return math.floor(number * (1 / pixelScaleFactor))
 end
 
+--**********************************
+--*********  Layouter  *************
+--**********************************
+
 local LayouterMetaTable = {}
 LayouterMetaTable.__index = LayouterMetaTable
 
--- a/lua/maui/layouthelpers.lua
+++ b/lua/maui/layouthelpers.lua
@@ -692,7 +692,7 @@ end
 
 function LayouterMetaTable:AtTopCenterIn(parent, offset, horizonalOffset)
     AtTopIn(self.c, parent, offset)
-    AtHorizontalCenterIn(self.c, parent, offset)
+    AtHorizontalCenterIn(self.c, parent, horizonalOffset)
     return self
 end
 
--- a/lua/maui/layouthelpers.lua
+++ b/lua/maui/layouthelpers.lua
@@ -501,17 +501,18 @@ function LayouterMetaTable:Hide()
 end
 
 function LayouterMetaTable:Color(color)
-    self.c:SetColor(color)
+    if self.c.SetSolidColor then
+        self.c:SetSolidColor(color)
+    elseif self.c.SetColor then
+        self.c:SetColor(color)
+    else
+        WARN(string.format("Unable to set color for control \"%s\"", self.c:GetName()))
+    end
     return self
 end
 
 function LayouterMetaTable:DropShadow(bool)
     self.c:SetDropShadow(bool)
-    return self
-end
-
-function LayouterMetaTable:SolidColor(color)
-    self.c:SetSolidColor(color)
     return self
 end
 
--- a/lua/maui/layouthelpers.lua
+++ b/lua/maui/layouthelpers.lua
@@ -1,5 +1,5 @@
 --*****************************************************************************
---* File: lua/modules/maui/layouthelpers.lua
+--* File: lua/modules/maui/lua
 --* Author: Chris Blackwell
 --* Summary: functions that make it simpler to set up control layouts.
 --*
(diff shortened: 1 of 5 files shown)


Garanas:
Could you copy / paste the code of the screenshots into your original post, instead of the screenshots?

lua/maui/layouthelpers.lua
```diff
@@ -474,4 +474,336 @@ end
...
+end
+
+--**********************************
+--*********  Layouter  *************
+--**********************************
+
+local LayouterMetaTable = {}
+LayouterMetaTable.__index = LayouterMetaTable
+
+function LayouterMetaTable:Disable()
```
Garanas, line 486: All these functions need a proper comment describing what it does, especially if we intend to make snippets as those are taken into account.

lua/maui/layouthelpers.lua
```diff
@@ -474,4 +474,336 @@ end
...
+    self.c:SetColor(color)
+    return self
+end
+
+function LayouterMetaTable:DropShadow(state)
+    self.c:SetDropShadow(state)
+    return self
+end
+
+function LayouterMetaTable:BitmapColor(color)
```
Garanas, line 506: These (and others) should just match the original name. As a few examples:
`LayouterMetaTable:TextColor` -> `LayouterMetaTable:SetColor`
`LayouterMetaTable:BitmapColor` -> `LayouterMetaTable:SetSolidColor`

That prevents contributors from having to learn a new syntax.

lua/maui/layouthelpers.lua
```diff
@@ -474,4 +474,336 @@ end
...
+    end
+    if state then
+        self.c:EnableHitTest()
+    else
+        self.c:DisableHitTest()
+    end
+    return self
+end
+
+function LayouterMetaTable:NeedsFrameUpdate(state)
```
Garanas, line 528: Do not use ambiguous names for values, I think this is a boolean. Call it a `bool` instead of `state`.

lua/maui/layouthelpers.lua
```diff
@@ -474,4 +474,336 @@ end
...
+end
+
+function LayouterMetaTable:Texture(texture)
+    self.c:SetTexture(texture)
+    return self
+end
+
+function LayouterMetaTable:HitTest(state)
+    if state == nil then
+        error(":HitTest requires 1 positional argument \"state\"")
```
Garanas, line 518: Do not use errors - use warnings. An error stops code executing, causing a gazillion lazy var errors. You can inform the user that something is wrong, and then tell them that it skipped processing the call in question.

lua/maui/layouthelpers.lua
```diff
@@ -474,4 +474,336 @@ end
...
+
+function LayouterMetaTable:ResetWidth()
+    ResetWidth(self.c)
+    return self
+end
+
+-- get control --
+
+function LayouterMetaTable:Get()
+    return self.c
```
Garanas, line 786: It would be interesting here to see if we can add some error-managing code to detect the lazy var loops that can happen. I'm not 100% confident, but I think every control inherits from the Control class. That resets the layout here:
 - https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/maui/control.lua#L33

It would be interesting to:
 - add an 'isDefault' flag to lazy vars, that is set to false after any Set function is called
 - Manually set the isDefault flag to true in the `ResetLayouts` function, after setting them to their default values
 - in this Get function, check whether (for example) both the left and the right definition are default. if they both are, then we are missing information. Give them a default value (left set to 0, right set to screen width - idk) and provide a warning indicating what you need to do to fix it.
 
And then for each combination of errors. The following are correct:
 - left and right defined
 - left and width defined
 - right and width defined
 - top and bottom defined
 - top and height defined
 - bottom and height defined
 
Anything else is incorrect and causes a lazy var loop.

Garanas (changes requested):
And we should also look into generating the snippets for this builder pattern.

4z0t:
> Could you copy / paste the code of the screenshots into your original post, instead of the screenshots?
Done!

lua/maui/layouthelpers.lua
Garanas: I also think it should check if control-specific properties exist, and if it doesn't, it gives the user a friendly warning instead of crashing.

4z0t:
@Garanas  I think it is possible, BUT:
* user can manually do a circular dependency, that is trackable by pcall, BUT
* parent control may have uncompleted layout whilest layouting child causing circular dependency
* I don't think it is worth doing this since finding circular deps is not that hard

Garanas:
> I don't think it is worth doing this since finding circular deps is not that hard

I disagree. Even with #3833 I frequently was at a loss what was causing the issue. Let alone for someone not as experienced as us that encounters a stream of errors.

Let's let it sink in a bit. Because having some more sane error messaging would help us, but also any UI modder and future contributor to understand what is going wrong.

lua/maui/layouthelpers.lua
4z0t: > I also think it should check if control-specific properties exist, and if it doesn't, it gives the user a friendly warning instead of crashing.

I saw `IsInstance` like function, how suitable is it?

Garanas: I was more thinking in terms of:

```lua
function LayouterMetaTable:DropShadow(state)
    self.c:SetDropShadow(state)
    return self
end
```
to
```lua
function LayouterMetaTable:DropShadow(state)
    if self.c.SetDropShadow then 
        self.c:SetDropShadow(state)
    else
        WARN("UI element is likely no text and therefore you can not set a drop shadow. Skipping this instruction.)
    end
    return self
end
```

4z0t:
excuse me where are my changes

lua/maui/layouthelpers.lua
4z0t: > I was more thinking in terms of:
> 
> ```lua
> function LayouterMetaTable:DropShadow(state)
>     self.c:SetDropShadow(state)
>     return self
> end
> ```
> 
> to
> 
> ```lua
> function LayouterMetaTable:DropShadow(state)
>     if self.c.SetDropShadow then 
>         self.c:SetDropShadow(state)
>     else
>         WARN("UI element is likely no text and therefore you can not set a drop shadow. Skipping this instruction.)
>     end
>     return self
> end
> ```

if user does regularly it like `bitmap:c:SetDropShadow(true)` he gets error anyway, so, I dont think skipping problems is a good idea.

4z0t:
Are you ok with this implementation of End?

Garanas:
> @Garanas also, what do you think about merging color method for both text and bitmap?

Sounds good to me.

Garanas:
> Are you ok with this implementation of End?

I'll look at it later this week - there is no rush, the release is not ready yet anyhow 😄 .

lua/maui/layouthelpers.lua
Ejsstiil: Combinations of loop errors, nice! This should go somewhere to Dev Wiki

4z0t:
Should I start making docs now or waiting for vs code extension?

PR #3836, opened 2022-05-03
Kick Observer Fix

Penguin5:
Fixes the kick observer bug that had to do with the incorrect observer index being  used when clicking a row on the observer list when team rating row(s) were present
Also removes the ability for the host to get the dialog to try to kick himself...


Garanas:
All good, well done. Thank you for being so spot on with the fix.

PR #3842, opened 2022-05-06
Add New Patchnotes Button

MrRowey:
Add a new patchnotes button next to where the options/mod manager button is and link to the new patchnotes screen.
image

And removed the ability to access the patchnotes from the Game Patch X text
image


Garanas:
Looking good 😄

PR #3848, opened 2022-05-08
Layouthelpers: fix functions' behaviour

4z0t:


Garanas:
Are these functions used anywhere? Do they still work accordingly?

4z0t:
No, they are not used anywhere

Issue #3282, opened 2021-05-12
AI not finishing builds on Blasted Rock


S7R7RPR:
Been a long time user of FAF and discovered a weird bug on Blasted Rock. Played against AIX and SorianAIX with the same issues. The commanders did some really weird things and then watching the replay I saw the commander starting a lot of builds and then just not finishing them. Not sure if it's addressable but figured I might as well report it!

https://replay.faforever.com/14493966
https://replay.faforever.com/14493990

Uveso:
Both games desync, so maybe i did not see how it was in reality;

1st game, UEF AI ACU left the factory building to attack the middle of the map
2nd game, both AI ACUs are interrupting the build because of enemy threat.
(The ACU should attack in this situation and not continue to build; One unfinished building would be normal)

Sadly i can't reproduce the error on my PC.
Blasted Rock is one of my AI test maps and i played hundreds of AI games on it.

Do me a favor; please install my AI mod and try the map again.
(My AI has extended debug function to trace down any AI errors)

https://forum.faforever.com/topic/350/

S7R7RPR:
Alright so I tried again on my own with both AIs and they performed more or less the same
https://replay.faforever.com/14499998

I then downloaded your AI from the vault and ran with that and the AI performed much much better.
https://replay.faforever.com/14500060

Based on the behavior comparison, I'm going to hazard a guess that it was a resource based bug. In the second game, I noticed the commanders kept just chilling and not doing anything for a bit. They started a few builds and moved on, but not nearly as many as before.

Uveso:
I was able to reproduce the error where the ACU was doing nothing.
It turned out that there was a small error inside the `ProcessBuildCommand` function.

I added a gamepatch to my AI and fixed the `ProcessBuildCommand`.
You can test it if you like. Just download and enable my develop AI mod.
You don't need to play against AI-Uveso, its also working for Sorian and GPG AI.

Please have in mind, this is a develop version of my AI, don't play it online!
[AI-Uveso-Mod-S7R7RPR.zip](https://github.com/FAForever/fa/files/6470467/AI-Uveso-Mod-S7R7RPR.zip)

In case you are interested in AI development, support or just information, join us on our AI developer Discord:
https://discord.gg/ChRfhB3

Garanas:
Closing this issue - @Uveso can you make a pull request to the repository with the changes you made to your mod to fix this?

Uveso:
sure, i will make the PR tomorrow.

PR #3851, opened 2022-05-08
Reduce unfinished buildings

Uveso:
Fix for #3282

The ACU can still stop building when enemies are nearby, but this is intended.

Note:
To resume the construction it's needed to fix the AIBehaviors.lua.CommanderThreadImproved() function.
At the moment this will be compensated by the "ManagerEngineerFindUnfinished" platoon.

Before:
PBC1

after:
PBC2

### DIFF
--- a/lua/platoon.lua
+++ b/lua/platoon.lua
@@ -3574,10 +3574,9 @@ Platoon = Class(moho.platoon_methods) {
             return
         end
         local aiBrain = eng.PlatoonHandle:GetBrain()
-
         if not aiBrain or eng.Dead or not eng.EngineerBuildQueue or table.empty(eng.EngineerBuildQueue) then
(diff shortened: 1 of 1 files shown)


Garanas:
Looking good - I'll playtest it tomorrow and merge accordingly. Thank you for supplying the fix.

Uveso:
The process build command gets interupted from outside but the function itself is working.
Mainly this patch should prevent the ACU from building nothing,
and engineers should not try to cap a mass spot where already another engineer is on the way.
The reduce in unfinished buildings was only a side effect.

Take a look at AIBehaviors.lua in CommanderThreadImproved() function.
As far as I remember the overcharge function breaks the build and the resume build doesn't work at all.

Test this without enemies so that the Overcharge function has no reason to interrupt the building process,
or spawn single enemy units into the AI base to interrupt it even more.

PR #3857, opened 2022-05-11
Improve performance of weaponry

Garanas:
See the forums for a full description:
 - https://forum.faforever.com/topic/3808/weapon-target-check-intervals


lua/system/blueprints-weapons.lua
```diff
@@ -0,0 +1,163 @@
...
+        weapon.AlwaysRecheckTarget = false 
+        weapon.TrackingRadius = 1.15
+        weapon.DummyWeapon = false 
+        weapon.ManualFire = false
+        return 
+    end
+
+    -- - process target check interval
+
+    weapon.TargetCheckInterval = math.min(
```
aeoncleanse, line 61: Move this section up to avoid doing countermeasures twice

lua/system/blueprints-weapons.lua
```diff
@@ -0,0 +1,163 @@
...
+            0.5 * (1 / (weapon.RateOfFire or 1))    -- based on attack rate
+        ,   (weapon.MaxRadius or 10) / 40           -- based on attack range
+    )   
+
+    -- except for counter measure weaponry
+    if weapon.RangeCategory == "UWRC_Countermeasure" then 
+        weapon.TargetCheckInterval = 0.5 
+    end
+
+    -- clamp value to something sane
```
aeoncleanse, line 71: All this could be done more cleanly using a math clamp at the calculation point above.

lua/system/blueprints-weapons.lua
```diff
@@ -0,0 +1,163 @@
...
+    weapon.TargetCheckInterval = 0.1 * math.floor(10 * weapon.TargetCheckInterval)
+
+    -- - process target tracking radius 
+
+    -- by default, give every unit a 15% target checking radius
+    weapon.TrackingRadius = 1.15
+
+    -- remove target tracking radius for non-aa weaponry part of structures
+    if isStructure and (weapon.RangeCategory ~= "UWRC_AntiAir") then 
+        weapon.TrackingRadius = 1.0
```
aeoncleanse, line 96: This will be quite a noticeable nerf to any PD or Arty with a slow turret rotation, particularly Ravager and Oblivion. Just make sure that's taken into account.

lua/system/blueprints-weapons.lua
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    -- give anti air a larger track radius
+    if weapon.RangeCategory == "UWRC_AntiAir" then 
+        weapon.TrackingRadius = 1.3 
+    end
+    
+    -- add significant target checking radius for bombers
+    if isBomber then 
+        weapon.TrackingRadius = 1.5
```
aeoncleanse, line 106: This will affect AA weapons on Bombers as well...

lua/system/blueprints-weapons.lua
```diff
@@ -0,0 +1,163 @@
+
+--- We don't want to change certain unit categories as they are quite sensitive to change
+local categoriesToSkip = { }
+categoriesToSkip.COMMAND = true
```
aeoncleanse, line 4: Why not a literal object declaration?

lua/system/blueprints-weapons.lua
```diff
@@ -0,0 +1,163 @@
+
+--- We don't want to change certain unit categories as they are quite sensitive to change
+local categoriesToSkip = { }
+categoriesToSkip.COMMAND = true 
+
+local weaponTargetCheckUpperLimit = 6000
+
+-- Multiplier depending on range category
+local weaponTargetCheckMultiplier = { }
```
aeoncleanse, line 9: Again, literal declaration here is probably easier to understand

lua/system/blueprints-weapons.lua
```diff
@@ -0,0 +1,163 @@
...
+    end
+
+    -- always allow experimentals to retarget
+    if isExperimental then 
+        weapon.AlwaysRecheckTarget = true 
+    end
+
+    -- do not allow bombers to suddenly retarget, as then they won't drop their bomb then
+    if isBomber then 
+        weapon.AlwaysRecheckTarget = false
```
aeoncleanse, line 132: good

lua/system/blueprints-weapons.lua
```diff
@@ -0,0 +1,163 @@
...
+        weapon.TargetCheckInterval = 0.1 
+    end
+
+    -- clamp value to something sane
+    if weapon.TargetCheckInterval > 10 then 
+        weapon.TargetCheckInterval = 10 
+    end
+
+    -- sanitize it
+    weapon.TargetCheckInterval = 0.1 * math.floor(10 * weapon.TargetCheckInterval)
```
aeoncleanse, line 87: Good

aeoncleanse (commented):
I'm really glad someone is finally taking the time to address this issue, it flagged years ago when we ran test scenarios through the profiler, but I never got around to actually fixing it.

If you want to do it really properly, I recommend running this logic against the files _outside_ of the game, by just running the lua through a simple standalone utility, baking the changes into the unit blueprint files. This way, any unit databases drawing from them will be accurate, and you avoid devs looking at the unit bp and wondering why their changes aren't sticking if they're unaware of this code.

lua/system/blueprints-weapons.lua
Garanas: You're correct - but it also prevents the following scenario:
 - https://forum.faforever.com/topic/3805/cybran-t2pd-having-problems-targeting-and-not-firing-at-targets-in-range

lua/system/blueprints-weapons.lua
Garanas: You're correct - I'll look into it

lua/system/blueprints-weapons.lua
Garanas: As in, inline?

lua/system/blueprints-weapons.lua
Garanas: I'll convert it as the pull request gets finalized. This is easier to work with

Garanas:
> If you want to do it really properly, I recommend running this logic against the files outside of the game, by just running the lua through a simple standalone utility, baking the changes into the unit blueprint files. This way, any unit databases drawing from them will be accurate, and you avoid devs looking at the unit bp and wondering why their changes aren't sticking if they're unaware of this code.

I agree, but it also leaves the illusion that those values are up for changes. As they have a significant influence on performance they should never be manually set to gain some arbitrary advantage.

PR #3863, opened 2022-05-14
xab3301

MadMaxFAF:
1st draft of activate animation

i think adding the spinning part will be cleaner to do in script

image

https://user-images.githubusercontent.com/62246181/168422579-a8ff29d7-f3fe-4768-8e0a-38411fa132f2.mp4


Garanas:
Looking good. We can do the spinning part in script, yes.

Garanas:
Spinning of the bottom part works, can you make it so that I can also spin the top part?

image

MadMaxFAF:
imo i think that will be too many spinning parts the top opens up and there are bones for effects there

MadMaxFAF:
yeah i was thinking about having the spinning being for on or off states and the opening part be for when it's using it's ability

Garanas:
At the moment it looks slightly off though, where we spin the bottom part but the top part is completely immobile. We can always test it, see how it looks.

Garanas:
here-is-gif-33

Current status quo 😄

PR #3873, opened 2022-05-17
T3 MMA Refactor Unit Naming

MrRowey:
- Updated Unit Names to Fit more seamlessly with the FA Naming scheme
- Moved the Effects in the Sera T3 MMA file to the `/Effects/Emitters` folder


Garanas:
I'm fine with losing some of the Github history on these files. But do you know whether these changes affect for example the unit database?

MrRowey:
the unit database should jsut need the update script running and it looks at all the BP file and should change accordingly

units/DEL0307/DEL0307_unit.bp
```diff
@@ -97,7 +97,7 @@ UnitBlueprint {
                     ScrollTreads = true,
                     TreadMarks = {
                         {
-                            BoneName = 'DELK002',
```
Garanas, line 100: Did you also change the bone names in the mesh?

MrRowey: Ooof that one slipped through and yes I rename mesh bones

MrRowey: xD that red is a deletsion thight Id missed it

Garanas: Good - I'll review it more thoroughly later this week.

MrRowey:
I suck to using `D` Over `U` for the Naming as they were Technically DLC to faf so when with that rather than classing them as Original units

speed2CZ:
Why would you not give them the same number? Since they are all T3 MAAs?

MrRowey:
> 

this was because of the way the factions are different due to some of the Base units already occupying that number . I can changed them again to make sure that they all have the same number but may fall into the same issues

MrRowey:
image
 Main Line T3 tanks are not he same Numbers so this should not be an issues Unless we end up going around Re numbering them all

Garanas:
I thought those were the main battle tanks

Garanas:
Oh, that is your point. Nevermind. 

The 'original' main battle tanks are Loyalists / Titians

MrRowey:
@Garanas So do i need to rename then all so they are the same (Cybran one wont fit with this) then ?

speed2CZ:
I dont see how they wont fit, they are all primary T3 mobile AA
Also what you should keep in mind when renaming units, you will need to rename all the AI files that use there units. Many coop missions also use them.

MrRowey:
i mean we can just move the effects but i do think that the units should have the proper naming scheme and not the one that they came with from the mod as it makes it easier to understand what the unit is based of its name ID . at the end of the day im not the one merging it if we don't want to make this change that fine. but is personally been a nucence that they arent using the same naming schem as the res of the units.

Garanas:
Speed2 is right. As an example, the blueprint name is hardcoded in the survival script of Packer, see his [unit tables](https://github.com/Packer/FAF-Survival-Genesis-Blank/blob/main/survival_genesis_blank.v0001/units/Units_Base.lua#L200). It can also be hardcoded into the `_save.lua` files of various maps, such as campaign and other scripted maps. We'd break all of that content with this change.

Changing the folder name / blueprint id of this unit is no longer an option. This should've been done when the unit was integrated.

PR #3874, opened 2022-05-19
Improve performance of intel management

Garanas:
There are some performance issues with intel management:
 - All units allocate a table of tables containing all the possible intel that they may propagate, even though they may not have that intel type
 - The functions that manage intel are not efficient: a closure is created on each change, inefficient use of tables and constant computation of values that are static on the majority of units.


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+    "nuxt.isNuxtApp": false
```
Garanas, line 2: What 😆

Garanas:
Some quick numbers, the current implementation allocates this table for all units:

```lua
self.IntelDisables = {
    Radar = {NotInitialized = true},
    Sonar = {NotInitialized = true},
    Omni = {NotInitialized = true},
    RadarStealth = {NotInitialized = true},
    SonarStealth = {NotInitialized = true},
    RadarStealthField = {NotInitialized = true},
    SonarStealthField = {NotInitialized = true},
    Cloak = {NotInitialized = true},
    CloakField = {NotInitialized = true}, -- We really shouldn't use this. Cloak/Stealth fields are pretty busted
    Spoof = {NotInitialized = true},
    Jammer = {NotInitialized = true},
}
```

The following numbers are from @KionX on Zulip:

```
Each entry in the hash table is 20 bytes.
The size of the hash table is rounded to the next power of 2 (64 records - 128 capacity).
Unused capacity remains reserved.

Each entry in the array is 8 bytes. The size of the array is not rounded. Unused capacity remains reserved.
```

Based on that we use:
 - 16 x 20 bytes for `IntelDisables` itself
 - 11 x 2 x 20 bytes for each sub table
 - 1 x 20 bytes for the entry into self
 ----------+
 -       780 bytes

Late game we can easily have 5000 units or more. They all have this intel table allocated, whether they need the entries or not. That means we have at least 5000 * 780 bytes = 3.9 megabytes of information that is solely used for intel management. That is a lot of data, especially as our game can only take up about 2gb of ram 😄

Garanas:
Closing this pull request - someone is free to pick it up again.

PR #3875, opened 2022-05-19
Reduce memory impact

Garanas:
From the GPG era, over the years on FAF and throughout the last year of FAF we've been consistently adding fields to units. As it turns out, these fields are expensive: each field takes up 20 bytes of information. In this pull request we carefully look at tables that come in large quantities (units, projectiles, props, weapons) and carefully evaluate if a field should exist or not.

### DIFF
--- a/.git-blame-ignore-revs
+++ b/.git-blame-ignore-revs
@@ -7,5 +7,6 @@ b5ad96a044b34f12b1c95121c4b49297fc0d243d
 # Changes all line endings to LF
 76f8ac36b322bae6c0145e9da7b56e3e5dabb1ca
 
-# Global white space cleanup
+# white space cleanups
 16d5cfebac6ddd19a3a9496b810beb9a1a73d950
+ce65d77916260cb29d021a036d484e84bd38eabd
--- a/lua/EffectUtilitiesAeon.lua
+++ b/lua/EffectUtilitiesAeon.lua
@@ -82,7 +82,7 @@ local function SharedBuildThread(pool, unitBeingBuilt, unitBeingBuiltTrash, unit
     if EntityCategoryContains(CategoriesHover, unitBeingBuilt) then 
 
         -- set elevation offset
-        offset = unitBeingBuilt.Elevation or 0
+        offset = unitBeingBuilt.Blueprint.Elevation or 0
 
         -- create a slider
         slider = CreateSlider(unitBeingBuilt, 0)
@@ -132,7 +132,7 @@ local function SharedBuildThread(pool, unitBeingBuilt, unitBeingBuiltTrash, unit
     end
 
     -- set correct shader of unitBeingBuilt so that it happens instantly after finishing
-    unitBeingBuilt:SetMesh(unitBeingBuilt.MeshBlueprint, true)
+    unitBeingBuilt:SetMesh(unitBeingBuilt.Blueprint.Display.MeshBlueprint, true)
 end
 
 --- The build animation for Aeon buildings in general.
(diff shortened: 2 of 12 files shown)


Garanas:
According to the numbers of KionX on Zulip:
```
Each entry in the hash table is 20 bytes.
The size of the hash table is rounded to the next power of 2 (64 records - 128 capacity).
Unused capacity remains reserved.

Each entry in the array is 8 bytes. The size of the array is not rounded. Unused capacity remains reserved.
```

One example from the pull request is to populate the `EventCallbacks` table as we add callbacks. In a non-campaign game, almost no unit uses callbacks of any kind. 

```
self.EventCallbacks = {
    -- OnKilled = {},
    -- OnUnitBuilt = {},
    -- OnStartBuild = {},
    -- OnReclaimed = {},
    -- OnStartReclaim = {},
    -- OnStopReclaim = {},
    -- OnStopBeingBuilt = {},
    -- OnCaptured = {},
    -- OnCapturedNewUnit = {},
    -- OnDamaged = {},
    -- OnStartCapture = {},
    -- OnStopCapture = {},
    -- OnFailedCapture = {},
    -- OnStartBeingCaptured = {},
    -- OnStopBeingCaptured = {},
    -- OnFailedBeingCaptured = {},
    -- OnFailedToBuild = {},
    -- OnVeteran = {},
    -- OnGiven = {},
    -- ProjectileDamaged = {},
    -- SpecialToggleEnableFunction = false,
    -- SpecialToggleDisableFunction = false,
    -- OnAttachedToTransport = {}, -- Returns self, transport, bone
    -- OnDetachedFromTransport = {}, -- Returns self, transport, bone
}
```

Late game we can easily have 5000 units. Using the numbers of Kionx, just this table on its own will occupy 5000 * 23 x 20 bytes = 2.3 megabytes of memory. I'm not sure if the table construct itself also reserves some memory. Could you tell us more about that @KionX ?

KionX:
5000 round up to power of 2 is 8192 * 23 x 20 = 3M

Garanas:
5000 is the unit count, all units would have their own instance of the callbacks table. That is what I was trying to compute. 

> table base size ~36 bytes(i didn't measure it exactly)
Even when it is empty?

KionX:
`Even when it is empty?`
Yes.

Garanas:
Does that mean that if we have 5000 units with each their own unique instance, that we end up with:
`5000 * 23 x 20 bytes + 5000 * 21 x 36 bytes = 6.08 megabytes`, if we'd count both the fields in `EventCallbacks` and the empt tables that it is referring too?

KionX:
Here are my calculations:
((EventCallbacks + 22 nested tables up to 32) * 20 + (EventCallbacks + 22 nested) * 36) * 5000
5000 * ((22^(32) * 20) + 23 * 36) = 5000 * (640 + 828) = 7340000 / 1024 / 1024 = 6Mebibytes(2^20) or 7Megabytes(10^6)

Garanas:
Test code for blueprint duplication:
```
local ok, msg = pcall(
    function()

        if true then 
            return 
        end

        local lookUp = { }

        -- keep the duplicate values in a table that has a weak reference to the value, e.g. if this
        -- is the only reference then they can get dropped.
        local duplicates = { }
        setmetatable(duplicates, { __mode = "v" })

        -- construct a lookup table, keep track of the duplicates
        for k, other in __blueprints do 
            if type(k) == 'string' then 
                table.insert(duplicates, other)
            else
                lookUp[other.BlueprintId] = k
            end
        end

        -- do a shallow copy from the one used by entity:GetBlueprint() to the more user friendly version
        for hash, id in lookUp do 
            if __blueprints[hash].BlueprintId == __blueprints[id].BlueprintId then 
                __blueprints[hash] = __blueprints[id]
            else 
                LOG(string.format("Error: %s doesn't match %i", hash, id))
            end
        end

        -- sanity check
        LOG(table.getn(duplicates))         -- one less because blueprints start at [0]
        LOG(table.getsize(__blueprints))
        LOG(table.getsize(lookUp))

        -- check if they get dropped, this should happen in a second or two. But they never do :sad:
        self.Trash:Add(ForkThread(
            function()
                while true do 
                    WaitSeconds(1.0)
                    LOG(table.getn(duplicates))
                end
            end
        ))
    end
)

reprol(msg)
```

Garanas:
After:
 - Running various AI vs AI games
 - Carefully looking at the changes, make sure I didn't miss anything
 - Look at co-op maps to ensure I didn't break a common access pattern

I think the pull request is ready - there will be more in the future, but this one is ready for now 😄

PR #3876, opened 2022-05-19
Adjust PDs

Tagada14:
Adjust the Tracking Radius, Turn Speed, and Target priorities for all PDs.
This PR is aimed at fixing the PD's not firing because they were aiming at targets just outside their range, the PD constantly changing its target thus lowering the DPS. 
his should also improve the performance a bit since we got rid of some target priorities.


Garanas:
Should this be on `fafdevelop` or on `fafbeta`?

Tagada14:
> Should this be on `fafdevelop` or on `fafbeta`?

Normally I would put this on FAF beta but this aims to fix a somewhat crucial bug and with it aiming at FAF Develop it has a higher chance of getting merged sooner.

units/UAB2101/UAB2101_unit.bp
```diff
@@ -223,7 +223,7 @@ UnitBlueprint {
                 '(ALLUNITS - SPECIALLOWPRI)',
             },
             TargetRestrictDisallow = 'UNTARGETABLE',
-            TrackingRadius = 1.15,
+            TrackingRadius = 1.05,
```
Garanas, line 226: The attack radius of tech 1 artillery is too close to the tracking radius of a tech 1 point defense. 

As an example, the Medusa has an attack radius of 30. The point defense has an attack radius of 26. 1,05 * 2,730. It wouldn't be unnatural for the Medusa (or any other tech 1 light artillery) to end up in that radius given that they slow down upon arrival, may need to move for other units or get bumped into the tracking radius. 

Therefore we either:
 - Reduce the tracking radius to 1.0
 - Remove the light artillery from its attack priorities

Tagada14: Done

Penguin5:
Doubling the yaw speed sounds like a significant buff to PD's, given that they usually seem to spend a significant portion of their engagement time turning (to target different units) in my experience.  I'm not opposed to such a change, but that seems more like a balance change to me.  I understand the logic of wanting to increase the yaw speed some to allow for quick rotation to high threat targets that it no longer tracks outside of weapon range, but I imagine a yaw increase of this magnitude is actually of significantly greater impact than the tracking radius change is.

Garanas:
This was extensively discussed with the balance team - the changes were made by Tagada. It is on FAF Develop because we're set to release at an earlier date. It fixes the issue as described in this forum topic:

- https://forum.faforever.com/topic/3805/cybran-t2pd-having-problems-targeting-and-not-firing-at-targets-in-range/2

PR #3879, opened 2022-05-22
Fix lobby edit?

4z0t:
Issue:
[game_17104954.log](https://github.com/FAForever/fa/files/8749744/game_17104954.log)


Garanas:
I'm not sure if this fixes it, but it won't break anything either

PR #3724, opened 2022-03-26
Optimize all projectile files

Garanas:
Description
-------------
Projectiles are not the most friendly entities on resources. They are the definition of garbage: they are created to be destroyed again. With this pull request we make sure to minimize the impact of projectiles. Some projectiles need almost no work as they completely depend on the projectile hierarchy. Others do a lot of work and are good targets for small performance gains.

This pull request is similar to https://github.com/FAForever/fa/pull/3444 , which has been integrated partially already.

Course of action
------------------
Claim a few projectiles by making a comment in this pull request. This prevents people from doing the same projectiles. There are a few important factors to take into account:

### LOD of the mesh

Some projectiles have a mesh defined but no LOD threshold set. The **L**evel **o**f **D**etail threshold is used to cull entities when they contribute to no more than a few pixels or when the strategic icon (the dots of projectiles) renders over the entire projectile. The value is called `LODCutoff` in the blueprint. When the LOD is not set the projectile is always rendered, regardless of the distance to the camera. As an example, take the blueprint of `AAALightDisplacementAutoCannonMissile01`:

```lua
    Display = {
        Mesh = {
            LODs = {
                {
                    ShaderName = 'TMeshAlpha',
                    -- if AlbedoName or MeshName are not defined then they use the one with the same name, in the same folder (of the base game).
                },
            },
        },
        StrategicIconSize = 1,
        UniformScale = 0.025,
    },
```

Here we'd want to add an entry that fits with the size of the mesh of the projectile. A Mantis has a LOD cutoff of 215. There are almost no projectiles larger than a Mantis. Therefore the average projectile should not exceed 215. The actual value depends on the projectile and whether it is visible once the strategic view is enabled (when zooming out). 

### Optimize scripts

A lot of projectiles have empty scripts. These are easy enough to tackle: not much to be done. Some projectiles have a lot of logic inside them. Primary candidates are projectiles that can split up into several projectiles. An example of this type of work is
https://github.com/FAForever/fa/pull/3718 where we tackled the projectiles of the Seraphim t1 mobile artillery.

In general it is important to:
 - Prevent table allocations (including the Vector calls)
 - Reuse tables as much as possible
 - Upvalue engine functions (do **not** upvalue Lua functions, as that may break mods)
 - Replace logic with mathematics when possible 
 - Remove logic if it has no use
 
And anything else that we discussed in the development series on Youtube. The following videos of the series are good to watch if you have no idea where to start:
[...]

PR #3880, opened 2022-05-22
Optimize projectiles 01

Ejsstiil:
https://github.com/FAForever/fa/pull/3724

Optimize following projectiles. Unused code, added LODCutoff where applicable.
TIFMortar01
TIFNapalmCarpetBomb01
TIFNapalmCarpetBomb02
TIFSmallYieldNuclearBomb01
TIMAntiMissile01
TIMAntiTorpedo01
TIMAntiTorpedo02
TIMMissileIntercerptor01
TIMSonarBuoyCreator01
TorpedoShipAeon01
TorpedoShipTerran01
TorpedoShipTerran02
TorpedoShipTerran03
TorpedoSubTerran01

### DIFF
--- a/projectiles/TIFMortar01/TIFMortar01_script.lua
+++ b/projectiles/TIFMortar01/TIFMortar01_script.lua
@@ -1,4 +1,8 @@
 -- Terran Mortar
 
-TIFMortar01 = Class(import('/lua/terranprojectiles.lua').TArtilleryProjectile) {}
+local TArtilleryProjectile = import('/lua/terranprojectiles.lua').TArtilleryProjectile
+
+---@class TIFMortar01 : TArtilleryProjectile
+TIFMortar01 = Class(TArtilleryProjectile) {}
+
 TypeClass = TIFMortar01
--- a/projectiles/TIFMortar01/TIFMortar01_script.lua
+++ b/projectiles/TIFMortar01/TIFMortar01_script.lua
@@ -1,9 +1,4 @@
---
 -- Terran Mortar
---
-local TArtilleryProjectile = import('/lua/terranprojectiles.lua').TArtilleryProjectile
-TIFMortar01 = Class(TArtilleryProjectile) {
-}
 
+TIFMortar01 = Class(import('/lua/terranprojectiles.lua').TArtilleryProjectile) {}
 TypeClass = TIFMortar01
-
--- a/projectiles/TIFNapalmCarpetBomb01/TIFNapalmCarpetBomb01_script.lua
+++ b/projectiles/TIFNapalmCarpetBomb01/TIFNapalmCarpetBomb01_script.lua
@@ -2,7 +2,7 @@
 
 local TNapalmCarpetBombProjectile = import('/lua/terranprojectiles.lua').TNapalmCarpetBombProjectile
 
--- Used by uea0103
+--- Used by uea0103
 ---@class TIFNapalmCarpetBomb01 : TNapalmCarpetBombProjectile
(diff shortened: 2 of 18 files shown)


projectiles/TIFMortar01/TIFMortar01_script.lua
```diff
@@ -1,9 +1,4 @@
---
 -- Terran Mortar
---
-local TArtilleryProjectile = import('/lua/terranprojectiles.lua').TArtilleryProjectile
-TIFMortar01 = Class(TArtilleryProjectile) {
-}
 
+TIFMortar01 = Class(import('/lua/terranprojectiles.lua').TArtilleryProjectile) {}
```
speed2CZ, line 3: putting the import into the Class call makes it less readable

projectiles/TIFNapalmCarpetBomb01/TIFNapalmCarpetBomb01_script.lua
```diff
@@ -1,9 +1,7 @@
...
 -- Terran Napalm Carpet Bomb
---
-
-local TNapalmCarpetBombProjectile = import('/lua/terranprojectiles.lua').TNapalmCarpetBombProjectile
-
-TIFNapalmCarpetBomb01 = Class(TNapalmCarpetBombProjectile) { }
 
+--unused???
+--[[
+TIFNapalmCarpetBomb01 = Class(import('/lua/terranprojectiles.lua').TNapalmCarpetBombProjectile) { }
```
speed2CZ, line 5: they can be used by mods, don't comment them out

speed2CZ (changes requested):

projectiles/TIFNapalmCarpetBomb02/TIFNapalmCarpetBomb02_script.lua
```diff
@@ -8,10 +8,6 @@
 --  Copyright © 2007 Gas Powered Games, Inc.  All rights reserved.
 -------------------------------------------------------------------------------
 
-local TNapalmHvyCarpetBombProjectile = import('/lua/terranprojectiles.lua').TNapalmHvyCarpetBombProjectile
```
Garanas, line 11: I'll update the issue / pull request.

All locals that are removed should be placed beneath the class. That prevents them from being part of a closure (yah) while maintaining mod support.

Mod files are 'appended' to the bottom of each file. Therefore they can use the original imports.

projectiles/TIFNapalmCarpetBomb01/TIFNapalmCarpetBomb01_script.lua
Garanas: I agree, the base game doesn't use all projectiles. But a mod may use it.

projectiles/TIFNapalmCarpetBomb02/TIFNapalmCarpetBomb02_script.lua
Ejsstiil: Why then remove them?

Garanas: Good point 😄

Garanas:
Made a decision, see also:

image

At no point should we remove top-file locals, we can move them to the bottom of the file however. At any point all the top-file definitions (outer scope) should remain available, as mods may depend on their existence.

Garanas (approved):
Looking good - fixed some of the comment / annotation style.

PR #3885, opened 2022-05-24
Hunter AI - fix args for AggressiveMoveToLocation

speed2CZ:
AggressiveMoveToLocation takes second optional arg for the squad to move, but GetHighestThreatPosition was feeding it threat value


Garanas:
This is a different file than `platoon.lua` that has a whole bunch of behaviors. Is this file solely used for scenario's, such as campaigns?

speed2CZ:
This hunter function most likely is campaign specific but others from this file might not be.

PR #3887, opened 2022-05-24
Add settings and annotations for vs code extension

### LINKED ISSUE #3858: Interest in a fork of the Lua extension for VS Code?
Hi, I've been watching some of the coding videos on youtube and I noticed that the Lua editing experience isn't that great. I also saw in the deployment setup that using the Lua extension is not recommended. Text here

> Note that we do not recommend the Lua language server. The game uses a slightly adjusted version of Lua. The syntax doesn't match. We also have a different import system. The Lua language server generates dozens of errors. And it is not equipped to work with our import system.

Would there be interest if I were to take an existing extension (with permission) and fork it for use for working on FA?

For reference I've been working on an extension for modding another game, you can find the repo [here](https://github.com/hahn-kev/ftl-xml-extension) if you would like. I'd love to help support the devs by working on an extension for them to help with errors and code quality etc.

hahn-kev:
This branch includes some changes to add settings required by the vs code extension and some examples of annotations, more still to be done.

Closes #3858


Garanas:
This is exciting. The repository of the extension can be found here:
 - https://github.com/hahn-kev/vscode-lua/releases/tag/v0.9

Garanas:
@hahn-kev If I understood our conversation correct, it would also support this new type of layout formatting that @4z0t is working on: https://github.com/FAForever/fa/pull/3835/files , as it is based on a metatable?

With regard to intellisense / introducing types?

hahn-kev:
for the failing test, we should probably just exclude that file, as it's not a file used by FA, it's a file used by the lua extension which is a totally different version of Lua.

Garanas:
I'll add it as an exception.

hahn-kev:
> We can use regular expressions to quickly create some basic annotation. As an example, with the pattern `(\w*) = Class\((\w*)\)` and the replacement:
> 
> ```
> ---@class $1 : $2
> $0
> ```
> 
> We can quickly create the hierarchy chain of a single file. See [635c2f0](https://github.com/FAForever/fa/pull/3887/commits/635c2f049becec9180a7a5141fb095155d595bec). It is not perfect, but it may save us quite a bit of work.

That looks like it works great! if only we could do that for the functions, but I suspect that might be a fair bit harder.

hahn-kev:
I also added a few more examples, in this case I declared `moho` as a global, and then in a few files I setup some of the methods, though I'm not clear on some cases like `AIBrain` it looks like theres another file `CAiBrain.lua` in the engine folder. Are those the methods defined on `moho.aibrain_methods`? If so then we should change how the class is setup now to be more like the Unit example.

Ejsstiil:
> We can use regular expressions to quickly create some basic annotation. As an example, with the pattern `(\w*) = Class\((\w*)\)` and the replacement:
> 
> ```
> ---@class $1 : $2
> $0
> ```
> 
> We can quickly create the hierarchy chain of a single file. See [635c2f0](https://github.com/FAForever/fa/pull/3887/commits/635c2f049becec9180a7a5141fb095155d595bec). It is not perfect, but it may save us quite a bit of work.

Really cool! Works for me too!

hahn-kev:
> We can use regular expressions to quickly create some basic annotation. As an example, with the pattern `(\w*) = Class\((\w*)\)` and the replacement:
> 
> ```
> ---@class $1 : $2
> $0
> ```
> 
> We can quickly create the hierarchy chain of a single file. See [635c2f0](https://github.com/FAForever/fa/pull/3887/commits/635c2f049becec9180a7a5141fb095155d595bec). It is not perfect, but it may save us quite a bit of work.

here's a new version, will not match if the previous line includes `@` which should prevent duplicate class definitions
`^([^@\n]*\n)((\w*) = Class\((\w*)\))` need to change the replace pattern too
```
$1---@class $3 : $4
$2
```

hahn-kev:
And here's one to annotate self on class functions, I doubt this will work correctly if there's more than 1 class in a file
`^((\w+) = Class[^]+\n[^@]+)(\n.+function\(self)`
replace:
```
$1
    ---@param self $2$3
```

hahn-kev:
Just added a bunch of annotations and minor tweaks for typing. The biggest change I did was introduce `moho.lua` in the engine folder to represent how the `moho` table is defined. I'm assuming this won't cause a problem in game and that everything in the engine folder is ignored by the game so this shouldn't cause any issues. 

@Garanas would you mind taking a look at how `moho.lua` is setup? Should now be able to use find references on it to see where and what is getting used. 

Also, is there somewhere that I can see all the categories in game? there's a bunch of errors about that and I'd love to clean it up.

Garanas:
> Also, is there somewhere that I can see all the categories in game? there's a bunch of errors about that and I'd love to clean it up.

What category exists depends on what categories are used in all the blueprint files combined. All the BlueprintIds (of units, at least) are categories too.

hahn-kev:
Alright, I've gone ahead and added some default categories. I made this list by running the game and iterating over all values in the `categories` table while the game was running and only outputting the names that were all uppercase (so some might have been missed idk). In doing this I noticed some weird stuff. Like for example there's both a `TECH3` and a `TECH_THREE` which are both used in lua and in blueprints, so that's probably an error.

If we'd like to do this another way I'm open to suggestions, we could also include all the default units in the game, though those don't seem to be used as often so maybe this is a decent solution?

Garanas:
Lets finalize this pull request and divide the annotation work over various pull requests. That makes it easier to review.

.vscode/settings.json
```diff
@@ -1,3 +1,33 @@
...
-}
\ No newline at end of file
+    "Lua.runtime.nonstandardSymbol": [
+        "continue",
+        "!="
+    ],
+    "Lua.completion.showWord": "Disable",
+    "Lua.runtime.special": {
+        "import": "require",
+        "doscript": "require"
```
Garanas, line 9: Are these default values all sane?

.vscode/settings.json
```diff
@@ -1,3 +1,33 @@
...
+    },
+    "Lua.runtime.version": "Lua 5.1",
```
Garanas, line 11: Shouldn't we use Lua 5.0? We don't fully support all the Lua 5.1 functionality, such as '#' as a replacement for table.getn

engine/Sim/Shield.lua
```diff
@@ -1,8 +1,5 @@
...
-
----
---
-function Shield:moho.shield_methods()
-end
-
+---@declare-global
+---@class moho.shield_methods
+local Shield = {}
+---TODO list shield methods
```
Garanas, line 4: There are no special shield methods - they are essentially an entity with a large (spherical) collision box.

.vscode/settings.json
hahn-kev: Sadly the extension doesn't support Lua 5.0, so that's not a setting. So I'll add a task to add it and then we can make adjustments as we like. For example the # support and some methods that were depreciated in 5.1

.vscode/settings.json
hahn-kev: Yeah, essentially it enables type and completion support for those methods.

hahn-kev:
Sounds good. Are we good to included all the changes in this PR now? Or do you want me to split some stuff out?

Garanas:
I'll go over it a few times, the majority of changes are related to the annotations. I assume you used the scripts that were posted in this topic to make the majority of these?

Garanas:
Shall I add the fa-plugin.lua in .vscode to the ignore file for the syntax checker?

PR #3889, opened 2022-05-24
Chat settings window fixes & small improvements

Ejsstiil:
1) Chat settings window is not displayed correctly, when more than 6 players is in game (or so).
With higher number of the players, bottom part of the window is hidden and you can't save changes.

Before:
image

After:
- you can also drag this window with mouse, which was not possible
- font size slider steps changed from 2 to 1, so you gain more precise font size

image

2) When you choose to "Show feed background" under the bare chat lines, there are an ugly paddings, which looks too "stripy": 
image

Before:
image
After:
image

3) Added a QOL button "Apply", so when you trying to find a proper Font size and Alpha, you can use "Apply" button so see your settings without need to always re-open Settings window:
Fig1:
image

### DIFF
--- a/lua/ui/game/chat.lua
+++ b/lua/ui/game/chat.lua
@@ -1459,9 +1459,16 @@ function CreateConfigWindow()
         index = index + 1
     end
 
+    local okBtn -- for reuse okBtn's.Onclick and for keeping buttons in logical order in the code flow (Apply + Reset | Ok + Cancel)
+    local applyBtn = UIUtil.CreateButtonStd(optionGroup, '/widgets02/small', '<LOC OPTIONS_0139>', 16)
+    LayoutHelpers.Below(applyBtn, optionGroup.options[index-1], 4)
+    LayoutHelpers.AtLeftIn(applyBtn, optionGroup)
+    applyBtn.OnClick = function(self) okBtn.OnClick(self, true) end -- re-use okBtn's click, but set noCloseDlg param
+
     local resetBtn = UIUtil.CreateButtonStd(optionGroup, '/widgets02/small', '<LOC _Reset>', 16)
     LayoutHelpers.Below(resetBtn, optionGroup.options[index-1], 4)
-    LayoutHelpers.AtHorizontalCenterIn(resetBtn, optionGroup)
+    LayoutHelpers.AtRightIn(resetBtn, optionGroup)
+    LayoutHelpers.ResetLeft(resetBtn)
     resetBtn.OnClick = function(self)
         for option, value in defOptions do
             for i, control in optionGroup.options do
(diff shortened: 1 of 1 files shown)


lua/ui/game/chat.lua
```diff
@@ -1483,15 +1490,17 @@ function CreateConfigWindow()
...
     LayoutHelpers.Below(okBtn, resetBtn, 4)
     LayoutHelpers.AtLeftIn(okBtn, optionGroup)
-    okBtn.OnClick = function(self)
+    okBtn.OnClick = function(self, noCloseDlg)
         ChatOptions = table.merged(ChatOptions, tempOptions)
         Prefs.SetToCurrentProfile("chatoptions", ChatOptions)
         GUI.bg:OnOptionsSet()
-        GUI.config:Destroy()
-        GUI.config = false
+        if (noCloseDlg or true) ~= true then -- preserve normal ok's behavior
```
Garanas, line 1500: I think `not noCloseDlg` achives the same result: if the value is not set, then we close the dialogue. Makes it easier to read.

Garanas, line 1500: I think `not noCloseDlg` achives the same result: if the value is not set, then we close the dialogue. Makes it easier to read.

Garanas:
I'm in favor of these changes - @4z0t could you give your opinion / review too?

Garanas:
I'm in favor of these changes - @4z0t could you give your opinion / review too?

lua/ui/game/chat.lua
Ejsstiil: `not noCloseDlg` would be dealing with uninitialized value, the `(noCloseDlg or true)` is the param defaulting to true

Garanas: But then you check with `~= true` - comparing boolean values like this makes code more confusing. 

Looking at it again: `(noCloseDlg or true)` after this noCloseDlg is always true, unless it is set to some non boolean value.

Garanas:
When opening it (on a new profile) it ends up at the very top:

image

Garanas:
Besides that - I like the changes. The apply button is a nice touch, once the starting position is fixed then I'll happily merge this.

Ejsstiil:
Strange - can replicate with the new profile. Any clue what is special about empty (new) profile?

Garanas:
> Strange - can replicate with the new profile. Any clue what is special about empty (new) profile?

It has no previous location. Therefore it defaults to something, and that default looks like this.

Ejsstiil:
> > Strange - can replicate with the new profile. Any clue what is special about empty (new) profile?
> 
> It has no previous location. Therefore it defaults to something, and that default looks like this.

Surprisingly not the case, this was checked at start, saved position is not even used.

lua/ui/game/chat.lua
```diff
@@ -1258,10 +1258,12 @@ function CreateConfigWindow()
         br = UIUtil.SkinnableFile('/game/panel/panel_brd_lr.dds'),
         borderColor = 'ff415055',
     }
-    GUI.config = Window(GetFrame(0), '<LOC chat_0008>Chat Options', nil, nil, nil, true, false, 'chat_config', nil, windowTextures)
+
+    local defPosition = Prefs.GetFromCurrentProfile('chat_config') or nil
+    GUI.config = Window(GetFrame(0), '<LOC chat_0008>Chat Options', nil, nil, nil, true, true, 'chat_config', defPosition, windowTextures)
```
Ejsstiil, line 1263: The issue was the 7th parameter - fixed position. In previous commit was set to false, as not fixed.
Seems that Window class has trouble to set drag handler

lua/ui/game/chat.lua
```diff
@@ -1515,6 +1517,13 @@ function CreateConfigWindow()
 
 
     GUI.config.Bottom:Set(function() return okBtn.Bottom() + 5 end)
+    if defPosition ~= nil then
+        GUI.config.Top:Set(defPosition.top)
+        GUI.config.Left:Set(defPosition.left)
+    else
+        GUI.config.Top:Set(function() return 90 end)
+    end
+    GUI.config:SetPositionLock(false) -- allow window to be draggable, didn't worked in Window() call
```
Ejsstiil, line 1526: This call assures the created window is draggable

4z0t:
See how I did it here https://github.com/4z0t/FAF-UI-Mods/blob/9c15fa1e045d1735ee560018254f12f6e15d760f/mods/better%20chat/modules/BCwindow.lua#L55

Garanas:
Tried and tested - it all appears to be good!

PR #3891, opened 2022-05-25
Sanitize unit blueprints

Garanas:
Adjusts / sanitizes various unit blueprint values in favor of performance.

### Guard scan radius

The guard scan radius is used when:
- A unit attack moves, it determines how far the unit remains from its target
- A unit patrols, it determines when the unit decides to engage its target

It could cause significant performance issues when used improperly, most noticeable during battles. Some extreme examples:

image
400 Cerberus vs 400 other Cerberus, outside of weapon tracking range - if they are in guard scan radius then they tank the sim

Unusually high values set to ASF and Mantis to show case the issue. It either results in stutters (asf example) or in a slow simulation (mantis example).

Note that the default value for ASF is set to 100, which makes the ASF example not as extreme as it appears as the entire enemy cloud of ASFs is easily inside that range of your cloud of ASFs, tanking the simulation as a result.

### Staging Platform Scan Radius

This value only makes sense when the unit in question is capable of staging air planes. It is set to 0 for any other unit.


Garanas:
A list of all changes as of commit [4616f41](https://github.com/FAForever/fa/pull/3891/commits/4616f411290241cfe65cbf091989ce6ad1392720)
- https://pastebin.com/ETpT6ZqQ

Garanas:
Unironically, we're not the first ones that did this. See also the changes in commit [6658271](https://github.com/FAForever/fa/pull/3891/commits/665827154a15f6f8aa41da4dacd00277e7801631). They didn't take into account structures, and some values still got set to unhealthy high values.

A list of all changes as of commit [c8327ce](https://github.com/FAForever/fa/pull/3891/commits/c8327ce70a9a82071f30451ede50988025e8871e)

- https://pastebin.com/i4QVjM9p

The value set in the blueprint itself is leading, and not overwritten. Any other value is set according to the logic as defined in [this file](https://github.com/FAForever/fa/blob/c8327ce70a9a82071f30451ede50988025e8871e/lua/system/blueprints-units.lua).

PR #3892, opened 2022-05-25
Sanitize values of air units

Garanas:
Adjusts the properties of air units to match the limitations of the engine and the aesthetics of the game. At the moment (commit fa380c03e0553dd8485bacb31ea458bedbbb250d) the branch is a play ground to test different values.

### Changes
All ASF have their values sanitized, this includes:
- Guard scan radius set to 18 (as per #3891)
- No target priorities except for `ALLUNITS`
- No target rechecking
- Improved allow / disallow list of valid targets

The footprint of ASF is adjusted depending on the faction. In practice this would be the same, but this makes easier experimentation.
- **A**eon ASF: default footprint
- **U**EF ASF: two times the regular footprint
- **C**ybran ASF: three times the regular footprint
- **S**eraphim ASF: four times the regular footprint

image

Note that the default footprint ends up with a lot of intersections of meshes:
image

With twice the space it is better, but still crowded:
image

With three times the space is more like it:
image

With four times the space it gets a bit much:
image

The more spacing we have, the less likely air fights become extremely dense and the better performance we'll have throughout the fight.

### Todo
- Some ASF have two weapons instead of one. That should be reduced to one weapon with two projectiles, similar to (for example) the Cybran Rocket bot.
- Apply the same spacing to all other aerial units


MrRowey:
No target priorities except for `ALLUNITS`

Im asuming the end goal will be to have some target prioritys ? EG. as your going to want to be able to shoot a strat over a anything else

Garanas:
Yes - they'll have more priorities in the future. We'll use this as a starting ground and improve accordingly.

Garanas:
Oh, fun fact: see the file changes. No superiority fighter preferred bombers over other air units: **except** the Seraphim ASF. All other ASFs only discouraged targeting intelligence related units, such as scouts 😄 .

speed2CZ:
Do the units stay further from each other even when flying? Or is it only when landed?

Garanas:
They're also more spaced out when they fly around.

Garanas:
Adapts the footprint size of air units according to:

- Bombers: size of 4
- Gunships: size of 3
- Other air units: size of 2

The footprint size used to be 1 for all air units. Experimental units are excluded.

Ejsstiil:
Don't like the idea of ignoring the definitions (Footprint) in *.bp's via code. Instead we should 'fix' corresponding *.bp in every intended unit.
This 'centralization' is no good

Garanas:
I disagree - some values are too important to not sanitize. 

We started a blueprint standardization process a while ago with formatting all the blueprint files using a tool made by @The-Balthazar . The same tool can be used to edit values in the blueprint files en-masse, similar to what we do here during blueprint loading. Ideally we start removing logic during blueprint loading and just process the blueprint files directly using that tool. But I lack the time to investigate it properly

Ejsstiil:
Formatting is a good thing, yes, even explicitly pull some default values back to blueprints is a good thing too.
But rendering some properties in bp useless by overriding them in the code is a bad idea, disagree here with you. Why they even do exists then?
Values in the blueprints should be the only one source of the truth.

Ejsstiil:
PS, not disputing the intention of normalizing the Footprints, but please let's do it in BP's

PR #3893, opened 2022-05-25
Fix tactical / strategical missile defense

Garanas:
Defines the `DesiredShooterCap` value for tactical missiles so that they function similar to how Anti Nukes work: one missile per target, unless more are required.

### Changes

Before looking at the gifs, Buzzkills and the Seraphim cruiser have the same rate of fire. Therefore two buzz kills should be able to neutralize two Seraphim cruisers.

Old situation where two Buzzkills are easily overwhelmed by two Seraphim cruisers:
here-is-gif-41

New situation where two Buzzkills pick their targets accordingly and manage to survive indefinitely:
here-is-gif-42

The issue was that the `DesiredShooterCap` value of the blueprint of the tactical missiles are not set. That defaults to 3. This pull requests changes it to the amount of health that a missile has, considering that all tactical missiles do 1 damage. 

With the old situation I'd need to have at least four buzz kills to get a reliable counter to the cruisers. With new old situation I'd need only two - as you'd expect when you look at the specs of the units in question.

### Rationale

Tactical missile defenses are unreliable and frustrating. Especially in large numbers (many tactical missiles, many tmd) they tend to make such extreme poor choices that they get overwhelmed. When all tactical missiles come in at once - even though you should have sufficient tmd (looking at rate of fire / damage and health of missiles) to properly neutralize the attack.

This is a broken feature - one that we can fix 😄  .


Garanas:
Some videos of ideas found [on the forums](https://forum.faforever.com/topic/4405/game-version-3741-sneak-peaks/27):

https://user-images.githubusercontent.com/15778155/179260773-7ebf6528-867e-4391-8483-5f8e7c894e99.mp4


https://user-images.githubusercontent.com/15778155/179260778-5f364ce3-126b-4d38-8839-a9a24c8abca5.mp4

PR #3899, opened 2022-05-26
Resourcebar visibility fix

Ejsstiil:
It is working now. Tested both on fresh profile and the normal one. Sorry ;)

Fixed red underlying warning texture which stayed visible after UI_ToggleGamePanels.

Before:
https://user-images.githubusercontent.com/36369441/170463095-46859726-4352-495b-8a55-057431ec18cb.mp4

After:
https://user-images.githubusercontent.com/36369441/170462573-32a86846-97b6-4e56-86c6-d23246def421.mp4

### DIFF
--- a/lua/ui/game/economy.lua
+++ b/lua/ui/game/economy.lua
@@ -168,6 +168,7 @@ function CommonLogic()
             group.expense:SetHidden(hidden)
             group.reclaimDelta:SetHidden(hidden)
             group.reclaimTotal:SetHidden(hidden)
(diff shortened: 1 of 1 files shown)


Garanas:
image

There's a few up top there too, but that is for another pull request 😄

Ejsstiil:
> There's a few up top there too, but that is for another pull request 😄
Yea! noticed it too yesterday. But its weird - it does that just for few initial 1-2 attempts. After that, its okay

PR #3905, opened 2022-05-29
Fixes script to allow for linux users

Spikey84:
This fixes `start_dev.sh` so that linux users can setup a development environment. It does add a few extra steps compared to a windows setup. First, you have to change `RunProton` to the location of the script on your machine (https://wiki.faforever.com/en/FAQ/Client-Setup this has a step in which u create it). Second, you will either have to remove the `/nomovie ` argument (incredibly laggy) or you have to hit escape once the game opens. Also the `/enablediskwatch` is needed to accomplish the in game code changes shown in @Garanas's video.


Garanas:
Great work @Spikey84 !

Garanas:
Great work @Spikey84 !

PR #3909, opened 2022-05-30
Hud Notifications leaks to the right view when split

Ejsstiil:
Notifications about ACU upgrades are not staying inside its own view - se video before.
This PR fixing that.

Tested:
- windowed mode
- fullscreen mode single monitor
- fullscreen on dual monitors

Before:

https://user-images.githubusercontent.com/36369441/170982064-68c5e0fe-ce45-4f87-b817-934853d744a1.mp4

After:

https://user-images.githubusercontent.com/36369441/170982095-0ba7b681-917e-49d9-834f-d6101ec0d141.mp4

### DIFF
--- a/lua/ui/notify/notifyoverlay.lua
+++ b/lua/ui/notify/notifyoverlay.lua
@@ -127,6 +127,12 @@ function createEnhancementOverlay(args)
         local worldView = import('/lua/ui/game/worldview.lua').viewLeft
         local pos = worldView:Project(overlay.pos)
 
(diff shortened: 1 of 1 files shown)


speed2CZ:
Wouldn't setting the correct parent or depth help here? Rather then calcualting the position like this.

Ejsstiil:
> Wouldn't setting the correct parent or depth help here? Rather then calcualting the position like this.

It's not computational, it's just a comparison here. But proper parent might help here, dunno

Garanas:
Are you sure you introduced all the required files? It has not changed after checking out on this branch.

PR #3910, opened 2022-05-30
Refactor BuffAffectsUnit to be easier to mod

The-Balthazar:
This refators BuffAffectsUnit to have the actual effects separated out into separate functions in a hash table that can be easilly added to with mods, whilst also stripping out the long inefficient `if elseif` chain.

This also adds one off warnings when buff types are missing.

More could be done to the individual buff functions, such as having them read the unit.bp field instead of calling `:GetBlueprint()`, but I wanted to get the basic infrastructure in rather than actually rewrite every function. (Also I think the Health one is actually bugged? But no one changes health that way.)

This rework also has the added benefit of still being theoretically compatible with older mods that hooked this function previously, but future mods could have an easier time of it.


Garanas:
Something odd happened 😄

Garanas:
I had a local branch named `patch-2` too

PR #3911, opened 2022-05-31
Massfabs panel fixes

4z0t:


Garanas:
What does this fix?

4z0t:
@Garanas It fixes UI being shown when all UI is hidden and position scale

PR #3912, opened 2022-05-31
Add remaining game files

Garanas:
Adds in all the remaining blueprint (`.bp`) and Lua (`.lua`) files from the following folders of the base game:

- effects.scd
- env.scd
- lua.scd
- meshes.scd
- mohodata.scd
- moholua.scd
- objects.scd
- props.scd
- units.scd

By doing so, we are independent of the following base game files:

- mohodata.scd
- moholua.scd
- lua.scd

These are no longer loaded in by the init files, as done by [1f7a8f5](https://github.com/FAForever/fa/pull/3912/commits/1f7a8f5ec6ac8d54969e61402293490f2a8481b3). We've not included all the remaining projectiles as that is done by https://github.com/FAForever/fa/pull/3724 .

Also guarantees that:

- instead of `[[` or `]]` `'` is used to construct strings
- all comments using `#` are replaced with `--`


Garanas:
Closely related to #3887 to improve support for the to-be intellisense extension.

PR #3926, opened 2022-06-04
Main menu ambient movie

Ejsstiil:
This PR resolves some missing features and finalizes already merged PR #3907.
It also adjusts the different behavior for **/nomovie** in conjuction of the Menu option "Main menu Background Movie"(Fig1):
- Game Options had already a dedicated option for BG movie, and now these two elements (Ambient and Movie) works together. It is no longer needed to use **/nomovie** in order to silence Video & Ambient under the Game Options
- this separation no longer affects any other in-game video material, which is the natural behavior for mentioned switch. Much better, when the only desired effect is just to "silence background video" in the Main Menu
- Game Options has got another "Command Line Override" for that (Fig1) with config-value preservation.
- it also removes the need of double-click, or pres Esc twice, when **/nomovie** has been used on commandline.

Fig1:
image


lua/ui/splash/splash.lua
```diff
@@ -3,7 +3,7 @@
 --* Author: Chris Blackwell
 --* Summary: create and control pre-game splash screens
 --*
---* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
```
Garanas, line 6: This happens sometimes - fix the copyright symbol by copying the one you see on the 'removed' line on Github.

Ejsstiil: Saw that - this one was not mine, but from the merge i had to do

Garanas: Ah, that was me then 😄

Ejsstiil: You can remove(discard) my merge before merging to devel? [d5b6ada](https://github.com/FAForever/fa/pull/3926/commits/d5b6ada03015a7a782df998c38a93aa5916d303e)

Garanas: no - we squash it regardless. So those commits are 'gone' after that anyway.

PR #3927, opened 2022-06-04
Auto jamming Toggle

Spikey84:
This will attempt to automatically toggle radar jamming occasionally. https://forum.faforever.com/topic/3866/jamming-ability-should-reset-when-vision-of-unit-is-lost

Very work in progress rn


lua/aibrain.lua
```diff
@@ -332,6 +337,43 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                   jammer:EnableUnitIntel('ToggleBit3', 'Jammer')
+                   
+                   
+                else
+                    RemoveJammer(jammer)
+                end
+
+            end
+            WaitSeconds(10)
+            CoroutineYield(1)
```
hahn-kev, line 373: doesn't `WaitSeconds` just call `CoroutineYield` internally, so are both needed here?

Garanas: WaitSeconds is defined here: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/simInit.lua#L26, which calls coroutine.yield at some point

lua/aibrain.lua
```diff
@@ -607,11 +606,10 @@ AIBrain = Class(moho.aibrain_methods) {
     --  val: true or false
     -- calls callback function with blip it saw.
     OnIntelChange = function(self, blip, reconType, val)
-        if reconType == 'LOSNow' then
+        if reconType == 'LOSNow' or econType == 'Omni' then
```
Garanas, line 609: Typo!

lua/aibrain.lua
```diff
@@ -344,15 +344,14 @@ AIBrain = Class(moho.aibrain_methods) {
...
         while true do 
 
             for i, jammer in self.jammers do
-                if not jammer:BeenDestroyed() then
-                    if jammer.resetJammer then
-                        
-                        LOG("ToggleJammer")
-                        -- run logic here
 
+                if not jammer:BeenDestroyed() then
```
speed2CZ, line 352: woudln't `jammer.Dead` do the same trick?

Garanas: No - there's a small window between a unit being destroyed and a unit being set to dead where any engine interaction can cause a crash. This was most notable when we used `unit.Dead` or the build animations in an attempt to evade an engine call, where we now often use `BeenDestroyed` instead.

lua/aibrain.lua
```diff
@@ -328,6 +333,41 @@ AIBrain = Class(moho.aibrain_methods) {
         self.PreBuilt = true
     end,
 
+    -- Jamming Switch Logic
+
+    AddJammer = function(self, unit)
```
Garanas, line 338: Use `TrackJammer` instead of `AddJammer`

lua/aibrain.lua
```diff
@@ -328,6 +333,41 @@ AIBrain = Class(moho.aibrain_methods) {
...
+        self.jammers[unit.EntityId] = unit
+    end,
+
+    RemoveJammer = function(self, unit)
```
Garanas, line 342: Similar, use `UntrackJammer`

lua/aibrain.lua
```diff
@@ -237,6 +237,11 @@ AIBrain = Class(moho.aibrain_methods) {
         -- they are capitalized to match category names
         local layers = { "LAND", "AIR", "NAVAL" }
         local techs = { "TECH2", "TECH3" }
+    
+        self.jammers = { }
```
Garanas, line 241: Entries in `self` should always be pascal case.

lua/aibrain.lua
```diff
@@ -328,6 +333,41 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                    if jammer.resetJammer then
+                        jammer:DisableUnitIntel('AutoToggle', 'Jammer')
+                        WaitSeconds(0.3)
+                        if not jammer:BeenDestroyed() then
+                            jammer:EnableUnitIntel('AutoToggle', 'Jammer')  
+                            jammer.resetJammer = false
+                        end
+                    end
+                else
+                    RemoveJammer(jammer)
```
Garanas, line 363: Better to untrack `OnDestroy` similar to how we start tracking during `OnCreate` of the unit itself. The if statement to check if the entity is still there is quite valid regardless.

lua/aibrain.lua
```diff
@@ -328,6 +333,41 @@ AIBrain = Class(moho.aibrain_methods) {
...
+
+            for i, jammer in self.jammers do
+
+                if not jammer:BeenDestroyed() then 
+
+                    if jammer.resetJammer then
+                        jammer:DisableUnitIntel('AutoToggle', 'Jammer')
+                        WaitSeconds(0.3)
+                        if not jammer:BeenDestroyed() then
+                            jammer:EnableUnitIntel('AutoToggle', 'Jammer')
```
Garanas, line 358: As this thread needs to remain alive throughout the game, we'll need to make sure that any external call can not crash the thread. We can use `pcall` to help us do this, you can look at `ToggleEnergyExcessUnitsThread` for inspiration.

lua/aibrain.lua
```diff
@@ -566,6 +606,15 @@ AIBrain = Class(moho.aibrain_methods) {
     --  val: true or false
     -- calls callback function with blip it saw.
     OnIntelChange = function(self, blip, reconType, val)
+        if reconType == 'LOSNow' or econType == 'Omni' then
+            if not val then
+                local unit = blip:GetSource()
+                if unit.Blueprint.Intel.JammerBlips > 0 then
+                    unit.resetJammer = true
```
Garanas, line 613: Unit fields (fields in `self`) should always be pascal case.

lua/aibrain.lua
```diff
@@ -328,6 +333,41 @@ AIBrain = Class(moho.aibrain_methods) {
...
+
+        while true do 
+
+            for i, jammer in self.jammers do
+
+                if not jammer:BeenDestroyed() then 
+
+                    if jammer.resetJammer then
+                        jammer:DisableUnitIntel('AutoToggle', 'Jammer')
+                        WaitSeconds(0.3)
```
Garanas, line 356: It appears this delay is required or the change is not being picked up. I suspect that the unit intel is disabled but only picked up / updated a few ticks later. After that we can re-enable it again. @KionX Do you know if intel is managed at a different tick rate than the simulation?

Garanas (changes requested):
Exciting - looking promising 😄

lua/aibrain.lua
KionX: Don't know.

Spikey84: Yea i could not get intel to toggle any faster, it also creates a cool effect of many popping on one after another. I was thinking maybe we create an array from each army that way if army 1 has 100 frigs and army 2 has one sparky, army 2 wont have to wait for army 1's frigs in order to reload the sparky. Thoughts?

lua/aibrain.lua
Spikey84: fixed, thanks!

lua/aibrain.lua
Spikey84: lol ig thats why it was not working, thanks

lua/aibrain.lua
Spikey84: done.

lua/aibrain.lua
Spikey84: Good to know.

Garanas:
@Spikey84 Make sure to hit the resolve button when you resolve them 😄 

And for future suggestions - note that these are suggestions. We can always discuss them if you happen to disagree.

lua/aibrain.lua
Garanas: Each army has its own brain, therefore each brain triggers their jammers on / off individually.

lua/aibrain.lua
Spikey84: What would the difference be? Just consistency?

Garanas: Just consistency, yes - I still need to do this for some other trackers too.

lua/sim/Unit.lua
```diff
@@ -341,6 +341,13 @@ Unit = Class(moho.unit_methods) {
         end
         -- Flags for scripts
         self.IsCivilian = armies[self.Army] == "NEUTRAL_CIVILIAN" or nil 
+
+        self.ResetJammer = false
```
Garanas, line 345: This field does not need to be set by default - if the field doesn't exist then it defaults to `nil`, which is equal to `false` in the average if statement.

For your idea: a single hash entry takes up about 20 bytes. We can easily have 6000+ units in a game. That means we have over 120 kilobytes for a flag that at most 1% of the units may need at any given point.

lua/aibrain.lua
Spikey84: I am unsure how this should work, should it be `ok, msg = pcall(jammer:DisableUnitIntel, 'AutoToggle', 'Jammer')`? what does this do?

Spikey84:
@Garanas all tested and working! Ready to be put on faf dev unless u see any issues.

lua/sim/Unit.lua
```diff
@@ -1957,6 +1964,11 @@ Unit = Class(moho.unit_methods) {
         end
 
         ChangeState(self, self.DeadState)
+
+        if self.Blueprint.Intel.JammerBlips > 0 then
```
Garanas, line 1968: Changing the state may kill the thread it is running on, this should be run before `ChangeState` runs.

Spikey84: Fixed it!

lua/aibrain.lua
Hdt80bro: Almost!
You'd do `local ok, msg = pcall(jammer.DisableUnitIntel, jammer, 'AutoTggle', 'Jammer')` (and again for EnableUnitIntel), since the colon syntax needs the actual function call to insert the calling table into, instead of a reference. What this does is make sure that any errors don't break the rest of the AIBrain. If the `pcall` *does* return an error, you would report it like this:
```lua
if not ok then 
    WARN("JammingFollowUpThread: " .. repr(msg))
end
```

Hdt80bro (approved):

PR #3933, opened 2022-06-08
Use playable area for map size

Hdt80bro:
Change the map size in the score panel to reflect the current playable area instead of the raw map bounds. Some maps will use a base size but edit the playable area to something smaller (e.g. use a 20x20 base size and then limit it to a 15x15). This will update the map size indicator in the score panel to show that.

### DIFF
--- a/lua/ui/game/score.lua
+++ b/lua/ui/game/score.lua
@@ -239,6 +239,8 @@ ShareDescriptionLookup["TransferToKiller"] = "lobui_0763"
 ShareDescriptionLookup["Defectors"] = "lobui_0767"
 ShareDescriptionLookup["CivilianDeserter"] = "lobui_0765"
 
+local mapSizeUI = false
+
 function SetupPlayerLines()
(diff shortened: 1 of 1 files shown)


Garanas:
Looking good!

image

lua/ui/game/score.lua
```diff
@@ -559,6 +572,13 @@ function SetupPlayerLines()
     Tooltip.AddControlTooltip(resModeSwitch.icon, {text = '', body = bodyText}, 1)
 end
 
+function MapSizeText(width, height)
+    -- round half down; this won't let fractional map sizes (e.g. from the map gen) make it needlessly long
+    local widthKm = math.ceil(width / 51.2 - 0.5)
+    local heightKm = math.ceil(height / 51.2 - 0.5)
+    return tostring(widthKm) .. "x" .. tostring(heightKm)
```
speed2CZ, line 579: Lua string concatenation convers numbers automatically, so `tostring` call is not really needed here.

Issue #3762, opened 2022-04-05
Refactor `layouthelpers.lua`


Garanas:
Description
-----------

Usually when we talk about performance we are referring to the sim side of the equation as that is what matters most in practice. It would be a shame however to forget about the UI side, in particular if there is low hanging fruit. For this issue we limit ourselves to `/lua/maui/layouthelpers.lua`. 

Course of action
----------------

There are a lot of math operations used. We can scope them as an upvalue to prevent the `GETGLOBAL` and the `GETTABLE` calls. Add the locals at the top of the file and replace the calls accordingly.

Test plan
---------

This one is a bit difficult to test and therefore will be checked by reviewing the code changes. By all means, when starting the game there should be no errors related to these changes.

Learning goals
--------------

Feel more comfortable working with the repository and making changes that impact dozens of other files.

4z0t:
Basically `math.floor` must be replaced with local one  (upvalued function) and `:Set(...)` that sets function replaced with `:SetFunction(...)`

Hdt80bro:
https://github.com/FAForever/fa/pull/3936#issue-1265083181

PR #3936, opened 2022-06-08
Refactor layouthelpers

Hdt80bro:
Refactor layouthelpers, as proposed in [issue#3762](https://github.com/FAForever/fa/issues/3762#issue-1193077875). Mostly just the things requested there, but comprehensively:
- Standardized comments
- Upvalued math.floor
- Replaced Set with SetFunction where appropriate
- Removed double divide by 2 in centering functions
- Moved optional argument checking into parent functions
- Reused tablelayouts in RightRelativeTo and BottomRelativeTo functions (though if the cost of setting extra local variables is more expensive than the table lookup, let me know)
- Removed redundant multiplication in InvScaleNumber

There's a lot a extra parentheses in here--I just left them because it doesn't matter, but I'd be interested to know if we have a style guideline for that.

### DIFF
--- a/lua/maui/layouthelpers.lua
+++ b/lua/maui/layouthelpers.lua
@@ -17,12 +17,12 @@
 local MathFloor = math.floor
 local MathCeil = math.ceil
 
+local Prefs = import('/lua/user/prefs.lua')
 -- Store and set the current pixel scale multiplier. This will be used when the
 -- artwork is scaled up or down so that offsets scale up and down appropriately.
 -- Note that if you add a new layout helper function that uses offsets, you need
 -- to scale the offset with this factor or your layout may get funky when the
 -- art is changed
-local Prefs = import('/lua/user/prefs.lua')
 local pixelScaleFactor = Prefs.GetFromCurrentProfile('options').ui_scale or 1
 
 --- Set new scale factor for absolute offset in layout functions
@@ -32,18 +32,9 @@ function SetPixelScaleFactor(newFactor)
 end
 
 --- Get scale factor for absolute offset in layout functions
----@return number pixelScaleFactor 
+---@return number pixelScaleFactor
 function GetPixelScaleFactor()
     return pixelScaleFactor
-end
-
---- Set absolute dimensions of a control
----@param control Control
----@param width? number no change if nil
----@param height? number no change if nil
-function SetDimensions(control, width, height)
-    SetWidth(control, width)
-    SetHeight(control, height)
 end
 
 --- Set absolute width of a control
@@ -64,13 +55,26 @@ function SetHeight(control, height)
     end
 end
 
+--- Set absolute dimensions of a control
+---@param control Control
+---@param width? number no change if nil
+---@param height? number no change if nil
+function SetDimensions(control, width, height)
+    SetWidth(control, width)
+    SetHeight(control, height)
+end
+
+
 --- Set the depth of a control higher than a parent
 ---@param control Control
 ---@param parent Control
 ---@param depth? integer defaults to 1
 function DepthOverParent(control, parent, depth)
-    depth = depth or 1
-    control.Depth:SetFunction(function() return parent.Depth() + depth end)
+    if depth and depth ~= 1 then
+        control.Depth:SetFunction(function() return parent.Depth() + depth end)
+    else
+        control.Depth:SetFunction(function() return parent.Depth() + 1 end)
+    end
 end
 
 --- Set the depth of a control lower than a parent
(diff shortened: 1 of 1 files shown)


Hdt80bro:
Oh, I didn't check for errors. One moment

Garanas:
Looking good - I'll give some more informed feedback tomorrow.

Hdt80bro:
Alright, I've gone ahead and done a full reorganization, which I'm guessing isn't what you're looking for, but there was just too much repeated code to not delegate the functions. These function shouldn't be performance critical (it's mostly the property Getters they set on the controls that affect performance), so I'm guessing a few extra function calls for the sake of organization are fine.
I'm also realizing that all of these circular dependency errors in the log aren't my fault--looks like there's something wrong with the base Control's Reset method?

Garanas:
> Alright, I've gone ahead and done a full reorganization, which I'm guessing isn't what you're looking for, but there was just too much repeated code to not delegate the functions.

That is good - it was long over due 😄 . We're currently introducing intellisense support, see also #3887 . Could you annotate the functions so that the extension can read them and give people hints on how to use them? See also:

- https://github.com/sumneko/lua-language-server/wiki/EmmyLua-Annotations

> I'm also realizing that all of these circular dependency errors in the log aren't my fault--looks like there's something wrong with the base Control's Reset method?

By default the horizontal (left / right / width) and the vertical (top / bottom / height) reference one another so that you only need to define two of each triplet. It is horrible from a development perspective as it is the most uninformative error message we have at FAF. With that said, if there weren't errors before and there are now then the adapted logic changes the behavior, instead of just improving readability / performance.

Hdt80bro:
Yeah, seeing the errors is what made me think that I did something, but for some reason the errors persist when I disable my test mod. So... I don't know. Can you confirm that you don't have the errors on the current deploy/fafdevelop build? Something may be up with mine.

Garanas:
Can you describe when you have the errors, share a log?

Hdt80bro:
[game_17245910.log](https://github.com/FAForever/fa/files/8867832/game_17245910.log)
The only thing I've modified here is turning on lazyvar.ExtendedErrorMessages

Garanas:
Are you testing the changes via a UI mod? Have you seen the [instructions](https://github.com/FAForever/fa/blob/deploy/fafdevelop/setup/setup-english.md) on how to setup a development environment?

Hdt80bro:
I was having issues with a UI mod, so I was using a sim mod, which was working. Setting up a development environment sounds a lot easier though... I'll get on that.

Either way, I'm pretty sure the issue in somewhere over here since it's happening consistently on all unmodded game branches, even after redownloading the gamedata. I'll just reinstall FAF and setup things up properly this time.

Hdt80bro:
Uh, I reinstalled both FAF & FA and the log still isn't clean from an unmodded default FAF-branch game. So I'm not sure what broke in the last twelve hours (I'm pretty sure it wasn't doing that before), but at least it isn't noticeable in-game. I'll have to look into in the morning.

4z0t:
I will take a look tomorrow, also rename floor to MathFloor since its usual upvalued name

Hdt80bro:
Okay, I've gone ahead and done both of your suggestions. Everything looks to be laid out fine in-game, but let me know if your error log doesn't look normal.

Hdt80bro:
Do we know if we can have `RelativeTo` and `AtCenterIn` (and thus `Layouter:AtCenterIn`, `Layouter:AtTopCenterIn`, and `Layouter:AtBottomCenterIn`) get their vertical and horizontal arguments switched? They are the only ones that go `Function(vertical, horizontal)` and it looks like only a handful of function calls would need to be modified.

Hdt80bro:
And with that, I'm ready for review again

lua/maui/layouthelpers.lua
```diff
@@ -309,42 +309,43 @@ end
 
 -- These functions reset a control to be calculated from the others
 
---- Resets a control's left to be calculated from its right and width
---- *Make sure Right and Width are defined!!!*
+--- Resets a control's left to be calculated from its right and width  
+--- **Make sure `Right` and `Width` are defined!!!**
```
Garanas, line 320:
```md
**Make sure `Right` and `Width` are defined!!!**
```

It should be:

```md
Make sure that either `control.Right` or `control.Height` is defined too.
```

Garanas: As the third can compute itself when two out of the triplet are defined.

Hdt80bro: The Reset functions actually undefine their member of the triplet, making the other two required.
```lua
control.Left:SetFunction(function() return control.Right() - control.Width() end)
```
Here we see that it sets `control.Left` to refer to both `control.Right` and `control.Width` (I'm presuming you said `control.Height` by accident, since the horizontal and vertical triplets are decoupled)

I'm also noticing that you removed the emphasis and added some specificity to the comment too. If you'd like, I can go through and change that.

Garanas: Ah - yes, my mistake. It should've been `Width`.

> I'm also noticing that you removed the emphasis and added some specificity to the comment too. If you'd like, I can go through and change that.

That would be great, and after that I think we're as good as we'll get it. @4z0t can you review this PR too?

Garanas:
I'll merge this after the release.

PR #3938, opened 2022-06-08
SC Lua debugger

Ejsstiil:
This PR re-introduces the in-game **Lua** **debugger**!

Even when this feature was always in the game, it was broken due to the missing graphics,
like buton icons, but most importantly - missing breakpoint markers. This lead to the never-ending
spam of error message due that fact and even when you managed to dismiss all that spam, the breakpoint markers were still invisible to the user. And error spam often start occurring again. 
Other than that, the default position and dimensions of the Debugger window was squeezed to its minimum,
and all internal elements had sizes close to zero.

### Fixed: ###
- restored UI icons, markers
- mount command for this in the `init_dev.lua`
- defaulted the position and dimensions of window and its components
- these defaults has to be prepared before the first use so we have to prepare and save them soon enough before the first use (i.e. on cmdline)

### To invoke the Lua debugger, you can use: ###
- Alt + F9 - the default keybind (this sometimes works only after you invoked binds menu (F1) first)
- in game console "SC_LuaDebugger"
- on cmdline /debug (most convenient for dev)

### Controls: ###
- double click on the line sets the breakpoint
- F5 resume execution
- F10 step-in (can't find any way how to do step-over)
- Ctrl + G goto line
- Ctrl + R reloads the file

#### Tips: ####
- If you have a breakpoint set in the frequently used includes, it can significantly slow the game in general, even when your breakpoint is not being hit, better you disable or delete all breakpoints, then set it just before you need it
- due to the above, sometimes game seems froze-up, but you can just wait (or kill game) and clear all breakpoints right after

image


Garanas:
Very exciting 😄 !

Garanas:
I've been thinking about whether we should make this available on other game types too, such as FAF and FAF Develop. The pull requests limits it to a development environment. Is there a specific reason for that?

Ejsstiil:
> I've been thinking about whether we should make this available on other game types too, such as FAF and FAF Develop. The pull requests limits it to a development environment. Is there a specific reason for that?

Was not sure about that. But yea, why not. It's only matter of adding the mount command to inits.

Garanas:
Can you enable the debugger in a multiplayer game, slowing down the game by introducing a breakpoint?

Garanas:
> Was not sure about that. But yea, why not. It's only matter of adding the mount command to inits.

It is not that simple I think - the `coderes` folder isn't shipped with the release. We'd need to do it via the `schook` folder if we want this.

KionX:
When I was fixing this debugger in 2020, I didn't seem to need to mount coderes.
I also had to unpack the game on the "right path".
And breakpoints were unbearably slow.
My assessment was - garbage.

Ejsstiil:
> Can you enable the debugger in a multiplayer game, slowing down the game by introducing a breakpoint?

Of course, you can enable Log during game, debugger, set breakpoints... when client is windowed, you can freeze game by invoking window menu (minimize, maximize, etc.), press alt + space, use traffic shaping.... etc
I know where you heading - the abusing, but the debugger IS already in the Game, regardless of /coderes or not :)

KionX:
I think it should only work in FAF Develop.

Ejsstiil:
I wrote it wrong - all the actions described above can be performed independently of the debugger is opened or not (except breakpoints). And the debugger is already there, invoked by ie. a hotkey. Unless we are able to deliberately disable it (via exe patch) it can't be isolated just for FAF Develop

Garanas:
The debugger is in the game, but without the icons it is near impossible to use. I agree with KionX - we should not include these files into the regular branch.

KionX:
I do not know a way to interfere in the game without breakpoints.

hahn-kev:
Not sure I really see a use case for having the debugger for non dev environment. Like is this so you could debug a bug that came up while one of us is playing some random game? or something else? Only a dev is really going to know how to use it anyway. I guess it would be nice for mod makers so they don't have to have a dev environment setup but they can still debug their code?

Garanas:
> I guess it would be nice for mod makers so they don't have to have a dev environment setup but they can still debug their code?

That would be the only use case, yes.

KionX:
Is it possible to debug replays?

Ejsstiil:
> In my opinion, the debugger is harmless without breakpoints.

This is still not an accurate description of the situation - you can set breakpoints, even if you don't see them in the GUI.

KionX:
Yeah. But it's still better not to help do it.
Or need to block access somehow.
Maybe the command line will help.

Ejsstiil:
> > I guess it would be nice for mod makers so they don't have to have a dev environment setup but they can still debug their code?
> 
> That would be the only use case, yes.

And is a good one I'd say - we can have better and more stable mods thanks to debugger

setup/bin/init_dev.lua
```diff
@@ -455,6 +455,9 @@ MountAllowedContent(fa_path .. '/gamedata/', '*.scd', allowedAssetsScd)
 -- get direct access to preferences file, letting us have much more control over its content. This also includes cache and similar
 MountDirectory(SHGetFolderPath('LOCAL_APPDATA') .. 'Gas Powered Games/Supreme Commander Forged Alliance', '/preferences')
 
+-- mount SC_LuaDebugger GUI graphics
+MountDirectory(locationOfRepository .. '/coderes', '/coderes')
```
KionX, line 459: MountDirectory(InitFileDir .. '/coderes', '/coderes') ?

Ejsstiil:
> Yeah. But it's still better not to help do it. Or need to block access somehow. Maybe the command line will help.

If there is anyone, who wants to harm the game to the others, there is nothing we can do to prevent that, he will always find the way. Its all about his bad motivation. Packet alteration, native debugger, you name it. This debugger is there already. As I said, without patching exe it can't disabled. GPG put it there and its been there since 2014. I'd no worry about it. Really

setup/bin/init_dev.lua
Ejsstiil: locationOfRepository

Ejsstiil:
Offtopic

Ejsstiil:
@Garanas could you please delete this? I no longer wish this PR to be merged.

Garanas:
After discussing it in private - at no point should the debugger become inaccessible to prevent abuse. It is has been in the game for years and that has worked out just fine. It is simply put too valuable to make inaccessible.

@Ejsstiil can you cull out the introduced files that are not used by the debugger? Once that is done, I'll happily merge this in.

PR #3941, opened 2022-06-11
UES0401 fix categories

Uveso:
added ANTIAIR to categories
removed double ANTINAVY


Garanas:
Would this affect builders and everything too? 😄

Uveso:
Oh yes!
It's like the butterfly effect, the change is small but you will play a totally different game! :)

(Ok, I lied - just a change for the threat calculation :D)

PR #3942, opened 2022-06-12
Initialize revamp scoreboard with starting play area

Hdt80bro:
Get https://github.com/FAForever/fa/pull/3933 working for the new score panel. I don't like how I had to do it though.


Garanas:
It was already working, see:

- https://github.com/FAForever/fa/blob/feature/scoreboard-revamp/lua/ui/game/score-alt.lua#L57
- https://github.com/FAForever/fa/blob/feature/scoreboard-revamp/lua/ui/game/score-alt.lua#L407

Can't make a new scoreboard and not include this fancy feature 😄 . The idea of the new scoreboard is to have its own user sync callback, see the first link. That is called every time the sync is. We can keep all the interactions with the scoreboard in this one file by doing so.

Garanas:
I saw your other [message](https://github.com/FAForever/fa/pull/3939#issuecomment-1153042289), it is correct that it doesn't work at the moment - but that is only because the scoreboard doesn't exist from the very start. If it does, then it is one of the first UI elements to be initialized and then it would all work just fine.

Thinking about it thought, when we set the new playable area we should maybe just update the session information too.

Garanas:
This looks better 😄

PR #3943, opened 2022-06-12
lod cut offs of bp's working to fix #3940

### LINKED ISSUE #3940: Sanitize LODCutoff of various blueprints
Description
-----------

As per #3912 all of the game files are now part of the repository. It revealed a couple of skeletons in the closet, one of these skeletons will be described with this issue: the **L**evel **o**f **D**etail (LOD) of meshes. We've been looking at these properties before, but then in relationship to [effects](https://github.com/FAForever/fa/pull/3720).

Separate mesh blueprints have an impact on the framerate of the game. We can draw a mesh through a draw call. In all simplicity, a draw call is a message from the CPU to the GPU to draw something. The more meshes we render, the more draw calls we have. The more draw calls we have, the lower the framerate. In order to keep the number of meshes we need to render in check we can apply various sorts of tricks. We'll discuss some of these out of interest.

### Culling 

This is easier to understand after watching [this video on Youtube](https://www.youtube.com/watch?v=VqH8kcmD-HI). Culling happens on the GPU by default to prevent doing computations on vertices (points on a mesh) that are outside of the viewport. We can cull earlier however - and instead of culling on vertices and triangles, we cull on draw calls. This is something a lot of games do and it is commonly referenced to as [Frustrum Culling](https://learnopengl.com/Guest-Articles/2021/Scene/Frustum-Culling). Every modern game engine supports this by default, and so does Supreme Commander according to a graphics study done by [Andrian Courreges](http://www.adriancourreges.com/blog/2015/06/23/supreme-commander-graphics-study/).

### Batching

A usual draw call is like walking to the grocery store for a single item. Walking to and from the grocery store can easily take you half an hour, but once you're at the grocery store you can retrieve the item in a minute. If we'd do this for every item that we need for our dinner recipe then it could easily take the entire day due to the overhead of walking to and from the grocery store.

Luckily, we can think ahead and make a list of items that we want to acquire from the grocery store. By doing so we can amortize the overhead of going to and from the grocery store over the number of items on our list. We can apply exactly the same idea to our draw calls: instead of rendering one mesh at a time we can queue up calls that are similar and render them all at once.

The metaphor is easier to understand after watching [this video on Youtube](https://www.youtube.com/watch?v=IrYPkSIvpIw). Luckily, the game already does this for us where props, projectiles and units of the same type are batched together.

### Level of Detail

Another method of culling is by gradually reducing the quality of an object as it gets further and further away from the user perspective. This is particularly relevant when the camera of the game allows you to look at content at great distances, such as in Supreme Commander via the strategical zoom. This value has to be set manually. It determines when it is still worth rendering the mesh in question. And that is where this pull request comes in: one skeleton in the closet is these LODCutoff values being all over the place. We'll need to sanitize that.

Course of action
----------------

You can find a complete list of skeleton-like values here:

- https://pastebin.com/0yduuv8U

You can ignore unit blueprint IDs (such as `ues0401`) and the majority of prop blueprints (those that end with `_prop`). We're primarily interested in sanitizing the values of separate mesh entities and of projectiles, as the average prop is already sanitized based on scale by #3662 . Because of this, you can find the majority of skeleton blueprints in the `effects` and the `projectiles` folder.

As an example we'll take the [Aeon adjacency node](https://github.com/FAForever/fa/blob/deploy/fafdevelop/effects/Entities/AeonAdjacencyNode/AeonAdjacencyNode_mesh.bp). It is described using this mesh blueprint:

```lua
MeshBlueprint {
    LODs = {
        {
            ShaderName = 'TMeshNoNormals',
        },
        UniformScale = 1,
    },
}
```

This appears to be quite empty, and that is because it is. The engine has a set of default values for various fields. The complete blueprint looks like this:

```text
  - BlueprintId: /effects/entities/aeonadjacencynode/aeonadjacencynode_mesh
  - BlueprintOrdinal: 3
  - Description: 
  - IconFadeInZoom: 0
  - LODs: table: 17B8D528
  -    1: table: 17B8D550
  -       AlbedoName: /effects/entities/aeonadjacencynode/aeonadjacencynode_albedo.dds
  -       LODCutoff: 1000
  -       LookupName: 
  -       MeshName: /effects/entities/aeonadjacencynode/aeonadjacencynode_lod0.scm
  -       NormalsName: 
  -       Occlude: false
  -       Scrolling: false
  -       SecondaryName: 
  -       ShaderName: TMeshNoNormals
  -       Silhouette: false
  -       SpecularName: 
  -    UniformScale: 1
  - SortOrder: 0
  - Source: /effects/entities/aeonadjacencynode/aeonadjacencynode_mesh.bp
  - StraddleWater: false
  - UniformScale: 1
```

Note that the layout is different because we printed it in game. This shows the issue: the `LODCutoff` value (that determines when to cull the mesh) is there, but its default value is set to a 1000! For your idea, that is the same value that is used for shields. And we're talking here about missiles and other small meshes that nowhere near correspond to the size of shields.

To fix this we can add the field ourselves:

```lua
-- Used by Aeon Adjacency nodes
MeshBlueprint {
    LODs = {
        {
            ShaderName = 'TMeshNoNormals',
            LODCutoff = 200,
        },
        UniformScale = 1,
    },
}
```

Note that the value is capital sensitive: it should be `LODCutoff`. It is also scope sensitive, it should be inside a table of the `LODs` table. The `LODs` table can have multiple entries. This isn't usually the case for these type of meshes.

Test plan
---------

The coarse of action is to create a new branch that is based off `deploy/fafdevelop`. You can adjust up to 4 to 6 blueprints per branch. For each change, make sure to indicate (as a comment in the blueprint, or as a comment on Github) where we can see the change in question. Take an example of the [pull requests](https://github.com/FAForever/fa/pull/3925) done by @Ejsstiil, they are very informative and because of that they are relatively easy to check and merge.

Learning goals
--------------

To be more comfortable with working with Git and the repository in general.

MadMaxFAF:
sets all adjacency nodes cut offs to 300

### DIFF
--- a/effects/Entities/AeonAdjacencyNode/AeonAdjacencyNode_mesh.bp
+++ b/effects/Entities/AeonAdjacencyNode/AeonAdjacencyNode_mesh.bp
@@ -2,6 +2,7 @@ MeshBlueprint {
     LODs = {
         {
             ShaderName = 'TMeshNoNormals',
+            LODCutoff = 300
         },
         UniformScale = 1,
     },
--- a/effects/Entities/CybranAdjacencyNode/CybranAdjacencyNode_mesh.bp
+++ b/effects/Entities/CybranAdjacencyNode/CybranAdjacencyNode_mesh.bp
@@ -2,6 +2,7 @@ MeshBlueprint {
     LODs = {
         {
             ShaderName = 'TMeshNoNormals',
+            LODCutoff = 300
         },
         UniformScale = 1,
     },
--- a/effects/Entities/ForceField01/ForceField01_mesh.bp
+++ b/effects/Entities/ForceField01/ForceField01_mesh.bp
@@ -1,8 +1,10 @@
+-- used by ueb0304
+
 MeshBlueprint {
     LODs = {
         {
             ShaderName = 'UEFQuantumGate',
-            LODCutoff = 1000,
+            LODCutoff = 500,
         },
     },
     SortOrder = 1000,
--- a/effects/Entities/SeraphimAdjacencyNode/SeraphimAdjacencyNode_mesh.bp
+++ b/effects/Entities/SeraphimAdjacencyNode/SeraphimAdjacencyNode_mesh.bp
@@ -2,7 +2,8 @@ MeshBlueprint {
     LODs = {
         {
             ShaderName = 'Seraphim',
-            LookupName = '/textures/environment/Falloff_seraphim_lookup.dds',            
+            LookupName = '/textures/environment/Falloff_seraphim_lookup.dds',    
+            LODCutoff = 300        
         },
     },
 }
--- a/effects/Entities/UEFAdjacencyNode/UEFAdjacencyNode_mesh.bp
+++ b/effects/Entities/UEFAdjacencyNode/UEFAdjacencyNode_mesh.bp
@@ -2,6 +2,7 @@ MeshBlueprint {
     LODs = {
         {
             ShaderName = 'TMeshNoNormals',
+            LODCutoff = 300
         },
         UniformScale = 1,
     },


MadMaxFAF:
all of effects/entities have been reviewed and done unless we want to change the shield cut off which i feel are about right at 1000

Garanas:
Shields are a different story, we'll have to tackle those separately. As an example, try attacking a shield boat with a frigate: the shield impact mesh is gigantic. That should be in line with the size of the impact mesh of a hover shield.

PR #3944, opened 2022-06-12
Add support for multiple upgrade animations

Garanas:
In particular relevant for tech 1 land factories that can either upgrade to an HQ or a support factory. These should have different upgrade animations.

### DIFF
--- a/units/UAB0101/UAB0101_script.lua
+++ b/units/UAB0101/UAB0101_script.lua
@@ -10,14 +10,6 @@
 
 local ALandFactoryUnit = import('/lua/aeonunits.lua').ALandFactoryUnit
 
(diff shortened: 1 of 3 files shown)


units/UAB0101/UAB0101_script.lua
```diff
@@ -10,6 +10,14 @@
 
 local ALandFactoryUnit = import('/lua/aeonunits.lua').ALandFactoryUnit
 
-UAB0101 = Class(ALandFactoryUnit) {}
+UAB0101 = Class(ALandFactoryUnit) {
+    GetUpgradeAnimation = function(self, unitBeingBuilt)
```
Garanas, line 14: To be removed before merging, this is an example

The-Balthazar (approved):

PR #3952, opened 2022-06-13
Update antiair and antinavy categories-Cybran Experimentals

maudlin27:
Update so units use the overlayantiair and overlayantinavy variants for antiair and antinavy where it is very weak at that role
-Soul ripper, Monkeylord and Megalith

### DIFF
--- a/units/URA0401/URA0401_unit.bp
+++ b/units/URA0401/URA0401_unit.bp
@@ -102,7 +102,6 @@ UnitBlueprint {
         'EXPERIMENTAL',
         'GROUNDATTACK',
         'NEEDMOBILEBUILD',
-        'ANTIAIR',
         'COUNTERINTELLIGENCE',
         'STEALTH',
         'VISIBLETORECON',
--- a/units/URL0402/URL0402_unit.bp
+++ b/units/URL0402/URL0402_unit.bp
@@ -74,8 +74,6 @@ UnitBlueprint {
         'BOT',
         'DIRECTFIRE',
         'AMPHIBIOUS',
-        'ANTIAIR',
-        'ANTINAVY',
         'NEEDMOBILEBUILD',
         'INTELLIGENCE',
         'COUNTERINTELLIGENCE',
--- a/units/XRL0403/XRL0403_unit.bp
+++ b/units/XRL0403/XRL0403_unit.bp
@@ -84,8 +84,6 @@ UnitBlueprint {
         'FACTORY',
         'DIRECTFIRE',
         'AMPHIBIOUS',
-        'ANTIAIR',
-        'ANTINAVY',
         'SNIPER',
         'NEEDMOBILEBUILD',
         'SONAR',


speed2CZ:
They do have those weapons, so they should keep those categories

Garanas:
I recall a comment from speed, but it appears to have been deleted. These experimental units should not be build for their anti air or anti navy purpose as they're extremely inefficient at it.

@maudlin27 could you think of a scenario where these changes would break some builder?
@speed2CZ could this break co-op in any fashion?

speed2CZ:
Afaik Mega has pretty good torps

Garanas:
In comparison to their costs they have horrible torpedo / AA capabilities. If I understood the discussion in Discord correct then AIs have difficulty dealing with these units because of these categories. I don't know the details, maybe @Uveso can jump in or Maudlin himself.

Uveso:
we use the unit categories for 2 reasons;
1. select/find a unit for building
2. calculate threat values

as exampe the Soul Ripper:

Using the category for building makes almost no sense.
You will not build a Soul Ripper because of his anti air capabilities.
So it does not need the ANTIAIR category for building.

But if you calculate threat, then you need to take the Soul Ripper into account.
So we need the OVERLAYANTIAIR category to check for the air threat.

So we use the normal categories for primary (strong) weapons
and overlay categories for secondary (weak) weapons.

maudlin27:
For some reason part of my comment got deleted - on the Megalith I meant to say I was borderline whether to consider it as deserving of the ANTINAVY category.  Doing some basic sandbox testing though it can kill 1/3 of its mass in T1 subs and survive with half health (while dying to half its mass in T1 subs) so probably just about justifies retaining ANTINAVY.

Garanas:
> For some reason part of my comment got deleted - on the Megalith I meant to say I was borderline whether to consider it as deserving of the ANTINAVY category. Doing some basic sandbox testing though it can kill 1/3 of its mass in T1 subs and survive with half health (while dying to half its mass in T1 subs) so probably just about justifies retaining ANTINAVY.

I'd say this is an arbitrary threshold. The question you should be asking is: would you build a Megalith to counter subs? And the answer is no - you'd make torpedo bombers.

Sprouto:
The categories themselves have no bearing on threat values - those are determined in the blueprint itself - and not by categories.  The ANTIAIR and ANTINAVY categories are not required for them to emit either air, or sub threat.   In that context they are not needed at all.

Where it impacts AI design most depends upon how you (the AI developer) choses to handle threat.   If your using ANTIAIR and ANTINAVY categories to locate units that can fight in those environments, then they are poor choices - and - they often conflict with the true role of the unit.    On the other hand, if you're using categories to interrogate enemy units as to their role, again - this is a poor choice - so removing those, and replacing them with OVERLAYxxx categories, would, at least, clean up any confusion.   Since you're likely already working from a list of units in the air, regardless of category - the true antiair value should be the sum of the air threat from all of them - regardless of category.   For every unit that may have ANTIAIR as a category, there are likely as many, or more, than do emit air threat - but don't have ANTIAIR category.

The IMAP (intel/threat) system built into the game uses core categories (LAND, AIR, NAVAL, MOBILE, STRUCTURE) to identify layers - but uses the threat values, on the blueprint, for every unit.   This is for, example, how mobile AA units are differentiated from ASF's - both will report as ANTIAIR threat - but only the ASF will report as AIR threat.   In this instance, you can see how the core category keeps them nicely separated - allowing the AI to understand the difference, if coded that way.

Uveso:
@Sprouto 

First, you are absolutly right in case of the base game.
For FAF i would say its only important that we dont build/select a unit by an AA category when we have low AntiAir capabilities on it.

But when i say i need the categories for threat, then i don't mean the threat value.
I check for ANTIAIR in general to make flight paths around AA.
There i dont need a value, just the presents of AA.
Also i don't use the imap threat, i am calculating the threat in my own function based on categories, techlevel and unit count.

In my case if we remove the 'ANTIAIR' for weak AA units/weapons then i still need the 'OVERLAYANTIAIR' to make my calculations to see if the area is free of AA units.

I hope that clarifies my statement a bit

PR #3961, opened 2022-06-15
new mesh and anim

### LINKED ISSUE #3957: Inconsistent transports Attachpoint bones
This bugs affects only AI, and currently the platoon AI for the units that aree to be transported can sometimes softlock because the AI is not able to load the units on the transports.

- I found a problem with at least T2 Aeon transport (uaa0104) and T3 UEF transport (xea0306)

**First some explanation of how transports work:**
Transport table in the bp determines how many T1 slots will T2 and T3 units take
```LUA
Transport = {
    AirClass = true,
    Class1Capacity = 12,  -- Num T1 units it can carry
    Class2AttachSize = 2, -- Class1Capacity / Class2AttachSize = Num T2 units it can carry (6 here)
    Class3AttachSize = 4, -- Class1Capacity / Class3AttachSize = Num T3 units it can carry (3 here)
    TransportClass = 10,
},
```
Transports have these 3 type of attach bones:
```LUA
Attachpoint_Lrg
Attachpoint_Med
Attachpoint
```
T3 units get attached to `Lrg` bone, T2 units to `Med` bone. Now this is combines with how many T1 bones they should take.
So when T3 unit is suppose to be loaded, it will also occupy 4 closest small attach bones (if one of the bones would be already occupied by T2 or T1 unit, then T3 unit can not be attached to this `Lrg` bone)

Now if we look at the bones structure of a working transport UEF T2:
ForgedAlliance_2022-06-15_10-45-41
you can see that the `Lrg` bone always have 4 small bones close to it. Evenly spread out.

In comparison to Aeon T2 transport, which is one of the broken ones:
ForgedAlliance_2022-06-15_10-50-13
The two `Lrg` bones on the left are fighting over the same 2 small bones. Cuz remember, it takes the 4 closest small bones.

So Aeon transport, despite of having 3 `Lrg` bones, can carry only 2 T3 units

Code that figures out for the AI how many units can be carried by a transport:
https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/AI/aiutilities.lua#L1259-L1278
```LUA
function GetNumTransportSlots(unit)
    local bones = {
        Large = 0,
        Medium = 0,
        Small = 0,
    }
    for i = 1, unit:GetBoneCount() do
        if unit:GetBoneName(i) ~= nil then
            if string.find(unit:GetBoneName(i), 'Attachpoint_Lrg') then
                bones.Large = bones.Large + 1
            elseif string.find(unit:GetBoneName(i), 'Attachpoint_Med') then
                bones.Medium = bones.Medium + 1
            elseif string.find(unit:GetBoneName(i), 'Attachpoint') then
                bones.Small = bones.Small + 1
            end
        end
    end

    return bones
end
```
This will find out that the Aeon T2 transport can carry 3 T3 units, even when it can carry only 2, because of the bone structure.

Then when the AI tries to load the units with this call
```LUA
IssueTransportLoad({T3_Unit_1, T3_Unit_2, T3_Unit_3}, {T2_Aeon_Transport})
```
It will not error out, but get stucked forever. The command is issued, but nothing happens.
The AI is waiting until all the units are in Idle state, indicating they got loaded into the transport, before moving out with the transports. But this never happens, because the loading command gets stucked.
ForgedAlliance_2022-06-15_11-10-00

My suggested solution is to move the bones in the transport in order for the `Lrg` and `Med` bones not be fighting over the same small bones.

MadMaxFAF:
new mesh and anim to fix #3957 and stop the unit from clipping through the ground


The-Balthazar:
If that's the same animation I posted with the wonky frame, here's the fix.
[UAA0104_Aopen01Fix.zip](https://github.com/FAForever/fa/files/8911656/UAA0104_Aopen01Fix.zip)

MadMaxFAF:
the model was clipping into the ground hence the mesh change

The-Balthazar:
aircraft clipping into the ground can be fixed in the bp.

MadMaxFAF:
i know but i tried that and weirdness ensued 
image
that is the standard model (w/o your animation)
with an elevation change from 8 to 40

MadMaxFAF:
also that is flat ground

units/UAA0104/UAA0104_unit.bp
```diff
@@ -139,7 +139,7 @@ UnitBlueprint {
             '<LOC ability_aa>Anti-Air',
             '<LOC ability_transport>Transport',
         },
-        AnimationPermOpen = '/units/UAA0104/UAA0104_aopen01.sca',
```
speed2CZ, line 142: Why is it called fix? The original anim should just be replaced

Garanas: To match the fact that we'll fix that during the next release. I also forgot to do the continental fix 😞

PR #3962, opened 2022-06-15
Feature/annotate campaign files

Garanas:
Annotates various campaign files to give a modern touch to campaign-related development. 

Progression:
- [ ] BaseManager.lua
- [ ] OpAI.lua
- [ ] Objectives2.lua
- [ ] ScenarioFramework.lua
- [ ] ScenarioUtils.lua
- [ ] Cinematics.lua

### DIFF
--- a/lua/AI/OpAI/BaseManager.lua
+++ b/lua/AI/OpAI/BaseManager.lua
@@ -544,6 +544,9 @@ BaseManager = ClassSimple {
     -- Functions for tracking the number of engineers working in a base manager
     ---------------------------------------------------------------------------
 
+    --- Add to the engineer count, useful when gifting the base engineers.
+    ---@param self BaseManager  # An instance of the BaseManager class
+    ---@param num number        # Amount to add to the engineer count.
     AddCurrentEngineer = function(self, num)
         if not num then
             num = 1
@@ -551,6 +554,9 @@ BaseManager = ClassSimple {
         self.CurrentEngineerCount = self.CurrentEngineerCount + num
     end,
 
+    --- Subtract from the engineer count
+    ---@param self BaseManager  # An instance of the BaseManager class
+    ---@param num number        # Amount to subtract from the engineer count.
     SubtractCurrentEngineer = function(self, num)
         if not num then
             num = 1
@@ -558,18 +564,30 @@ BaseManager = ClassSimple {
         self.CurrentEngineerCount = self.CurrentEngineerCount - 1
     end,
 
+    --- Retrieve the engineer count
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@return number               # Number of active engineers
     GetCurrentEngineerCount = function(self)
         return self.CurrentEngineerCount
     end,
 
+    --- Retrieve the maximum number of engineers, the base manager won't build more engineers than this
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@return number               # Maximum number of engineers for this base manager
     GetMaximumEngineers = function(self)
         return self.EngineerQuantity
     end,
 
+    --- Add an engineer to the engineer pool of the base manager
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@param unit Unit             # Engineer to add
     AddConstructionEngineer = function(self, unit)
         table.insert(self.ConstructionEngineers, unit)
     end,
 
+    --- Remove an engineer from the engineer pool of the base manager
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@param unit Unit             # Engineer to remove
     RemoveConstructionEngineer = function(self, unit)
         for k, v in self.ConstructionEngineers do
             if v.EntityId == unit.EntityId then
@@ -579,26 +597,44 @@ BaseManager = ClassSimple {
         end
     end,
 
+    --- Defines the maximum number of construction engineers
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@param num number            # New maximum number of construction engineers
     SetMaximumConstructionEngineers = function(self, num)
         self.MaximumConstructionEngineers = num
     end,
 
+    --- Retrieves the maximum number of construction engineers
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@return number               # Maximum number of construction engineers
     GetConstructionEngineerMaximum = function(self)
         return self.MaximumConstructionEngineers
     end,
 
+    ---comment
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@return integer
     GetConstructionEngineerCount = function(self)
         return table.getn(self.ConstructionEngineers)
     end,
 
+    ---comment
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@param bool any
     SetConstructionAlwaysAssist = function(self, bool)
         self.ConstructionAssistBool = bool
     end,
 
+    ---comment
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@return unknown
     ConstructionAlwaysAssist = function(self)
         return self.ConstructionAssistBool
     end,
 
+    ---comment
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@return boolean
     ConstructionNeedsAssister = function(self)
         if not self:ConstructionAlwaysAssist() or self:GetConstructionEngineerCount() == 0 then
             return false
@@ -606,6 +642,10 @@ BaseManager = ClassSimple {
         return true
     end,
 
+    ---comment
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@param unit any
+    ---@return boolean
     IsConstructionUnit = function(self, unit)
         if not unit or unit.Dead then
             return false
@@ -620,6 +660,9 @@ BaseManager = ClassSimple {
         return false
     end,
 
+    ---comment
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@param num any
     SetPermanentAssistCount = function(self, num)
         if num > self.EngineerQuantity then
             error('*Base Manager Error: More permanent assisters than total engineers')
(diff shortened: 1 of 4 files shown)


Garanas:
I broke it there for a second 😃

lua/AI/OpAI/BaseManager.lua
```diff
@@ -73,14 +121,15 @@ local BuildingCounterDefaultValues = {
...
-BaseManager = Class {
-    -- TO DO LIST:
-    -- Expansion at mass positions (Probably a must, but may not fit here - perhaps an OpAI which is enable/disable through bool)
-    -- Maybe build T2 shields if no T3 engs are available for Seraphim/UEF
-    -- Artillery, Nukes, and TML control (At least toggling nukes and TML required, much want for control of where to attack)
-    -- Engineer counts when needing higher tech level engineers (possibly not needed since recovery is hard and not always wanted)
+---@class BaseManager
+---@field Trash TrashBag
+---@field AIBrain AIBrain
+
```
hahn-kev, line 127: this line 127 needs to be removed, for some reason it seems to break the class definition for `BaseManager` when it's removed then autocomplete on self inside functions works fine and shows all the other methods

lua/AI/OpAI/BaseManager.lua
```diff
@@ -4,6 +4,54 @@
...
 ----------------------------------------------------------------------
 
+---@alias SaveFile "AirAttacks" | "AirScout" | "BasicLandAttack" | "BomberEscort" | "HeavyLandAttack" | "LandAssualt" | "LeftoverCleanup" | "LightAirAttack" | "NavalAttacks" | "NavalFleet"
+
+-- types that originate from the map
+
+---@class MarkerChain                           # Name reference to a marker chain as defined in the map
+---@class Area                                  # Name reference to a area as defined in the map
+---@class Marker                                # Name reference to a marker as defined in the map
+---@class UnitGroup                             # Name reference to a unit group as defined in the map
```
hahn-kev, line 14: not totally sure what you mean by types that originate from the map, but we should try and give these proper types. Right now they're not much help to the developer as they are, you call them all names, so are they strings? if they're a type that's returned by some map method, like `map:getMarker(1)` then we should also annotate that map method with the same type to link them together to help the developer, right now they might not know where to get a Area class instance or Marker.

lua/AI/OpAI/BaseManager.lua
```diff
@@ -4,6 +4,54 @@
...
+
+-- types commonly used in repository
+
+---@class BuildCondition                        # { file, function, parameters }
+---@class FileFunctionRef                       # { file, function }
```
hahn-kev, line 19: Similar with these, though it looks like maybe they could be a couple different types. In which case I'd do something like this `@alias FileFunctionRef string|fun(someArg:number):string` that says that a `FileFucnctionRef` can be either a string (maybe a file name? just guessing here) or it can be a function who's first argument is a number and the function returns a string.

lua/AI/OpAI/BaseManager.lua
```diff
@@ -181,10 +230,16 @@ BaseManager = Class {
...
-    --     GroupName = Priority, -- Name not in quotes
-    -- }
+    --- Initialises the base manager. 
+    ---@see See the functions StartNonZeroBase, StartDifficultyBase, StartBase or StartEmptyBase to the initial state of the base
+    ---@param self BaseManager          # An instance of the BaseManager class
+    ---@param brain AIBrain             # An instance of the Brain class that we're managing a base for
+    ---@param baseName UnitGroup        # Name reference to a unit group as defined in the map that represnts the base, usually appended with _D1, _D2 or _D3 
+    ---@param markerName Marker         # Name reference to a marker as defined in the map that represents the center of the base
+    ---@param radius number             # Radius of the base - any structure that is within this distance to the center of the base is considered part of the base
+    ---@param levelTable any            # A table of { { UnitGroup, Priority } } that represents the priority of various sections of the base
```
hahn-kev, line 240: it looks like from how it's used this should have the type `table<UnitGroup, number>`

hahn-kev (changes requested):
I did a bit of research and it doesn't seem there's a good way to leave comments on fields of classes unfortunately.

lua/AI/OpAI/BaseManager.lua
Garanas: These are defined in the map, as part of the `_save.lua` file. That was my intend, it is in the end a string. Initially I wanted to use an alias, but that only shows it as a string.

lua/AI/OpAI/BaseManager.lua
Garanas: Ah, no - this is an array-based table with two entries: the file to look for a function, and a function name. They are both strings. I'm all for a better alternative.

hahn-kev: hum, that is a bit tricky. It looks like we can at least do this
```lua
---@class FileName: string
---@class FunctionName: string
---@class FileFunctionRef # { file, function }
---@field [1] FileName
---@field [2] FunctionName
```

so the first 2 lines just create 2 simple classes that are just strings, this basically the same as an alias, but the names will be preserved as a sort of comment/hint to the developer. Then for the `FileFunctionRef` class we declare it as having 2 fields which are just index 1 and 2 and we can then specify the type of them. This is actually quite similar to how you can define tuples in ts.

lua/AI/OpAI/BaseManager.lua
hahn-kev: ok, I think that makes sense. Since it's a string just indicate that the class extends a string like this `@class Area: string` this is pretty much the same as an alias but the name doesn't get removed.

lua/AI/OpAI/BaseManager.lua
Garanas: I wasn't aware you could define index-based fields too, that is great!

lua/ScenarioPlatoonAI.lua
```diff
@@ -58,8 +40,8 @@ function BuildOnce(platoon)
 end
 
 --------------------------------------------------------
---  function: DefaultOSBasePatrol = AddFunction
---      parameter 0: string: platoon = "default_platoon"
+---  function: DefaultOSBasePatrol = AddFunction
+---      parameter 0: string: platoon = "default_platoon"
 --------------------------------------------------------
 function DefaultOSBasePatrol(platoon)
```
Garanas, line 46: @speed2CZ could you tell me more about this function?

lua/AI/OpAI/BaseManager.lua
hahn-kev: me either, I found out when I tested it lol

lua/AI/OpAI/BaseManager.lua
```diff
@@ -181,10 +239,16 @@ BaseManager = Class {
...
-    -- {
-    --     GroupName = Priority, -- Name not in quotes
-    -- }
+    --- Initialises the base manager. 
+    ---@see See the functions StartNonZeroBase, StartDifficultyBase, StartBase or StartEmptyBase to the initial state of the base
+    ---@param self BaseManager          # An instance of the BaseManager class
+    ---@param brain AIBrain             # An instance of the Brain class that we're managing a base for
+    ---@param baseName UnitGroup        # Name reference to a unit group as defined in the map that represnts the base, usually appended with _D1, _D2 or _D3 
+    ---@param markerName Marker         # Name reference to a marker as defined in the map that represents the center of the base
+    ---@param radius number             # Radius of the base - any structure that is within this distance to the center of the base is considered part of the base
```
speed2CZ, line 248: Radius is used for managing factories, engineers and forming platoons

lua/AI/OpAI/BaseManager.lua
```diff
@@ -349,7 +463,13 @@ BaseManager = Class {
         return self.OpAITable[name]
     end,
 
-    -- Add a group created in the editor to the base manager to be maintained by the manager
+    --- Adds a build group to the base manager that it needs to maintain
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@param groupName UnitGroup   # Name reference to a unit group as defined in the map that represents the unit group to build 
+    ---@param priority number       # Priority that indicates how important this build group is in comparison to others
+    ---@param spawn boolean         # Flag that indicates whether the build group is immediately spawned
+    ---@param initial boolean       # ??
```
speed2CZ, line 471: if true, it skips sorting the build groups by priority
it should be false (or rather nil) for any group that is added after the base is initialized

lua/AI/OpAI/BaseManager.lua
```diff
@@ -393,47 +527,76 @@ BaseManager = Class {
...
     end,
 
+    --- Retrieves the radius of the base manager, which is used to search for factories and engineers
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@return number
     GetRadius = function(self)
         return self.Radius
     end,
 
+    --- Defines the radius of the base manager, which is used to search for factories and engineers
```
speed2CZ, line 544: and forming platoons

lua/AI/OpAI/BaseManager.lua
```diff
@@ -393,47 +527,76 @@ BaseManager = Class {
...
+    ---@return nil
     SetRadius = function(self, rad)
         self.Radius = rad
     end,
 
     ---------------------------------------------------------------------------
     -- Functions for tracking the number of engineers working in a base manager
     ---------------------------------------------------------------------------
+
+    --- Add to the engineer count, useful when gifting the base engineers.
```
speed2CZ, line 556: This is used internally by the main base manager engineer thread to keep track of how many engineers are in the base

lua/AI/OpAI/BaseManager.lua
```diff
@@ -443,33 +606,55 @@ BaseManager = Class {
         end
     end,
 
+    --- Defines the maximum number of construction engineers
```
speed2CZ, line 609: constructions engineers are those that can start new construction projects, others can still assist them

lua/AI/OpAI/BaseManager.lua
```diff
@@ -443,33 +606,55 @@ BaseManager = Class {
...
     end,
 
+    ---comment
+    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@return integer
     GetConstructionEngineerCount = function(self)
         return table.getn(self.ConstructionEngineers)
     end,
 
+    ---comment
```
speed2CZ, line 630: `self.ConstructionAssistBool` determines if the engineers that are not constructions engineers or permanently assisting engineers, should assist construciton projects

lua/AI/OpAI/BaseManager.lua
```diff
@@ -484,42 +669,74 @@ BaseManager = Class {
         return false
     end,
 
+    ---comment
```
speed2CZ, line 672: Permanent assist engineers are assigned to only assist factories and are evenly split among all factories, these engineers won't do any other tasks in the base.

lua/ScenarioPlatoonAI.lua
```diff
@@ -11,42 +11,24 @@ local ScenarioFramework = import('/lua/ScenarioFramework.lua')
...
+    return humans
 end
 
 --------------------------------------------------------
---  function: BuildOnce = AddFunction
---      parameter 0: string: platoon = "default_platoon"
+---  function: BuildOnce = AddFunction
+---      parameter 0: string: platoon = "default_platoon"
 --------------------------------------------------------
 function BuildOnce(platoon)
```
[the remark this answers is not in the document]
speed2CZ: It's an extra function that can be set to run on the platoon when the platoon is formed. Setting it's priority to 0 so it won't be rebuild again when it dies.

lua/ScenarioPlatoonAI.lua
speed2CZ: This is some vanilla campaign function. The logic for the vanilla platoons is complicated and hard to manage, so I prefer to replace it to the FA format rather than investigating too much what it does and how.

lua/ScenarioPlatoonAI.lua
```diff
@@ -494,15 +482,11 @@ function PatrolThread(platoon)
...
---      PatrolRoute - List of locations to patrol
---      PatrolChain - Chain of locations to patrol
---  function: RandomPatrolThread = AddFunction
---      parameter 0: string: platoon = "default_platoon"
---------------------------------------------------------------------
+--- Platoon patrols a random set of locations
+--- | Platoon data value    | Description   |
+--- | --------------------- | ------------- |
+--- | PatrolRoutes          | List of locations to patrol
+--- | PatrolChains          | Chain of locations to patrol
```
speed2CZ, line 489: PatrolRoute or PatrolChain, not plural

PR #3967, opened 2022-06-17
ADD LODCutoff in Projectiles-3

7h3-M0rPh3u5:
ADD LODCutoff in Projectiles bp for bombs, missiles and torpedos : UEL0001, UEB2108, UES0304, UEL0111, XEL0306, UEL0201, UEB2302, UEB2401

The mainpart here is the lodcutoff add to shells that are used by all UEF tanks and ships.


Garanas:
Example of the issues that we're fixing:


_Frigates having strange pathing when finished building_


_Battleships having strange pathing when finished building_

Garanas:
I wasn't even aware the MA12 Striker used mesh-based projectiles 😄

PR #3972, opened 2022-06-19
Add the ability to force a weapon retarget on selected units through a hotkey

Garanas:
As per #3857 weapons no longer retarget as usual. The average scenario does not require this, but especially at the start of the game there are moments where a weapon retargeting is critical. With this pull request we introduce the capability to force a weapon retarget.

As weapon retargeting is not cheap there is a cooldown of one second per unit at which it can retarget its weapons. To confirm to the player that it occurred a small print message is displayed on screen indication how many weapons are rechecking their targets.


Garanas:
image

As an example of the output. The hotkey can be found below the `Orders` header:

image

PR #3978, opened 2022-06-19
Fix the tactical missile enhancements

Garanas:
As reported [on the forums](https://forum.faforever.com/topic/4139/game-version-3738/6) there was an issue with the tactical missile enhancement. The fix was not turning the death weapon into a manual fire weapon. I don't understand why this fixes the issue - the death weapon and the tactical missile appear to be unrelated.

But here we are.


Garanas:
After writing about it to Balthazar - I think a unit can have only one weapon with a manual fire. And that the first weapon it finds is the one it uses. In other words: the user on the forums was triggering the **death weapon** of his ACU and not the tactical missile weapon.

PR #3981, opened 2022-06-20
Code Annotation

MrRowey:
First got at annotating the code for the Intellisense plugin.

Their are a few that are still not defined just wanted to ensure I'm on the right path

### DIFF
--- a/engine/Core.lua
+++ b/engine/Core.lua
@@ -1,6 +1,8 @@
 ---@declare-global
 ---Module: Core
 -- @module Core
+---@alias Vector unknown
+
 
 ---
 --  Create a target object
@@ -161,15 +163,13 @@ end
 function OrientFromDir(vector)
 end
 
----
---  Create a point vector(px,py,pz, vx,vy,vz)
----@param px number
----@param py number
----@param pz number
----
----@param vx number
----@param vy number
----@param vz number
+---Create a point vector(px,py,pz, vx,vy,vz)
+---@param px Vector
+---@param py Vector
+---@param pz Vector
+---@param vx Vector
+---@param vy Vector
+---@param vz Vector
 function PointVector(px,py,pz, vx,vy,vz)
 end
 
@@ -179,9 +179,13 @@ end
 function RPCSound(sound)
 end
 
----
---  Create a 2d Rectangle (x0,y0,x1,y1)
-function Rect()
+
+---Create a 2d Rectangle (x0,y0,x1,y1)
+---@param x0 integer
+---@param y0 integer
+---@param x1 integer
+---@param y1 integer
+function Rect(x0,y0,x1,y1)
 end
 
 ---
@@ -286,9 +290,10 @@ end
 function Trace(enable)
 end
 
----
---  Addition of two vectors
-function VAdd()
+---Addition of two vectors
+---@param v1 Vector
+---@param v2 Vector
+function VAdd(v1, v2)
 end
 
 ---
@@ -316,15 +321,15 @@ end
 
 ---
 --  Distance between two 3d points (v1,v2)
----@param v1 number
----@param v2 number
+---@param v1 Vector
+---@param v2 Vector
 function VDist3(v1,v2)
 end
 
 ---
 --  Square of Distance between two 3d points (v1,v2)
----@param v1 number
----@param v2 number
+---@param v1 Vector
+---@param v2 Vector
 function VDist3Sq(v1,v2)
 end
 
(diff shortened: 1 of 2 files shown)


engine/Core.lua
```diff
@@ -304,12 +316,16 @@ end
 
 ---
 --  Distance between two 3d points (v1,v2)
-function VDist3()
+---@param v1 number
+---@param v2 number
```
speed2CZ, line 320: these are vectors, not numbers

engine/Core.lua
```diff
@@ -304,12 +316,16 @@ end
...
+function VDist3(v1,v2)
 end
 
 ---
 --  Square of Distance between two 3d points (v1,v2)
-function VDist3Sq()
+---@param v1 number
+---@param v2 number
```
speed2CZ, line 327: these are vectors, not numbers

lua/AI/AIBehaviors.lua
```diff
@@ -55,7 +54,7 @@ function CDRRunAway(aiBrain, cdr)
                         if canTeleport then
                             IssueTeleport({cdr}, runSpot)
                         else
-                            cmd = plat:MoveToLocation(runSpot, false)
+                           local cmd = plat:MoveToLocation(runSpot, false)
```
speed2CZ, line 57: this local should be removed as it's not used

lua/AI/AIBehaviors.lua
```diff
@@ -384,6 +408,10 @@ function AirUnitRefit(self)
     end
 end
 
+---comment
+---@param unit Unit
+---@param plan integer
```
speed2CZ, line 413: plan is a string with the plan name to use

lua/AI/AIBehaviors.lua
```diff
@@ -384,6 +408,10 @@ function AirUnitRefit(self)
...
+---@param data integer
```
speed2CZ, line 414: data is a table with platoon data

engine/Core.lua
MrRowey: Hmmm okay Vector is not a Type I'm able to select @Garanas is this something that needs adding then ?

MrRowey:
@speed2CZ What Parameter is base then as could not fine where that is set

lua/AI/AIBehaviors.lua
```diff
@@ -608,15 +632,9 @@ CommanderOverrideCheck = function(self)
...
--- Args:
--- platoon - the platoon
--- Description:
--- Finds the experiemental unit in the platoon (assumes platoons are only experimentals)
--- Returns:
--- experimental or nil
--------------------------------------------------------
+---Finds the experiemental unit in the platoon (assumes platoons are only experimentals)
+---@param platoon Platoon
+---@return boolean                # experimental or nil
```
hahn-kev, line 637: you marked this as returning a boolean, when it looks like it actually returns a unit or nil. You can indicate that like this `@return Unit|nil` that way the person calling this function knows they should check for nil before using the value.

engine/Core.lua
hahn-kev: If you're not sure you can always create an alias which we can change later. So like at the top of Core you could add `---@alias Vector unknown` then any place a vector is used you can name it that, then later on we can update the alias to be the correct type.

engine/Core.lua
```diff
@@ -161,7 +163,14 @@ end
...
 --  Create a point vector(px,py,pz, vx,vy,vz)
-function PointVector()
+---@param px number
+---@param py number
+---@param pz number
+---
+---@param vx number
+---@param vy number
+---@param vz number
+function PointVector(px,py,pz, vx,vy,vz)
```
hahn-kev, line 173: make sure to specify a return type for this function and any other functions.

hahn-kev (commented):
Looks like a good start, there's some functions where the comment is just `comment` but I'm assuming those are ones you haven't gotten to yet. Don't forget to specify a return type on your core functions, vector is tricky because it's a type that hasn't been defined yet, but I left a comment about that.

4z0t:
You must add `---@return` too for better readability and understanding

MrRowey:
> You must add `---@return` too for better readability and understanding

from what jip told me if you do the --- for the auto complete and it dose not give a return then its not needed

engine/Core.lua
```diff
@@ -159,9 +163,14 @@ end
 function OrientFromDir(vector)
 end
 
----
---  Create a point vector(px,py,pz, vx,vy,vz)
-function PointVector()
+---Create a point vector(px,py,pz, vx,vy,vz)
+---@param px Vector
+---@param py Vector
+---@param pz Vector
```
speed2CZ, line 169: p - x y z are elements of one vector, it's not 3 vectors

engine/Core.lua
```diff
@@ -329,12 +350,17 @@ end
...
+---@param y Vector
+---@param z Vector
+function Vector(x,y,z)
 end
 
 ---
 --  Create a vector (x,y)
-function Vector2()
+---@param x Vector
+---@param y Vector
```
speed2CZ, line 362: numbers not vectors

PR #3985, opened 2022-06-22
Adjust naval target priorities

Spikey84:
Here is my proposed changed to naval target priorities. Below is a list of changes, i will log changes as i make them.

- All frigs, destros, and battleships with aa have had their TP simplified (removed t3 bomber in favor of general bombers)
- All Battleships now target T3 navy before targeting generic navy.
- Tempest's TP have been changed to the following
```
'NAVAL MOBILE EXPERIMENTAL',
'NAVAL MOBILE TECH3',
'NAVAL MOBILE',
'EXPERIMENTAL',
'TECH3',
'ALLUNITS',
```
                
More soon

PR #3986, opened 2022-06-23
Balance/adjust naval target priorities

Garanas:
Copied from #3985 
 
 Here is my proposed changed to naval target priorities. Below is a list of changes, i will log changes as i make them.

- All frigs, destros, and battleships with aa have had their TP simplified (removed t3 bomber in favor of general bombers)
- All Battleships now target T3 navy before targeting generic navy.
- Tempest's TP have been changed to the following
```
'NAVAL MOBILE EXPERIMENTAL',
'NAVAL MOBILE TECH3',
'NAVAL MOBILE',
'EXPERIMENTAL',
'TECH3',
'ALLUNITS',
```
                
More soon


Garanas:
Speaking of, I think that:
- Tech 3 battleships should target Tech 3 Navy, Tech 2 Navy and then the remainder
- Tech 2 destroyers should target Tech 2 Navy and then the remainder

Which is in line with Turin's proposal where units prefer to attack other units of the same tech level.

Garanas:
What if we'd make the Seraphim Destroyer prefer hover over attacking tech 1 frigates?

MrRowey:

PR #3996, opened 2022-06-24
T3 land target priorities

Spikey84:
Hopefully the format of this pr is better than the lost one :P

This PR has my proposed T3 land TP changes and simplifications. All changes will be explained below and updated as more are added. Any balance team members are welcome to commit to this PR (I think thats how this works lol).

Changes
- Refactored all T3 MAAs to be more intuitive (for example hitting asf before t1 bombers)
- Simplified T3 assault bots to target t1 and 2 land equally with pd (im not sure i like this)
- Simplified long range land units
- Made percys more consistent with other assault bots (this may need to be changed)


Garanas:
How did you manage to get a conflict on every change you made

PR #3995, opened 2022-06-23
Refactor effect utilities part 1

Hdt80bro:
Adds the optimized functions in part 1 of https://github.com/FAForever/fa/issues/3757 in `EffectUtilitiesGeneric.lua` and imports them to `EffectUtilities.lua`.
Also adds annotations to both files.
Looks like an optimization for `ApplyWindDirection` slipped in too.

PR #3998, opened 2022-06-25
Refactor effect utilities part 2

Hdt80bro:
Cleaned up the functions and deprecated all of the optimized functions.
~~Some of the function parameters' names got changed, but I foolishly put the annotations on part 1's PR... will change the annotation parameters to match after both PR's merge.~~ Ah, there we go
I have not applied Chebyshev recursion in `SpawnBuildBots` yet, but I did replace one of the trig functions with a cheaper `math.sqrt`.
The only logical difference that should change anything is that the Aeon teleport-out effects are now offset vertically like everyone else, while they weren't before. It might have been a bug. Looking at it, I actually can't tell the difference in-game.


Garanas:
Due to merging of #3995 there are a lot of conflicts - can you solve those @Hdt80bro ?

Hdt80bro:
I knew it'd happen, the way I was mixing changes to files between PR's... it should be all good now.

Hdt80bro:
So was the Aeon portion of `PlayTeleportOutEffects` (at https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/EffectUtilities.lua#L1210) supposed to always be offset like the rest of the factions? I can't visually tell a difference from applying `OffsetEmitter(0, Yoffset, 0)` and not.

lua/EffectUtilities.lua
```diff
@@ -8,174 +8,220 @@
...
+local unpack = unpack
+
+local MathAbs = math.abs
+local MathCeil = math.ceil
+local MathPow = math.pow
+local MathMax = math.max
+local MathMin = math.min
+local MathSin = math.sin
+local MathSqrt = math.sqrt
+local MathTau = 2 * math.pi
```
4z0t, line 29: `MATH_TAU` or `mathTau`

Hdt80bro: Is the idea here that `math.tau` isn't in the library, so it shouldn't be upvalued like the rest of the library?

4z0t: It is a constant, not a function 🙂

Hdt80bro: Alright, I can see that. It's like I'm defining a new constant, not upvaluing a constant.

Hdt80bro: Since it's a definition instead of an upvalue, ideally I'd like it to be below all upvalues... but it just fits so nicely with the other math upvalues.

Garanas:
I'm going to run a few games with this, see how it ends up. I'll try and merge it later this week

Hdt80bro:
@Garanas how'd it go?

Garanas:
Good question, I can't recall. Not sure if I end up doing that in the end. Given the amount of changes for the next release (which is on Sunday) I'll merge this in once the release is considered stable

Issue #3997, opened 2022-06-24
Medusa don't stun if they die before projectile hits.


RabidPope:
If a medusa fires a shot and then dies, the shot wont stun.

To reproduce, make a medusa fire a shot and ctrl k it.

Tried it with the aeon bomber and it does not have the same bug.
![Medusa shot](https://user-images.githubusercontent.com/55444317/175719656-3f7c5960-019c-42f0-b3a6-75ed22ae9261.png)

Hdt80bro:
Here it is: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Projectile.lua#L596
The destroyed Shimmer falling through the air still counts as a launcher, but a Medusa on the ground turns into a wreck immediately.

RabidPope:
So im guessing the stun value needs to be stored in a variable on the projectile when fired?

Hdt80bro:
The problem is that area buffs need an instigator to determine the army, and when the launcher dies, there's nothing left. You *could* use the projectile itself, except that it isn't a unit. This is a requirement that makes sense for literally all buffs except `STUN`. All the `AddBuff` method does with the unit when it's a stun type is get its army, which we can easily extracted from the projectile. I'll put together a PR.

PR #4004, opened 2022-06-25
Let destroyed units' projectiles stun enemies

Hdt80bro:
Fix https://github.com/FAForever/fa/issues/3997
I'd prefer to change `utilities.GetTrueEnemyUnitsInSphere` to accept an army instead of a Unit, but that would create a little extra overhead and confuse `AddBuff`'s arguments.

### DIFF
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -51,7 +51,6 @@ local EntitySetMaxHealth = _G.moho.entity_methods.SetMaxHealth
 local EntitySetHealth = _G.moho.entity_methods.SetHealth
 local EntityGetPositionXYZ = _G.moho.entity_methods.GetPositionXYZ
 local EntityDestroy = _G.moho.entity_methods.Destroy
-local UnitAddBuff = import('/lua/sim/Unit.lua').Unit.AddBuff
 
 local ProjectileGetLauncher = _G.moho.projectile_methods.GetLauncher
 
@@ -600,7 +599,7 @@ Projectile = Class(moho.projectile_methods) {
                         if not target and v.BuffType == 'STUN' and isAreaEffect then
                             -- all the AddBuff method will do with the instigator for stunning is
                             -- get its Army anyway; set that and we're good
-                            UnitAddBuff({Army = self.Army}, v, self:GetPosition())
+                            import('/lua/sim/Unit.lua').Unit.AddBuff({Army = self.Army}, v, self:GetPosition())
                         end
                     end
                     -- Check for target validity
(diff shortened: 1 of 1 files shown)


lua/sim/Projectile.lua
```diff
@@ -51,6 +51,7 @@ local EntitySetMaxHealth = _G.moho.entity_methods.SetMaxHealth
 local EntitySetHealth = _G.moho.entity_methods.SetHealth
 local EntityGetPositionXYZ = _G.moho.entity_methods.GetPositionXYZ
 local EntityDestroy = _G.moho.entity_methods.Destroy
+local UnitAddBuff = import('/lua/sim/Unit.lua').Unit.AddBuff
```
Garanas, line 54: AddBuff is a Lua function - we shouldn't cache those into an up value. These are often hooked. Caching them via an up value may break those hooks.

Garanas:
This solves this issue, but I think we should overhaul stun-like damage all together in another pull request.

At the moment the criteria for receiving the damage is not the same as for applying the stun, see also https://github.com/FAForever/fa/pull/3987 . Damage is calculated as a sphere-collision box query, where as `GetTrueEnemyUnitsInSphere` only takes into account the center of the unit we're testing against. That causes inconsistencies.

What if we'd apply another damage instance where we have:
- Damage type 'Stun'
- Damage amount as duration (every 1 point of damage is 1 tick of stun)
- Pass the buff as 'instigator' to pass the information to check if we should stun the unit or not

And then take care of the stun logic in OnDamage of the Unit class? We can then exclude the damage for props and shields.

lua/sim/Projectile.lua
Hdt80bro: The issue is that the only object we could call it on is nil.

Hdt80bro: I thought that `import` resolved all hooks before it returns? Or are you talking about in general, in case a mod updates the method mid-game for some reason?

Hdt80bro:
That sounds like a much better solution. However, how would we handle the different targeting types buffs can have? Would we just leave in Stun as a buff type for self-applications, and then also have a stun damage type for area-of-effect?

Garanas:
You have `Damage` and `DamageArea` to make a distinction. We'll leave in the old method as mods expect that. The stun mechanic of the base game would then just work properly.

lua/sim/Projectile.lua
Garanas: No - it does not. An example is if a subclass hooks into `AddBuff` (to prevent being able to be stunned, as an example) but we still reference the base class implementation here.

Hdt80bro: Hmm, alright, I've removed the upvalue

PR #4005, opened 2022-06-26
T2 land target prios

Spikey84:
Simplifies and refactors T2 lands TPs. List of changes below, will update as more are made. 

- Fixed pillar being different from other tanks and took out restriction on targeting tmd.
- Simplified Flack, i would appreciate feedback on this change.
- Simplified MML, also would like some feedback.
- Fixed tmd on range bots.
- Fixed mongoose targeting issue (see thread on bal team discord)


Spikey84:
Redone in https://github.com/FAForever/fa/pull/4062

PR #3918, opened 2022-05-31
Initial Chrono Dampener effect based on implementation of Deribus

Garanas:
after: 

https://user-images.githubusercontent.com/15778155/171293892-d0d1bf9b-bc36-418e-b135-edad83d30ea4.mp4

PR #4008, opened 2022-06-26
Overhaul stun logic to match the stun radius with the damage radius

Garanas:
As discussed in #4004 the stun radius does not match the damage radius. As a result we have #3478 and https://github.com/FAForever/fa/pull/3987 that try and fix this radius. But to no avail, as the query is fundamentally different:

- For damage the query is a sphere -> collision box intersection
- For stun the query was a sphere -> unit center intersection

As a result, the stun radius will always be off depending on the unit that you're applying the stun to. This pull request introduces a new damage type (coined 'Stun') that is ignored by entities except for units that the stun is applicable to.

### DIFF
--- a/lua/sim/DefaultDamage.lua
+++ b/lua/sim/DefaultDamage.lua
@@ -8,24 +8,119 @@
 --**  Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
 --****************************************************************************
 
--- upvalue for performance
+-- scope as upvalue for performance
 local Damage = Damage 
 local DamageArea = DamageArea
 
--- cache for performance
+local Entity = import('/lua/sim/Entity.lua').Entity
+
+--- A cache of categories for stun allow / disallow strings
+local CategoriesCache = { }
+
+--- A dummy entity that is used to pass over stun parameters via stun instances
+---@class StunEntity
+---@field CategoriesAllowed Categories
+---@field CategoriesDisallowed Categories | boolean
+local StunEntity = Class(Entity) {
+
+    CategoriesAllowed = categories.ALLUNITS,
+    CategoriesDisallowed = false,
+
+    --- Defines the categories that we're allowed to stun, caching the parsing. Defaults to categories.ALLUNITS
+    ---@param self StunEntity       # Instance of StunEntity
+    ---@param cats string | nil     # String of categories, defaults to categories.ALLUNITS
+    SetAllowedCategories = function (self, cats)
+
+        -- if no categories given, we apply to all units
+        if not cats then
+            self.CategoriesDisallowed = categories.ALLUNITS
+            return
+        end
+
+        local cached = CategoriesCache[cats]
+        if cached then
+            self.CategoriesAllowed = cached
+        end
+
+        cached = ParseEntityCategory(cats)
+        CategoriesCache[cats] = cached
+        self.CategoriesAllowed = cached
+    end,
+
+    --- Defines the categories that we're allowed to stun, caching the parsing. Defaults to false
+    ---@param self StunEntity       # Instance of StunEntity
+    ---@param cats string | nil     # String of categories, defaults to false
+    SetDisallowedCategories = function (self, cats)
+
+        -- if no categories given, we do not exclude units
+        if not cats then
+            self.CategoriesDisallowed = false
+            return
+        end
+
+        local cached = CategoriesCache[cats]
+        if cached then
+            self.CategoriesDisallowed = cached
+        end
+
+        cached = ParseEntityCategory(cats)
+        CategoriesCache[cats] = cached
+        self.CategoriesDisallowed = cached
+    end,
+}
+
+--- Create an instance for each army
+local StunEntityInstances = { }
+for k, brain in ArmyBrains do 
+    StunEntityInstances[k] = StunEntity({ Army = brain:GetArmyIndex() })
+end
+
+--- Applies an area stun at the given location, caches the parsing of categories
+---@param position Vector           # Origin of the stun
+---@param duration number           # Duration of the stun, in seconds
+---@param radius number             # Radius of the stun, in oGrids
+---@param allowed string | nil      # String representation of allowed categories, nil allows all units
+---@param disallowed string | nil   # String representation of disallowed categories, nil disallows no units
+---@param army number               # Army where the stun damage originates from
+function ApplyAreaStun (position, duration, radius, allowed, disallowed, army)
+    StunEntityInstances[army]:SetAllowedCategories(allowed)
+    StunEntityInstances[army]:SetDisallowedCategories(disallowed)
+
+    DamageArea(StunEntityInstances[army], position, radius, 10 * duration, 'Stun', false)
+end
+
+--- Applies a single-target stun, caches the parsing of categories
+---@param target Unit               # Target to stun
+---@param duration number           # Duration of the stun, in seconds
+---@param allowed string | nil      # String representation of allowed categories, nil allows all units
+---@param disallowed string | nil   # String representation of disallowed categories, nil disallows no units
+---@param army number               # Army where the stun damage originates from
+function ApplyStun (target, duration, allowed, disallowed, army)
+    StunEntityInstances[army]:SetAllowedCategories(allowed)
+    StunEntityInstances[army]:SetDisallowedCategories(disallowed)
+
+    Damage(StunEntityInstances[army], target, 10 * duration, 'Stun')
+end
+
+-- scope as upvalue for performance
 local VectorCache = Vector(0, 0, 0)
-local MathFloor = math.floor 
-local CoroutineYield = coroutine.yield 
+local CoroutineYield = coroutine.yield
 
 local EntityBeenDestroyed = _G.moho.entity_methods.BeenDestroyed
 local EntityGetPositionXYZ = _G.moho.entity_methods.GetPositionXYZ
 
---- Performs damage over time on a unit.
+--- Applies damage over time on a single target
+---@param instigator Unit | Projectile | Weapon | Entity
+---@param unit Unit         # Unit to apply damage instances to
+---@param pulses number     # Number of damage instances
+---@param pulseTime number  # Time between damage instances
+---@param damage number     # Damage amount
+---@param damType string    # Damage type such as 'Normal' or 'Force'
+---@param friendly boolean  # Flag to indicate we can damage allied units
 function UnitDoTThread (instigator, unit, pulses, pulseTime, damage, damType, friendly)
 
     -- localize for performance
     local position = VectorCache
-    local DamageArea = DamageArea
     local CoroutineYield = CoroutineYield
 
     -- convert time to ticks
@@ -42,7 +137,15 @@ function UnitDoTThread (instigator, unit, pulses, pulseTime, damage, damType, fr
     end
 end
 
---- Performs damage over time in a given area.
+--- Applies damage over time in an area
+---@param instigator Unit | Projectile | Weapon | Entity
+---@param position Vector       # Location to apply damage instances at
+---@param pulses number         # Number of damage instances
+---@param pulseTime number      # Time between damage instances
+---@param radius number         # Damage radius
+---@param damage number         # Damage amount
+---@param damType string        # Damage type such as 'Normal' or 'Force'
+---@param friendly boolean      # Flag to indicate we can damage allied units
 function AreaDoTThread (instigator, position, pulses, pulseTime, radius, damage, damType, friendly)
 
     -- localize for performance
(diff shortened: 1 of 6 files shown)


Garanas:
Still need to process other stun-based weapons of the base game.

Tagada14:
I like the overhaul but would it be possible to still stun targets that have a personal shield? Otherwise, it would make Obsidians really strong and an Aeon vs Aeon match-up with Chrono would be awkward, to say the least.

Garanas:
Yes, that is possible! I'll add it in.

Garanas:
The only unit that is skipped is the chrono field, as that should be tackled with #3918 once that is more finalized.

Hdt80bro:
It looks like you've flattened the stun durations for the different tech levels. I know adding more buffs with different target restrictions is a yucky way to it, but I wonder if we could change that with this PR by adding different stun durations for different tech levels in the same buff, and then pull that stun value out of the buff depending on the affected unit tech level in the `OnDamage` method. It'd definitely reduce the amount of fine-tuning you could do with `TargetAllow` and `TargetDisallow` though.

Garanas:
I flattened them because they felt unintuitive from a user perspective. The UI doesn't indicate the duration of the stun, yet some units are 'unstunned' sooner than others. 

I'd be interested in having some tech-based curve on stuns if it is easy to understand for the user. But until that day, I'd rather the stun duration to be 'reliable' for the user: it either stuns, or doesn't stun. And if it stuns, it stuns for the same duration for each instance.

RabidPope (approved):
Pretty nice. 
Downloaded and tried it out. Tested the following for medusa and sacu:

No stun through shields. 
Stun on personal shields. 
Stun after death.

Will try and test the other units later.
Only thing wrong I found was the sacu does not explode.

Tagada14:
I don't think that the stun duration should be flattened for different tech tiers. I understand your UX concern but I think that the balance reasons outweigh it  as with a flat number you either need to keep it low enough to not make the stun OP vs higher tech tiers and then it's pretty useless vs tech 1 and feels underwhelming over all. Personally I don't think it's a big deal for UX nor is it super unintuitive. Higher tech units are more advanced and therefore less vulnerable to the EMP.

Garanas:
The only two units that are not flat already are the Medusa (url0103) and the SACU (url0301), there are (in my opinion) better ways of applying a curve than trying to stun multiple times in an area. Especially when we're talking about the overhead in terms of performance 😄 .

Note that typically higher tech units do not get stunned at all. As an example, the Medusa doesn't stun tech 3+ units.

Hdt80bro:
Would we be interested in a global stun curve of something like 
image
(don't worry, I already have code for it)
We would then set the area buff code to apply stun time based on the curve (and setup stun resistances for either the tech levels or tech level differences, and also set stun falloffs for the launchers), instead of creating multiple area buffs with different stun times and specific targets.

Garanas:
What would the Lua code be that computes that?

Hdt80bro:
```lua
-- this function would ideally be optimized for various support parameter counts, and allow multiple returns
-- I've left most of the code unoptimized, except for adding memoization (a critical piece of performance)
function Memoize(fn, cache)
    cache = cache or {}
    return function(...)
        local key = table.concat(args, ',')
        local stored = cache[key]
        if stored then
            return stored
        end
        value = fn(unpack(args))
        cache[key] = value
        return value
    end
end
-- basis of the stun curve. the rest of the code just standardizes the curve with respect to the falloff parameter
-- by shifting and scaling by the inflection point
function RawStunCurve(falloff, resistance)
    resistance =- resistance
    return math.pow(1 + falloff, resistance/ (1 + math.pow(falloff, resistance))
end
-- don't have a closed formed expression for this, so just plug
--    x = (p^(x + 1) - p^(-x) + p - 1)/(ln((p + 1)^(p - 1) p^(-2 p - 1)) + p^(x + 1) ln(p + 1) + ln(p/(1 + p))/p^x)
-- into your favorite equation solver to get a few values (where p is the falloff), and use logarithmic interpolation for
-- everything in-between (it probably doesn't even need to be that accurate, since the values only range from -.6 to -.35)
local inflections = {
    [1] = -0.584187070149687,
    [1.25] = -0.583036548171388,
    [1.5] = -0.580405674598629,
    [2] = -0.573272758504428,
    [3] = -0.557521806487837,
    [4] = -0.542838508621603,
    [5] = -0.529782219491405,
    [6] = -0.518212959135337,
    [7] = -0.507903184616362,
    [10] = -0.482647062730128,
    [15] = -0.452541624988572,
    [20] = -0.430952063220271,
    [23] = -0.420535274239504,
    [30] = -0.401039455921266,
    [40] = -0.380599943010585,
    -- any higher falloff parameters are going to have a bad time being the base of a power function
}
function StunCurveInflection(falloff)
    local below, above = 1, 99999
    local belowValue, aboveValue = inflections[1], 0
    for k, v in inflections do
        if k < falloff then
            if k > below then
                below = k
                belowValue = v
            end
        else
            if k < above then
                above = k
                aboveValue = v
            end
        end
    end
    -- logarithmic interpolation
    local dx = math.log(falloff / below)
    local dy = math.log(above / below)
    local infl = dx / dy * (aboveValue - belowValue) + belowValue
    return infl
end
StunCurveInflection = Memoize(StunCurveInflection, inflections)
function StunCurveScale(falloff)
    return RawStunCurve(falloff, StunCurveInflection(falloff))
end
StunCurveScale = Memoize(StunCurveScale)
function StunModifier(falloff, resistance)
    -- if falloff is below 1, the curve actually begins to reverse its shape
    local inflection = StunCurveInflection(falloff)
    return RawStunCurve(falloff, resistance + inflection) / StunCurveScale(falloff)
end
StunModifier = Memoize(StunModifier)
function StunTime(dur, falloff, resistance)
    return dur * StunModifier(falloff, resistance)
end
```

Garanas:
@Tagada14 we need to make a call on two things, one of them being whether we want curve-based damage, such as the suggestion of @Hdt80bro 

And, even if it would make some fights awkward - for consistency, I think any unit protected by some shield should not be stunnable. After all - they are protected by a shield. It'd add a bit of depth to the game, and give units with a shield a clear buff (that they lack when gaining vet, as the shield isn't taken into account there).

Tagada14:
Curve-based stun duration would be nice to have but it's not a necessity. Units with personal shields getting stunned are IMO a must-have. It would break balance a lot and in my opinion be more unintuitive (You played zoomed out and you see 3/4 of your units get stunned while a couple of tanks are not, only after a couple of seconds do you remember that those tanks are obsidian so they have shields thus they are not getting stunned)

lua/sim/DefaultDamage.lua
```diff
@@ -8,24 +8,119 @@
...
--- cache for performance
+local Entity = import('/lua/sim/Entity.lua').Entity
+
+--- A cache of categories for stun allow / disallow strings
+local CategoriesCache = { }
+
+--- A dummy entity that is used to pass over stun parameters via stun instances
+---@class StunEntity
+---@field CategoriesAllowed Categories
+---@field CategoriesDisallowed Categories | boolean
```
Hdt80bro, line 23: We now have have a proper `EntityCategory` to use

lua/sim/DefaultDamage.lua
```diff
@@ -8,24 +8,119 @@
...
 local EntityGetPositionXYZ = _G.moho.entity_methods.GetPositionXYZ
 
---- Performs damage over time on a unit.
+--- Applies damage over time on a single target
+---@param instigator Unit | Projectile | Weapon | Entity
+---@param unit Unit         # Unit to apply damage instances to
+---@param pulses number     # Number of damage instances
+---@param pulseTime number  # Time between damage instances
+---@param damage number     # Damage amount
+---@param damType string    # Damage type such as 'Normal' or 'Force'
```
Hdt80bro, line 118: We now have a `DamageType` to use (same for `AreaDoTThread`'s parameter)

lua/sim/DefaultDamage.lua
```diff
@@ -8,24 +8,119 @@
...
+---@param target Unit               # Target to stun
+---@param duration number           # Duration of the stun, in seconds
+---@param allowed string | nil      # String representation of allowed categories, nil allows all units
+---@param disallowed string | nil   # String representation of disallowed categories, nil disallows no units
+---@param army number               # Army where the stun damage originates from
+function ApplyStun (target, position, duration, allowed, disallowed, army)
+    StunEntityInstances[army]:SetAllowedCategories(allowed)
+    StunEntityInstances[army]:SetDisallowedCategories(disallowed)
+
+    Damage(StunEntityInstances[army], position, target, 10 * duration, 'Stun')
```
Hdt80bro, line 102: I know I'm a sucker for optimization, but we could store `StunEntityInstances[army]` into a local variable. I suspect many tiny things like this will add up to helping performance.

lua/sim/DefaultDamage.lua
```diff
@@ -8,24 +8,119 @@
...
+
+        -- if no categories given, we apply to all units
+        if not cats then
+            self.CategoriesDisallowed = categories.ALLUNITS
+            return
+        end
+
+        local cached = CategoriesCache[cats]
+        if cached then
+            self.CategoriesAllowed = cached
```
Hdt80bro, line 42: You need a `return` in here to actually make the cache useful!

lua/sim/DefaultDamage.lua
```diff
@@ -8,24 +8,119 @@
...
+
+        -- if no categories given, we do not exclude units
+        if not cats then
+            self.CategoriesDisallowed = false
+            return
+        end
+
+        local cached = CategoriesCache[cats]
+        if cached then
+            self.CategoriesDisallowed = cached
```
Hdt80bro, line 63: Same, need a `return`!

units/URL0303/URL0303_Script.lua
```diff
@@ -80,28 +81,19 @@ URL0303 = Class(CWalkingLandUnit) {
...
-        -- Now handle our special buff and FX
-        local original_bp = table.deepcopy(self:GetBlueprint().Buffs)
-        local bp
-        for k, v in original_bp do
-            if v.Add.OnDeath then
-                bp = v
-            end
-            if self.ChargingInProgress then
-                bp.Duration = bp.DurationWhenCharging
-            end
```
Hdt80bro, line 96: Thank you so much, this looked awful!

Hdt80bro:
@Garanas I think that in order to get this wonderful change into the works, we need to table the ancillary discussions about shield stunning and curve flattening for another time. Those are separate things we can come back to later if we'd like. For now, I think that means leaving those two behaviors alone in this PR--which, I know... means re-adding several inefficient area buffs to get that tiered stun effect. Just think of it as something to motivate a future solution, or something.

Garanas:
I'll pick this up when I'm back! I'm not looking forward to adding back the inefficient area buffs - @Tagada14 could you discuss with the balance team whether stuns should or should not pierce shields? The old implementation allowed them to pierce shields, this implementation prevents that for most shields, except for personal shields.

Penguin5:
I think it intuitively makes sense for shields to block stuns coming from outside of the shields, and it seems like a nice feature for increased depth and strategic potential.  Also, if necessary, I figure that units with shields could be rebalanced some to accommodate such a change.

Hdt80bro:
At this point, I doubt the balance conversation will be resolved. So we need to remove all behavioral discrepancies this PR produces as best as possible and somehow convince the balance team that dome shields blocking stunning is a good idea.

Garanas:
For now, this PR is in limbo.

RabidPope:
Do we really need to update all the units at once? Would it be easier to get this through if it was just the t1 units, mainly the medusa? It gets used in any game that has a cybran player, gives up a large chunk of dps to have a stun that does not work half the time. A flat 2 seconds stun on t1 and t2 makes sense. Its on the lower side of its stun duration and will work when it should.

Tagada14:
In my opinion, stuns should not be able to pierce Dome shields (eg. Static shield generators, mobile shields) but they should stun units with personal shields. I will discuss it with the rest of the team as well.

Strogoo:
изображение

Ok, so to make it penetrates shields we need some AOE thing that ignores shields and applies damage to units at the same time. Actually, this is how allies aoe damage works right now. What if we use DamageArea() with dummy (or any) unit from opposite team as `instigator` and with `damageFriendly = true`?

Garanas:
I have not thought of trying that flag, I'll test it.

MrRowey:
After Speaking with Jip We are closing this one for now to deal with at a later date

lL1l1:
@Garanas Stunning through shields is now possible with `ShieldAbsorptionValues`. Should be as simple as setting the absorption value for the stun damage type to 0 for the default and static shield absorption types.

PR #4015, opened 2022-07-04
Re: Remove redundant functions

Hdt80bro:
Everything's better the second time around!

### DIFF
--- a/lua/aeonunits.lua
+++ b/lua/aeonunits.lua
@@ -71,19 +71,7 @@ AFactoryUnit = Class(FactoryUnit) {
 ---------------------------------------------------------------
 --  AIR STRUCTURES
 ---------------------------------------------------------------
-AAirFactoryUnit = Class(AirFactoryUnit) {
-    StartBuildFx = function(self, unitBeingBuilt)
-        AFactoryUnit.StartBuildFx(self, unitBeingBuilt)
-    end,
-
-    OnPaused = function(self)
-        AFactoryUnit.OnPaused(self)
-    end,
-
-    OnUnpaused = function(self)
-        AFactoryUnit.OnUnpaused(self)
-    end,
-}
+AAirFactoryUnit = Class(AirFactoryUnit, AFactoryUnit) {}
 
 ---------------------------------------------------------------
 --  AIR UNITS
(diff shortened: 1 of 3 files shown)


Hdt80bro:
To continue the conversation, `defaultunit.lua`
```lua
-- AIR FACTORY UNITS
AirFactoryUnit = Class(FactoryUnit) {}

-- LAND FACTORY UNITS
LandFactoryUnit = Class(FactoryUnit) {}

SeaFactoryUnit = Class(FactoryUnit) {
    DestroyUnitBeingBuilt = function(self)
        if self.UnitBeingBuilt and not self.UnitBeingBuilt.Dead and self.UnitBeingBuilt:GetFractionComplete() < 1 then
            self.UnitBeingBuilt:Destroy()
        end
    end,
}
```
matches the structure of `aeonunits.lua`
```lua
AAirFactoryUnit = Class(AirFactoryUnit, AFactoryUnit) {}

ALandFactoryUnit = Class(LandFactoryUnit, AFactoryUnit) {}

ASeaFactoryUnit = Class(SeaFactoryUnit, AFactoryUnit) {
    StartBuildFx = function(self, unitBeingBuilt)
        local thread = self:ForkThread(CreateAeonFactoryBuildingEffects, unitBeingBuilt, self.BuildEffectBones, 'Attachpoint01', self.BuildEffectsBag)
        unitBeingBuilt.Trash:Add(thread)
    end,
}
```
where now the Aeon factory units are using multiple inheritance from their faction's shared factory unit and their layer's default factory unit. Land and air behave the same, and navy does something a little different in each. Aeon is unique is being able to completely encapsulate their build effects into a helper class we can inherit, since all of the layer's factories have the same build effects.

`seraphimunits.lua` does something similar, but those classes aren't empty because they add layer-specific build effects, so it's quite useful there. The build effects in `terranunits.lua` are a little simpler so they just added all the logic directly to the factory classes without an intermediate helper class; same with `cybranunits.lua` since it's pretty uniform buildeffects-wise with the buildbots.

There's definitely repeated common code we could pull out of the rest of the factions' factories, but only for two of the three layers:
 * Seraphim factories have common build rotors for Land and Naval factories.
 * Cybran factories have common build manipulators for Land and Air factories.
 * UEF factories have common build arms for Air and Naval factories. All layers also have the same `CreateBuildEffects` method too, which would make `TLandFactoryUnit` empty if pulled into a common class.

Garanas:
The base branch was off - my mistake.

PR #4016, opened 2022-07-04
Re-implement game results

Garanas:
Re-implements the code that manages the game results that are send to the server. A few big changes:
 - The brain is no longer responsible for sending game results
 - The file `/lua/victory.lua` is now `/lua/sim/matchstate.lua` and is responsible for sending game results
 - The limit of the number of game results to send has been removed
 - The 'score' game result is removed entirely

Throughout the game we scan for the conditions of the game mode. We'll take assassination as an example. For that game mode the command unit needs to be alive, if it no longer is then the game mode should end. We scan the brains over time to see if they meet this requirement. 

For each brain scanned there is a small delay. If a brain is considered to be defeated (no units alive that are required for the game mode) then it rechecks all of the brains again to ensure that we didn't miss any cascading effects. Once we find no additional defeated brains we flag the considered brains as defeated, process the fact that the brain is defeated and send their game results via the sync to the server.

Once we've dealt with brains that are defeated we check if all the remaining brains agreed to a draw. If that is the case then we consider the brains to be in a 'drawn' state, process that fact and send their game results via the sync to the server. The game ends shortly after.

If not all brains agree with a draw we check if all the remaining brains are part of the same team, and they all agreed to an allied victory. If both of these conditions are true we process the fact that they are all victorious and send their game results via the sync to the server. The game ends shortly after.

If no brains remain (all brains are defeated) then the game ends shortly after.

One particular issue of the previous implementation is the check to `unit.Dead` instead of `IsDestroyed(unit)`. The flag `unit.Dead` may not have been set yet, even though the unit is considered to be destroyed. This is likely due to the execution order in Lua. See also https://github.com/FAForever/fa/pull/3794 that had a similar issue for build effects. 

Relevant quote of @Askaholic from Zulip about how the server operates:

```
It's important to know that 'victory' and 'defeat' are actually misnomers, as they don't mean 'you won the game' or 'you lost the game' but rather 'your ACU survived' and 'your ACU died'.

I think of the processing as happening in 3 stages:

Collect all of the individual reports for each army. Every player who is still connected to the game will report a result for each army when it dies.
Resolve all of the individual army results. So if 3 players reported 'defeat' for army 1 and one player reported 'victory' for that army, the army result will be resolved as 'defeat'. This is the step where we try to determine who's ACU is dead and who's ACU is still alive.
Resolve the team results by looking at the resolved army results. Here we check which team still has ACU's alive if any and make a determination about the game as a whole. If no ACU's are alive then the game is a draw, if only one team has ACU's alive then that team wins, otherwise if both teams have ACU's alive with 'VICTORY' results then we can't determine a winner. In the very rare case that the players used the diplomacy menu to end the game in a draw, all players in the game need to either be defeated or have a 'draw' army result.

Also note that the only results the server looks at are victory, defeat, draw, and mutual_draw. Anything not listed in this enum is ignored: https://github.com/FAForever/server/blob/a13889fe079d8c478a1841ad5f60e5147f3bb557/server/games/game_results.py#L25. That means the score result types are completely ignored.

That's because those 'score' results were supposed to be used to tell us how many ACU kills an army got. (Why they aren't called 'kills' instead of 'score' is beyond me)

Step 2 above happens here: https://github.com/FAForever/server/blob/a13889fe079d8c478a1841ad5f60e5147f3bb557/server/games/game_results.py#L126

Step 3 happens here: https://github.com/FAForever/server/blob/a13889fe079d8c478a1841ad5f60e5147f3bb557/server/games/game_results.py#L248
```

### DIFF
--- a/lua/sim/MatchState.lua
+++ b/lua/sim/MatchState.lua
@@ -30,7 +30,7 @@ local function CollectDefeatedBrains(aliveBrains, condition, delay)
         if criticalUnits then
             -- critical units found, make sure they all exist properly
             local oneCriticalUnitAlive = false
-            for k, unit in criticalUnits do
+            for _, unit in criticalUnits do
                 if (not IsDestroyed(unit)) and (unit:GetFractionComplete() == 1) then
                     oneCriticalUnitAlive = true
                     break
@@ -90,7 +90,7 @@ local function MatchStateThread()
 
     -- consider all non-civilian brains to be alive and kicking
     local aliveBrains = { }
-    for k, brain in ArmyBrains do
+    for _, brain in ArmyBrains do
         local index = brain:GetArmyIndex()
         if not ArmyIsCivilian(index) then
             aliveBrains[index] = brain
@@ -126,7 +126,7 @@ local function MatchStateThread()
                 brain:OnDefeat()
 
                 -- communicate to the server that this brain has been defeated
-                table.insert(Sync.GameResult, { k, string.format("%s %i", 'defeat', -10) })
+                table.insert(Sync.GameResult, { k, "defeat -10" })
 
                 -- stop considering it a brain that is still alive
                 aliveBrains[k] = nil
@@ -139,7 +139,7 @@ local function MatchStateThread()
 
             -- check for draw
             local draw = true
-            for k, brain in aliveBrains do
+            for _, brain in aliveBrains do
                 draw = draw and brain.OfferingDraw
             end
 
@@ -153,7 +153,7 @@ local function MatchStateThread()
                     brain:OnDraw()
 
                     -- communicate to the server that this brain has been defeated
-                    table.insert(Sync.GameResult, { k, string.format("%s %i", 'draw', 0) })
+                    table.insert(Sync.GameResult, { k, "draw 0" })
 
                     -- stop considering it a brain that is still alive
                     aliveBrains[k] = nil
@@ -165,7 +165,7 @@ local function MatchStateThread()
 
             -- check for win
             local win = true 
-            for k, brain in aliveBrains do
+            for _, brain in aliveBrains do
                 win = win and brain.RequestingAlliedVictory
             end
 
@@ -185,7 +185,7 @@ local function MatchStateThread()
                     brain:OnVictory()
 
                     -- communicate to the server that this brain has been defeated
-                    table.insert(Sync.GameResult, { k, string.format("%s %i", 'victory', 10) })
+                    table.insert(Sync.GameResult, { k, "victory 10" })
 
                     -- stop considering it a brain that is still alive
                     aliveBrains[k] = nil
--- a/lua/sim/MatchState.lua
+++ b/lua/sim/MatchState.lua
@@ -26,15 +26,25 @@ end
 local function CollectDefeatedBrains(aliveBrains, condition, delay)
     local defeatedBrains = { }
     for k, brain in aliveBrains do
-
-        local criticalUnits = brain:GetListOfUnits(brain, condition)
-        if criticalUnits then 
-            for k, unit in criticalUnits do 
-                if not IsDestroyed(unit) and unit:GetFractionComplete() == 1 then
-                    defeatedBrains[k] = brain 
+        local criticalUnits = brain:GetListOfUnits(condition)
+        if criticalUnits then
+            -- critical units found, make sure they all exist properly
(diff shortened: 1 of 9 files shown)


Garanas:
This change is a result of a discussion on [Zulip](https://faforever.zulipchat.com/#narrow/stream/203478-general/topic/Incorrect.20draw) and a two hour long talk with Sheikah 😄

Garanas:
Re-implements the code that manages the game results that are send to the server. A few big changes:
 - The brain is no longer responsible for sending game results
 - The file `/lua/victory.lua` is now `/lua/sim/matchstate.lua` and is responsible for sending game results
 - The limit of the number of game results to send has been removed
 - The 'score' game result is removed entirely

Throughout the game we scan for the conditions of the game mode. We'll take assassination as an example. For that game mode the command unit needs to be alive, if it no longer is then the game mode should end. We scan the brains over time to see if they meet this requirement. 

For each brain scanned there is a small delay. If a brain is considered to be defeated (no units alive that are required for the game mode) then it rechecks all of the brains again to ensure that we didn't miss any cascading effects. Once we find no additional defeated brains we flag the considered brains as defeated, process the fact that the brain is defeated and send their game results via the sync to the server.

Once we've dealt with brains that are defeated we check if all the remaining brains agreed to a draw. If that is the case then we consider the brains to be in a 'drawn' state, process that fact and send their game results via the sync to the server. The game ends shortly after.

If not all brains agree with a draw we check if all the remaining brains are part of the same team, and they all agreed to an allied victory. If both of these conditions are true we process the fact that they are all victorious and send their game results via the sync to the server. The game ends shortly after.

If no brains remain (all brains are defeated) then the game ends shortly after.

One particular issue of the previous implementation is the check to `unit.Dead` instead of `IsDestroyed(unit)`. The flag `unit.Dead` may not have been set yet, even though the unit is considered to be destroyed. This is likely due to the execution order in Lua. See also https://github.com/FAForever/fa/pull/3794 that had a similar issue for build effects.

lua/sim/MatchState.lua
```diff
@@ -0,0 +1,209 @@
...
+---@param delay number          # Delay between each brain to spread the load over various ticks
+---@return AIBrain[]            # Table of brains that are considered defeated, can be empty
+local function CollectDefeatedBrains(aliveBrains, condition, delay)
+    local defeatedBrains = { }
+    for k, brain in aliveBrains do
+        local criticalUnits = brain:GetListOfUnits(condition)
+        if criticalUnits then
+            -- critical units found, make sure they all exist properly
+            local oneCriticalUnitAlive = false
+            for k, unit in criticalUnits do
```
Askaholic, line 33: `k` not used in the loop? Also using `k` in the outer loop so this is a bit confusing.

```suggestion
            for _, unit in criticalUnits do
```

lua/sim/MatchState.lua
```diff
@@ -0,0 +1,209 @@
...
+    -- determine game conditions
+    local condition = Conditions[ScenarioInfo.Options.Victory]
+    if not condition and not (ScenarioInfo.Options.Victory == 'sandbox') then
+        SPEW("Unknown victory condition supplied: " .. ScenarioInfo.Options.Victory .. ", victory condition defaults to sandbox.")
+        return
+    end
+
+    -- consider all non-civilian brains to be alive and kicking
+    local aliveBrains = { }
+    for k, brain in ArmyBrains do
```
Askaholic, line 93: `k` not used in the loop?

lua/sim/MatchState.lua
```diff
@@ -0,0 +1,209 @@
...
+            end
+        end
+
+        -- loop through the brains that are still alive to check for alliance differences
+
+        if table.getsize(aliveBrains) > 0 then 
+
+            -- check for draw
+            local draw = true
+            for k, brain in aliveBrains do
```
Askaholic, line 142: `k` not used in the loop?

lua/sim/MatchState.lua
```diff
@@ -0,0 +1,209 @@
...
+                    aliveBrains[k] = nil
+                end
+
+                CallEndGame()
+                break
+            end
+
+            -- check for win
+            local win = true 
+            for k, brain in aliveBrains do
```
Askaholic, line 168: `k` not used in the loop?

Askaholic (commented):
Seems OK, but will need to test it with the server to be able to tell if anything broke.

PR #4009, opened 2022-06-26
Annotate campaign related functions pt. 2

Garanas:

PR #4018, opened 2022-07-04
Improved Mod Manager UI

HUSSAR-mtrela:
Improved UI and added new features to the Mod Manager:

- mod sorting
- mod searching
- save/load a list of favorite mods
- compact/expanded view of mod list
- links to mod source/websites
- added optional styling arguments to tooltip functions

Mod Manager in compact view:
image

Mod Manager in expanded view: 
image

Searching for mods with a "black" keyword:
image

Sorting mods by their name:
image

Save Favorite Mods:
image

Load Favorite Mods:
image

Mod Manager show:
-  links to mod info website (if mod_info.lua has url setting)
-  links to github website (if mod_info.lua has github setting)

image


lua/ui/game/tooltip.lua
```diff
@@ -20,39 +20,42 @@ local Keymapping = import('/lua/keymap/defaultKeyMap.lua').defaultKeyMap
...
 local createThread = false
 
-function CreateMouseoverDisplay(parent, ID, delay, extendedBool, width, forced)
-
-    -- -- values used throughout the function
-
+-- creates a tooltip box from ID table and with optional parameters
+-- @param - ID is a table, e.g. { text = 'tooltip header', body = 'tooltip description' } 
+-- @param - extendedBool indicates whether to just create tooltip header or also tooltip description
+-- @param - width is optional width of tooltip or it is auto calculated based on length of header/description
```
Garanas, line 26: The format of the new annotations is `--- @param name type description`, as an example:
`---@param width number Optional width of tooltip or it is auto calculated based on length of header/description`.

See also: https://github.com/FAForever/fa/pull/4009/files as an example

lua/ui/game/tooltip.lua
```diff
@@ -193,34 +218,43 @@ function CreateExtendedToolTip(parent, text, desc, width)
...
         local tempTable = false
 
         if desc then
-            tooltip.desc[1] = UIUtil.CreateText(tooltip, "", 12, UIUtil.bodyFont)
+            tooltip.desc[1] = UIUtil.CreateText(tooltip, "", descFontSize, UIUtil.bodyFont)
             tooltip.desc[1].Width:Set(tooltip.Width)
             if text == nil then
-                tooltip.desc[1].Top:Set(tooltip.Top)
-                tooltip.desc[1].Left:Set(tooltip.Left)
+--                tooltip.desc[1].Top:Set(tooltip.Top)
```
Garanas, line 250: Prefer to remove code instead of commenting it

Garanas (commented):
The difference in your screenshots (where the star icon becomes green) doesn't happen:

image

There's some empty space over at the end, we can take that out

image

Great work in general - very exciting!

lua/ui/game/tooltip.lua
HUSSAR-mtrela: ok, I will correct API comments

Garanas:
Done 👍

HUSSAR-mtrela:
@Garanas I fixed all issues that you found:
- validating domain of URLs before calling OpenURL()
- styling of favorite icon when Random faction is selected
- formatting of API comments in tooltip.lua
- removed dead/commented out code

Garanas (changes requested):
image

The list count doesn't adjust when you toggle filters on / off

image

When you toggle on the favorite mods the list resets, and you end up at the top again. That feels odd

The default color of the stars should be darker, to have a significant difference when you've chosen a random faction. I tried `ff434343`, that works good

@HUSSAR-mtrela Do you think you can find the time to do this in the next few days? I'd like to include this with the release, but for that I'd need to merge it sooner than later 😄

HUSSAR-mtrela:
@Garanas alright, I fixed scroll bar when mods are filtered and I used your color for unchecked favorite toggle. Also, I fixed counter for blacklisted mods and mods missing dependency. 

However, I did not change resting of mod list to top after toggling on the favorite mods because the same happens when filtering mods since both filtering mods and loading favorite mods are using the same function for refreshing mods list. Changing this behavior would be risky and it would add additional complexity. Ideally, this change would be much easier done when I replace (in another PR/release) the custom scroll control used in the Mod Manager with existing Grid control which has built-in scrolling.

Garanas:
Sounds good - and it is looking good. Great work!

4z0t:
Really like new look, however:
1. I'd really appreciate if it is done in new style, so, there is a class that handles it, lines  has class too and so on
2. Use Layouter instead of LayoutHelpers
3. Star's background doesnt match with ingame one 
image

lua/ui/lobby/ModsManager.lua
```diff
@@ -127,18 +162,19 @@ local modsTags = {
 function CreateDialog(parent, isHost, availableMods, saveBehaviour)
```
4z0t, line 162: must be a Class

lua/ui/lobby/ModsManager.lua
```diff
@@ -155,31 +191,109 @@ function CreateDialog(parent, isHost, availableMods, saveBehaviour)
     LayoutHelpers.AtHorizontalCenterIn(subtitle, dialogContent, 0)
     LayoutHelpers.AtTopIn(subtitle, dialogContent, 26)
 
+    CreateSortComboBox()
+
+    -- creating a toggle for activating favorite mods
+    local favoriteModsToggle = CreateFavoriteButton(dialogContent, 25, false, true)
+    LayoutHelpers.AtLeftIn(favoriteModsToggle.Icon, dialogContent, 18)
```
4z0t, line 198: forget LayoutHelpers, use Layouter

Garanas: This PR builds on legacy code, it is not a reimplementation. For now, this is fine

Issue #2342, opened 2017-12-26
GC tractor claw errors


speed2CZ:
```
WARNING: Error running lua script: Failed to attach entity url0303 to ual0401 on bone 21
         stack traceback:
         	[C]: in function `AttachBoneTo'
         	...\users\odstr\documents\github\fa\lua\aeonweapons.lua(132): in function <...\users\odstr\documents\github\fa\lua\aeonweapons.lua:112>
```
Seems like multiple GCs trying to steal each other's units that are being sucked. Sometimes they target the unit with beam as well.

Fastest way to reproduce is to spawn counple of GCs and some T3 units and still errors will slowly start popping out.

FAETHER:
Is that why GC's sometimes refuse to use their magnets? Or that bug has been already fixed?

PR #4020, opened 2022-07-05
Adjust behavior of tractor beams

Garanas:
This pull request tackles:
 - The units being sucked up now leave wreckages
 - The tractor beam weapons actually properly function, as the previous implementation could cause them to not work at all
 
And it pimps the behavior slightly:
 - More effects to give it a bigger 'oof' factor
 - Unit that is being sucked up rotates randomly, like the Aeon Power Generators do. Gives a feeling of 'helplessness' when the unit occasionally manages to aim and fire a shot or two at the colossus

A future pull request is required to pimp the behavior even further, where we use a projectile instead of a slider. That prevents the unit from suddenly moving all over the place when the colossus rotates.

https://user-images.githubusercontent.com/15778155/177477541-bd580f3f-87a2-430e-a6e9-0825886a1de9.mp4


MrRowey:
#2342

Garanas:
To fix that particular error I'd need to re-implement it from the ground up.

Garanas:
Closed in favor of #4040

PR #4021, opened 2022-07-05
Adjust annotation class hierarchy of engine classes

Garanas:


Garanas:
Classes are based on printing the functions when they are converted into a Lua class, see also:
[class-data-dump.txt](https://github.com/FAForever/fa/files/9062818/class-data-dump.txt)

PR #4022, opened 2022-07-06
Improve intel management

Hdt80bro:
Changes intel from checking the energy levels every half second to only using the AIBrain's `EnergyDependingEntity` callbacks. This also involves some cooperation from the enhancement code, since some enhancements add or create intel traits (it also keeps the radar flickering synced up to the enhancement completion). ~~In order to not change the radar flickering behavior, I ended up partially restoring that half-second loop since the flickering was originally created by an interaction between it and the low energy level disabler. This does mean we now have freedom over the flicker duty cycle, which I've taken the liberty of implementing via `Blueprint.Intel.DeactivateTime` (to counterpart `ReactivateTime`).~~ *Intel no longer flickers and `ReactivationTime` is used as a reactivation delay after the player stops power stalling*

A unit's `IntelDisables` are now only allocated for what they need; it is no longer the case that all units create the table with all fields, despite most units having nothing to do with intel, and even those that do not using all of the fields anyway. You will find that `IntelDisables` is now very often `nil`.

There may have been a bit of refactoring as well...


lua/sim/Unit.lua
```diff
@@ -2132,32 +2175,22 @@ Unit = Class(moho.unit_methods) {
...
             -- Values can be setting up manually via bp.
-            if bp.VeteranMass then
-                self.Sync.manualVeterancy = {
-                    [1] = bp.VeteranMass[1],
-                    [2] = bp.VeteranMass[1] + bp.VeteranMass[2],
-                    [3] = bp.VeteranMass[1] + bp.VeteranMass[2] + bp.VeteranMass[3],
-                    [4] = bp.VeteranMass[1] + bp.VeteranMass[2] + bp.VeteranMass[3] + bp.VeteranMass[4],
-                    [5] = bp.VeteranMass[1] + bp.VeteranMass[2] + bp.VeteranMass[3] + bp.VeteranMass[4] + bp.VeteranMass[5],
-                }
+            local vetMass = bp.VeteranMass
```
Garanas, line 2178: One of the other refactorings spotted

lua/sim/Unit.lua
```diff
@@ -2641,41 +2674,43 @@ Unit = Class(moho.unit_methods) {
     -- present to zero, and when going from zero disablers to one.
 
     DisableUnitIntel = function(self, disabler, intel)
+        if not self.IntelDisables then
+            self.IntelDisables = {}
+        end
         local function DisableOneIntel(disabler, intel)
```
Garanas, line 2680: Best to not have this has a function inside a function - because of that it creates a new closure every time `DisableUnitIntel` is called

lua/sim/Unit.lua
```diff
@@ -2798,78 +2828,32 @@ Unit = Class(moho.unit_methods) {
...
-            local enh = self.Blueprint.Enhancements
-            if enh then
-                for k, v in enh do
-                    if self:HasEnhancement(k) and v.MaintenanceConsumptionPerSecondEnergy and v.MaintenanceConsumptionPerSecondEnergy > 0 then
-                        bpVal = v.MaintenanceConsumptionPerSecondEnergy
-                        break
-                    end
-                end
-            end
+    IntelFlickerThread = function(self)
```
Garanas, line 2831: I don't think intel should flicker - just like shields, if you have no power it should be just turned off.

Garanas (commented):
Looking good - I'll review it more in-depth this weekend.

lua/sim/Unit.lua
Hdt80bro: Ah, good catch.

lua/sim/Unit.lua
Hdt80bro: I was trying my best to preserve current behavior--I think lots of players are used to radar flickering as a visual indicator of a power stall. Note that it will also affect cloak and stealth flickering.
It's built into the blueprint with `ReactivationTime`--do you want to remove that altogether?

Garanas: Reactivation time is fine, but it should run once you have sufficient power again. The flickering makes the game more difficult to read and to understand, same as it did for the shields (where it would flicker on and catches a tactical missile).

Hdt80bro: Alright, I can remove it. Not dealing with flickering simplifies a couple other things too.

Garanas:
I'll pick this up when I'm back 👍

lua/sim/Unit.lua
```diff
@@ -2810,78 +2864,39 @@ Unit = Class(moho.unit_methods) {
...
-            end
-        end
-        if self.IntelThread then
-            self.IntelThread = nil
+    --- Create the intel reactivation thread, called when the player stops power stalling by the AIBrain.
+    --- This unit only gets added as an energy dependent entity to the AIBrain when the unit has intel.
+    ---@param self Unit
+    OnEnergyViable = function(self)
+        -- Not sure how the thread could already be active, since energy can only become viable after
+        -- it's been depleted, and that removes any threads. Good to be safe though.
```
Garanas, line 2885: The correct answer to this is 'mods'

lua/sim/Unit.lua
```diff
@@ -2193,32 +2244,22 @@ Unit = Class(moho.unit_methods) {
...
-                }
-                local defaultMult = techMultipliers[self.techCategory] or 2
+                local defaultMult = veterancyMultipliers[self.techCategory] or 2
 
                 self.Sync.myValue = math.max(math.floor(bp.Economy.BuildCostMass * (bp.VeteranMassMult or defaultMult)), 1)
             end
         end
 
-        self:EnableUnitIntel('NotInitialized', nil)
+        self:EnableUnitIntel('Construction')
```
Garanas, line 2262: This causes all fields to have at least an empty table.

Garanas (changes requested):
Took a lot longer than expected - apologies for that.

The pull request works, but there's an issue. I've added a routine to log the intel state of a unit, you can find it in the hotkeys menu by searching for `intel` . For each unit all fields have at least an empty table due to the call to enabling intel after construction. Do you reckon there's a way to prevent this from happening? As an example, an Aurora should have an empty intel table.

edit: as an example, the `IntelDisabler` table looks like this for an Aurora:

```
INFO:  - Cloak: table: 1DCBA758
INFO:  - CloakField: table: 1DCBAA78
INFO:  - Omni: table: 1DCBA348
INFO:  - Radar: table: 1DCBA1B8
INFO:  - RadarStealth: table: 1DCBAB90
INFO:  - RadarStealthField: table: 1DCBA488
INFO:  - Sonar: table: 1DCBAF28
INFO:  - SonarStealth: table: 1DCBA0A0
INFO:  - SonarStealthField: table: 1DCBAD98
INFO:  - Spoof: table: 1DCBA618
```

Garanas:
Fixed the issue I mentioned. There's one more though: when you change the script bits (for say, a radar) then it adds in all remaining intel types into the table. See also:

```
    OnScriptBitSet = function(self, bit)
        if bit == 0 then -- Shield toggle
            self:PlayUnitAmbientSound('ActiveLoop')
            self:EnableShield()
        elseif bit == 1 then -- Weapon toggle
            -- Amended in individual unit's script file
        elseif bit == 2 then -- Jamming toggle
            self:StopUnitAmbientSound('ActiveLoop')
            self:DisableUnitIntel('ToggleBit2', 'Jammer')
        elseif bit == 3 then -- Intel toggle
            self:StopUnitAmbientSound('ActiveLoop')
            self:DisableUnitIntel('ToggleBit3', 'RadarStealth')
            self:DisableUnitIntel('ToggleBit3', 'RadarStealthField')
            self:DisableUnitIntel('ToggleBit3', 'SonarStealth')
            self:DisableUnitIntel('ToggleBit3', 'SonarStealthField')
            self:DisableUnitIntel('ToggleBit3', 'Sonar')
            self:DisableUnitIntel('ToggleBit3', 'Omni')
            self:DisableUnitIntel('ToggleBit3', 'Cloak')
            self:DisableUnitIntel('ToggleBit3', 'CloakField') -- We really shouldn't use this. Cloak/Stealth fields are pretty busted
            self:DisableUnitIntel('ToggleBit3', 'Spoof')
            self:DisableUnitIntel('ToggleBit3', 'Jammer')
            self:DisableUnitIntel('ToggleBit3', 'Radar')
        elseif bit == 4 then -- Production toggle
            self:OnProductionPaused()
        elseif bit == 5 then -- Stealth toggle
            self:StopUnitAmbientSound('ActiveLoop')
            self:DisableUnitIntel('ToggleBit5', 'RadarStealth')
            self:DisableUnitIntel('ToggleBit5', 'RadarStealthField')
            self:DisableUnitIntel('ToggleBit5', 'SonarStealth')
            self:DisableUnitIntel('ToggleBit5', 'SonarStealthField')
```

Preferably, it wouldn't do that. Question is whether it is worth fixing?

Hdt80bro:
Well, it's very easy to fix. The question is whether it impacts mods dynamically enabling intel fields after the disable, as they then won't have the proper disables set on the field. This is already a problem for the construction disabler, but I think it's safe to presume that no radar field will be active while it's building.

Garanas:
The solution still expands the intel table to everything when disabling and enabling. It won't make it for this patch, maybe we need to rebuild how intel works from the ground up. What if we can solely enable / disable what the unit is able to produce (at any point in time)?

Garanas:
And there's also issues with animations and effects that trigger, even though the intel is disabled by say energy.

Garanas:
@Hdt80bro I am closing this, we'll need to find the time for a better solution someday.

PR #4024, opened 2022-07-08
Annotate engine/User

Hdt80bro:
Annotates `engine/User*` (and annotates a few things around it--sorry, I never can do just one thing, though it does make sense to also do annotation work on relevant code while it's right there). Specific things covered:
 * Fully documents all user-side classes inside of the `moho` table
   * Adds [`CDiscoveryService`](https://github.com/Hdt80bro/fa/blob/annotate-engine/user.lua/engine/User/CDiscoveryService.lua), which was hiding in the syntax errors of [`/engine/User.lua`](https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/User.lua#L1097-L1114)
   * Adds all remaining files in `/User` into the `/engine/moho.lua` table, and forms them into annotated classes
   * Adds a bunch of methods that were missing (in particular, `Destroy` was common) 
 * Created classes for things related to User function arguments, usually inside of a different but relevant file
   * `ArmiesTable` `ArmyInfo` something for GetArmiesTable() to return
   * `BuildTemplate` `BuildTemplateBuilding` `BuildQueue` gets used in setting units' queues
   * `Client` gets returned by GetSessionClients() 
   * `CommunicationData` used in broadcasting and sending messages
   * `EconData` returned from UserUnit.GetEconData()
   * `EconomyResourceValues` `EconomyTotals` used in GetEconomyTotals()
   * `GameData` used to launch the game
   * `GameOptions` <key, value> pairings of Scenario Options, used after the game launches
   * `KeyEvent` part of Edit callbacks
   * `MeshInfo` used in AddCommandFeedbackBlip()
   * `MissileInfo` returned from UserUnit.GetMissileInfo()
   * `ModInfo` mostly just `mod_info.lua` after it's gone through the mod utilities
   * `OrderInfo` holds information about the actual command cap
   * `PatchedDepositResult` from the binary-patched function to find deposits in an area
   * `Peer` holds raw connection data from the GetPeer functions
   * `PlayerData` part of the GameData, holds lobby options for the players/observers 
   * `RangeOverlay` information about a type of range ring
   * `RolloverInfo` returned from GetRolloverInfo() for the selected info and unit view
   * `ScenarioOption` `ScenarioOptionValue` used in the in-lobby option factory before the game launches
   * `SimCallback` one of the ways the Sim gets updated from the User end
   * `UIScenarioInfo` `UIScenarioConfiguration` mostly just `map_scenario.lua` after it's gone through map utilities
   * `WatchedGameData` `WatchedPlayerData` part of the lobby implementation

I opted to not add comments to functions that I can't meaningfully add to, like
```lua
---@return Cursor
function GetCursor() end
```
either where the function is self-explanatory enough that I'd just rewrite the function name and parameters as a comment, or where I don't know the function well enough to not end up doing that anyway.


lua/options/options.lua
```diff
@@ -1028,7 +1028,7 @@ options = {
 
                     aaoptions = GetAntiAliasingOptions()
 
-                    aamax = 0
+                    aahigh = 0
```
Garanas, line 1031: not sure why this was changed, but best to keep annotation pull requests about annotations. It makes them a lot easier to merge

Hdt80bro: It fixes a typo I saw and, 'just felt like fixing while I was there'. I really didn't want to take the time and effort to come back later to make a new branch and PR just for that. I *do* tend to have more of a breadth-first approach to things than is helpful to everyone else... I'm having a hard time keeping this PR to just the `/engine/User` files as it is.

It makes it so that the loop immediately below it properly defines `aahigh` for use in setting <https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/options/options.lua#L1066>. Looking at it, we don't even have an ultra-high fidelity preset, so that if-branch never gets called anyway (which... explains why this went unnoticed).

All in all, it's entirely moot in every regard possible on my part and I know you're right (and knew it when I did it too). I'll undo it.

lua/ui/lobby/data/gamedata.lua
```diff
@@ -1,11 +1,34 @@
+--- Watched version of `GameData` that can easily change with the lobby
+---@class WatchedGameData
```
hahn-kev, line 2: why don't you have this class inherit from `GameData` and just add the modified fields?

lua/ui/game/wlduiprovider.lua
```diff
@@ -3,11 +3,12 @@
 --* Author: Chris Blackwell
 --* Summary: Responds to wld UI events and show appropriate UI
 --*
---* Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
+--* Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
 --*****************************************************************************
 
 -- Functions exposed to lua
 
+---@class WldUIProvider
```
hahn-kev, line 11: you should indicate that this class inherits from `moho.WldUIProvider_methods`

lua/ui/lobby/data/gamedata.lua
Hdt80bro: Because... I didn't think of that. That's a good idea!

Garanas:
Freeze this pull request - otherwise it becomes too big to manage. Only process feedback at this point.

engine/User/CLobby.lua
```diff
@@ -2,90 +2,109 @@
...
+
 ---
---  void CLobby.ConnectToPeer(self,address,name,uid
-function CLobby:ConnectToPeer()
+---@param data CommunicationData
+function CLobby:BroadcastData(data)
+end
+
+---
+---@param address string | number
```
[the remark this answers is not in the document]
Hdt80bro: > Freeze this pull request - otherwise it becomes too big to manage. Only process feedback at this point.

So it *did* become too big to manage.
Good thing I was done (that's why I opened it for review)

Hdt80bro:
> Freeze this pull request - otherwise it becomes too big to manage. Only process feedback at this point.

So it *did* become too big to manage.
Good thing I was done! (that's why I opened it for review)

engine/User/CMauiControl.lua
```diff
@@ -2,126 +2,126 @@
...
----
---  Control:Destroy() -- destroy a control.
+--- Destroy a control
 function CMauiControl:Destroy()
 end
 
----hit testing will be skipped for this control
----@param recursive boolean?
+--- Hit testing will be skipped for this control
+---@param recursive? boolean
```
Garanas, line 27: defaults to true I believe

Hdt80bro:
Alright, this has been open long enough--I'm going to merge it in a day. Since it's just basic annotation stuff, it won't break anything and will allow us to further integrate some of the work into the code base. If people don't like it, we can change it. I think this is the approach we should be taking for annotations in general (think: Wikipedia contributions).

Garanas:
Good call - lets keep them smaller in the future than this though.

Hdt80bro:
Yeah... I'll try

PR #4037, opened 2022-07-10
Alternative, more performing vision entities

Garanas:
Reduces the number of table and field allocations to zero, with the exception of allocating the entity itself. The old vision marker is still available.

Todo:
 - [ ] Convert all existing vision markers to this new vision marker


Garanas:
@Spikey84 if you're interested in picking up something - then take this! You'll need to finalize the PR so that all units that use the old vision marker (usually via their projectiles) use the new vision marker. It has less table allocations and allows you to only set / adjust what is actually being used.

You can leave the old vision marker in other scripts, such as campaign related scripts. They either usually persist a lot longer, or are used in such a low quantity that the change is negligible. While at the same time the chance of introducing bugs is quite high, and it is difficult to test. 

Just look at the unit / projectile scripts, from the top of my head:

- Tech 1 UEF artillery (see its projectile)
- Tech 2 Aeon bomber (see its projectile)
- Tech 1 Seraphim scout (see its OnImpact)

To find others, you can `CTRL + SHIFT + F` to search for `VizMarker` through the entire repo, see where it shows up!

The Tech 3 Quantum Optics Facility is a bit more involved and its use is negligible again (there is usually only 1, if any in a game). You can skip that unit. And by all means - if the code has a bug, feel free to fix it as you see fit. Make sure to adjust the annotations if that is required.

PR #4038, opened 2022-07-10
Annotating of cAIBrain and AI related structures / functions

Garanas:


engine/Sim/CAiBrain.lua
```diff
@@ -68,37 +51,33 @@ end
...
+---@param location Position
 function CAiBrain:CanBuildStructureAt(blueprintID, location)
 end
 
---- Returns true in case terrain is not blocking weapon fire from attackPosition to targetPosition.
--- @param attackPosition  Table with position {x, y z}
--- @param targetPosition position Table with position {x, y z}
--- @param arcType Types: 'high', 'low', 'none'.
--- @return true/false
+--- Returns true if the terrain is **not*8 blocking weapon fire from the attack position to the target position
```
relent0r, line 60: This is actually the other way around, it returns true is terrain is blocking weapon fire. Its incorrect in the original documentation.

engine/Sim/CAiBrain.lua
```diff
@@ -44,19 +28,18 @@ end
...
---- Orders the builder to build a unit.
--- @param builder Unit, (engineer) to use.
--- @param blueprintID Unit's bp ID to build, example: 'ueb0101'.
--- @param buildLocation Table {x, z, 0}.
--- @param buildRelative: true/false. true = build coordinates are relative to the starting location, false = absolute coords
--- @return true/false
+--- Order a unit to build a structure
+---@param builder Unit
+---@param blueprintID string
+---@param buildLocation Position
```
relent0r, line 34: Can you add a small bit here to describe the position. It goes { x, z, rotation }.

engine/Sim/CAiBrain.lua
```diff
@@ -270,6 +269,12 @@ end
 -- @param needToBeIdle true/false Unit has to be idle (appears to be not functional).
 -- @param requireBuilt true/false defaults to false which excludes units that are NOT finished (appears to be not functional).
 -- @return tblUnits Table containing units.
+
+--- Returns a list of units that match the categories
+---@param category Categories
+---@param needToBeIdle boolean
+---@param requireBuilt boolean Appears to be not functional
+---@return Unit[]
```
relent0r, line 277: Don't quite understand why you changed to Unit[] indicating a list rather than the table that gets returned.

engine/Sim/CAiBrain.lua
Garanas: What parameter is that, the rotation?

engine/Sim/CAiBrain.lua
Garanas: It is syntactically the same - Unit[] is a 'list' in terms of that only the array section of a table is populated

engine/Sim/CAiBrain.lua
speed2CZ: There's no rotation, it's classic {x, y, z} position

lua/platoon.lua
```diff
@@ -23,6 +23,8 @@ local SPAI = import('/lua/ScenarioPlatoonAI.lua')
 --for sorian AI
 local SUtils = import('/lua/AI/sorianutilities.lua')
 
+---@alias PlatoonSquads 'Attack' | 'Artillery' | 'Guard' 'None' | 'Scout' | 'Support' | 'Unassigned'.
```
speed2CZ, line 26: missing | betwen Guard and None

engine/Sim/CAiBrain.lua
```diff
@@ -160,29 +139,45 @@ end
...
 
 --- Returns list of factories at location.
 -- @param location Table with location, it's not a position but location created by PBMAddBuildLocation function.
 -- @param radius Number in game units.
 -- @return tblUnits List of factories.
+
+--- Returns a list of factories at a location
+---@param location Position
+---@param radius number
+---@return FactoryUnit
```
speed2CZ, line 180: return type should be a list, also keeping the old description?

engine/Sim/CAiBrain.lua
```diff
@@ -294,12 +299,12 @@ end
...
--- @param category Unit's category, example: categories.TECH2 .
--- @param position Table with position {x, y, z}.
--- @param radius Number in game units.
--- @param alliance Types: 'Ally', 'Enemy', 'Neutral'.
--- @return Number.
+--- Returns the number of units around a position that match the categories
+---@param category Categories
+---@param position Position
+---@param radius number
+---@param alliance 'Ally' | 'Enemy'
```
speed2CZ, line 306: what about Neutral?

engine/Sim/CAiBrain.lua
relent0r: For this particular function its rotation. You'll notice in all the default code they will often statically set it as zero.
e.g `aiBrain:BuildStructure(eng, whatToBuild, {buildLocation[1], buildLocation[3], 0}, buildRelative)`

When I was writing custom engineer code I used the getsurfaceheight to populate the value thinking it was height and all the structures ended up rotating however many degrees the number was. I assumed that originally the game supported rotating structures when building like other rts but for whatever reason they disabled it.

Garanas: I'm going to test that out 😄 , that would be exciting!

engine/Sim/CAiBrain.lua
Garanas: Ah, yes - forgot about that!

engine/Sim/CAiBrain.lua
Garanas: I'll add it!

Hdt80bro:
In general, anywhere you have an annotation type of `Categories` (or `categories` as a typo) it should be replaced with the temporary type `CategorieType`. However, I'd recommend going ahead and using its actual type, `moho.EntityCategory` which has not been created yet. I'll prioritize the relevant files out of my PR into a new PR, so that reviewing and merging it is quicker.

Hdt80bro:
You are missing the `AddArmyStat(statname,val)` method

Garanas:
This particular PR requires a bit of help from one of the AI developers - I'll pick it up once I'm back.

Garanas:
It isn't complete yet, but decided to merge as other people have tried annotating the same file

PR #4040, opened 2022-07-11
Re-implement the tractor beams of the colossus

Garanas:
Completely re-implements the tractor beams, fixing all the critical issues with them. Similar to #4020, closes https://github.com/FAForever/fa/issues/2342 .

This pull request tackles:

- The units being sucked up now leave wreckages
- The tractor beam weapons actually properly function, as the previous implementation could cause them to not work at all

And it pimps the behavior slightly:

- More effects to give it a bigger 'oof' factor
- Unit that is being sucked up rotates randomly, like the Aeon Power Generators do. Gives a feeling of 'helplessness' when the unit occasionally manages to aim and fire a shot or two at the colossus

https://user-images.githubusercontent.com/15778155/178206449-965c7bb8-2e88-4aed-8625-b62d33e6bc13.mp4

### DIFF
--- a/effects/emitters/tractor_01_emit.bp
+++ b/effects/emitters/tractor_01_emit.bp
@@ -1,0 +1,155 @@
+EmitterBlueprint {
+	BlueprintId = 'aeon_gate_01',
+	Lifetime = -0.20,
+	Repeattime = 0.20,
+	TextureFramecount = 1.00,
+	Blendmode = 3.00,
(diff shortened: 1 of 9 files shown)


Garanas:
https://user-images.githubusercontent.com/15778155/178301025-f809186b-a536-4256-8301-fa47c7e60646.mp4
Due to a script error, Titan being cute

PR #4041, opened 2022-07-11
Add Core files

Hdt80bro:
Adds [`moho.EntityCategory`](https://github.com/Hdt80bro/fa/blob/add-core-files/engine/Core/EntityCategory.lua) and classes it to [`EntityCategory`](https://github.com/Hdt80bro/fa/blob/add-core-files/engine/BlueprintCategories.lua#L4), which replaces the temporary `CategorieType`

Moves `CSound` from `/engine/Sim/` to `/engine/Core/`

Adds the three `/engine/Core/` files to the `moho.lua` table

Adds the missing non-uppercase categories to the documented [`categories`](https://github.com/Hdt80bro/fa/blob/add-core-files/engine/BlueprintCategories.lua) (barring the unit name & file categories automatically added by the engine)


engine/Core/EntityCategory.lua
```diff
@@ -0,0 +1,26 @@
+---@declare-global
+---@class moho.EntityCategory
+local EntityCategory = {}
+
+--- Return the union between two categories
+---@param augend moho.EntityCategory
+---@param addend moho.EntityCategory
+---@return moho.EntityCategory
+function EntityCategory.__add(augend, addend)
```
hahn-kev, line 9: do these get detected by the extension properly when adding 2 categories together?

Hdt80bro: It does not. It still thinks that `CAT1 + CAT2` is a number.

engine/BlueprintCategories.lua
```diff
@@ -424,4 +450,29 @@ categories = {
 ---|"VERIFYMISSILEUI"
 ---|"VISIBLETORECON"
 ---|"VOLATILE"
----|"WALL"
\ No newline at end of file
+---|"WALL"
+---
+---| "AdvancedCoolingUpgrade"
```
Garanas, line 455: What are these?

engine/Core/EntityCategory.lua
hahn-kev: ok, in the next version of the extension we will be able to annotate these like this: `---@operator div(moho.EntityCategory): moho.EntityCategory` but it's not supported yet.

Hdt80bro: Right on!

engine/BlueprintCategories.lua
Hdt80bro: `'Cybran'` is generated by a typo for [Cybran PoW Facility](https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/XRC1502/XRC1502_unit.bp#L20)
The rest are generated by sACU presets adding their [enhancement names to the `CategoriesHash`](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/system/Blueprints.lua#L581) for some reason, then getting [unhashed back to 'Categories`](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/system/Blueprints.lua#L624), and thus getting fed back to the engine (since it can still modify blueprints at that point).
I'm pretty sure these are all unintentional, so I'll remove them.

Hdt80bro: The sACU enhancements' names *are* really confusing in an entity category context (e.g. the entity category `Overcharge` means "An entity with the Overcharge enhancement")--could they be redone or would that break compatibility? It'd also be better to remove the hardcoded unit names for robustness. For example, the expression to disable antiair land units uses `url0301_AntiAir` instead of `SUBCOMMANDER * NaniteMissileSystem` like some of the surrounding code, which I'd like to refactored to something like `PRESET_ENH_NaniteMissileSystem` (the `SUBCOMMANDER` category is redundant since it doesn't matter who has the enhancement, just that it has it).

Hdt80bro: *Aaaahhh I'm digging deeper into the weeds and want to redo how unit restrictions are set up, but I know I shouldn't*

Hdt80bro:
Oh, and this will require updating all pending PR's to use `EntityCategory` instead of `CategorieType` (any use of `Categories` should already be replaced). Actually, why have `Categories` then? It shouldn't ever be used, aside from its use to annotate `categories`.

PR #4047, opened 2022-07-14
Conform `/Engine/Sim/` to Moho

Hdt80bro:
Conforms documented Sim engine prototypes and Lua classes to what's inside Moho. Leaves `AiBrain` alone, as that is being handled separately.


engine/Sim/CManipulator.lua
```diff
@@ -1,22 +0,0 @@
----@declare-global
----@class moho.manipulator_methods
-local CManipulator = {}
```
Garanas, line 3: What is the idea of moving this to `IAniManipulator` ?

Hdt80bro: `IAniManipulator` is the engine name for the underlying class; since the rest of the files use the engine name, I updated this to match.

Garanas: ah, I see. I was not aware. The name felt strange 😄

Hdt80bro: Yeah, the engine names are a little... haphazard

PR #4048, opened 2022-07-15
Add color support

Hdt80bro:
`EnumColor` may want to instead be moved to a different file, rather than being shoved to the bottom.

### DIFF
--- a/engine/Core.lua
+++ b/engine/Core.lua
@@ -23,6 +23,8 @@
 ---@field x1 number
 ---@field y1 number
 
+---@alias Color string # `EnumColor` or hexcode like `'RrGgBb'`, or `'AaRrGgBb'` with transparency
+
 --- Constructs a rectangle, usually the first point is in the top-left corner and the second is in the bottom-right corner
 ---@param x0 number
 ---@param y0 number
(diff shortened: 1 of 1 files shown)


Garanas:
Great find - wasn't aware it was possible beyond the primary colors.

Garanas:
Great find - wasn't aware it was possible beyond the primary colors. I've added EnumColor as a type, what are your thoughts on that?

Hdt80bro:
I did that at first, but realized that it clogs up the type information, and for some reason adds the comment on `Color` to everything.
image
I think this is fine
image
Since hexcodes are strings, I figure the type information of `EnumColor` is already a subset of that, so we don't loose documentation.
However, if we could somehow specify a formatted string as a type, I'd be open to that. We could do something like
```
---@alias Color OpaqueColor | TransparentColor | EnumColor
---@alias OpaqueColor string["XXXXXX"] # somehow specify a string of length 6 containing only hex characters 
---@alias TransparentColor  string["XXXXXXXX"] # same, but for length 8
---@alias EnumColor # all 140 options...
```
Nevertheless, we still couldn't keep it from expanding since it's an `@alias` not a `@class`, and we couldn't make it a `@class` because those can't extend literals or aliases.

TL;DR I think it'd be better to leave it in the description comment because it's cleaner

PR #4049, opened 2022-07-16
Annotate `/Engine/Sim.lua` creation functions

Hdt80bro:


engine/Sim.lua
```diff
@@ -1096,13 +1234,13 @@ end
 function Warp(entity, location, orientation)
 end
 
---- TODO.
----@diagnostic disable-next-line: lowercase-global
+---
+---@return Entity
```
Garanas, line 1238: This doesn't return anything, it adds a reference to the _c_object to the specs provided

Hdt80bro: You're right, those should be parameters!

PR #4051, opened 2022-07-16
Annotate `/engine/Sim` army functions

Hdt80bro:


engine/Sim/Entity.lua
```diff
@@ -2,7 +2,7 @@
 ---@class moho.entity_methods
 local Entity = {}
 
----@alias Army number
+---@alias Army number | string
```
Garanas, line 5: There are functions that specifically expect a number, and those that specifically expect a string. We need to make a distinction between those

Hdt80bro: I saw that, but when I went through, I couldn't find any that require either specifically, aside from the Army intializers. I think `Army` behaves a lot more like `Bone` does--you can use either the army index or army name. I can do more testing though.

PR #4057, opened 2022-07-16
Annotate `engine/Sim.lua`

Hdt80bro:
Includes some changes in https://github.com/FAForever/fa/pull/4052 to comprehensively annotate `Sim.lua` (merge that one first)


engine/Sim.lua
```diff
@@ -172,7 +173,7 @@ end
 function CreateBuilderArmController(unit, turretBone, barrelBone, aimBone)
 end
 
---- Creates a collision detection manipulator, calls the function `self.OnAnimTerrainCollision(self, bone, x, y, z)` when a bone that is being watched collides with the terrain
+--- Creates a collision detection manipulator
```
Garanas, line 176: `calls the function `self.OnAnimTerrainCollision(self, bone, x, y, z)` when a bone that is being watched collides with the terrain` is good for understanding how this function is used in practice!

engine/Sim.lua
```diff
@@ -475,89 +476,92 @@ end
 function DamageRing(instigator, location, minRadius, maxRadius, damage, damageType, damageFriendly, damageSelf)
 end
 
---- Get DEBUG info for UI selection.
--- TODO.
+--- Gets DEBUG selection
+---@return Unit[]
```
Garanas, line 480: As the selection of each player is different, this sounds like something that can cause a desync. Mark it as `deprecated`

engine/Sim.lua
```diff
@@ -475,89 +476,92 @@ end
...
+---@param z number
+---@param sizeX number
+---@param sizeZ number
+---@param elevation number
+function FlattenMapRect(x, z, sizeX, sizeZ, elevation)
 end
 
---- Deletes sscouted icons from the target area.
--- If the area is in a radar range, it will switch back to default unscouted icons.
+--- Deletes scouted icons from the target area.
```
Garanas, line 548: `Removes all recon blips from the target area, if the area is in radar range it generates unseen recon blips`

Garanas (commented):
Looking good - keep it up!

engine/Sim.lua
Hdt80bro: I think it'd only cause a desync if used to synchronize code. It should be fine for debug purposes, or even if used in synchronous code as long as its result is sent to the other players. I'll add a warning to it nonetheless.

Issue #3888, opened 2022-05-24
Shield miscalculations


Penguin5:
https://replay.faforever.com/17121095
It appears that at 13:14, a com with ~2200 hp fully under an active parashield with 100% shield strength and sufficient energy to maintain the shield, dies when 2 notha shots hit the shield bubble of the parashield.

![image](https://user-images.githubusercontent.com/72904368/170110398-54cf3930-f2a2-4f1f-9910-726274b1cd94.png)
![image](https://user-images.githubusercontent.com/72904368/170111058-c4c6049b-69a9-43da-8ae9-e886bcb9d902.png)

Garanas:
Closing this issue as I consider it to be unfixable, neither from Lua or from an engine patch.

Issue #3153, opened 2020-08-09
Seraphim T3 MAA Lightning Tank Shoots Through Shields


ThomasHiatt1:
Like the title says, the Seraphim T3 MAA "Uyanah" Lightning Tank shoots land units through shields unimpeded. I'm pretty sure this is unintended and is quite overpowered. It was reported by BlackKiwi89 on the FAF discord and I verified it myself by testing in a sandbox game.

MrRowey:
@Garanas this is still an issue, just did some testing the shield bubble is taking from what looks to be about 5% of the damage and the rest is let through the shield to its intended target and just zap it. Also, notice while I had the log open every times it fired it spams this warning in the log

![image](https://user-images.githubusercontent.com/20344151/145051021-f8e39977-29f7-4f2d-aa0c-8d8afde4f13a.png)

Garanas:
Good find - I'll see if I can fix this before the 20th.

MrRowey:
Tested on Update version of FAF Develop -  Waring dose not show :) but still kills units though shield

Garanas:
I found out why this happens - they hacked in how the damage works. Because of that it bypasses any logic, in fact you can make it do double damage (damage an intermediate target and the target you're aiming for).

PR #4058, opened 2022-07-19
Fix various issues with beams and projectiles

Garanas:
This pull requests consists of several linked issues. Closes https://github.com/FAForever/fa/issues/3153, closes https://github.com/FAForever/fa/issues/3888

## Anti air tracking and damage radius

https://user-images.githubusercontent.com/15778155/179852097-0ca5970f-54d1-41c1-86a9-b1b6f4086716.mp4

Unreliable, tracking with leading and too little damage radius: the scout survives

https://user-images.githubusercontent.com/15778155/179852112-ae1087a7-ee85-4313-ba3d-a2ffabc97749.mp4

Reliable, tracking without leading and proper applied damage radius: the scout dies


Garanas:
That is an interesting side effect 😄 , after discussing it with Spikey we've chosen a different direction: we add the velocity of the target the project impacts with to make the damage move 'along' with the entity it does damage to.

Garanas:
A visual of the offset in damage:

image
_A scout that is turning, red marks the old damage location where as blue marks the new damage location_

image
_A scout at maximum velocity, red marks the old damage location where as blue marks the new damage location_

image
_A scout at maximum velocity, red marks the old damage location where as blue marks the new damage location. The collision boxes are also rendered (in blue), clearly showing that the old damage location is no longer in reach of the spy plane_

PR #4059, opened 2022-07-20
Refactor Scenario Framework part 1

Hdt80bro:
Covers [a bit more than its fair share of] the first third of annotating the scenario framework. Incidentally, these functions are basically pulled from `scenariotriggers.lua`, so I also ended up annotating those copied functions as well. And when I say "annotate" I kinda mean "refactor"--but don't worry, its *mostly* just variable renaming. This one definitely became larger than I thought it was going to be, even breaking it into chunks... I'll reevaluate my methods.


Garanas (approved):
Looks great!

Garanas:
@speed2CZ could you find the time to look over these changes too?

lua/ScenarioFramework.lua
```diff
@@ -712,81 +601,93 @@ function SetupMFDSync(movieTable, text)
         tempData.text = tempText
         LOG("ERROR: Unable to find name in string: " .. text .. " (" .. tempText .. ")")
     end
-
-    local timeSecs = GetGameTimeSeconds()
-    tempData.time = string.format("%02d:%02d:%02d", math.floor(timeSecs/360), math.floor(timeSecs/60), math.mod(timeSecs, 60))
-    tempData.color = 'ffffffff'
+    -- local timeSecs = GetGameTimeSeconds()
```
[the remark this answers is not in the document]
Hdt80bro: Ah, I was still testing to see if `GetGameTime()` was the perfect solution to whatever mess was going on before (there was--count 'em--two bugs in that line). Turns out that it is!

speed2CZ (changes requested):
Please test your changes first and debug it, then I'll review it.

Hdt80bro:
The IDE is really good about letting you about problems with annotations--are you talking about making sure I renamed parameters correctly or the small logical equivalences I did?

Hdt80bro:
Hhm, FAF is currently completely broken as to anything to do with the campaign, so it can't be properly tested. I've summarized the kinds of logical changes here if you'd like to review them manually. Would you prefer that I revert them instead?
```diff
- function PlatoonMoveRoute(platoon, route, squad)
-     for _, v in route do
-         if type(v) == 'string' then
-             if squad then
-                 platoon:MoveToLocation(ScenarioUtils.MarkerToPosition(v), false, squad)
-             else
-                 platoon:MoveToLocation(ScenarioUtils.MarkerToPosition(v), false)
-             end
-         else
-             if squad then
-                 platoon:MoveToLocation(v, false, squad)
-             else
-                 platoon:MoveToLocation(v, false)
-             end
-         end
-     end
- end
+ function PlatoonMoveRoute(platoon, route, squad)
+     for _, node in route do
+         if type(node) == 'string' then
+             node = ScenarioUtils.MarkerToPosition(node)
+         end
+         platoon:MoveToLocation(node, false, squad)
+     end
+ end

- function GroupProgressTimerThread(group, time)
-     ...
-     v:SetWorkProgress(currTime/time)
-     ...
- end
+ function GroupProgressTimerThread(units, time)
+     ...
+     local prog = currTime / time
+     ...
+     unit:SetWorkProgress(prog)
+     ...
+ end

- function Dialogue(dialogueTable, callback, critical, speaker)
-     local canSpeak = true
-     if speaker and speaker.Dead then
-         canSpeak = false
-     end
- 
-     if canSpeak then
-         ...
-     end
- end
+ function Dialogue(dialogueTable, callback, critical, speaker)
+     if not (speaker and speaker.Dead) then
+         ...
+     end
+ end

- SetupMFDSync(movieTable, text)
-     ...
-     local timeSecs = GetGameTimeSeconds()
-     tempData.time = string.format("%02d:%02d:%02d", math.floor(timeSecs/360), math.floor(timeSecs/60), math.mod(timeSecs, 60))
-     tempData.color = 'ffffffff'
-     if movieTable[4] == 'UEF' then
-         ...
-     end
-     ...
- end
+ SetupMFDSync(movieTable, text)
+     ...
+     tempData.time = GetGameTime()
+     if movieTable[4] == 'UEF' then
+         ...
+     else
+         tempData.color = 'ffffffff'
+     end
+     ...
+ end

*variable storage changes not included
- function PlayDialogue()
-     ...
-     if v.text and not v.vid then
-         if not v.vid then
-             DisplayMissionText(v.text)
-         end
-     end
-     ...
-     local vidText = ''
-     if v.text then
-         vidText = v.text
-     end
-     ...
-     SetupMFDSync(movieData, vidText)
-     ...
- end
+ function PlayDialogue()
+     ...
+     if text and not vid then
+         DisplayMissionText(text)
+     end
+     ...
+     text = text or ""
+     ...
+     SetupMFDSync(movieData, text)
+     ...
+ end

- function AreaTriggerThread(callbackFunction, rectangleTable, category, onceOnly, invert, aiBrain, number, requireBuilt, name)
-     ...
-     for _, v in rectangleTable do
-         table.insert(recTable, ...)
-     end
- 
-     ...
-     for _, v in recTable do
-         ...
-         totalEntities[table.getn(totalEntities) + 1] = ve
-         ...
-     end
-     ...
-     local numEntities = table.getn(totalEntities)
-     if numEntities > 0 then
-     ...
-     if number and ((amount >= number and not invert) or (amount < number and invert)) then
-     ...
-     if (triggered and not invert and not number) or (not triggered and invert and not number) or (triggered and number) then
-     ...
- end
+ function AreaTriggerThread(callback, rectangles, category, onceOnly, invert, aiBrain, unitCount, requireBuilt, name)
+     ...
+     for i, rect in rectangles do
+         recTable[i] = ...
+     end
+ 
+     ...
+     local totalEntityCount = 0
+     for _, rect in recTable do
+         ...
+         totalEntityCount = totalEntityCount + 1
+         totalEntities[totalEntityCount] = entity
+         ...
+     end
+     ...
+     if totalEntityCount > 0 then
+     ...
+     if unitCount and ((amount >= unitCount) ~= invert) then
+     ...
+     if (not unitCount and (triggered ~= invert)) or (triggered and unitCount) then
+     ...
+ end

- function ThreatTriggerAroundUnitThread(callbackFunction, aiBrain, unit, rings, onceOnly, value, greater, name)
-     ...
-     if greater and threat >= value then
-         -- stuff
-     elseif not greater and threat <= value then
-         -- same stuff
-     end
-     ...
- end
+ function ThreatTriggerAroundUnitThread(callback, aiBrain, unit, rings, onceOnly, value, greater, name)
+     ...
+     if (greater and threat >= value) or (not greater and threat <= value) then
+         -- that stuff
+     end
+     ...
+ end

- function TimerTriggerThread(callbackFunction, seconds, name, displayBool, onTickFunc)
-     ...
-     local ticking = true
-     local targetTime = math.floor(GetGameTimeSeconds()) + seconds
- 
-     while ticking do
-         onTickFunc(targetTime - math.floor(GetGameTimeSeconds()))
- 
-         if targetTime - math.floor(GetGameTimeSeconds()) < 0 then
-             ticking = false
-         end
- 
-         WaitSeconds(1)
-     end
-     ...
- end
+ function TimerTriggerThread(callback, seconds, name, display, onTickSecond)
+     ...
+     local targetTime = math.floor(GetGameTimeSeconds()) + seconds
+     onTickSecond(seconds)
+ 
+     while true do
+         WaitSeconds(1)
+         local time = targetTime - math.floor(GetGameTimeSeconds())
+         onTickSecond(time)
+         if time < 0 then
+             break
+         end
+     end
+     ...
+ end

- function CreateArmyStatTrigger(callbackFunction, aiBrain, name, triggerTable)
-     ...
-     if string.find (triggerData.StatType, "Economy_") then
-         aiBrain:GetArmyStat(triggerData.StatType, 0.0)
-     else
-         aiBrain:GetArmyStat(triggerData.StatType, 0)
-     end
-     ...
- end
+ function CreateArmyStatTrigger(callbackFunction, aiBrain, name, triggerTable)
+     ...
+     aiBrain:GetArmyStat(trigger.StatType, 0)
+     ...
+ end

- function UnitToPositionDistanceTriggerThread(cb, unit, marker, distance, name)
-     ...
-     local fired = false
-     while not fired do
-         ...
-         fired = true
-         if name then
-             cb(TriggerManager, name, unit)
-             return
-         else
-             cb(unit)
-             return
-         end
-         ...
-     end
- end
+ function UnitToPositionDistanceTriggerThread(callback, unit, marker, distance, name)
+     ...
+     while true do
+         ...
+         if name then
+             callback(TriggerManager, name, unit)
+         else
+             callback(unit)
+         end
+         return
+     end
+ end

- function CreateUnitCapturedTrigger(cbOldUnit, cbNewUnit, unit)
-     if cbOldUnit then
-         unit:AddUnitCallback(cbOldUnit, 'OnCaptured')
-     end
-     if cbNewUnit then
-         unit:AddUnitCallback(cbNewUnit, 'OnCapturedNewUnit')
-     end
- end
+ function CreateUnitCapturedTrigger(cbOldUnit, cbNewUnit, unit)
+     unit:AddOnCapturedCallback(cbOldUnit, cbNewUnit)
+ end

- function VariableBoolCheckThread(cb, varName, value, name)
-     if value then
-         ...
-         cb(TriggerManager, name)
-     else
-         ...
-         cb(TriggerManager, name)
-     end
- end
+ function VariableBoolCheckThread(callback, varName, invert, name)
+     if invert then
+         ...
+     else
+         ...
+     end
+     callback(TriggerManager, name)
+ end

- function MissionNumberTriggerThread(cb, value, name)
-     ...
-     while not ScenarioInfo.VarTable['Mission Number'] == value do
-     ...
- end
+ function MissionNumberTriggerThread(callback, value, name)
+     ...
+     while ScenarioInfo.VarTable['Mission Number'] ~= value do
+     ...
+ end

- function CreateUnitGivenTrigger(cb, unit)
-     TriggerFile.CreateUnitGivenTrigger(cb, unit)
- end
+ CreateUnitGivenTrigger = TriggerFile.CreateUnitGivenTrigger

- function CreateThreatTriggerAroundUnit(callbackFunction, aiBrain, unit, rings, onceOnly, value, greater)
-     return ForkThread(ThreatTriggerAroundUnitThread, callbackFunction, aiBrain, unit, rings, onceOnly, value, greater)
- end
+ function CreateThreatTriggerAroundUnit(callbackFunction, aiBrain, unit, rings, onceOnly, value, greater, name)
+     return ForkThread(ThreatTriggerAroundUnitThread, callbackFunction, aiBrain, unit, rings, onceOnly, value, greater, name)
+ end

- function CreateTimerTrigger(cb, seconds, displayBool)
-     timerThread = TriggerFile.CreateTimerTrigger(cb, seconds, displayBool)
-     return timerThread
- end
+ function CreateTimerTrigger(callback, seconds, name, display, onTickSecond)
+     timerThread = TriggerFile.CreateTimerTrigger(callback, seconds, name, display, onTickSecond)
+     return timerThread
+ end
```
The only API "breaking" changes are the last three
* `ScenarioFramework.lua` no longer directly calls `scenariotriggers.lua` functions, so changing functions in `scenariotrigger.lua` instead of `ScenarioFramework.lua` won't update them
* adding support for the previously unused `name` parameter in all of the trigger functions means that any calls to these that derive their last argument from unadjusted multiple-return function calls will result in the callback function being called using the triggermanager + trailing name argument version
* fixing the parameters in `CreateTimerTrigger` might mean that mods that, for some bizzare reason, rely on it being wrong need to be updated

Hdt80bro:
Alright, I've done some testing and none of the errors are mine

Garanas:
Can you copy some of the errors you encountered?

Hdt80bro:
Oooh, rats, I already moved back to fafdevelop. I can test a campaign map again the morning if you'd like.

Hdt80bro:
Sorry about all that @speed2CZ , I was still figuring out how to use the campaign branch. Everything's good now!

lua/sim/ScenarioUtilities.lua
```diff
@@ -92,27 +96,28 @@ function GetMarkerChain(name)
...
+---@param areaName Area
+---@return Rectangle
+function AreaToRect(areaName)
+    local area = Scenario.Areas[areaName]
     if not area then
         error('ERROR: Invalid area name')
     end
     local rectangle = area.rectangle
-    return Rect(rectangle[1],rectangle[2],rectangle[3],rectangle[4])
+    return Rect(rectangle[1], rectangle[2], rectangle[3], rectangle[4])
```
4z0t, line 120: Wouldn't `unpack` be better?

Garanas: I'm not sure if this is the case - doesn't unpack essentially do the same, but then with the overhead of a function call?

Hdt80bro: It's a Lua function though, not an engine function, so I think it should be a little faster? Should I run some benchmarks?

Hdt80bro: Oh you're right, it looks like its up to twice as fast to do those 4 table accesses than to unpack it

lua/ScenarioFramework.lua
```diff
@@ -844,40 +783,43 @@ function EndGameWaitThread(callback)
...
+    SimUIVars.SaveEnhancementRestriction(restrict)
+    import('/lua/enhancementcommon.lua').RestrictList(restrict)
+    Sync.EnhanceRestrict = restrict
 end
 
--- Returns true if all units in group are dead
-function GroupDeathCheck(group)
-    for _, v in group do
-        if not v.Dead then
+--- Returns if all units in the group are dead
```
speed2CZ, line 805: Return true if

lua/ScenarioFramework.lua
```diff
@@ -844,40 +783,43 @@ function EndGameWaitThread(callback)
...
+        if not unit.Dead then
             return false
         end
     end
     return true
 end
 
--- Iterates through objects in list.  if any in list are not true ... returns false
+--- Returns if the list is entirely truthy
+---@param list boolean[]
```
speed2CZ, line 818: Where did you find that it's a list of booleans?

speed2CZ (changes requested):

lua/ScenarioFramework.lua
```diff
@@ -5,135 +5,149 @@
...
+
+---@class MovieTable
+---@field [1] string path/name
+---@field [2] string bank
+---@field [3] string cue
+---@field [4] string faction
+
+-- these need to be global because other files refer to them for some reason
+
+Cinematics = import('/lua/cinematics.lua')
```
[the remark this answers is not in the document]
Hdt80bro: If making things local broke our base FA code, I'm scared to think how much that would break other mods. I'd prefer to leave them as-is for now.

lua/ScenarioFramework.lua
```diff
@@ -5,135 +5,149 @@
...
-end
-
--- Single Line timer Trigger creation
--- Fire the <cb> function after <seconds> number of seconds.
--- you can have the function repeat <repeatNum> times which will fire every <seconds>
--- until <repeatNum> is met
 local timerThread = nil
-function CreateTimerTrigger(cb, seconds, displayBool)
-    timerThread = TriggerFile.CreateTimerTrigger(cb, seconds, displayBool)
+--- Creates a UI timer that fires `callback` after `seconds`.
```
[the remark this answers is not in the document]
Hdt80bro: It depends on the `display` parameter--I can add that documentation in

lua/ScenarioFramework.lua
```diff
@@ -5,135 +5,149 @@
...
+--- You can have the function repeat `repeatNum` times which will fire every `seconds`.
```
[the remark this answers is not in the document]
Hdt80bro: Good catch! I blindly copied the original comment, but appears to have been misplaced for `CreateUnitDamagedTrigger`

lua/ScenarioFramework.lua
```diff
@@ -405,267 +373,213 @@ function GetCatUnitsInArea(cat, area, brain)
...
+---@param squad? string
 function PlatoonMoveRoute(platoon, route, squad)
-    for _, v in route do
-        if type(v) == 'string' then
-            if squad then
-                platoon:MoveToLocation(ScenarioUtils.MarkerToPosition(v), false, squad)
-            else
-                platoon:MoveToLocation(ScenarioUtils.MarkerToPosition(v), false)
-            end
+    for _, node in route do
```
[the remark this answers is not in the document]
Hdt80bro: It can be either a position vector or a marker name--are we thinking that `position` encompasses idea that enough?

lua/ScenarioFramework.lua
Hdt80bro: I just looked at how it runs, since it's never used. I originally interpreted it as an array, but yeah, you're right, it could be a regular table with keys, meaning that the values could just be objects that we're checking to be non-nil / false. Should we mark at as `unknown` then?

lua/ScenarioFramework.lua
Hdt80bro: We decided on the imperative mood (so it'd end up being "Returns true if...").
I actually removed the "true/false"s from all of my comment on the basis of succinctness (I see it kind of like using `if bool == true then` instead of `if bool then`) and robustness (`if obj then` isn't entirely equal to `if obj == true then` when the object isn't a boolean)
What are your thoughts about it?

lua/ScenarioFramework.lua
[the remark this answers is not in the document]
Hdt80bro: Hold on, you're completely right! The parameter name misled me--it merely appears to control whether or not the onTickSecond function is called.

lua/ScenarioFramework.lua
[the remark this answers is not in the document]
Hdt80bro: Ah--the part I missed was that "as-is" currently means localized. I missed that they were originally local, sorry

lua/ScenarioFramework.lua
speed2CZ: My point is that the current comment is confusing. Normaly when you say that the function returns, it means it ends without returning anything. And here the comment says that the function return when all units are dead,

Hdt80bro: Huh. I've never thought to read it like that. This is why we have reviews!
I change it in this PR, though I can't promise I'll go back through and change it in everything else I've done.

Garanas: That is a good point, I didn't read it like that either.

Hdt80bro:
Looks like all the reviews are resolved and this has become inactive--I'll merge soon.

Garanas:
Hold up - allow me to run it over one more time tonight 😄

Hdt80bro:
@Garanas How'd it go?

Garanas:
Lets merge it after the release is stable

PR #4061, opened 2022-07-20
T3 land target prios fix

Spikey84:
Hopefully the format of this pr is better than the lost one :P

This PR has my proposed T3 land TP changes and simplifications. All changes will be explained below and updated as more are added. Any balance team members are welcome to commit to this PR (I think thats how this works lol).

Changes

    Refactored all T3 MAAs to be more intuitive (for example hitting asf before t1 bombers)
    Simplified T3 assault bots to target t1 and 2 land equally with pd (im not sure i like this)
    Simplified long range land units
    Made percys more consistent with other assault bots (this may need to be changed)


Hdt80bro:
What's the thought process behind making T3 MAA target transports before experimentals?

Spikey84:
> What's the thought process behind making T3 MAA target transports before experimentals?

The thought is that a czar could easily screen for a drop. The drop could then land while the czar takes a tiny bit of damage. The damage required to kill the transport is so little that its worth the less dps going into the experimental in order to easily prevent the drop from landing. But im open to alternatives ofc.

Hdt80bro:
Nah, that sounds good. Continentals can tank more shots than the T2 transports, so it just means that those can screen for air experimentals--which shouldn't come up too often since UEF doesn't have any.

Spikey84:
Targeting laden transports would be nice, but it kinda goes against the goal of simplifying targeting. If someone knows a way to do this i would like to test it.

MrRowey:

PR #4062, opened 2022-07-20
T2 Land Target Priority (take 2)

Spikey84:
Simplifies and refactors T2 lands TPs. List of changes below, will update as more are made.

    Fixed pillar being different from other tanks and took out restriction on targeting tmd.
    Simplified Flack, i would appreciate feedback on this change.
    Simplified MML, also would like some feedback.
    Fixed tmd on range bots.
    Fixed mongoose targeting issue (see thread on bal team discord)


MrRowey:

PR #4068, opened 2022-07-23
Refactor scenario framework part 3

Hdt80bro:
Finishes up the scenario framework annotations. A little bit of refactoring, particularly in the offmap scripts--I tested it though and it works fine.


Hdt80bro:
@Garanas are there any concerns about renaming the offmap functions?

Garanas:
No, there are not.

Hdt80bro:
Looks like this one got left in the incubator--I'll merge this after the release

PR #4069, opened 2022-07-23
Refactor sim utils

Hdt80bro:
Refactors `/lua/SimUtils.lua`. Breaks down the rebuild unit process into chunks and makes unit cap sharing work with unlocked teams.


lua/sim/Prop.lua
```diff
@@ -315,6 +315,18 @@ Prop = Class(moho.prop_methods) {
             EntitySetCollisionShape(self, shape, centerx, centery + sizey, centerz, sizex, sizey, sizez)
         end
     end,
+    
+    RevertCollisionShape = function(self)
```
Garanas, line 319: Was this function missing / referenced somewhere?

Hdt80bro: It lets [blocking entities](https://github.com/Hdt80bro/fa/blob/refactor-sim-utils/lua/SimUtils.lua#L594) be able to handle Units and Prop the same way, so that we don't have to track of both types separately because we revert the collision differently.

Hdt80bro:
If you test these changes, make sure you look at unit transferring processes--in particular, rebuilt unfinished units from fullshare

Garanas (approved):

Garanas:
Took a while, but here we are. Had a subtle issue in it.

The primary discovery for me in this pull request is that `SimUtils` essentially only exists for the logic of transferring units 😄

PR #4070, opened 2022-07-23
Add unit utilities

Hdt80bro:
As you might have guessed from the branch name, this is in preparation for reworking the sacrifice bonus (specifically because the engine never intended for support commanders to have presets), but these utilities are probably nice to have in general. There are three main changes:
* Abstracts the total mass cost calculation out of the veterancy code into `Unit:GetTotalResourceCosts()`. It also appears that the previous calculation did not take enhancement prerequisites into account, so that's been fixed.
* Adds the concept of a "default work order," explained in the comments as "something that an assisting unit would be able to directly do for the unit without inheriting work: constructing, enhancing, upgrading, factory building, repairing, or nothing (in that order)"--these are the actions that can happen when the sacrifice command is given. Maybe I should have included that in the comments too
  * This excludes silo missile building, mobile building, shield assisting, reclaiming, capturing, and teleporting
* Provides a method to calculate the progress bonus a sacrifice will have. Note that it doesn't cap the returned progress to the actual bonus that will happen at that given moment for the current default work order. Leaving that to the user increases robustness and is more useful (and I may have plans for the progress spilling over into the next default work order when the sacrifice bonus is too large...)

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2674,7 +2674,6 @@ Unit = Class(moho.unit_methods) {
     OnStartSacrifice = function(self, targetUnit)
         self:SetUnitState("Sacrificing", true)
         EffectUtilities.PlaySacrificingEffects(self, targetUnit)
-        self.SacrificeThread = self:ForkThread(self.DoSacrificeThread, targetUnit)
         self.SacrificeTargetUnit = targetUnit
     end,
 
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -3077,53 +3077,6 @@ Unit = Class(moho.unit_methods) {
         return health / maxHealth
     end;
 
-    --- Adds progress to the default work order for this unit. A default work order is something that
-    --- an assisting unit can directly do for the unit without inheriting work: constructing it,
-    --- enhancing it, upgrading it, factory building from it, and repairing it (in that order).
-    ---@param self Unit
-    ---@param prog number
-    ---@return number progressLeft
-    AddDefaultWorkProgress = function(self, prog)
-        LOG("adding default work progress: " .. prog)
-        if self:GetFractionComplete() < 1.0 then
-            local before = self:GetFractionComplete()
-            SimUtils.AddConstructionProgress(self, prog)
-            return prog - (self:GetFractionComplete() - before)
-        end
-
-        if self.WorkItem or self:IsUnitState("Upgrading") or self:IsUnitState("Building") then
-            local newProgress = self:GetWorkProgress() + prog
-            if newProgress > 1.0 then
-                prog = newProgress - 1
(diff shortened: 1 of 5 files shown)


lua/sim/Unit.lua
```diff
@@ -6,37 +6,41 @@
...
--- cached categories for performance
 local UpdateAssistersConsumptionCats = categories.REPAIR - categories.INSIGNIFICANTUNIT     -- anything that repairs but insignificant things, such as drones
 
--- upvalue for performance
-local rawget = rawget 
-local IsAlly = IsAlly
-local armies = ListArmies()
+
+
+local DeprecatedWarnings = { }
```
Garanas, line 43: This one should be unused, and can be removed

lua/sim/Unit.lua
```diff
@@ -2622,12 +2671,16 @@ Unit = Class(moho.unit_methods) {
         end
     end,
 
-    OnStartSacrifice = function(self, target_unit)
-        EffectUtilities.PlaySacrificingEffects(self, target_unit)
+    OnStartSacrifice = function(self, targetUnit)
+        self:SetUnitState("Sacrificing", true)
+        EffectUtilities.PlaySacrificingEffects(self, targetUnit)
+        self.SacrificeThread = self:ForkThread(self.DoSacrificeThread, targetUnit)
```
Garanas, line 2677: The function `DoSacrificeThread` doesn't exist

Garanas (approved):
Looks good - works as intended. It does feel like the pull request is not complete. Yet, it doesn't cause an error so I'm not entirely sure what is going on 😄

lua/sim/Unit.lua
Hdt80bro: Doh! I forgot to remove that

Hdt80bro:
Yeah, it's just groundwork for my future plans to rework sacrificing, so it doesn't add any functionality (besides fixing the total unit cost not including prerequisite enhancements). I thought I'd try to keep things small, since I usually struggle with that.

lua/sim/Unit.lua
Garanas: Yes, but it also appears this function didn't get called. Are you sure it gets called when a unit starts sacrificing?

Hdt80bro: Yes, I was testing a while back and `OnStartSacrifice` works fine--the effects still play and everything. I actually started to implement my rework (inspiring https://github.com/FAForever/fa/pull/4069), but was able to scale it back to something manageable. I'll continue again once these two get closed.

lua/sim/Unit.lua
Hdt80bro: Not quite--there's still a couple uncommented out deprecation warnings in the class.

Garanas:
I'm fine with merging this - do as you see fit @Hdt80bro

Issue #3433, opened 2021-09-20
Support commanders dust cloud explosion happens when underwater


Garanas:
![image](https://user-images.githubusercontent.com/15778155/133986033-c982d019-64be-41db-93bf-a133193aac7a.png)

When a support commander dies underwater the dust cloud explosion still happens. This shouldn't be the case.

Hdt80bro:
What should happen when the sACU is in the shoal? Would we need to rework the effects to make it transition from having a smoke ring to not? Otherwise, if it's a hard transition, like if we simply guarded [this](https://github.com/FAForever/fa/blob/deploy/fafdevelop/effects/Entities/SCUDeath01/SCUDeath01_script.lua#L49) when the unit's underwater, it might look funny.

MrRowey:
If the unit is under the water would think that just behaving the unit explosion with out dust cloud makes sense

Hdt80bro:
Sorry, I meant if the unit is only partially underwater. What happens then? When do we stop drawing the smoke ring?

MrRowey:
well unless their a way to having different stages where the unit is fully under the water compared to paddling on the beach I would assumes if in water no smoke ?

Hdt80bro:
Sure, we can try that
![image](https://user-images.githubusercontent.com/37224614/180617563-834a9235-58a7-4ef5-9463-1c81ea96be0e.png)
This look good?
```lua
if position[2] + 2 > GetSurfaceHeight(position[1], position[3]) then
    self:ForkThread(self.CreateOuterRingWaveSmokeRing)
end
```

MrRowey:
Yes this looks pretty good in my option as soon as the SACU head is under the water the make sense for their to be not dust cloud

@Garanas thought with this way of implementing it ?

Garanas:
Looks good to me - in the future we should have a different explosion for when the SACU is below water. My thoughts are something [like this](https://science.howstuffworks.com/explosion-land-water1.htm), in particular the image. We can re-use some of the current naval destruction effects for that, and spawn those effects on the surface.

Hdt80bro:
The image is nice and all, but now I've read the article and want to revamp all explosions to be more destructive underwater...
*ignore the impulse*
I'll put in that change

PR #4071, opened 2022-07-23
Remove smoke ring on underwater SACU death

Hdt80bro:
Closes https://github.com/FAForever/fa/issues/3433

### DIFF
--- a/effects/Entities/SCUDeath01/SCUDeath01_script.lua
+++ b/effects/Entities/SCUDeath01/SCUDeath01_script.lua
@@ -46,7 +46,9 @@ SCUDeath01 = Class(NullShell) {
     EffectThread = function(self)
         local army = self:GetArmy()
         local position = self:GetPosition()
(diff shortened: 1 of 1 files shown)


MrRowey:
Will test it on my end

Hdt80bro:
Perfect! Right now, the `2` in `if position[2] + 2 > GetSurfaceHeight(position[1], position[3]) then` comes from the smoke ring being created 2 units above the commander (the `2` hidden in `...X * OffsetMod , 2, Z * OffsetMod...` [here](https://github.com/FAForever/fa/blob/deploy/fafdevelop/effects/Entities/SCUDeath01/SCUDeath01_script.lua#L81)).
If you find it looks better with a different number (e.g. because the smoke ring has height) then let us know.

MrRowey:
Based on my testing this works as descried and the current paraments are good :)

Hdt80bro:
Perfect, we can merge this then

PR #4073, opened 2022-07-24
Optimize range shaders

KionX:
Another optimization range rings and vision.
TwoSidedStencil hardware support is required.
Since DX9 is very old, I think there will be no problems.

### DIFF
--- a/effects/range.fx
+++ b/effects/range.fx
@@ -26,9 +26,9 @@ Vertex vertexShader( float3 vertex : POSITION0, float2 coeff : TEXCOORD0, float2
 
 technique Cast
 {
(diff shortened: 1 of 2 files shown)


Garanas:
image
_FAF game type_

image
_FAF Develop game type_

image
_FAF Develop exe + this PR_

Garanas:
I've tried to find out what hardware does and does not support it, but couldn't find a list. The first few hits are from 2001 up to 2003 with regards to papers. That is twenty years ago! Let's give this a try - is the pull request final @KionX ?

KionX:
Yes, final.

PR #4076, opened 2022-07-24
Again range shader optimize

KionX:
I missed one little thing. This will add a lot of FPS.


Garanas:
From a technical perspective, what does this change exactly?

KionX:
Disables unnecessary writing to the stencil buffer.

PR #4078, opened 2022-07-26
Adjust transport collision hitboxes

### LINKED ISSUE #4066: Adjust behavior of transports
Description
-------------

All transports in the base game have an extensive hit box to prevent attached units from soaking up direct fire, causing the transport to live a lot longer than it should. A side effect of this is that splash damage can often not reach the attached units, in turn flak and other splash-damage based weapon often only deal damage to the transport and not to the attached units.

image
_An example of the hitbox, any flak damage that hits the transport from below will not damage the attached units_

Course of action
------------------

We'll be fixing this into something that the balance team can agree with. This involves two parts:

- (1) The hitbox of a transport nicely fits the model again
- (2) We disable collisions on units that are attached

We can approach (1) by adjusting the following blueprint values of a unit for each transport:

- `SizeX`
- `SizeY`
- `SizeZ`
- `CollisionShape`
- `CollisionOffsetX`
- `CollisionOffsetY`
- `CollisionOffsetZ`

We can approach (2) via `OnAttachedToTransport` and `unit.OnDetachedFromTransport`, where we can set the flag `unit.DisallowCollisions` to true and false respectively. We'll also need to adjust the `OnKilled` of the transport to guarantee that the flag is always reset properly.

Useful hotkeys:

- `CTRL + ALT + SHIFT + C`: Show collision boxes of units

Test plan
----------

Changing the flag `unit.DisallowCollisions` to `true` can have quite the impact if the unit retains that flag. Therefore it is critical that at any point the flag should be turned false if the unit is no longer attached. This requires quite some testing, especially in the rare scenario where an attached unit manages to survive a transport crash. I'm not sure how that happens, but it does.

Learning goals
----------------

Feeling more comfortable working with the repository.

pobiega:
Attempts to fix #4066

I first implemented an `OnKilled` function for transports that iterated the cargo to make sure `DisallowCollisions` was set to false, but in testing it seemed `OnDetachedFromTransport` handled it in all cases I could come up with - I found no way to kill a 
transport without `OnDetachedFromTransport` being called on each attached unit.

Also not entirely sure how much "free space" is desired to be left above the model, it seems to vary quite a bit before (and after) changes and per model.

### Badly taken hitbox screenshots
Aluminar
Aluminar
Chariot
Chariot
Continental
Continental 1
Continental 2
Courier
Courier
Dragonfly
Dragonfly
Sky Lifter
Sky Lifter
Skyhook
Skyhook
Vish
Vish
Vishala
Vishala

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -4202,6 +4202,7 @@ Unit = Class(moho.unit_methods) {
             self:DisableDefaultToggleCaps()
 
         end
+        self.DisallowCollisions = true
         self:DoUnitCallbacks('OnAttachedToTransport', transport, bone)
     end,
 
@@ -4210,6 +4211,7 @@ Unit = Class(moho.unit_methods) {
         self:EnableShield()
         self:EnableDefaultToggleCaps()
         self:TransportAnimation(-1)
+        self.DisallowCollisions = false
         self:DoUnitCallbacks('OnDetachedFromTransport', transport, bone)
     end,
 
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -4202,7 +4202,6 @@ Unit = Class(moho.unit_methods) {
             self:DisableDefaultToggleCaps()
 
         end
-        self.DisallowCollisions = true
         self:DoUnitCallbacks('OnAttachedToTransport', transport, bone)
     end,
 
@@ -4211,7 +4210,6 @@ Unit = Class(moho.unit_methods) {
         self:EnableShield()
         self:EnableDefaultToggleCaps()
         self:TransportAnimation(-1)
-        self.DisallowCollisions = false
         self:DoUnitCallbacks('OnDetachedFromTransport', transport, bone)
     end,
 
--- a/lua/system/dkson.lua
+++ b/lua/system/dkson.lua
@@ -46,13 +46,6 @@ json = { version = "dkjson 2.5" }
 
 local _ENV = nil -- blocking globals in Lua 5.2
 
-pcall (function()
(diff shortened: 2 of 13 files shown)


Garanas:
Apologies for the late response - I was on holidays. The results look great, I'll review it later today 😄

Garanas:
I've been looking at it, so far so good. One issue I noticed is that when the transports are landed (without cargo) they can be too close to the ground, causing projectiles (from interceptors, for example) to collide with the terrain instead of the transport ship. This is due to a bug: the engine chooses the last target that it hits, instead of the first. The projectiles of interceptors fly so fast that they intercept with two targets: the transport and the terrain. And the terrain is 'last', and therefore chosen.

I'll fix the projectile speed of interceptor weaponry in another pull request. Can you extend the collision box of all transports vertically by 0.4, you shouldn't need to adjust the offset as we're extending up instead of down.

image
_An example of the extended hitbox_

pobiega:
@Garanas All transports extended by 0.4, except Chariot as I saw you did that one.

Garanas:
After a short discussion with the balance team we decided to remove the disabling of collisions again - this is a simulated game, therefore if they catch a shot or two that is fine.

This was your first issue, what are your thoughts on it? Was it convenient to do, did you feel some information was missing?

pobiega:
The issue was very detailed and described the problem well, so it was actually pretty easy to get going. The hardest thing was getting the in-game cheat panel working, as Geforce Experience apparently blocks many of the ingame debug tools keybinds by default, haha.

I'm not super familiar with Lua, but I'm a C# developer professionally, so git/vsc/lsp stuff was well known. I'll probably do more issues further on, if possible :)

PR #4086, opened 2022-08-02
Extend blueprint annotations

Hdt80bro:


lua/system/class.lua
```diff
@@ -439,36 +417,41 @@ function ConstructClass(bases, specs)
 
     class.__index = class
 
+    local staticInit = class.__static_init
+    if staticInit then
+        staticInit(class)
```
Garanas, line 422: What is this function call, it doesn't appear to exist at the moment

lua/system/class.lua
```diff
@@ -181,22 +175,18 @@ function Class(...)
     -- }
 
     -- Class ({ field=value, field=value, ... })
-    if IsSimpleClass(arg) then 
-        local class = ConstructClass(nil, arg[1] )
-
-        -- set the meta table and return it
+    if IsSimpleClass(arg) then
+        ---@type fa-class
+        local class = arg[1]
```
Garanas, line 180: I don't understand this change

lua/sim/Buff.lua
```diff
@@ -469,7 +530,7 @@ end
 -- A key -> function table for buffs, uses the buffName parameter
 local UniqueBuffs = { }
 UniqueBuffs['SeraphimACURegenAura'] = BuffRegenFieldCalculate
-UniqueBuffs['SeraphimAdvancedACURegenAura'] = BuffRegenFieldCalculate
```
Garanas, line 472: Are we sure this doesn't break anything, or has it been broken all this time? 😄

Hdt80bro: Pretty sure that's been broken this whole time. The buff itself is created by the unit script, so that's been working.

Garanas: We'll see what happens then, interesting - wrote this to fix a balance patch.

lua/system/class.lua
Hdt80bro: Ahh, one of my pre-staged changes. `__static_init` lets the class modify itself once its been constructed (particularly to modify `__call` or `__index`).
For example, it'd structure the aliasing/lazy-evaulation built into [Statistics.lua](https://github.com/Hdt80bro/fa/blob/augment-profiler/lua/shared/statistics.lua#L768-L814). (Now, arguably, some of those features shouldn't exist because it encourages improper usage of the class...)

lua/system/class.lua
Hdt80bro: I clarified the role `ConstructClass` has in the class (which is actually the second argument to the function--`arg[1]`). The logical change actually occurs just below, where I set the metatable to be set *before* the class is constructed, since that's usually what makes sense.

lua/system/class.lua
Garanas: I'm not sure about this. I understand it may have value, but modifying classes directly outside of the hooking structure (as this function would allow) has always been discouraged. In particular, if people end up tweaking functions (of the class) it may break the hierarchy that is used to determine whether a function has been inherited.

Hdt80bro: Actually, the `__static_init` function is still a field, so it can still be modified and even inherited! (I understand that the very idea of "static inheritance" doesn't make sense to a lot of people though).

As to tweaking how an object views a field, I agree that it can be easily be misused. While that goes for most things in general (read as: *code is very easy to break*), this does have a greater propensity to be difficult to fix. Note that someone could have easily done this before, so this merely gives the user another access point to do it. I get discouraging that however.

In this case (lazy evaluation of the statistics cache) I found it to be the cleanest and most robust solution that allows for extensibility. If you're *really* against letting classes modify their own perception of data, then I can look for another solution.

Garanas: All code is easy to break - but we shouldn't make it any easier. 

> Note that someone could have easily done this before, so this merely gives the user another access point to do it. I get discouraging that however.

How would they have done this before, by manually editing the class in the file? 

> In this case (lazy evaluation of the statistics cache) I found it to be the cleanest and most robust solution that allows for extensibility. If you're really against letting classes modify their own perception of data, then I can look for another solution.

I'll think it over for a few days.

Hdt80bro: > How would they have done this before, by manually editing the class in the file?

Yeah, that's actually what I ended up doing for the that StatCache class I referenced

Hdt80bro: In summary, this doesn't make it easier to break code--it's intended common use isn't even to alter metatable (though it would actually make such uses conform to the class hierarchy). I imagine it more like static initializers in other OOP langauge--a place for calling static functions and setting fields.

If we *don't* want this method to be inherited, then we could either abandon the idea, or exclude it from inheritance and setup a wrapper to intercept new additions to the class. That way we can inform the logic that checks if something has been inherited of the change to the class. The alternative is to set the metatable of the Class (and not its objects) to not accept new additions to the class.

I didn't actually realize that Classes check their methods against their bases to detect cycles--I just thought derived classes pulled everything from its bases and that was that. However, if classes aren't supposed to be modified after construction, then it sounds like we already have several issues: (1) the second alternative I presented should already be in place, (2) we can't use CPP-style method class implementations (though I think we only use those with classes that don't call `Class` at the moment), (3) the current hierarchy check is needlessly complex and can be greatly simplified.

Anyways, I'll take a look at how things work more carefully, now and in the future. It's sounding like this change is bringing up several points that require a large change to address, so I'll save it for another time and revert this change.

Issue #3730, opened 2022-03-26
Benchmarks


Garanas:
Description
-----------

The repository contains various [benchmarks](https://github.com/FAForever/fa/tree/deploy/fafdevelop/lua/benchmarking/benchmarks). At the moment it requires [a mod](https://gitlab.com/supreme-commander-forged-alliance/other/profiler) to run these benchmarks. The mod is not intuitive to use. 

The profiler window has a tab preserved for benchmarks. This tab needs to become functional.

Course of action
----------------

The content of the panel would be roughly the following:

![image](https://user-images.githubusercontent.com/15778155/160252156-13e24fc2-4b48-4087-9b21-191cc4ad99d4.png)
_10/10 drawing of the idea_ 

On the left side there is:
 - A button to find all benchmarks (that are in the benchmark folder)
 - A drop down to choose a file with benchmarks
 - A drop down to choose a function from the file of benchmarks
 - A button to run the benchmark
 
 On the right side there is:
  - The time taken to run the benchmark
  - The byte code of the benchmark
  
The profiler is split over three components:
 - The [sim-side](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Profiler.lua) of the profiler
 - The [ui-side](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/game/Profiler.lua) of the profiler
 - The [shared components](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/shared/Profiler.lua) between ui and sim of the profiler
 
On the sim side the running of benchmarks is already implemented. See the functions [FindBenchmarks](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Profiler.lua#L164) and [RunBenchmarks](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Profiler.lua#L253). The information needs to be synced to the UI side. The information needs to be stored on the UI side. And the UI side needs to be able to visualize the information.

The profiler is written using a class structure as introduced by @4z0t . Please adhere to it.

Test plan
---------

This should be one pull request. The pull request will solely implement the benchmarking functionality as described. It should therefore at a minimum support:
 - Searching for files with benchmarks
 - Being able to run benchmarks
 - Being able to see the results of benchmarks
 
And above all, it should be friendly to use. This will be the primary tool to understand how certain aspects of the game impact performance. As an example, it would be great if it recalls the selection of the second drop box when you change the selection of the first drop box.

PR #4087, opened 2022-08-02
Revamp benchmarking

Hdt80bro:
Adds benchmarking to the profiler and debug classes for functions.

Fixes https://github.com/FAForever/fa/issues/3730


Hdt80bro:
I'm completely open to UI or stats changes by the way. I'm pretty happy with how the bytecode displays (although it requires a lot of work to parse--we're one step away from our own Lua virtual machine at this point), but that's open to feedback too (e.g. if you have a different way to represent jumped-to instructions than just prefixing them with '>').

Garanas:
I don't know what half of the statistics that you included means in practice. I'll have to look that up 😄 ! I don't have any initial feedback on the work done, we may need to tweak some benchmarks to not last a second (have a smaller number of iterations).

lua/shared/Profiler.lua
```diff
@@ -1,49 +1,623 @@
...
+            ["other"]  = {},
+            ["upval"]  = {},
+        },
     }
-end
\ No newline at end of file
+end
+
+-- the profiler benchmarks must be guarded so that players can't maliciously send requests
+local devs = {"jip"}
```
Garanas, line 91: You can add yourself to this

lua/sim/Profiler.lua
```diff
@@ -243,48 +274,40 @@ function FindBenchmarks(army)
...
-local benchmarkOutput = { }
-local benchmarkRuns = 30
-
-function RunBenchmarks(info)
-
-    -- localize for performance
-    local abs = math.abs
+function RunBenchmark(fileIndex, benchmarkIndex)
+    if not benchmarkThread then
+        benchmarkThread = ForkThread(RunBenchmarkThread, fileIndex, benchmarkIndex)
```
Garanas, line 282: Add a LOG statement indicating what benchmark is being run

lua/system/repr.lua
```diff
@@ -55,6 +59,10 @@ local function IsLazyVar(t)
 end
 
 local function _FormatHeader(t)
+    -- everything else will break with modules if it tries to read fields as non-existent global variables
```
Garanas, line 62: The same applies to a vector if I recall correct

lua/ui/game/Profiler.lua
```diff
@@ -128,10 +139,32 @@ function ReceiveBenchmarks(data)
...
+        if benchmarkState.SelectedBenchmark ~= 0 then
+            runButton:Enable()
+        end
+    end
+    benchmarkRunning = false
 end
 
---- Toggles the profiler on / off in the simulation, sends along the army that initiated the call
+-- doesn't work!
+--function ReceiveBenchmarkProgress(data)
```
Garanas, line 156: You can remove the function all together!

Garanas:
Absolutely great work 👍 , I hope this changes our perception as to what is and is not a performance improvement. In particular, using upvalues to cache a table access has not shown to be as effective in practice. Running those benchmarks now reveals that too in some cases.

lua/system/repr.lua
Hdt80bro: The problem with modules is that non-existent global variables throw an error (and modules are loaded in a new environment, so all of their fields get this treatment), so any of the checks to see what type of object it is will fail. It just so happens that the first check is for Vector and so it ends up throwing the error there--I don't think vectors specifically have this problem.

lua/ui/game/Profiler.lua
Hdt80bro: I was *really* hoping to be able to sync benchmark iteration progress back to the UI so that there's a visual indicator to the user that something's happening (instead of the static "Running..." text currently there). It turns out that the Sync table only updates at the end of function calls however. So it looks like the only way to get progress to propagate back to the UI is if each benchmark sample is instigated by the UI, and that sounds rather messy.

Can you think of a way to get this to working? I left the code commented out in case someone else knows more about sync behavior.

Hdt80bro:
I already set a metadata system for the benchmarks (e.g. for display like [this](https://github.com/Hdt80bro/fa/blob/augment-profiler/lua/benchmarking/benchmarks/math-generic.lua#L15-L25)), so it'd be a simple thing to add an iteration count to it if we need certain benchmarks to use fewer iterations.

Hdt80bro:
As for the statistics... it's looking like we might need to reevaluate what information is actually useful (e.g. Excess Kurtosis [how much more "tailedness" the data has compared to the normal distribution] is always about -1.5). I rewrote the statistics file more as a library that we could reuse elsewhere, so we shouldn't feel bad about not using everything it has to offer.

lua/ui/game/Profiler.lua
Garanas: The sync is synced every tick, you can add a wait between iterations on the sim-side. It would then show up accordingly on the UI side.

Hdt80bro:
There we go--I synced benchmark progress (though now it's a little slower) and implemented tooltips and custom benchmark runs

Garanas:
```
WARNING: Error running OnMouseoverItem script in CScriptObject at 1b4a0c00: ...\jip\documents\delete-me\fa\lua\ui\game\profiler.lua(927): attempt to call method `AddressToString' (a nil value)
         stack traceback:
         	...\jip\documents\delete-me\fa\lua\ui\game\profiler.lua(927): in function `BytecodeTooltip'
         	...\jip\documents\delete-me\fa\lua\ui\game\profiler.lua(677): in function <...\jip\documents\delete-me\fa\lua\ui\game\profiler.lua:674>
````

It appears to be related to a tooltip, in particular this line causes it:

image

These return invalid results:

image

It appears to load the file in the UI, causing this particular up value to fail:

image

We can change the benchmark.

Garanas:
@Hdt80bro do you have the intention to finish this pull request? If not, then you can close it

Hdt80bro:
I do.

MrRowey:
@Hdt80bro Any update on completing this as it been over a year since you said you were ?

MrRowey:
Closing this PR For now @Hdt80bro if you still intend to Complete this then you can reopen it as it has been 14 months since the last updates.

PR #4088, opened 2022-08-02
UnitView StackOverFlow issue

4z0t:
Fixes error where assisting each other units (repairing) cause stackoverflow.
image
by @Hdt80bro


Hdt80bro:
It appears that this (now that I've updated it) operates closer to the original `GetRolloverInfo` function. Can you confirm? There shouldn't be any differences between the selection info when you hover over a selected unit and when you don't, and there shouldn't be any errors anymore.

Garanas:
Appears to work like a charm with these changes

image

PR #4092, opened 2022-08-04
Fix the guard return radius of naval units

### LINKED ISSUE #1629: Battleship, cruiser patrol / scan radius
I've seen with Cybran and Sera battleship, if something gets close to their range when they are on a patrol, they persuade the target way beyond the patrol route.

Same goes for Sera curiser, it had same patrol as the battleship

Some example:
https://puu.sh/rW0AG/b2f00f4b61.png

Garanas:
The guard return radius is used when a unit that is patrolling engages, and follows a unit. Once it exceeds the distance, it returns back to the patrol area. This radius was set to 130 for almost all naval units, causing them to take a journey of their own until their target is destroyed or they have moved a distance of more than 130.

The return distance is now set to:

- 40 for tech 1 naval units
- 30 for tech 2 naval units
- 15 for tech 3 naval units
- 0 for experimental naval units and strategic submarines

Closes #1629

### DIFF
--- a/units/UAS0102/UAS0102_unit.bp
+++ b/units/UAS0102/UAS0102_unit.bp
@@ -1,4 +1,7 @@
 UnitBlueprint {
+    AI = {
+        GuardReturnRadius = 40,
+    },
     Audio = {
         AmbientMove = Sound {
             Bank = 'UAS',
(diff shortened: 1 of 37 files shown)


Garanas:
Could someone confirm I got all the naval units, and set their return radius according to the description?

MrRowey:
UEF is missing XES0205 Shield Boat

PR #4095, opened 2022-08-05
Partial Share

Penguin5:
Adds a new share condition that transfers buildings and engineers like in full share, but kills other units like in share until death.


Garanas:
Seems fair enough to me - I'll test it and merge it this weekend!

lua/aibrain.lua
```diff
@@ -852,19 +852,29 @@ AIBrain = Class(moho.aibrain_methods) {
...
                         end
-                        local units = self:GetListOfUnits(categories.ALLUNITS - categories.WALL - categories.COMMAND, false)
-                        TransferUnfinishedUnitsAfterDeath(units, indexes)
+                        if Categories then
+                            units = self:GetListOfUnits(Categories, false)
+                        else
+                            units = self:GetListOfUnits(categories.ALLUNITS - categories.WALL - categories.COMMAND, false)
+                        end
+                            TransferUnfinishedUnitsAfterDeath(units, indexes)
                     end
```
Sheikah45, line 870: This will never run with partialShare because it is all inside the shareOption == 'FullShare' if statement.

Penguin5: That was just for handling the optional Categories input that I added where categories were used in that function...  Would you like me to revert that optional Categories part within the Fullshare part there?

Sheikah45: I am just confused what this is used for then. Why would other categories be used for full share?

Penguin5: There is no current use for it.  I added it for compatibility in case someone used the function with Fullshare in the future or with a mod and wanted/expected the optional Category input to work for it.

Garanas: Due to how the functions are setup they are not hookable by a mod.

Sheikah45: Just seems a little counter-intuitive / adding complexity for no clear reason to me, so I would revert it but leave that to Jip.

Also I think all variables should be camelCase not PascalCase so categories not Categories I believe.

Penguin5: Doing categories instead of Categories didn't work properly in my testing due to messing up stuff like categories.ALLUNITS... I will revert the bit in the Fullshare part though.

lua/aibrain.lua
```diff
@@ -952,6 +959,20 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                for index, brain in brains do
+                    local units = brain:GetListOfUnits(categories, false)
+                    if units and not table.empty(units) then
+                        for _, unit in units do
+                            if unit.oldowner == owner then -- The unit was built by me
+                                unit:Kill()
+                            end
+                        end
+                    end
+                end
```
Sheikah45, line 973: Why are we iterating through all the units for all brains for the categories rather than checking the categories of the units in sharedUnits[owner] like is done in the regular KillSharedUnitsInCategories. 

This would allow us to only look at the actual shared units rather than iterating over every unit of every brain.

Sheikah45: Also this doesn't reset the sharedUnits table for the owner brain

lua/aibrain.lua
```diff
@@ -852,19 +852,25 @@ AIBrain = Class(moho.aibrain_methods) {
...
                         TransferUnfinishedUnitsAfterDeath(units, indexes)
                     end
 
                     for k, brain in brains do
-                        local units = self:GetListOfUnits(categories.ALLUNITS - categories.WALL - categories.COMMAND, false)
+                        if Categories then
+                            units = self:GetListOfUnits(Categories, false)
+                        else
+                            units = self:GetListOfUnits(categories.ALLUNITS - categories.WALL - categories.COMMAND, false)
+                        end
```
Sheikah45, line 873: also rather than break the naming convention just give a different name like categoriesToTransfer

lua/aibrain.lua
```diff
@@ -880,7 +886,8 @@ AIBrain = Class(moho.aibrain_methods) {
             end
 
             -- Sort the destiniation brains (armies/players) by rating (and if rating does not exist (such as with regular AI's), by score, after players with positive rating)
-            local function TransferUnitsToHighestBrain(brains)
+            -- optional category input (default of everything but walls and command)
+            local function TransferUnitsToHighestBrain(brains, categories)
```
Sheikah45, line 890: this is named categories so will have a name conflict as well. Best to rename

lua/aibrain.lua
Sheikah45: Same categories name conflict

Garanas:
Adding mod support to this local functions (local in the scope of another function) has no benefit - they can't be accessed due to their scope. 

At some point in the future, these should be refactored.

lua/aibrain.lua
Penguin5: Alright, I'm changing it to just use a modified KillSharedUnits instead.

Penguin5:
So, can this be merged?

Hdt80bro:
@Penguin5 now that we're finally done with the release and hotpatching, we have time to look through all of the open pull requests again. Could you resolve the conflicts so we can merge this?

Penguin5:
Done @Hdt80bro

Hdt80bro (approved):

PR #4097, opened 2022-08-06
Auto Tac Missile Start

Spikey84:
Makes all ACUs and SACUs that can make tac missile start construction by default.

### DIFF
--- a/units/XSL0301/XSL0301_unit.bp
+++ b/units/XSL0301/XSL0301_unit.bp
@@ -1,6 +1,7 @@
 UnitBlueprint {
     AI = {
         InitialAutoMode = true,
+        
         GuardScanRadius = 26, -- allows for factory reclaim bug 
         TargetBones = {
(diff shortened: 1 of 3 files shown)


Spikey84:
Sorry for so many commits :)

Garanas:
Amount of commits is fine - we squash them anyway 😄

units/XSL0301/XSL0301_unit.bp
```diff
@@ -1,10 +1,12 @@
 UnitBlueprint {
     AI = {
+        GuardScanRadius = 26, -- allows for factory reclaim bug
```
Garanas, line 3: Do we use the SACUs for this?

Spikey84: This is in the SCAU but i had nothing to do with adding it, it was just a conflict that i was trying to resolve

Issue #3042, opened 2020-04-01
cybran shield + hyves - bad eco bug


adminxp:
replay 34+ min https://content.faforever.com/faf/vault/replay_vault/replay.php?id=11372599
recorded video of issue: https://www.youtube.com/watch?v=Ph8BpAL7sDc&feature=youtu.be
description: hyves started to assist shield - doing nothing, shield was not upgraded by them nor healed, just a lot of build power with huge mass consumption for nothing.
If I had no paragon that would exhuast my eco in few seconds.

Garanas:
It appears this happens when the shield has taken damage, you start upgrading (clicking the button) but the hives start assisting between you clicking the button and the upgrade actually happening.

Garanas:
No, it is different. You can reproduce it by:

- Shield of structure takes damage
- Set engineers to assist the structure, healing the shield
- Upgrade the structure during while the shield is being healed
- Bug as described

Now, what can we do about it 😄

PR #4103, opened 2022-08-07
Fix a shield assist bug where resources were sent into Limbo

Garanas:
Closes https://github.com/FAForever/fa/issues/3042

A shield bug which can be reproduced by:
- Shield of structure takes damage
- Set engineers to assist the structure, healing the shield
- Upgrade the structure while the shield is being healed
- Bug as described

https://user-images.githubusercontent.com/15778155/183289862-772505f4-1244-4325-9444-af6b1b520dd6.mp4

_Without the fix applied, resources are consumed but not applied_

https://user-images.githubusercontent.com/15778155/183289863-a8aa8af1-fd9f-483f-98c1-90cd83b71cce.mp4

_With the fix applied, shield is assisted accordingly_

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -314,8 +314,26 @@ StructureUnit = Class(Unit) {
 
         if performUpgrade and order == 'Upgrade' then
             ChangeState(self, self.UpgradingState)
+
+            -- Fix an oddly specific bug to structures with damaged shields:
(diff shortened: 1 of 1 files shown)


Garanas:
@KionX  / @Strogoo what are your thoughts?

KionX:
Perhaps during the upgrade of the building, the unit will be recreated and the assist does not know about it.
It can potentially create a lot of unpredictable problems.
In this case, updating the order is the right decision.

PR #4111, opened 2022-08-08
Reduce memory impact of weapons

Garanas:
Reduces the overall memory impact of weaponry. More to come in future pull requests.


Garanas:
Also adjusts the target check interval of some of the beam weaponry

PR #4112, opened 2022-08-08
Fix the save / load functionality of the game

### LINKED ISSUE #3877: Offline saves of missions don't load correctly
Running the coop missions offline gives the advantage save and load the game later. Most people play online, but few still use it.
Now it's broken that when you load the save, the objective panel is missing and you see the normal skirmish score. Also the AI breaks with some errors and you can't continue the mission.

I did test it on the first aeon mission (Joust)
- start the game offline in skirmish mode, save/load

Some quick testing showed that it used to work on patch [3723](https://github.com/FAForever/fa/releases/tag/3723)
(this might not be the last working patch either)

I didn't have time to track down the exact commit that broke it.

Garanas:
The pattern reverted in this pull request cause issues when you save / load the game. The up value (read: local in file-scope) pattern on a file-scope applied to tables causes issues. The reference doesn't appear to be updated properly. As a result, the file ends up with an empty table (decoupled from the actual table) that is then used.

Subtle little bug 😄 

I've scanned the code for other places where I used this pattern. I found them in `class.lua` and in the init files. The latter is irrelevant.

closes #3877

### DIFF
--- a/lua/system/class.lua
+++ b/lua/system/class.lua
@@ -42,15 +42,15 @@
 -- function: 1DE7E1C0 (BuilderParamCheck, id = 1) = { base instance }
 
 -- It allows us to track a function back to the base instance.
-local Hierarchy = {}
+Hierarchy = {}
 
 --- Debug utilities
 
 local enableDebugging = false
 
-local HierarchyDebugLookup = {}
-local HierarchyDebugLookupCFunctions = {}
-local HierarchyDebugLookupCount = {}
+HierarchyDebugLookup = {}
+HierarchyDebugLookupCFunctions = {}
+HierarchyDebugLookupCount = {}
 
 local function PrintHierarchy()
     -- cache for performance
--- a/lua/system/import.lua
+++ b/lua/system/import.lua
@@ -1,6 +1,20 @@
--- Copyright � 2005 Gas Powered Games, Inc.  All rights reserved.
+-- Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
 --
 -- Implement import()
+
+-- note that actual modules cannot upvalued because upvalues are not reset when we load a save file
+local StringLower = string.lower
+local StringSub = string.sub
+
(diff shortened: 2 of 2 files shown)


Garanas:
Not a lot of people have the co-op dev environment setup. @Hdt80bro / @speed2CZ could either of you confirm it works again?

Garanas:
image
Amount of saving it took to find the relevant changes

speed2CZ:
Objectives are not properly loaded. They are not shown in the UI, units are not highlighted

Garanas:
image

After loading, are you sure you checked out properly?

Garanas:
image

restarted the game, loading both from the campaign menu and from the skirmish load - seems to work fine. The game does forget that I had cheats on 😄 

image

Also the indicator entities show fine

Hdt80bro:
Just a note, the other solution for performance critical code would be to use `debug.setupvalue` to reset the upvalue--this would be a little more work though

Garanas:
It is fine - this is a loss we can take. Did you have the co-op setup too, could you check and compare with and without this PR?

Hdt80bro:
Sure

Hdt80bro:
Appears to be functioning fine to me--it looks like only actual modules cannot be upvalued, everything else is fine, so I went ahead and changed that.

Hdt80bro:
Interestingly, this also fixes the load menu to use campaign saves when you initialize the game from the campaign load screen.

speed2CZ:
Idk what Im doing wrong, but still the same result

Garanas:
What map are you testing on? And if you make changes in the repo (either of them), are they registered in-game?

speed2CZ:
Correct branch, changes do show up. tested FA missions 1, 2, and some custom oned, always the same result

Garanas:
Do you have time to call later today, then we can screen share. See what is going on!

speed2CZ:
I wont have time for at least another week

PR #4113, opened 2022-08-08
Reduce projectile speed of air to air weapons

Garanas:
Allows interceptors to more reliably fire and hit air units that have landed. Otherwise the projectiles end up going so fast that they collide with both the air unit and the ground. Due to the collision bug, the game takes the last collision and therefore the interceptors 'miss' their target.


Garanas:
When testing, particularly noticeable on the tech 1 aeon transport that likes to dig deep into the ground.

PR #4115, opened 2022-08-10
Improve attack move callback

Penguin5:
This filters down the attack move callback to not clear the orders of non-mobile units (so giving this sort of attack move order won't cancel factory/mex/etc build queues).


Garanas:
What is the benefit of this over the approaches that are already available?

Penguin5:
The benefit is being able to set a custom keybind for attack move.  Especially given the frequency of use of attack move, it is nice to be able to set a one-key custom keybind for it, rather than having to use alt-RMB to do an attack move order.

Garanas:
@Tagada14 can you or someone of the balance team confirm the use of this? It seems a bit much to me, better suited for a UI mod

Penguin5:
The sim callback can't be adjusted like this PR does via UI mod afaik.

Garanas:
I agree, then we'd limit it to that specific change but leave the additional key binds for a UI mod.

lua/SimCallbacks.lua
```diff
@@ -565,7 +565,8 @@ end
 
 Callbacks.AttackMove = function(data, units)
     if data.Clear then
-        IssueClearCommands(units)
+        -- filtering down to mobile units is to avoid clearing building queues by mistake
+        IssueClearCommands(EntityCategoryFilterDown(categories.MOBILE, units))
```
Sheikah45, line 569: This filtering seems like it would be unexpected behavior from a UX perspective

Garanas: Factories do not respond to these type of commands. Try having a unit + a factory in your selection, and then manually use the attack move buttons. The factory responds quite different, cancelling the new orders

Tagada14:
I think this would be better as a UI mod, especially given the differences between it and regular AM.

PR #4119, opened 2022-08-12
Extend functionality of selection sets (control groups)

Garanas:
Implements the majority of suggestions by Blodir [from this post on the forums](https://forum.faforever.com/topic/2420/mod-request-fix-control-group-behavior?_=1660143954512). All the functionality is optional, it will default to what everyone is used to.

## Control groups

Control groups in Supreme Commander has always been a bit messy, especially when you look at how other strategy games approach it. We're streamlining their behavior, without removing the capability to have the old behavior. You can find the exact details on Github:

### Deprecated UI mods

The following UI mods are known to collide with these changes, and are therefore deprecated:

- `group_split`
- `Control Group Zoom Mod`
- `additionalControlGroupStuff`

Make a post in this thread if you think some of these features are worth keeping around and we'll discuss whether the UI mod should be re-implemented or that we can integrate the behavior into the game.

### Settings

With these changes, there are three places where you can adjust the behavior. Lets go over them one by one.

- `Set behavior`

Triggers when you try to create a control group with your current selection. The settings allow you to 'steal' units from other control groups, or the old behavior where a unit can be part of multiple control groups

- `Factory behavior`

Triggers when a factory is assigned to a control group and it starts producing a unit. The settings allow you to create units that are not attached to a control group, or the old behavior where a unit inherits the control group of the factory.  

- `Double tap behavior`

Triggers when you select a control group twice. The settings allow you to choose between a series of behaviors:
- - **none**: nothing happens, the behavior is disabled
- - **translate-zoom**: the old behavior, where the camera is translated and zoomed in / out to see the entire control group
- - **translate-zoom-if-lower**: the old behavior, but it does not zoom in. It only translates the camera and zooms out as required.
- - **translate`: the old behavior**, but it does not change the zoom at all. It only translates the camera.

These changes give you full control over what some of the UI mods implemented separately.

- `Double tap decay`

Allows you to define the amount of time (in miliseconds) when something is considered a double tap. Defaults to the old value (which is 2 seconds).

### Hotkeys

For completeness sake I'll include all the related hotkeys. Some of these are unchanged from the old behavior.

- `['set_group1'] = '<LOC key_desc_0015>Set Group 1'`

Defines the units for control group 1. Steal behavior can be adjusted using the `Set behavior` in the settings.

- `['group1'] = '<LOC key_desc_0004>Recall Group 1'`

Recalls the control group by selecting it. 

- `['revert_selection_set'] = 'Restore selection to the one before recalling a control group'` 

New behavior that allows you to switch back to the previous selection when you previously recalled a control group. 

- `['append_group1'] = 'Append control group 1 to current selection'`

The old append behavior, where the control group is added to the current selection.

- `['add_selection_to_selection_set1'] = 'Append the current selection to control group 1'`

An alternative to the append behavior, that allows you to add the current selection to the control group.

- `['combine_and_select_with_selection_set1'] = 'Combine selection and control group 1 and select the control group'`

An alternative to the append behavior that allows you to merge the current selection and the control group and then immediately recalls the control group.


Garanas:
image

Garanas:
Blodir evaluated the branch, it appears to be good. Could someone else review it too? Just a matter of trying the options, see if what is supposed to happen, happens

Tagada14:
This is beyond awesome, I tried to tackle the issue of factories adding units to their control groups but couldn't find the code responsible for it and gave up. I am really glad you managed to make this work!

Garanas:
I've added hotkeys to choose the exact append behavior, regardless of the option set. But they won't show up. Does anyone know why? Previous hotkeys always showed up just fine

Hdt80bro:
I recommend creating localization entries for the options--I actually got tired of syncing my changes to the localization text to the string file when I did localization for the profiler, so I created a utility to automatically pull (and number!) LOC's from files (inasmuch as we can have access to them). I suspect we could automate it even further than that too. Note that I'm not asking you to translate everything, just to leave as an option for someone to.

Hdt80bro:
> I've added hotkeys to choose the exact append behavior, regardless of the option set. But they won't show up. Does anyone know why? Previous hotkeys always showed up just fine

It appears that only items in the `UserKeyMap` or `UserDebugKeyMap` table keys in the user preferences are used for `GetKeyLookup`

Garanas:
> > I've added hotkeys to choose the exact append behavior, regardless of the option set. But they won't show up. Does anyone know why? Previous hotkeys always showed up just fine
> 
> ~It appears that only items in the `UserKeyMap` or `UserDebugKeyMap` table keys in the user preferences are used for `GetKeyLookup`~ _Uh, this is actually exactly what we want_ Everything looks to be fine, so I'm guessing this got figured out

ah, yes - my apologies 😄

PR #4123, opened 2022-08-13
Remove UI rally points of factories

Garanas:
image
Rally points show up when you select a factory, but it is difficult to distinguish to which factory they belong

image
Best to just use the command overview, note that a rally point is **not** a command, it is just a mesh that follows the last command of a factory


Penguin5:
Why not have both (the command overview, with the rally point at the end)?

Hdt80bro:
Aww, I loved those. I understand your reasoning though. Could we leave in the rally point if there's just one factory selected?

Garanas:
I wasn't aware of them until yesterday. Were you actively using them?

Hdt80bro:
I'm just used to seeing them. There are times that showing the command graph would clutter the screen, so I'm happy to quickly see the rally point of a factory I select because I don't have to visually untangle any graphs--and also because it's immediate/automatic. It's also a really fast way to see if any of the factories are still using their default waypoints.

Garanas:
Fair enough, then we'll leave it in

PR #4124, opened 2022-08-13
Improved cursor / command behaviors

Garanas:
This pull request basically implements what Strogo shared [in this forum post](basically properly implements https://forum.faforever.com/topic/3372/random-ui-improvements-ui-mod/30). These changes are generally considered good, hence there's no need to not give it to everyone.

In order to properly implement it the `worldview.lua` required a bit of an overhaul. Any UI mod that previously interacted with `worldview.lua` will likely no longer work.

## Ignore mode

https://user-images.githubusercontent.com/15778155/184528485-40d27d4e-e171-4755-97a1-0bea3c4699ae.mp4

When you hold `CTRL` and not `SHIFT` you limit the possible commands to `Move` and `Attack move`. All other commands (reclaim, attack unit, etc) are ignored. You can not queue these commands

## Reclaim

https://user-images.githubusercontent.com/15778155/184528519-8b9bc35a-0f5d-417f-8e8f-94d0ed81f1d6.mp4

When you specifically try to reclaim the tolerance whether the mouse hovers of a prop (or unit) is lowered, allowing you to issue reclaim orders a lot easier without introducing accidental move orders

## Tactical / nuke launch

https://user-images.githubusercontent.com/15778155/184528568-f0df26c0-a0f4-43cd-bd40-87657b0b42ba.mp4

When you try to issue a tactical / nuke launch they will always ignore all units and props. This should make it a lot easier to hit exactly where you intend to hit. As a consequence, it also makes aiming when zoomed out less trivial

## Disabled reclaim cursor

https://user-images.githubusercontent.com/15778155/184528714-0e9c6bdd-4c4c-43de-8eca-ce83f751c231.mp4

At a certain distance you can no longer issue reclaim commands, the cursor now reflects that by being disabled. It looks horrible at the moment, already asked someone to make a better alternative


Garanas:
@Strogoo could you explain to me the logic behind not showing the reclaim cursor when you're in command mode, but not hovering over reclaim? Your UI mod does something very complicated, and I can't entirely understand the logic of it

MrRowey:
Tested All Seems to work As Explained :)

Strogoo:
@Garanas My thoughts was that it's easier to aim at props with normal cursor and then switching to default reclaim cursor shows user that reclaim order will be applied now, kinda helps to reduce useless clicks

Garanas:
I understand the idea, but not your implementation

Hdt80bro:
```lua
local rmbOrder = wv:GetRightMouseButtonOrder()

if rmbOrder and rmbOrder ~= "RULEUCC_Move" then
    StartCommandMode('order', {name = 'RULEUCC_Reclaim', easyRec = true})
else
    StartCommandMode('build', {name = 'f', fake = true, easyRec = true})
end
```
@Garanas this is the main part. If the command behavior is set to reclaim, but the current order would be a move order, then it's a "fake reclaim order".
```lua
if modeData.fake then
    self.Cursor = {'/mods/RUI/textures/cursorIdle.dds', 2, 2}
else
    self.Cursor = {'/mods/RUI/textures/cursorReady.dds', 15, 15}
end
```
Then, we catch if the reclaim order is fake here to update the cursor

Strogoo:
Yep, it is based on `GetRightMouseButtonOrder()`. If RMB-order is not a "Move" it means cursor is over reclaimable object. When starting a fake "build" order (using fake "build" as left-click blocker),  we pass some flag in its data to check for it later and apply the right cursor.

It took me a while to understand my own code when I decided to upload the mod after a long break tbh :D. It was even more messy in the past

PR #4126, opened 2022-08-15
adjust typing of Class so it works again

hahn-kev:
Basically a while ago I merged in new changes to the extension, and that broke how our class function worked. Looks like some people have tried to fix it with some annotations, but that didn't really seem to work.

This should be a working fine, please confirm we're not seeing any warnings like this:
image


MrRowey:
Can Confirm this Fixes the issuse

PR #4130, opened 2022-08-16
Annotations for defaultunits.lua

MrRowey:
Annotated most of the file any that I was unsure of have left as Unknown

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -35,6 +35,7 @@ StructureUnit = Class(Unit) {
     FxDamage2 = {EffectTemplate.DamageStructureFireSmoke01, EffectTemplate.DamageStructureSparks01},
     FxDamage3 = {EffectTemplate.DamageStructureFire01, EffectTemplate.DamageStructureSparks01},
 
+    ---@param self StructureUnit
     OnCreate = function(self)
         Unit.OnCreate(self)
         self:HideLandBones()
@@ -46,6 +47,7 @@ StructureUnit = Class(Unit) {
     end,
 
     --- Hides parts of a mesh that should be visible when the structure is made on water
+    ---@param self StructureUnit
     HideLandBones = function(self)
         if self.LandBuiltHiddenBones and self.Layer == 'Land' then
             for _, v in self.LandBuiltHiddenBones do
@@ -57,6 +59,7 @@ StructureUnit = Class(Unit) {
     end,
 
     --- Rotates the structure towards the enemy, primarily used for point defenses
+    ---@param self StructureUnit
     RotateTowardsEnemy = function(self)
 
         -- retrieve information we may need
@@ -128,6 +131,9 @@ StructureUnit = Class(Unit) {
         self:SetRotation(degrees)
     end,
 
+    ---@param self StructureUnit
+    ---@param builder Builder
+    ---@param layer string
     OnStartBeingBuilt = function(self, builder, layer)
         Unit.OnStartBeingBuilt(self, builder, layer)
 
@@ -145,6 +151,9 @@ StructureUnit = Class(Unit) {
         end
     end,
 
+    ---@param self StructureUnit
+    ---@param builder Builder
+    ---@param layer string
     OnStopBeingBuilt = function(self, builder, layer)
         Unit.OnStopBeingBuilt(self, builder, layer)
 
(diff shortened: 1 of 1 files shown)


lua/defaultunits.lua
```diff
@@ -288,6 +312,9 @@ StructureUnit = Class(Unit) {
     end,
 
     -- Modified to use same upgrade logic as the ui. This adds more upgrade options via General.UpgradesFromBase blueprint option
+    ---@param self StructureUnit
+    ---@param unitBeingBuilt StructureUnit
```
Garanas, line 316: Doesn't have to be a structure, a factory produces units

lua/defaultunits.lua
```diff
@@ -395,6 +422,9 @@ StructureUnit = Class(Unit) {
         end,
     },
 
+    ---@param self StructureUnit
+    ---@param builder Builder
```
Garanas, line 426: Better to have `Unit` instead of `Builder`

lua/defaultunits.lua
```diff
@@ -427,26 +460,41 @@ StructureUnit = Class(Unit) {
         Unit.StopBeingBuiltEffects(self, builder, layer)
     end,
 
+    ---comment
+    ---@param self StructureUnit
+    ---@param unitBeingBuilt StructureUnit
```
Garanas, line 465: Can also be a regular `Unit`

Garanas: Same for all the others

lua/defaultunits.lua
```diff
@@ -465,6 +513,8 @@ StructureUnit = Class(Unit) {
         self:CreateTarmac(true, true, true, orient, currentBP, currentBP.DeathLifetime or 300)
     end,
 
+    ---@param self StructureUnit
+    ---@param wreckage Wreckage
```
Garanas, line 517: Better to use `Prop`

lua/defaultunits.lua
```diff
@@ -803,6 +881,9 @@ FactoryUnit = Class(StructureUnit) {
         end
     end,
 
+    ---@param self FactoryUnit
+    ---@param unitBeingBuilt FactoryUnit
```
Garanas, line 885: Can also be a regular unit

lua/defaultunits.lua
```diff
@@ -832,6 +916,10 @@ FactoryUnit = Class(StructureUnit) {
     end,
 
     --- Adds a pause between unit productions
+    ---@param self FactoryUnit
+    ---@param productionpause FactoryUnit
```
Garanas, line 920: this is a number (it is used in the wait function)

lua/defaultunits.lua
```diff
@@ -1844,6 +2079,9 @@ BaseTransport = Class() {
         unit:OnAttachedToTransport(self, attachBone)
     end,
 
+    ---@param self BaseTransport
+    ---@param attachBone string
```
Garanas, line 2083: can also be a number, there's a separate bone class / alias somewhere. @Hdt80bro do you know its name?

lua/defaultunits.lua
```diff
@@ -1899,18 +2146,24 @@ BaseTransport = Class() {
 --- Base class for air transports.
 ---@class BaseTransport : AirUnit
 AirTransport = Class(AirUnit, BaseTransport) {
+
+    ---@param self BaseTransport
```
Garanas, line 2150: This should be an `AirTransport`, that inherits from both `AirUnit` and `BaseTransport`

Garanas:
Best to use `any` if you're unsure, instead of `unknown`

lua/defaultunits.lua
Hdt80bro: `Bone`

lua/defaultunits.lua
```diff
@@ -1652,6 +1865,9 @@ AirUnit = Class(MobileUnit) {
         end
     end,
 
+    ---@param self AirUnit
+    ---@param new AirUnit
```
Garanas, line 1869: these are not air units, they are strings

lua/defaultunits.lua
```diff
@@ -1759,6 +1988,10 @@ AirUnit = Class(MobileUnit) {
     end,
 
     --- Called when the unit is killed, but before it falls out of the sky and blows up.
+    ---@param self AirUnit
+    ---@param instigator AirUnit
```
Garanas, line 1992: instigator can be any unit

PR #4132, opened 2022-08-16
Fix The Mass Left Behind when the upgrade kennel is destroyed

MrRowey:
- Set BuildCostMass to `1050` from `500` to allow for the Differential Upgrade Cost Calculation Parameter to work correctly
- Set BuildCostEnergy to `2500` from `5250`
- Add `DifferentialUpgradeCostCalculation = true`

### DIFF
--- a/units/XEB0204/XEB0204_unit.bp
+++ b/units/XEB0204/XEB0204_unit.bp
@@ -123,7 +123,7 @@ UnitBlueprint {
         UniformScale = 0.05,
     },
     Economy = {
-        BuildCostEnergy = 2500,
+        BuildCostEnergy = 5250,
         BuildCostMass = 1050,
         BuildRate = 20,
         BuildTime = 2000,
--- a/units/XEB0204/XEB0204_unit.bp
+++ b/units/XEB0204/XEB0204_unit.bp
@@ -124,9 +124,10 @@ UnitBlueprint {
     },
     Economy = {
         BuildCostEnergy = 2500,
-        BuildCostMass = 500,
+        BuildCostMass = 1050,
         BuildRate = 20,
         BuildTime = 2000,
+        DifferentialUpgradeCostCalculation = true,
         EngineeringPods = {
             {
                 CreateWithUnit = true,


MrRowey:
This is not a Balance issues it clearly a Bug @Garanas your though here

MrRowey:
> I feel like this is more of a bug than a balance issue. A Kennel should not have 50 less mass when 500 mass is put into it.

and it should be more than 50 as it should cost 1500 so you getting the cost of the base upgrade and second upgrade mass

Garanas:
This is a bug, that has an impact on balance. Hence, it needs the balance tag

MrRowey:
i dont see how this effect balance you literally getting back the mass you should have in the first place ?

Hdt80bro:
> This is a bug, that has an impact on balance. Hence, it needs the balance tag

I can see that. It's the concept of "this affects the currently balanced gameplay environment" instead of "this merely brings it to parity with the current balance vision." I'm not being terribly consistent then--I suspect half of my bug fixes are related to balance in that case (e.g. making unit cap sharing work with unlocked teams in <https://github.com/FAForever/fa/pull/4069>)

MrRowey:
sure, but then you could jsut argue eveything that you touch can affect ballance in some what shap or form

MrRowey:
But i do think in the future @Tagada14 if you adding the balance tag for a PR like this to give your reasoning to why. As anything that has the balance tag tends to take twice as long to get sorted out.

Hdt80bro:
> sure, but then you could jsut argue eveything that you touch can affect ballance in some what shap or form

I think the idea is that stuff like UI improvements, code optimizations, refactors, & annotations, dev tools, or new QoL features don't affect balance, but something like this changes actual numbers that could hypothetically influence the outcome in a competitive game (even if it's a bug).

Regardless, it's just a label to help us organize things, and I see no reason why this particular PR should need lots of attention--it's a bug and should be pretty uncontroversial.

Garanas:
Just because it impacts balance doesn't mean that we can't fix it on our own. Tagada is adding the tag to every PR that touches balances (like this one) so that if things have changed then he can more easily find out why.

units/XEB0204/XEB0204_unit.bp
```diff
@@ -124,9 +124,10 @@ UnitBlueprint {
     },
     Economy = {
         BuildCostEnergy = 2500,
-        BuildCostMass = 500,
+        BuildCostMass = 1050,
         BuildRate = 20,
         BuildTime = 2000,
+        DifferentialUpgradeCostCalculation = true,
```
Garanas, line 130: I'm not sure how this field works, bit shouldn't the energy also be adjusted?

Hdt80bro: Yup, you're right! It's handled [here](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/game.lua#L46-L51) by the way

Hdt80bro:
Sorry for blowing up that conversation by the way

PR #4137, opened 2022-08-18
Working prototype on launching missiles without cancelling orders

Garanas:
As requested by the balance team, a prototype of the feature. It isn't complete yet, needs some additional checks before firing the weapon.

https://user-images.githubusercontent.com/15778155/185418565-906f553a-206a-4d23-9b3c-d5f9c3ded5e1.mp4

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -727,6 +727,48 @@ do
     end
 end
 
+do 
+    Callbacks.GroundFireManualWeapons = function(data, units)
+        units = SecureUnits(units)
+        for k, unit in units do 
+            -- unit:GroundFireManualWeapons(data.position)
+            local weapon = false
+            local weapons = unit.Blueprint.Weapon
+            for k, blueprint in weapons do
+                if blueprint.ManualFire and blueprint.CountedProjectile then
+
+                    if blueprint.EnabledByEnhancement then
+                        local enhancements = SimUnitEnhancements[unit.EntityId]
+                        if enhancements then
+                            for _, enh in enhancements do
+                                if enh == blueprint.EnabledByEnhancement then
+                                    LOG("hellooo")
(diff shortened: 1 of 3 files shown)


lua/SimCallbacks.lua
```diff
@@ -727,6 +727,48 @@ do
...
+                if blueprint.ManualFire and blueprint.CountedProjectile then
+
+                    if blueprint.EnabledByEnhancement then
+                        local enhancements = SimUnitEnhancements[unit.EntityId]
+                        if enhancements then
+                            for _, enh in enhancements do
+                                if enh == blueprint.EnabledByEnhancement then
+                                    LOG("hellooo")
+                                    weapon = unit:GetWeaponByLabel(blueprint.Label)
+                                    break
```
Hdt80bro, line 747: This only breaks out of the inner loop, not the outer one.

And so people argue that this is the only place `goto`s have in languages without named breaks & continues: to break out of outer loops.

Unfortunately, we don't get those in Lua 5.0, but a `return` can do the same thing if the code is moved to a new function (which is usually what should have already been done if there's that much nesting and indentation).

Garanas: Ah, whoops 😄 !

I'll fix that, but the concept works regardless!

lL1l1:
Would be interesting to apply to overcharge if possible.

Garanas:
Yes, it would work in a similar fashion.

MrRowey:
@Garanas any idea if this pr will be completed?

MrRowey:
@Garanas Do we see this becoming a feature at any point or would it just be safe to close this Prs for now as it sat dormant for over 2 years

MrRowey:
I am going to close this PR as there has been no work done with it since 2022. It can be re-opened if someone is going to take it on and complete the idea

PR #4147, opened 2022-08-18
Start refactoring `sim\unit.lua`

Hdt80bro:


Hdt80bro:
Too outdated to revive

PR #4157, opened 2022-08-20
Fix armies color functionality

Garanas:
The hotkey is now functional again and it is disabled during a replay (as it causes an error in both the current, and the new implementation).


Garanas:
woops, mmhh

PR #4159, opened 2022-08-20
Do not gift units that are not finished

Garanas:
Also includes a partial changelog 😄 

Fixes the issue where units that are being upgraded are gifted twice, in-place. See also:  https://replay.faforever.com/17811485

image

### DIFF
--- a/changelog-dev.md
+++ b/changelog-dev.md
@@ -3,8 +3,55 @@
 
 ## Bug fixes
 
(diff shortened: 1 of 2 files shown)


Garanas:
@KionX could this be related to your changes on gifting units in the engine? Because it popped up in this last week, while we didn't touch any of this logic.

KionX:
Yes it is.

PR #4064, opened 2022-07-22
Migrate `CalculateBallisticAcceleration.lua` to `defaultweapons.lua`

Hdt80bro:
Addresses [3739](https://github.com/FAForever/fa/issues/3739) (yes I tested this--the results are the same)
As requested, the byte code. It's longer because the `MuzzleSalvoSize==1` case has been optimized to not cache or include salvo logic.
```
(  42)    0 - GETUPVAL       2    0    0
(  42)    1 - MOVE           3    1    0
(  42)    2 - CALL           2    2    2
(  43)    3 - TEST           2    2    1
(  43)    4 - JMP            0    2
(  44)    5 - LOADK          3    0
(  44)    6 - RETURN         3    2    0
(  46)    7 - GETUPVAL       3    1    0
(  47)    8 - GETTABLE       4    2  251
(  51)    9 - GETUPVAL       5    2    0
(  51)   10 - MOVE           6    1    0
(  51)   11 - CALL           5    2    4
(  52)   12 - GETUPVAL       8    3    0
(  52)   13 - MOVE           9    2    0
(  52)   14 - CALL           8    2    4
(  54)   15 - GETUPVAL      11    4    0
(  54)   16 - MOVE          12    2    0
(  54)   17 - CALL          11    2    2
(  56)   18 - LOADNIL       12   16    0
(  59)   19 - TEST          11   11    0
(  59)   20 - JMP            0   18
(  59)   21 - GETUPVAL      17    5    0
(  59)   22 - MOVE          18   11    0
(  59)   23 - CALL          17    2    2
(  59)   24 - TEST          17   17    0
(  59)   25 - JMP            0   13
(  61)   26 - GETUPVAL      17    2    0
(  61)   27 - MOVE          18   11    0
(  61)   28 - CALL          17    2    4
(  61)   29 - MOVE          14   19    0
(  61)   30 - MOVE           9   18    0
(  61)   31 - MOVE          13   17    0
(  62)   32 - GETUPVAL      17    3    0
(  62)   33 - MOVE          18   11    0
(  62)   34 - CALL          17    2    4
(  62)   35 - MOVE          16   19    0
(  62)   36 - MOVE           9   18    0
(  62)   37 - MOVE          15   17    0
(  62)   38 - JMP            0   10
(  65)   39 - GETUPVAL      17    6    0
(  65)   40 - MOVE          18    0    0
(  65)   41 - CALL          17    2    2
(  65)   42 - MOVE          12   17    0
[...]

PR #4162, opened 2022-08-21
Re-introduce `table.insert` over using a separate counter

Garanas:
The table was used by multiple barrels separately, they were overwriting each others manipulators. See also https://github.com/FAForever/fa/pull/4064/files , and in particular this change: https://github.com/FAForever/fa/pull/4064/files#diff-29a4964e78571ea3d37fe1cd130083a802cf3a9341d1909864d9f9445af592f6R498


Garanas:
image

image

PR #4165, opened 2022-08-22
Guard `Blueprint.General` for command caps

Hdt80bro:


lua/system/blueprints-units.lua
```diff
@@ -193,7 +193,10 @@ local function PostProcessUnit(unit)
 
     -- # Fix being able to check for command caps
 
-    unit.General.CommandCapsHash = table.deepcopy(unit.General.CommandCaps)
+    local unitGeneral = unit.General
+    if unitGeneral and unitGeneral.CommandCaps then
+        unitGeneral.CommandCapsHash = table.deepcopy(unitGeneral.CommandCaps)
+    end
```
Garanas, line 199: needs an else, so that there's at least an empty table

Hdt80bro: Is that what you're thinking of?

PR #4167, opened 2022-08-23
Mods manager improvements

4z0t:
- [x] After closing mods manager being not host you can no longer type in chat
- [x] Cleanup pattern matching


Garanas:
best to open a new one now

PR #4168, opened 2022-08-23
Fix internal state of HQs

Garanas:


Garanas:
Woops - I was on fafbeta 😄

PR #4175, opened 2022-08-25
Add PBR blender shader nodes

BlackYps:
I hope I make this PR against the right branch.
This adds a blender file that contains shader nodes to use the textures of the game units in a PBR shader workflow.
The shaders are still work in progress, but pretty usable already.


Garanas:
Branch is correct! I didn't quite expect the file to be so big, it is 83 megabytes. Are textures and models included too?

BlackYps:
Yes, I included some models, because the textures vary between models and then I packed all the textures, so the materials don't break when you transfer the file to a different machine.
I think having all these assets is convenient if one wants to play around with the shader. If we just want to save the shader for posterity, we could do with one unit per faction.

BlackYps:
UEF, Cybran and Aeon are now done for the moment. It might be that it looks odd on a unit that I have not tested, but it is too much work to import every single model. It's probably smarter to correct it once an issue with a unit surfaces.
The Seraphim shader is still a bit of a mess, but so is the in-game shader. It primarily needs an artistic direction how the Seraphim units should look like, before work can continue there.
From my point of view this can be merged now.

Garanas:
@MrRowey could you try the list of instructions in this PR?

MrRowey:
> @MrRowey could you try the list of instructions in this PR?

Will try tomorow

Garanas:
Perhaps, due to the size, it is better to make a PR for this repository:

- https://github.com/FAForever/Art-Assets

BlackYps:
I just checked: A file with the nodes only would only be around 2 MB big. If we include one small unit per faction it's maybe 6 MB. The amount of units in the current file is not necessary. I just left them in because I saw no reason to remove them

Garanas:
Waiting for Rowey to report back and then we're good.

MrRowey:
image
@Garanas Instruction work :)

BlackYps:
I wrote the guide with basic blender proficiency in mind. That was probably a bit too optimistic. Thank you for detailing the guide. Might as well add how to correctly scale the model too. I will do that later today, but if you want to you can do that too

MrRowey:
Feel free to add the as id rather someone who know blender a bit more to do it right way xD

MrRowey:
Oh also based on these instruction i did create it in to a video format here for people who prefer that https://youtu.be/pq9os0NhfB4

BlackYps:
There seems to be a volume cutoff for your mic that is overtuned. The audio is frequently missing for a moment.

BlackYps:
If you wait with re-recording the video until this PR is merged, then you can show exactly where to get the files from. Also, a blender tip: when navigating for the textures, there is a recently used folders list in the bottom left corner that you can just click. Then you don't need to manually navigate to the correct unit folder every time.

Garanas:
Could you add the video to the PR? We can always update it in the past again. After that I'll merge it

BlackYps:
I just pushed a final update to the shaders. I feel they are finished now.

PR #4176, opened 2022-08-25
Annotation work /File Cleaning part 5

MrRowey:
Last Files for lua/sim file
- Projectile.lua
- AttackMove.lua
- EnhanceTasklua
- GiveTask.lua
- TargetLocation.lua


lua/sim/Projectile.lua
```diff
@@ -225,7 +225,7 @@ Projectile = Class(moho.projectile_methods) {
 
     --- Called by the engine when the projectile is killed, in other words: intercepted
     ---@param self Projectile
-    ---@param instigator Unit
+    ---@param instigator Projectile
```
Hdt80bro, line 228: The instigator is the unit that launched the projectile, so this should still be `Unit`

PR #4180, opened 2022-08-25
refactor AI economy conditions

relent0r:
## Description of the changes
This PR replaces some of the default economy condition functions. By default all conditions perform a call to the function aiutilities.lua AIGetEconomyNumbers. This function runs 13 economy functions and does a few other calculations then returns a table. The economy condition then takes the economy property its interested in from this.
For instant build conditions this gets called each time the build condition monitor performs its condition check. This PR removes the excess calls by making the build conditions only call the functions they need for the data required.

## Test setup for the changes
Perform an AI game and record the number of times the economy based functions are called using the built in profiler. Then rerun the same game via replay with the refactored code. Ensure no desync due to errors and compare the number of times the functions are called. The refactored set should be less by a number of thousands over a period of time.

Note : I haven't touched the EconOverTime functions or StorageMax functions as these require extra calculations that are done within the AIGetEconomyNumbers function. There are also a bunch of function calls in the file SorianInstantBuildConditions. Some are custom and others are duplicates, would recommend removing the duplicates and re pointing the builders to the default conditions via a different PR.

This is the number of calls made prior to refactor over a 10 minute game.
FunctionCallsPreRefactor

This is the number of calls made post refactor across the same 10 minute game.
FunctionCallsPostRefactor

### DIFF
--- a/lua/editor/EconomyBuildConditions.lua
+++ b/lua/editor/EconomyBuildConditions.lua
@@ -14,7 +14,7 @@ local GetEconomyStoredRatio = moho.aibrain_methods.GetEconomyStoredRatio
 local GetEconomyIncome = moho.aibrain_methods.GetEconomyIncome
 local GetEconomyRequested = moho.aibrain_methods.GetEconomyRequested
 local GetEconomyStored = moho.aibrain_methods.GetEconomyStored
-local ParagonCat = categories.ENERGYPRODUCTION * categories.EXPERIMENTAL * categories.STRUCTURE
+local ParagonCat = categories.STRUCTURE * categories.EXPERIMENTAL * categories.ECONOMIC * categories.ENERGYPRODUCTION * categories.MASSPRODUCTION
 
 local AIUtils = import('/lua/ai/aiutilities.lua')
 
--- a/lua/editor/EconomyBuildConditions.lua
+++ b/lua/editor/EconomyBuildConditions.lua
@@ -8,6 +8,13 @@
 --**
 --**  Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
 --****************************************************************************
+
+local GetEconomyTrend = moho.aibrain_methods.GetEconomyTrend
+local GetEconomyStoredRatio = moho.aibrain_methods.GetEconomyStoredRatio
+local GetEconomyIncome = moho.aibrain_methods.GetEconomyIncome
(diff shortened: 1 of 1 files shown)


lua/editor/EconomyBuildConditions.lua
```diff
@@ -237,7 +233,7 @@ end
 ---@param EnergyEfficiency integer
 ---@return boolean
 function GreaterThanEconEfficiencyOverTime(aiBrain, MassEfficiency, EnergyEfficiency)
-    if HaveGreaterThanUnitsWithCategory(aiBrain, 0, 'ENERGYPRODUCTION EXPERIMENTAL STRUCTURE') then
+    if HaveGreaterThanUnitsWithCategory(aiBrain, 0, categories.ENERGYPRODUCTION * categories.EXPERIMENTAL * categories.STRUCTURE) then
```
Garanas, line 236: You can cache these categories too, as an upvalue. See also: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/benchmarking/benchmarks/categories.lua

Garanas:
Good use of the profiler to back up your changes 👍 . The changes appear good to me, @Uveso could you glance over it too?

lua/editor/EconomyBuildConditions.lua
relent0r: good point, I'll update to include that.

Uveso:
looks good.

i would use as paragon cat this:
```categories.STRUCTURE * categories.EXPERIMENTAL * categories.ECONOMIC * categories.ENERGYPRODUCTION * categories.MASSPRODUCTION```

only using ENERGYPRODUCTION EXPERIMENTAL STRUCTURE will lead to some buildings that are recognized as paragon but are only modded energy structures like the Black Ops "Philadelphia" (Stellar Generator)
There are also some T4 energy generator mods that will also make problems.

relent0r:
> categories.STRUCTURE * categories.EXPERIMENTAL * categories.ECONOMIC * categories.ENERGYPRODUCTION * categories.MASSPRODUCTION

Ok will update to use this.

Garanas:
Is the check specifically for the paragon? You can also check for the unit id, all unit ids are categories too.

Uveso:
This check should be for paragon like units, not paragon only.

With unit ID this would not recognize units from my "Quantum Resource Generators" mod, "paragon for all" mod etc.

PR #4181, opened 2022-08-25
Add annotation classes

Hdt80bro:
Adds most of the non-game classes to the annotation system. This clashed the Maui and UI checkboxes, so the class in `/lua/ui/control/checkbox.lua` has been renamed to `UICheckbox`. Also made all Maui imports non-relative. And restored the copyright symbols...

Most importantly, and the entire reason I started this, that awful `Categories` class that keeps popping up has been finally removed.


lua/maui/histogram.lua
```diff
@@ -14,7 +14,7 @@ Class methods:
             [2] = ...
         }
 --]]
-local Control = import('control.lua').Control
+local Control = import('/lua/maui/control.lua').Control
 
 ---@class Histogram : moho.histogram_methods, Control
```
Garanas, line 19: this one should be flagged as deprecated, as it does not work

lua/maui/mauiutil.lua
```diff
@@ -3,15 +3,15 @@
...
 
-local Checkbox = import('checkbox.lua').Checkbox
+local Checkbox = import('/lua/maui/checkbox.lua').Checkbox
 
 -- Create a radio group, which is a group of checkboxes and associtated labels that have
 -- a mutually exclusive checked state.
-RadioGroup = Class
-{
+---@class RadioGroup
+RadioGroup = Class() {
```
Garanas, line 14: Should be a SimpleClass, for consistency

Garanas (changes requested):
Minor feedback, feel free to merge as you see fit after processing / discussing it.

lua/maui/histogram.lua
Garanas: (the class)

PR #4187, opened 2022-08-28
Strategic view

Garanas:
image

Experiments with an alternative to the usual dual view. Functionality that is still required:
- [ ] Being able to choose the left or right side
- [ ] Being able to track units in the separate views
- [ ] Being able to swap the main view with a side view (and back)
- [ ] Being able to bind a camera position to a side view
- [ ] Being able to lock a side view (camera wise)
- [ ] Make it look better by adding some black border between the views

And preferably, but proven difficult:
- [ ] The UI should only be in the main view, and not overlap with the side views


MrRowey:
@Garanas can you close if you don't intend to carry on with it or is anyone else willing to look at finishing it off ?

Garanas:
No, we'll close it 😃 !

PR #4189, opened 2022-08-29
Increase muzzle velocity and aoe for uef and seraphim gunship

SomeSwkoll:
This fixes the issue where a single flak can circle dodge vs uef and seraphim gunships and take no damage. In my tests muzzle velocity increases alone were insufficient to counter circle dodge so a combination of aoe increases and muzzle velocity increases to achieve the desired result.

Muzzle velocity 40 -> 60
Damage radius 0 -> 1

### DIFF
--- a/units/UEA0203/UEA0203_unit.bp
+++ b/units/UEA0203/UEA0203_unit.bp
@@ -280,6 +280,7 @@ UnitBlueprint {
             BallisticArc = 'RULEUBA_None',
             CollideFriendly = false,
             Damage = 20,
+            DamageRadius = 1,
             DamageType = 'Normal',
             DisplayName = 'Hells Fury Riot Gun',
             FireTargetLayerCapsTable = {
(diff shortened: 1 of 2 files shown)


Garanas:
The branch should point to `deploy/fafbeta`.

Note that increasing the muzzle velocity increases the chances of having the collision bug. Especially the t1 Aeon artillery is prone to that: if a projectile ends up colliding with multiple things at once in a frame, it ends up picking the last thing it hit. For a gunship shooting at the t1 Aeon artillery this is the ground, which is why some units are really good at 'evading damage while walking in a straight line'

Giving it a damage radius eliminates the problem. My point is that these changes require each other - if you increase the muzzle velocity, a damage radius (to some degree) is required.

4z0t:
Giving aoe allows killing reclaim and dealing damage to stacked units which might be not as intended. May be very effective against engies groups as well.

PR #4193, opened 2022-08-30
Hive & Kennel Rebalance

FemtoZetta:
Kennels currently are very bad and basically no one uses them. Hives are too strong in some cases and provide better versatility in the late game than engineers can provide (like instantly changing your bp to something different; high concentration of bp in a small area). Recently we had a few discussions about it on Discord and that's why I decided to make an attempt at rebalancing.

This rebalance aims to make the Kennel more useful and makes losing drones a lot less punishing. With the better build power Kennels are now more mass efficient than Hives (still a lot less than the current Hives though), and losing the drones isn't so bad anymore. The health buff makes them able to at least take 1 or 2 hits from AA without dying like before.

Hives get a big build power nerf (still better than current Kennel) and a slight build range nerf. This should make them worse at being super concentrated build power, but they should still be a useful building for the lategame where mass efficiency is less important than versatility.

Old Hive range:
grafik

New Hive range:
grafik

### DIFF
--- a/units/XEA3204/XEA3204_unit.bp
+++ b/units/XEA3204/XEA3204_unit.bp
@@ -110,8 +110,8 @@ UnitBlueprint {
         AirThreatLevel = 0,
         ArmorType = 'Light',
         EconomyThreatLevel = 1,
-        Health = 50,
-        MaxHealth = 50,
+        Health = 100,
+        MaxHealth = 100,
         RegenRate = 0,
         SubThreatLevel = 0,
         SurfaceThreatLevel = 0,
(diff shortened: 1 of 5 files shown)


4z0t:
I understand those changes: you are trying to make equal these 2 units. But changes you want to do are very sharp and crucial. Hives have their advantage as being very strong stationary build power, on the other hand kennels allow their drones be everywhere on the map and reach things engies cannot due to their collision model. If you want to buff kennels or nerf hives you'd need to do it smoothly. And yeah, balance discussion should be done somewhere else, but I hope you got the point.

Garanas:
@FemtoZetta best to rebase using the latest deploy/fafbeta

Garanas:
Great - @Tagada14 all yours

Tagada14:
For now we will be going with a smaller adjustment (https://github.com/FAForever/fa/pull/4216), just giving Kennels their deserved buff. In the future, probably in a couple of months, we will be looking further into Hives and Kennels.
I will leave the PR open for now and revisit it in around a month.

Tagada14:
#4644
Decided to go with a different set of changes

PR #4197, opened 2022-09-03
Let sparkies get radar/sonar

Penguin5:
Enable sparkies to build t1/t2 UEF radar and sonar


maudlin27:
I get the radar addition (even if I'd go for T1 radar or boosting sparky's built in radar instead) since you need intel for combat buildings to work, but what has building mexes got to do with a combat engineer theme? Building resource generating structures feels like it should remain with the core engineers.

Also why radar but not sonar? The unit itself is amphibious and it can build torpedo launchers so if it's considered 'in-theme' to let it build radar then it feels like it should get sonar as well.

Garanas:
I agree, let it build radar and sonar but leave out the extractors.

Garanas:
@Penguin5 best to rebase using the latest `deploy/fafbeta`

Tagada14:
The PR view is wrong despite the rebase, I had the same thing happen to me yesterday, and only closing the PR and creating a new one helped

PR #4199, opened 2022-09-03
Terrain scanning feature

Garanas:
todo 🤠


Garanas:
This is less trivial than we thought it was initially. There is a large delay, which requires us to have some smart algorithm to predictively scan bits of the height map.

PR #4205, opened 2022-09-08
OnMissileImpactTerrain changes required for functional callback

relent0r:
## Description of the changes

This PR corrects a small typo and makes some changes to enable callback functionality for the previously created callback `OnMissileImpactTerrain`. I'm not 100% on initializing the self.Callbacks table in the OnCreate function for the unit class but it seemed like the way to go.
I'll work my way through testing the other OnMissile callbacks against the AI when I can get time.

## Test setup for the changes
This PR was tested against the RNGAI sim mod by the following method.
Within the tactical missile platoon function it will loop through all available launchers.
It will then perform the following.
```
                if not tml.terraincallbackset then
                    tml:AddMissileImpactTerrainCallback(missileTerrainCallbackRNG)
                    tml.terraincallbackset = true
                end
```

The callback function

```
function MissileCallbackRNG(unit, targetPos, impactPos)
    if unit and not unit.Dead and targetPos then
        if not unit.TargetBlackList then
            unit.TargetBlackList = {}
        end
        unit.TargetBlackList[targetPos[1]] = {}
        unit.TargetBlackList[targetPos[1]][targetPos[3]] = true
        return true, "target position added to tml blacklist"
    end
    return false, "something something error?"
end
```

When the tactical missile platoon is evaluating targets it will perform the following check against each tactical missile launcher is has which allows it to skip the target if its position is blacklisted from that particular launcher.
```
if tml.TargetBlackList then
        if tml.TargetBlackList[targetPosition[1]][targetPosition[3]] then
              RNGLOG('TargetPos found in blacklist, skip')
              continue
        end
end
```

Attached is the visual of the terrain block check performed. The red T2 extractor is behind terrain. The AI's TML fires a missile, after hitting the terrain the launcher platoon will add the blacklisted position to the launcher unit and will no longer fire missiles at it.

MissileLaunch

Logs were generated to confirm the launcher had added the target position to its blacklist and when the launcher skipped the target during target acquisition.
Logs generated within callback validating launcher UnitId, target position and impact position.
```
info: MissileCallbackRNG
info: xsb2108
info: 
info:  - 1: 199.5
info:  - 2: 21.578125
info:  - 3: 266.5
info: 
info:  - 1: 205.69364929199
info:  - 2: 27.374061584473
info:  - 3: 249.92608642578
info: TargetPos Added to blacklist
```

Logs generated when blacklisted unit position found.
```
info: TargetPos found in blacklist, skip
```

### DIFF
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -266,7 +266,7 @@ Projectile = Class(moho.projectile_methods) {
 
         -- callbacks for launcher to have an idea what is going on for AIs
         local categoriesHash = self.Blueprint.CategoriesHash
-        if categoriesHash['TACTICAL'] or categoriesHash['STRATEGICAL'] then
+        if categoriesHash['TACTICAL'] or categoriesHash['STRATEGIC'] then
             -- we have a target, but got caught by terrain
             if targetType == 'Terrain' then
                 if not IsDestroyed(self.Launcher) then
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -103,7 +103,7 @@ local cUnit = moho.unit_methods
 ---@field Army Army
 ---@field UnitId UnitId
 ---@field EntityId EntityId
----@field EventCallbacks function[]
+---@field EventCallbacks table<string, function[]>
 Unit = Class(moho.unit_methods) {
 
     Weapons = {},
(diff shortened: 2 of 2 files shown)


Garanas:
This looks good, apologies for the mistakes that were made. I'll review this during the weekend!

Garanas:
It took one weekend longer, but here we are. I messed up the naming convention, it should've been `EventCallbacks` and not `Callbacks`. I fixed that, together with your changes it looks good now.

PR #4206, opened 2022-09-08
Fix props breaking bomber targeting

Hdt80bro:
* Props no longer break the targeting of bombers
* The salvo data is now reset once the salvo is done
* Improved free-falling of bomb clusters in the failure mode
* More hyper-optimization of `CalculateBallisticAcceleration`
* Some annotations


Hdt80bro:
Wrong base

PR #4210, opened 2022-09-08
Optimize lazyvar & Cleanup annotations

Hdt80bro:


lua/lazyvar.lua
```diff
@@ -1,125 +1,229 @@
...
+        end
     end
 end
 
+--- Sets the value
 function LazyVarMetaTable:SetValue(value)
-    local dirtyList = {}
-    self:SetDirty(dirtyList)
+    if value == nil then
+        error("You are attempting to set a LazyVar's value to nil, don't do that!")
```
Garanas, line 138: this stops execution, the `value = 0` line is never executed. Better to use warnings if you intend execution to continue

Hdt80bro: That's what I thought, but I've seen execution continue. `pcall` can do some weird things.

4z0t:
Why not separate PRs? 😩

Hdt80bro:
It's the annotation creep... I intended to do just the lazyvar stuff, but then I saw easy annotation work that applied to that, and then there was a little more... at the end, I really didn't want to go through and remove all of the annotations (which generally shouldn't problematic anyway).

lua/ui/game/Profiler.lua
```diff
@@ -148,7 +148,10 @@ list = false
 
 --- Opens up the window
 function OpenWindow()
-
+    local t = import("/lua/maui/group.lua").Group(GetFrame(0))
+    LOG(t._c_object)
```
Garanas, line 152: There are LOG statements here

lua/ui/game/commandmode.lua
```diff
@@ -112,7 +112,7 @@ end
 function EndCommandMode(isCancel)
 
     --- ???
-    if ignoreSelection then
+    if ignoreSelection or not modeData then
```
Garanas, line 115: What does this change do?

lua/ui/game/orders.lua
```diff
@@ -982,7 +982,7 @@ end
 --- initialStateFunc is a function that gets called once the control is created and allows you to set the initial state of the button
 ---      the function should have this declaration: function(checkbox, unitList)
 --- extraInfo is used for storing any extra information required in setting up the button
----@type table<string, OrderInfo>
+
```
Garanas, line 985: This seems incorrect, why remove the type?

lua/ui/lobby/data/watchedvalue/watchedvaluetable.lua
```diff
@@ -42,14 +42,16 @@ WatchedValueTable = ClassSimple {
...
                 -- limit logging only to changes of the WatchedValueTable
-                if LoggingEnabled and not LoggedChanges[msg] then
-                    LoggedChanges[msg] = true
-                    LOG(msg)
+                local value = _store[key]
+                if LoggingEnabled then
+                    local msg = 'WatchedValueTable __index function(wvt, ' .. repr(key) .. ') '  .. tostring(value)
+                    if not LoggedChanges[msg] then
+                        LoggedChanges[msg] = true
+                        LOG(msg)
```
Garanas, line 51: I don't see why we'd want to LOG these changes

lua/version.lua
```diff
@@ -1,4 +1,7 @@
-local Version = '3744'
+local Version = "3744"
+---@alias PATCH "3744"
+---@alias VERSION "1.5.3744"
```
Garanas, line 3: More work when doing a release!!! 😄

lua/system/trashbag.lua
```diff
@@ -24,21 +24,21 @@
...
     -- http://lua-users.org/wiki/GarbageCollectionTutorial
     -- http://lua-users.org/wiki/WeakTablesTutorial
     __mode = 'v',
 
-    --- Add an entity to the trash bag.
-    Add = function(self, entity)
-
+    --- Adds an entity to the trash bag
+    ---@param self TrashBag
+    ---@param trash Destroyable
```
Garanas, line 40: Now I understand the new base class, good call 👍

lua/lazyvar.lua
```diff
@@ -1,125 +1,229 @@
...
+    ---@type LazyVar
+    ---@diagnostic disable-next-line:assign-type-mismatch,miss-symbol,exp-in-action,unknown-symbol
+    local result = {&4 initial} -- preallocate table with hashsize=4, arraysize=1
     setmetatable(result, LazyVarMetaTable)
-    if initial == nil then 
-        result[1] = 0
-    else
-        result[1] = initial
+    local WeakKeyMeta = WeakKeyMeta
+    do
```
Garanas, line 220: why is this in a `do ... end` block?

lua/version.lua
Hdt80bro: yup! Luckily they're all in the same place.

lua/lazyvar.lua
Hdt80bro: It reduces the stacksize by putting `used_by` out of scope by the time `uses` gets used and thus they end up taking the same register.

lua/system/trashbag.lua
Hdt80bro: This meant that lazyvar & co needed to be updated to support `OnDestroy` for consistency. I also made the lazyvar destructor destroy it's value, though I'm not entirely sure that's safe.

lua/ui/lobby/data/watchedvalue/watchedvaluetable.lua
Hdt80bro: I was so disappointed with how that file was written when I found it. I started to completely rewrite it, but managed to convince myself that it wasn't worth breaking the Lobby in the process when we're only a month from redesigning it. Instead, I painfully left each and every behavior as I found it.

lua/ui/game/Profiler.lua
Hdt80bro: Ah, good catch. Thought I removed those...

lua/ui/game/commandmode.lua
Hdt80bro: It fixes a latent bug I discovered where it ends up writing fields to `false`. I found it because I added the metatable cleanup operation to booleans.

lua/ui/game/orders.lua
Hdt80bro: Manually typing tables shadows any information about their values which can be important since each `OrderInfo` can have a different set of fields. This does mean that table accesses to it aren't typed as an `OrderInfo`, but since it's a local variable used only twice--and through a `table.deepcopy` operations that strip all table annotation types at that--none of it matters anyway. In general, I prefer to rely on implicit typing when it improves readability.

lua/system/trashbag.lua
Hdt80bro: `Destroyable` also represents how the engine handles controls by the way

lua/ui/game/commandmode.lua
Garanas: I see now, good call

lua/ui/lobby/data/watchedvalue/watchedvaluetable.lua
Garanas: Understood 😄 , I am proud of you

lua/system/trashbag.lua
Garanas: The value is usually a number or a function. I don't think those can be destroyed

lua/lazyvar.lua
Garanas: And there's no overhead of using that?

lua/ui/game/orders.lua
Garanas: Fair enough 👍

lua/system/trashbag.lua
Hdt80bro: Yeah, I check for it to exist first before doing that--https://github.com/FAForever/fa/pull/4210/files#diff-400e5297b564eef68495418887792b9916dd8d86033ff35cb08c069416aaa4e6R193-R205

lua/lazyvar.lua
Garanas: That is interesting to know 😄 !

Hdt80bro:
@Garanas how are you feeling about this one?

PR #4211, opened 2022-09-09
Using the DifferentialUpgradeCostCalculationCost for Radar

MrRowey:
using the `DifferentialUpgradeCostCalculationCost` for Radar to make the reclaim more sense the Mass / Energy Values can be debated by the balance team. currently I've just added the previous tier to the next to have the same cost as they are now if uses the upgrade mechanic and makes more expensive if built straight.

### DIFF
--- a/units/UAB3104/UAB3104_unit.bp
+++ b/units/UAB3104/UAB3104_unit.bp
@@ -132,10 +132,11 @@ UnitBlueprint {
         UniformScale = 0.25,
     },
     Economy = {
-        BuildCostEnergy = 30000,
-        BuildCostMass = 2400,
+        BuildCostEnergy = 3432,
+        BuildCostMass = 2660,
         BuildTime = 2575,
         MaintenanceConsumptionPerSecondEnergy = 2000,
+        DifferentialUpgradeCostCalculationCost = true;
         RebuildBonusIds = {
             'uab3104',
         },
--- a/units/UAB3201/UAB3201_unit.bp
+++ b/units/UAB3201/UAB3201_unit.bp
@@ -136,14 +136,15 @@ UnitBlueprint {
         UniformScale = 0.25,
     },
     Economy = {
-        BuildCostEnergy = 3600,
-        BuildCostMass = 180,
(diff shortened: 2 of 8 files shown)


Hdt80bro:
Two code notes, you dropped a zero for `UAB3104`'s energy cost, and it's `DifferentialUpgradeCostCalculation` instead of `DifferentialUpgradeCostCalculationCost`.

In terms of balance, I suspect that sonar should be included if this is something we decide to do. However, I'm not sure it is. Currently, only factories and engineering stations use differential costs.

For factories, it's a way to encourage upgrading so that you don't self-destruct a useless lower tier factory for the reclaim if you build it as a new factory, or end up with less total build power if you upgrade instead building it as a new factory (which would cost the same without the differential calculation)--all without messing with the reclaim value of the factory.

For engineering stations, you can't build the higher tiers, so the differential cost is purely there to make the cheap upgrade not reduce its reclaim value by making it equal to the total resources put into the building.

The rest of the upgradable buildings can be built directly, but have reasons to cost the same as building a new structure. For example, shields and mass extractors are both location specific (building a new shield doesn't cover the same area so placement density is a concern, and mass extractors can't be placed willy-nilly--they need to be upgraded) and, more importantly, both function while upgrading. This means that you can immediately gain the benefits of the lower tier building before you could build the higher tier one, but you sacrifice the cost of that lower tier building to achieve that. Radars and sonars behave the same way.

In my opinion, this last fact is the most important. Radars, sonars, mass extractors, and shields all provide their benefit while upgrading, while factories and Hives do not which is the difference (Kennels *usually* can, but again, the upgraded version can't be directly built, so it's a function of making the cheap upgrade cost not impact the reclaim value).

MrRowey:
@Hdt80bro as long as balance team is happy with this one ill do the sonar one

Tagada14:
I don't see a reason for adding this, T1 and T2 radars don't leave wrecks anyway (because they get overkilled), so this would only affect OMNI. This would also mess with the current values so I see only downsides and creating more work for the balance team without any upside for the change

PR #4214, opened 2022-09-13
Reorder Lobby Options

Penguin5:
Just a QoL change that makes share condition the top option in its options list and allow observers third (as opposed to both being lower down on the list)


Garanas:
Less scrolling - good thinking

PR #4198, opened 2022-09-03
Kennel Buff

Penguin5:
Decreases kennel mass cost by 20%
Decreases drone rebuild mass and energy costs by 90%

PR #4216, opened 2022-09-13
Nerf Kennel Costs

Penguin5:
Decreases kennel mass cost by 20%
Decreases drone rebuild mass and energy costs by 60%
Decrease drone rebuild time by 33.33%
Increases drone elevation from 2 to 3
(replacement of #4198)

### DIFF
--- a/units/XEA3204/XEA3204_unit.bp
+++ b/units/XEA3204/XEA3204_unit.bp
@@ -110,8 +110,8 @@ UnitBlueprint {
         AirThreatLevel = 0,
         ArmorType = 'Light',
         EconomyThreatLevel = 1,
-        Health = 6,
-        MaxHealth = 6,
+        Health = 50,
+        MaxHealth = 50,
         RegenRate = 0,
         SubThreatLevel = 0,
         SurfaceThreatLevel = 0,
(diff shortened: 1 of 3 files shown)


Hdt80bro:
Looking at other drone costs, the Kennel drone definitely needed to be cheaper. 90% is a *huge* decrease though--what's the thought process behind making it cost half as much as an air scout?

Penguin5:
For perspective, the other recent PR that rebalances kennels (#4193) actually reduces drone rebuild mass and energy costs by 94% and gives them 100 health... Drone rebuild costs are just so ridiculous at present that my PR is actually the more conservative one.

Tagada14:
@Penguin5 Could you decrease the buff to the Kennel drones please?
I would keep it at 1kE and 100 mass for now. Also decrease the BT to 500.
Could you also look into increase their elevation a bit? It's an idea to make them less vulnerable to random AoE.
Increasing their hp to 50 should also be a small QoL buff while not going over board.

Penguin5:
@Tagada14 I made the changes you requested.

Tagada14:
Looks good, have you tested how big of an effect the elevation change has?

Penguin5:
I did basic testing of it. It can be changed further if you prefer. I changed the elevation value in the bp from 2 to 3.  Here's how it looks:  
image

image

PR #4219, opened 2022-09-14
Successive upgrades in hotbuild via keybind

ChessBerry:
Hello everyone, 

this PR is a proposed feature addition to hotbuild, namely being able to queue successive upgrades using the upgrade keybind, instead of having to use a mouse click.

Upgrading e.g. a T1 factory to T3 in one go previously required a mouse click to queue the upgrade to T3, if the upgrade to T2 was not finished. 
Now, pressing the upgrade hotkey on your keyboard a second time does queue the upgrade to T3, regardless of if the T2 upgrade is finished or not.

I tested it with factories, radars and sonars of different factions, and even with adding units into a factory queue between the T2 and T3 upgrade and it seems to work as intended.

TODO: I don't know how to adjust the icon in the command card so that it shows the correct keybind as, right now, the T3 upgrade doesn't show a keybind at all if the T2 upgrade is queued but not finished yet.

Tell me what you think :)

### DIFF
--- a/lua/keymap/hotbuild.lua
+++ b/lua/keymap/hotbuild.lua
@@ -577,13 +577,13 @@ end
 -- Doing this fully recursively is probably overkill as the support fac / HQ is the only time when there are more than
 -- two possible upgrade paths, and the cybran shield is the only building with more than two successive upgrades, but
 -- hacking it together didn't feel right, hence the 'better' recursive solution.
-function issueUpgradeCommand(upgrades, buildableCategories)
+function IssueUpgradeCommand(upgrades, buildableCategories)
     local success = false
     for _, upgr in upgrades do
         local bp = __blueprints[upgr]
         successiveUpgrades = upgradeTab[bp.BlueprintId] or {bp.General.UpgradesTo}
         if successiveUpgrades then
-            success = issueUpgradeCommand(successiveUpgrades, buildableCategories)
+            success = IssueUpgradeCommand(successiveUpgrades, buildableCategories)
             if success then
                 break
             end
@@ -610,11 +610,11 @@ function buildActionUpgrade()
         local bp = unit:GetBlueprint()
         -- If upgradeTab[bp.BlueprintId] returns a table, there are two or more possible upgrades, e.g. HQ and 
         -- support factory. If instead bp.General.UpgradesTo returns a string, then there is only one possible 
-        -- upgrade, e.g. for a radar. For our helper function issueUpgradeCommand we want a table, hence the { } 
+        -- upgrade, e.g. for a radar. For our helper function IssueUpgradeCommand we want a table, hence the { } 
         -- brackets
         local upgrades = upgradeTab[bp.BlueprintId] or {bp.General.UpgradesTo}
         SelectUnits({unit})
-        local success = issueUpgradeCommand(upgrades, buildableCategories)
+        local success = IssueUpgradeCommand(upgrades, buildableCategories)
         if not success then
             result = false
         end
--- a/lua/keymap/hotbuild.lua
+++ b/lua/keymap/hotbuild.lua
@@ -607,7 +607,6 @@ function buildActionUpgrade()
                 -- upgrade is already queued.
                 IssueBlueprintCommand("UNITCOMMAND_Upgrade", second_upgrade, 1, false)
                 success = true
-                break
             elseif EntityCategoryContains(buildableCategories, cmd) then
                 -- Queue the upgrade of the selected structure
                 IssueBlueprintCommand("UNITCOMMAND_Upgrade", cmd, 1, false)
--- a/lua/keymap/hotbuild.lua
+++ b/lua/keymap/hotbuild.lua
@@ -579,8 +579,8 @@ end
 -- hacking it together didn't feel right, hence the 'better' recursive solution.
 function IssueUpgradeCommand(upgrades, buildableCategories)
     local success = false
-    for _, upgr in upgrades do
-        local bp = __blueprints[upgr]
+    for _, u in upgrades do
+        local bp = __blueprints[u]
         local successiveUpgrades = upgradeTab[bp.BlueprintId] or {bp.General.UpgradesTo}
         if successiveUpgrades then
             success = IssueUpgradeCommand(successiveUpgrades, buildableCategories)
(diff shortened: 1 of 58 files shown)


Garanas:
Looks like a good quality of life change. What is the behavior that you describe in the TODO? I don't entirely understand what it means / what you're trying to achieve. A screenshot can help.

ChessBerry:
Let's say I have hotbuild-upgrade bound to M and I select a T1 factory. I will see the following in my command card (or whatever it's called):
Forged Alliance 14_09_2022 14_23_30 (2)
(Ignore the keybinds for all the other units)

If I now queue the upgrade to a T2 HQ, I will get access to the T2 build suite for unit queueing purposes, but notice that there is no keybind for the upgrade to the T3 HQ on the T3 HQ icon:
Forged Alliance 14_09_2022 14_23_40 (2)

After the upgrade to T2 is done, the T3 icon _does_ get the appropriate keybind:
Forged Alliance 14_09_2022 14_23_52 (2)

What is TODO is to give the icon in the intermediate step, when the T2 upgrade is queued but not finished, where I drew the red arrows, to correct keybind symbol, in this case the letter M. 
I don't know how to do that do as I haven't been able to find the appropriate place in the code governing these icon changes

4z0t:
Afaik it looks what blueprints it can build and upgrade into in construction.lua (have a nice day looking through __this__ code...)

ChessBerry:
Took me a while, but I found where the hotbuild upgrade icons are set in gamemain.lua. The icons and successive upgrades should now work correctly, even for factories or unconventional structures, like the four times upgradeable cybran shield.

Garanas:
Exciting, not sure why that code is there. I'll look at this tonight, see if I can break it 😄

ChessBerry:
I phrased that a bit misleadingly: The icons themselves are not directly set in gamemain.lua, but the table that governs what is a valid upgrade target (according to hotbuild) is.

Also yes, please break it, there is for sure stuff I've overlooked that could be improved :P

Garanas:
It doesn't work for build mode, but I think I'm the only user of that. @Hdt80bro do you happen to use hotbuild, and if so can you review this?

Hdt80bro:
Sorry, I don't

Garanas:
@4z0t do you use hotbuild, if so can you review this?

4z0t:
Yes, It works! Nice job!

lua/keymap/hotbuild.lua
```diff
@@ -571,32 +571,50 @@ function buildActionUnit(name, modifier)
...
+-- Recursively finds and tries to order possible upgrades from highest tech to lowest tech. This allows queuing upgrades
+-- successively using keybinds, even if the previous upgrade isn't finished yet.
+-- Doing this fully recursively is probably overkill as the support fac / HQ is the only time when there are more than
+-- two possible upgrade paths, and the cybran shield is the only building with more than two successive upgrades, but
+-- hacking it together didn't feel right, hence the 'better' recursive solution.
+function issueUpgradeCommand(upgrades, buildableCategories)
+    local success = false
+    for _, upgr in upgrades do
+        local bp = __blueprints[upgr]
+        successiveUpgrades = upgradeTab[bp.BlueprintId] or {bp.General.UpgradesTo}
```
4z0t, line 584: Did you want it to be `local` variable?

lua/keymap/hotbuild.lua
```diff
@@ -571,32 +571,50 @@ function buildActionUnit(name, modifier)
...
 end
 
--- Does not upgrade T1 facs that are currently upgrading to T2 to T3 when issued
+-- Helper function for the buildActionUpgrade() function below. 
+-- Recursively finds and tries to order possible upgrades from highest tech to lowest tech. This allows queuing upgrades
+-- successively using keybinds, even if the previous upgrade isn't finished yet.
+-- Doing this fully recursively is probably overkill as the support fac / HQ is the only time when there are more than
+-- two possible upgrade paths, and the cybran shield is the only building with more than two successive upgrades, but
+-- hacking it together didn't feel right, hence the 'better' recursive solution.
+function issueUpgradeCommand(upgrades, buildableCategories)
```
4z0t, line 580: PascalCase

lua/keymap/hotbuild.lua
```diff
@@ -571,32 +571,50 @@ function buildActionUnit(name, modifier)
...
+    local success = false
+    for _, upgr in upgrades do
```
4z0t, line 582: Dont short names

4z0t (changes requested):

lua/keymap/hotbuild.lua
ChessBerry: So,
```
for _, u in upgrades do
```
would be better?

lua/keymap/hotbuild.lua
ChessBerry: yes, there is no reason for successiveUpgrades to be global

lua/keymap/hotbuild.lua
ChessBerry: PascalCase for the issueUpgradeCommand function? All other functions in hotbuild.lua are camelCase though. Is the new, proper style guide to make all new functions PascalCase?

Garanas: Yes - we're trying to get a standard going. New content should use pascal case.

lua/keymap/hotbuild.lua
Hdt80bro: It's a matter of style--the specific tradition 4z0t is referring to is one that uses one-letter variable names for loop iterators.  That's fine when you don't have lots of nested loops like you have here, but when there are--or when the variable does lots of heavy lifting--I prefer to see it spelled out completely (unless that creates a variable that has only a one-letter *difference* from another variable [e.g. `upgrade` and `upgrades`]). Abbreviations in general are fine. We don't actually a unifying standard on variable names, so pretty much whatever you do (within reason) is fine.

ChessBerry: I changed 'upgr' to 'u' as mentioned above

Garanas (approved):

ChessBerry:
Anything still missing that's preventing a merge? 

If I understand correctly, I implemented the last requested change in 7a4c18c above

4z0t (approved):

Garanas:
Yeah, I was discussing this in Zulip. Github doesn't understand when the requested changes are applied, hence we need 4z0t to do another review.

PR #4221, opened 2022-09-16
Add experimental features to the Atlantis

Garanas:
Adjusts two properties of the Atlantis:
- (1) Collision box that fits the unit mesh
- (2) Allow anti air weapons to fire while underwater

## Collision Box 

image
_Left is the adjusted collision box, right is the current collision box_ 

Reduces the collision box of the Atlantis. Vertically it extended a lot, way beyond the mesh of the Atlantis. As a consequence the Atlantis is one of the easiest submarines to hit with ground fire. This is now more difficult, but still possible.

## Anti air weapons

https://user-images.githubusercontent.com/15778155/190701246-4ef041e7-a69d-4b1a-b856-64af80a06e4c.mp4

Allows the anti air weapons of the Atlantis to fire while they are underwater. This would be unique - as an experimental unit should be. This significantly increases its anti air possibilities, as it can remain safely underwater. Ideally, I'd also like it to retarget the sam flayers when the target is destroyed. This is an experimental unit after all. A lot of salvo's are lost because the 2nd salvo is launched right before the first salvo destroyers the original target.

Visually there's still some work to do - I'll do that once it is accepted.


Tagada14:
I like the changes, will look further into it once the balance patch is finished and deployed.

Tagada14:
The Balance team discussed this feature and (2) will not be implemented.
While it's a cool idea it would go against the general balance of subs and AA (look at T3 sera subs). Atlantis is extremely tanky, especially for a submarine. Torpedo bombers are the primary weapon to deal with subs and Atlantis would be a hard counter to them. It would also heavily overlap in its role with the T2 Cruiser + Bulwark which is not desirable. We believe that it will be more interesting and healthier for the gameplay to pursue the role of a long rage sub for the Atlantis and changes in that direction are discussed, possibly including the (1).

PR #4222, opened 2022-09-16
Add experimental features to the CZAR

Garanas:
Adjusts one property of the CZAR:
- (1) Allows its primary weapon (the ever devastating beam) to target submerged units

## Quantum Beam Generator

https://user-images.githubusercontent.com/15778155/190704724-49f4a601-c86f-4c7b-ac5f-7fbd16c96162.mp4

Allows the quantum beam generator to target submarines as usual. The splash damage of the beam is so significant that it can damage then anyway, but currently it requires you to manually ground fire. This is an experimental unit - it should be able to decipher how it can target submarines.

Visually it doesn't look that great yet, we'll fix that once it is merged


Penguin5:
Does this allow it to do any more damage to submerged units than it presently can?

Garanas:
No, it just allows it to enable its beam and aim for them as it would with any other unit.

speed2CZ:
Dont all the other units need to ground fire as well?
This also makes the depth charges kinda obsolete.

Tagada14:
The Balance team has discussed this feature and decided against it.
It would remove a micro aspect which is not the direction we want to go. Supcom already puts very little emphasis on micro and ground firing is a decent micro depth that we don't want to remove.

PR #4225, opened 2022-09-16
Make Harbinger behave like SACU's

Hdt80bro:
As https://github.com/FAForever/fa/pull/4218 points out, there's no good way to solve Harbingers stopping to reclaim with Attack-Move without an engine patch. This PR makes it happen by applying a new category, `SACU_BEHAVIOR`, to the Harbinger. This works due to https://github.com/FAForever/FA-Binary-Patches/pull/13 renaming the `SUBCOMMANDER` category to `SACU_BEHAVIOR` in the engine (and we prevent compatibility issues by automatically adding `SACU_BEHAVIOR` to any unit with `SUBCOMMANDER`). There are no other side-effects because Harbingers cannot rebuild structures.

### DIFF
--- a/lua/system/blueprints-units.lua
+++ b/lua/system/blueprints-units.lua
@@ -1,6 +1,10 @@
 --- Post process a unit
 ---@param unit Unit
 local function PostProcessUnit(unit)
+    if table.find(unit.Categories, "SUBCOMMANDER") then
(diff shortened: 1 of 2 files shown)


Garanas:
Communicated it to the balance team, great work on an alternative to removing it!

Garanas (approved):

PR #4226, opened 2022-09-17
Decouple UI intialization and laying out

Hdt80bro:
This PR introduces Layouter support for the `Layout()` pattern by having the Layouter call it on `End()`. This moves laying-out out of the initializer so that controls can be completely created before they're laid out--something very handy when `End()` requires the entire component hierarchy to be complete to function properly. This also eliminates the need for a component to layout itself to layout its children, and thus increases modularity. In addition, controls can now set their layout to depend on children attributes without needing to interlace their layouts in a confusing manner.

See the [documentation](https://github.com/Hdt80bro/fa/blob/layout-pattern/lua/maui/layouthelpers.lua#L1090-L1175) for more details.

### DIFF
--- a/engine/User/UserUnit.lua
+++ b/engine/User/UserUnit.lua
@@ -167,7 +167,7 @@ function UserUnit:IsIdle()
 end
 
 ---
----@param category moho.EntityCategory
+---@param category CategoryName
 ---@return boolean
 function UserUnit:IsInCategory(category)
 end
--- a/lua/lazyvar.lua
+++ b/lua/lazyvar.lua
@@ -2,6 +2,8 @@
 -- LazyVar module
 --
 
+---@alias Lazy<T> T | LazyVar<T> | fun(): T
+
 local TableInsert = table.insert
 
 local iscallable = iscallable
@@ -16,11 +18,9 @@ ExtendedErrorMessages = false
 local EvalContext = nil
 local WeakKeyMeta = { __mode = 'k' }
 
----@class LazyVar : Destroyable, OnDirtyListener
+---@class LazyVar<T> : {[1]?: T, compute?: fun(): T}, Destroyable, OnDirtyListener
 ---@operator call: fun(): any
----@field [1]? any
 ---@field busy? boolean
----@field compute function
 ---@field trace? string
 ---@field used_by table<LazyVar, boolean>
 ---@field uses table<LazyVar, boolean>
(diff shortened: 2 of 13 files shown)


Garanas:
@4z0t could you look into this? Then we can merge it later this week, for us all to experiment with.

4z0t:
At first glance I can tell that it is going to break lots of things of mine and some ui code in game since you totally removed previous code. Im gonna take a look deeply later.

4z0t:
I would suggest to create new file for it / add to existing

Hdt80bro:
Oh, I forgot that you originally added this for your UI mods. Yeah, I'll definitely revert those changes.

> I would suggest to create new file for it / add to existing

I really want to move the Layouter to a separate file, but that'd break it even more. Unless you're talking about making a separate Layouter file in addition to what we already have? I'd like to keep the repository unified, so I'd rather not.

Hdt80bro:
@4z0t There, now the only two things to discuss about what I added are:
  * If using the reused layouter without calling `:End()` should warn you when you try to reuse the layouter again to catch accidental interlacing layout errors (though it still works by returning a new Layouter)
  * If we want to keep the layouter attributes, in case we really do want to add typed layouters in a separate file like you were suggesting

4z0t:
When **End()** was added it was meant to be optional and even unnecessary thing and thats why I dont use it. Now you are making it mandatory, which causes all mods I made produce Warn logs.

4z0t:
Anyway It is ok for with this new pattern, however I'm afraid of **GetControl** and its behavior.

Hdt80bro:
> Now you are making it mandatory, which causes all mods I made produce Warn logs.

Yea, your mods will still work just fine, they'll just produce warnings now. If you don't want that we can remove that layer of validation (though I'd rather not) and make `End()` work purely to validate layout positions and call `Layout`.

> however I'm afraid of GetControl and its behavior

Can you elaborate? The idea is to make a Layouter behave like its control much more closely and extend support for other layout objects to have containing controls. I nearly added the `GetControl` bypass to the base layout helpers, but decided that it was redundant for the current Layouter utilization.

There was one other thing I wanted to added, but figured would have been too much right now--that was making the parent control in the layout helper functions default to the control's actual parent to reduce verbosity.

4z0t:
> Can you elaborate?
If control has a field **c** its gonna break.

Hdt80bro:
Ah yes, I was going to rename the field to something more agreeable but it didn't make it to this revision. How does `layoutControl` sound?

4z0t:
Mb not a field but an index like lazyvar does: `self[1]`?

Hdt80bro:
Oh that'd be an even easier way to break it then, especially if they try to pass in an array--interestingly, it would make LazyVars pointing to a control work for laying out though. Do we think it'd be better to check for the metatable or a control getter method? That begins to sound like a hack or a similar problem. I think with a field like `layoutControl`, if the object that's passed in has the same field defined as a Layouter then they probably want it to work on that control anyway. Duck typing for the win!

4z0t:
Zero index then? 😅

PR #4228, opened 2022-09-17
Add key action localization directives

Hdt80bro:
Allows for key actions to be localization directives like `"<LOC loadingtip>Press {k toggle_key_bindings} to open the key mapper"`

### DIFF
--- a/lua/keymap/keymapper.lua
+++ b/lua/keymap/keymapper.lua
@@ -92,6 +92,27 @@ function GetCurrentKeyMap()
     return GetUserKeyMap() or GetDefaultKeyMap()
 end
 
+--- Returns the current key binding for an action, or `false` if not found
+---@param action string
+---@return string | false
+function GetCurrentKeyBinding(action)
+    local TableFind = table.find -- luckily, this handles nil tables
+
+    local binding = TableFind(Prefs.GetFromCurrentProfile("UserKeyMap"), action)
+    if binding then return binding end
+    binding = TableFind(Prefs.GetFromCurrentProfile("UserDebugKeyMap"), action)
+    if binding then return binding end
+
+    local defaultKeyMap = import(GetDefaultKeyMapName())
+
+    binding = TableFind(defaultKeyMap.defaultKeyMap, action)
+    if binding then return binding end
+    binding = TableFind(defaultKeyMap.debugKeyMap, action)
+    if binding then return binding end
+
+    return false
+end
+
 function ClearUserKeyMapping(key)
     if not key then return end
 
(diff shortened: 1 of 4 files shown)


Garanas:
That is interesting, is this to make the tooltips more consistent? Does it work when you have multiple keys, as an example: `shift + k`?

Hdt80bro:
Yes, there's great potential in tooltips, though honestly I was inspired for the loading tips.

Modifiers are actually already part of the key binding name, for example action `"fac_group1"`'s default binding is `"Ctrl-Shift-1"`--the localization code in the keymapper is mostly there to normalize modifier order (though it also calls `LOC` to get the proper names of the keys).

Garanas:
How do I test these changes? 😄

Hdt80bro:
Frankly, I just replaced `/lua/ui/help/tooltips.lua` with something like
```lua
Tips = {"Press {k toggle_key_bindings} to open the key mapper"}
```
Though there's probably a more accessible `LOC()` you can use

ChessBerry:
@Garanas @Hdt80bro 
For some unknown reason, this PR, specifically 562208e38c610c7fe8694c3176b0ac4ff6a4d2d9, breaks the ecomanager mod completely as it turns all non-HUD graphics completely black.

With 562208e38c610c7fe8694c3176b0ac4ff6a4d2d9:
Forged Alliance 24_09_2022 19_12_49

With a09fa3454f4173644e88c99bfc64d213c205ba32, the commit just before 562208e38c610c7fe8694c3176b0ac4ff6a4d2d9:
Forged Alliance 24_09_2022 19_16_25

Of course, development on the faf repo shouldn't be constrained by some random mod.
There are a decent number of people using that mod though, even if just just for the mex overlay feature. so before this is pushed onto the main FAF branch, a PSA to turn it off is probably necessary.

I think the problem is with ecomanager's scoreboard changes that have been kinda broken for years. No promises, but if I can figure out what's going wrong, I'll push a fixed version of ecomanager to the vault.

Ps: If this isn't the right place to post notes likes this, please direct me to where it would be more appropriate

ChessBerry:
Figured it out:

In on of ecomanager's modules it calls the UIUtil.CreateText as:

```
UIUtil.CreateText(AutoshareContainer, 0, 11, UIUtil.bodyFont)
```

this seemingly doesn't work anymore, while changing the second argument to a string
```
UIUtil.CreateText(AutoshareContainer, '0', 11, UIUtil.bodyFont)
```
returns the functionality to where it was before.

If I understand correctly, this is a result of you changing the LOC function in Localization.lua, which is called from within to CreateText:
```
--* return a text object with the appropriate font set
function CreateText(parent, label, pointSize, font, dropshadow)
    label = LOC(label) or LOC("<LOC uiutil_0000>[no text]")
    font = font or buttonFont
    local text = Text(parent, "Text: " .. label)
    text:SetFont(font, pointSize)
    text:SetColor(fontColor)
    text:SetText(label)

    if dropshadow then
        text:SetDropShadow(true)
    end
    return text
end
```

to expect a string, and only a string, as argument, removing the implicit casting, from e.g. int to string, that happened with the old LOC in the StringSub function.

Was this an intentional change?

Hdt80bro:
Yes, this was an intentional change intended to promote type safety. You can see that I was able to change all of the instances of numbers being passed into `LOC` in our repo for the PR due to it.

However, if this *does* completely break mods, then we obviously can't have that. I'll revert to the previous behavior.

PR #4232, opened 2022-09-18
Optimize Reclaim Labels attempt 69

4z0t:
Optimizes Labels display to reduce FPS drops. 
* Selects only reclaim props in view for project
* Combines nearest props depending on camera zoom

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -89,7 +89,6 @@ deprecatedMods["DynamicReclaimGrouping"] = true
 deprecatedMods["EzReclaim"] = true
 deprecatedMods["OnScreenReclaimCounter"] = true
 deprecatedMods["ORV"] = true
-deprecatedMods["ReclaimFix"] = true
 deprecatedMods["SmartReclaimSupport"] = true
 
 -- typical FA packages
--- a/init.lua
+++ b/init.lua
@@ -60,22 +60,52 @@ end
 -- mods that have been integrated, based on folder name 
 local integratedMods = { }
 integratedMods["nvidia fix"] = true
+
 integratedMods = LowerHashTable(integratedMods)
+
+-- mods that are deprecated, based on folder name
+local deprecatedMods = { }
+deprecatedMods["simspeed++"] = true
+deprecatedMods["#quality of performance 2022"] = true
+
+-- as per #4119 the control groups (called selection sets in code) are completely overhauled and extended feature-wise,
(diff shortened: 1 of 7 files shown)


Garanas:
Managed to go from 90 fps (before 6fdd6f5) to 200 fps (after 6fdd6f5) when rendering reclaim on the Ditch. There was a lot of overhead:
- Each reclaim label had their own instance of an update and display function
- Each reclaim point had a field (`inPlayableArea`) that can be reconstructed by knowing the label is not `OutsidePlayableAreaReclaim`
- Each reclaim label had a field (`proj`) that allocated a new table (of two elements) each frame
- Each reclaim label created two closures each frame (by using the LayoutHelpers function, instead of directly updating the value)

And there's still a lot of overhead:
- Pooling is nice, but only if you do not re-allocate the entire thing
- The function `project` is called in `OnFrame` for each label (the big culprit!!), while that is technically only required when we move the camera
- Combining of reclaim labels makes the entire procedure `O(n^2)`, while we can have it `O(nk)` where `k` is strictly smaller than `n` by using a grid structure. Do not overcomplicate this with a quad tree, it is no longer required.

And in general: should we show reclaim labels when we move the camera? It causes a few issues as the camera position (and the updating of the labels) is not performed each frame, but each tick. Especially when in split view this is noteable.

Another issue is that the old code was not really moddable. As a result, every UI mod that interacts with the reclaim labels usually just copied the source file. For people to take note of these changes we'll need to deprecate those mods using the init file. As a result, we should also:
- (1) make this implementation more moddable (more use of functions that have a single, simple purpose)
- (2) add some of features of the UI mods that we're deprecating (like the coloring, total reclaim value on screen)

Garanas (changes requested):
Oh, and please annotate all future changes 😄

Garanas:
reclaim-faf
_on deploy/faf_

reclaim-before-6fdd6f5
_commit 9beeb6d_

reclaim-after-6fdd6f5
_commit 6fdd6f5_

Great work @4z0t, let us make it even better the upcoming days and merge it in over the weekend!

4z0t:
Reclaim fix and EzReclaim shouldn't be there, their functions are displaying lables.

Garanas:
`EzReclaim` hard-hooks one of the files that we're working on. It shouldn't be doing that. I'll remove `Reclaim fix` from the deprecation list

Garanas (approved):

PR #4234, opened 2022-09-19
Fix "reveal civilians" not resetting civilians to neutral

Hdt80bro:
Got broken in the recall patch; this fixes it

### DIFF
--- a/lua/sim/ScenarioUtilities.lua
+++ b/lua/sim/ScenarioUtilities.lua
@@ -605,8 +605,6 @@ function InitializeArmies()
             local setup = armySetups[strArmy]
             local brain = GetArmyBrain(strArmy)
 
-            --LOG("*DEBUG: InitializeArmies, army = ", strArmy)
-
             local econ = tblData.Economy
             SetArmyEconomy(strArmy, econ.mass, econ.energy)
 
@@ -721,9 +719,6 @@ function InitializeScenarioArmies()
         if tblData then
             local setup = armySetups[strArmy]
             local brain = GetArmyBrain(strArmy)
-
-            LOG("*DEBUG: InitializeScenarioArmies, army = ", strArmy)
-
 
             local econ = tblData.Economy
             SetArmyEconomy(strArmy, econ.mass, econ.energy)
--- a/lua/sim/ScenarioUtilities.lua
+++ b/lua/sim/ScenarioUtilities.lua
@@ -641,19 +641,18 @@ function InitializeArmies()
 
             ----[ irumsey                                                         ]--
             ----[ Temporary defaults.  Make sure some fighting will break out.    ]--
-            for iEnemy, strEnemy in tblArmy do
+            for iEnemy, _ in tblArmy do
                 -- only do it once for each pair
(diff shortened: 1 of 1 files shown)


Hdt80bro:
Why don't AI's get to see civilians?

Garanas:
I don't know, @relent0r would you want AIs to be able to see civs like humans can?

relent0r:
I actually got sick of AI being unable to see civilians a while back and did a little hack that would reveal civilians (set them as allies) for half a second then revert.

So yes I'd love for AI to be able to see Civilians by default as currently without the above they just suicide a bunch of early units into PDs.

Hdt80bro:
Oh, well it'd be trivially easy to get them to see civilians--we actually explicitly exclude them from the reveal.

relent0r:
Ha that sounds like discrimination :) and an odd decision to make. I'm not sure if there would be any consequences for coop games though, that's the only reason I would imagine they would be excluded.

Garanas:
> Ha that sounds like discrimination :) and an odd decision to make. I'm not sure if there would be any consequences for coop games though, that's the only reason I would imagine they would be excluded.

That is an interesting question, we'll find out I assume.

lua/sim/ScenarioUtilities.lua
```diff
@@ -618,11 +617,11 @@ function InitializeArmies()
 
             local armyIsCiv = setup.Civilian
 
-            if armyIsCiv and civOpt ~= 'neutral' and strArmy ~= 'NEUTRAL_CIVILIAN' then -- give enemy civilians darker color
+            if armyIsCiv and civOpt ~= "neutral" and strArmy ~= "NEUTRAL_CIVILIAN" then -- give enemy civilians darker color
```
Garanas, line 620: Is the lua plugin telling you to make these changes?

lua/sim/ScenarioUtilities.lua
```diff
@@ -728,7 +722,7 @@ function InitializeScenarioArmies()
             local setup = armySetups[strArmy]
             local brain = GetArmyBrain(strArmy)
 
-            LOG('*DEBUG: InitializeScenarioArmies, army = ', strArmy)
+            LOG("*DEBUG: InitializeScenarioArmies, army = ", strArmy)
```
Garanas, line 725: This LOG statement can go

lua/sim/ScenarioUtilities.lua
```diff
@@ -606,7 +605,7 @@ function InitializeArmies()
             local setup = armySetups[strArmy]
             local brain = GetArmyBrain(strArmy)
 
-            --LOG('*DEBUG: InitializeArmies, army = ', strArmy)
+            --LOG("*DEBUG: InitializeArmies, army = ", strArmy)
```
Garanas, line 608: This comment can go too

Garanas (approved):

lua/sim/ScenarioUtilities.lua
Hdt80bro: No, that's just my personal preference =)

PR #4236, opened 2022-09-20
Intelisence minor fixes

MrRowey:
Some minor changes

### DIFF
--- a/lua/options/options.lua
+++ b/lua/options/options.lua
@@ -73,12 +73,12 @@ optionsOrder = {
 
 local Prefs = import('/lua/user/prefs.lua')
 local SetMusicVolume = import('/lua/UserMusic.lua').SetMusicVolume
-local savedMasterVol = false
-local savedFXVol = false
-local savedMusicVol = false
-local savedVOVol = false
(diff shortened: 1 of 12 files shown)


Hdt80bro (approved):

lua/options/options.lua
```diff
@@ -73,20 +73,20 @@ optionsOrder = {
 
 local Prefs = import('/lua/user/prefs.lua')
 local SetMusicVolume = import('/lua/UserMusic.lua').SetMusicVolume
-local savedMasterVol = false
-local savedFXVol = false
-local savedMusicVol = false
-local savedVOVol = false
+local savedMasterVol = nil
```
Garanas, line 76: `nil` is not always the same as `false`, did you confirm there are no if statements that specifically rely on this being `false`?

MrRowey: Ive only changed ones that needed a Nil check

MrRowey: Most the ones I've changes were where it was being detected as Bool when it was something else

Garanas (approved):

PR #4237, opened 2022-09-21
Change `self.Fun(self, ...)` to `self:Fun(...)`

Hdt80bro:
This is more efficient bytecode-wise in addition to being more maintainable:

```self.Fun(self)```

compiles to
```
0 - GETTABLE       1    0  250
1 - MOVE           2    0    0
2 - CALL           1    2    1
```

while

```self:Fun()```

compiles to
```
0 - SELF           1    0  250
1 - CALL           1    2    1
```

The speed savings are marginal.

For those looking to win regex golf, this should be fit by the pattern `\b(?!table\b)(?!string\b)(\w+)\s*\.\s*(\w+)\s*\(\s*\1\s*(,\s*|(\)))` ->`$1:$2($4` (though make sure to leave the benchmarking functions the same)

### DIFF
--- a/engine/Sim.lua
+++ b/engine/Sim.lua
@@ -174,7 +174,7 @@ end
 function CreateBuilderArmController(unit, turretBone, barrelBone, aimBone)
 end
 
---- Creates a collision detection manipulator, calls the function `self.OnAnimTerrainCollision(self, bone, x, y, z)`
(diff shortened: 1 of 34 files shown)


Garanas (approved):
We should probably document this type of performance choices on the wiki.

PR #4240, opened 2022-09-21
added a missing 0 for an offset

MrRowey:


Hdt80bro (changes requested):
Unfortunately, this is valid syntax. I add the 0 whenever I come across it so it reads better, but I currently count 534 instances of this already in the repo. Probably not worth an entire commit, but you can experiment with a global replacement using the regex `([^-]-|[^\w\.%-])\.(\d)` -> `$10.$2`. Otherwise, this might get closed unfortunately.

MrRowey:
huh Interning would not that thought that was valid feel free to close this then it just looked odd with out the zero. Regex always looks like random magic need to read up on how all those random thing together = something xD

Hdt80bro:
Yeah, VSC doesn't do a good job syntax highlighting decimals without a leading 0

PR #4247, opened 2022-09-30
Physically Based Rendering

Blodir:
Implements physically based shading for meshes.

We've had to work within the limitations of the engine:
* There's only one direct light: the sun. Settings for the sun are changed on a per map basis by the map creator.
* We are relying on existing environment maps, which are of varying quality. Notably some are missing the lower half (it's all black) entirely, which causes big issues with the shading.
* We are using mip levels of the environment map to approximate diffuse irradiance maps as we can't precompute anything more accurate without engine changes
* We have replaced the (unused) anisotropic texture with a precomputed environment BRDF lookup texture

Resources:
* https://learnopengl.com/PBR/Theory
* https://learnopengl.com/PBR/IBL/Diffuse-irradiance
* https://learnopengl.com/PBR/IBL/Specular-IBL
* https://blog.selfshadow.com/publications/s2013-shading-course/

image
image
image
Untitled
Note that the lighting setting of the maps varies wildly and we have not set on a good mapping for it, which is why in the last example the PBR shading looks darker than the original.

Also note that while we are trying to infer albedo/metallic/roughness/ao information from the original textures shipped with the game, those textures are very inconsistent at times. For optimal results each texture has to be manually checked.

(Jip) Todo before merge:
- [ ] fix Nomads overriding the mesh shader


Garanas:
Exciting initial start, wonder what it looks like on Seraphim units 😄 .

In terms of pre-computing the roughness, BlackYps mentioned we can use mip maps for that. You can generate initial mip maps using Image Magick, as separate files. And then combine them again. I experimented with that for maps to reduce the detail of a texture as you zoom out so that it looks less repetitive 😄 

cubemap

An example cube map, extracted from fa steam. I'm not entirely sure what order the images are but it is difficult to imagine they choose a different order than the usual cube maps. Maybe @The-Balthazar has an idea about it

Blodir:
image
image

Blodir:
image

Garanas:
@Blodir  / @BlackYps Lets finalize the shader part of this task. Aim is to merge this pull request. We'll complete the work (actually assign units the shaders) in another pull request, in the near future when we start working on the textures of individual units.

PR #4249, opened 2022-10-01
Add in various suggestions of people

Garanas:
Adds a hotkey to select all mass extractors that are upgrading.

Adds a hotkey to select all radars.

Adds a hotkey to filter your selection (of engineers) to one unit that represents the highest tech. All other engineers assist the selected unit. Command units are ignored, as it isn't trivial to take into account enhancements.

https://user-images.githubusercontent.com/15778155/193425009-2badf18b-4dc2-4077-977f-3d0f7a30a568.mp4


lua/keymap/misckeyactions.lua
```diff
@@ -392,4 +393,77 @@ end
...
+    if selection then
+
+        -- try and find extractors that are upgrading
+        local upgrading = { }
+        for k, unit in selection do
+            if unit:GetWorkProgress() > 0 then
+                table.insert(upgrading, unit)
+            end
+
+            if next(upgrading) then
```
4z0t, line 416: Why is this here?

Garanas: to check if the array has at least one entry in it

4z0t: Use table.empty instead. And why would you check it each iteration?

Hdt80bro: `next` is more efficient, although I understand the concern about loss of conveyed intention. Although, if we cared about optimizing code, there are several other thing we could do here as well. As for why it is checked each iteration, it can change each iteration.

4z0t: Selecting units is very expensive operation, I'd recommend do it in the end...

Garanas: it is supposed to happen in the end.

Garanas:
We'll tackle the 'get into transports' hotkey with another pull request

PR #4253, opened 2022-10-09
Attempt at fixing engineer behavior for Seraphim air factories

Garanas:
In response to [this forum post](https://forum.faforever.com/topic/4965/sera-airfab-engineers-are-not-selectable-in-the-unloading-process). I always found it annoying too 😄 

Appears there is no easy solution with the existing approach (where the engineer is 'tethered' off), but why not just 'blink' it off? Like a rabbit from a magicians hat:

https://user-images.githubusercontent.com/15778155/194754212-5e56257a-3f06-4dbf-bdb5-1baf4731bbb0.mp4

I'm not entirely sure about the effects being used, open to suggestions for alternative effects. But this would be the only viable solution that I can see so far.

### DIFF
--- a/lua/seraphimunits.lua
+++ b/lua/seraphimunits.lua
@@ -128,17 +128,12 @@ SAirFactoryUnit = Class(AirFactoryUnit) {
                 unitBuilding:DetachFrom(true)
                 self:DetachAll(self.Blueprint.Display.BuildAttachBone or 0)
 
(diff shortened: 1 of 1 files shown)


magge-faf:
Thanks for your time and effort - I like the idea and looks more natural to me. The "air tethering-drag-and-drop_ish_-way" was always a bit odd to look at. (counts for all air fabs). If there is no unusual side effect doing the "rabbit way", I am all in to test it further.

PR #4262, opened 2022-10-10
Fix a few teleportation glitches

Garanas:
Fixes a few teleportation glitches by checking if the unit can still teleport at the start and end of the teleportation sequence.

```
ACU starts with teleportation enhancement

ACU queues transport order, then teleportation order
Transport drops anywhere
ACU initiates the teleport but it able to do whatever it feels like
```

```
ACU starts with teleportation enhancement

ACU queues up a move order, queues up switching the enhancement, queues up a move order and then a teleport order
ACU moves, switches the enhancement, moves and then teleports even if it lost the enhancement
```

```
ACU starts with teleportation enhancement

ACU queues up the teleportation too close to the map border, ACU ends up outside the map being un targetable.
```


lua/defaultunits.lua
```diff
@@ -2548,72 +2548,6 @@ CommandUnit = Class(WalkingLandUnit) {
     -- TELEPORTING WITH DELAY
     -------------------------------------------------------------------------------------------
 
-    ---@param self CommandUnit
-    ---@param teleporter any
-    ---@param location Vector
-    ---@param orientation Quaternion
-    InitiateTeleportThread = function(self, teleporter, location, orientation)
```
Garanas, line 2555: There was no difference between this version, and the one in the unit class

Hdt80bro:
Most likely - it reads statically from the blueprint, so it won't be as dynamic as that solution.

Issue #4258, opened 2022-10-09
Add division to score board for matchmaker games


BlackYps:
The latest client sends the division name of the player to the game if applicable. PR: https://github.com/FAForever/downlords-faf-client/pull/2793

Now the game needs to pick that name up to display it in the score board of the game instead of the rating. We could also or instead display a 20x40px image that petric made that visually represents the division.

The division info should also be saved in the replay. 
Can we even add more data to the replay?

Garanas:
We can add data to the replay by doing a `SimCallback`. Where are the images, I think we should use those

BlackYps:
Would you rather have a link to the server location or the physical files?

BlackYps:
Icons on server are here: https://content.faforever.com/divisions/icons/small/goldI_small.png I can also provide the physical files if that is better

Garanas:
Batching them into a zip would be preferred, I can't properly navigate that link

BlackYps:
[icons.zip](https://github.com/FAForever/fa/files/9758496/icons.zip)
@4z0t was also interested in implementing this

4z0t:
Closing?

BlackYps:
No?

PR #4263, opened 2022-10-12
Add Divisions to scoreboard

4z0t:
Related to this #4258, however can't test it since everybody must have this code on their side in order to see division of every player ~~(by default all gm's lmao) and position must be corrected for sure.~~ (Works in autolobby only) Decided to not change base scoreboard layout leaving it for modders.


4z0t:
@Garanas I can leave only the fact of divisions, because this PR waits for too long

4z0t:
@Garanas what are we doing with this one?

4z0t:
It also requires other division textures.

Hdt80bro:
Is this supposed to add divisions to the scoreboard or lobby? I see code that sets up both, but not actually doing either.

4z0t:
@Hdt80bro it adds only the fact that player has division. Those are provided through gameoptions.

BlackYps:
It's supposed to be for the scoreboard. Maybe we can show them in the lobby as well at a later point, but for now the focus is the scoreboard

4z0t:
I don't see a better place for divisions in base scoreboard.

BlackYps:
It's supposed to replace rating

4z0t:
In base scoreboard rating is concatenated with nickname. What i mean is current code base isn't suitable for any changes like this. I will add in my scoreboard for sure 🙂

Garanas (approved):

PR #4265, opened 2022-10-13
Implement first round of ai fixes and improvements

relent0r:
## Description of changes

This PR implements the first of a series of AI fixes and improvements.
The PR focuses on the following.

Small adjustment to the base template for the 'Normal AI' so that it will build a couple more engineers and one more land factory.

**Improve the AttackForceAI by the following**
Add a maximum platoon size count to stop attack force platoons growing beyond a comfortable size. A PlatoonFull property will be set when the platoon goes beyond its size limit. It will default to 40 if unset.
Add size restrictions to the MergeWithNearbyPlatoons platoon function by way of a maxPlatoonCount param that will check if the platoon is already at maximum size.

**Improve the EconomyMonitor**
Implements a new Economy Over Time thread that will provide a 30 second history (compared to the previous 8) of economy data and will allow more types of economy over time metrics. I've kept the sample length small to minimise impact to the AI behaviour. 
Updated the GetEconomyNumbers function to leverage this data when it is called for backwards compatibility.
Update existing EconomyOverTime build conditions to leverage this data and cut down on excess GetEconomyNumbers calls.

**Improve ACU behavior**
Add a combat property that will be set when the ACU goes into combat and removed after it returns to base. This property will enforce no engineer jobs are assigned to the ACU when it is in combat. This resolves the issues where the ACU would try to build structures while in combat.
Add a structure completion function to the default acu threads to allow the acu to complete structures correctly after it has come out of combat. This will avoid the problem of the acu starting and not completing excess numbers of structures when it is constantly being called into and out of combat due to units popping into range.

## Test setup for the changes
Testing these changes requires logging and observation. 
I have logged the changes to confirm logic is triggering when required and economy data is being passed correctly. 
Platoons that use the updated code will move along paths without re-orientating as much as the direction is pre calculated, this will result in quick and smoother platoon movement. The platoons should not go beyond a maximum of 40 units so that movement issues are limited. Though it must be noted that 40 units will still cause issues on certain maps. The max platoon counts could be limited in the future through the builders but will require tuning so that game play experience is not negatively impacted.
The ACU changes should be easily observed when comparing to the previous behaviours. The acu should no longer try to start building structures while in combat and should complete any structures that it will building when it went into combat but were not finished.

Note : The EconomyOverTime changes will impact the Sorian AI as it leverages the same economic data but based on my testing using a sample length of 30 seconds does not cause any negative impact. No other changes impact the Sorian AI.

### DIFF
--- a/lua/AI/aiattackutilities.lua
+++ b/lua/AI/aiattackutilities.lua
@@ -815,7 +815,6 @@ function AIPlatoonSquadAttackVector(aiBrain, platoon, bAggro)
                 -- force reevaluation
                 platoon.LastAttackDestination = {attackPos}
             else
-                local pathSize = table.getn(path)
                 -- store path
                 platoon.LastAttackDestination = path
                 -- move to new location
--- a/lua/AI/aiattackutilities.lua
+++ b/lua/AI/aiattackutilities.lua
@@ -819,12 +819,10 @@ function AIPlatoonSquadAttackVector(aiBrain, platoon, bAggro)
                 -- store path
                 platoon.LastAttackDestination = path
                 -- move to new location
-                for wpidx,waypointPath in path do
-                    if wpidx == pathSize or bAggro then
-                        platoon:AggressiveMoveToLocation(waypointPath)
-                    else
-                        platoon:MoveToLocation(waypointPath, false)
-                    end
+                if bAggro then
+                    platoon:IssueAggressiveMoveAlongRoute(path)
+                else
+                    platoon:IssueMoveAlongRoute(path)
                 end
             end
         end
--- a/lua/AI/aiutilities.lua
+++ b/lua/AI/aiutilities.lua
@@ -34,16 +34,12 @@ function AIGetEconomyNumbers(aiBrain)
     econ.MassStorage = aiBrain:GetEconomyStored('MASS')
 
     if aiBrain.EconomyMonitorThread then
-        local econTime = aiBrain:GetEconomyOverTime()
-
-        econ.EnergyRequestOverTime = econTime.EnergyRequested
-        econ.MassRequestOverTime = econTime.MassRequested
-
-        econ.EnergyIncomeOverTime = SUtils.Round(econTime.EnergyIncome, 2)
-        econ.MassIncomeOverTime = SUtils.Round(econTime.MassIncome, 2)
-
-        econ.EnergyEfficiencyOverTime = math.min(econTime.EnergyIncome / econTime.EnergyRequested, 2)
-        econ.MassEfficiencyOverTime = math.min(econTime.MassIncome / econTime.MassRequested, 2)
+        econ.EnergyRequestOverTime = aiBrain.EconomyOverTimeCurrent.EnergyRequested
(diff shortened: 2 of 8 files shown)


lua/AI/aiutilities.lua
```diff
@@ -34,16 +34,12 @@ function AIGetEconomyNumbers(aiBrain)
...
-
-        econ.EnergyIncomeOverTime = SUtils.Round(econTime.EnergyIncome, 2)
-        econ.MassIncomeOverTime = SUtils.Round(econTime.MassIncome, 2)
-
-        econ.EnergyEfficiencyOverTime = math.min(econTime.EnergyIncome / econTime.EnergyRequested, 2)
-        econ.MassEfficiencyOverTime = math.min(econTime.MassIncome / econTime.MassRequested, 2)
+        econ.EnergyRequestOverTime = aiBrain.EconomyOverTimeCurrent.EnergyRequested
+        econ.MassRequestOverTime = aiBrain.EconomyOverTimeCurrent.MassRequested
+        econ.EnergyIncomeOverTime = aiBrain.EconomyOverTimeCurrentEnergyIncome
+        econ.MassIncomeOverTime = aiBrain.EconomyOverTimeCurrentMassIncome
```
Uveso, line 40: same as EconomyOverTimeCurrentEnergyIncome
(missing point?)

Garanas:
Thank you for the quick review @Uveso

lua/AI/aiutilities.lua
relent0r: Thanks Uveso, I'll update.

lua/AI/aiattackutilities.lua
```diff
@@ -819,12 +819,10 @@ function AIPlatoonSquadAttackVector(aiBrain, platoon, bAggro)
                 -- store path
                 platoon.LastAttackDestination = path
                 -- move to new location
-                for wpidx,waypointPath in path do
-                    if wpidx == pathSize or bAggro then
-                        platoon:AggressiveMoveToLocation(waypointPath)
-                    else
-                        platoon:MoveToLocation(waypointPath, false)
-                    end
+                if bAggro then
```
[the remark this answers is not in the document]
relent0r: Ah yes thanks, your right about the aggressive move at the end of the path, I'd forgotten about that bit. I'll need to add another param to Jips movealongroute function to support that.

Now that you mention it, I just spotted a little type on mine 6837 of the platoon.lua that I need to fix up as well.
@Garanas I assume that line is supposed to read for ``k = 1, count do`` and not ``for k = fi1rst, count do``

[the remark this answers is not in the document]
relent0r: I have added the final aggressive move to the IssueAggressiveMoveAlongRoute function and removed the unneeded variable pathSize from the AIPlatoonSquadAttackVector function.

lua/AI/aiutilities.lua
relent0r: Have corrected these typos.

Uveso (approved):
Didn't find any other errors

Hdt80bro:
No, go ahead and merge

PR #4271, opened 2022-10-15
Add new economy conditions for AI, fix ai ACU from not completing structures after multiple interuptions

relent0r:
## Description of changes
This PR adds two new builder conditions.
**GreaterThanEconEfficiencyCombined**
It replaces the standard over time efficiency conditions for all AI structure, assist builders (does not change unit builders). The goal of this condition is for the AI to take into account the over time efficiency and current efficiency economic situation. This will help the AI make better economic decisions during economy fluxuations.

GreaterThanEconIncomeOverTime
It replaces the standard EconIncome condition for all AI builders. This will allow the AI to take income account over time income metrics so that small spikes from reclaim will not trigger builders expecting a consistent income requirement.

**Fix to ACU structure completion during multiple combat interupts**
It was found after the last code update that while the ACU was completing structures after a single combat call when it was interrupted multiple times the CDRFinishUnit function would potentially detect the ACU as idle and set the UnitBeingBuiltBehavior property to false when the structure was not completed. Also added a flag to stop the acu getting new engineer task while UnitBeingBuiltBehavior was not completed.

## Test setup for the changes
Logging can be used to validate the correct economy metrics are being collected.
Observation will only show in certain situations when the instant economic situation does not match the over time one.

The ACU fix can be observed when the ACU has partially built a structure but it call away for combat, returns and start completing the structure then is call away again. The ACU will no longer ignore the previous structure. If the structure is destroyed or completed by another engineer in the mean time the UnitBeingBuiltBehavior will be set to false on the next ACu thread loop.


relent0r:
@Uveso are you able to review this one please. I've got another for category string cleanup but don't want to cause a merge conflict with this one.

Uveso (approved):
Looks good, works good :)
Tested with GPG and Sorian as usual.

PR #4279, opened 2022-10-20
Add new translations

Yuchenjimmy:


Garanas:
I've re-translated a few back to english using Google Translate - looks good! If more is coming, could you make those in a separate pull request? Then we can merge this one in

Garanas (approved):

Yuchenjimmy:
I’m still working on it, but kind of busy these three days, next update may not be too soon. I also translate for the download client, both simplified Chinese and traditional Chinese had been updated, please apply them at next announcement, thank you!

Yuchenjimmy:
An new update, a new pull request? Understood.

Garanas:
Yes - a separate pull request for each would be perfect

PR #4284, opened 2022-10-22
Refactor all default AI builders removing all string categories from builder conditions.

relent0r:
## Description of changes
This PR refactors all default AI builder conditions removing all string categories and replacing with user data categories.

The default codebase has a very inconsistent category type, it has both string and user data category types for all builders. Any strings users to categories for builder conditions must be parsed into userdata before the condition is run.

For example
```
function FactoryComparisonAtLocation(aiBrain, locationType, unitCount, unitCategory, compareType)
    local factoryManager = aiBrain.BuilderManagers[locationType].FactoryManager
    local testCat = unitCategory
    if type(unitCategory) == 'string' then
        testCat = ParseEntityCategory(unitCategory)
    end
    if not factoryManager then
        WARN('*AI WARNING: FactoryComparisonAtLocation - Invalid location - ' .. locationType)
        return false
    end
    local numUnits = factoryManager:GetNumCategoryFactories(testCat)
    return CompareBody(numUnits, unitCount, compareType)
end
```
So any time a condition is passing string categories the game performing `ParseEntityCategory` function calls. This is done 100s (thousands?) of times for each loop of the builder condition monitor thread. This PR will save the engine 1000s of function calls and make the builder condition format consistent across the default AI builders.

There is also a fix for an error in the platoon.lua file where an instance of GetPlatoonUnits was incorrect when platoons were trying to merge.

Note : Nothing was done for the Sorian AI. 

Last Note : there will be a follow up PR for converting enginerbuildai in the future to category data but the number of times those are called by comparison are much fewer.

Test setup for the changes
The Profiler can be used to see the difference in the number of times the function call is made. I have provided comparative screenshots of the results based on before and after.

With the string categories there are 7763 ParseEntityCategory function calls within a 6 minute period.
After the string categories are converted to user data the number of function calls drops to 1432 for the same time period.

CategoryConversion
StringCategories


Garanas:
Good call!

I'm wondering: how are build conditions evaluated? Does it stop evaluating a builder when the first condition is invalid, or does it try all of them regardless? 

```
            { UCBC, 'HaveLessThanUnitsInCategoryBeingBuilt', { 1, categories.MASSEXTRACTION } },
            { EBC, 'GreaterThanEconIncomeOverTime', { 7, 50 } },
            { IBC, 'BrainNotLowPowerMode', {} },
            { EBC, 'GreaterThanEconEfficiencyCombined', { 0.9, 1.2 }},
            { UCBC, 'HaveGreaterThanUnitsWithCategory', { 5, categories.MASSEXTRACTION }},
```

As an example, it seems to be that `BrainNotLowPowerMode` is a lot cheaper to check then `HaveLessThanUnitsInCategoryBeingBuilt` or `HaveGreaterThanUnitsWithCategory`, and if the script stops evaluating when one condition fails it should first evaluate the cheaper ones.

(this is not for this pull request, but just in general - if this is the case then I'll make a pull request too)

Garanas (approved):

Garanas:
Appears to be this function, from `/lua/sim/builder.lua`:

```lua
    ---@param self Builder
    ---@return boolean
    BuilderConditionTest = function(self)
        for k,v in self.BuilderConditions do
            if not self.Brain.ConditionsMonitor:CheckKeyedCondition(v, self.ReportFailure) then
                self.BuilderStatus = false
                if self.ReportFailure then
                    LOG('*AI DEBUG: ' .. self.BuilderName .. ' - Failure Report Complete')
                end
                return false
            end
        end
        self.BuilderStatus = true
        return true
    end,
```

Garanas:
Or not, I'm not entirely sure where they get evaluated.

relent0r:
From memory you've got it pretty much right.

https://github.com/FAForever/fa/blob/f900096f05c62c30d412921a2de18edefe8bfd4c/lua/sim/BrainConditionsMonitor.lua#L192

This thing here loops across all the conditions to assess. (there is a small difference between engineers/factories and platoon formers tho)

The other thing s on line 135 of that file. It treats certain file names differently, certain condition files are treated instantly and must be evaluated every time they are checked. Others are cached for 7 seconds.

So having the cheapest/most likely to fail conditions first improves the condition checks, and having the conditions checking flags rather than performing all sorts of logic and function calls improves it too. In my AI I have a few threads that run every x number of seconds and just have the conditions check the flag status rather than have multiple instances of the logic running every x ticks depending on how long the condition monitor is taking to perform a full loop.

Then for every new base that gets spun up for the AI another set of managers is also spun up. Usually the drop is sim speed in AI games that is not directly related to unit activity is related to the number of bases and managers that are running at any one time.

PR #4287, opened 2022-10-23
Adjust factory behavior when building and finished building a unit

Garanas:
This pull request fixes various issues with factories:
- (1) A factory that assists another factory uses the rally point of the other factory to compute the correct roll off point
- (2) A factory can now spin the unit to the correct roll off point when the unit is finished building, see also:

https://user-images.githubusercontent.com/15778155/197396006-eac6737a-aa3b-4012-be22-0579abb667a4.mp4

- (3) A factory no longer waits for the unit that is being built to finish its first order. That order can get cancelled by a player (or an AI) and by doing so, artificially decreasing the 'roll off costs'. Instead, the function takes into account the distance of the unit to the build pad. If unit is sufficiently far away it will start building the next unit

https://user-images.githubusercontent.com/15778155/197396129-d25ee02c-f486-4746-8cba-9432647b494a.mp4

All in all, a reduction of the frustrations surrounding of units leaving the factories.

### DIFF
--- a/engine/Core.lua
+++ b/engine/Core.lua
@@ -201,7 +201,7 @@ function IsAlly(army1, army2)
 end
 
 --- Checks if the C-side of an object is destroyed / de-allocated
----@param entity InternalObject
+---@param entity? InternalObject
 ---@return boolean
 function IsDestroyed(entity)
 end
--- a/engine/Core/Blueprints/UnitBlueprint.lua
+++ b/engine/Core/Blueprints/UnitBlueprint.lua
@@ -26,6 +26,7 @@
 
 --- read more here: https://wiki.faforever.com/en/Blueprints
 ---@class UnitBlueprint: EntityBlueprint
+---@field CategoriesHash table<string, boolean>
 ---@field AI UnitBlueprintAI
 ---@field Air UnitBlueprintAir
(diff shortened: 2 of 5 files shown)


Garanas:
image
Old branch

Garanas:
image
This pull request

Hdt80bro:
Making the building unit slowly turn towards which ever direction the rolloff for the rally point would be at is something I've been thinking about for a while. Making the unit jump-face the direction looks too sudden though (I have the same issue with submarines jumping layers when built), and it does make sense for there to be a rolloff penalty for players that move the rally point the tick before the unit finishes building.

Garanas:
There is no rotating duration to prevent clipping with the factory. Especially naval factories will have this issue. Air factories (transports) will have this issue too. There is no event when the rally point changes, otherwise we could rotate during the build. I do not expect this to happen often. And if it does, you're likely not looking at the factory.

This pull request exists because of the poor pathing of units. In particular, their inability to rotate on the same spot properly. 

About the submarines: I'm sure we can find a solution for that too.

Hdt80bro (approved):
Well, sounds good then. As usual, we can change it later if people start complaining :)

PR #4288, opened 2022-10-23
FAF UI Framework Part 2

4z0t:
* Dict
* Escape Cover

### DIFF
--- a/lua/shared/Containers/Dict.lua
+++ b/lua/shared/Containers/Dict.lua
@@ -1,0 +1,230 @@
+---@type fun(data:table?) : Dict
+local _Dict
+---@class Dict
+---@field _data table<any,any>
+---@overload fun(data:table?):Dict
+Dict = ClassSimple
+{
+    ---Creates dict from given table
+    ---@param self Dict
+    ---@param data? table
+    __init = function(self, data)
+        self._data = {}
+        if data then
+            local d = self._data
+            for k, v in data do
+                d[k] = v
+            end
+        end
+    end,
+
+    ---Returns value by key
+    ---@generic K
+    ---@generic V
+    ---@param self Dict
+    ---@param key K
+    ---@return V
+    Get = function(self, key)
+        return self._data[key]
+    end,
+
+    ---Checks whether dict contains value
+    ---@param self Dict
+    ---@param value any
+    ---@return boolean
+    Contains = function(self, value)
+        for _, v in self._data do
+            if v == value then
+                return true
+            end
+        end
+        return false
+    end,
+
+    ---Creates array of values from Dict
+    ---@param self Dict
+    ---@return table
+    Values = function(self)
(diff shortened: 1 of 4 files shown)


Garanas:
Is this a dictionary?

4z0t:
Yes, it is.

4z0t:
Well, I can allow wrapping table, so, there wont be allocations, but wrapper

Garanas:
No - the reason I was asking is because it should then just be called a `Dictionary`, instead of `Dict`

lua/shared/Containers/Dict.lua
```diff
@@ -0,0 +1,230 @@
...
+    ---@param self Dict
+    ---@param data? table
+    __init = function(self, data)
+        self._data = {}
+        if data then
+            local d = self._data
+            for k, v in data do
+                d[k] = v
+            end
+        end
```
Hdt80bro, line 18:
```lua
local d = {}
self._data = d
if data then
    for ...
end
```
Personal preference :) -- this would be slightly faster for calls that provide `data` and slightly, slightly slower for calls that don't.

lua/shared/Containers/Dict.lua
```diff
@@ -0,0 +1,230 @@
...
+    ---Dict Iterator
+    ---@param self Dict
+    ---@return fun(tbl: table<any,any>, key:any):any,any
+    ---@return table<any,any>
+    Iter = function(self)
+        return self.Next, self._data
+    end,
+
+    ---Next function for dict data
+    Next = next,
```
Hdt80bro, line 158: If this is expected to be general-purpose, it should also be used in the library methods.

Hdt80bro (changes requested):
Couple of renaming & logical considerations, but the biggest thing is with use in synchronization code: if keys are allowed to be of any type, iteration order depends on a value's hash which can different for players for functions, tables, and userdata. So either limit keys to be strings (which are internalized by Lua and hash consistently) and booleans/numbers (which inherently have well-defined hashes) or document a warning and hope the user understands the risks (which is usually a forlorn one).

4z0t:
@Hdt80bro well, as was mentioned before and had been pointed out, `Containers` are primarily created for UI, not for sim.

4z0t:
In first part there is [README](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/UI-Framework/README.md)

Hdt80bro:
I'd recommend completing that first then and resolving the rest of the problems--I'd also recommend a general restructuring. I'd also advise not separating documentation from what it's documenting (especially across directories), or at least pointing readers of the code to the readme. Comments are invaluable.

4z0t:
By @Hdt80bro 's concerns I decided to put these classes only as part of UI for now. It is UI Framework for obvious reasons. 🙂

PR #4289, opened 2022-10-23
Match defeat time with pre-3744 patch

Garanas:
The defeat time was inadvertently increased. It should now better match what it was before. Also adds in log statements as to when a player is defeated, won or drew.


4z0t:
Is there a callback about defeat/victory and draw into user side?

Garanas:
There is not, would you want one?

PR #4291, opened 2022-10-23
Allow turrets to rotate to the nearest threat again

Garanas:
As per #3891 structures have no guard scan radius. The rotation logic still relied on that. It now defaults to 40 for Tech 1, 60 for Tech 2 and 120 for Tech 3. If no threats are in range, the structure will aim towards the center.


Hdt80bro:
Looks like the Lighting Storm hover PR made it in here ;)

Garanas:
Woops, you are correct. Sneaky commits. I'll fix it tomorrow

PR #4294, opened 2022-10-24
Add path checking thread for AI. Add Build condition for use in build…

relent0r:
## Description of changes
This PR introduces a path checking thread for the default AI using the navmesh class. It starts when the AI starts the BaseMonitorInitialization function(does not affect the Sorian AI). Its primary goal is to provide the AI information on which movement layers it can get to an enemies base with.

It will create and maintain a table against the AI brain with information on each enemy and which base can path to the enemy start position with which layer. This can then be accessed by builder conditions in order to allow the AI a more dynamic production capability.

This PR intentionally limits its use to just an example builder so that only the functionality needs to be reviewed. Builder implementation and balance will be performed later.

Use case examaples.
Changing Land factory build limits on island maps
Changing land unit production when land paths are not available.
Changing amphib unit production when land paths are available.
Changing early transport production on island maps.
Limiting mobile land experimental production on maps where no pathing is available.
etc

Example of the table.

info: CanPathTo Table 
info:  - Daedalus (AI: Adaptive): table: 2AF42AA0
info:  -    Alexandra (AI: Sorian): table: 2AF429B0
info:  -       ARMY_2: Land
info:  -       ARMY_3: Amphibious
info:  -       ARMY_4: Land
info:  -       ARMY_5: Amphibious
info:  -       ARMY_6: Amphibious
info:  -       MAIN: Amphibious

This shows the name of the AI at the root table. Then the name of the enemy.
It will then show which movement layer can get to the enemy start position from each of its bases.
I've attached a screenshot showing the real world visual. The AI believes it can get to the enemy AI via amphibious from its main base and 3 expansions. It believes it can get to the enemy via land for two of its expansions that are on the enemies land mass. (please note that the AI establishes a base at the time that an engineer executes an expansion builder not when the base is actually established so some entries are from bases that did not get successfully established yet)

## Test setup for the changes
Create AI's games across various map types. The developer used a combination of land, island, and plateau based maps to validate the correct movement layers were being created against the table. You can perform a log of the  aiBrain.CanPathToEnemy table to get a list of the current AI's, their enemies and the positions they can path from to the enemy start position.

CanPathToThreadCheck


lua/AI/aiutilities.lua
```diff
@@ -2959,3 +2959,82 @@ function AIFindFurthestExpansionAreaNeedsEngineer(aiBrain, locationType, radius,
...
+                --LOG('CanPathToCurrentEnemy: Land path to the enemy found! LAND map! - '..OwnIndex..' vs '..EnemyIndex..''..' Location '..k)
+                aiBrain.CanPathToEnemy[OwnIndex][EnemyIndex][k] = 'Land'
+            else
+                -- we have no path from the nav mesh....
+                --LOG('CanPathToCurrentEnemy: no path returned ')
+                local amphibPath, amphibReason = NavUtils.CanPathTo('Amphibious', locPos, {enemyX,0,enemyZ})
+                if not amphibPath then
+                    --LOG('CanPathToCurrentEnemy: No land or amphib path detected')
+                        -- No land or amphib path, we are likely on a plateu and cant go anywhere without transports.
+                        aiBrain.CanPathToEnemy[OwnIndex][EnemyIndex][k] = 'NoPath'
```
Garanas, line 3030: What if it would default to `Air`, instead of `NoPath`?

lua/AI/aiutilities.lua
```diff
@@ -2959,3 +2959,82 @@ function AIFindFurthestExpansionAreaNeedsEngineer(aiBrain, locationType, radius,
 
     return retPos, retName
 end
+
+
+--- Monitors pathing from each AI base to the current enemy start position. Used for determining which movement layers can attack an enemy.
+---@param aiBrain AIBrain
+function CanPathToCurrentEnemy(aiBrain)
```
Garanas, line 2966: Shouldn't this be a function inside the brain class? It solely relies on the brain (as its input)

lua/AI/aiutilities.lua
relent0r: Thats a good point, and a good idea it will fit with the other naming conventions.

lua/AI/aiutilities.lua
relent0r: Yea thats also a good point. I'll shift it.

lua/aibrain.lua
```diff
@@ -4960,7 +4961,85 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                    --LOG('CanPathToCurrent we have an enemy')
+                    enemyX, enemyZ = self:GetCurrentEnemy():GetArmyStartPos()
+                    -- if we don't have an enemy position then we can't search for a path. Return until we have an enemy position
+                    if not enemyX then
+                        coroutine.yield(30)
+                        break
+                    end
+                else
+                    --LOG('CanPathToCurrent no current enemy')
+                    coroutine.yield(30)
```
Garanas, line 4997: Better to use WaitTicks - that is more commonly used throughout the code base and it refers to `coroutine.yield`, see also: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/simInit.lua#L25

lua/aibrain.lua
```diff
@@ -4960,7 +4961,85 @@ AIBrain = Class(moho.aibrain_methods) {
...
+        -- added amphib vs nopath results so we can tell when we are trapped on a plateu
+        coroutine.yield(Random(5,20))
+        local NavUtils = import('/lua/sim/NavUtils.lua')
+        if not self.CanPathToEnemy then
+            self.CanPathToEnemy = {}
+        end
+
+        while true do
+            --LOG('Start path checking')
+            --LOG('CanPathToCurrent Table '..repr(aiBrain.CanPathToEnemy))
```
Garanas, line 4980: Better to remove code in comments, makes it easier to read / understand for other developers

lua/aibrain.lua
```diff
@@ -4960,7 +4961,85 @@ AIBrain = Class(moho.aibrain_methods) {
...
+            
+            --We are getting the current base position rather than the start position so we can use this for expansions.
+            for k, v in self.BuilderManagers do
```
Garanas, line 4983: For my understanding, this will check what is applicable for each base (each builder manager is a base)?

relent0r: Yes thats right. It needs to validate each base position since expansions can be on different land masses and may or may not be able to path via the different movement layers.

Originally I was going to just write this as something that fires each time a base is created but then I realised that as soon as the current enemy changes then all the path results could change with it.

lua/aibrain.lua
relent0r: ok have done this.

lua/aibrain.lua
```diff
@@ -4960,7 +4961,85 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                        break
+                    end
+                else
+                    --LOG('CanPathToCurrent no current enemy')
+                    coroutine.yield(30)
+                    break
+                end
+
+                -- Get the armyindex from the enemy
+                local EnemyIndex = ArmyBrains[self:GetCurrentEnemy():GetArmyIndex()].Nickname
```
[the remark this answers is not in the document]
relent0r: I don't have an answer for this  :) Would you rather it be the army index since thats also unique?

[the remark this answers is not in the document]
Garanas: Yes, that seems to me to be the better approach.

lua/aibrain.lua
relent0r: Ok have changed this for consistency.

lua/aibrain.lua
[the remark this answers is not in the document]
relent0r: have updated to use the army indexes instead of nicknames.

Garanas (approved):

Garanas (approved):

PR #4298, opened 2022-10-25
Add query to retrieve label of area

### LINKED ISSUE #4296: Query for graph area (UniqueLabel?) at position.
## Describe the query
I'm not sure on the naming conventions for this. Based on the NavGenerator lua I believe they are called labels so for the sake of readability I'll call them that for the rest of this description.

Function to query which label a position is currently assigned. This query will provide metadata for other objects such as markers or AI objects such a bases. 
For example when an AI creates a naval base it will query the label for the naval position, this will allow it to know if other positions are on the same pond or different ponds and which label is allocated to those ponds. 

```lua
---@param layer String
---@param position Vector
---@return Label
function GetCurrentLabel(layer, position)

end
```

## Visualize the query
This screenshot shows which layer points would have labels. Each of the 9 land masses would have separate labels/grapharea names. The query would return the label/graph for each land mass if a position on them is queried with the layer type of 'Land' or 'Amphibious' or 'Air' is used(not actually sure if Air should be valid as there will always only be one label to my knowledge).

## True / false positives
True will return the label.
If a label cannot be returned a reason may be provided. There are certain situations where a label my not be available such as when the position is out of the map bounds, or when the positions is between two labels such as the side of a mountain. I'm not totally sure how you would handle that. 

I considered requesting a 'closest' parameter that would then provide the closest label to that position if none were available at the position queried.
In my current logic when a position does not return a zone then I would get 8 points around the position and query those to see if one would return a valid zone as a fallback. This situation happens when a platoons position is inside a mountain or another area where no zone/label is defined so while its a niche situation it does happen.

Label10

Garanas:
Closes #4296

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -831,7 +831,6 @@ Callbacks.NavToggleScanLayer = function(data)
 end
 
 Callbacks.NavToggleScanLabels = function(data)
-    LOG("ToggleScanLabels")
     import("/lua/sim/NavDebug.lua").ToggleScanLabels(data)
 end
 
@@ -841,4 +840,8 @@ end
 
 Callbacks.NavDebugPathTo = function(data)
     import("/lua/sim/NavDebug.lua").PathTo(data)
+end
+
+Callbacks.NavDebugGetLabel = function(data)
+    import("/lua/sim/NavDebug.lua").GetLabel(data)
 end
--- a/lua/shared/NavGenerator.lua
+++ b/lua/shared/NavGenerator.lua
@@ -14,6 +14,10 @@ LayerColors = {
     Air = 'add8e6'
 }
 
+---@class NavDebugGetLabelState
+---@field Position Vector 
+---@field Layer NavLayers
+
 ---@class NavDebugPathToState
 ---@field Origin Vector 
 ---@field Destination Vector 
--- a/lua/sim/NavDebug.lua
+++ b/lua/sim/NavDebug.lua
@@ -67,6 +67,13 @@ function PathTo(data)
     PathToState = data
 end
 
+---@type NavDebugGetLabelState
+local GetLabelState = { }
+
+function GetLabel(data)
+    GetLabelState = data
+end
+
 function ScanOver(mouse, layer)
     if mouse then
         local over = NavGenerator.NavGrids[layer]:FindLeaf(mouse)
@@ -210,6 +217,25 @@ function Scan()
                     end
                 end
             end
+        end
+
+        if GetLabelState.Position then
+            local label, msg = NavUtils.GetLabel(GetLabelState.Layer, GetLabelState.Position)
+            if label then
+                local color = Shared.LabelToColor(label) or 'ffffff'
+                DrawCircle(GetLabelState.Position, 3.9, color)
+                DrawCircle(GetLabelState.Position, 4.0, color)
+                DrawCircle(GetLabelState.Position, 4.1, color)
+            else
+                DrawCircle(GetLabelState.Position, 3.9, '000000')
+                DrawCircle(GetLabelState.Position, 4.0, 'ffffff')
+                DrawCircle(GetLabelState.Position, 4.1, '000000')
+            end
+
+            Sync.NavDebugGetLabel = {
+                Label = label,
+                Msg = msg
+            }
         end
 
         WaitTicks(2)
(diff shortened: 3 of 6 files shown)


Garanas:
image

relent0r:
I can test this tonight by replacing my existing logic. Should make it easy to validate.
Based on what I'm seeing. To mark the label for mass points in my customer resource point table it should be as simple as doing this at the start of the game.

```
for k, v in AdaptiveResourceMarkerTableRNG do
    if not v.Labels then
        v.Labels = {} 
        if RUtils.PositionInWater(v.position)
            v.Labels.Water = NavUtils.GetLabel('Water', v.position)
        else
            v.Labels.Land = NavUtils.GetLabel('Land, v.position)
        end
    end
end
```

Garanas:
Yes, but perhaps it is best if we implement this in the base game? We can just add this information to each extractor, regardless of whether your AI is running or not.

relent0r:
> 

I like this idea in concept. Would require a bit more thinking as I'd like to fix the markers that AI are seeing on adaptive maps vs what is reality. Using the ScenarioInfo cache is not currently reliable for this reason. Having the marker table populated and label set during the CreateResourceDeposit function would help this situation. Though now that I'm looking it seems like you've already done this from line 388 in the MarkerUtils file so at least for resource markers this looks like an easy change if the AI is using the marker cache.

Garanas:
>Having the marker table populated and label set during the CreateResourceDeposit function would help this situation. Though now that I'm looking it seems like you've already done this from line 388 in the MarkerUtils file so at least for resource markers this looks like an easy change if the AI is using the marker cache.

You're correct, that is what the marker utilities are made for 😄 , I can adjust that code to populate these values as extractors are placed on the map. The navigational mesh is generated before the resources are placed.

relent0r:
I did testing on this from a few different places. I tested this on the map Valenvaron since it has multiple land masses and under water mass points so that i could test the different logic types that we might use labels for. I didn't get time to test what multiple naval ponds look like.

First was the base managers. So each AI base manager would attempt to add the label to a base attribute using its position and its layer (Land or Water).
Logging indicated that this was fine. It does point out that the developer must be aware that label numbers are not unique else they could be confused that the same labels are returned for multiple layers.

Main base example
info: Graph Area set as 2 for baseLayer type Land
Naval Expansion example
info: Graph Area set as 2 for baseLayer type Water
Land Expansion examples
info: Graph Area set as 11 for baseLayer type Land
info: Graph Area set as 16 for baseLayer type Land
info: Graph Area set as 4 for baseLayer type Land

I then tested setting labels on mass markers during the CreateResourceDeposit function call but this failed with the follow error.
info: No label reason Invalid layer type - this is likely a typo. The layer is case sensitive

I'm guessing the nav generator hadn't completed created the roots?

I then tested setting the labels a little bit later in the game on mass points.
For land masspoints this worked with logs
info: Label against masspoint is 2 for layer land
info: Label against masspoint is 2 for layer land
info: Label against masspoint is 12 for layer land
info: Label against masspoint is 16 for layer land
info: Label against masspoint is 16 for layer land

but for Water based mass points it failed with the following reason. I was using the mass marker position as it was when the resource point was created so I'm unsure why these were being returns as unpathable.
info: No Label reason is Position is unpathable
info: No Label reason is Position is unpathable

The logic I was using was

```
if GetTerrainHeight(marker.position[1], marker.position[3]) < GetSurfaceHeight(marker.position[1], marker.position[3]) then
            AdaptiveResourceMarkerTableRNG[masspoint].RNGArea = NavUtils.GetLabel('Water', marker.position)
            local label, reason = NavUtils.GetLabel('Water', marker.position)
            if not label then
                RNGLOG('No Label reason is '..reason)
            else
                RNGLOG('Label against masspoint is '..AdaptiveResourceMarkerTableRNG[masspoint].RNGArea..' for layer water')
            end
        else
            AdaptiveResourceMarkerTableRNG[masspoint].RNGArea = NavUtils.GetLabel('Land', marker.position)
            local label, reason = NavUtils.GetLabel('Land', marker.position)
            if not label then
                RNGLOG('No Label reason is '..reason)
            else
                RNGLOG('Label against masspoint is '..AdaptiveResourceMarkerTableRNG[masspoint].RNGArea..' for layer land')
            end
        end
```

Garanas:
@relent0r all labels are now unique across layers. I'll add in the resource marker logic in another pull request

PR #4300, opened 2022-10-26
Fix/ai factory manager not removing factories after upgrade.

relent0r:
Description of changes
This PR fixes a long standing issue with the default factory manager. When a factory is upgraded the original factory is never destroyed or removed from the AI's factory manager list. This in turn causes the FactoryLessAtLocation and FactoryGreaterAtLocation conditions to return incorrect results. Any AI logic that queries factory counts from this list will be invalid after the first factory has upgraded.

Note : I wasn't sure about adding the table rebuild function. But it seemed like a valid requirement for maintaining the factory list table.

Test setup for the changes
With the default code add logging to the factory manager that will provide the list of factories that the AI has (or a count). When the AI upgrades a factory you will see that the factory list increases by one instead of remaining the same because the old factory is still in the list. This in turn makes any factory counts invalid.
With this fix applied the number will remain correct after factory upgrades are completed.

### DIFF
--- a/lua/sim/FactoryBuilderManager.lua
+++ b/lua/sim/FactoryBuilderManager.lua
@@ -282,7 +282,7 @@ FactoryBuilderManager = Class(BuilderManager) {
         for k,v in guards do
             if not v.Dead and v.AssistPlatoon then
                 if self.Brain:PlatoonExists(v.AssistPlatoon) then
-                    v.AssistPlatoon:ForkThread(v.AssistPlatoon.EconAssistBodyRNG)
+                    v.AssistPlatoon:ForkThread(v.AssistPlatoon.EconAssistBody)
                 else
                     v.AssistPlatoon = nil
                 end
--- a/lua/sim/FactoryBuilderManager.lua
+++ b/lua/sim/FactoryBuilderManager.lua
@@ -278,19 +278,24 @@ FactoryBuilderManager = Class(BuilderManager) {
     ---@param factory Unit
     FactoryDestroyed = function(self, factory)
         local guards = factory:GetGuards()
+        local factoryDestroyed = false
         for k,v in guards do
             if not v.Dead and v.AssistPlatoon then
                 if self.Brain:PlatoonExists(v.AssistPlatoon) then
-                    v.AssistPlatoon:ForkThread(v.AssistPlatoon.EconAssistBody)
+                    v.AssistPlatoon:ForkThread(v.AssistPlatoon.EconAssistBodyRNG)
                 else
                     v.AssistPlatoon = nil
                 end
             end
         end
         for k,v in self.FactoryList do
-            if v == factory then
+            if (not v.Sync.id) or v.Dead then
                 self.FactoryList[k] = nil
+                factoryDestroyed = true
             end
+        end
+        if factoryDestroyed then
+            self.FactoryList = self:RebuildTable(self.FactoryList)
         end
         for k,v in self.FactoryList do
             if not v.Dead then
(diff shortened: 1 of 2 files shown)


lua/sim/FactoryBuilderManager.lua
```diff
@@ -288,10 +289,14 @@ FactoryBuilderManager = Class(BuilderManager) {
             end
         end
         for k,v in self.FactoryList do
-            if v == factory then
+            if (not v.Sync.id) or v.Dead then
```
Garanas, line 292: What is the idea behind checking the sync id?

relent0r: It was something that was done after a discussion with Sprouto when I asked him how he approached this problem in LOUD way back so I'm not actually sure if its an absolute requirement.

Sprouto: Sync.id is a functionless way of knowing if a unit has been completely removed from the game - as v.Dead should be, but as I discovered, there are edge cases where it may still be processing the death process (especially when hooking is involved) and hasn't been set to dead.    It's a wash as to which one you may wish to use - but I trust Sync.id as the final word.

relent0r: Happy to remove the sync id check btw Jip. I never tested without so never validated if that condition would ever come up.

Garanas: If you need to be sure that something is considered destroyed then you can use the global [IsDestroyed](https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Core.lua#L206). If that flags true, then the c-object is de-allocated, and the unit no longer exists (in memory).

In Lua, these are the two events you check for:
- `unit.Dead`, flagged during [OnKilled](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L1405) function
- `unit.Sync.id`, removed during [OnDestroyed](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L2229)

When a tick is processed there are a few phases:

```
- (...)
- (7) Engine determines what units are to be de-allocated (destroyed) (`IsDestroyed` returns true`
- (8) Engine processes all Lua routines
- - (...)
- - (8.9) Engine processes Lua routine that checks if unit `a` is dead via `.Dead` (returns false)
- - (9.0) Engine processes Lua routine that flags unit `a` as dead
- - (9.1) Engine processes Lua routine that checks if unit `a` is dead via `.Dead` (returns true)
- (...)
```

Because each routine (and therefore: each unit or thread) is processed one-by-one, it is very well possible that some routine is checking the `.Dead` or `.Sync.id` and get a false value, even though the unit is technically in the process of being destroyed. It is just not registered yet.

Long story short: if you need to have a safe, guaranteed check to see if a unit is destroyed then we cannot rely on any Lua logic as we have no idea of the order of execution. The use of `.Sync.id` only makes the gap smaller, as it is set during `OnDestroy` instead of `OnKilled`. 

The only safe, guaranteed check is `IsDestroyed` or `entity:BeenDestroyed`. I think internally these are equivalent.

Garanas: One example of a subtle bug that was introduced by using `.Dead` instead of `IsDestroyed` is the draw bug that was in ladder for a long time: they used `.Dead` to check. S

ometimes not both routines calling `OnKilled` was run before we checked whether both armies were defeated, and as a result one army was considered defeated _the next tick_ , and in turn they gained a victory (and the other party a defeat). By using `IsDestroyed` we fixed this subtle bug.

Garanas: And in terms of performance: yes, this is more expensive. But it significantly reduces the number of subtle bugs that are impossible to find because the behavior (read: the order in which the routines are processed) is not deterministic between games.

I'd always use `IsDestroyed` for AI code, instead of `.Dead` (or `.Sync.id`).

Garanas (approved):

Garanas (approved):

PR #4303, opened 2022-10-27
Add new translations.

Yuchenjimmy:
Add new translations. Also, several strings cannot be replaced in game, such as "Recalculate CPU rating" and "Run cpu benchmark", etc; and some of the strings seems missing from the strings_db.lua file, like "at_Default="default"". Please tell me how to solve it, thanks in advance.


Garanas:
You're correct - some are missing. That means they do not have a LOC entry. Let me guide you through on how to fix that for the next pull request.

Using Visual Studio Code you can search through all files in the repository. This assumes that you have it open as a folder:

image

Then you can use `CTRL + SHIFT + F` to search through all files. As an example:

image

The area marked with red is the search results. The area marked with yellow is the LOC id that is missing from your translation file. To add it you create a new line and write: `lobui_0426="..."`, where `...` would be your translation.

Could you again continue in a new pull request? I'll review this one and merge it later today

Yuchenjimmy:
Thanks for you guidance! But sorry I'm a little confused, lobui_0426="Recalculates your CPU rating." does exist in the Traditional Chinese loc file, which is in the folder "TW". Though I editted it, nothing changed, even editting the US loc file wasn't work.

Yuchenjimmy:
But "at_Default="Default"" still can't be replaced. Loc file code is at_Default="預設" and here is the in-game image.
[https://imgur.com/oed4Qtx](url)

Garanas:
Ah, I understand you now. I'll look into it tonight for you

Yuchenjimmy:
Ok, thank you.

Garanas:
@Yuchenjimmy I'll look at this after the release of this Sunday

Yuchenjimmy:
Thank you!

Hdt80bro:
Sorry this has been open for so long, I don't think we have anyone that can properly review translations. I took a look with Google Translate though, and it seems accurate (If you feel up for it, in the future, you might be able to do the simplified Chinese translation at the same time since I think it's pretty similar to the traditional Chinese used in Taiwan--don't feel like you have to though). I'll merge this in later this week.

Regarding the default values in combo boxes, it looks like it currently hardcoded as "(default)" unfortunately:
image
I'll open a pull request in a bit to fix that.

Hdt80bro:
By the way, I'm interested in typography but am only a little familiar with the writing system from my time with Japanese. It looks like the original translation is woefully inconsistent with punctuation and the like (for example, it uses all 3 possible comma styles: half-width comma + space `, ` full-width comma `，` and ideographic comma `、`)--if you know the modern graphical standards of writing in Chinese and the typographical expectations people have for it in games, you might be able to set the standard we use in Supreme Commander.

Hdt80bro (approved):

Yuchenjimmy:
Sorry for the really late response. And yes, the vanilla string looks pretty strange, as what you mentioned, there are also a few of spaces appearing between sentences. But maybe I'd like to correct it lastly, translating is much more important.

Yuchenjimmy:
There are some terms being used differantly between Simplified Chinese and Traditional Chinese nowadays, but I think it's not a big problem. I would use translator after finishing Traditional Chinese translations.

Yuchenjimmy:
Also, I found the "Extended graphics" in Video setting seems doesn't have a loc.
fa-deploy-fafdevelop - Microsoft Visual Studio 2023_1_15 上午 12_30_36

Garanas:
We can add that in 👍

PR #4306, opened 2022-10-28
Base manager cleanup

4z0t:


Garanas:
This appears to be just formatting, is that correct?

4z0t:
Annotations, some logic fixes

Garanas:
Could you pinpoint the logic fixes?

Garanas (approved):

Garanas:
@4z0t what if we merge these in, before you make more changes? Further changes are then part of another pull request

4z0t:
Is there a rush? If it is too necessary then ofc merge, but there are no that many changes. I'm a bit confused about fa-coop repo, that I can't change now 😔

Garanas:
The fa-coop repo is in my opinion one giant mistake 😄 

The only reason it should exist is to adjust the balance, but that is about it. All the other logic should at some point be transferred to this repository as it is only confusing and less trivial to make a release.

Garanas:
And there is no rush, but I prefer small pull requests over large ones. And this one is starting to become large

4z0t:
Just imagine Advanced Scenario Framework PR for coop :)

Garanas:
I'd prefer if you'd make the PR for the base repository 😄

4z0t:
What happened?

lua/AI/OpAI/BaseManager.lua
```diff
@@ -316,861 +365,928 @@ BaseManager = ClassSimple {
         self.Initialized = true
         if not brain.BaseManagers then
             brain.BaseManagers = {}
-            brain:PBMRemoveBuildLocation(false, 'MAIN') -- Remove main since we dont use it in ops much
+            brain:PBMRemoveBuildLocation(nil, "MAIN") -- remove main since we don't use it in ops much
```
Garanas, line 368: Is this technically the same?

lua/AI/OpAI/BaseManager.lua
```diff
@@ -316,861 +365,928 @@ BaseManager = ClassSimple {
...
     FindGroup = function(self, groupName)
-        for num, data in self.LevelNames do
+        for _, data in self.LevelNames do
             if data.Name == groupName then
                 return data
             end
         end
         return false
-    end,
+    end;
```
Garanas, line 658: There's a `;` instead of a `,` , what is the idea of that?

Garanas:
It appears to be primarily formatting. Are there any functional changes too?

lua/AI/OpAI/BaseManager.lua
```diff
@@ -288,859 +365,928 @@ BaseManager = ClassSimple {
...
-    ---@param self BaseManager      # An instance of the BaseManager class
+    ---@param self BaseManager
     ---@return boolean
     ConstructionNeedsAssister = function(self)
-        if not self:ConstructionAlwaysAssist() or self:GetConstructionEngineerCount() == 0 then
-            return false
-        end
-        return true
-    end,
+        return self:ConstructionAlwaysAssist() or self:GetConstructionEngineerCount() == 0
```
speed2CZ, line 768: Not only this change makes this condition less readable, you also managed to change the logic of it so it's broken now.

lua/AI/OpAI/BaseManager.lua
```diff
@@ -288,859 +365,928 @@ BaseManager = ClassSimple {
...
-                    -- No requirement and stop available, return upgrade name
-                    else
-                        return upgradeName
-                    end
-                end
-            else
-                error('*Base Manager Error: ' .. self.BaseName .. ', enhancement: ' .. upgradeName .. ' was not found in the unit\'s bp.')
+            if not bpUpgrade then
+                error("*Base Manager Error: " .. self.BaseName .. ", enhancement: " .. upgradeName .. ' was not found in the unit\'s bp.')
+                continue
```
speed2CZ, line 1070: No more code is executed after `error`

lua/AI/OpAI/BaseManager.lua
```diff
@@ -288,859 +365,928 @@ BaseManager = ClassSimple {
...
+    end;
 
-    -- Sets a group's priority
+    --- Sorts build groups by priority
+    ---@param self BaseManager
+    SortGroupNames = function(self)
+        table.sort(self.LevelNames, self.GroupSorter)
+    end;
+    GroupSorter = function(a, b)
+        return a.Priority < b.Priority
```
speed2CZ, line 1141: This sorts the group from lowest to highest priority, changing the original logic, which was from highest to lowest.

speed2CZ (changes requested):
Overall I can't really approve of these changes. It's called cleanup but it makes everything much less readable. Trying to optimize a lot of things that runs usually just once.
On top of that some logic changes that were surely not tested. I've caught a couple of those but I wouldn't be surprised if there were more.

Then there are some weird changes to formatting, like replacing ' ' with " ", why?
; instead of ,

My doubt of how much the individual things were tested were reinforced further by the fact that the whole base manager is just broken.

lua/AI/OpAI/BaseManager.lua
xXEddieXxx: Less readable for me, i found not much of a difference in conditions. http://lua-users.org/wiki/ExpressionsTutorial

lua/AI/OpAI/BaseManager.lua
Hdt80bro: This is not always the case, like how `pcall` doesn't always catch errors

lua/AI/OpAI/BaseManager.lua
Hdt80bro: Mmh, good catch, should be `self:ConstructionAlwaysAssist() and self:GetConstructionEngineerCount() ~= 0`
Again, I disagree about the readability, especially when the returns are as inconsistent as this file.

Hdt80bro:
> It's called cleanup but it makes everything much less readable

It's interesting that you say that; to me, it was horrendous before and much better now. Looks like this is a good trial run of the new (code style contributions)[https://github.com/FAForever/fa/pull/4318] I'm writing - I'm excited to hear your thoughts.

Of course, a few things were focused on optimization that weren't mean to increase readability (like the caricature that was `LoadDefaultBaseEngineers` that I reduced by 468 table creations)--it's all clean up though.

You're right that I didn't test it as much as I should have, though I'm not sure that pointing out that I likely couldn't have to begin with makes that worse. I'll take a look again tonight to make sure it's logically consistent - alternatively, I can remove most of those changes if we don't want to deal with that right now or just revert the commit.

Hdt80bro:
I'll let this be 4z0t's PR; we can address the other concerns in the future.

Hdt80bro:
I don't agree with all of the formatting, but overall it's an improvement, so I'll merge this

PR #4307, opened 2022-10-28
Re-organize game options 

Garanas:
Adds in various headers and moves around the options. A lot of them part of the 'interface' header even though they clearly influence gameplay. Others are in the gameplay header even though they clearly only influence what is being rendered on screen.

This will also incorporate the defaults as discussed here:
- https://forum.faforever.com/topic/3248/default-game-options?_=1666983419904

@Hdt80bro what are your thoughts on removing (in other words: by default, enabling) some of the options? As an example, sharing templates across all factions seems like a trivial option. Another is being able to see more extensive information of the unit you're hovering over (like being able to see the amount of shield hp and shield or hp regen).

---
Current organization:
**Gameplay**
  * **Camera controls**
      * Screen Edge Pans Main View
      * Arrow Keys Pan Main View
      * Pan Speed
      * Zoom Wheel Sensitivity
      * Accelerated Pan Speed Multiplier
      * Keyboard Rotation Speed
      * Accelerated Keyboard Rotate Speed Multiplier
  * **Templates**
      * Accept Build Templates
      * Template Rotation
      * All Race Templates
      * Factory Build Queue Templates
      * Visible Template Names
  * **Control groups**
      * Set behavior
      * Factory behavior
      * Double tap behavior
      * Double tap decay (in ms)
  * **Commands**
      * Ignore mode via CTRL
      * Automated Structure Ringing
      * Automatic Extractor Selection
  * **Selection threshold**
      * Default selection threshold
      * Reclaim selection threshold
      * Replay selection threshold
  * **Cursor features**
      * Depth scanning

**Interface**
  * Language
  * UI Scale
  * **Rendering**
      * Always Render Strategic Icons
      * Show Waypoint ETAs
      * Display Subtitles
      * Display World Border
      * Show Lifebars of Attached Units
      * Use Factional UI Skin
      * Bigger Strategic Build Icons
      * Always Render Custom Names
      * Force Render Enemy Lifebars
  * **Misc**
      * Multiplayer Taunts
      * Template Name Cutoff
  * **Interactions**
      * Construction Tooltip Information
      * Display Tooltips
      * Display Loading Tips
      * Economy Warnings
      * Tooltip Delay
      * Quick Exit
      * Lock Fullscreen Cursor To Window
      * Main Menu Background Movie
      * Use Factional UI Font Color
      * Enable Cycle Preview for Hotbuild
      * Cycle reset time (ms)
      * Draggable Build Queue
      * Show Factory Queue on Hover
      * Middle Click Avatars
      * Single Unit Selected Info
      * Zoom Pop Distance
      * Seperate Idle Builders
      * Display more Unit Stats
      * Show Armament Build in Factory Menu
      * Improved Unit deselection
      * Hotkey Labels
  * **Spawn menu**
      * Team Columns
      * Filter Columns
      * Split Sources
      * Type Filter Mode
      * Include No-tech Filter
      * Include ACU/Paragon Filter
      * Filter By Menu Sort


Hdt80bro:
## Option Removal

My initial reaction is to leave them in so that players can have something closer to the vanilla if they'd prefer that. I realize that we already change so much in FAF that that's a moot point though. So probably a better metric for leaving options in is whether or not changing it can interfere with how someone can play the game and the tools they have available to them. For example, template rotation: some people want to drag-build templates, others want to be able to (at least partially) rotate the templates. One option strictly prohibits the other.

The question then becomes to what extent does always enabling options like all-faction templates or extended unit information have the potential to bring unwanted effects to the player? I suspect not very much.

### All Race Templates

First off is a note about how this is poorly named and should be All Faction Templates. It seems that the original implementer was used to race terminology such as used in SC2 or BFME. But the four factions in Supreme Commander aren't primarily differentiated by what race they are, or even their species (if "race" is in a more general sense to contrast Seraephim); UEF, Aeon, and Cybran are all human (well, an argument can be made for Cybran's being Cyborgs, but I'm not touching that as it relates to race) so the original terminology of "faction" is more appropriate.

The only reason I can come up with for not immediately merging all-faction templates is if someone has extremely developed play styles for each faction that means completely different build templates that would clutter up the template space if merged--which is a distinct possibility for templates that include structures that behave differently for each faction (mostly defensive structures). So at the very least, only templates that completely contain structures that are identical in each faction should be automatically merged.

This still leaves the option of merging *all* templates or not however. So at the end of the day, we would conservatively only end up with an additional third option: "Only identical templates". But do we want to be conservative in that regard though? I tend to lean towards "yes" as a player, though I've noticed that you, as a game designer, have a tendency to want to simplify the game and remove potential options the player has to deal with :)

Note that all-faction templates is somewhat incomplete as it is--it does a raw string replacement for each blueprint instead of using the factional blueprint converter (this has been brought up for Cybran T3 Shields).

### Display more Unit Stats

This one's pretty benign. The only harm that one could possibly bring is a little more screen clutter. And in terms of what we're already doing to the UI without giving the player choice over what's happening, this should logically already be set to true and removed.

Actually, there's several other options like this: "Show armament information in build menu", "Construction tooltip information", "Single unit selected info", "Show Factory Queue on Hover"... so we would either need to group these together more clearly and add options for the rest of the UI changes we have (e.g. reclaim info, detailed veterancy, scoreboard info...) or remove them *all*. As a player of agency, I lean towards the former, though, again, I understand if you want the latter.

### Spawn menu

Something else I see as possibly removing is the Spawn menu stuff. It's a lot of clutter for not a lot of benefit to most people. In additional, take the other menus: the chat menu has a separate popup for its settings, the multi-functional menu drops down intel rings, the console has its own settings menu, the profiler will have its own tab for options... the *cheat* menu (as it is more properly called anyway, since it does more than just create units) should definitely be shunted to its own configs panel too.

### Factional UI

I'm also seeing "Use factional UI font color"--we definitely haven't been building the UI with that in mind, so that one doesn't work as expected at all.

"Use faction skin" also seems redundant due to skin cycling.

At the very least, these should be next to each other.

### Input

Arrow Keys Pan Main View--we either need to find a proper way to remove this from the console and into the hotkeys menu, or set this to true and remove it / put the option in the hotkeys menu in case of conflicts. More on how we should handle input later...

## Gameplay & Interface Dichotomy

The categorization between gameplay and interface wasn't very clear to begin with. An interface describes the boundary between two distinct things. But if we define a separate layer for gameplay, then what exactly are the two things the interface is providing a boundary for? The best answer I got is the player and the UI. However, defining an interface in terms of another interface that ultimately describes the other thing we want to contrast it against is not at all useful. Some interface options go the other way too, belonging in the `Video` tab (at least, once renamed to "Display"). It's a blurry line, especially for input options, as those affect how the player interact with both the game and the UI.
  * For example, I see the cursor features as being purely rendering cosmetics that would belong to the interface under the traditional dichotomy. However, they do indeed interact with the world and influence how the player plays the game, so could be viewed as gameplay options too.
 
### Input
 
In fact, that last point brings up a good consideration: perhaps `Input` should be its own tab, separate from `Gameplay` and `Interface` to avoid that issue. Since this is usually where keybindings go, this also raises the question of why hotkeys (which is yet another item that has developed as to render its name somewhat misleading) aren't considered options as well and in the options menu (the answer, of course, is an implementation issue with how "hotkeys" are bound in-game in a separate window). And why the options menu doesn't have the collapse and search features hotkeys do...
  * This would move all of the keyboard, mouse, Camera, input delay etc... stuff out of gameplay, but introduces a new issue that some categories are now split between three tabs:

### Aspect vs Category Division

The other thing to consider is that dividing options by their *aspect* (whether they affect how the player plays the game or how the player views the game; again, a blurry line) could be less useful for some options than grouping them by their *category* (a set of options that affect the same property).
  * For example, you currently put all but one template-related options in `Gameplay > Templates`, but this leaves one lonely template option in `Interface > Misc` out of place and hard to find.

One solution is to allow the user to navigate both perspectives by merging every option in `Gameplay` and `Interface` to a tab called `Options` (and renaming the "Options" menu to "Settings" to avoid confusion) and provide toggle that changes subtabs from grouping by aspect and category, along with a search bar like hotkeys & notify do. In fact, this would make an excellent entry vector to integrate hotkeys (to be renamed to "keybindings") *and* "Notify Management" to the settings menu and allow for greater cohesion between them (e.g. being able to put control group options right next to their keybindings, if desired). This sounds like a lot of work we don't want to do, however.

## Suggested Grouping

**Gameplay**
  * **Camera controls** - these really do feel like interface options to me, but keeping the amount of options in each tab balanced is good :)
      * Screen Edge Pans Main View
      * Zoom Wheel Sensitivity
      * Pan Speed - swap to put next to multiplier
      * Accelerated Pan Speed Multiplier
      * Keyboard Rotation Speed
      * Accelerated Keyboard Rotate Speed Multiplier
      * Zoom Pop Distance - this needs looking into, but I think it belongs here
  * **Templates**
      * Accept Build Templates - we *really* need to clarify what this does and that players can send templates
      * Template Rotation
      * All Race Templates
      * Factory Build Queue Templates
      * Visible Template Names
      * Template Name Cutoff - moved to be with its friend :)
  * **Control groups** - these need renaming / description updates to make them more clear
      * Set behavior
      * Factory behavior
      * Double tap behavior
      * Double tap decay (in ms)
  * **Commands**
      * Ignore mode via CTRL
      * Automated Structure Ringing
      * Automatic Extractor Selection
  * **Selection**
      * Middle Click Avatars - moved from `Interface`
      * Separate Idle Builders - moved from `Interface`, fixed typo
      * Improved Unit deselection - moved from `Interface`
      * Default selection threshold
      * Reclaim selection threshold
      * Replay selection threshold
  * **Cursor features** - I know you're planning to add more features to make this warrant a separate group :)
      * Depth scanning

**Interface**
  * Language\*
  * Display Subtitles - I noticed you moved this to `Rendering`, but it really belongs next to `Language`
  * UI Scale\* - We really went through all that effort to make the UI scale update dynamically and it ended up requiring a restart this whole time??
  * **HUD** - rename from "Render" since some options really had nothing to do with world rendering
      * Show Waypoint ETAs - really doesn't belong anywhere else
      * Show Lifebars of Attached Units - put lifebars together
      * Always Show Enemy Lifebars - renamed from "Force Render Enemy Lifebars"
      * Always Show Custom Names - change "Render" to "Show" to avoid confusion
      * Always Show Strategic Icons - same^
  * **Building**
      * Bigger Strategic Build Icons - I also wanted the icon options together, so placing it here (just below always show strat icons) is a compromise
      * Draggable Build Queue
      * Hotkey Labels
      * Enable Cycle Preview for Hotbuild
      * Cycle reset time (ms)
  * **UI**
      * Display Tooltips
      * Tooltip Delay
      * Use Factional UI Skin
      * Use Factional UI Font Color
      * Economy Warnings
  * **Additional Information** - consider if we really do want to keep these options
      * Display more Unit Stats
      * Single Unit Selected Info
      * Construction Tooltip Information
      * Show Armament Build in Factory Menu
      * Show Factory Queue on Hover
  * **Misc** - misc should be at the bottom
      * Main Menu Background Movie
      * Show Loading Tips - don't use "display" to avoid confusion
      * Quick Exit
      * Multiplayer Taunts

**Display** - rename from "Video"
  * Primary Adapter
  * Secondary Adapter
  * Lock Fullscreen Cursor To Window - moved from `Interface`
  * Fidelity Presets
  * Render Sky
  * Display World Border - moved from `Interface`
  * Fidelity
  * Shadow Fidelity
  * Anti-Aliasing
  * Texture Detail
  * Level Of Detail
  * Vertical Sync
  * Bloom Render
  * Experimental graphics

## Notes

Several options are linked (like "Show Tooltips" and "Tooltip Delay"), and we could do a better job hiding options that don't make sense / integrating them into a single option.

We should probably visually indicate which options require a restart.

The settings menu is kind of small and each option takes up a little more space than it needs to. This ends up making options kind of unwieldy to mange in general. The hotkey & notify managers do a much better job handling this, and we should probably update the settings menu to match at some point.

I've noticed that several of the options are missing localization and have dummy tooltips.

### Current Options Structure

Is indeed truly awful and in need of reworking:

**Gameplay**
  * Zoom Wheel Sensitivity
  * Always Render Strategic Icons
  * Construction Tooltip Information
  * Show Waypoint ETAs
  * Multiplayer Taunts
  * Screen Edge Pans Main View
  * Arrow Keys Pan Main View
  * Pan Speed
  * Accelerated Pan Speed Multiplier"
  * Keyboard Rotation Speed
  * Accept Build Templates
  * **Commands**
      * Ignore mode via CTRL
  * **Cursor features**
      * Depth scanning
  * **Selection threshold**
      * Default selection threshold
      * Reclaim selection threshold
      * Replay selection threshold
      * Automated Structure Ringing
      * Automatic Extractor Selection

**Interface**
  * Language
  * UI Scale
  * Display Subtitles
  * Display World Border
  * Display Tooltips
  * Display Loading Tips
  * Economy Warnings
  * Tooltip Delay
  * Quick Exit
  * Lock Fullscreen Cursor To Window
  * Main Menu Background Movie
  * Show Lifebars of Attached Units
  * Use Factional UI Skin
  * Use Factional UI Font Color
  * Enable Cycle Preview for Hotbuild
  * Cycle reset time (ms)
  * Bigger Strategic Build Icons
  * Template Rotation
  * Draggable Build Queue
  * Show Factory Queue on Hover
  * Middle Click Avatars
  * All Race Templates
  * Single Unit Selected Info
  * Zoom Pop Distance
  * Factory Build Queue Templates
  * Seperate Idle Builders
  * Visible Template Names
  * Template Name Cutoff
  * Display more Unit Stats
  * Always Render Custom Names
  * Force Render Enemy Lifebars
  * Show Armament Build in Factory Menu
  * Improved Unit deselection
  * Hotkey Labels
  * **Control groups**
      * Set behavior
      * Factory behavior
      * Double tap behavior
      * Double tap decay (in ms)
  * **Spawn menu**
      * Team Columns
      * Filter Columns
      * Split Sources
      * Type Filter Mode
      * Include No-tech Filter
      * Include ACU/Paragon Filter
      * Filter By Menu Sort

**Video**
  * Primary Adapter
  * Secondary Adapter
  * Fidelity Presets
  * Render Sky
  * Fidelity
  * Shadow Fidelity
  * Anti-Aliasing
  * Texture Detail
  * Level Of Detail
  * Vertical Sync
  * Bloom Render
  * Experimental graphics

Garanas:
Good suggestions, I'll re-order them. You're referring to Vanilla, by far more than half of our current options are introduced because they were UI mods. As an example, this is the number of options in Vanilla:

image

image

Garanas:
@Hdt80bro Processed your suggestions, some still require word changes (including loc)

PR #4308, opened 2022-10-28
Nav mesh - integrate navigational mesh with marker utilities

Garanas:
image

Stores the nav layer (land, or amph) and the nav label of resource markers. Requires the use of the `MarkerUtilities` framework. The framework is now also further annotated and a few bugs fixed.

You can find the map / marker utilities here:

image


relent0r:
An example of a map that has issues with the mass marker labels.

image

Garanas:
What map is that?

Garanas:
Problem solved via #4313 , which is a band-aid for now, as it doesn't guarantee that it doesn't happen.

PR #4310, opened 2022-10-29
Fix and cleanup OtherArmyUnitCountBuildConditions.lua

4z0t:


lua/editor/OtherArmyUnitCountBuildConditions.lua
```diff
@@ -9,128 +9,122 @@
...
 --****************************************************************************
 
+---comment
 ---@param aiBrain AIBrain
----@param targetBrain string
----@param numReq integer
+---@param targetBrains string[]
+---@param numReq number
 ---@param category EntityCategory
+---@param compareType '>='|"<="|'=='|'>'|'<'|
```
Hdt80bro, line 17: We should really alias the type inside `compareType` one of these days...
also, add something like `defaults to ">="` to the parameter and mark it as optional

lua/editor/OtherArmyUnitCountBuildConditions.lua
```diff
@@ -9,128 +9,122 @@
...
-    local testBrain = ArmyBrains[GetFocusArmy()]
+function FocusBrainBeingBuiltOrActiveCategoryCompare(aiBrain, numReq, categories, compareType)
     local num = 0
-    for k,v in categories do
-        num = num + testBrain:GetBlueprintStat('Units_BeingBuilt', v)
-        num = num + testBrain:GetBlueprintStat('Units_Active', v)
+    local tblArmy = ListArmies()
+    for iArmy, strArmy in pairs(tblArmy) do
+        if ScenarioInfo.ArmySetup[strArmy].Human then
+            local testBrain = GetArmyBrain(strArmy)
```
Hdt80bro, line 105: Would using `ArmyBrains[iArmy]` work as well?

lua/editor/OtherArmyUnitCountBuildConditions.lua
```diff
@@ -9,128 +9,122 @@
...
-            testBrain = v
-            break
+function BrainsCompareNumCategory(aiBrain, targetBrains, numReq, category, compareType)
+    local num = 0
+
+    local targetBrainSet = {}
+    for _, brain in targetBrains do
+        if brain == 'HumanPlayers' then
+            local tblArmy = ListArmies()
+            for _, strArmy in pairs(tblArmy) do
```
Hdt80bro, line 26: Looks like the code you copied was wrong, should be `for _, strArmy in ipairs(ListArmies()) do` since it's an array (not that there will end up being a difference)

lua/editor/OtherArmyUnitCountBuildConditions.lua
```diff
@@ -9,128 +9,122 @@
...
+                if ScenarioInfo.ArmySetup[strArmy].Human then
+                    targetBrainSet[ScenarioInfo.ArmySetup[strArmy].ArmyName] = true
```
Hdt80bro, line 28: I recommend storing `ScenarioInfo.ArmySetup[strArmy]` to a local to improve readability and efficiency.

lua/editor/OtherArmyUnitCountBuildConditions.lua
```diff
@@ -9,128 +9,122 @@
...
+                end
+            end
+        else
+            targetBrainSet[brain] = true
+        end
+    end
+
+    for brain, _ in targetBrainSet do
+        for _, testBrain in ArmyBrains do
+            if testBrain.Name == brain then
```
Hdt80bro, line 38: How about
```lua
for _, testBrain in ipairs(ArmyBrains) do
    if targetBrainSet[testBrain.Name] then
```

Hdt80bro (commented):
Haven't tested it--I don't know enough about `SimObjectives` to know if the fix is proper--but it looks like it works!

lua/editor/OtherArmyUnitCountBuildConditions.lua
Hdt80bro: I was imagining
```lua

Comment (Hdt80bro) [reply lua/editor/OtherArmyUnitCountBuildConditions.lua]: I was imagining
```lua
local armySetup = ScenarioInfo.ArmySetup[strArmy]
if armySetup.Human then
    targetBrainSet[armySetup.ArmyName] = true
end
```
to also capture the `[strArmy]` table access, but what you did was actually better! If we were *really* interested in optimization, we could combine the ideas and do something like
```lua
local armySetups = ScenarioInfo.ArmySetup
...
for _, strArmy in ipairs(tblArmy) do
    local setup = armySetups[strArmy]
    if setup.Human then
        targetBrainSet[setup.ArmyName] = true
    end
end
```
though it's not worth holding up just for that

lua/editor/OtherArmyUnitCountBuildConditions.lua
Hdt80bro: The feedback for `ListArmies()` actually applied here as well, but I don't like repeating the same comments so I generally only mark the first occurrence. I'm beginning to see that it's less effective to add comments that are ambiguous whether they're line-specific or file-specific. Let me know what you think. It also looks like we could have done the same optimization with `ScenarioInfo.ArmySetup`--and I still wonder if the table access through `ArmyBrains` that's theoretically faster than a function call through `GetArmyBrain` does the same thing.

All-in-all, the 3 things I'm bringing up aren't important, so I'm good to merge this.

Hdt80bro (approved):

lua/editor/OtherArmyUnitCountBuildConditions.lua
```diff
@@ -9,128 +9,127 @@
...
-            break
-        end
-    end
-    local numUnits = testBrain:GetCurrentUnits(category)
-    if numUnits <= numReq then
-        return true
-    else
-        return false
-    end
+function BrainGreaterThanOrEqualNumCategory(aiBrain, targetBrain, numReq, category)
```
speed2CZ, line 87: `BrainGreaterThanOrEqualNumCategory` is suppose to take single brain as a input, Since as the name suggest it's `BrainGreater` not `BrainsGreater`
Same for the others. So the single brain should be converted into a table before sending it into `BrainsCompareNumCategory`

4z0t: Well, I see that all references to it use `targetBrain` as table of strings, thus you are not right.

speed2CZ: Most of the conditions operate on a single brain, `BrainsCompareNumCategory` was a first condition added to support multiple brains.
I've searched for the use cases and there is some copy pasted error in shadow's missions where he's using the wrong functions for testing multiple brains while it should go all into `BrainsCompareNumCategory`

Garanas: @4z0t could you pick this up and fix it with a new pull request?

4z0t: @Garans already opened a PR

4z0t: #4319

PR #4311, opened 2022-10-29
MiscBuildConditions.lua cleanup

4z0t:


Hdt80bro (approved):
Looking much better!
Some of your simplifications no longer guarantee that a boolean is always returned (particularly for existence checks), but this is for the editor, so it's probably fine. In that regard, I'm not sure you need to be as careful about returning `false` instead of `nil` as you are--in fact, since LuaPlus doesn't throw errors for `nil` attributes, equality (but not comparison) checks can be simplified from `a.b and a.b == c or false` to `a.b == c` (comparisons don't like nils however, so they still need to check for existence). In the same vein, I'd also store the values of table accesses to a local variable for reuse, but what you have is fine as well.

Garanas:
This is not for the editor - these conditions were initially defined in the editor but are actively used by AIs. @relent0r or @Uveso 
 could you give us your thoughts on these changes too?

relent0r:
I'm don't comprehend the condition monitor as well as Uveso so he'll have the best understanding on any negative effects. These are used by both the skirmish and campaign AI.
Is the performance gain worth the increased effort in troubleshooting these during normal AI development? e.g if the AI isn't doing something that I think it should currently its easy to add logs to all the conditions to figure out which ones are failing, like this it would be more effort to perform the same troubleshooting.

4z0t:
Do I have to add `or false` everywhere needed?

Garanas:
@4z0t as relent0r states these changes make it less trivial to debug. What is the purpose of cleaning up these files, beyond the code style?

4z0t:
@Garanas it reduces amount of bytecode due to removing if statement

Garanas:
Yes, but is that a noticeable performance improvement in practice?

relent0r:
> 

I'd like to get Uveso's thoughts on that. The code that performs the checks lives here.
https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/BrainConditionsMonitor.lua

It didn't 'look' like there was anything that would trip up if the response was nil. Anyone troubleshooting the condition monitor would need to be aware that a false response is no longer guaranteed.

Hdt80bro:
> @Garanas it reduces amount of bytecode due to removing if statement

Actually, the bytecode is exactly the same: Lua comparisons don't produce values on the stacks, so anything like
```lua
local a = b > c
```
will get translated to
```
GREATER b c
    JUMP 1
LOADBOOL a true
LOADBOOL a false
```

4z0t:
I would say they differ, take a look at 
```lua
function c(a,b)
return a > b
end
```
and

```lua
function c(a,b)
if a > b then
return true
end
return false
end
```
last one has one more op.

Hdt80bro:
This also means the statement
```lua
a and b or c
```
more properly evaluates as something like
```
if a then
    if b then
        yield b
    end
end
yield c
```
instead of
```
if a then
    yield b
end
return c
```
or, in other words, if `b` is `nil` or `false` then whatever's in `c` will get yielded

Hdt80bro:
> last one has one more op.

Mmh, that goes back to the Lua VM does not treat booleans expressions (or even booleans themselves) as fluently as the Lua language does, and that the compiler is unable to return anything that isn't in a register. So `return true` will translate as `LOADBOOL a true; RETURN a`. So having one return statement will indeed have one less opcode (which is untrue for returning registers).

When considering performance, in my previous testing, `return a and b or c` will return `c` faster than when unrolling it (due to being able to reuse the register assignment in the test for the return instead of moving it for the return), but will also return `b` slower (due to the extra testing it does). This can be used for manual branch prediction, but that's highly unrecommended in general. The difference in function size is negligible (4 bytes), especially compared to how often it gets used and the size of the rest of the codebase / memory used. Making worthwhile optimizations is something I struggle with myself ;) (which you can probably guess from how much I've investigated this in the past).

Uveso (approved):
I like the changes.
Even if the changes for the Lua interpreter workflow are small, it will benefit from the intensive use by the Brain Condition Monitor.

While testing i don't saw any issue with false/NIL.

It's true, with these changes, it's harder to debug these features.
But these functions are so simple, it should not be a problem to recreate the whole function with trace/debug lines.

From my point of view, this PR is a welcome optimization.

Hdt80bro:
@4z0t I've pushed my suggestions based on the discussion so far--you can revise further (or revert of my changes :) ) as you'd like

Hdt80bro:
It does look like `LessThanMapWaterRatio` and `GreaterThanMapWaterRatio` are backwards by the way--does anyone have any code that uses those functions?

relent0r:
> It does look like `LessThanMapWaterRatio` and `GreaterThanMapWaterRatio` are backwards by the way--does anyone have any code that uses those functions?

Not sure why you believe they are backwards, 
GreaterThanMapWaterRatio =  if greater than x return true
LessThanMapWaterRatio = if less than x return true

I don't use these conditions but I do use the underlying c function GetMapWaterRatio. Since water in maps doesn't change I set an attribute on the aiBrain on map start and just reference that (its between 0 and 1, x by 100 to get the percentage etc).

I would be tempted to do the same with the default AI since I can see that nothing is actually using these function conditions within FAF but I'll be implementing some understanding of how much water is on a map for the AI so it can set a priority on its naval expansions, though there are a couple of other metrics (like being able to hit mass point markers from water) that I would want it to take into account as well.

relent0r:
Just as an example. We don't have priority functions enabled for the default faf code currently. But with mine I use a dynamic priority adjustment as part of the AI builder like this (p.s not a high level programmer so there are probably more efficient ways).

```
local NavalExpansionAdjust = function(self, aiBrain, builderManager)
    if aiBrain.MapWaterRatio < 0.20 then
        return 0
    elseif aiBrain.MapWaterRatio < 0.30 then
        return 200
    elseif aiBrain.MapWaterRatio < 0.40 then
        return 400
    elseif aiBrain.MapWaterRatio < 0.60 then
        return 650
    else
        return 750
    end
end
```

Hdt80bro:
> Not sure why you believe they are backwards,
> GreaterThanMapWaterRatio = if greater than x return true
> LessThanMapWaterRatio = if less than x return true

It's just not the operator that matter, its the argument order too. Those functions appear to take the parameters exactly backwards. Compare them to the rest of the comparison functions, some of which I actually just swapped the operands and operators around for so that *they* made sense
```lua
function LessThanMapWaterRatio(aiBrain, num)
    return aiBrain:GetMapWaterRatio() < num
end
```
This function will return when `num` is *greater* than the map water ratio

In comparison, the old function
```lua
function MissionNumberLessOrEqual(aiBrain, num)
    if ScenarioInfo.MissionNumber and num >= ScenarioInfo.MissionNumber then
        return true
    else
        return false
    end
end
```
uses the greater-than-or-equal-to operator, which is the "wrong" operator, but is still right because the arguments are also swapped.

relent0r:
Ah yup I get what you mean. There are alot of those cases across the condition files. I always assumed different cultures might interpret positioning differently similar to how Japanese read books from the back to the front :)

PR #4313, opened 2022-10-30
Nav mesh - introduce dynamic compression threshold

Garanas:
The compression threshold is used to prevent small neighbors that usually have little impact on the final paths, while they drastically push the memory requirements for the navigational mesh.

With this pull request we make it dynamic: 
- 5x5, 10x10, 20x20 maps have a 2x2 threshold for land, amph and hover. Naval has a 4x4 threshold
- 40x40, 80x80 maps have a 4x4 threshold for land, amph and hover. Naval has a 8x8 threshold

The idea behind it is that usually as maps increase in size the number of narrow corridors is essentially reduced to zero.


Garanas:
This is a band-aid solution to what @relent0r describes here: https://github.com/FAForever/fa/pull/4308#issuecomment-1296019233

PR #4320, opened 2022-11-01
More annotations / code changes

MrRowey:
Filles that have been affected in this PR
- `lua/sim/EngineerManager.lua` = annotation work
- `effects/Entities/SCUDeath01/SCUDeath01_script.lua` = File Tidy-up & Replaced Deprecated code
- `lua/aeonprojectiles.lua` = Some Commenting changes
- `lua/aeonweapons.lua` = Some Annotations work
- `lua/aibrain.lua` = added missing annotations & some commenting  rework
- `lua/editor/AMPlatoonHelperFunctions.lua` = Move functions around so functions could see each other
- `lua/editor/EconomyBuildConditions.lua` = move import to bottom of page
- `lua/sim/EngineerManager.lua` = Some annotation and code work
- `lua/ui/campaign/campaignmanager.lua` = commenting rework
- `lua/ui/campaign/campaignmovies.lua` = commenting rework
- `lua/ui/campaign/operationvars.lua` = commenting rework
- `lua/user/tasks/Tasks.lua` = Move functions around so functions could see each other


lua/sim/EngineerManager.lua
```diff
@@ -190,7 +190,7 @@ EngineerManager = Class(BuilderManager) {
     ---@param pauseVal number
     ---@param unitCheckFunc any
     ---@param category EntityCategory
-    ---@return boolean
+    ---@return boolean | number
```
Hdt80bro, line 193: Similar comments with how booleans are used for signalling for the next several methods; `@return number | true`

lua/sim/EngineerManager.lua
```diff
@@ -671,7 +672,7 @@ EngineerManager = Class(BuilderManager) {
 
     ---@param self EngineerManager
     ---@param unit Unit
-    ---@param finishedUnit boolean
+    ---@param finishedUnit Unit|UserUnit
```
Hdt80bro, line 675: `EntityCategoryContains` does work with both `Unit` and `UserUnit`, but each Lua state only uses one of them. In this case, engineer manager is in the sim, so it'll be `Unit`.

Hdt80bro (changes requested):
Nice work!

lua/aeonweapons.lua
```diff
@@ -135,9 +135,9 @@ ADFTractorClaw = Class(Weapon) {
...
     ---@param self ADFTractorClaw
-    ---@param blip Blip | Unit
-    ---@return Unit | boolean
-    GetUnitBehindTarget = function(self, blip)
+    ---@param blip Blip
+    ---@return Unit | false | Blip --[[@as Unit]]
+        GetUnitBehindTarget = function(self, blip)
         if IsUnit(blip) then
             -- return the unit
             return blip
```
Hdt80bro, line 143: This function always returns the Unit of the Blip (unless it doesn't have one), so the return type should be `@return Unit | false`. When I was talking about adding `--[[@as Unit]]`, that would have gone on the actual return on line 143 (not the annotation), like `return blip --[[@as Unit]]` in case an actual Unit got passed instead of a Blip (but it isn't necessary to add it). Come to think of it, this also means that the `blip` parameter should have the type `Blip | Unit` as well.

lua/aeonweapons.lua
```diff
@@ -519,7 +519,7 @@ AANDepthChargeBombWeapon = Class(DefaultProjectileWeapon) {
 
     ---@param self AANDepthChargeBombWeapon
     ---@param bone Bone
-    ---@return Projectile
+    ---@return Projectile | nil
```
Hdt80bro, line 522: This isn't `nil`, `CreateProjectile` just doesn't have the correct return type.

lua/aeonweapons.lua
```diff
@@ -551,7 +551,7 @@ AANDepthChargeBombWeapon02 = Class(DefaultProjectileWeapon) {
 
     ---@param self AANDepthChargeBombWeapon02
     ---@param bone Bone
-    ---@return Projectile
+    ---@return Projectile | nil
```
Hdt80bro, line 554: ^

lua/aeonweapons.lua
```diff
@@ -583,7 +583,7 @@ AANTorpedoCluster = Class(DefaultProjectileWeapon) {
 
     ---@param self AANTorpedoCluster
     ---@param bone Bone
-    ---@return Projectile
+    ---@return Projectile | nil
```
Hdt80bro, line 586: ^

lua/aibrain.lua
```diff
@@ -3473,20 +3472,20 @@ AIBrain = Class(moho.aibrain_methods) {
...
-            numFactories = table.getn(gates)
+            numFactories = TableGetn(gates)
         end
 
         return numFactories
     end,
 
     ---@param self AIBrain
-    ---@param platoon Platoon
+    ---@param platoon Unit
```
Hdt80bro, line 3488: It looks like this is actually a type we haven't named yet (set up around lines 2269-2297). Leave as `any` for now.

lua/sim/EngineerManager.lua
```diff
@@ -226,7 +226,7 @@ EngineerManager = Class(BuilderManager) {
     ---@param pauseVal number
     ---@param unitCheckFunc any
     ---@param category EntityCategory
-    ---@return boolean
+    ---@return boolean | number
```
Hdt80bro, line 229: `@return number | true`

lua/sim/EngineerManager.lua
```diff
@@ -313,11 +313,12 @@ EngineerManager = Class(BuilderManager) {
     ---@param unit Unit
     ---@param econ any unused
     ---@param pauseVal number
-    ---@return boolean
+    ---@return boolean | number
```
Hdt80bro, line 316: ^

lua/user/tasks/Tasks.lua
```diff
@@ -2,35 +2,65 @@
 
 local AbilityDefinition = import("/lua/abilitydefinition.lua").abilities
 
+AbilityUnits = {}
+AbilityRangeCheckUnits = {}
 
-function GetUnitsScript(TaskName, SelectedUnits, data)
```
Hdt80bro, line 6: I discourage moving functions around like this since it breaks up the logical flow of the file, even though it "creates" error in the extension--that is a problem with the extension, not the code.

Hdt80bro (changes requested):

MrRowey:
@Hdt80bro ill get to these when i can :) got 2 week before i break up for Christmas so will look at them then :)

MrRowey:
@Hdt80bro this one should be done now

Hdt80bro (approved):

PR #4321, opened 2022-11-01
init.lua Option to deprecate mods by version number

ErikMekkes:
This might be an interesting addition ?

If it is not something you'd like, there are two fixes from the earlier 2 commits that should be applied regardless:
- `line 81: deprecatedMods = LowerHashTable(deprecatedMods)` should be moved down after the latest new entries
- `line 88: deprecatedMods["EzReclaim"] = true` should be "EzR" as that is the Mod's folder name

Tested without issues against the 83 mods I currently have in my mods folder.
Properly handles integers and the annoying 1.1 != 1.10 and 1.9 < 1.10 convention of version numbering.


Garanas:
I approve - would help a lot. 

> so supporting a simple major.minor string like you have now seems like a better contribution goal for this PR.

I agree, let us start simple.

init.lua
```diff
@@ -353,7 +363,92 @@ local function MountModContent(dir)
...
+        if not info_file then
+            LOG("Mod info file is not readable")
+            continue
+        end
+        local mod_version = nil
+        local mod_version_a = nil
+        local mod_version_b = nil
+        while true do
+            -- read next line until out of lines
+            local line = info_file:read("*line")
```
Garanas, line 381: if I recall correct you can load in other files the 'usual' way. You'll just need to be careful to not pollute the global environment of the init file (and by doing so, accidentally allow mods to edit it one way or another).

ErikMekkes: after reading up on this... I can indeed completely avoid adding to the global environment with this pattern:
```
local mod_info = loadfile(dir .. "/" .. mod .. "/" .. infoFile)
local mod_info_env = { }
setfenv(mod_info, mod_info_env)
mod_info()
```
This allows access to mod_info_env.version which is fully local and wont be present / accessible anymore outside the current function scope.

Unfortunately loadfile brings back a problem with the version notation, it preserves the number type, making it a float for which I couldn't think of a good way to parse it back into major.minor notation that preserves `1.1 != 1.10` as they both become 1.1000000238419. I avoided this by picking a string value for `deprecatedMods["name"] = "1.1"` instead of a `1.1` number value.

it would be possible to avoid it again by also making the mod_info version values a string, but that is too crazy a change to expect every mod_info to update the version value... unless i'm missing more tricks I think that brings me back to line reading for now 😟 But i did learn some environment tricks 😄

Hdt80bro: Rats, it'll have to do then

Hdt80bro:
@ErikMekkes sorry to hijack your PR like this, but we're currently redoing the contributor workflow, so I thought I'd help you out instead of making you jump through a bunch of hoops (we're also somewhat on time-crunch this week). Does everything still look good?

(Don't worry @Garanas , I'll copy the shared init files before merging)

Hdt80bro (approved):

ErikMekkes:
No problem, opened it to show the idea and a basic functional version so upgrade it as much as you like. I've checked out your changes on my end, much cleaner and still working fine 👍

re: future generalization, I've created a working version class that can be loaded into init.lua, it supports the regular comparison operators. Could use a better name and it might be a bit overkill for now, but maybe nice if more use cases for version numbers show up.

[VersionClass.lua.txt](https://github.com/FAForever/fa/files/9919320/VersionClass.lua.txt)

Garanas:
Note that `init.lua` is technically not used, except by Nomads I think.

Do we have mods to test this with? And what if the version field does not represent a number, but say the character `b`? And what if it is missing all together?

It may sound silly, but if the init file crashes then the player ends up with a broken game.

ErikMekkes:
I've been careful at avoiding errors from unexpected version formatting lines or values

> Note that init.lua is technically not used, except by Nomads I think. Do we have mods to test this with?

I didnt mention this properly, but my testing was done by copying over the changes to the separate init_dev.lua I use to launch my fa dev workspace. I wasn't sure which init's (if any) you would want changed so I only committed and pulled this one. 

I've used that to launch normal games on the latest fafdevelop version while adding bunch of deprecated version variation lines for various mods as i was testing ways to handle the major.minor / float logic. Resulting code does not have any issues reading a version from the 83 mods i have accessible in my folder. (when it was still trying to retrieve the version number for all of them, only does this for deprecated mods since hdt's latest edits).

I did not encounter any alphabetic version numbers in any real mods, but can confirm (tested) that they do not result in a crash. currently if not marked as deprecated they won't be even be checked, if they were marked deprecated and need to be checked then tonumber() results in a disabled mod with a log message that it was disabled because it does not produce a valid version number, and cant be checked against the required one.

Garanas (approved):

Hdt80bro:
> re: future generalization, I've created a working version class that can be loaded into init.lua, it supports the regular comparison operators. Could use a better name and it might be a bit overkill for now, but maybe nice if more use cases for version numbers show up.

I like the idea - for now, I'm leaning towards getting these changes merged in and if we find we want more generalization we can open a new PR

Hdt80bro:
Actually @ErikMekkes , before I merge, what was the reasoning behind making mod processing no longer use the lowercase version of the mod name?

ErikMekkes:
ah yea we could re-add that, I was having errors from LowerHashTable() while experimenting with keys but that's no longer a problem.

I didn't worry much about it as they're still quite error prone anyway and should probably just be double checked carefully. (found a misspelled one and LowerHashTable was in the wrong place before new entries, breaking all the newly added ones)

PR #4322, opened 2022-11-01
Changes some using files to prevent Intellisense popups

MrRowey:


units/DAA0206/DAA0206_script.lua
```diff
@@ -1,16 +1,12 @@
...
+-- File :  /cdimage/units/DAA0206/DAA0206_script.lua
+-- Author(s): Dru Staltman, Eric Williamson, Gordon Duclos, Greg Kohne
+-- Summary : Aeon Guided Missile Script
+-- Copyright © 2007 Gas Powered Games, Inc.  All rights reserved.
+------------------------------------------------------------------------------
+
 local AAirUnit = import('/lua/aeonunits.lua').AAirUnit
 local DefaultProjectileWeapon = import('/lua/sim/defaultweapons.lua').DefaultProjectileWeapon
-local EffectTemplate = import('/lua/EffectTemplates.lua')
-local EffectUtils = import('/lua/effectutilities.lua')
```
Garanas, line 13: If they are unused it is better to put them at the bottom of the file

units/DSLK004/DSLK004_script.lua
```diff
@@ -1,19 +1,16 @@
...
+-- Copyright © 2008 Blade Braver!
+------------------------------------------------------------------------------
 
 local SLandUnit = import('/lua/seraphimunits.lua').SLandUnit
 --local CollisionBeamFile = import('/lua/kirvesbeams.lua')
 local DefaultBeamWeapon = import('/lua/sim/DefaultWeapons.lua').DefaultBeamWeapon
 --local Dummy = import('/lua/kirvesweapons.lua').Dummy
 local EffectTemplate = import('/lua/EffectTemplates.lua')
 
-local CollisionBeam = import('/lua/sim/CollisionBeam.lua').CollisionBeam
```
Garanas, line 16: Is this a safe change on its own?

units/UAL0001/UAL0001_script.lua
```diff
@@ -61,8 +61,8 @@ UAL0001 = Class(ACUUnit) {
             local bp = self:GetBlueprint().Enhancements[enh]
             local bpEcon = self:GetBlueprint().Economy
             if not bp then return end
-            self:SetProductionPerSecondEnergy(bp.ProductionPerSecondEnergy + bpEcon.ProductionPerSecondEnergy or 0)
```
Garanas, line 64: This change isn't logically the same, what am I missing that the extension would suggest this?

units/DSLK004/DSLK004_script.lua
MrRowey: The Variable is not even being called in this particular file. by renaming removed the Intellisense saying that it uses else where under a different name

units/UAL0001/UAL0001_script.lua
MrRowey: Logically it is ? I've done this one before and hdt said it was fine as your doing the bp calculation together first or it = 0 ? or am i miss interrupting how this should be working as the Intellisense is happy once i did this.

units/UAS0401/UAS0401_script.lua
```diff
@@ -12,7 +12,7 @@ local ADFCannonOblivionWeapon = WeaponsFile.ADFCannonOblivionWeapon02
 local AANChronoTorpedoWeapon = WeaponsFile.AANChronoTorpedoWeapon
 local AIFQuasarAntiTorpedoWeapon = WeaponsFile.AIFQuasarAntiTorpedoWeapon
 
-local CreateAeonTempestBuildingEffects = import("/lua/effectutilities.lua").CreateAeonTempestBuildingEffects
```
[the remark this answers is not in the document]
MrRowey: xD then we need to change that file name then as this is a common error that the Intellisense though with the file name being "different"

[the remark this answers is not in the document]
Garanas: Ah, gotcha. Maybe we can adjust the plugin instead to always recommend lower path import strings. @hahn-kev would that be possible?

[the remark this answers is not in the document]
MrRowey: this can be both either it says the FileName is diffrent or that the Import itself is diffrent EG `explosion` in one and then `Explosion` in another file all thats diffrent is the capital

units/URA0107/URA0107_Script.lua
```diff
@@ -8,7 +8,8 @@
 -- **  Copyright © 2006 Gas Powered Games, Inc.  All rights reserved.
 -- ****************************************************************************
 
-local AirTransport = import("/lua/defaultunits.lua").AirTransport
+local AirTransport = import('/lua/defaultunits.lua').AirTransport
+local Explosion = import('/lua/defaultexplosions.lua')
 local explosion = import("/lua/defaultexplosions.lua")
```
Hdt80bro, line 13: I'm not sure leaving both in here is necessary for a file this small. Of course, it'd be the safer thing to do, but at that point you might as well not rename it I think.

Hdt80bro (changes requested):

units/DSLK004/DSLK004_script.lua
Hdt80bro: I think that was due to the filename having different capitalization in different files, which is incidentally something we just fixed. I don't see the intellisense error, so I don't think renaming this is necessary.

units/UAL0401/UAL0401_Script.lua
```diff
@@ -179,3 +177,6 @@ UAL0401 = Class(AWalkingLandUnit) {
 }
 
 TypeClass = UAL0401
+
+-- Kept for Mod Backwards Compatability
+local Utilities = import("/lua/utilities.lua")
```
Hdt80bro, line 182: If you rename it from lowercase, it doesn't keep any compatibility <--  check the spelling on the word too ;)
Likewise, perhaps we need `local explosion = Explosion` down here too (and something similar in the rest of the files we rename as well).

Hdt80bro (changes requested):

Hdt80bro (approved):
Looking good! @Garanas does renaming import modules (e.g. `explosion` t o`Explosion`) cause compatibility concerns?

MrRowey:
@Hdt80bro anything i need to do here to get this finalised ?

Garanas:
I see I got pinged. Renaming is a compatibility issue, unless the file is not used by any of the most common mods.

MrRowey:
@Hdt80bro @Garanas revert the import renaming so this should be all good to merge

Issue #4295, opened 2022-10-24
Nav mesh - improve testability and performance of the Heap data structure


Garanas:
Description
-------------

A priority queue is required for path finding. We [implemented](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/NavDatastructures.lua#L27) this as a heap. This heap is specifically written to support path finding, it is not a generalized heap. This reduces the overhead of the heap.

If you're unfamiliar with the data structure it is best to read up about it:
- https://www.youtube.com/watch?v=t0Cq6tVNRBA
- https://www.geeksforgeeks.org/introduction-to-heap-data-structure-and-algorithm-tutorials/

Note that we use a min-heap: the smallest element is at the top.

Course of action
----------------

This issue consists of two parts:
- [x] (1) Introduce a test suite for the heap, that is run as part of the build workflow of Github
- [ ] (2) Improve the performance of the heap

For (1) you can use the [existing tests](https://github.com/FAForever/fa/tree/deploy/fafdevelop/tests) as an example. Additional test files are automatically picked up. You'll need to write various tests to proof various properties. As a few examples:
- If you add 10 items, the heap has a size of 10
- If you add 10 items, the items are in the correct order
- If you add 10 items, and then extract 10 items they are in a sorted order from small to large
- (...)

There can be other interesting properties to test. But these three are the most important properties: they help us make sure that the heap operates as expected.

For (2) there are various places where we can easily skip table operations when we efficiently use the `Size` value of the heap. As an example, take this snippet:

```lua
    ---@param self NavPathToHeap
    ---@param index number
    Heapify = function(self, index)
        -- find the left / right child and the parent.
        local parent = index
        local left = self:ToLeftChild(index)
        local right = self:ToRightChild(index)

        -- find the best of the two, we assume the left child is.
        local min = left

        -- if there is a right child, then there always has to be a left child. Hence, we can now assume there's both a right and a left child.
        -- compare the two: if right is smaller, then assign min = right. Else, keep min on left.
        if self.Heap[right] then
            if (self.Heap[right].AcquiredCosts + self.Heap[right].ExpectedCosts) < (self.Heap[left].AcquiredCosts + self.Heap[left].ExpectedCosts) then
                min = right
            end
        end

        -- if there is a child, compare the lowest child (is given due to code above) with our parent. 
        -- If it's smaller, switch the parent and the lowest child.
        if self.Heap[min] then
[...]

PR #4323, opened 2022-11-02
[4295] Part zero - fix bugs found in luft.lua

Zjonn:
## Introduction
Please have in mind that this is my first contact with **Lua**.
I've started working on first task from #4295 and found out that `equal` function in Luft is not working (always returns `true`).
I created this PR to address found problems and provide potential improvements

### Luft
- `to.equal` node had assigned `test` function instead of `Test`, so testing function was not invoked, and tests using it returned `true` - fixed

- ~~Each `PathNode` in `Luft` has `To` field, I assume it was added to describe how nodes should be combined, e.g.:
`Test.expect().to.equal()` is correct
`Test.expect().to.equal.equal()` is not
The `PathNode:GetNext(word) -> PathNode` at the end of the function was always checking if there is matching `PathNode` for current word in list of all possible nodes instead of `To` nodes of the current node, so as long as there was some Node matching current word the sequence was correct. 
In the PR the check is removed.~~
- ~~Removed nonexisting nodes from PathNodes  `To` and added missing~~

### Colors
- After fixing equal funciton tests started failing, as I see:
  - in the unit tests `HSVtoHSLsimp` was incorrect - fixed
  - in the colors implmentation `HSLtoHSV` was incorrect - fixed

### DIFF
--- a/lua/shared/color.lua
+++ b/lua/shared/color.lua
@@ -476,7 +476,7 @@ function HSLtoHSV(hue, sat, lit)
     if val == 0 then
         return hue, 0, 0
     end
-    return hue, 2 - 2 / (1 + sat), val
+    return hue, 2 - 2  * lit / val, val
 end
 
 
--- a/tests/luft.lua
+++ b/tests/luft.lua
@@ -41,7 +41,7 @@ local Assertion = {}
 ---@field Alias? string
 ---@field NotName? string
 ---@field NotAlias? string
----@field Chain?  fun(a: LuftAssertion)  called after simple chaining e.g. a`.b`.c 
+---@field Chain?  fun(a: LuftAssertion)  called after simple chaining e.g. a`.b`.c
 ---@field ChainCall?  fun(a: LuftAssertion, ...)  called after chained calling e.g. a`.b()`.c
 ---@field Test?  fun(...): boolean        function to test with all arguments (from the starting values, calling arguments, and support values); expects both `FailString` formatters to be defined
 ---@field Parameters number               number of parameters the test function has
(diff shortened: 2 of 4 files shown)


Zjonn:
@Garanas please take a look

Hdt80bro:
Oh this one is probably my fault then--I just rewrote the testing module :)

I'll take a look later tonight

Hdt80bro:
Well done! These some good catches, and helped me find a couple other bugs too (such as the `total_module_` variables only updating at the start of a test). The only comment I have the new behavior (with the knowledge I have about how I wanted it to work) is that you've removed some of the natural language of English I intended to have in, but appear to have only partially implemented (some of it still needs fixing too like the "to not" paths). I'm curious if that's something we even want to have though, e.g. is doing
```lua
luft.expect(x).to.be.equal.to(y)
luft.expect(y).to.be.close.to(y)
luft.expect(y).to.be.an.integer()
```
more desirable than
```lua
luft.expect(x).to.be.equal(y)
luft.expect(y).to.be.close(y)
luft.expect(y).to.be.integer()
```

Garanas:
I'm impartial on either approach, the only thing I can think of is that we shouldn't deviate too much from how other testing suites operate. That would make it more difficult to use.

Zjonn:
Personally I prefer shorter statements, also less complexity in untesed module sounds helpfully.
But for now I will remove changes related to Nodes chaining and it will be up to you @Hdt80bro what to do with them next.

Hdt80bro (approved):

Hdt80bro:
That's alright, I'll merge this and fix the other you revealed in a separate PR

Issue #4136, opened 2022-08-18
Refactor all import paths to lower case


Garanas:
Description
------------

This is a non-trivial issue and it likely requires an external program to process the excessive amount of changes that it proposes. But before we dive into that, lets make sure we understand what the issue is.

The file import system is defined in Lua, it can be found [here](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/system/import.lua). For ease of reading, here's a copy:

```lua
function import(name)
    local modules = __modules

    -- (1) attempt to find the module without lowering the name
    local existing = modules[name]
    if existing then
        return existing
    end

    -- (2) attempt to find the module when lowered the name
    name = StringLower(name)
    existing = modules[name]
    if existing then
        return existing
    end

    -- inform the devs that we're loading this module for the first time
    if informDevOfLoad then
        SPEW("Loading module '", name, "'")
    end
    
    -- (...)
    
end
````

When we import a file multiple times then we return the same instance of that file. We do this via caching. The caching requires a key, the key used is the path in question. A file should be referenced by only one key. 

We've introduced an early exit (1) to check if we can find the module without lowering the key (the name / path). If that doesn't work, we lower it (to guarantee there's one key to a file) and (2) check if we can find the module this time. If we can't, then we're seeing this module for the first time so we'll have to create a new module.

The issue is that lowering a string has a complexity of `O(n)` where `n` is the length of the string. Hence, we'd really like to prevent that. As a few examples of possible inputs:

- 35: `/lua/ui/game/layouts/score_mini.lua`
- 26: `/lua/terranprojectiles.lua`
- 64: `/projectiles/adfshielddisruptor01/adfshielddisruptor01_script.lua`

Note that these have all already been lowered by the code - imagine having to do that each time the Shield Disruptor creates a projectile!

Course of action
------------------

Every _single_ import statement in the game repository should be lower case. There are 4600 instances over 1401 files, therefore it is an impossible task to do by hand. Luckily, it isn't difficult to automate the task either. To solve this issue you'll need to write a smart regular expression [(1)](https://vim.fandom.com/wiki/Changing_case_with_regular_expressions#) / [(2)](https://hatoum.com/blog/2017/12/6/using-regular-expressions-to-change-case) that takes the input and returns the same characters, but lowered.

As useful tool to help you: https://www.regexpal.com/

Other approaches are welcome too, but this one is easiest to re-apply. Note that any solution should be idempotent.

Test plan
---------

[...]

PR #4325, opened 2022-11-03
Cleanup imports

Hdt80bro:
Changes all import to lowercase and be absolute

Close https://github.com/FAForever/fa/issues/4136


Garanas:
_github preview instantly crashes_

What approach did you take to tackle this? I'd like to know, so that we can maybe re-use it for other cases too

Garanas:
I can't approve it because the interface crashes, but it appears to work 👍

Hdt80bro:
I used the regex search & replace built into VSCode. I'm, um, not willing to share the exactly process I went through (I may or may not have learned about `\L` *after* I was done), but a reusable approach for the future that could have saved 30 minutes is:
* To change all paths to absolute, search using `import\s*\(\s*(['"])(?!/)([\w\\./-]*)\1\s*\)`
  * There were about 20 of these not in the `/lua/` directory; I manually changed those
  * The rest can be replaced with `import("/lua/\L$2")`
  * There's probably a better way to do this with better tools
* To change all imports to lowercase, `import\s*\(\s*(['"])([\w\\./-]*)\1\s*\)` -> `import("\L$2")`
* There are a few imports that have concatenation logic in them or are derived from functions that are not caught with the above process. At this point, all strings containing both uppercase letters and a slash (so we only view strings that look paths) can be manually reviewed using `(['"])([a-z_\\./]*[A-Z][\w.]*[\\/][\w\\./]*|[\w\\./]*[\\/][\w\\./]*[A-Z][\w\\./]*)\1` (the `|` in the middle is to catch both when the uppercase letter is before and after the slash)

PR #4332, opened 2022-11-04
Update deprecated mods with the correct version number

Garanas:
With thanks to #4321 we can now deprecate a specific version of a mod. This updates the remainder of the deprecated mods to have a version number, with the one exception that I couldn't find any more 😄

### DIFF
--- a/init.lua
+++ b/init.lua
@@ -63,7 +63,6 @@ integratedMods["nvidia fix"] = true
 
 integratedMods = LowerHashTable(integratedMods)
 
-
 -- take care that the folder name is properly spelled and Capitalized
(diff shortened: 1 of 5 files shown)


Garanas:
@Crotalus  this also adds your economy manager to the deprecation list. The reason is in `commandmode.lua`, where you need to add this if statement: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/game/commandmode.lua#L119

Without it, this line will crash: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/game/commandmode.lua#L151

As we no longer allow booleans to have fields attached to them.

PR #4341, opened 2022-11-06
Polish collapse arrows

Hdt80bro:
~~*sigh*~~
~~I got the collapse arrows for the voting panel and mass fab panel to hide correctly, but now the mass fab panel auto-shows at the beginning of the game, despite trying to replicate the voting panel as best I can.~~

~~Couple other things I threw in here include adding tooltips and key actions to the collapse arrows, and adding a delay to the unit destruction on recall so that the final game state can be viewed when there's only two teams. I'm pretty sure that last bit is currently completely breaks recalling.~~

~~Oh well.~~

Now just refines the collapse arrows, making them only appear when not disabled, and adds tooltips & key actions to them.

As part of testing, the large volume of unnecessary sync data for mass fabs was removed; turned out to not be necessary, but probably good for the environment. The only other non- collapse arrow change is fully removing the partially removed recall camerawork.


lua/sim/Recall.lua
```diff
@@ -23,6 +23,7 @@ function init()
                         IsAlly(focus, index) and
                         brainWith.BrainType ~= "Human" and
                         not ArmyIsCivilian(index)
+                          and false
```
Garanas, line 26: Are these intentional?

Garanas:
There are too many changes for a few hours before the release, this will be part of the next iteration

lua/sim/Recall.lua
Hdt80bro: Yeah, there's several testing changes in here so I can test it without going through the hassle of opening several clients at once and waiting for the gate-in cooldown.

lua/aibrain.lua
```diff
@@ -608,8 +619,17 @@ AIBrain = Class(moho.aibrain_methods) {
...
+                if lastSyncTable.on ~= syncTable.on
+                    or lastSyncTable.off ~= syncTable.off
+                    or lastSyncTable.totalEnergyConsumed ~= syncTable.totalEnergyConsumed
+                    or lastSyncTable.totalEnergyRequired ~= syncTable.totalEnergyRequired
+                    or lastSyncTable.totalMassProduced ~= syncTable.totalMassProduced
+                then
+                    Sync.MassFabs = syncTable
+                    -- swap the data buffers
+                    syncTable, lastSyncTable = lastSyncTable, syncTable
+                end
```
Hdt80bro, line 632: @Garanas is this sync data reduction something you'd prefer to wait on for a simple collapse arrow change?

Garanas: This is fine

Garanas:
Not sure if it is introduced by this PR, but there's two issues:

image

At the moment I have only one fabricator, and it states there are no deactivated fabricators. But I still am 'missing' 100 energy. 

image

When all fabricators are disabled manually then the window no longer shows.

Garanas:
🤔 , sounds like a bug then. But beyond the scope of this pull request

Issue #4344, opened 2022-11-06
Fix TMD spinning around its axis unable to hit missile issue


EvildrewFAF:
**Describe the feature**
Add a minimum range to all TMDs except Aoen. Fixes the issue with TMDs spinning around unable to hit missiles that hit them (UEF most notably). Add MinRadius = 1.5 to the weapon in the blueprint section.

**Expected behavior**
When a missile flies through or hits the TMD, it doesn't spin around on its own axis unable to hit the missile and targets another incoming missile which it can hit.

Garanas:
We discussed this in private - @Tagada14 this is technically a balance change, but I don't think it will have any consequences practically. The TMD can't at missiles that are that close.

Tagada14:
If you don't think it will have any consequences then go ahead

PR #4375, opened 2022-11-08
Fix TMDs spinning around their axis

Garanas:
Closes https://github.com/FAForever/fa/issues/4344


Garanas:
Appears it needs to be slightly more for the spinning to go away properly

PR #4376, opened 2022-11-08
Cache vector in ObjectiveArrow.lua

4z0t:


Hdt80bro:
@Garanas is changing the signature of `BounceThread` to accept the saved offset and vector safe to do so that it doesn't have to go through instance fields?

Hdt80bro:
Actually, we could move `BounceTime` from the instance as well if it's not used anywhere else (which it doesn't look like it is, but, well, mods are weird sometimes).

4z0t:
I can make this vector a local variable as well as offset

lua/objectiveArrow.lua
```diff
@@ -19,25 +19,26 @@ ObjectiveArrow = Class(Entity) {
...
                 yOff = (self.Size * unitScale) / 2.0 + extents.Max.y - spec.AttachTo:GetPosition().y
                 yOff = yOff + 0.5
             else
                 yOff = spec.Size / 2.0 + 0.5
             end
 
-            self:SetParentOffset(Vector(0,yOff,0))
             self.SavedOffset = yOff;
-            ForkThread(self.BounceThread,self)
+            self.SavedVector = Vector(0, yOff, 0)
```
Garanas, line 39: Probably best to just pass the offset to BounceThread, and allocate a vector in the local scope of the thread. That way there is no 'global' scope that can mess things up.

Garanas: @Hdt80bro ^, I think that matches your suggestion

Hdt80bro (approved):

Hdt80bro:
I had been staring at this too long, I needed to put my changes somewhere so I could move on. I managed to exclude all optimizations I had stored that would have reduced readability such as Chebyshev recursion and variable reusing though.

4z0t:
@Hdt80bro I think this is redundant at this point

Hdt80bro:
What else are you thinking covers these changes?

PR #4379, opened 2022-11-09
Weapon Priority Fix

ErikMekkes:
Investigated weapon priority issues after reports of the advanced targeting priorities mod being broken.

Tested and noticed they worked correctly for certain units (t1 tanks) but not others (acu, experimentals). dug in a bit and saw it must be a difference between the limited category parser and the full category parser.

confirmed by switching them around that the issue was in the parser used, guessing this was an autocomplete typo after splitting up the category parsing into full and limited sections in #4338.

### DIFF
--- a/lua/WeaponPriorities.lua
+++ b/lua/WeaponPriorities.lua
@@ -31,7 +31,7 @@ function ParseTableOfCategories(inputString)
             full = { }
             for k, category in categories do
                 if not (category == '') then
-                    local parsed = cachedTablePriorities[category] or ParseEntityCategory(category)
+                    local parsed = cachedTablePriorities[category] or ParseEntityCategoryProperly(category)
                     cachedTablePriorities[category] = parsed
                     if parsed then
                         table.insert(full, parsed)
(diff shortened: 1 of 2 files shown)


ErikMekkes:
comma splitting seems to be fine, my suspicion is that ParseEntityCategoryProperly in lua/sim/CategoryUtils fails for simple single entries. The section used for limited categories hardcodes in a sub operator and that section does not result in empty tables for simple single entries without operators:
`ParseEntityCategoryProperly(string.format('(%s) - COMMAND', category))`

guessing ParseEntityCategoryProperly is missing a branch to return a result if it is a simple single category entry.

ParseEntityCategory does work for these simple entries, but i'm guessing it is not able to handle operators or had some other issue?

ErikMekkes:
Should be fully resolved now with the tokenizer bugfix

ErikMekkes:
should be fully resolved now with tokenizer fix

Hdt80bro (approved):

Garanas:
Great 👍 , I made the same observations but didn't find the time (day before the release) to investigate. This will be part of the hot fix of next weekend

PR #4380, opened 2022-11-09
Tweak size and add in depth

Garanas:


Garanas:
The terrain shader changes should not have made it in 😄

PR #4386, opened 2022-11-11
Keep track of radar structures

Garanas:
@relent0r with regards to our discussion about radar management, I'd take the approach of this pull request.

Roughly speaking it:
- (1) Keeps track of radars at each tech level
- (2) Adds a build condition to check for nearby radars of 1 tech higher

(1) is cheap performance wise as usually a player (and an AI) will not have hundreds of radars. (2) is cheap because once you have an omni somewhere it usually covers the entire map. Meaning the condition will usually only pass for one radar. Except for 20x20 maps, where radars near the edge may also get a green light for an upgrade. Which is what you'd want, as then you carefully expand the radar presence.

Note that (2) should likely be a build condition of some kind, but I'm not familiar enough with those to make that work. I imagine a simple radar manager (similar to a factory manager) may be sufficient though, no build conditions required.

### DIFF
--- a/lua/editor/UnitCountBuildConditions.lua
+++ b/lua/editor/UnitCountBuildConditions.lua
@@ -1449,3 +1449,32 @@ function ForcePathLimit(aiBrain, locationType, unitCategory, pathType, unitCount
     end
     return true
 end
+
+
+-- TODO: where does this fit best?
+
+--- Buildcondition to decide if radars should upgrade based on other radar locations.
+---@param aiBrain AIBrain
+---@param locationType string
+---@param radarTech string
+---@return boolean
+function ShouldUpgradeRadar(aiBrain, locationType, radarTech)
+
+    -- loop over radars that are one tech higher
+    local basePos = aiBrain.BuilderManagers[locationType].Position
+    local otherRadars = aiBrain.Radars[radarTech]
+    for _, other in otherRadars do
+        -- determine if we're too close to higher tech radars
+        local range = other.Blueprint.Intel.RadarRadius
+        if range then
+            local squared = 0.64 * (range * range)
+            local ox, _, oz = other:GetPositionXYZ()
+            local dx = ox - basePos[1]
+            local dz = oz - basePos[3]
+            if dx * dx + dz * dz < squared then
+                return false
+            end
+        end
+    end
+    return true
+end
--- a/lua/editor/UnitCountBuildConditions.lua
+++ b/lua/editor/UnitCountBuildConditions.lua
@@ -1450,9 +1450,6 @@ function ForcePathLimit(aiBrain, locationType, unitCategory, pathType, unitCount
     return true
 end
 
-
--- TODO: where does this fit best?
-
 --- Buildcondition to decide if radars should upgrade based on other radar locations.
 ---@param aiBrain AIBrain
 ---@param locationType string
--- a/engine/Core/Blueprints/UnitBlueprint.lua
+++ b/engine/Core/Blueprints/UnitBlueprint.lua
@@ -26,6 +26,9 @@
 
 --- read more here: https://wiki.faforever.com/en/Blueprints
 ---@class UnitBlueprint: EntityBlueprint
+---@field TechCategory 'TECH1' | 'TECH2' | 'TECH3' | 'EXPERIMENTAL'
+---@field LayerCategory 'LAND' | 'AIR' | 'NAVAL'
+---@field FactionCategory 'UEF' | 'CYBRAN' | 'AEON' | 'SERAPHIM'
 ---@field CategoriesHash table<string, boolean>
 ---@field AI UnitBlueprintAI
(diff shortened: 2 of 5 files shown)


relent0r:
You may be better putting the table updates into the addunit and removeunit functions within the EngineerManager class.

These are triggered using the CreateUnitDestroyedTrigger and CreateUnitCapturedTrigger so will be reliable and only apply to AI. The AI when it builds things will use engineers that are associated with a specific builder mananger, which from a build condition perspective is about all we can use to determine a position we would like to check from.

Each engineer manager maintains a ConsumptionUnits table which things like radars would be added to when they are built.
By table its structured like this.
```
self.ConsumptionUnits = {
            Engineers = { Category = categories.ENGINEER, Units = {}, UnitsList = {}, Count = 0, },
            Fabricators = { Category = categories.MASSFABRICATION * categories.STRUCTURE, Units = {}, UnitsList = {}, Count = 0, },
            Shields = { Category = categories.SHIELD * categories.STRUCTURE, Units = {}, UnitsList = {}, Count = 0, },
            MobileShields = { Category = categories.SHIELD * categories.MOBILE, Units = {}, UnitsList = {}, Count = 0, },
            Intel = { Category = categories.STRUCTURE * (categories.SONAR + categories.RADAR + categories.OMNI), Units = {}, UnitsList = {}, Count = 0, },
            MobileIntel = { Category = categories.MOBILE - categories.ENGINEER - categories.SHIELD, Units = {}, UnitsList = {}, Count = 0, },
        }
```

So from a single source of truth perspective it may be worth looking towards this. I haven't personally used the ConsumptionUnits table in my use of the AI but it is maintained.

For the actual build condition it looks good, instead of passing the unit to it you could pass the LocationType of the base which is running the condition monitor. That way it can look up its position and use that since the unit won't be available to it.

One advantage to the route you've taken with the data table (being attached to the brain) is that its not connected to a builder manager so if we get to the point where we figure out ways of making the AI build radars where it 'should' instead of just at bases it would provide a more flexible location for other logic to use.

Garanas:
I've looked at your suggestion but I'm not sure if I agree.

It appears to me that using the engineer manager would limit the use of this table. The engineer manager is often populated by AI files, such as a platoon function, factory builder manager and when constructing buildings through AI functions. That would limit the use of the table as it seems that only AIs will have this table defined.

If we always guarantee that the table is properly populated then we can also use it in a broader sense. For example, the issue we're solving here is managing our own radars so that they do not overlap too much. But we can use the same data to find radars that we should try to snipe, _without_ doing an expensive call that enumerates all units to find if a radar exists or not. 

But the current approach would be less viable for builder conditions, I'm not sure how those work if you can't pass in the unit (or blueprint) of what you're trying to build.

relent0r:
I agree that having it on the brain offers more flexibility. Just pointing out that we would have multiple unit tables being maintained.

The builder conditions get passed the brain so access the table is fine. 
For the unit itself, we could pass the building type that wants to be built such as T2Radar, T3Radar and use the buildingtemplate to lookup the blueprint.  That would support mods well enough I think.

Using your function there I can put together a condition that will work. Am I right that you are looking for a radar that can overlap the new radars position within the first 60% of its current radius?

Garanas:
> Using your function there I can put together a condition that will work. Am I right that you are looking for a radar that can overlap the new radars position within the first 60% of its current radius?

Yes, that is arbitrarily chosen though. The idea is that you want redundancy to some degree. Feel free to suggest what it should be instead

relent0r:
> > Using your function there I can put together a condition that will work. Am I right that you are looking for a radar that can overlap the new radars position within the first 60% of its current radius?
> 
> Yes, that is arbitrarily chosen though. The idea is that you want redundancy to some degree. Feel free to suggest what it should be instead

Actually I like that number. It means that a T1 radar will still be upgraded if the resulting radar would provide intel in an area that the AI doesn't already have, same with T3 on 20km+ maps.

relent0r:
I've updated this to function with the AI and run a couple of test games. No errors and confirmed that it was giving the correct bool depending on the condition.

The only flaw is that because the TECH levels are static. A T1 radar won't respect T3 when deciding to upgrade to T2. But T2 will respect T3 when looking to upgrade to T3.

I'm not totally sure if thats a bad thing when it comes to radar redundancy. But it completely stops the big issue of the AI upgrading all its T2 radars (the 2k power requirement per radar does bad things) at its expansions when it doesn't need them causing big power drops and many excess pgens being built.

lua/editor/UnitCountBuildConditions.lua
```diff
@@ -1448,4 +1448,33 @@ function ForcePathLimit(aiBrain, locationType, unitCategory, pathType, unitCount
         return false
     end
     return true
+end
+
+
+-- TODO: where does this fit best?
```
Garanas, line 1454: You can remove this comment

PR #4391, opened 2022-11-11
Water shader improvements

BlackYps:
current shader:
old

new shader:
new

The main visual difference is that the overtuned sky reflections for low angles are now more sensible and the weird interpolation near the shores is now gone. This could also help with properly identifying the shore line on some maps.

On a technical level the shader now ignores the values of unitreflectionAmount and skyreflectionAmount that can be set by the map.
I tested it on some maps, but more testing would be good to make sure this doesn't produce unexpected results. I mainly tested on Seton's Clutch, Roanoke Abyss, Saltrock Colony 8 Players, Enfilade and Dual Gap Reloaded.


Garanas:
Looks a lot better to me, I'll test it on some of my own maps.

Garanas (approved):

Garanas:
Experimented a bit, water looks a lot more consistent. I can take the loss of 'proper' lava, as that is not used at the moment anyway.

effects/water2.fx
```diff
@@ -378,8 +378,11 @@ float4 HighFidelityPS( VS_OUTPUT inV,
    	//Schlick approximation for fresnel
     float NDotV = saturate(dot(viewVector, N));
 	float F0 = 0.08;
-    float fresnel = F0 + (1.0 - F0) * pow(1.0 - NDotV, 5.0);         
-    refractedPixels = lerp(refractedPixels, reflectedPixels, fresnel);
+    float fresnel = F0 + (1.0 - F0) * pow(1.0 - NDotV, 5.0);
+
+	if (skyreflectionAmount == 1.5)
```
Garanas, line 383: Please add a comment here, stating why it is there

Garanas: Why do we want this exception here exactly?

Garanas:
It happened again 😞 , I thought I switched the branch. Apologies 😄

Garanas:
Back to just being about water

Garanas:
I tried to force push the fix, but it didn't affect this branch. Not sure if I force pushed your local branch on accident.

effects/water2.fx
BlackYps: If the map author didn't change the skyreflectionAmount, it defaults to 1.5 which is way too bright. With this we can change the default while still preserving the value that the map author deliberately set. I tested it on various maps and there are some maps that set it to even less than one and it does look nicer to preserve that most of the time.

PR #4326, opened 2022-11-03
Create tests for NavDatastructures.lua

Zjonn:
PR covers first task from #4295

It contains black box unit tests for `sim/NavDatastructures.lua`, covered functions:
- `IsEmpty`
- `Clear`
- `Insert`
- `ExtractMin`

Issue #4314, opened 2022-10-30
Nav mesh - re-implement the quad tree compression methods


Garanas:
Description
-------------

The compression is a key component of the navigational mesh. It significantly reduces the amount of memory and computations during run-time. The initial implementation works, but performance-wise it can be improved. We're talking about the `Compress` method of the `CompressedLabelTree` class. At the moment of writing, you can find it on line 281:
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/NavGenerator.lua#L353

The issue of the initial implementation is that it works top-down instead of bottom-up.  This causes a lot of data to be re-read. This is best explained with a drawing:

![delete-me-thrice](https://user-images.githubusercontent.com/15778155/198872024-f8fa7994-9b8d-4f97-9eac-600f203f0cab.png)

Say we have a 4x4 square. The bottom right corner is invalid, marked as red. The rest of the square is valid, marked as blue. The current implementation reads the entire square, only to find out that the last read is invalid. The entries that are read are marked as yellow. The algorithm splits the area in four equal squares, and then starts reading each square separately again. The entries that are read for each square are marked as yellow.

On a map we're not talking about 4x4 squares, but about 32x32, 64x64, 128x128 or even 256x256 squares. That means we can potentially do a lot of re-reading when a square is not entirely valid.

Course of action
------------------

Re-implement the compress method using a bottoms-up approach. This is a recursive implementation:
- (1) We first compute all children, up to the base case: the compression threshold. These represent the initial leafs.
- (2) We check each child to determine if they have the same label. If they all have a label assigned and they are all the same (either `-1` for invalid, or `0` for valid) then we cull the children and assign that label to the current node. That node then becomes a new leaf. 

With this approach we guarantee that we only read each cell of the square once. We do introduce some additional class instantiations that we then potentially immediately trash - in the future we'll be looking at pooling those to prevent thrashing the garbage collector too much.

The description is a bit rough - if you have questions, you can ask so in the comments. 

Test plan
----------

The output should be exactly the same, but the generator should be faster. The changes should be limited to the `Compress` function of the `CompressedLabelTree` class.

Note that the pooling is **not** part of this issue, it will be a separate issue and needs to be solved using a separate pull request.

Learning goals
--------------

Learn to work with algorithms and data structures.

Zjonn:
Hey @Garanas, I will take care of this issue, you can assign it to me

Garanas:
[...]

PR #4399, opened 2022-11-13
Improve performance of the Heap data structure

### LINKED ISSUE #4295: Nav mesh - improve testability and performance of the Heap data structure

Zjonn:
Closes #4295 

## Brief
Main goad of the #4295  was to improve performance of the Heap datastructure, how it went?
Well, I got about ~~11% of performance improvement~~ 60% after @Garanas suggestions, tested using following scenario:
```lua
...
---@param heap NavPathToHeap
local function AddAndPop(heap)
    for i = 1, 5000, 1 do
        for _ = 1, 4, 1 do
            heap:Insert(AsLeaf(math.random()))
        end
        for _ = 1, 1, 1 do
            heap:ExtractMin()
        end
    end
    heap:Clear()
end
...
```

## Changes
- As suggested in the issue `ExpectedCost` was replaced with `TotalCost`
- `Heapify` and `Rootify` functions in NavDatastructures have been replaced by their iterative versions.
- Adjusted UTs to use `TotalCost` instead of `ExpectedCost`

### DIFF
--- a/lua/sim/NavDatastructures.lua
+++ b/lua/sim/NavDatastructures.lua
@@ -69,58 +69,59 @@ NavPathToHeap = ClassSimple {
 
     ---@param self NavPathToHeap
     Heapify = function(self)
+        local heap = self.Heap
+        local heap_size = self.HeapSize
+
         local index = 1
         -- find the left / right child
-        local left = self:ToLeftChild(index)
-        local right = self:ToRightChild(index)
+        local left = 2 * index
+        local right = 2 * index + 1
         -- if there is no left child it means we restored heap properties
-        while self.Heap[left] do
+        while left <= heap_size do
             local min = left
 
             -- if there is a right child, compare its value with the left one
             -- if right is smaller, then assign min = right. Else, keep min on left.
-            if self.Heap[right] and (self.Heap[right].TotalCosts < self.Heap[left].TotalCosts) then
+            if heap[right] and (heap[right].TotalCosts < heap[left].TotalCosts) then
                 min = right
             end
-            
-            -- if min has higher value than the index it means we restored heap properties 
+
+            -- if min has higher value than the index it means we restored heap properties
             -- and can break the loop
-            if self.Heap[min].TotalCosts > self.Heap[index].TotalCosts then
+            if heap[min].TotalCosts > heap[index].TotalCosts then
                 return
             end
             -- otherwise, swap the two values.
-            self:Swap(index, min)
+            local tmp = heap[min]
+            heap[min] = heap[index]
+            heap[index] = tmp
             -- and update index, left and right indexes.
             index = min
-            left = self:ToLeftChild(index)
-            right = self:ToRightChild(index)
+            left = 2 * index
+            right = 2 * index + 1
         end
     end,
 
     ---@param self NavPathToHeap
     Rootify = function(self)
+        local heap = self.Heap
         local index = self.HeapSize
-        local parent = self:ToParent(index)
+        -- index / 2
+        local parent = index >> 1
         while parent >= 1 do
             -- if parent value is smaller than index value it means we restored correct order of the elements
-            if self.Heap[parent].TotalCosts < self.Heap[index].TotalCosts then
+            if heap[parent].TotalCosts < heap[index].TotalCosts then
                 return
             end
             -- otherwise, swap the values
-            self:Swap(parent, index)
-            -- and update index and parend indexes
+            local tmp = heap[parent]
+            heap[parent] = heap[index]
+            heap[index] = tmp
+            -- and update index and parent indexes
             index = parent
-            parent = self:ToParent(index)
+            -- parent / 2
+            parent = parent >> 1
         end
-    end,
-
-    ---@param self NavPathToHeap
-    ---@param a number
-    ---@param b number
-    Swap = function(self, a, b)
-        local l = self.Heap[a]
-        self.Heap[a] = self.Heap[b]
-        self.Heap[b] = l
     end,
 
     ---@param self NavPathToHeap
@@ -133,29 +134,5 @@ NavPathToHeap = ClassSimple {
         else
             WARN("given object to Heap was nil!")
         end
-    end,
-
-    ---@param self NavPathToHeap
-    ---@param index number
-    ---@return number
-    ToParent = function(self, index)
-        -- index / 2
-        return index >> 1
-    end,
-
-    ---@param self NavPathToHeap
-    ---@param index number
-    ---@return number
-    ToRightChild = function(self, index)
-        -- 2 * index + 1
-        return 2 * index + 1
-    end,
-
-    ---@param self NavPathToHeap
-    ---@param index number
-    ---@return number
-    ToLeftChild = function(self, index)
-        -- 2 * index
-        return 2 * index
     end,
 }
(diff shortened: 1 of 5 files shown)


Askaholic:
Btw this would be an ideal candidate for writing some unit tests as there are no dependencies so importing the module should be really easy

Zjonn:
Already did it in #4326

Askaholic:
Awesome! :+1:

lua/sim/NavDatastructures.lua
```diff
@@ -63,59 +62,56 @@ NavPathToHeap = ClassSimple {
...
         return value
     end,
 
     ---@param self NavPathToHeap
-    ---@param index number
-    Heapify = function(self, index)
-        -- find the left / right child and the parent.
-        local parent = index
+    Heapify = function(self)
+        local index = 1
```
Garanas, line 72: What if you also store `self.Heap` in a local, e.g. `local heap = self.Heap` ? It prevents a lot of repetitive table operations, same for the `rootify` function

lua/sim/NavDatastructures.lua
```diff
@@ -63,59 +62,56 @@ NavPathToHeap = ClassSimple {
...
+        -- find the left / right child
         local left = self:ToLeftChild(index)
         local right = self:ToRightChild(index)
+        -- if there is no left child it means we restored heap properties
+        while self.Heap[left] do
```
Garanas, line 77: I don't think we need to check for whether the element exists, but instead we can simply check if `left <= self.HeapSize` or even better - store the heap size in a local and compare with that. If the heap is correct, then at no point should `self.Heap[left]` have a value when it is outside of the heap (e.g., `self.HeapSize`)

lua/sim/NavDatastructures.lua
```diff
@@ -63,59 +62,56 @@ NavPathToHeap = ClassSimple {
...
 
     ---@param self NavPathToHeap
-    ---@param index number
-    Heapify = function(self, index)
-        -- find the left / right child and the parent.
-        local parent = index
+    Heapify = function(self)
+        local index = 1
+        -- find the left / right child
         local left = self:ToLeftChild(index)
```
Garanas, line 74: What if you replace these with their definitions, e.g. `left = 2 * index` and `right = 2 * index + 1`?

Garanas: In Lua a function call is relatively expensive, especially when it involves `self`. See also: 
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/benchmarking/benchmarks/function-origin.lua

Where `IsDead` refers to: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L5324

Garanas (commented):
A few suggestions to find more performance. After that we'll merge this in, great work 👍

lua/sim/NavDatastructures.lua
Zjonn: Thanks for the idea. This changed the 11% improvement to almost ~60%

lua/sim/NavDatastructures.lua
Zjonn: Done

lua/sim/NavDatastructures.lua
Zjonn: Done

lua/sim/NavDatastructures.lua
Garanas: This change on its own, or the combination of the changes?

Zjonn: Mainly this change, the other two had a much smaller impact, about 4%

Garanas (approved):

Garanas:
@Zjonn great work 👍 

I'm hoping I can interest you to pick up this issue too: https://github.com/FAForever/fa/issues/4314

Similar to the issue you've been working on here, it involves an algorithm that can do with a touch or two from a performance perspective.

PR #4401, opened 2022-11-13
ForkThreadable class

4z0t:
Uses common class for self-forkthreading.

### DIFF
--- a/lua/AI/AttackManager.lua
+++ b/lua/AI/AttackManager.lua
@@ -1,5 +1,5 @@
 local Utilities = import("/lua/utilities.lua")
-
+local TrashForkThreadable = import("/lua/shared/ForkThreadable.lua").TrashForkThreadable
 -- ATTACK MANAGER SPEC
 --{
 --    AttackCheckInterval = interval,
@@ -26,8 +26,8 @@ local Utilities = import("/lua/utilities.lua")
 --     AMPlatoons = { AMPlatoonName, AMPlatoonName, etc },
 -- },
 
----@class AttackManager
-AttackManager = ClassSimple {
+---@class AttackManager : TrashForkThreadable
+AttackManager = Class(TrashForkThreadable) {
     brain = nil,
     NeedSort = false,
     PlatoonCount = { DefaultGroupAir = 0, DefaultGroupLand = 0, DefaultGroupSea = 0, },
@@ -36,16 +36,6 @@ AttackManager = ClassSimple {
         self.Trash = TrashBag()
         self.brain = brain
         self:Initialize(table)
-    end,
-
-    ForkThread = function(self, fn, ...)
-        if fn then
-            local thread = ForkThread(fn, self, unpack(arg))
-            self.Trash:Add(thread)
-            return thread
-        else
-            return nil
-        end
     end,
 
     Initialize = function(self, attackDataTable)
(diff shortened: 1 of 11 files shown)


lua/AI/AttackManager.lua
```diff
@@ -78,7 +68,8 @@ AttackManager = ClassSimple {
         local atckCond = {}
         if not AttackConds then
             AttackConds = {
-                { '/lua/editor/platooncountbuildconditions.lua', 'NumGreaterOrEqualAMPlatoons', {'default_brain', 'DefaultGroupAir', 3} },
+                { '/lua/editor/platooncountbuildconditions.lua', 'NumGreaterOrEqualAMPlatoons',
```
Garanas, line 71: Did your formatter process this file? 😄

4z0t: Yeah, accidentally

Hdt80bro: Ohhhh, is *that* what's been adding all of those noisy spaces?

Garanas: Could undo the formatting? It adds some single keywords: https://github.com/FAForever/fa/pull/4401/files#diff-d99bfe6b1eb282cf605c30f4b3b3fcdbc988ec36423d419fbc3bbc89908dc663R267 , which looks like this in the editor:

image

That doesn't make it more readable to me

Garanas:
I understand the class, but the pattern is _highly inefficient_, it involves a lot of table creations and unpacking.

4z0t:
It is used in key classes of the game, wdym?

4z0t:
Well, I think we can only specify this in comments that class is very impactful on performance and not recommended to use.

Garanas (approved):

Hdt80bro (changes requested):
I'm liking the reduction in repeated code! That's always good. Several concerns however:
* `ForkThreadable` never gets used, and attempting to do so implies an intention to produce code with memory leaks--it should likely be removed.
* Right now, the `TrashForkThreadable` class only serves to unify the fork method; it supplies other methods to create the trash bag and clean it up, but they never get called (and shouldn't, as managing the trashbag is not the class's purpose), so it lacks proper encapsulation. Of course, we also don't currently have an inheritable trashing class, so leaving the `TrashForkThreadable` class as an incomplete type would be fine--it just shouldn't pretend to be complete.
* The naming is unnatural--"ForkThreadable" modifies the wrong verb, and prepending "Trash" to that doesn't add the adjective to the right word. I recommend something like "ThreadForkable" and "TrashThreadForkable" (if you do end up removing the former class, the latter would be renamed to it).
* As long as behavioral inheritance is the goal, the trashbag version's annotation class should inherit the other's (though not the class itself, if you keep both).
* And of course, you'll need to undo your formatter :)

Jip is right about the inefficiency of the pattern and the implications supplying these classes would have on making it easier for people to write inefficient code. However, an extra table creation and 2 function calls are usually negligible to what's happening inside the thread, and definitely one of the lowest priorities compared to the rest of the poorly written code. They shouldn't be optimized right now. There's also the thought that if we don't provide a way, someone else might try to do it themself in a way that ends up being worse :)

Over all, I think this is a good change that could help with maintenance in the future.

4z0t:
I made it `ForkThreadable` because it is simpler to understand that it does something with function `ForkThread`.

Garanas:
> ForkThreadable never gets used, and attempting to do so implies an intention to produce code with memory leaks--it should likely be removed.

I think threads self-clean up once they have finished running the function, but I have no data to back that up. The only reason threads are added to the trashbag is to guarantee that threads are destroyed that heavily rely on `self` (and will crash, the moment `self` has no valid c object behind it)

4z0t:
I wanted to add Trash allocation and destruction in ctor (__init) and dtor (Destroy) but I'm not sure it is necessary.

Garanas:
I'm closing this pull request - I'm in favor of the pattern, but it needs to be applied entirely. At the moment the init function of the new abstraction is not always called, for example.

PR #4412, opened 2022-11-16
add logic for army pool detection and disband blocking

### LINKED ISSUE #4411: AI ArmyPool being disbanded during games.
**Describe the bug**
The AI maintains a special platoon called the 'ArmyPool'. This is where all units go before they are pulled into platoons.
While doing some other work I realized that something was clearing the commands for all structures (extractors, factories etc) randomly. This happens during the PlatoonDisband function.
After troubleshooting I've found that something is disbanding the ArmyPool during games. I've done full logging and traced it to at least two places.
1. The ACU thread, https://github.com/FAForever/fa/blob/80e1544d27889d4a93114c9229122d5a5c4ac233/lua/AI/AIBehaviors.lua#L361

2. The ProcessBuildCommand function.
https://github.com/FAForever/fa/blob/80e1544d27889d4a93114c9229122d5a5c4ac233/lua/platoon.lua#L3618

The ACU thread issue is relatively simple to fix. The ProcessBuildCommand one is a little harder as there are multiple things that call it including engineer callbacks. Which is where I think this problem may be coming from.

**To Reproduce**
Reproducing is easy as it will happen during AI games. It may not happen in EVERY game but I've noticed that its pretty consistent to happen at least once, and often multiple times over a game that spans 15+ minutes. But it required extensive logging to find the occurrences as the symptoms are not obvious.

**Expected behavior**
The ArmyPool should never be disbanded as this interrupts the AI's ability to create new platoons and clears the commands of any units in that ArmyPool.

**Additional context**
This issue has been relatively invisible as not many units have functionality that is interrupted. The most well known (by well know I mean noticed by 2 or 3 people) is that factories will randomly cancel building orders. Its not noticeable in the T1 stage because of the build speeds but when a T3 unit or support commander build is cancelled half way through it becomes more noticeable.

I'm currently looking at ways to resolve this issue. The most obvious solution is to make sure the platoon is checked within the PlatoonDisband function to make sure that it is not the army pool before it continues.

relent0r:
Closes #4411 
This PR is designed for the next major patch release, should not be included in a hotfix.

## Description of changes
This PR changes adds an ArmyPool detection to the PlatoonDisband function. It is being put in place to stop units from accidentally disbanding the ArmyPool.
This is a relatively brute force method but any units trying to disband the army pool should not be treated as playing nice. I've added a warning for the AI developer so they will know when one of their units are doing this.

Note : because this sets the bool in the InitializeSkirmishSystems function then this has no impact to Sorian.

## Test setup for the changes
Start a skirmish with the default AI. If a units or platoon trys to disband the ArmyPool it will be blocked and a warning will be logged.
From my own testing I saw some interesting results. On a map like Point of Reach V4 I got 800 instances of this warning, where as another 10km map with no water I got none for the test.

### DIFF
--- a/lua/platoon.lua
+++ b/lua/platoon.lua
@@ -247,6 +247,10 @@ Platoon = Class(moho.platoon_methods) {
 
     ---@param self Platoon
     PlatoonDisband = function(self)
+        if self.ArmyPool then
+            WARN('AI WARNING: Platoon trying to disband ArmyPool')
+            return
+        end
         local aiBrain = self:GetBrain()
         if self.BuilderHandle then
             self.BuilderHandle:RemoveHandle(self)
--- a/lua/platoon.lua
+++ b/lua/platoon.lua
@@ -249,6 +249,7 @@ Platoon = Class(moho.platoon_methods) {
     PlatoonDisband = function(self)
         if self.ArmyPool then
             WARN('AI WARNING: Platoon trying to disband ArmyPool')
+            --LOG(reprsl(debug.traceback()))
             return
         end
         local aiBrain = self:GetBrain()
--- a/lua/aibrain.lua
+++ b/lua/aibrain.lua
@@ -1511,6 +1511,7 @@ AIBrain = Class(moho.aibrain_methods) {
         -- TURNING OFF AI POOL PLATOON, I MAY JUST REMOVE THAT PLATOON FUNCTIONALITY LATER
         local poolPlatoon = self:GetPlatoonUniquelyNamed('ArmyPool')
         if poolPlatoon then
+            poolPlatoon.ArmyPool = true
             poolPlatoon:TurnOffPoolAI()
         end
 


Garanas (approved):

lua/platoon.lua
```diff
@@ -247,6 +247,10 @@ Platoon = Class(moho.platoon_methods) {
 
     ---@param self Platoon
     PlatoonDisband = function(self)
+        if self.ArmyPool then
+            WARN('AI WARNING: Platoon trying to disband ArmyPool')
+            return
```
Garanas, line 252: You can add a traceback to have an idea _where_ it got called from with: `reprsl(debug.traceback())`

relent0r: ah I haven't used that before. What sort of information would I expect it to give me?
From testing I couldn't get much information as to where the platoon disband came from without adding logs into the corresponding function.

Garanas: The complete stacktrace. Best to try it out and see what happens, just add that snippet to pieces of code. Note that it is expensive and should strictly be used to make (tracking of) errors more obvious.

relent0r: It dumps this.
```
info: stack traceback:
info:         g:\code\fa\lua\platoon.lua(253): in function <g:\code\fa\lua\platoon.lua:250>
info:         (tail call): ?
info:         g:\code\fa\lua\platoon.lua(3852): in function <g:\code\fa\lua\platoon.lua:3751>
```

Which in this scenario tells me which function it happened in and which function called it. So it does help identify where it came from. The why is a different story :) But will very useful. I'll leave this commented out given its expense so that AI Devs can re-enable when troubleshooting as it will identify correctly where the problem came from.

I'm pretty confident callbacks and queued commands are responsible for this.

Garanas: It appears a recursive call is responsible: that is what a tail call is. Glad you're aware of it now. Note that you can keep it enabled in cases like these, where something happens that shouldn't be happening - it will make it easier to investigate game logs.

PR #4413, opened 2022-11-16
AI Tuning. Economic mass focus. Minimum essential units.

relent0r:
This PR is designed for the next major patch release, should not be included in a hotfix.

**Description of changes**
This PR changes tunes some of the default AI's builders. It aims at decreasing the likelyhood that the AI will put itself into a mass stall and be unable to get out of it for prolonged periods.

A minimum mass storage requirement is placed on most non essential structure based builders and shield upgrade formers.

Adds minimum builders for the first air factory, land factory and hydro(within 160 units) to make sure these are made as early as possible.

This is a test by observation change. 
Comparatively speaking.
The AI will stall its mass less than it used to and when it does it shouldn't continuously make the situation worse. 
The AI should be more consistent about making sure it has at least 1 factory of air and land within the first 5 mins. 
The AI will be consistent with rebuilding T1 Air and Land factories (this is in preparation for support factory integration) if it has no non support factories.
The AI should produce engineers when it is power stalling in the T1 phase up to the engineer cap of the base.
The AI will dedicate two engineers to mass points within a radius of 40 at game start. 

**These changes will make some other AI weaknesses more apparent. Issues noticed in testing.**
In testing it was noted that the AI is more likely to stall its power in the T1 phase due to the increased mass in storage and the fact that there is only one T1 pgen builder and no assist builder to scale quicker in the T1 phase.
The naval rush AI base template seems very broken and is often selected when the AI is set to adaptive on an island map. 
The AI can get into a run away effect with its mass extractor upgrades due to the builders conditions limiting spend on unit production.

The next tuning pr will tackle power building and scaling in the early game as a primary focus.


lua/AI/AIBuilders/AIEconomicBuilders.lua
```diff
@@ -1250,6 +1261,28 @@ BuilderGroup {
     -- =====================================
     --     T1 Engineer Resource Builders
     -- =====================================
+    Builder {
+        BuilderName = 'T1 Hydrocarbon Engineer Single',
+        PlatoonTemplate = 'EngineerBuilder',
+        Priority = 1000,
+        BuilderConditions = {
+                --DUNCAN - commented out
```
Garanas, line 1269: Better to remove commented code, we have git history to see the history of a file

lua/AI/AIBuilders/AIEconomicBuilders.lua
```diff
@@ -1994,11 +2028,31 @@ BuilderGroup {
 BuilderGroup {
     BuilderGroupName = 'EngineerMassBuildersHighPri',
     BuildersType = 'EngineerBuilder',
+    Builder {
+        BuilderName = 'T1ResourceEngineer 40',
+        PlatoonTemplate = 'EngineerBuilder',
+        Priority = 1005,
+        InstanceCount = 2,
+        BuilderConditions = {
+                --{ UCBC, 'EngineerLessAtLocation', { 'LocationType', 3, categories.ENGINEER * (categories.TECH2 + categories.TECH3) }},
```
Garanas, line 2037: Same here.

lua/AI/AIBuilders/AIExperimentalBuilders.lua
```diff
@@ -217,6 +220,7 @@ BuilderGroup {
             { UCBC, 'HaveLessThanUnitsInCategoryBeingBuilt', { 1, categories.EXPERIMENTAL}},
             { UCBC, 'HaveLessThanUnitsInCategoryBeingBuilt', { 1, categories.NUKE * categories.STRUCTURE}},
             { UCBC, 'HaveGreaterThanUnitsWithCategory', { 0, categories.ENERGYPRODUCTION * categories.TECH3}},
+            { EBC, 'GreaterThanMassStorageCurrent', { 150 }},
```
Garanas, line 223: If I am correct the conditions are evaluated one-by-one, where the builder is rejected the moment it finds an invalid condition. From a performance perspective it is better to fail on simple checks quick, before running more expensive checks.

Without asking you to re-order all the conditions, could you add in future conditions with respect to how expensive they are? The average economy condition is (very) cheap in comparison to unit conditions: they should be at the top of the list.

Garanas (changes requested):

lua/AI/AIBuilders/AIExperimentalBuilders.lua
relent0r: Yes your right, line 223 is cheaper than any of the conditions on 220,221,222. I did reorder a bunch of the others but I also missed some. I'll try and reorder what I've got here.
The general rule is to have a happy medium between making sure the most likely to fail condition will come first, but also trying to have the cheapest first.

In this case GreaterThanMassStorageCurrent is one function call vs many for the HaveLessThanUnitsInCategoryBeingBuilt condition.

Garanas: I think the order should depend on the complexity of the condition. As an example:

- `{ EBC, 'GreaterThanMassStorageCurrent', { 150 }},` has a complexity of O(1), because it doesn't rely on any list or other recursive datastructure
- `{ UCBC, 'HaveLessThanUnitsInCategoryBeingBuilt', { 1, categories.NUKE * categories.STRUCTURE}},` has a complexity of O(n) because internally it relies on `aiBrain:GetListOfUnits(categories.CONSTRUCTION, false)`, and then iterates over the results. A list (or table) is a recursive structure, meaning it can grow and shrink: these conditions get more and more expensive as the game progresses, as `n` relies on the number of units.

All unit conditions should be at the bottom (checked last), even if they would usually fail fast too (it still has to iterate over all the current units to determine that no unit matches a category!), while checking one value is always cheap.

lua/AI/AIBuilders/AIEconomicBuilders.lua
relent0r: have removed comments where appropriate.

lua/AI/AIBuilders/AIExperimentalBuilders.lua
relent0r: i've reordered all conditions in the ones that I've changed. Thought it was better to just do it right the first time.

relent0r:
Just as a quick note. The MIBC condition results are cached for 7 seconds compared to the EBC or UCBC conditions.

Garanas (approved):

PR #4420, opened 2022-11-18
Refactor shadow orders code

d-colwell:
Split out Shadow Orders code to its own module to allow shadow orders to be processed by multiple processors without needing to add more order registries to the command loop


Garanas:
A change like this would break compatibility with UI mods that rely on the functions and tables to be present in `spreadattack.lua`.

Garanas:
Closing this pull request as there has been no activity the past two weeks.

PR #4421, opened 2022-11-18
Turn all imports of build conditions into lower case

Garanas:
The import function is a lot more efficient when the entire string is lower case. It allows us to skip this string operation:

- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/system/import.lua#L51

These build conditions are constantly being imported throughout the game. The faster this is, the better 😄 

@relent0r could you review this PR, cross-reference some of the changes and see if they are correct? I used a regular expression, so if one is correct then they should all be correct.


Garanas:
And fixed I think 😄

relent0r:
Looking at the imports they seem fine. The local vars are all unchanged so no impact to anything.
I didn't know about the lower case thing actually, Just assumed that GPG was full of windows devs :)

I'll be honest I don't fully understand the why of what your doing in the BrainConditionsMonitor. You take both the upper and lower case versions of the path names and merge them into a table? I assume its related to the cFilename param that is passed to that GetConditionKey function.

Anyway it looks like you've got something wrong in it. I logged everything going through this if statement and nothing is coming through as an instant condition.

```
if instantBuildConditions[cFilename] then
            newCondition = InstantImportCondition()
        else
            newCondition = ImportCondition()
        end
```

relent0r:
On a side note. I never really paid attention to it before. But the Skirmish AI is loading condition files that are never used except in the campaign.

Garanas:
`table.cat` didn't quite work as expected, it returned:

```
INFO:  - 1: true
INFO:  - 2: true
INFO:  - 3: true
INFO:  - 4: true
INFO:  - 5: true
INFO:  - 6: true
INFO:  - 7: true
INFO:  - 8: true
```

While it should be:

```
INFO:  - /lua/editor/EconomyBuildConditions.lua: true
INFO:  - /lua/editor/InstantBuildConditions.lua: true
INFO:  - /lua/editor/SorianInstantBuildConditions.lua: true
INFO:  - /lua/editor/UnitCountBuildConditions.lua: true
INFO:  - /lua/editor/economybuildconditions.lua: true
INFO:  - /lua/editor/instantbuildconditions.lua: true
INFO:  - /lua/editor/sorianinstantbuildconditions.lua: true
INFO:  - /lua/editor/unitcountbuildconditions.lua: true
```

Technically the entire if statement is bogus, as a mod may accidentally write it as `/lua/Editor/EconomyBuildConditions.lua` (note the `E` of editor instead of `e`), which would be a simple typo to make for say a mod.

Garanas:
> On a side note. I never really paid attention to it before. But the Skirmish AI is loading condition files that are never used except in the campaign.

What files are these?

Garanas:
@relent0r it appears to work now 👍

relent0r:
Yup looks good.

As for the unused condition files.

local PCBC = '/lua/editor/PlatoonCountBuildConditions.lua'
local OAUBC = '/lua/editor/OtherArmyUnitCountBuildConditions.lua'


There are also a couple that I don't understand.
These are loaded on every builder file but I don't see it used. I don't even know why you would want it in them.
local PlatoonFile = '/lua/platoon.lua'
local AIAddBuilderTable = import("/lua/ai/aiaddbuildertable.lua")

My only thought is potentially mods?

relent0r (approved):

Garanas:
Yes, we learned the hard way that mods not just rely on logic, but sometimes also on imports 😄

PR #4422, opened 2022-11-18
Support Layer Tiers & Add Backwards Compatibility for FactoryBuilderManager.lua

Azraeel:
Adds Layered Tiers to the Factory Builder Manager allowing factories to have a much faster response as their list of builders is decreased to searching for example

"Land & Tech 3" instead of "Land".


Azraeel:
@Garanas @relent0r 

(Also i have no idea why github decided to delete the entire file and replace it??

Garanas:
Sounds like a line ending problem

PR #4425, opened 2022-11-18
Support Layer Tiers & Add Backwards Compaibility For FactoryBuilderManager.lua

Azraeel:


Garanas:
I do not have the understanding to review this - could @Uveso or @relent0r help out?

relent0r:
Did you test this @Azraeel before putting in the PR?

The way I'm reading it. Is that there's an optional brain flag called `HasLayerTBuilders`.
Once this is set to true then the FactoryManager expects all builders to define which tier they are to be built on and the manager won't accept anything except either that or 'All'.

When I tried to test this the manager would not build anything other than what was defined first.
e.g If I set this. BuilderType = 'LandT1', 'LandT2',
I got the builder from T1 land factories but not T2.

Do you think you could do a bit of a write up in the PR description explaining how we should be testing this with some examples as either I'm doing it wrong or its not working correctly.

Is the benefit purely a performance thing or are you also using this to enforce production limits without builder conditions? Do you have it setup in Swarm at all so I could see it fully configured?

Azraeel:
@relent0r 

I tested it although it only works with 1 of the strings for example builderType = 'LandT1', 
Thats how its intended to work you can still use all tho

I generally dont have the time to really do in-depth testing like i used too but it doesnt work with Brackets because its not parsing multiple strings within the brackets only 1 string is allowed currently. Although I'd like for you to be able to do LandT1, LandT2, & LandT3 for example.

Garanas:
@Azraeel when a change is made it is important to make sure you've tested it and describe how you tested it. Especially for an abstract change like this. Otherwise, we are all left guessing what your intent was, and what you have tested so far.

Dragun123:
As a note I’ve gotten a working version of this already in SCTA. And can easily modify it to be usable version for FA Vanilla factions.

relent0r:
From how I've read its current functionality. The benefit it provides currently is to make sure T3 builders don't get looked at by t1/t2 factories. Without it supporting multiple tiers only a certain number of builders could take advantage of it yet unless you wanted them to be restricted to only that tier of units.

If you set the others to any then you'd lose the current land/air/sea types.

I'd suggest a bit more work on this one so it will support multiple types before merge. Doesn't sound like Azraeel has much time though, not sure if @Dragun123 is interested in working on this code or if we'd rather look at the avenue he took for SCTA.

In order for the default AI to take advantage of it someone would also need to redo all the builders to specify tiers.

I haven't asked Sprouto about the real world gains vs effort from this in regards to the factory manager as this was originally done in LOUD as a way to allow factories to minimize the numbers of builders the factories are checking.

PR #4430, opened 2022-11-19
Feature/ai structure manager

relent0r:
**Description of changes**
This PR implements the Structure Mananger. Along with a first implementation of its use which is extractor upgrades.

**Goal of the Structure Manager**
The structure managers primary goal is to allow the AI to perform structure upgrades without using the Platoon Former Manager. Currently all upgrades leverage the platoon former. This places stress on the manager who's primary design goal is to form platoons of units for combat etc. By moving upgraders away from the platoon former the AI will have more responsive platoon formers and a shorter condition loop.

The other main advantage of the Structure Mananger is that its a centralized logic thread so there is only a single source of triggers. Historically all the builder conditions are checked individually which can/will result in the AI being unable to control which upgrades take place without complex builder conditions and even then the nature of conditions could mean multiples triggers could happen when they are unwanted.

I've tried to make the Structure Manager as self contained as possible. So most data is held within it where possible. This is to try and avoid future developers making changes to other parts of the AI without realising they may be impacting this class.

**Current framework**
The manager is desired to be modular in its approach so they other types of structure upgrades can be added easily. Currently it doesn't really do that. I would like assistance on how to better make this a part of it without losing the advantages of logic potential for each structure type.
It currently will spin up a single structure upgrade thread which runs on a 3 second loop if no upgrades are available and 6 seconds if they are.
Compared to the builder manager equivalents this is a much slower thread loop but consistent. The default condition monitors loop time can range from 0.15 seconds to 40 seconds depending on the strain on the condition monitor.

**The Extractor thread**
I've put an excessive amount of comments in the extractor thread so the readers both technical and non technical will understand what is going on.

For Players wanting to review logic. You want to be looking at 
ExtractorUpgradeThread -- Responsible for upgrade triggers.
SelectClosestExtractor -- Responsible for selecting a specific extractor to upgrade once the trigger has happened.
UpgradeExtractor -- Responsible for the upgrade thread that runs on each extractor, will pause extractor upgrades if economy is hurting up to a certain time limit.

**Help Wanted**
I'm open to any and all technical advice on improving the framework and structure of this class as I don't have much experience with lua classes and general best practice programming.
I'm open to advice for improving logic triggers and other game play based decision making. I think some of the advanced FAF players could help alot in this area.

Notes about trigger logic and why I made some of the decisions.
It must be taken into account that this logic needs to be able to function across a range of maps and gameplay styles.
From maps where the player only gets 4 or 5 extractors to maps where they get many.
From gameplay styles like ladder to team play to turtle styles.

There may well be other metrics (like mass point share, team game spawn positions etc) that could help influence the logic and I'm open to advice on this as well . The goal is to try and have the AI more intelligently manage its structure upgrades without getting stuck down a certain style of play or map type. Keep in mind that people still play the game on potatoes so performance is a consideration.

This branch contains all changes required to make the Structure Manager function for all default AI so you just need to start a skirmish game with a default AI and it will be operational. The existing builders for extractor are commented out but will be removed before this is merged.
Note that the extractor upgrades will try to work with the AI's current economic situation. But the if the AI is making bad decisions (which is does) then it will struggle to get upgrades done. 
I've done tuning for the AI's early game to assist the structure manager in this way. But the T3 phase can cause issues. Mainly due to the wild swings that different construction projects create on the economy and the mass consumption a engineer will have on one project vs another.

**Future functionality**
I'd like the Structure Manager to more easily be able to have more structure types added rather than having to build now threads of each type. For things like extractors and factories I'm not sure that is possible but for things like radars and shields it might be.
For extractors I'd like the manager to take into account imap threat and rather than selecting the closest extractor, selecting the best.

### DIFF
--- a/lua/AI/AIBaseTemplates/ChallengeMain.lua
+++ b/lua/AI/AIBaseTemplates/ChallengeMain.lua
@@ -32,7 +32,7 @@ BaseBuilderTemplate {
         'EngineerMassBuildersHighPri',
                 
         -- Extractors
-        'ExtractorUpgrades',
+        --'ExtractorUpgrades',
 
         -- ACU Builders
         'Default Initial ACU Builders',
--- a/lua/AI/AIBaseTemplates/NormalMain.lua
+++ b/lua/AI/AIBaseTemplates/NormalMain.lua
@@ -32,7 +32,7 @@ BaseBuilderTemplate {
         'EngineerMassBuildersHighPri',
                 
         -- Extractors
-        'ExtractorUpgrades',
+        --'ExtractorUpgrades',
 
         -- ACU Builders
         'Default Initial ACU Builders',
--- a/lua/AI/AIBaseTemplates/RushExpansionBalancedFull.lua
+++ b/lua/AI/AIBaseTemplates/RushExpansionBalancedFull.lua
@@ -44,7 +44,7 @@ BaseBuilderTemplate {
         'T3NukeDefenseBehaviors',
 
         -- Extractors
-        'Time Exempt Extractor Upgrades Expansion',
+        --'Time Exempt Extractor Upgrades Expansion',
 
         -- ==== NAVAL EXPANSION ==== --
         'NavalExpansionBuilders',
--- a/lua/AI/AIBaseTemplates/RushExpansionLandFull.lua
+++ b/lua/AI/AIBaseTemplates/RushExpansionLandFull.lua
@@ -44,7 +44,7 @@ BaseBuilderTemplate {
         'T3NukeDefenseBehaviors',
 
         -- Extractors
-        'Time Exempt Extractor Upgrades Expansion',
+        --'Time Exempt Extractor Upgrades Expansion',
 
         -- ==== NAVAL EXPANSION ==== --
         'NavalExpansionBuilders',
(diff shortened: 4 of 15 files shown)


Garanas:
It feels to me that the structure manager can be simplified quite a bit, without losing its flexibility. I'll try and make a draft this week of my ideas on it.

relent0r:
> It feels to me that the structure manager can be simplified quite a bit, without losing its flexibility. I'll try and make a draft this week of my ideas on it.

Cool, happy to take on ways to improve it. Lots of potential for the various structure types, especially from the perspective of taking load off the condition monitor. I guesstimated over 300 various conditions being passed through the condition monitor for upgrades currently.

relent0r:
note to self. Noticed this bit of code that is actually changing the name of the AI personality so the DefaultAI check wont work all the time.

```
function SetupMainBase(aiBrain)
    local base, returnVal, baseType = GetHighestBuilder(aiBrain)

    local per = ScenarioInfo.ArmySetup[aiBrain.Name].AIPersonality
    ScenarioInfo.ArmySetup[aiBrain.Name].AIBase = base
    if per != 'adaptive' and per != 'sorianadaptive' then
        ScenarioInfo.ArmySetup[aiBrain.Name].AIPersonality = baseType
    end

    LOG('*AI DEBUG: ARMY ', repr(aiBrain:GetArmyIndex()), ': Initiating Archetype using ' .. base)
    AIAddBuilderTable.AddGlobalBaseTemplate(aiBrain, 'MAIN', base)
    aiBrain:ForceManagerSort()
end
```

relent0r:
Cancelling this PR as the work Jip is doing will cover it.

PR #4432, opened 2022-11-20
Nav mesh - add in label metadata

Garanas:
The metadata of a label can contain information such as the size of the area with that label, the number of extractors and a reference to that extractor.

Todo:

- [x] debug function via the UI


relent0r:
Did you have an example of what the Area number might look like for a corresponding map? Point of reach v4 is a good map for this logic.

If I was going to have the AI leverage this for deciding on expansion establishment.

There are main islands with spawns and lots of mass points over large areas. Then there are small islands will low numbers of mass points. This meta data would allow me to prioritize (I think) which islands should have bases established on first and which ones are even worth the investment.

This metadata should also allow generation of AI expansion markers for maps that don't have any markers.

Garanas:
I'm working on the UI to visualize the data. Then we can test it and see what the data looks like.

On the topic of expansion markers: we'll need to have a discussion as to what an expansion marker is - what is the criteria? I understand the idea, but for us to generate those locations (or rather: areas?) we'll need a criteria.

relent0r:
> I'm working on the UI to visualize the data. Then we can test it and see what the data looks like.
> 
> On the topic of expansion markers: we'll need to have a discussion as to what an expansion marker is - what is the criteria? I understand the idea, but for us to generate those locations (or rather: areas?) we'll need a criteria.

Since our time zones dont line up very well. I've started writing up this as a way to try and describe the various AI markers and what they are used for. I think we aim for the primary 3 (Large expansion, Expansion and Naval Area) first. I've described the criteria as we worked out when Uveso was doing his original marker generator.
https://github.com/relent0r/RNGAI/wiki/NavMesh-Definitions---Expansion-Markers-and-other-AI-markers

This is just a starter, we can discuss some of the more intricate aspects of position selection next time we chat.

Garanas:
https://user-images.githubusercontent.com/15778155/203727703-26f5e060-186f-42b4-b51d-39fca4b6e2bd.mp4

@relent0r see also the video, ready for you to check. The area is in square kilometers.

relent0r:
> 

Did a quick test and there is something off about the mass point count. I 'think' its one short quite often. Its coming up as an odd number on most maps where it should be even.

Unrelated to this, so I can create a separate issue for it. But on this map it thinks it can path to a place that it can't.

Regor VI Highlands

Garanas:
@relent0r try again, there was a subtle bug in the generator.

relent0r:
> @relent0r try again, there was a subtle bug in the generator.

Retested and it works much better now. Mass point counts are correct (and you even fixed that issue on regor highlands).

awesome work!

Garanas:
@relent0r try the air pathing now, re-introduced the corner neighbors

relent0r:
> @relent0r try the air pathing now, re-introduced the corner neighbors

Ah yes that looks like a much smoother path line across the various layers.

PR #4434, opened 2022-11-20
Uef bubble shield Rework

Spikey84:
Makes the UEF Bubble Shield upgrade a parallel upgrade to the personal shield. All stats are placeholders until otherwise said. I just threw in the ones from my mod testing this feature.  

Screenshot from 2022-11-20 18-50-39


MrRowey:
@Spikey84 were going ahead with this change but few changes needed
- Same Cost as the Personal Shied
- Shield to have 9k hp

MrRowey (approved):

PR #4436, opened 2022-11-22
AI Factory Builder Retune phase 1

relent0r:
**Description of changes**
This PR tunes the default AI's factory builders and removes the pointless complexity of the original.

Remove some of the factory builder groups and make sure the main factory builder group provides the same functionality. 

Add a minimum factory builder so that on land maps where the AI has pathing to the enemy it will get 4 land factories quicker. This is still checking the factory cap on the base template so certain AI personalities like easy will remain with a low count.

Decrease the mass to factory requirements for the main bases for the T1 phase so the AI will get the minimum factory counts up in the early game rather than having to wait for the mid T2 phase.

Makes changes to some base templates where the number of land factories was too low.

Changes the eco requirements for transports in general. The requirements were so low that the AI could only build transports when its eco was hurting. The transport builders need alot more work and they are needlessly complex and excessive. But that is for another PR.

Add a multi tier engineer platoon template. This will allow the first available engineer to take on certain builders. Rather than needing a specific tier of engineer.

Added a pathing check condition for builders that want to checking if a specific path is available to the current enemy. Used for a factory builder and will be used in future for amphib tank decision making etc.
Also fixed a logic flaw for the pathing conditions when an AI doesn't have a current enemy.

Test setup for the changes
This is an observation based test. The AI should be better about making sure it has a viable number of land factories during the T1 phase rather than waiting until the T2 phase depending on resources. AI should not be quite as excessive about building transports, but still bad.

### DIFF
--- a/lua/AI/AIBaseTemplates/ChallengeExpansion.lua
+++ b/lua/AI/AIBaseTemplates/ChallengeExpansion.lua
@@ -20,7 +20,7 @@ BaseBuilderTemplate {
         'T1EngineerBuilders',
         'T2EngineerBuilders',
         'T3EngineerBuilders',
-        'EngineerFactoryConstruction',
+        'EngineerFactoryConstructionExpansion',
         
         -- ==== LAND UNIT BUILDERS ==== --
         'T1LandFactoryBuilders',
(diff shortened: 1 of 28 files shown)


relent0r:
Note. Added a missing property to the `T1 Engineer Find Unfinished` builder called 'AssistUntilFinished'. Missing this was causing the engineer to drop out of its platoon function and then pick up the builder again every 10 ticks or so thrashing the engineer manager.

Garanas (approved):

PR #4438, opened 2022-11-23
Fix: Seraphim torpedo bombers doing more than 750 damage Fixes #4261 and delete redudant code.

### LINKED ISSUE #4261: Seraphim torpedo bombers doing more than 750 damage sometimes
**Describe the bug**
Seraphim torpedo bombers to more damage to some units, depending on the angle/spot at which their torpedos hit.

**To Reproduce**
Spawn an aeon carrier and an enemy sera torp bomber. Have the torp bobmer attack the carrier once from a horizontal direction, with the torpedos hitting the carrier's sides, and it will do 834 damage, about 11% more than it's supposed to:
Forged Alliance 10 10 2022 20_25_46 (2)

If the torpedo approaches the carrier from a vertical direction, hitting it's front or back, it does 750 damage as expected:
Forged Alliance 10 10 2022 20_26_09 (2)

**Expected behavior**
Regardless of firing angle or hit location, each torpedo bomber salvo should deal a consistent 750 damage.

**Additional context**
2 minute replay exhibiting the bug: 18245313  
No sim mods were active, and at least one other person in discord could reproduce my findings.  
I haven't exhaustively tested all torp bombers, but it the bug seems to be confined to the sera torp bomber and is most easily visible on larger ships. Might work for other ships/targets too though.

xXEddieXxx:
The bug is way more prominent than i first thought the seraphim torpbomber deals 833 dmg to almost all t3 Naval units and does this by every third ore fourth run(quite common) , this is because after the calculation of the Damage, the original projectile and the childprojectiles share the same Damagetable, so if the parent actually hits the target, it deals damage as an additional projectile 750(9*ca.83(rounded) + ca.83 =833

to fix this, i tried to remove the Damage from the Parentprojectile so if the timing is off and it hits the Target, no additional damage will be dealt. 
I actually dont know if this is the best way to do it and im open for a better solution, but it fixes the bug! 

To test this, just spawn in a t3 ship(i suggest t3 Aoen Battleship  because of lacking AA). I also tested it with a range of other Naval units.

## Additional context
I also deletet some redudant conditions and changed one value, that was not making any sense. Also some refactoring on the numbers was done to stick to our convention  (.1 ->0.1).

Please if you know a more beautiful way to get rid of the Damage value of the parentprojectile tell me :D

### DIFF
--- a/projectiles/SANHeavyCavitationTorpedo04/SANHeavyCavitationTorpedo04_script.lua
+++ b/projectiles/SANHeavyCavitationTorpedo04/SANHeavyCavitationTorpedo04_script.lua
@@ -22,23 +22,11 @@ SANHeavyCavitationTorpedo04 = Class(SHeavyCavitationTorpedo) {
 
         PauseUntilTrack = function(self)
                 local distance = self:GetDistanceToTarget()
-                local waittime
                 local turnrate = 360
-                -- The pause time needs to scale down depending on how far away the target is, otherwise
-                -- the torpedoes will initially shoot past their target.
-                if distance > 6 then
-                        waittime = .1 --0.45
-                        if distance > 12 then
-                                waittime = .1--0.7
-                                if distance > 18 then
-                                        waittime = 0.1--1
-                                end
-                        end
-                else
-                        waittime = .2
+                if distance < 6 then
                         turnrate = 720
                 end
-                WaitSeconds(waittime)
+                WaitSeconds(0.1)
                 self:SetMaxSpeed(14)
                 self:TrackTarget(true)
                 self:SetTurnRate(turnrate)
@@ -49,6 +37,6 @@ SANHeavyCavitationTorpedo04 = Class(SHeavyCavitationTorpedo) {
         local mpos = self:GetPosition()
         local dist = VDist2(mpos[1], mpos[3], tpos[1], tpos[3])
         return dist
-    end,
+        end,
 }
 TypeClass = SANHeavyCavitationTorpedo04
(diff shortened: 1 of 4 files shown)


xXEddieXxx:
can a reviewer change this PR back to draft i actually misclicked and cant convert it back :D Thanks

Garanas:
I've given it a thought and I think this solution is fine, The moment the projectile splits it should no longer do any damage.

Garanas:
@xXEddieXxx is there anything else that you'd like to add? And have you confirmed that this fixes the issue? If so, you can set it as ready for preview.

xXEddieXxx:
@Garanas Hey if this is the best solution (setting the table to nil) im fine with it. I would not add anything else to that, i dont like to pack to many things in. Trying to stick to the scope of the PR/Issue.

xXEddieXxx:
@Garanas to your second point. Is there any way to write a Test for that? I Confirmed it visually by staring at the display for 40 Minutes :D

Garanas:
Anything that directly interacts with the engine is not testable, as we can't properly mock the engine. The answer therefore is no.

projectiles/SANHeavyCavitationTorpedo04/SANHeavyCavitationTorpedo04_script.lua
```diff
@@ -22,23 +22,11 @@ SANHeavyCavitationTorpedo04 = Class(SHeavyCavitationTorpedo) {
...
-                if distance > 6 then
-                        waittime = .1 --0.45
-                        if distance > 12 then
-                                waittime = .1--0.7
-                                if distance > 18 then
-                                        waittime = 0.1--1
-                                end
-                        end
-                else
-                        waittime = .2
```
Hdt80bro, line 38: This one isn't `.1`, so you've changed the wait time in this case. I've no idea if this is actually what we want, as the original scaling of the wait time over distance seems to have been completely flattened.

xXEddieXxx: This is actually the weird part of the code i mentioned above. the Waittime should decrease by distance(it didnt). This else statement covers the smallest possible distance and for some reason the wait time gets increased?? I was pretty confused by it and changed it. The tests didnt show any difference in the beavior of the torpedos and i think it should stay this way, the code complety disagreed with the comment that was written there before.

Garanas (approved):

PR #4441, opened 2022-11-23
T3 Arty Adjustment

Spikey84:
- E cost of T3 arty is multiplied by 1.5.
- E cost to reload the arty is multiplied by 2. (Confirmed not to take longer than arty reload with full adjacency)

### DIFF
--- a/units/UAB2302/UAB2302_unit.bp
+++ b/units/UAB2302/UAB2302_unit.bp
@@ -99,7 +99,7 @@ UnitBlueprint {
         UniformScale = 0.125,
     },
     Economy = {
-        BuildCostEnergy = 915000,
+        BuildCostEnergy = 1372500,
         BuildCostMass = 73200,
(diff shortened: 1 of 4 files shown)


Garanas:
I'm not sure if people know that stalling energy will cause the artillery to not fire. When you increase this, there should be some visual indicator that the artillery is stalling, similar to how shields or radars have a visual indicator. As an example: this can be achieved using effects, that emit according to ratio satisfied that tick.

Tagada14:
I agree it would be nice to have an indicator, would the game team be interested in adding that (can be in this or separate PR)? Unless Spikey knows how to and wants to do that as well.

Garanas:
> would the game team be interested in adding

No, not quite. A game needs to communicate its state properly. That is part of balance too.

Spikey84:
I have created an issue to deal with the communication of the game state issue. Afaik this can be merged.

PR #4442, opened 2022-11-23
GC Claw Adjustment

Spikey84:
Compensates for the GC's claws working correctly. To accomplish this the claws are changed to not instantly kill units that enter them, now they do 720 damage per second until the unit is killed. The pull velocity remains the same. This results in the gc being slightly better than its original form before the claw change.  All data used to come to this conclusion can can be found [here](https://docs.google.com/spreadsheets/d/1zmevPyRNKlI2gxxqL646FgmVULIta2bRDR_0A4iJCYA/edit?usp=sharing). These numbers have not yet been passed by Tagada and are subject to change. Below are examples of my testing.

Screenshot from 2022-11-23 17-32-10

Screenshot from 2022-11-23 16-05-27
Screenshot from 2022-11-23 15-55-27

https://replay.faforever.com/18584864

### DIFF
--- a/lua/aeonweapons.lua
+++ b/lua/aeonweapons.lua
@@ -51,9 +51,9 @@ ADFTractorClaw = Class(Weapon) {
     BeamFx = { EffectTemplate.ACollossusTractorBeam01 },
 
     SliderVelocity = {
-        TECH3 = 1,
-        TECH2 = 7,
-        TECH1 = 13,
+        TECH3 = 10,
+        TECH2 = 13,
+        TECH1 = 16,
     },
 
     --- Adds logic to catch edge cases
@@ -232,6 +232,14 @@ ADFTractorClaw = Class(Weapon) {
                 WaitTicks(1)
 
                 if not IsDestroyed(unit) then 
+                    
+
+                    self:MakeVulnerable(target)
+                    while not IsDestroyed(target) and not IsDestroyed(unit) do
+                        WaitSeconds(0.1)
+                        Damage(unit, bonePosition, target, 1, "Normal")
+                    end
+
                     CreateLightParticle(unit, muzzle, self.Army, 4, 2, 'glow_02', 'ramp_blue_16')
                     Explosion.CreateScalableUnitExplosion(target, 3, true)
 
@@ -247,6 +255,7 @@ ADFTractorClaw = Class(Weapon) {
                     -- create thread to take into account the fall
                     self:ForkThread(self.TargetFallThread, target, trash, muzzle)
                     self:ResetTarget()
+                    LOG("0")
                 else 
                     self:MakeVulnerable(target)
                     trash:Destroy()
--- a/lua/aeonweapons.lua
+++ b/lua/aeonweapons.lua
@@ -127,7 +127,7 @@ ADFTractorClaw = Class(Weapon) {
     OnInvalidTargetThread = function(self)
         self:ResetTarget()
         self:SetEnabled(false)
-        WaitSeconds(0.4)
+        WaitSeconds(10)
         if not IsDestroyed(self) then
             self:SetEnabled(true)
         end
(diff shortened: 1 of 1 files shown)


Garanas (changes requested):
This will make them feel broken again. Highly suggest reducing the speed of units being tractored in instead.

Tagada14:
Yes, that was the intention. I think to achieve the goal of the GC being balanced and the visuals being good we will need to strike a balance between increasing the reload time (to eg. 2-5 sec) and increasing how long it takes to suck in a unit (if you make it too long it will also look bad, I wager that you could increase that by a maximum of 1.5-2 times)

Garanas:
It is not about visuals being good. It is about the player being able to understand the state of the unit (why is it not firing?). This pull request adds in a significant delay between units, but I think it does it the wrong way. But if there is a delay between units, some 'charge' effect would tell the player what is going on.

Tagada14:
If the reload is reasonably short eg. <4 seconds then you don't really need a reload bar or anything like that. The weapon is firing cause it's reloading as all weapons do. I agree that some visual effects would be nice but they shouldn't and won't stop us from balancing a unit. If someone else (or Spikey if he feels like it) wants to add that then that's fine. It can be either in this or in a separate PR.

Garanas:
> I agree that some visual effects would be nice but they shouldn't and won't stop us from balancing a unit.

I'll mimic your tone: If you're not willing to do that and cause the unit to feel broken again then **I will stop you** from 'balancing' the unit that I spent two days fixing.

Communicating the game state is key. That is part of the user experience. The same experience that we both should guard.

Spikey84:
I have updated this. Now as @Garanas suggested I am reducing the speed of the claws for t3. This results in a similar time (see the spreadsheet) with the only change being that units will shoot once in a while duing the sucking process. The shooting and sucking issue seems to not be a big deal since it only really buffs high alpha units since they can shoot once per rotation and still do epic damage. If this is a problem we should be able to stop the unit from shooting while being sucked. Does this satisfy everyone involved @Tagada14 @Garanas ?

BlackYps:
> Communicating the game state is key. That is part of the user experience. The same experience that we both should guard.

When you are zoomed out then you don't see the hypothetical visuals anyway. When the claws are consistently dragging units then I think people can infer that there is a reload time and we don't necessarily need additional visuals

Spikey84:
@BlackYps 
With the current version of the pr the units are just sucked slower, does a great job showing the timing between shots and fixed the balance.

Tagada14:
This PR is currently completely broken
image

Tagada14:
It seems like the GC is partially broken overall. @Garanas 
See: https://replay.faforever.com/19082871
The Percivals being sucked in often go underground, if we reduce the suck-in speed then this issue gets worse since they spend a significant amount of time underground.

Tagada14:
Another issue that is made visible in this PR is that GC can somehow mess up the targetability of a unit. I suspect that this happens when it starts sucking a unit in and then it drops it or something. But it results in the GC not being able to hit the target at all.

Garanas:
That units end up going underground has always been an issue with how the tractors work. When you reduce the speed it just becomes more visible.

Spikey84:
I am working on making it so the suck is as fast as before but the unit slowly dies once it reaches the claw. Currently the issue is that the GC beam targets the unit in the claw still. Will try to have this done within the next few days.

Garanas:
Do you need help?

Spikey84:
Sure, i left something in discord about how to prevent the beam from targeting a unit but still allow it to be damaged, but it must have been lost in other chats. It seems like if i can get that sorted then make sure the death of the unit is handled correctly ill be all set. After that i will just adjust the damage until its closer to the old version.

Spikey84:
@Tagada14 @Garanas This version seems to be balanced to the old version, easy to read the state of the claws, and not broken looking. I tried to find edge cases and fixed any i saw. Look good?

Garanas:
https://user-images.githubusercontent.com/15778155/215031479-aa92eae5-46ab-4f3b-afb9-fcb77955973c.mp4

Garanas (approved):
I've not tested for edge cases, but it looks good 👍

Tagada14:
I will test this later today and merge it if everything works. Great job!

Spikey84:
@Garanas very nice effects!

PR #4454, opened 2022-11-28
AI utility function and rallypoint set change

relent0r:
**Description of changes**
This PR implements a small utility function for the AI to help it perform a few task.
The ShiftPosition function will lerp towards or away from a position based a distance specified. It will be used for simple position shifts and also for avoiding enemy units when that logic is implemented.

The MapCenterPoint attribute is added to the AI brain so that is can easily query it for position references. This will be used for a few things.

The rallypoint logic change. When a map does not contain any rally point markers currently the AI will pick a random position at 30 units somewhere from the current. When used for rally points it means that it can often be placed in front, behind or to the side of a base. Apart from the extra time spent moving it also means that large formations that are formed up at the rally point will have to try and move through the AI base at times. 
Also any platoon returning to base have uncertain return points unless a base center point is used, which causes the same issues. This behaviour can be seen with the attackforceai when it can't find an attack position. It will return to the center point of the base and eventually get stuck. With a more reliable rally point set I can make platoons return to the rally point instead of the base center point and have confidence they won't end up with the same issue.

**How to test**
Start an AI skirmish game on a map that has no rally point markers (crash site for example). Once the AI has built its first factory the rally point should be set towards the map center point at around 25 units from the base center point. 25 units was found to be the sweet spot for various map types and avoiding being too close. Keep in mind that the AI has a rally point monitor that will shift the rally point should a structure be placed on it.

note that we are not using the navmesh yet so a map with no rally point usually means a map with no markers at all. I currently use Uveso's marker generator for these scenarios but it does not create rally point markers. So for this particular scenario it is useful for the test.

This is not a perfect solution for all maps with no rally point markers but it is much better than the current.

p.s I suck at maths, the shift position utility function was gifted by Softles.
p.p.s This doesn't solve the problem of naval factories setting their waypoint to the base one.  I believe this only happens when the naval rush template is selected which we shouldn't be doing anyway.

### DIFF
--- a/lua/AI/aiutilities.lua
+++ b/lua/AI/aiutilities.lua
@@ -2971,7 +2971,7 @@ function ShiftPosition(pos1, pos2, dist, reverse)
     --if the reverse bool is set it will go in the oposite direction e.g towards/away
     --It is multipurpose, used for simple vector3 lerps and enemy avoidence logic
     if not pos1 or not pos2 then
-        WARN('*AI WARNING: ShiftPosition Missing positions')
+        WARN('*AI WARNING: ShiftPosition missing positions')
     end
     local delta
     if reverse then
--- a/lua/AI/aiutilities.lua
+++ b/lua/AI/aiutilities.lua
@@ -2959,3 +2959,30 @@ function AIFindFurthestExpansionAreaNeedsEngineer(aiBrain, locationType, radius,
 
     return retPos, retName
 end
+
+---@param table pos1
+---@param table pos2
+---@param integer dist
+---@param bool reverse
+---@return table
+function ShiftPosition(pos1, pos2, dist, reverse)
+    --This function will lerp a position in two ways
+    --By default it will shift from pos2 to pos1 at the specified distance    
+    --if the reverse bool is set it will go in the oposite direction e.g towards/away
+    --It is multipurpose, used for simple vector3 lerps and enemy avoidence logic
+    if not pos1 or not pos2 then
+        WARN('ShiftPosition* Missing positions ')
+    end
+    local delta
+    if reverse then
+        delta = VDiff(pos1,pos2)
+    else
+        delta = VDiff(pos2,pos1)
+    end
+    local norm = math.max(VDist2(delta[1],delta[3],0,0),1)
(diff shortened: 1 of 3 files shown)


Garanas:
It feels to me the rally point can end up in a mountain, or am I being silly here?

Garanas:
What we could do instead is the following:
- Generate a path from the factory that we set the rally point for, to some target (enemy spawn?)
- Find the node in the path that is in the golden ratio, sufficiently far away but not too far

I'm not sure how often this function is being called. But it can be cached, and refreshed every minute if required

relent0r:
> 

Yes your right, there is potential for a rally point to be in a mountain. I thought about that pathing avenue as well and it is a good one since it will take into account maps like forbidden pass where the best rally point pos isn't in a straight line towards the center. Though that particular map has its own rally point markers so it wouldn't leverage this code.
The rally point is created at the time that the first factory of a base is built so the question would be what position would you use at a path destination? I couldn't use an enemy spawn since then it would only be valid on land only maps and I also wouldnt know if an enemy had been selected or was available. I couldn't use the center point since I wouldn't know if it was pathable. You'd likely end up with if else statements for the various scenarios.

Ideally you actually want to have different rally points depending on the factory type. e.g having an air rally point be behind the base would be better than in front etc.
LOUD does a perimeter set of rally points around the base but that would have a dependency of having certain map layouts and designs so while it works great for LOUD it wouldn't for faf and its many strange map types and sizes. 

I did a bunch of testing on various maps without rallypoint markers and since the distance is only 25 then the chances of hitting a mountain are slim. I think maybe a 5km map like Craig Dunes would have that issue if it didn't already have rally point markers.

With the maps I have the bulk of them already have rally point markers so its a bit of a limited use case. Except for the case of map gen maps, none of them have rally point markers so this weakness really stands out.

I'm happy to take suggestions for how to make this better. I just couldn't think of anything else that would work better with the various scenarios.

Garanas (approved):
Approving this as it is already (a lot) better then what we have now, especially for maps with no rally point markers. We can revise it at a later date with a better implementation when we encounter maps with suboptimal choices.

PR #4457, opened 2022-11-28
Cube maps fixes

BlackYps:
I don't have permission to push to the fa repo directly.
I changed the faction specific textures and added this one as default texture: https://polyhaven.com/a/clarens_midday
It might be necessary to make the terrain a bit lighter, but I failed to do it quickly in Gimp. After some testing on some maps we can decide if we think that is necessary, but for now I think we can merge this and the parent PR from Madmax.

Once we rebased the PBR branch on this we can test the PBR shaders on any map.


Garanas:
is `textures/environment/DefaultEnvCube_TEST.dds` for testing, or for production too?

BlackYps:
I have no idea. It's possible that someone was using it for their map, so I changed it as well

BlackYps:
It's a base game file if that is your question

PR #4458, opened 2022-11-28
Add 'Spawn' to search for spawn markers that are in use by players

### LINKED ISSUE #4453: GetMarkersByType not backwards compatible with Blank_Markers
Jip this is for your reference

**Describe the bug**
When using the GetMarkersByType the resulting response doesn't match the default function.

When requesting  a Blank_Marker type the resulting table looks like this.

info: tempMarkers 
info:  - color: ff800080
info:  - orientation: table: 216E1F78
info:  -    1: 0
info:  -    2: 0
info:  -    3: 0
info:  -    type: VECTOR3
info:  - position: table: 216E1F28
info:  -    1: 284.5
info:  -    2: 25.083560943604
info:  -    3: 57.5
info:  -    type: VECTOR3
info:  - prop: /env/common/props/markers/M_Blank_prop.bp
info:  - type: Blank Marker

When using the GetMarkers function which pulls straight from Scenario.MasterChain._MASTERCHAIN_.Markers we get a resulting table of

info:  - Name: 2
info:  - Position: table: 23F639D8
info:  -    1: 227.5
info:  -    2: 25.134349822998
info:  -    3: 454.5
info:  -    type: VECTOR3

I would like to request that for Blank_Markers a Name attribute be added with the ARMY number assigned so that default logic will be able to leverage the resulting table.

The default logic will do this. 
```
local tempMarkers = AIGetMarkerLocations(aiBrain, 'Blank Marker')
        for k, v in tempMarkers do
            LOG('tempMarkers '..repr(v))
                table.insert(markerList, {Position = v.Position, Name = v.Name})
            end
        end
```

Its not the end of world. But I know that modded AI will be leveraging some of these default functions so I can't change them without impacting others.

Garanas:
Closes #4453

```
INFO:  - Spawn: table: 19880FC8
INFO:  -    Count: 1
INFO:  -    Markers: table: 19880D48
INFO:  -       1: table: 13B58050
INFO:  -          color: ff800080
INFO:  -          name: ARMY_1
INFO:  -          orientation: table: 13B58028
INFO:  -             1: 0
INFO:  -             2: 0
INFO:  -             3: 0
INFO:  -             type: VECTOR3
INFO:  -          position: table: 18B0E280
INFO:  -             1: 178.5
INFO:  -             2: 47.776359558105
INFO:  -             3: 182.5
INFO:  -             type: VECTOR3
INFO:  -          prop: /env/common/props/markers/M_Blank_prop.bp
INFO:  -          type: Blank Marker
```

### DIFF
--- a/lua/sim/MarkerUtilities.lua
+++ b/lua/sim/MarkerUtilities.lua
@@ -20,14 +20,17 @@
 local StringSplit = import("/lua/system/utils.lua").StringSplit
 local TableDeepCopy = table.deepcopy
 
+---@alias MarkerType 'Mass' | 'Hydrocarbon' | 'Spawn' | 'Air Path Node' | 'Land Path Node' | 'Water Path Node' | 'Ampibious Path Node' | 'Transport Marker' | 'Naval Area' | 'Naval Link' | 'Rally Point' | 'Expansion Area' | 'Protected Experimental Construction'
+
 ---@class MarkerData
 ---@field size number
 ---@field resource boolean
 ---@field type string
 ---@field orientation Vector
 ---@field position Vector
----@field color Color | nil 
+---@field color Color | nil
 ---@field adjacentTo string         # used by old pathing markers to identify the neighbors
+---@field name? string              # used by spawn markers
 ---@field NavLayer NavLayers        # Navigational layer that this marker is on, only defined for resources
 ---@field NavLabel number | nil     # Navigational label of the graph this marker is on, only defined for resources and when AIs are in-game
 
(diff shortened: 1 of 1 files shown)


Garanas:
@relent0r this would add 'Spawn' as a valid marker type, which includes the markers that correlate to positions occupied by a player. Or do you want it to return all spawns, regardless of whether it is occupied?

relent0r:
> @relent0r this would add 'Spawn' as a valid marker type, which includes the markers that correlate to positions occupied by a player. Or do you want it to return all spawns, regardless of whether it is occupied?

Spawn is a great filtered table. Only question would be how this is returned for the various spawn types in the lobby options. e.g if the player sets the spawn to 'Optimal Balance' would the spawn table still return according to the various spawns? The default AI goes to alot of effort when spawn locations are considered random which I remember causes logic issues. If the marker type of Spawns was always reliably returning a table of valid positions then it'd allow a degree of simplification.

relent0r (approved):

Issue #4459, opened 2022-11-29
Keybind option for "select all navy, not sonar"


Deribus:
**Describe the feature**
A keybind option for "select all navy, not sonar"

**Expected behavior**
I don't want to send my poor T3 sonar to its death every time I select all navy and forget to deselect sonar. Expected behavior should be pretty clear

Garanas:
@xXEddieXxx This one is a good-first-issue, do you have the time and interest to pick this up?

xXEddieXxx:
@Garanas will Look into it

xXEddieXxx:
@Garanas  is this closed ? :D

Garanas:
Oh, yes 😄

Garanas:
And good job on the pull request 👍

PR #4463, opened 2022-11-30
feat: Add new Keybind option "Select all navy, without mobilesonar"/ make the original beavior consistent #4459

xXEddieXxx:
## Description of the changes
I added a new Keybind option :"select_naval_no_sonar". 
It selects all navy units except sonar platforms (Ships with sonar trait still get selected, mabye the naming needs to be refactored).

I also changed the t2 seraphim sonar platform that actually was not selected when the select Navy keybind was used.(in contrast to all other mobile sonars)
It is now consisten with all sonarplatfroms that when u select all navy they also get selected. 

## Test setup for the changes
Spawn in as many navy units and sonarplatforms as you like and try the new Keybindoption. 

## Additional context
If there is anything off with description or naming just tell me i will change it :)


4z0t:
> @4z0t is there no way to subtract categories using the console command `UI_SelectByCategory`?

@Garanas no, only union and intersection.

Garanas (approved):

lua/keymap/keydescriptions.lua
```diff
@@ -326,6 +326,7 @@ keyDescriptions = {
     ['select_air_no_transport'] = '<LOC key_desc_0253>Select air but no transports',
     ['select_air_transport'] = '<LOC key_desc_0254>Select only transports',
     ['lock_zoom'] = '<LOC key_desc_0255>Lock the zoom to current level',
+    ['select_naval_no_sonar'] = 'Select all Naval units except Sonarplatforms',
```
Garanas, line 329: Is 'Sonarplatforms' one word?

xXEddieXxx: I will check it after work and change it acordingly. :D

lua/keymap/misckeyactions.lua
```diff
@@ -17,7 +17,9 @@ function lockZoom()
     end
 end
 
-
+function navalNoSonar()
+    UISelectionByCategory("NAVAL + MOBILE", false, false, false, false) SelectUnits(EntityCategoryFilterDown(categories.ALLUNITS - categories.MOBILESONAR, GetSelectedUnits()))
```
4z0t, line 21: This plus doesn't do anything actually, remove it to not confuse anyone.

4z0t (changes requested):

lua/keymap/misckeyactions.lua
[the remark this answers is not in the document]
xXEddieXxx: Can you explain ?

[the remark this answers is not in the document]
Garanas: There's no need to cache this. The code path is not critical.

@xXEddieXxx this is caching of catories:
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultunits.lua#L25-L27

It prevents re-computing the categories arithmetic's. Those arithmetic's are engine calls and are therefore expensive. Caching the computation is useful when you're dealing with critical paths (paths that are run so often that you want to get everything out of it that you can possibly get, performance-wise)

Garanas: If you happen to run into it - yes. But don't go looking for it 😄 !

xXEddieXxx: Love that sentence ! xD

lua/keymap/keydescriptions.lua
Garanas: image

I think it is two separate words. Could you adjust that? After that we can merge.

xXEddieXxx:
ok about my last PR i investigated a little bit more and found, that someone wrote a smart select(wich is used rarely), wich gives you the opportunity to use a Keybind and give it string attriubutes also substraction as you mentioned @Garanas. 
So i deleted my own function and used this instead, i dont know if this is better but it looked nicer to me then having a special function for each Keybind, that are all doing the same.(like my navalnosonar function or the Airnotransport function)
I would like your thoughts about this aswell @4z0t the function looked pretty old can this be used ?

Garanas:
I'm not a huge fan of smart selection because of the manual parsing. But for the sake of a hotkey - it is fine 😃

Garanas (approved):

PR #4466, opened 2022-12-01
Performance - clean up sync and score thread

Garanas:
I've not found out why yet, but these two functions:
- [OnSync](https://github.com/FAForever/fa/blob/performance/cleanup-sync/lua/UserSync.lua#L38)
- [ResetSyncTable](https://github.com/FAForever/fa/blob/performance/cleanup-sync/lua/SimSync.lua#L23)

Are called repeatedly at these events:
- (1) a sim beat
- (2) a (new) unit selection
- (3) end of moving camera

Where (1) makes sense, but (2) and (3) are unexpected. Why that happens is a mystery, but it did result in a lot of additional table trashing in both the sim, UI and the sync. Therefore, the first few commits factor out as many 'default' tables that are expected in the sync. And the last commit 25546e16f88d0a1dd4a2ffa87756865c073cb102 cleans up the sync instead of re-allocating it. This allows us to re-use memory.

The last commits 25546e16f88d0a1dd4a2ffa87756865c073cb102 and b6f13529e3dbf01743e14f89faa8c32c24ba4666 also clean up some of the score tracking. Specifically, it removes the blueprint information tracking for individual units. As achievements do not work it merely bloats up the sync and the total memory usage. We'll introduce it back by just sending the brain statistics at the end of the game, as found here:
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/aibrain.lua#L127-L164

Previous related PRs:
- https://github.com/FAForever/fa/pull/3091

edit: with thanks to @KionX , this is the full list of when sync is called:
- camera move
- unit selected/deselected
- ordering a unit
- pause/unpause
- remove build order of factory
- set factory waypoint


lua/sim/score.lua
```diff
@@ -218,16 +211,6 @@ local function ScoreThread()
             Score.general.lost.mass = brain:GetArmyStat("Units_MassValue_Lost", 0).Value
             Score.general.lost.energy = brain:GetArmyStat("Units_EnergyValue_Lost", 0).Value
 
-            for unitId, stats in brain.UnitStats do
```
Garanas, line 221: Note that this is just a deep copy of the unit statistics available in the brain. The achievements available in the client do not take into account time. See also:

image

That means it doesn't need to be accumulated like this.

PR #4468, opened 2022-12-02
Fixed default terrain type skin function

The-Balthazar:
Not having the second argument in the terrain type be `true` causes it to destroy any c-objects on the mesh to be destroyed. This includes weapons, thrusters, sliders, and so on. It's an unused feature, but is fully implemented aside from this one typo.


Garanas (approved):

BlackYps:
What could this unused feature do?

The-Balthazar:
It could let us do things like this, based on the terrain type of the map.
image

PR #4469, opened 2022-12-02
Cybran navy factories

MadMaxFAF:


Garanas:
Looking great 👍 

image

Garanas:
And can you add a bone where these pillars are, in the center? Then we can put a decal where it meets with the floor.

MadMaxFAF:
yeah i can adjust that bit sticking out of the water not sure exactly where you want a bone added?

Garanas (approved):

PR #4471, opened 2022-12-03
Localize and fix coloring of combo defaults

Hdt80bro:
* Adds localization for the " (default)" label for combo boxes. I should have only added localization support, but I tried to add the database keys based on other translation work available.
* Fixes the coloring of the title not being yellow when the default item is immediately selected
* Moves the ad-hoc callbacks to the combo class
* Other refactoring of the file
* Adds some annotation stuff in the surrounding area


lua/ui/controls/combo.lua
```diff
@@ -508,92 +644,37 @@ BitmapCombo = Class(Group) {
...
-                    local sound = Sound({Cue = self.mClickCue, Bank = "Interface",})
-                    PlaySound(sound)
-                end
-            end
-            self:OnEvent(event)
-
-            return eventHandled
-        end
-
-        self.OnDisable = function(self)
```
Garanas, line 568: Interesting how these were added manually, for each instance 😄

Garanas (approved):
works in-game, can't produce any errors

Garanas:
@Hdt80bro 

This changes to the combobox are apparently not entirely logically equivalent, as it appears this breaks the minimap functionality where you can switch to the spot. I'm talking about these:

image

Don't ask why it uses a combo box, I don't know. But can you look into this?

Hdt80bro:
Uh, sure

Hdt80bro:
The logical difference comes from closing the methods with the `self` upvalue from the initializer instead of using the called method control. This revealed a latent bug [here](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/lobby/lobby.lua#L4812-L4814) where the event handler was not being called on the right control.

PR #4474, opened 2022-12-03
Features for order / cursor behavior

Garanas:
The alt order, which is attack move, now takes precedence over everything. This is in particular useful for engineers and harbingers, where you usually just want to attack move instead of ordering them to reclaim a tree.

The attack ground order now ignores all props and units, similar to the tactical and strategical missiles orders. As a side effect you can no longer choose the 'Return fire' option in the order panel:

image

It was not possible to determine which was which. The return fire option represents attack move. This is already covered by the alt key, which is now always functional when used.

And last but not least, when you are in the move command mode (pressed the hotkey to move) it ignores the 'convert move to patrol' order, allowing you to just issue move orders without having to worry about accidentally causing your units to patrol.


Garanas:
These ideas originate from:
- https://forum.faforever.com/topic/4705/upcoming-developers-patch/75
- https://forum.faforever.com/topic/5237/holding-alt-should-disable-the-reclaim-cursor/3

And from a private discussion with Babel.

ChessBerry:
Gave this a quick test run and both additions work exactly as you'd want them to.

Also I really like the solution to disable the 'convert move to patrol' only when the move command is held, as then it's clear that we wanted a move command and a move command exactly.

Great work as always!

Hdt80bro:
> As a side effect you can no longer choose the 'Return fire' option in the order panel

Um, I can't get out of hold-fire now. Once I select that option, I'm stuck not shooting.

Hdt80bro:
Ohh, are you transforming the toggle so I need to select ground fire to shoot again? That's so confusing--just make it a regular toggle for hold-fire then.
My other concern then is that abstracting away all this and removing fire state removes functionality built into the engine, though I'd need to go in and test more thoroughly what exactly that would be. It definitely makes `SetFireState` more interesting to deal with.

Garanas:
> > As a side effect you can no longer choose the 'Return fire' option in the order panel
> 
> Um, I can't get out of hold-fire now. Once I select that option, I'm stuck not shooting.

I seem to be able to do that, switching back and from hold fire. There's still the ground attack option, see the screenshot.

> My other concern then is that abstracting away all this and removing fire state removes functionality built into the engine, though I'd need to go in and test more thoroughly what exactly that would be. It definitely makes SetFireState more interesting to deal with.

The only thing the return fire state did was act as an attack move, which is what the ALT key does too.

Hdt80bro:
> The only thing the return fire state did was act as an attack move, which is what the ALT key does too.

It turns out that this is untrue--return-fire will only stop for units that add it as a shooter but attack-move will stop for any unit. I have other feedback for this change even if that wasn't the case, so let's not mess with fire states in this PR (since that's not its focus anyway).

Hdt80bro:
Some cubemaps snuck in here by the way :)

Hdt80bro:
> Can you describe how you tested this?

Move some units with return-fire by some power generators: they won't stop.
Attack-move some units with any fire state except ground-fire by: they now stop.
(My exact description about only stopping for active shooters is based on conjecture when I noticed a return-fire unit stopping the precise moment an artillery shell was launched at it--I didn't feel like it was important enough to get a rigorous set of conditions for when they stopped, just that it's different from attack-move)

Garanas:
I see, I was not aware. I'll make sure to revert the disable. Subtle change 😄

Garanas:
@ChessBerry this is merged along with some logging. The executable on FAFDevelop also includes the 'do not issue reclaim order on empty patches of ground'. Could you test these changes, see if you can break them or cause unexpected behavior?

In particular:

```
Introduces various new features to make it easier to issue orders:
- CTRL no longer toggles formation move, instead it represents a move command that ignores everything else
- ALT now represents an attack move command that ignores everything else
- Ground attack can no longer snap to props or units

In addition, when you hold CTRL or ALT the command mode (e.g., a reclaim order via hotkey) is cached and restored. This allows you to enable the reclaim mode, click a few props to reclaim, hold ctrl to issue move order, release ctrl, click more props to reclaim - all without entering reclaim mode again.
```

If you encounter any issues, post a log here and ping me.

ChessBerry:
@Garanas It seems like there was an unintentional(?) change to how the new ALT interacts with hotkeys:

On 'faf' if you press ALT+some-key to use a hotkey you've set up in your hotkey menu, e.g. alt+q for repair, and keep pressing ALT, it gives you the hotkey, in this case repair.

On 'faf develop', if you press the same ALT+q and keep pressing ALT, it gives you the attack command instead of repair.


This is a problem as now every hotkey that is bound with ALT requires twice the actions to function than it did before: First pressing the key combination and then releasing it. 

I was wondering during the last weeks why sometimes my hotkeys weren't doing their job correctly, and it turns it was probably this behavior screwing me in the faf develop games.

PR #4475, opened 2022-12-04
AI - fix incorrect .Result checks and replace with .Status

relent0r:
**Description of the changes**

I realised that the .Result attribute for player brain 'defeat' was changed to .Status and 'Defeat' but none of the AI was updated.

This PR changes the checks to be .Status and 'Defeat'

I'd like someone to sanity check this is correct as we now have other status like recalled etc. It may not be an issue if the 'Defeat' will be the only other status where the game will continue.

### DIFF
--- a/lua/AI/AIBuilders/SorianStrategyBuilders.lua
+++ b/lua/AI/AIBuilders/SorianStrategyBuilders.lua
@@ -867,9 +867,9 @@ BuilderGroup {
             local enemies = 0
             local allies = 0
             for k,v in ArmyBrains do
-                if v.Result ~= "defeat" and not ArmyIsCivilian(v:GetArmyIndex()) and IsEnemy(v:GetArmyIndex(), aiBrain:GetArmyIndex()) then
+                if v.Status ~= "Defeat" and not ArmyIsCivilian(v:GetArmyIndex()) and IsEnemy(v:GetArmyIndex(), aiBrain:GetArmyIndex()) then
                     enemies = enemies + 1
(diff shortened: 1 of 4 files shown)


Hdt80bro:
"Recalled" should be the only new status, and it's specifically designed to not work with AI's--so I don't think you need to worry about that.

Hdt80bro (approved):

Garanas (changes requested):
It feels to me it is better to use either of these:
- https://github.com/FAForever/fa/blob/feature/add-messages-to-interactions/lua/aibrain.lua#L1287-L1290
- https://github.com/FAForever/fa/blob/feature/add-messages-to-interactions/engine/Sim.lua#L49-L51

In particular the latter is true when `SetArmyOutOfGame` is called.

@Hdt80bro should we extend the brain function to include a check for `SetArmyOutOfGame`?

Garanas:
Extended the is defeated function, please use that instead @relent0r

relent0r:
> Extended the is defeated function, please use that instead @relent0r

Ah ok, I'll update the PR to have instances use the IsDefeated function. That should make it future proof.

Hdt80bro:
> @Hdt80bro should we extend the brain function to include a check for SetArmyOutOfGame?

Good call, that's an entirely separate mechanism from the status that's supposed to be meaningfully correlated.

relent0r:
@Garanas Have updated (and fixed a small bug I spotted in testing)

Garanas (approved):

PR #4477, opened 2022-12-04
Print chat message when game is paused / unpaused

Garanas:
image

### DIFF
--- a/lua/ui/game/clientutils.lua
+++ b/lua/ui/game/clientutils.lua
@@ -1,4 +1,26 @@
 
+--******************************************************************************************************
+--** Copyright (c) 2024 FAForever
+--** 
+--** Permission is hereby granted, free of charge, to any person obtaining a copy
+--** of this software and associated documentation files (the "Software"), to deal
+--** in the Software without restriction, including without limitation the rights
+--** to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+--** copies of the Software, and to permit persons to whom the Software is
+--** furnished to do so, subject to the following conditions:
+--** 
+--** The above copyright notice and this permission notice shall be included in all
+--** copies or substantial portions of the Software.
+--** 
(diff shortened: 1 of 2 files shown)


Garanas:
@Hdt80bro suggestions on if we need to, and if so how to localize this? Sending the message in russian to everyone doesn't make a lot of sense, for example. But neither does English, for other languages

Maaach:
well I just saw this by accident but couldn't you just sent a direct/private message to each individual player in their preferred language instead of one same message to all players

Garanas:
That is quite possible, yeah

Hdt80bro:
Auto-localizing chat messages is definitely a problem, and taunts & notify already suffer from this. Sending automated messages through the chat system to individual players based on their localization might work, but as a kludge has several problems. One is that each client must now query each other player for their localization, and then trust that other players will respect it. This introduces an inversion of control, where we have new points in the UI where the client is no longer in charge of its own locale. A more fundamentally one is in the presentation, in that information is given as a player that they did not send (i.e. just like the resource giving, taunt, and notify messages already do... at some point we *will* need a better way to present all this). It would be better to make this distinction at more a fundamental level by transforming the chat box into a message box, where additional messages that are not player chats can be displayed in it in a way that makes it clear what each is. So, instead of it saying `<Player> to all: Paused the game` it might be more like `<Player> paused the game` in a different color to signify "system" text.

This particular implementation also raises the question of whether pausing is something that should be handled via the chat system, or if it should be polled from the sim--there are different timing considerations to each approach, particularly when the game has stalled.

4z0t:
>  suggestions on if we need to, and if so how to localize this? Sending the message in russian to everyone doesn't make a lot of sense, for example. But neither does English, for other languages

@Garanas the way taunts are processed.

Hdt80bro:
Maybe we could do it more pleasantly than the callback mess that Taunts end up creating--I think the way notify works is a little better, but still rather ad hoc.

PR #4479, opened 2022-12-06
Aeon land factory HQ/Support visual reworks

The-Balthazar:
Visual rework of Aeon land factories, using (close to) the original models for the HQs and cut down models for the support factories.

Bonus: removed some redundant blueprint values, and fixed the T3 support having the Cybran unit shader.

Currently this missing updated distance models, but that will mostly be hard to notice, and also the pevious lod1s were never actually updated to represent the old HQ's anyway.


Garanas:
From the logs:

```
INFO: Can't find texture "/units/zab9501/zab9501_normalsTS.dds" -- trying fallback.
INFO: Can't find texture "/units/zab9501/zab9501_normalsTS.dds" -- trying fallback.
INFO: Can't find texture "/units/zab9501/zab9501_normalsTS.dds" -- trying fallback.
INFO: Can't find texture "/units/zab9601/zab9601_normalsTS.dds" -- trying fallback.
INFO: Can't find texture "/units/zab9601/zab9601_normalsTS.dds" -- trying fallback.
```

image

Beyond that, they look great 👍 , the only question I have is whether it is possible to push these side arms up a notch:

image

That way it clips less with the Harbinger, and it also adds to the animation for the t2 -> t3 support.

Garanas (approved):

PR #4484, opened 2022-12-09
Fix minor issue with lazyimport

4z0t:


4z0t:
@Garanas can you confirm it doesnt produce errors now?

Garanas:
Why not undo the change you made: removing line 19 and re-importing the command mode again?

Hdt80bro:
Jip is right, the problem isn't that lazy imports broke something, it's that caching the import removes the ability to use the reloaded module when it becomes dirty. This is a problem when using either `import` or `lazyimport` as dependency tracking is explicitly designed to stop after the file has finished loading. Note that because `lazyimport` doesn't load the imported file when initialing loading the file, it also doesn't create any `used_by` dependencies that would then be reloaded by dirtying the file. Adding more dependencies elsewhere doesn't fix this fundamental problem (even if it did, it should be noted that it would be considered a kludge).

4z0t:
Lazyimport is designed for static loading of modules and not for dynamic unfortunately. I understand your concerns but lazy import is a tool and tools have field of application.

Garanas:
I understand, and we're not removing the ability to lazy import. But the field of application for this particular tool is quite limited.

Garanas:
I've reverted the accidental change where we lazy load command mode.

Garanas (approved):

Hdt80bro:
I wonder if we should add more safety to it to avoid future confusion (e.g. immediately loading engine files, re-enabling dependency tracking while loading, disabling while untracked / removing the function after module loading). It doesn't seem worth it to me though--most of the problems are already inherent in how we handle imports.

PR #4485, opened 2022-12-10
Prevent loading navigational mesh by default and fix issues with M27AI

Garanas:
Prevents the navigational mesh from always loading when an AI is detected. Instead, it only loads when an AI specifically mentions it requires it. This is set in `aitypes.lua` or for mods in their `CustomAIs_v2` folder setup, as an example - changing this:
```lua
        {
            key = 'RNGStandardExperimental',
            name = "<LOC RNG_0002>AI: RNG Standard Experimental",
        },
```
To this:
```
        {
            key = 'RNGStandardExperimental',
            name = "<LOC RNG_0002>AI: RNG Standard Experimental",
            requiresNavMesh = true
        },
```

Will guarantee that the navigational mesh is generated at the start of the game. The `Generate` function of the generator by default prevents generating a new mesh if one already exists - this allows maps, mods or other scripts that may rely on the generate to call this by default without having to check if it is already generated.

This pull request also removes various checks in the builder files as those cause issues for M27AI.

### DIFF
--- a/lua/AI/AIBuilders/AIAirAttackBuilders.lua
+++ b/lua/AI/AIBuilders/AIAirAttackBuilders.lua
@@ -7,10 +7,6 @@
 --**  Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
 --****************************************************************************
 
-if not ScenarioInfo.GameHasAIs then
-    WARN("AIAirAttackBuilders loaded in a non-ai game from:" .. reprs(debug.traceback()))
-end
-
 local BBTmplFile = '/lua/basetemplates.lua'
 local BuildingTmpl = 'BuildingTemplates'
 local BaseTmpl = 'BaseTemplates'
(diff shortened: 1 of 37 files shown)


Garanas:
@relent0r added in a way to properly detect if we're dealing with a base game AI, see also:
- https://github.com/FAForever/fa/pull/4485/files#diff-514781ee9a4283ed2420eb74be449318d9f307f381cb34a11f563b69d5f2fe01R226-R228

As an example of the output:

```
INFO: * OnCreateAI: AIPersonality: (easy)
INFO: ARMY_2
INFO: true
DEBUG: Detected an AI with skirmish systems: ARMY_2, enabling AI functionality
INFO: * OnCreateAI: AIPersonality: (m27ai)
INFO: ARMY_3
INFO: nil
```

relent0r:
> 

Nah no reason to have them on Sorian, I'm not doing anything for it that would require these.

So for me detecting for the default I just need to say `if brain:IsBaseAI then`. Looks nice and simple.

relent0r:
@Garanas I gave this a test with my AI but no navmesh gets generated.

The changes look like this.
https://github.com/relent0r/RNGAI/blob/experimental/lua/AI/CustomAIs_v2/RNGAI.lua

PR #4489, opened 2022-12-11
Reduce memory allocations of props

Garanas:
Overall reduces the number of table fields that props create over their life span. I'll write out the exact changes soon.

Removes the following fields:
- `SyncData` when the prop can not have a label (usually individual trees)
- `Dead` was set when `OnDestroy` was called. This is an unreliable method regardless - instead it now just calls `BeenDestroyed`
- `IsWreckage` is now set in the metatable
- `CanBeKilled` if a prop is invulnerable (it has the property 'INVULNERABLE') then it is assigned a separate prop class instead
- `CanTakeDamage` see above
- `hasLabel` was set if prop has a label, but the existence of `SyncData` replaces that
- `CollisionRadius` was used by wrecks to catch an issue with sharing unfinished units. There is a separate solution for this
- `CollisionSizeX` see above
- `CollisionSizeY` see above
- `CollisionSizeZ` see above
- `CollisionCenterX` see above
- `CollisionCenterY` see above
- `CollisionCenterZ` see above
- `CollisionShape` see above

To give an idea of the reduction of memory, take the following maps:
- Archsimkats Valley (10x10): 8K+ initial props
- Mangrove Delta (20x20): 11K+ initial props
- Crossfire Canal (20x20): 21K+ initial props

Based on the numbers as found here: https://github.com/FAForever/fa/pull/3875

We would prevent on a regular prop the following properties: `Dead`, `CanBeKilled`, `CanTakeDamage`, `hasLabel`, and for some: `SyncData`. That would be at least 5 * 20 bytes worth of data for the keys, along with the values it would be another 16 bytes which make up for 116 bytes. Multiply that by 20K and we easily save about 2 mega bytes worth of memory. Let alone when we take into account SyncData, which would be an empty table worth of 36 bytes: that pushes the number to 2.96 megabytes!

For wrecks it is even worse, as by default the collisions were stored too. That takes up another 8 fields, which also pushes the hash bit of the table to 32 entries instead of only 16.

I'm not _that_ good with these numbers, but in general it is quite a nice improvement 😄 !

### DIFF
--- a/engine/Core/Blueprints/PropBlueprint.lua
+++ b/engine/Core/Blueprints/PropBlueprint.lua
@@ -1,8 +1,12 @@
 ---@meta
 
 ---@class PropBlueprint : EntityBlueprint
+---@field Audio PropBlueprintAudio
 ---@field Display PropBlueprintDisplay Display information for the unit
 ---@field Defense PropBlueprintDefense Defense information for the unit
+---@field Economy PropBlueprintEconomy 
+
+---@class PropBlueprintAudio
 
 ---@class PropBlueprintDisplay
 ---@field MeshBlueprint string Name of mesh blueprint to use. Leave blank to use default mesh.
@@ -14,3 +18,4 @@
 ---@class PropBlueprintEconomy
 ---@field ReclaimMassMax number    Max Reclaimable mass resource.
 ---@field ReclaimEnergyMax number  Max Reclaimable Energy resource.
+---@field ReclaimTimeMultiplier number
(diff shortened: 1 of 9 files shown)


Garanas:
I'd argue that we could go even further:
- `CachePosition` may be obsolete
- `SyncData` with a minor refactoring we can completely eliminate this table / entry
-  `Trash` this one is slightly more tricky as it feels like it should be there regardless, but it is barely used at the moment and the places where it is used it isn't strictly necessary
- `Blueprint` not strictly required, is only referenced in 4 functions

And with that, we'd have 7 entries for a regular prop - reducing the hash section of the table to only 8 entries 😄

Garanas:
When loading `Crossfire Canal` from scratch I can occasionally get below 300 committed memory with this pull request. Without it, it is usually at about 302 or 303.

Garanas:
For now, this is it for props. Usually during the end game there aren't _that_ many props around anymore as they get reclaimed.

PR #4491, opened 2022-12-11
Partial Share Bug Fix

Penguin5:
Import UnitsAnalyzer.ContainsCategory


Garanas:
Are you expecting the categories to be string based? It seems it is better to use this: https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Core.lua#L97

Penguin5:
Alright, @Garanas, I switched it to EntityCategoryContains.

PR #4495, opened 2022-12-12
Attempt at fixing issues with nav generator not running

Garanas:


Garanas:
Tested and confirmed by Relent0r

PR #4499, opened 2022-12-13
Remove `NaturalProducer` from `C-D2 "Rover"`

### LINKED ISSUE #4498: UEF drone wrongly marked as `NaturalProducer`
**Describe the bug**
The C-D2 "Rover" has the [`NaturalProducer` attribute ](https://github.com/FAForever/fa/blob/develop/units/UEA0003/UEA0003_unit.bp#L133) under `Economy` despite producing no resources. 

**Expected behavior**
It does not produce resources so should not have this attribute.

**Additional context**
The [only other units that have this](https://github.com/search?l=&q=naturalproducer+repo%3AFAForever%2Ffa&type=code) are ACUs and sACUs.

haifron:
`C-D2 "Rover"` produces no resources, so should not have this attribute. The only units that should have it are the ACUs and sACUs.

Closes #4498

### DIFF
--- a/units/UEA0003/UEA0003_unit.bp
+++ b/units/UEA0003/UEA0003_unit.bp
@@ -130,7 +130,6 @@ UnitBlueprint {
         BuildableCategory = {
             'BUILTBYTIER3ENGINEER UEF',
         },
-        NaturalProducer = true,
     },
     Footprint = {
         MaxSlope = 0.25,


Garanas (approved):

Garanas:
I'm not sure if this value does anything. But it is technically correct to remove it from this unit.

Garanas:
If these are similar blueprint changes like these then it is fine to make a pull request. We prefer smaller pull requests. We use issues primarily to discuss larger changes and feature requests.

PR #4500, opened 2022-12-13
Fix option keys and improve their naming scheme

Garanas:
Not exactly ideal, but I suppose we should've reviewed the settings better before merging them 😄

### DIFF
--- a/lua/options/options.lua
+++ b/lua/options/options.lua
@@ -391,26 +391,26 @@ options = {
 
             {
                 title = "<LOC selectionsets0001>Steal from other control groups",
-                key = 'selection_sets_add_behavior',
+                key = 'steal_from_other_control_groups',
                 type = 'toggle',
                 default = false,
                 custom = {
                     states = {
-                        {text = "<LOC _Yes>No", key = false },
-                        {text = "<LOC _No>Yes", key = true },
+                        {text = "<LOC _No>No", key = false },
+                        {text = "<LOC _yes>Yes", key = true },
                     },
                 },
             },
 
             {
                 title = "<LOC selectionsets0004>Add to factory control group",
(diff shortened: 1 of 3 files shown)


lua/options/options.lua
```diff
@@ -391,26 +391,26 @@ options = {
...
-                key = 'selection_sets_add_behavior',
+                key = 'steal_from_other_control_groups',
                 type = 'toggle',
                 default = false,
                 custom = {
                     states = {
-                        {text = "<LOC _Yes>No", key = false },
-                        {text = "<LOC _No>Yes", key = true },
+                        {text = "<LOC _No>No", key = false },
+                        {text = "<LOC _yes>Yes", key = true },
```
Hdt80bro, line 400: Needs to be `_Yes`

Hdt80bro:
I've found that renaming preference keys causes problems, since the preferences file still contains the old key and lingers (though it only warns you for rogue key actions I believe).

Garanas:
Yes, people will need to re-enable it by toggling it again. Or are we talking about other issues, besides keys lingering in the options file?

I renamed them because they:
 - made no sense when I read the if statement
 - would be outdated because we switch the values up

Hdt80bro:
No, those are both concerns I was talking about. However, I didn't make the connection that you changed the logic and thus current preferences would flip. We might be able to solve both the problems of requiring players to do ultimately needless work and leaving dead keys in the file by introducing a version migration system, then adding a function that sets the new key and deletes the old one for this update--it might even be a good way to help mitigate future update pains for all parties involved, though setting up the infrastructure to check for that might not be worth it yet for one feature fix. Something to think about, I suppose.

Garanas:
For now - something to think about.

PR #4503, opened 2022-12-14
Break cycles in dependencies

Garanas:
@relent0r it appears co-op had the same issues 😄


speed2CZ (approved):

relent0r:
> @relent0r it appears co-op had the same issues 😄

yay not alone!

PR #4504, opened 2022-12-15
Remove redundant lines in engineer blueprints

haifron:
## Issue details
`NeedToFaceTargetToBuild` defaults to `false` when not set in a bp. These lines are therefore unnecessary.

**Additional Information**
They are only there as these attributes were changed from `true` to `false` in 7f3824270 (september 3 2011) to incorporate features from a mod called "CBFP" (community bug fix patch) to: "make the engineers build faster".

**My testing**
I tested this & it runs fine, and the units' build behaviour is the same as before (which makes sense given only some ACUs even have this line & we clearly know they build the same).

## Testing setup:
1. spawn each of the units whose bps have been changed:
`UEL0001`, `UAL0105`, `UAL0001`, `XSL0105`, `XSL0208`, `XSL0309`, `XSL0001`
2. Order them to build something behind them
3. They should not (do not) turn around (although the `BuildBone` will rotate to face the build site)
- To compare to the `NeedToFaceTargetToBuild = true` option, spawn a mantis & order it to assist a unit building, the entire unit will have to move & turn such that the body is facing the construction site.
 
>**Fun fact:**
The only unit with `NeedToFaceTargetToBuild = true` is the mantis.
If you were to set `NeedToFaceTargetToBuild = False` for the mantis it will look like how a hive builds.

### DIFF
--- a/units/UAL0001/UAL0001_unit.bp
+++ b/units/UAL0001/UAL0001_unit.bp
@@ -306,7 +306,6 @@ UnitBlueprint {
         },
         MaxBuildDistance = 10,
         NaturalProducer = true,
-        NeedToFaceTargetToBuild = false,
         ProductionPerSecondEnergy = 20,
         ProductionPerSecondMass = 1,
         StorageEnergy = 3900,
--- a/units/UAL0105/UAL0105_unit.bp
+++ b/units/UAL0105/UAL0105_unit.bp
@@ -202,7 +202,6 @@ UnitBlueprint {
             'BUILTBYTIER1ENGINEER AEON',
         },
         MaxBuildDistance = 5,
-        NeedToFaceTargetToBuild = false,
         SacrificeEnergyMult = 0.6,
         SacrificeMassMult = 0.6,
         StorageEnergy = 0,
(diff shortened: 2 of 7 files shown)


haifron:
welp I changed the branch name on my repo and that closed this PR. Didn't know that would do that.
>oops

Garanas:
Best to branch off, and make a pull request from that branch.

Hdt80bro:
There's so much fluff & obsolete parameters left in the blueprints--in general, booleans are always optional and default to `false` (except for a few weird cases). My perspective is that most of the redundant information there serves as documentation, but we definitely do need to go through the database with a comb one of these days. I'm actually working on documenting the annotation classes for blueprints right now, so that might help in the future. Thoughts @Garanas ?

Garanas:
I'm all for culling data that is not used. But it only makes sense to me if the value either:
- is semantically wrong, and therefore should be corrected (by removing it)
- the value is unused, and also has no default (removing it would save a table entry)

Beyond that, removing data that is set to the default value has little benefit in practice - as the game will set the table field regardless.

Garanas:
Oh, and then there's the unit database. I'm not sure how that reacts to entries missing that are set to their default value.

Garanas (approved):

PR #4505, opened 2022-12-15
AI Naval tuning

relent0r:
**Description of changes**
This PR retunes some aspects of the AI naval play. 

- Fixes a an errors where the AI would not upgrade beyond T2 if the navalrush ai was selected.
- The AI would use the Turtle expansion base template for naval expansions, naval area markers are now excluded.
- Allows the adaptive AI to use the NavalExpansionLarge base template, previously it would build the naval factory and then do nothing else.
- Decrease factory counts on the expansion builder for T1 Naval so the AI did not ruin its economy.
- Add logic for detecting which mass markers could be hit by frigates and if there are under water mass markers. Happy for someone to help make this better. Its weakness currently is that it will detect a position in water but not understand the water depth so it may decide a raid position even though the frigate couldn't get close enough to shoot it due to water depth. Edit. Modified so it will require a minimum 1.1 depth.
- Add a priority function that will adjust depending on map water ratio and naval raiding capability.
- Adding of a submarine based mass marker raiding platoon, requires underwater mass points.
- Add platoon template for submarine mass marker raiders
- Add support for naval platoons to use the guardmarker platoon function.
- Add a Layer property to bases so the AI can easily distinguish between land and naval builder managers.
- Add a layer check to the returntobaseai function so that naval units will return to naval bases and land units with not return to naval bases.

Some of the testing here requires logging, and some observation.
You can log the bool brain attributes to validate if under water mass points are detected correctly and if the frigate raiding mass points are detected and added to the frigate raid table correctly (note the naval raiding logic does not use this table yet).
The other changes are mainly observation, making sure that the correct AI are able to use naval bases.

There are still a many issues with naval play but this will help the AI from completely ruining its naval chances instantly.

p.s in regards to the naval expansion builder, the priority function could be tuned better/ access more data to try and make a better decision for the situation. the default code base doesn't use priority functions but they have value as a way to enable/disable builders and improve builder selection based on available data. I believe if we had a good data set rather than function calls then you could use them in ways that are similar to the unused strategy manager but without all the performance cost.

### DIFF
--- a/lua/AI/AIBaseTemplates/NavalExpansionLarge.lua
+++ b/lua/AI/AIBaseTemplates/NavalExpansionLarge.lua
@@ -76,7 +76,7 @@ BaseBuilderTemplate {
         end
 
         local personality = ScenarioInfo.ArmySetup[aiBrain.Name].AIPersonality
-        if personality == 'rushnaval' then
+        if personality == 'rushnaval' or personality == 'adaptive' then
             return 100
         end
 
--- a/lua/AI/AIBaseTemplates/RushExpansionNaval.lua
+++ b/lua/AI/AIBaseTemplates/RushExpansionNaval.lua
@@ -13,7 +13,7 @@ BaseBuilderTemplate {
         -- ==== ECONOMY ==== --
         -- Factory upgrades
         'T1BalancedUpgradeBuildersExpansion',
-        'T1BalancedUpgradeBuildersExpansion',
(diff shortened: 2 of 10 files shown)


Garanas:
I did not find the time to review this yet, @Hdt80bro do you have a spare hour or two?

Garanas (approved):

PR #4506, opened 2022-12-15
Clean up terrain types

Hdt80bro:
I'm not entirely happy with how this is, but I've been staring at it too long to feel like it'd be useful to do that any longer. The important bits are:
* Cleans up `/lua/TerrainTypes.lua` (particularly whitespace and empty tables)
* Annotates terrains types and the effect types they can contain
* Cleans up how `/lua/sim/Unit.lua` deals with terrain type effects--maybe...


lua/sim/Unit.lua
```diff
@@ -40,6 +40,13 @@ local UpdateAssistersConsumptionCats = categories.REPAIR - categories.INSIGNIFIC
 
 local DeprecatedWarnings = { }
 
+local DefaultTerrainType = GetTerrainType(-1, -1)
+local EmptyTable = {}
```
Garanas, line 44: Could you help me understand the need of this?

Garanas (approved):
Appears to cull quite a few tables - looking good. Just one comment to process.

lua/sim/Unit.lua
Hdt80bro: Instead of creating a new empty table every time an effect isn't found, it returns this same table. In order to preserve the sandboxed nature creating a new table would have, assignment to it is prohibited. Ideally, we could move this setup to the config file so that we could do this everywhere we're required to return an empty table instead of nil, but--baby steps.

Garanas: I see 👍 , I'll give it a test or two this week and merge it if it runs fine

Garanas (approved):

PR #4507, opened 2022-12-15
Structure terrain slope tilting

The-Balthazar:
Added the option for terrain slope tilting for structures that aren't getting rotated towards the enemy, and haven't flattened the terrain already.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -133,6 +133,67 @@ StructureUnit = Class(Unit) {
 
         if EntityCategoryContains(StructureUnitOnStartBeingBuiltRotateBuildings, self) then
             self:RotateTowardsEnemy()
+        elseif not bp.Physics.FlattenSkirt then
+            local bp = self.Blueprint
+            local function GetTerrainSlopeAngles(pos, box)
+                --Sanitise box into a table
+                if type(box) == "number" or type(box) == "nil" then
+                    box = {(box or 1)*0.5, (box or 1)*0.5}
+                --If we made it here, we already know its a table
+                elseif type(box[1]) == "number" and type(box[2]) == "number" then
+                    box = {box[1] * 0.5, box[2] * 0.5}
+                else
+                    return
+                end
+                --Get heights
+                local Heights = {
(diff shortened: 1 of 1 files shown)


Hdt80bro:
It looks like you plopped down some utility functions you needed inside `OnStartBeingBuilt()`--this makes them non-extensible, unreusable, and expensive. I recommend either adding them to the utilities library instead or incorporating their functionality into the function (removing the unused features). There's a couple of other concerns in them as well, but we *still* haven't gotten our guidelines updated, so that's not really fair to point out right now :)

Hdt80bro:
I'm curious, why go through two `Rotator`s instead of directly changing the unit orientation? Does that not properly update other stuff like collisions or meshes?

The-Balthazar:
I should put them in some form of utils file, I just wanted to get something up before I had to do life stuff. They're from my [terrain utils file from my mod](https://github.com/The-Balthazar/BrewLAN/blob/master/mods/BrewLAN/lua/TerrainUtils.lua). 

I didn't use set orientation because that involved using quaternians.

Hdt80bro:
Ah, gotcha. Someone has actually already solved that math for us, so if you'd prefer, you can transform your rotation using `EulerToQuaternion(roll, pitch, yaw)` first.

The-Balthazar:
I've got it working with `SetOrientation`, and while it does mean the collision lines up with the mesh and I don't have to deal with the origin being a non-standard orientation in the mesh, it does cause some graphical weirdness: The selection highlight gets weirdly offset as the incline increases, and my mod has a shielded extractor that has the bubble offset a different kind of weirdly, which indicates that there might also be other things that go a bit weird.

Hdt80bro:
Huh, that might mean that there's more bugs to fix then. Have you found anything else besides the selection box and attached entities causing problems solved by rotators?

The-Balthazar:
I did find that it's possible to skew the collision box with it, but that was with manually input quaternions `(1,1,1,1)`. I don't know how realistic that is an orientation, and I've been a bit busy the last few days.

Garanas:
I can't seem to be able to push back onto this branch 🤔 

image

I also do not entirely understand why it says `hdt80bro/fa.git`

Hdt80bro:
*Huh*

PR #4509, opened 2022-12-15
Add feature to snap to unit for observers / replays

Garanas:
Allows you to attach a camera position to a unit, such as an ACU. You can then quickly switch to and from the ACU. This isn't as useful in regular play, but for casters it can be a tool to quickly navigate between points of interest.

- [ ] make 'return to previous camera position' work
- [ ] allow up to four instances
- [ ] update all descriptions related to camera hotkeys
- [ ] LOC all descriptions related to camera hotkeys


engine/Core.lua
```diff
@@ -178,7 +178,7 @@ function GetMovieDuration(localFileName)
 end
 
 ---
----@param id UnitId
+---@param id UnitId | EntityId
```
Hdt80bro, line 181: I'm pretty sure `UnitId` is being used as the blueprint id string right now (which ends up confusing the name), so this should just `EntityId`. The last holdout looks like `GetUnitId`, which should also return an `EntityId` then.

Garanas: oh, I think you're right on that one

PR #4516, opened 2022-12-18
Add `/skirmish` arg to open skirmish

### LINKED ISSUE #4514: Run a game and open offline lobby with selected map
Add args for run a game right into the lobby. This is equivalent to the action: `run the game and click on "Skirmish" button`.

Desired arguments: `/offline`, `/selectedMap`

The `/map` argument is not suitable, as it run the match immediately. Some will need to configure the lobby (win conditions, mods, etc.)

Related with https://github.com/FAForever/downlords-faf-client/issues/2372

Marc-Spector:
Closes #4514

### DIFF
--- a/lua/ui/skirmish/skirmish.lua
+++ b/lua/ui/skirmish/skirmish.lua
@@ -1,0 +1,15 @@
+function openOfflineLobby(mapFolderName)
+    local playerName = import("/lua/user/prefs.lua").GetCurrentProfile().Name or "Unknown"
+    local scenarioFilePath = creteScenarioFilePath(mapFolderName)
+    local lobby = import("/lua/ui/lobby/lobby.lua")
+    lobby.CreateLobby('None', 0, playerName, nil, nil, GetFrame(0), function() ExitApplication() end)
+    lobby.HostGame(playerName .. "'s Skirmish", scenarioFilePath, true)
+end
+
+function creteScenarioFilePath(mapFolderName)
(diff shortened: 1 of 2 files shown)


lua/ui/skirmish/skirmish.lua
```diff
@@ -0,0 +1,29 @@
...
+end
+
+local function parseMapFolderNameFromArgs()
+    local mapFolderName
+    local mapFolderName = GetCommandLineArg("/skirmish", 1)
+    if not mapFolderName then
+        error("set map folder name for /skirmish arg")
+    end
+    return mapFolderName[1]
+end
```
Sheikah45, line 16: Isn't there already a /map command? Why duplicate it instead of just making this a boolean command and using both?

Marc-Spector: Please read it https://github.com/FAForever/downlords-faf-client/pull/2873/files#diff-3d47503b30b47f5a259049a2e45e1eb555b502bd1f9cc68c936394a0d8fc6f1fR51-R52

I cannot change behavior of /map arg. The arg is not exit in any lua files

Sheikah45: Ah gotcha, I missed that. And I guess since it is offline mode you don't want to use the ice adapter and the host game method that we currently use to set the map in the lobby correct?

Although since we introduced the gameoption flag you could set the scenario file directly using that possibly

PR #4517, opened 2022-12-18
Projectiles Fil Clean-up

MrRowey:
Clean Up of the projectile files removing unused commented code first


projectiles/AANDepthCharge01/AANDepthCharge01_script.lua
```diff
@@ -64,7 +63,6 @@ AANDepthCharge01 = Class(ADepthChargeProjectile) {
     end,
 
     OnImpact = function(self, TargetType, TargetEntity)
-        --LOG('Projectile impacted with: ' .. TargetType)
```
Garanas, line 67: For this particular file, you can also refactor the VizMarker into VizMarkerOpti, you can search for examples. The opti version allocates as few tables as possible.

projectiles/AANDepthCharge03/AANDepthCharge03_script.lua
```diff
@@ -62,7 +54,6 @@ AANDepthCharge03 = Class(ADepthChargeProjectile) {
     end,
 
     OnImpact = function(self, TargetType, TargetEntity)
-        --LOG('Projectile impacted with: ' .. TargetType)
```
Garanas, line 65: Same as the previous comment.

projectiles/AANDepthCharge01/AANDepthCharge01_script.lua
```diff
@@ -1,6 +1,5 @@
---
 -- Depth Charge Script
---
```
Garanas, line 3: One could also wonder what the practical value is of `EnterWaterMovementThread = function(self)` for this and similar projectiles, is it worth waiting 1 tick (which is - fun fact - the same tick, but at the very end) and create a separate thread for that all together?

It probably is not

projectiles/AANTorpedoCluster01/AANTorpedoCluster01_script.lua
```diff
@@ -41,25 +36,22 @@ AANTorpedoCluster01 = Class(ATorpedoCluster) {
...
             proj:PassDamageData(self.DamageData)
-        end            
-        
+        end
+
         local pos = self:GetPosition()
         local spec = CachedSpecifications
         spec.X = pos[1] 
         spec.Z = pos[3]
-        spec.Army = self.Army
```
Garanas, line 55: Sounds like I touched this file before. But the same applies: you can refactor the vision marker into the opti version

projectiles/AANTorpedoClusterSplit01/AANTorpedoClusterSplit01_script.lua
```diff
@@ -42,14 +36,10 @@ AANTorpedoCluster01 = Class(ATorpedoCluster) {
         end
         self:ForkThread(self.EnterWaterMovementThread)
     end,
-
```
Garanas, line 45: Same here, is the enter water movement thread _really_ doing anything? It doesn't even contain a wait statement!

projectiles/AIFFragmentationSensorShell02/AIFFragmentationSensorShell02_script.lua
```diff
@@ -1,48 +1,42 @@
...
 local RandomFloat = import("/lua/utilities.lua").GetRandomFloat
 
 AIFFragmentationSensorShell02 = Class(AArtilleryFragmentationSensorShellProjectile) {
-               
     OnImpact = function(self, TargetType, TargetEntity) 
         if TargetType != 'Shield' then
 	        local FxFragEffect = EffectTemplate.Aeon_QuanticClusterFrag02 
             local bp = self:GetBlueprint().Physics
-	        
+
```
Garanas, line 16: You can always replace `self:GetBlueprint()` with `self.Blueprint` for projectiles

projectiles/AIMFlare01/AIMFlare01_script.lua
```diff
@@ -1,22 +1,15 @@
...
+------------------------------------------------------------------------
 local Flare = import("/lua/defaultantiprojectile.lua").Flare
 local AIMFlareProjectile = import("/lua/aeonprojectiles.lua").AIMFlareProjectile
 
 AIMFlare01 = Class(AIMFlareProjectile) {
-
     OnCreate = function(self)
         AIMFlareProjectile.OnCreate(self)
-        --LOG('*DEBUG: FLARE CREATED')
-        --self:SetCollisionShape('Sphere', 0, 0, 0, self:GetBlueprint().Physics.FlareRadius)
```
Garanas, line 18: Same here (a few lines below) - you can factor out the `GetBlueprint`

Garanas (commented):
A lot more is possible then you're doing now. This particular sweep isn't so much about cleaning up the files, but also looking at the logic and wondering: do we really need this _here_?

For example: all the OnEnterWater functions create splash effects. If they all do this, why don't we do it in the base class for example?

A few other valid improvements:
- using the opti vision marker, located in the same file
- reducing table entries
- reducing engine calls (`self:GetArmy` -> `self.Army`, `self:GetBlueprint` -> `self.Blueprint` , `self:GetLauncher` -> `self.Launcher`)
- replace `WaitSeconds` with `WaitTicks`, but with corrected values (usually * 10, except for `WaitSeconds(0.1)` which is `WaitTicks(2)` , see also the definition of `WaitSeconds` in `simInit.lua`)
- and a lot more, if you closely look at the hierarchy.

And like that, a lot more is possible then you're doing right now. Not that what you're doing is bad, but it feels to me a lot more can be done.

projectiles/AANDepthCharge01/AANDepthCharge01_script.lua
MrRowey: Ive searched for VisMarkerOpti and cant seem to see a Example is this somthing that in a upcoming pr ?

Garanas: Yes, that seems about right 👍

projectiles/AANDepthCharge01/AANDepthCharge01_script.lua
```diff
@@ -1,84 +1,54 @@
...
+    FxEnterWater= {
+        '/effects/emitters/water_splash_ripples_ring_01_emit.bp',
+        '/effects/emitters/water_splash_plume_01_emit.bp',
+    },
 
     OnCreate = function(self)
         ADepthChargeProjectile.OnCreate(self)
         self.HasImpacted = false
-        self:ForkThread(self.CountdownExplosion)
+        self.ForkThread(self.CountdownExplosion)
```
Garanas, line 17: This is not a safe change on its own - see also:
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Projectile.lua#L885-L893

You'll also need to add it to the trash bag

projectiles/AANDepthCharge01/AANDepthCharge01_script.lua
```diff
@@ -1,84 +1,54 @@
...
     end,
 
     CountdownExplosion = function(self)
-        WaitSeconds(self.CountdownLength)
-
+        WaitTicks(self.CountdownLengthInTicks)
         if not self.HasImpacted then
-            self:OnImpact('Underwater', nil)
+            self.OnImpact('Underwater', nil)
```
Garanas, line 23: There is no need to change this. Note that this change is also not logically equivalent. Please revert all changes of the same type (`:` -> `.`, back to `:`)

Garanas (changes requested):
The two conversations left need to be looked at before we can merge this.

projectiles/AANDepthCharge01/AANDepthCharge01_script.lua
MrRowey: I've just revert back to : from . for now 

I'm assuming the change would be something like

```lua
self.Trash:ForkThread(self.CountdownExplosion)
```

Garanas: No, it would be:
-  `self.Trash:Add(ForkThread(self.CountdownExplosion, self))`

when you use `:` what actually happens is that the argument before `:` (which is `self` in this context) is appended as the first argument to whatever you're calling after the `:`

Garanas (approved):

PR #4521, opened 2022-12-21
aeon navy factories

MadMaxFAF:
still got the lod 1's to do


Garanas:
image

I feel like we can do better - this loses the 'Aeon touch' that other Aeon units have: the white plating. What if we make the support factory smaller? I understand that it messes with the support arms, but I'm sure we can deal with that too.

MadMaxFAF:
image
image
image

Garanas (approved):
Must have been the most difficult unit to tackle - but the results are amazing. Great work

PR #4523, opened 2022-12-21
Introduces function to determine allocated bytes for any value

Garanas:
This is useful to understand the impact of tables (or values in general) - a long missing factor in our hunt towards better performance. Looking forward to feedback on this function.

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -96,45 +96,6 @@ SyncMeta = {
     end,
 }
 
---- Determines the size in bytes of the given element
----@param element any
----@param ignore table<string, boolean>     # List of key names to ignore of all (referenced) tables
----@param seen? table<table, any>           # List of strings and / or tables that we've seen so far. Should be nil by default
----@return integer
-function ToBytes(element, ignore, seen)
-
-    -- has no allocated bytes
-    if element == nil then
-        return 0
-    end
-
-    -- applies to tables and strings, to prevent counting them multiple times
-    local seen = seen or { }
-    if seen[element] then
-        return 0
-    end
-
-    -- determine size
-    local allocatedSize = debug.allocatedsize(element)
-    if allocatedSize == 0 then
-        return 8
-    elseif type(element) ~= 'table' then
-        seen[element] = true
-        return allocatedSize
-    end
-
-    -- at this point we know it is a table
-    seen[element] = true
-    for k, v in element do
-        if not ignore[k] then
-            allocatedSize = allocatedSize + ToBytes(v, ignore, seen)
-        end
-    end
-
-    return allocatedSize
-end
-
-
 local cUnit = moho.unit_methods
 ---@class Unit : moho.unit_methods
 ---@field Brain AIBrain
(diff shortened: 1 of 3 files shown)


lua/sim/Unit.lua
```diff
@@ -96,6 +96,45 @@ SyncMeta = {
...
+---@return integer
+function ToBytes(element, ignore, seen)
+
+    -- has no allocated bytes
+    if element == nil then
+        return 0
+    end
+
+    -- applies to tables and strings, to prevent counting them multiple times
+    local seen = seen or { }
```
4z0t, line 112: It is local already

lua/sim/Unit.lua
```diff
@@ -216,6 +255,14 @@ Unit = Class(moho.unit_methods) {
     OnCreate = function(self)
         local bp = self:GetBlueprint()
 
+        ForkThread(
+            function()
+                WaitSeconds(1.0)
+                LOG(ToBytes(self, { Blueprint = true, Brain = true, Audio = true, MovementEffects = true }))
+                LOG(ToBytes(__blueprints))
+            end
+        )
```
4z0t, line 264: Test code.

lua/sim/Unit.lua
```diff
@@ -96,6 +96,45 @@ SyncMeta = {
...
+    -- at this point we know it is a table
+    seen[element] = true
+    for k, v in element do
+        if not ignore[k] then
+            allocatedSize = allocatedSize + ToBytes(v, ignore, seen)
+        end
+    end
+
+    return allocatedSize
+end
```
4z0t, line 135: Duplicate

4z0t (changes requested):

lua/sim/Unit.lua
Garanas: Test code too 😄

lua/sim/Unit.lua
Garanas: Parameters are, I'll adjust it

Garanas:
Fun fact with the navigational mesh, on the map `Syrtis Major` the entire datastructure takes up about 33 megabytes:

image

On Seton's Clutch this is about 48 megabytes. This means that the compression we do works quite well.

About 30% of the memory that we use originates from these lines:
- https://github.com/FAForever/fa/blob/feature/bytes-function/lua/sim/NavGenerator.lua#L652-L670

If we remove those lookup tables:
- Syrtis Major: 33mb -> 21mb
- Seton's Clutch: 48mb -> 30mb

Oof 😄

KionX:
This should exec only for tables: https://github.com/FAForever/fa/pull/4523/files#diff-b7cb9a852b1ea5bdbe28863255f535c0924f4446962075087db5bbe9a9b71314R54

Garanas:
As far as I can see it only executes that for tables a this moment

KionX:
if type(v) == 'table' then

Garanas:
I'm all up for an alternative if it is more accurate!

As far as I can tell, the recursion only applies to tables at the moment, unless you're making a distinction between the hash and array sections of a table.

PR #4524, opened 2022-12-22
Adds default repeat build option for factories

### LINKED ISSUE #4346: Feature Request: Repeat Build is on for new factories toggle.
Implement the UI party feature that toggles on repeat build on every factory build as an option, like right clicking on T2 mexes and highers automattically rings it with mass storages.

When you build a new factory --> The repeat build will be toggled on by default, it does not change behavior and can be toggled manually off after, but just makes every new factory start with it.

Example of UI party working in game.
ezgif-2-fabbee3e1c

I've noticed some bugs on my game with UI Party recently and would like to remove it, but this feature is too much important for me to remove it, so I'm hoping for it to no longer be necessary for me.

haifron:
Closes #4346

### New logic:
- When you first select a factory, it is toggled to repeat build if the option is enabled.
- This is controlled by `Options > Interface > Building > Factories Default Repeat Build`. It is `false` by default.

This implementation is smoother than the UI-Party implimentation since it doesn't require changing the user's selection to make the change. To achieve that I've made the assumption there's no way to give a queue to a factory before selecting it.

### Extra information / cheaper implementation
This implementation does loop through the new units in the selection to check for factories, which is similar to a loop in `"rallypoint.lua".OnSelectionChanged()` so it might be cheaper if just implemented there, but I didn't want to add it there initially since the functionality is totally irrelevant to `rallypoint.lua` so it feels a tad spaghetti-y. I would like to know what you guys think.

I would've much rather only executed this once on factory completion, rather than check for it on each new selection,, but I don't know of any way to access `UserUnit.ProcessInfo`  from the `OnComplete` functions, which usually deal with the `Unit` class.

<details><summary>

##### if implemented in `rallypoint.lua` this just adds an `if` in a loop

</summary>

  ```
function OnSelectionChanged(selection)
     if not beatFunctionAdded then
         import("/lua/ui/game/gamemain.lua").AddBeatFunction(OnBeat)
         beatFunctionAdded = true
     end

     ClearAllRallyPoints()

     local factories = EntityCategoryFilterDown(categories.STRUCTURE * categories.FACTORY, selection)
     if factories then
        for _, factory in factories do
            AddRallyPoint(factory)

+         if not factory.HasBeenSelected then
+             factory:ProcessInfo('SetRepeatQueue','true')
+             factory.HasBeenSelected = true
+         end

        end
     end
end

  ```
</details>

### Testing setup
With the option enabled/disabled, build a factory. When you select the factory it should be set to repeat build if the option is enabled, or not if the option is disabled.


Garanas:
I don't have time to go into details now, but there's a separation between the sim and the UI. Any field (`HasBeenSelected`) you set in the sim is not available in the UI. The reason this works is because this statement:
- https://github.com/FAForever/fa/pull/4524/files#diff-4d24f0ee6330335ec1aa4beab16e827c5930fc86fa6ce5cbf5aaf85471412e32R664

Returns true whether the value is `false` or `nil`, and in this case it is `nil` because you're using the value in the ui layer, which is oblivious of the value in the sim layer.

See also:
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Sim/Unit.lua
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/User/UserUnit.lua

haifron:
Ah ok that explains a lot of the confusion I was having 😛 thanks for your explanation. I've therefore removed the changes to `/lua/defaultunits.lua` since they were unnecessary. I'll just believe it was thanks to my genius that it worked anyway.

Do you have an opinion on whether the added `gamemain.lua` code should stay there, or be moved to `rallypoint.lua`?

###### That force-push was an accident and I've undone it 🙃

lua/ui/game/gamemain.lua
```diff
@@ -657,6 +657,18 @@ function OnSelectionChanged(oldSelection, newSelection, added, removed)
     if not isOldSelection then
         import("/lua/ui/game/selection.lua").PlaySelectionSound(added)
         import("/lua/ui/game/rallypoint.lua").OnSelectionChanged(newSelection)
+        if options.repeatbuild then
```
Garanas, line 660: Best to do this like this:
- `Prefs.GetFromCurrentProfile('options.cursor_splash_damage') == 'on'`

But then with the correct values of course. That way if a player changes the preferences they don't need to restart the game.

lua/ui/game/gamemain.lua
```diff
@@ -657,6 +657,18 @@ function OnSelectionChanged(oldSelection, newSelection, added, removed)
...
+            local factories = EntityCategoryFilterDown(categories.STRUCTURE * categories.FACTORY, added) -- find all newly selected factories
+            if not table.empty(factories) then
```
Garanas, line 662: There's no need to check for this - if it empty then the for loop will simply not iterate.

lua/ui/game/gamemain.lua
```diff
@@ -657,6 +657,18 @@ function OnSelectionChanged(oldSelection, newSelection, added, removed)
     if not isOldSelection then
         import("/lua/ui/game/selection.lua").PlaySelectionSound(added)
         import("/lua/ui/game/rallypoint.lua").OnSelectionChanged(newSelection)
+        if options.repeatbuild then
+            local factories = EntityCategoryFilterDown(categories.STRUCTURE * categories.FACTORY, added) -- find all newly selected factories
```
Garanas, line 661: Is there a reason why you use the `added` parameter and not the `newSelection` parameter?

Garanas (changes requested):

lua/ui/game/gamemain.lua
haifron: Oh perfect, I was wondering how to get that behaviour.

lua/ui/game/gamemain.lua
haifron: Yes, `added` = `newSelection` - `oldSelection` essentially. Since we only need to check for factories that have *never* been selected before, looking at anything that was in `oldSelection` is totally redundant - we literally just selected them.

This at worst saves no time (nothing was reselected, i.e `newSelection` == `added`), and at best saves a lot of time (a lot of units were reselected, i.e `newSelection` ≫ `added`).

Garanas: I agree - good call

lua/ui/game/gamemain.lua
```diff
@@ -657,6 +657,16 @@ function OnSelectionChanged(oldSelection, newSelection, added, removed)
     if not isOldSelection then
         import("/lua/ui/game/selection.lua").PlaySelectionSound(added)
         import("/lua/ui/game/rallypoint.lua").OnSelectionChanged(newSelection)
+        if Prefs.GetFromCurrentProfile('repeatbuild') then
```
Garanas, line 660: needs to be `options.repeatbuild`

Garanas:
We need to look into how we want to manage code that works on a selection change in general. This solution is fine for now, but we may put it somewhere else in the (near) future. I just don't know where to put it yet.

There's a lot of functionality that works with the current selection in some fashion.

lua/ui/game/gamemain.lua
haifron: B) fixed, cheers

lua/options/options.lua
```diff
@@ -831,6 +831,18 @@ options = {
                   inc = 100,
                 },
             },
+            {
+                title = "<LOC OPTIONS_0287>Factories Default Repeat Build",
```
Garanas, line 835: I'd argue that this is a gameplay feature and not an interface feature, as it immediately impacts gameplay. Perhaps put it below the `Commands` header?

Garanas (approved):

lua/options/options.lua
haifron: Sure 👍

Garanas:
Played a few games and I can't play without this anymore - great work @haifron

Issue #4324, opened 2022-11-02
Change AI multiplier types and defaults


relent0r:
Note : I would target the Feb 2023 release for these changes

**Describe the feature**
Its long been considered that the cheat multiplier by default is too high, and that having a separate cheat and build multiplier that are adjustable have negative impacts on the AI.

After discussion with another AI dev, they'd like the buildmult and cheatmult to remain separate and I can see their pov. But I'll leave the description here.

~~I would like to suggest that we make some changes to this.
First change is to remove having the two separate options and just have a single one that does both.
Instead of having a lobby option for BuildMult and CheatMult, we simply have CheatMult.~~

~~My understanding is that the changes required are relatively low. In the SetupCheat function we could just setup the buffAffects.BuildRate.Mult to use the CheatMult configuration item. Then some other changes in regards to build conditions etc.~~

~~Doing this would resolve confusion around setting up the AI with lobby options. It is somewhat unknown that adjusting one without adjusting the other will cause AI to perform badly.~~
~~e.g if build is above cheat then it will crash its economy, like wise if cheat is above build it will overflow its economy.~~

~~The only times configuring cheat over build would benefit the AI is when its badly tuned to crash its economy easily. 
LOUD has taken this approach and found user life is happier for it.~~

The second change is the lower the default value for the multiplier from 2.0 to 1.5 (or perhaps lower I'm not sure). We see many users pose questions about the AI where they don't know the lobby options exist so are stuck either playing against an AI that no one below a 1500 rating on specific maps is going to beat or against a non cheating AI that they roll over. The AI player base is a silent one so I imagine there are many others that hit this and just don't communicate it with the community while getting subpar experience.

The 3rd change is (this is pending more AI work that will improve AI intel gathering) that we disable the OMNI cheat by default(option IntelCheat).

The expectation is that the current AIX is setup the way it is because of the severe short comings in the AI's capability when FA was released, with the planned changes coming to the AI over the coming months I believe these defaults currently are and will be more detrimental to the non serious users player experience. Experienced players will know that they can adjust lobby options to provide a more challenging experience should they desire it.

Garanas:
This sounds good. Would this involve adjusting the lobby options, or would it also involve adjusting logic surrounding the AI?

The lobby options are defined here:
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/lobby/lobbyOptions.lua

relent0r:
[...]

PR #4525, opened 2022-12-22
Allow negative aix cheat multipliers

ANIALLATOR114:
### Description of changes:
I saw this mentioned here but couldn't find an open issue/pr for the change. https://github.com/FAForever/fa/issues/4324#issuecomment-1304788478

This PR changes the range of selectable aix cheat values to include negative multipliers. 0.5->0.9.

This also required a change to the default selection value as 5 entries are being added to the start of the list. The same default of 1.5 is retained.

This will allow players to choose a harder cheating aix in terms of resources, but lowering their build speed to allow more time for the player to react. Alternatively a player could restrict an aix's resources but retain its build speed.

### Testing:
Open a lobby and observe the default value has retained at 1.5 and now values less than 1.0 are selectable.
image
image

Open a skirmish match with aix and observe their resource production and build speed match the value set in the lobby. 
I set 0.5 for both in this example.
image

https://user-images.githubusercontent.com/116189545/209223565-782f6e79-05e2-4ce2-8264-c7c6fe2866a5.mp4

ps - This is my first contribution to FAF so please let me know if I'm missing anything here!

### DIFF
--- a/lua/ui/lobby/lobbyOptions.lua
+++ b/lua/ui/lobby/lobbyOptions.lua
@@ -595,26 +595,28 @@ globalOpts = {
 ---@type ScenarioOption[]
 AIOpts = {
    {
-        default = 6,
+        default = 11,
         label = "<LOC aisettings_0001>>AIx Cheat Multiplier",
         help = "<LOC aisettings_0002>Set the cheat multiplier for the cheating AIs.",
         key = 'CheatMult',
         value_text = "%s",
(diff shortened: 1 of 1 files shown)


Garanas:
Looking good! I'll give it a closer look this weekend

Garanas (approved):

Garanas:
@ANIALLATOR114 one minor not: you can use 'closing keywords' to automatically close the issue you're trying to solve when the pull request is merged. As an example: closes #4324

PR #4526, opened 2022-12-23
Fix Techlevel-Categories discrepancies

haifron:
## Description of the changes
There were a few bps whose `bp.General.TechLevel` didn't match the tech level in `bp.Categories`. With this PR all of the main game bps should have the proper tech levels in `bp.General.TechLevel` (unless both the `Categories` tech level & the `General` tech level were wrong already 🙃)

## Extra info
Tech1 == Basic
Tech2 == Advanced
Tech3 == Secret
Experimental == Experimental


Garanas:
I wasn't aware that this is the relationship between the tech level and the tech. Where did you learn about this?

haifron:
I don't remember honestly, I have seen it mentioned somewhere though 🤔. Either way this holds true for *every* blueprint - _except_:
- the ones in this PR
- the ACU blueprints which have `RULEUTL_EXPERIMENTAL` in `bp.General.TechLevel` but no tech level in `bp.Categories`.

i.e with the exceptions of the ACU blueprints, I can say every main game blueprint has:
- one of `TECH1`, `TECH2`, `TECH3`, `EXPERIMENTAL` in `bp.Categories`
- one of `RULEUTL_...` `Basic`, `Advanced`, `Secret`, `Experimental` in `bp.General.TechLevel`

and every one besides the 9 changed here have these two matching according to what I specified in the OP.

So it seems almost certain these few outliers are mistakes.

Garanas:
@Hdt80bro could you look into the exe to see if this value is referenced? Just out of curiosity, to see if it has any impact.

Hdt80bro:
It doesn't look like `TechLevel` is used by the engine (nor its `RULEUTL_*` enums--hence why they appear [here](https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Core/Blueprints/UnitBlueprint.lua#L3) instead), only for unused test dump purposes in Lua. The tech categories definitely are used though.

PR #4538, opened 2022-12-27
Add table.random

4z0t:
Adds table.random function returning random entry from it. Replaces code duplication.

### DIFF
--- a/lua/AI/AIBehaviors.lua
+++ b/lua/AI/AIBehaviors.lua
@@ -979,7 +979,7 @@ function FatBoyBuildCheck(self)
 
     -- Randomly build T3 MMLs, siege bots, and percivals.
     local buildUnits = {'uel0303', 'xel0305', 'xel0306', }
-    local unitToBuild = buildUnits[Random(1, table.getn(buildUnits))]
+    local unitToBuild = table.random(buildUnits)
 
     aiBrain:BuildUnit(experimental, unitToBuild, 1)
     WaitTicks(1)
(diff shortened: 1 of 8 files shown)


Hdt80bro:
Ahh, perfect, I've been wanting to do something like this for a while! You missed a bunch of places you could have used this, so I'll commit those in a bit (though I'll leave the supposedly performance-critical places alone, as well as the ones with individually hard-coded size ranges). There are several places that would benefit from a map-key version or a clamped version as well, though those are better left for another time :)

I was a little confused at first about what this function did until I looked at it, so I'd go the route that `GetRandomEntry` did to make it less confusing (doing that also removes the overhead of the table access to `table`--luckily, adding this function doesn't increase the size of `table` to the next power of two, so the access time itself won't be more expensive). This is the pattern the string utility methods follow as well.

4z0t:
`table.getrandomentry`? `table.getrandom`?`table.randomentry`?

Hdt80bro:
If you want to inject it into the table library like we've been doing thus far, I like `table.randomentry` best, but that doesn't make it clear that it only works for arrays though (I also usually see the term "entry" to mean an entire key-value pair of a map), because we've since bloated `table` for use with associative tables as well. However, at this point in development there's not quite a good solution, so it'll have to do. With how we've been treating the global table, I'd still prefer it to be a proper utility function like `GetRandomTableEntry` (in addition to the reasons above), though trying to bring order to decades of disorganized development is likely a losing battle at this point.

Garanas:
It seems to me that this is fine for a first version, I like the simplification of code. Usually random elements are taken over the array section (effects, for example). I do not recall one example where it also takes into account the hash part

Hdt80bro:
The only one I saw was picking a random faction

Garanas (approved):

PR #4552, opened 2023-01-04
Weapons

MrRowey:


lua/aeonweapons.lua
```diff
@@ -537,8 +530,8 @@ AANDepthChargeBombWeapon = ClassWeapon(DefaultProjectileWeapon) {
         }
 
         if proj and not proj:BeenDestroyed() then
-            proj:PassDamageData(damageTable)
```
Garanas, line 540: This substitution is not logically equivalent! It would be:
`proj:PassDamageData(damageTable)` `->` ` proj.PassDamageData(proj, damageTable)`

But we've learned that it is not faster to do that - as this is an internal Lua functionality, unlike the class implementations of Forkthread. Best to revert these changes.

lua/cybranweapons.lua
```diff
@@ -354,7 +340,7 @@ CIFNaniteTorpedoWeapon = ClassWeapon(DefaultProjectileWeapon) {
     CreateProjectileForWeapon = function(self, bone)
         local proj = self:CreateProjectile(bone)
         local damageTable = self:GetDamageTable()
-        local bp = self:GetBlueprint()
+        local bp = self.Blueprint
         local data = {
```
Garanas, line 344: Could you investigate what this data is used for?

Garanas (changes requested):
Bring back the `:` for those projectile function calls and I think we're good. I'll look at it again this evening.

lua/cybranweapons.lua
MrRowey: Will Lookin to via Seperat PR

lua/aeonweapons.lua
```diff
@@ -491,8 +484,8 @@ AIFArtilleryMiasmaShellWeapon = ClassWeapon(DefaultProjectileWeapon) {
...
-            proj:PassData(data)
+            proj.PassDamageData(damageTable)
```
Garanas, line 487: missed these

lua/terranweapons.lua
```diff
@@ -102,15 +100,12 @@ TDFIonizedPlasmaCannon = ClassWeapon(DefaultProjectileWeapon) {
 ---@class TDFHiroPlasmaCannon : DefaultBeamWeapon
 TDFHiroPlasmaCannon = ClassWeapon(DefaultBeamWeapon) {
     BeamType = CollisionBeams.TDFHiroCollisionBeam,
-    FxMuzzleFlash = {},
```
Garanas, line 105: Removing `FxMuzzleFlash` will change the behavior, and give the weapon the default muzzle as defined by the `DefaultProjectileWeapon` class

Garanas (changes requested):
You've removed some empty tables, but not all of them guarantee the same behavior. Some are empty to override the base effects, usually of the `DefaultProjectileWeapon` class.

Garanas (approved):

PR #4555, opened 2023-01-05
Improve the retrieval of the (nearest) valid leaf of a position

Garanas:
Units tend to get themselves to edge cases quite quickly. As a result a query for a path can easily end up failing because the origin location is not 'valid', even though it technically is valid.

Specifically, this pull requests adds in:

- (1) Neighbors for invalid leaves
- (2) Special utility functions to retrieve grid and leaf
- (3) Further reduction of navigational mesh resolution

In case of (1), a leaf that is invalid does not look for corner elements. As an example, it only searches for:

```text
0|x|0
x| |x
0|x|0
```

While the corners are not considered valid.

### DIFF
--- a/lua/shared/NavGenerator.lua
+++ b/lua/shared/NavGenerator.lua
@@ -98,6 +98,10 @@ end
 ---@param label number
 ---@return string
 function LabelToColor(label)
+    if label == -1 then
+        return 'ff0000'
+    end
+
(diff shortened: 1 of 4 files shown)


Garanas:
This can be tested using the navigational mesh UI, specifically a location such as this is now considered valid:

image

Garanas:
@maudlin27 this should fix the issue of edge cases (near ramps / cliffs) not having a label assigned

PR #4565, opened 2023-01-08
Naval adjustments

Tagada14:
Aeon Frigate:
Decrease cost by 10 mass
Increase range by 2

Aeon Destroyer
Decrease HP by 300
Decrease range by 10

Cybran Frigate:
Increase mass cost by 10
Lower Range by 2

Cybran Destroyer:
Increase HP by 450

UEF Battleship:
Increase ms by 0.25
Increase Turn Rate by 2.5

Tempest:
Increase mass cost by 2k
Change BT/mass ratio to 50% that of BS

For a long time, Aeon frigates were absolute garbage while their Destro's were insanely good while Cybran had the exact opposite with OP frigate and weak destroyer. These changes redistribute the power of those factions while retaining the units' unique traits and further enhancing them by adjusting the ranges of the frigate. We hope it will also introduce more micro potential in frigate fights, especially with Cybran vs Aeon.

With the changes to Bulwark which should move its usage more to T2 rather than T3 stage UEF Battleships are left out to dry. We are making them slightly less clunky so that hopefully they can dodge at least some shot if you micro them.

Tempest has been very powerful for a long time and it's getting a well-deserved nerf. You will need more BP/infrastructure ready to build one instead of it acting like a possible quick mass dump after spamming destroyers and skipping T3 Navy completely.


ComradeStryker:
What of the Atlantis?
Would it be tackled later on with the more detailed navy rebalance?

---

~ Stryker

PR #4567, opened 2023-01-08
Adjust engineers

Tagada14:
Increase BP of T2 and T3 engineers to increase their efficiency compared to T1 engies.
T2 12.5 -> 13 (10.4 -> 10 m/BP)
T3 30->32.5 (10.4 -> 9.6 m/BP)


ComradeStryker:
Sparky was left out?
Why not increase its BP (by 0.5 to even it out to 18) as well?

---

~ Stryker

Azraeel:
I appreciate this balance change so much its unreal

PR #4570, opened 2023-01-08
Adjust pathing parameters of transports to improve (un)loading

Garanas:
Adjusts the various air parameters of transports to make them more responsive (1), more 'alive' (2) fix a few minor stats hiccups (3) and allow submarines to hit transports that landed on water (4)

(1) was accomplished by allowing transports accelerate and turn faster, while reducing the average elevation of the transport and their hover height. During testing they now consistently pick up and drop units. A side effect is that they are easier to micro, especially relevant early game against that first interceptor.

(2) is accomplished by allowing transports to 'bank forwards' as they move around. This appears to be purely a visual improvement, especially when they land.

(3) some transports had an elevation of 8, while others had an elevation of 15. It is a lot easier for anti air to hit a unit at an elevation of 8, then it is to hit a unit at an elevation of 15. They now all have:

- Tech 1: elevation of 8
- Tech 2: elevation of 9
- Tech 3: elevation of 10

(4) some transports had a positive collision extent in the Y direction. As a consequence, their collision box was above the water when lading on water. This made it impossible for submarines to hit the transport.


Garanas:
I'd like to note that I also think that the maximum speed of transports is quite excessive:

- Tech 1 has 10
- Tech 2 has 14.3
- Tech 3 has 15

I'd argue that these numbers would be more legit:

- Tech 1 has 10
- Tech 2 has 11
- Tech 3 has 14

As a transport is - one way or another - loaded with units and therefore carries a significant weight. The reduction of air speed would also act as a nerf for (1), where transports would be more maneuverable.

PR #4571, opened 2023-01-08
Adjust pathing parameters for engineers

Garanas:
Fixes various hiccups for engineers, including bad targeting bones (1), allow all engineers to move backwards and for a reasonable distance (2), improve pathing of engineers by making them accelerate / brake / turn faster (3) and all hover units should wobble slightly (4)

The entire point of this pull request is to reduce the 'bad pathing' that you end up having due to a cluttered base and engineers constantly bumping into things.

(1) was accomplished by simply removing the targeting bones. When there are no targeting bones the center of the collision extent (without taking into account any offsets) is used. This is better because:

- The targeting bones of the Seraphim engineers were lower than the center of the engineer. As a result, they are more difficult to hit for low arc / beam weapons when the terrain is bumpy
- The targeting bones of the Aeon engineers were also part of an animation, which can cause a hiccup when a weapon is aiming, causing it to overshoot as the animation plays

(2) guarantees that the behavior is the same for all engineers. It also allows engineers to move backwards a reasonable distance, such that even for a factory (or similar large structures) the engineer would prefer to move backward if able instead of turning around

(3) combined with (2) means the build site is cleared a lot quicker as not only there are fewer hiccups in terms of pathing, they are also resolved faster because engineers can accelerate / brake a lot faster

 (4) is purely aesthetical, where the hover engineers now wobble slightly 😄 
 
This will change how an engineer feels and reacts. From a competitive perspective the engineer can evade projectiles a lot easier. From a practical perspective the engineers just feel a lot less clunky. In my opinion the new statistics also fit the nature of a construction unit: it should be more nimble than the average unit.


Tagada14:
When I played around with the engineers today I came to similar numbers to achieve a good feeling while not making them too fast. Overall this is exactly what the Balance team wanted to achieve so after a bit of testing I will merge this.
Great work Jip!

PR #4572, opened 2023-01-08
Make stationary stealth fields more useful

Garanas:
Stationary stealth fields is one of those units that I never see around usually. In this pull request we give them a buff (1) , and on top of that we also add in some faction-specific flavors (2).

(1) All stealth fields have a larger stealth radius. They no longer provide sonar stealth, as that doesn't make sense at all. I was also unaware of them providing it. See the changes for the exact stats.

(2) All stealth fields could use a bit of a 'faction flavor', therefore:

- Aeon: similar to the ACU radar upgrade, the stealth field provides a large vision radius
- UEF: similar to other jamming units, the stealth field provides jamming blips. Unintentionally but quite nice, these jamming blips do not look like units - they look like structures 😄 
- Cybran: a larger stealth field than all the other factions
- Seraphim: a jamming crystal behavior, where instead of providing stationary blips they'd move around just like the jamming crystal. 

The jamming crystal one is not part of this pull request - yet. I can implement that when it is accepted.

### DIFF
--- a/units/UEB4203/UEB4203_unit.bp
+++ b/units/UEB4203/UEB4203_unit.bp
@@ -140,14 +140,9 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        RadarStealthFieldRadius = 32,
+        RadarStealthFieldRadius = 26,
         ReactivateTime = 5,
-        VisionRadius = 20,
-        JamRadius = {
-            Max = 30,
-            Min = 0,
-        },
-        JammerBlips = 6,
+        VisionRadius = 22,
     },
     Interface = {
         HelpText = '<LOC ueb4203_help>Stealth Field Generator',
--- a/units/UEB4203/UEB4203_unit.bp
+++ b/units/UEB4203/UEB4203_unit.bp
@@ -140,7 +140,7 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        RadarStealthFieldRadius = 26,
+        RadarStealthFieldRadius = 28,
         ReactivateTime = 5,
         VisionRadius = 22,
     },
--- a/units/UEB4203/UEB4203_unit.bp
+++ b/units/UEB4203/UEB4203_unit.bp
@@ -140,10 +140,14 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        RadarStealthFieldRadius = 24,
+        RadarStealthFieldRadius = 32,
         ReactivateTime = 5,
-        SonarStealthFieldRadius = 24,
         VisionRadius = 20,
+        JamRadius = {
+            Max = 30,
+            Min = 0,
+        },
+        JammerBlips = 6,
     },
     Interface = {
         HelpText = '<LOC ueb4203_help>Stealth Field Generator',
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -3030,7 +3030,7 @@ Unit = Class(moho.unit_methods) {
 
     ---@param self Unit
     ---@param disabler string
-    ---@param intel string
+    ---@param intel IntelType
     DisableUnitIntel = function(self, disabler, intel)
         local function DisableOneIntel(disabler, intel)
             local intDisabled = false
@@ -3078,7 +3078,7 @@ Unit = Class(moho.unit_methods) {
 
     ---@param self Unit
     ---@param disabler string
-    ---@param intel string
+    ---@param intel IntelType
     EnableUnitIntel = function(self, disabler, intel)
         local function EnableOneIntel(disabler, intel)
             local intEnabled = false
@@ -3122,16 +3122,19 @@ Unit = Class(moho.unit_methods) {
         end
 
         if intEnabled then
-            self:OnIntelEnabled()
+            self:OnIntelEnabled(intel)
         end
     end,
 
     ---@param self Unit
-    OnIntelEnabled = function(self)
+    ---@param type IntelType
+    OnIntelEnabled = function(self, type)
     end,
 
     ---@param self Unit
-    OnIntelDisabled = function(self)
+    ---@param disabler string
+    ---@param type IntelType
+    OnIntelDisabled = function(self, disabler, type)
     end,
 
     ---@param self Unit
--- a/units/UAB4203/UAB4203_unit.bp
+++ b/units/UAB4203/UAB4203_unit.bp
@@ -167,10 +167,9 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        RadarStealthFieldRadius = 24,
+        RadarStealthFieldRadius = 32,
         ReactivateTime = 5,
-        SonarStealthFieldRadius = 24,
-        VisionRadius = 20,
+        VisionRadius = 32,
     },
     Interface = {
         HelpText = '<LOC uab4203_help>Stealth Field Generator',
--- a/units/UAB4203/UAB4203_unit.bp
+++ b/units/UAB4203/UAB4203_unit.bp
@@ -167,9 +167,9 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        RadarStealthFieldRadius = 24,
+        RadarStealthFieldRadius = 26,
         ReactivateTime = 5,
-        VisionRadius = 24,
+        VisionRadius = 26,
     },
     Interface = {
         HelpText = '<LOC uab4203_help>Stealth Field Generator',
--- a/units/UAB4203/UAB4203_unit.bp
+++ b/units/UAB4203/UAB4203_unit.bp
@@ -167,9 +167,9 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        RadarStealthFieldRadius = 32,
+        RadarStealthFieldRadius = 24,
         ReactivateTime = 5,
-        VisionRadius = 32,
+        VisionRadius = 24,
     },
     Interface = {
         HelpText = '<LOC uab4203_help>Stealth Field Generator',
--- a/units/URB4203/URB4203_unit.bp
+++ b/units/URB4203/URB4203_unit.bp
@@ -141,9 +141,9 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        RadarStealthFieldRadius = 42,
+        RadarStealthFieldRadius = 32,
         ReactivateTime = 5,
-        VisionRadius = 26,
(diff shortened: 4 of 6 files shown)


Garanas:
image

Example of the jamming 😄 !

units/UEB4203/UEB4203_unit.bp
```diff
@@ -140,10 +140,14 @@ UnitBlueprint {
         UnitWeight = 1,
     },
     Intel = {
-        RadarStealthFieldRadius = 24,
+        RadarStealthFieldRadius = 32,
         ReactivateTime = 5,
-        SonarStealthFieldRadius = 24,
         VisionRadius = 20,
+        JamRadius = {
```
Tagada14, line 146: Vision radius for UEF stealth gen should also probably be 26

ComradeStryker:
For the UEF stealth field -

This is a huge nerf for the structure even though you're trying to improve how it would work with the faction theme.
A stealth field should hide what's underneath their coverage area... Not make it more obvious. 

Unless the jamming ability is separate from the stealth ability, then it's a giant nerf for the structure and how it works within the UEF. 

If the change goes through, you may as well rename the structure to "Jamming Field".

It's a cool idea.. but only if this new ability is switchable.
"Stealth Mode" or "Jamming Mode". Or better yet, have both abilities with a Toggle button. 
So, you could have either or both active. Hiding whatever units are underneath it and displaying the blips.

Speaking of blips. I think more blips are needed if it is to become an effective ability for this Stealth Field.
A Sparky has 8 extra radar blips... why build something more expensive if a unit can do it for far cheaper?

---

For other factions... nice ideas but I'm concerned about the Cybran stealth field - it already is the cheapest one and quickest to build... you're improving the best stealth field even more so by increasing the range of it, significantly without other drawbacks?

---

~ Stryker

Garanas:
To preserve that, we could:

 - (1) By default increase all radia, as that is the core idea of this pull request
 - (2) Add upgrades for faction specific flavors: Aeon gains more vision, UEF gains jamming blips that look like buildings, Cybran loses maintenance costs and Seraphim gains flying blips

speed2CZ:
I'd say they are already interesting enough, even before any buff. Omni never reaches into your main base. And placing couple of stealth fields in your base help hide what you're up so, since you dont see clumps of engineers around new exps or game enders right away.

ComradeStryker:
Thing is, stealth fields are almost never used in a player's base - 
They're used offensively 70 or 80% of the time.
To hide a tac base, an ACU, a PD creep (for those of you who do that.), etc.

Sure, stealth fields can be used in bases to help hide a project, like a game-ender or an early Monkeylord rush; (as @speed2CZ mentioned.)
However, these defensive fields don't really need the new changes.
The best benefit of these changes is to add them to offensive stealth fields. 

---

@Garanas, This may be a little far-fetched, not gonna lie, but may be an idea worth looking into:

Perhaps the UEF and Sera stealth field could instead work like the Seraphim's ACU with the regen field?
The UEF is all about HP, right? So, why not have the Stealth Field buff the HP of units (and structures?) in its range by 10-15%?
No HP regen buff, just HP capacity buff.

The Sera Field could have a 5-7.5% HP buff but with a minor regen effect as well.
+5 HP/S regen.

---

But... the reason I say this is far-fetched, is because this doesn't really fit with the description of the structure, being a Stealth Field, and all.
Sure, jamming is nice, but it isn't anywhere near as useful as the Stealth these structures provide.
If the structures, instead, offered their abilities first, and then be upgraded to stealth fields... this could work, better.
But, this would drastically change how these Stealth Fields would work at a glance, so... it's difficult to say.

---

~ Stryker

Garanas:
I like the idea of boosting some other attribute, I'm also not sure if they are still stealth fields at that point though

Nojoke185:
hows about this....

default steath gen behaviour but with option to upgrade to allow to project to another map area (like the eye of Rhianne) a small steathed area?

Garanas:
https://user-images.githubusercontent.com/15778155/211414655-72ca377e-69d7-4d4b-bea4-d46121f0b40e.mp4

Improved the fx for the sera stealth field 😄

ComradeStryker:
I thought of a few ideas, hopefully, one of them might stick.

---

All Factions: 
Increase their stealth range as per the post by default.

(Cybran should still have the largest field.)

---

Aeon: 
Sensor Upgrade:

Increase vision

---

UEF: 
Upgrade to Shield & Jammer

The shield upgrade wouldn't be as strong as a dedicated T2 shield but it would offer some protection to structures and units. 
When the Shield is up, Stealth is down and Jamming takes over.

Example:
image

---

Cybran: 
Cloak Field

Cloak Radius is very small (Matching that or even less than the build range of a T1 engineer.)
Only would affect units. (Structures and Exps would be exempt.)
(Large energy consumption (-300E))

Example:
image

---

Seraphim:
Regen Field:

+5 HP/s for all units (and structures?) in the imidate area.
(Outer yellow circle of this image.)

image

---

If need be, the UEF could get its own variant of a "Jamming Field" as a standalone structure.
I always found it strange why the UEF ACU didn't have a Jamming upgrade - this could be a way to introduce it to the faction. (T1 structure perhaps?)

---

These ideas would obviously have to be balanced properly with each faction in mind, but, these are the main gists of it.

---

~ Stryker

MrRowey:
I remember looking into this extensively and notice the ring that the steal field shows is not what the field actually covers and it a disparity for all factions. here you can see the wall sections that are visible from a t2 radar

**AEON**
image
image

**UEF**
image
image

**CYBRAN**
image
image

**SERA**
image
image
***Ignore Dark Gray Blips their the extra ones reclaimed from Cybran one***

As you can see each one the field dosnet starts till 3 wallsecion in from the right hand side witch is a lot and 1 from the bottom

Garanas:
I remember something about that, yeah

maudlin27:
I like the idea of increased vision for Aeon; for UEF also agree that Jamming for UEF+Sera could be a negative.  If a faction specific theme is wanted what about just going with the 'UEF = higher health' theme and giving the stealth gen 1k hp for UEF.
Similarly Sera's theme could be taking a bit of each faction but not doing it as well (a bit like it's combat scout), so it's stealth is worse than cybran but better than Aeon+UEF, it's vision is worse than Aeon but better than the others, it's health is 2nd best, and it costs a bit more?

Hdt80bro:
I like the different ideas presented here, but it feels a bit like a mod trying to extrapolate out complete unit diversity across all factions on the basis of a few interesting differences. I don't think we would need to force every factions' units to be special because of the quirks of the others, but I do understand the lack of compelling reasons to use stealth field generators (except in a few niche circumstances). I see a more fundamental problem with them, but adding abilities might also work, though harder to get right.

Enough has been said about the radar jamming :)

Adding more vision range to the Veil feels like it goes against the Aeon faction design of unit specificity, especially since "seeing stuff" and "making stuff not seen" are rather opposite. They would have a separate building with the sole purpose of providing a vision radius in a certain area... ;-) Actually, if anything, since Cybran has the Soothsayer and Aeon has the Eye of Rhianne, then UEF and Seraphim are the only factions "missing" a vision structure, so one of them "should" get the added vision radius (though, again, forcing factionally complete specialization feels like mod territory to me and a little weird here).

The building costs, maintenance, and health do already follow the original faction stat themes, but I don't see a problem with exaggerating them a bit if we think the balance would improve. If we do add the stealth field radius to that, the power maintenance should probably also be changed to match, similar to shields (though with the Cybran shield anti-affinity inverted of course). Perhaps not quite to same extent however, as there's also no shield HP counterpart for stealth to diversify.

As far as removing the sonar stealth field, I don't see any reason to. Other stealthed units work just as well against sonar as they do radar, and it makes sense that the technology would be designed to allow for that. Splitting the stealth field over radar and sonar is a little odd, since personal stealth doesn't work that way, but I can't think of a reason why we need to nerf them based purely on that decoupling. If we did remove the sonar stealth field from stealth field generators, then we should also remove it from the Deceiver, Mermaid, Flood XR, and Monkeylord for consistency, and I don't see that happening.

---

To me, the biggest problem with the stealth field generators is the space density they require relative to the benefit they provide. Based only on the economy cost, the gains from plopping down a stealth field here & there should be enough benefit to be worthwhile (though definitely not enough to be considered unbalanced), but I suspect that we don't see them used all that much because they also cost a disproportionately large footprint that competes for shield space, pathfinding, and adjacency (in the long term, placing deceivers everywhere would be a more efficient way to stealth your things [for *any* metric: space density, cost, health, maintenance, mobility...], but the point of the game when you can do that is usually after stealth has much use).

This is actually half of how higher tech units are balanced as well: lower tier units do deal more damage per cost, but you can't fit enough of that damage in one spot because of how spread out it is (they have comparatively low damage density, which is also due to having shorter range and lower survivability) so the more advanced units end up being effective at dealing larger amounts of damage.

The shield generator space density is just about right (taking just enough room compared to its shield dome to be useful while forcing diminishing returns--shield spillover seems to finely tune the diminishing return ratio to something even more balanced) and has more physical benefit so that it takes priority (it might be that players just don't feel the effect of enemy lack of intel enough to think stealth generators are worth it too). However, then adding a stealth field to that space (which takes just as much room as the shield generator) doesn't leave much room for the other buildings you're trying to protect. This means we only see it when "not getting hit" isn't as important as "not getting shot at" (e.g. when setting up that first fire base on *some turtley maps with established metas*, incursions deep inside enemy territory that can be very easily stopped if known, or protected development such as game enders you want the enemy to have as little time to react to).

There's not really a good solution for this at this point in time. It's a little late in history to change the size (e.g. templates and compatibility), but the options that would help aren't optimal anyway:
* As a 2x2 building (like a radar) instead of 6x6 (like a shield generator), overlapping stealth fields has no added benefit that the increased space density could provide, but then there's no reason to ever not build it (and the models wouldn't fit anyway). It needs some penalty to add a tactical tradeoff such that it needs to be considered and not be yet another thing players need to always worry about.
* As a 3x3 building (like an SMD), the odd size messes up structure packing and adjacency. I also suspect this is still a little too small.
* No 4x4 building currently exists and sounds awful to get working (not to mention the mod incompatibility of introducing a new size). Otherwise, this might have been ideal.

---

Anywhooo, it's much too late to apologize for bluntness or any perceived harshness (I never intend that), but I'll at least acknowledge the particularness of the things I have to say. As usual, take it as advice you can choose to consider when making your decisions.

Hdt80bro:
@MrRowey this is because intel still uses the grid-based system from Vanilla Sup Com--the perfect-looking vision rings are due to shaders and only visual
image
image

MrRowey:
Intresing

Garanas:
I'm closing this one - I'll copy the effect changes over to another pull request.

PR #4576, opened 2023-01-10
Reduce impact of intel management on simulation

Garanas:
After two attempts it feels like this is a solution that we can work with. See also:

- https://github.com/FAForever/fa/pull/3874
- https://github.com/FAForever/fa/pull/4022

It appears to catch all situations:

- (1) Units with no intel have no intel status table
- (2) Units with only 'free' intel have no intel status table
- (3) Units with some free, and some maintenance-related intel only take into account the intel that is part of the maintenance
- (4) Units with enhancements are taken into account, even if they don't have the enhancement yet
- (5) Whether you start with or without energy, all is good

(1) is by far the biggest concern: a lot of units have no intel besides vision, they do not require the data structure to track it. The original data structure is about 1300 bytes, multiply that by 5K units and you end up with 7+ wasted megabytes. (2) means that units that require no tracking also have no intel status, think in terms of scouts and frigates. And (3) guarantees that if a unit has some free, and some non-free intel then only the non-free intel is taken into account. Lastly (4) was a bit of a headache, but this solution appears to work reliably and only requires us to track the intel types that are adjusted by the enhancements.

(4) does assume that all intel of enhancements is manually disabled in the unit script. This is the case for every unit in the base game since GPG, I suspect mod authors copied that pattern. If they did not, then so be it.

There's still some todo's:

- [x] Add back in the recharge time
- [x] Remove debugging statements
- [x] Refactor `EnergyDependingUnits` from being array based to hash based
- [x] ~~Refactor `EnergyDependingUnits` to no longer rely on an economy event, instead just query the brain each tick~~
- [x] Refactor whatever is in `OnCreate` of the `IntelComponent` to the blueprint creation phase: that way we don't have to compute it as we go in-game
- [x] Look into the cloak (field) exceptions
- [x] Refactor all `OnIntelEnabled` and `OnIntelDisabled` to also accept the intel type


Garanas:
@Hdt80bro do you have time to review this?

Sprouto:
This is a valuable change - provided it's seamless with modded units with customized intel features.   The memory saved is rather substantial, on paper, if not in game - but as pointed out - much of it is redundant data structure, especially for those units with FreeIntel, or only one or two intel types.   The cloak feature, in conjunction with mod units that grant 'cloak' on a temporary basis to units, is going to be tricky to implement.

Garanas:
What units would have such behavior? I can toy with them, see how they operate.

Garanas:
image

image

Appears to work as expected with the cloak field generator.

PR #4578, opened 2023-01-10
Add ability to invert panning direction with middle mouse button

Garanas:
Adds option to invert panning with (the middle) mouse (button)


MrRowey:
@lL1l1 can you check this one out see if it works and if it dose we can add it in

lL1l1 (approved):
It works.

PR #4579, opened 2023-01-11
Refactoring Entities

MrRowey:


speed2CZ:
Why are you adding 1 tick everywhere?
You've also converted several `WaitSeconds(0.1)` to `WaitTicks(11)`, which is quite wrong.
This looks like just creating bugs from working code to me rather than improving anything.

Garanas:
Rowey is inlining `WaitSeconds` , as defined:

```lua
---@param n number
function WaitSeconds(n)
    if n <= 0.1 then
        WaitTicks(1)
        return
    end
    WaitTicks(n * 10 + 1) -- <- the +1
end
```

The pull request still needs to be reviewed - it is far from being merged in.

@MrRowey please do not further extend this pr, I was almost done reviewing the first 16 and then poof 😄

MrRowey:
> @MrRowey please do not further extend this pr, I was almost done reviewing the first 16 and then poof 😄

😆 Yeh this pr wont get any more changes to it I'm done with adding to it :)

PR #4583, opened 2023-01-11
Allow all experimentals to walk backwards

Garanas:
As the title states - allows all (land) experimentals to move backwards. They will all have the same backup distance. Similar to the Megalith, all non-fatboy experimentals will move backwards at half their usual movement speed.

### DIFF
--- a/units/XSL0401/XSL0401_Script.lua
+++ b/units/XSL0401/XSL0401_Script.lua
@@ -146,6 +146,18 @@ XSL0401 = Class(SWalkingLandUnit) {
         self:Destroy()
     end,
 
+    OnMotionHorzEventChange = function(self, new, old)
+        SWalkingLandUnit.OnMotionHorzEventChange(self, new, old)
+
+        if (old == 'Stopped') then
+            local bpDisplay = self:GetBlueprint().Display
+            if bpDisplay.AnimationWalk and self.Animator then
+                self.Animator:SetDirectionalAnim(true)
+                self.Animator:SetRate(bpDisplay.AnimationWalkRate)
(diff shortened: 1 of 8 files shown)


Garanas:
We could add it to the Atlantis, Tempest and Scathis too.

units/XSL0401/XSL0401_Script.lua
```diff
@@ -146,6 +146,18 @@ XSL0401 = Class(SWalkingLandUnit) {
         self:Destroy()
     end,
 
+    OnMotionHorzEventChange = function(self, new, old)
```
4z0t, line 149: Why is it copied 3 times?

4z0t (changes requested):

units/XSL0401/XSL0401_Script.lua
Garanas: because I'll make a general solution if the balance team accepts the idea behind the change

The-Balthazar:
Is there any specific reason for the animations to not be set to reversible anyway, even for units that can't walk backwards?

Garanas:
I don't know, I was thinking about that too.

RabidPope:
Hi. Sorry if this is the wrong place to ask. Could the same be done for mantis and ilshies? They tend to get stuck.

Garanas:
Yeah, it is possible - not sure if this would allow them to get unstuck though. I'll experiment with it

4z0t:
@Garanas should we close this?

Garanas:
Yeah - I'm in no rush on these changes

PR #4586, opened 2023-01-14
Integrate 'NoShake' UI mod

Garanas:
The mod is not deprecated - there is no technical reason to do so.

image


Garanas:
@Hdt80bro I'm not sure whether this is an interface item, or a gameplay item. What are your thoughts?

Hdt80bro:
Probably `Gameplay` with the rest of the camera controls--it's an in-world thing, I view `Interface` as UI stuff.

PR #4589, opened 2023-01-15
Initial draft for threat threshold for navigational mesh

Garanas:
image

Mapped for testing to the `PathTo` debug functionality.

image


relent0r:
@Garanas Just yelling out. I still have it on the agenda to fully test this.

Garanas:
@relent0r this is ready for testing and, if all is good, for a merge!

image

There is a separate tab to help debugging and understanding the new pathing function. You can toy around with the different threat functions, threat radia and the threat threshold. You can also pick the layer you'd like to path over. You can check the threat values (to a rough degree) using the map utilities window. I found it to be unsufficient when doing more thorough testing. I'll make a new soon that (for example) displays the current threat values of a given cell, like the reclaim debugging ui does for reclaim values.

image

image

image

image

Garanas:
An interesting discovery is that the existing pathing algorithm does not appear to take into account point defense. Those structures do not have the `LAND` category. The old function as found here:

- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/AI/aiattackutilities.lua

Usually uses the platoon layer (`LAND`, `AMPHIBIOUS`, `AIR`, and `NAVAL`) as the threat identifier. As a result:

- All `LAND` and `AMPHIBIOUS` and `AIR` checks are entirely blind of structure-based defenses as they have neither of these categories (they are `STRUCTURE`)
- An `AMPHIBIOUS` platoon could be entirely blind of a `LAND` platoon that consists solely of that category (and vice versa)

Instead, they should've used:

- `ANTISURFACE` for anything that may pose a threat to land
- `ANTIAIR` for anything that may pose a threat to air
- `ANTISUB` for anything that may pose a threat to submarines (and then combine it with `ANTISURFACE`)

Garanas:
@relent0r here is an abstract snippet on how to use this. If you have the plugin (if not, then go get it!) then the type system should help you with the possibilities as you write the code! The first three arguments are similar to that of `PathTo`

```lua
local NavUtils = import("/lua/sim/NavUtils.lua")

-- these depend on the context
local brain = ...
local pathingOrigin = ...
local pathingDestination = ...
local pathingLayer = ...
local threatThreshold = ...
local threatRadius = ...

-- retrieve the threat function we want to use, see "/lua/shared/NavGenerator.lua" for the full list
local threatFunc = NavUtils.ThreatFunctions['Air']

-- then do the call, the results are similar to "NavUtils.PathTo" except that this one takes into account the threat value
local path, n, distance = NavUtils.PathToWithThreatThreshold(
    pathingLayer,
    pathingOrigin,
    pathingDestination,
    brain,
    threatFunc,
    threatThreshold,
    threatRadius
)

-- do something with the path...

```

PR #4603, opened 2023-01-19
Code Tidy-up

MrRowey:
- Fix Naming of Explosion effects
- removed unessary empty lines

### DIFF
--- a/effects/Explosion/Explosion01_a_script.lua
+++ b/effects/Explosion/Explosion01_a_script.lua
@@ -1,10 +1,4 @@
---
 -- script for projectile BoneAttached
---
 local NullShell = import("/lua/sim/defaultprojectiles.lua").NullShell
-
-Explosion01_a = Class(NullShell) {
-}
-
+Explosion01_a = Class(NullShell) {}
 TypeClass = Explosion01_a
-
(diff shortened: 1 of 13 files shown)


effects/Explosion/Explosion01_a_script.lua
```diff
@@ -1,10 +1,4 @@
...
---
 local NullShell = import("/lua/sim/defaultprojectiles.lua").NullShell
-
-Explosion01_a = Class(NullShell) {
-}
-
-TypeClass = Explosion01_a
-
+Explosion01_a = Class(NullShell) {}
+TypeClass = Explosion01_a
```
Garanas, line 4: I'm quite confident that these are not used. The only reference we have is in [defaultexplosions.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultexplosions.lua). Maybe best to just remove these all together.

Garanas (approved):
Just annotation changes - they appear to be fine.

effects/Explosion/Explosion01_a_script.lua
MrRowey: So do should these be replace by something or remove the inserts of the meshes altogether a-d even where it is called in defaulteexplosions.lua

PR #4612, opened 2023-01-21
Fix hooking issues for mods

Garanas:
The use of `lazyimport` appears to break hooking.


Uveso:
Tested and its working perfect!

Issue #4601, opened 2023-01-19
Build Menu Resets selected options when platoon changes size (units die)


piisawheel:
**Describe the bug**
See Title.

**To Reproduce**
Go into a game.  Create 5 cybran support commanders.  Select a *different* option on the build menu (maybe you want to see the t2 cerberus for example).  Destroy 1 Unit (maybe put him near an enemy turret so he dies while you have the platoon selected - or self d and select the platoon before the countdown).  The menu has reset back to tier 3 (the default) because the platoon changed sizes (1 unit was killed).  This is *extremely* annoying to those of us with low apm, who, in the middle of a firefight, have to spam t2 and then hope we can select the cerberus turret before 1 of your 50 units dies and reset the build menu back to t3.  I've selected the t3 torpedo launcher so many times in survival it's not funny, and almost frustrated me to the point of rage quitting the game I was in (survival ffa, not leaving a team screwed).

**Expected behavior**
When units are already selected, and the squad changes size *not due to player interaction* (units die), the menu should not reset, unless no units are alive that can build that tier.

Garanas:
We'll pick it up 👍

PR #4617, opened 2023-01-22
Fix reset of construction menu when selecting a structure

Garanas:
Also closes https://github.com/FAForever/fa/issues/4601 . The reset was caused by the option `gui_all_race_templates`, which represents `All Faction Templates`. The sequence was:

- Click structure to build
- Update UI: no templates, so 1 less active tab
- Add templates from all faction templates option
- Update UI: templates, so 1 more active tab

As a result, the construction menu 'changed' from 5 to 4 tabs, and then back from 4 to 5 tabs. Because the number of tabs changed the default tab (which is the highest tech, up to tech 3) was reselected. The code is quite spaghetti, but it was missing the sauce. So to fix it we add an if statement that skips checking for a change in templates.

The 'quick fix' for those that encounter this is to give each faction at least 1 template that you made with that faction.

### DIFF
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -398,7 +398,6 @@ function CreateTabs(type)
     local desiredTabs = 0
     -- Construction tab, this is called before fac templates have been added
     if type == 'construction' and allFactories and options.gui_templates_factory ~= 0 then
-        LOG("Set templates")
         -- nil value would cause refresh issues if templates tab is currently selected
         sortedOptions.templates = {}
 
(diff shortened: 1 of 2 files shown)


lua/ui/game/construction.lua
```diff
@@ -408,7 +408,9 @@ function CreateTabs(type)
             local numActive = 0
             for _, tab in controls.tabs do
                 if sortedOptions[tab.ID] and not table.empty(sortedOptions[tab.ID]) then
-                    numActive = numActive + 1
+                    if tab.ID != 'templates' then
```
speed2CZ, line 411: `~=` ?

Garanas: Yeah, that works too.

PR #4618, opened 2023-01-22
Adjust elevation of strategic bombers to make them easier to hit

Garanas:
Adjusts the elevation from `20` to `14`. All the other units could really use a checkup too, they are all over the place. Some transports have an elevation of around 10, others of around 18. The latter lands slower and is a lot more difficult to hit.

In general I'd say:

- Tech 1 at a height of 10 - 12, with an exception for scouts (maybe 14)
- Tech 2 at a height of 12 - 14
- Tech 3 at a height of 14 - 16, with an exception for scouts (maybe 18)
- Tech 4 is at a height of 16 - 18

But that is for the future.

### DIFF
--- a/units/UAA0304/UAA0304_unit.bp
+++ b/units/UAA0304/UAA0304_unit.bp
@@ -222,7 +222,7 @@ UnitBlueprint {
             LAYER_Sub = false,
             LAYER_Water = false,
         },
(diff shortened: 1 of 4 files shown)


Garanas:
A quick test near the edge of the weapon radius:

image

image

PR #4625, opened 2023-01-24
Reduce impact of walls on simulation

Garanas:
Drastically reduces the number of table fields for a wall structure. They now only have what is expected of every unit. Sadly, to do this they'd either inherit from a DummyUnit or break the hierarchy of the usual unit class. In this pull request I've chosen for the second approach, but I'm open to suggestions.

### DIFF
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -280,10 +280,12 @@ function TransferUnitsOwnership(units, toArmy, captured)
     -- add delay on turning on each weapon 
     for _, unit in newUnits do
         -- disable all weapons, enable with a delay
-        for k = 1, unit.WeaponCount do
-            local weapon = unit:GetWeapon(k)
-            weapon:SetEnabled(false)
-            weapon:ForkThread(TransferUnitsOwnershipDelayedWeapons)
+        if unit.WeaponCount then
+            for k = 1, unit.WeaponCount do
+                local weapon = unit:GetWeapon(k)
+                weapon:SetEnabled(false)
+                weapon:ForkThread(TransferUnitsOwnershipDelayedWeapons)
+            end
         end
     end
 
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -6,6 +6,7 @@
 -----------------------------------------------------------------
 local Entity = import("/lua/sim/entity.lua").Entity
 local Unit = import("/lua/sim/unit.lua").Unit
+local DummyUnit = import("/lua/sim/unit.lua").DummyUnit
 local explosion = import("/lua/defaultexplosions.lua")
 local EffectUtil = import("/lua/effectutilities.lua")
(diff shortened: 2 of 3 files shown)


Garanas:
@The-Balthazar what are your thoughts?

The-Balthazar:
Either approach will break any mod that expects regular units that can be interacted with in normal ways to inherit from Unit. Dummy units generally can't be reclaimed, or damaged, or targetted. Clearing OnPreCreate alone will probably break several of my mods.

The-Balthazar:
Doesn't this as written break the slope-to-terrain thing for them?

Garanas:
> Doesn't this as written break the slope-to-terrain thing for them?

No, as they inherit from a structure and that part is inherited.

> Either approach will break any mod that expects regular units that can be interacted with in normal ways to inherit from Unit. Dummy units generally can't be reclaimed, or damaged, or targetted. Clearing OnPreCreate alone will probably break several of my mods.

I'm aware, therefore the discussion. How would we tackle this best?

The-Balthazar:
So I think probably having a flag for like "SimpleUnit", that is true for them, maybe hard coded on the unit script level, that more complex effects within functions check against, that way it will still inherit everything, and the stuff you don't want in will go.
> No, as they inherit from a structure and that part is inherited.

Isn't that in OnCreate, or OnStopBeingBuilt, that you're overwriting there?

Garanas:
> So I think probably having a flag for like "SimpleUnit", that is true for them, maybe hard coded on the unit script level, that more complex effects within functions check against, that way it will still inherit everything, and the stuff you don't want in will go.

I like this idea - maybe we should have done that for the dummy drones too.

> Isn't that in OnCreate, or OnStopBeingBuilt, that you're overwriting there?

It is in `OnStartBeingBuilt`

The-Balthazar:
I think it's fine to have a clear delination between "Simple" units and "Dummy" units, with simple being units that aren't expected to do anything other than have basic interactions done on it, and dummy units being only units for technical reasons, and not being damagable, reclaimable, or otherwise interactable outside of things specifically trying to interact with them. Like walls vs cybran build effect, and ferry beacons.

The slope effect should probably be in something that triggers when the unit is spawned off, like OnCreate.

It's a bit of an aside but: I also think super basic reference stuff like `self.UnitId` should be defined in OnPreCreate or hard coded in unit-level scripts, but that's another point. There's a lot of small order of opperation things that that would fix, like people hooking OnCreate, and calling something that uses those before they're defined, when in the old system of it using `GetUnitId` it would still work.

Garanas:
I'd be open to that idea, I'll adjust this pr accordingly

Garanas:
> The slope effect should probably be in something that triggers when the unit is spawned off, like OnCreate.

Tackled that in: https://github.com/FAForever/fa/pull/4636

Issue #4610, opened 2023-01-21
Refactor shields to use the `ShieldEffectsComponent`


Garanas:
Description
-----------

There are various shields in the game. They all have their own setup to create effects, but each setup bottles down to introducing various effects to a specific bone. We can refactor this to:

- (1) Reduce code duplication
- (2) Proper use of trash bags

We introduced a component system for these type of tasks. You can find it in [defaultcomponents.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultcomponents.lua), and in particular the [ShieldEffectsComponent](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultcomponents.lua#L7-L31). It contains all the logic required to efficiently create and destroy shield effects.

Course of action
----------------

Refactor the remaining shield units to use the shield effects component. Do not touch units specifically designed for campaign, such as the Aeon Palace. All units you refactor should use the shield effects component for their shield effects. Note that rotators, sliders and other manipulators are too unit specific to generalize, and therefore they remain in the individual unit classes.

You can use [ueb4301](https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UAB4202/UAB4202_script.lua), [ual0307](https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UAL0307/UAL0307_script.lua), [ueb4202](https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UEB4202/UEB4202_script.lua) and [uel0307](https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UEB0307/UEB0307_script.lua) as examples. 

Test plan
---------

You can create the units using the unit cheat menu. You can find your binding in the hotkeys menu:

![image](https://user-images.githubusercontent.com/15778155/213861016-38224fb4-67ee-4373-93d6-7773aed0b6be.png)

Learning goals
--------------

Learn to work with Lua and the repository. If you use the [extension](https://github.com/FAForever/fa-lua-vscode-extension) then note that the extension can make mistakes - it is not perfect. Always evaluate what it is trying to tell you, and whether you should ignore that.

scarress:
Can i continue on this issue?
I have tested some changes locally addressing the emitters not being placed correctly for cybran shield gens and for UEF shield bot.
Seems like the the other pr, is not progressing atm.

Garanas:
@MrRowey can he pickup this issue?

PR #4628, opened 2023-01-24
Structure/Units Shield using Component File

MrRowey:
This Pr is going towards #4610 

Confirmed Working and Tested

- [ ] URB4202
- [ ] URB4204
- [ ] URB4205
- [ ] URB4206
- [ ] URB4207
- [ ] XSB4202
- [ ] XSB4301
- [ ] XSL0307


Garanas (changes requested):
This pull request has not been tested.

As a few examples:

- `ShieldEffectsBag` was overwritten when the unit was done building, see my commit
- `ShieldEffectsBone` is not defined as a class value, therefore all the effects are on the ground

image

It is very important that you test changes in-game, as you make them. Especially these changes that are on a unit-to-unit basis: you can spawn in the unit and observe the behavior.

MrRowey:
It seems this part of the Cybran shields have a table.insert which is not Appling the effect to the bone hence them being spawned on the floor

```lua
        for k, v in self.ShieldEffects do
            table.insert( self.ShieldEffectsBag, CreateAttachedEmitter(self, 'Shaft', self.Army, v))
        end
```

PR #4635, opened 2023-01-25
Fix direction of adjacency beams

Garanas:
All adjacency beams currently take energy from what they are adjacent of, instead of providing it to 😄 !

https://user-images.githubusercontent.com/15778155/214690159-a962d71d-993f-41d9-91ae-2db53dd87435.mp4


Hdt80bro (approved):

speed2CZ:
This makes storages "send" resources into pgens and mexes

Garanas:
Ah, good point. I'll add an if statement to check the type

PR #4636, opened 2023-01-25
Orientate structures to match terrain when created

Garanas:
Previously the orientation was set when the unit starts being built. But when units are spawned in they never get built, and therefore they would not orientate properly. Now structures will always orientate if they are set to do so.


Garanas:
image

This also allows it to work using the cheat menu 😃

PR #4638, opened 2023-01-26
Fix incorrect substitution in shader

Garanas:
https://user-images.githubusercontent.com/15778155/214900254-c39c437d-6629-4029-befe-87518ca9f14e.mp4


Garanas:
@BlackYps this is related to the PBR changes, could you check if the view direction is still correct?

BlackYps:
I can see that we inadvertantly changed how vertex.position gets calculated, but I don't see that used by the uef build shader, which confuses me massively.

You need to calculate position after viewDirection, otherwise you break it. Just pull that first changed line down.

BlackYps:
I just realized that many other vertex shaders also use the old incorrect method of calculating the viewDirection. We should change them to the new correct method as well.

BlackYps (approved):
Looks good now

PR #4640, opened 2023-01-26
Improve memory management of treads

Garanas:
An interesting one, every time a unit would: 

- Transition to or from 'Stopping' movement state
- Transition to a new terrain type
- Get beamed up / down from a transport

A unit with treads would:

- Re-allocate the `TreadThreads` table multiple times
- Kill all treads, if they exist
- Allocate new threads, if we need them

And then something like this happens, that constantly triggers the transition events:

https://user-images.githubusercontent.com/15778155/214961016-2da33897-3604-48df-ade9-14eb819cac83.mp4

On top of that, this pull request decreases the time that a splat (a tread mark) exists by tech while increasing the LOD by tech. See also: 

- https://github.com/FAForever/fa/pull/4640/files#diff-4efe2c03fd02522260393e6dde9105da93b49718bc4100b67d194f8b06084c35R34-R46

### DIFF
--- a/lua/defaultcomponents.lua
+++ b/lua/defaultcomponents.lua
@@ -65,7 +65,8 @@ TreadComponent = ClassSimple {
             self:AddThreadScroller(1.0, treads.ScrollMultiplier or 0.2)
 
             local treadMarks = treads.TreadMarks
-            if treadMarks and self.TerrainType.Treads ~= 'None' then
+            local treadType = self.TerrainType.Treads
+            if treadMarks and treadType and treadType ~= 'None' then
                 self:CreateTreads(treadMarks)
             end
         end
--- a/lua/defaultcomponents.lua
+++ b/lua/defaultcomponents.lua
@@ -1,4 +1,3 @@
-
 ---@class ShieldEffectsComponent : Unit
 ---@field Trash TrashBag
 ---@field ShieldEffectsBag TrashBag
@@ -6,7 +5,7 @@
 ---@field ShieldEffectsScale number
 ShieldEffectsComponent = ClassSimple {
 
-    ShieldEffects = { },
+    ShieldEffects = {},
     ShieldEffectsBone = 0,
     ShieldEffectsScale = 1,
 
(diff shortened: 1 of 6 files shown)


Garanas:
Note that the system doesn't actually work properly: try moving 200 tanks at once and you'll see what I mean. There appears to he a maximum to the number of splats being rendered, and that maximum is relatively low in comparison to the number of splats generated.

BlackYps:
I think that's the same reason why building tarmacs etc. disappear sometimes when tilting the camera. The render limit kicks in. I guess they implemented it to not stress the hardware too much, but today it shouldn't be an issue anymore. Probably needs an engine patch to increase it though.

Do I see this correctly that different tread textures per terrain type are possible?

Hdt80bro:
Yes, *any* effect for any terrain type can be changed, see https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/TerrainTypes.lua

Garanas:
> I think that's the same reason why building tarmacs etc. disappear sometimes when tilting the camera. The render limit kicks in. I guess they implemented it to not stress the hardware too much, but today it shouldn't be an issue anymore. Probably needs an engine patch to increase it though.

Tarmacs are decals, they work different. I'm not entirely sure what the exact difference is.

Garanas:
> Do I see this correctly that different tread textures per terrain type are possible?

Yes, but it didn't quite work like that. At the moment terrain would only disable treads (no need for treads on rocks or metal, for example). As a result the majority of generated maps have no treads. I'm wondering whether that is why generated maps always felt faster than regular maps. With this pull request we change that - treads always work, even when the terrain type tells you not to.

lua/sim/Unit.lua
```diff
@@ -3814,14 +3818,6 @@ Unit = ClassUnit(moho.unit_methods) {
         end
     end,
 
-    ---@param self Unit
-    ---@param pos Vector
-    ---@return TerrainTreadType
-    GetTTTreadType = function(self, pos)
```
Hdt80bro, line 3820: We can't remove this method, it's still used elsewhere causing errors (and probably by mods as well)

lua/defaultcomponents.lua
```diff
@@ -20,12 +19,120 @@ ShieldEffectsComponent = ClassSimple {
...
+
+        while true do
+            while not self.TreadSuspend do
+                CreateSplatOnBone(self, treadOffset, treadBone, treadTexture, sizeX, sizeZ, lod, duration, army)
+                WaitTicks(interval)
+            end
+
+            SuspendCurrentThread()
+            self.TreadSuspend = nil
+            WaitTicks(interval)
```
Hdt80bro, line 135: This delays tread creation by one interval any time the unit moves again

Hdt80bro (changes requested):
It doesn't look like this will work. Adding back the missing function is trivial, but letting treads underwater, on snow, and on tarmacs looks very unnatural. Since threads are very lightweight, I'm interested in the performance gains. You might be able to get more by consolidating all treads for the unit into one thread, instead of one for each. The Fatboy treads also looks very noisy since it always overlaps both full treads, so I'm not sure that's better.

lua/sim/Unit.lua
Garanas: I'll add that back in

lua/defaultcomponents.lua
Garanas: Visually that is fine - during the first interval the unit barely moves.

Garanas:
@Hdt80bro I don't understand.

> Adding back the missing function is trivial, but letting treads underwater, on snow, and on tarmacs looks very unnatural

I agree with the water, but treads have always shown on tarmacs. See also: 

image

Testing on `Arctic Refuge` and you're correct: the treads do not show on snow. Odd - in my personal perception they should be visible on snow too, but maybe a different type of tread. I'll add in the terrain type check again.

> You might be able to get more by consolidating all treads for the unit into one thread, instead of one for each. 

I like the idea, but that would be something for a second iteration. This first iteration is already a good improvement memory wise on the current implementation.

lua/defaultcomponents.lua
Hdt80bro: I can notice it--it's logically an error, just remove it

Hdt80bro:
Ah, my bad, [these](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/TerrainTypes.lua#L1568-L1609) are left over from vanilla
image
Still, not making tread marks on regular concrete, underwater, on snow makes sense (treads look like the dirt being kicked up, something ice and snow doesn't have--some snow types looks like they're only a thin layer and we could change them, but the tread would probably still look different than normal)

Garanas:
We could look into adjusting the terrain type:

image

But how do we change them back, after the tarmac is gone 🤔

lua/defaultcomponents.lua
Garanas: adjusted it to waiting just a single tick

Hdt80bro: Why do you think you need it?

Garanas:
To me the fatboy doesn't look that bad

image

BlackYps:
> But how do we change them back, after the tarmac is gone

Maybe on destruction of the building, when the tarmac is set for removal with a waiting time period. It's not perfect, as treads would show up as soon as the building is destroyed, but while the building is still alive it would work.

lua/defaultcomponents.lua
Garanas: Having an infinite loop with no wait statement is asking for trouble.

Garanas:
@BlackYps it wasn't so much about when to change it, more about what to change it (back) to?

Garanas:
I'll be merging these changes in as they are a good improvement over the old implementation, without changing the old behavior.

PR #4621, opened 2023-01-22
SCU rebalance

TurinturambarFAF:
Nerfs the raw SCU HP and dps to the level of a vanilla ACU and also nerfs the raw SCU bp from 56 to 40 and compensates these changes by buffs of the respective upgrades.

See: #4642

PR #4642, opened 2023-01-27
SCU Adjustments

Tagada14:
#4621 

To do: Change descriptions of upgrades (maybe also some icons) to match what the upgrades do.


MrRowey:

RabidPope:
Just a thought: In addition to these changes, would there perhaps be a chance to consider lowering base acu cost and adding it to the cost of the ras upgrade? That way you can get RAS coms for same cost but they are weaker while making combat coms more viable. Also makes sera coms cheaper for extra income.

PR #4647, opened 2023-01-28
Minor fix to mercy targeting

haifron:
The mercy was allowed to target units on the `Air` layer (i.e flying units) while it was on `Land`/`Water` i.e landed. Now it cannot.

I wasn't able to get it to actually hit an flying unit, but you could give it an order to attack them & it would just cancel the order after taking off, which could lead to players thinking they've given a valid order but they'd misclicked on an air unit.


Garanas:
Good find!

Garanas (approved):

PR #4668, opened 2023-02-03
Fix trees not falling when colliding with units

Garanas:


Garanas:
This was a nasty one - as the game continues it would dip your fps quite a bit due to the number of individual entities to render!

Issue #4666, opened 2023-02-02
UEF mobile shield is missing effects


BlackYps:
The unit normally has an effect in the middle of its shield generator when it is active. These effects seem to have been removed on accident.

PR #4669, opened 2023-02-03
Fix logging of intel and Parashield not having its effects

Garanas:
Closes https://github.com/FAForever/fa/issues/4666


Garanas:
@MrRowey made a similar mistake here 😄

PR #4674, opened 2023-02-03
Add check to prevent canpathtoenemy from running for non baseai

relent0r:
Description of changes

This PR adds a base ai check to the CanPathToCurrentEnemy function that is run during AI initialization.

This is to prevent the navmesh getting loaded when 3rd party AI's are using the default BaseMonitorInitialization thread


relent0r:
As a side note, When I was testing this (with the Uveso AI being present) I had issues with the hook of MiscBuildConditions.lua file.

Garanas:
Could you elaborate on the issues?

relent0r:
> 

Not related to this PR. Just something I noticed. I'll explore them more on wednesday once I've got a bit more time.

PR #4675, opened 2023-02-04
Dynamically compute LODs of units like we do with props

Garanas:
This pull request consists of three parts:

- (1) Dynamically computes the LODs of units
- (2) Dynamically determine the LODs of tarmacs of units
- (3) Fix tarmacs switching up when structure starts / stops upgrading

(1) is only partially complete, a second pull request will finalize it. In this pull request we force the re-computation of all LODs of all units. It uses a similar approach to props, as found in: https://github.com/FAForever/fa/pull/3662 . The second pull request will only compute the LOD value of units that have none, after which we remove all the pre-defined cutoffs of units and props. That way mods we do not 'enforce' these rules on mods.

(2) is a visual thing, where previously:

- (a) Cutoff was the same for different types of decals
- (b) Cutoff was the same, regardless of size

(a) could introduce noise when zoomed out, especially for the Cybran normals. (b) doesn't make a lot of sense, where the decal pops into view sooner than a power generator does, but later than a factory would. We streamline this such that:

- All albedo decals share the same cutoff as the LOD1 mesh blueprint of the unit, if present
- All normal / glow decals share the same cutoff as the LOD0 mesh blueprint of the unit, if present

As a result decals fade in more naturally, where we initially only see the albedo and then as we zoom in we get rewarded with the details provided by the normals decal.

(3) is the technical bug that caused me to look into this. It was reported by Madmax, but I'm sure everyone notices it the moment they know decals can switch up as you start upgrading. This no longer happens - instead decals are inherited and the structure that is responsible ends up destroying them.

### DIFF
--- a/engine/Core/Blueprints/ProjectileBlueprint.lua
+++ b/engine/Core/Blueprints/ProjectileBlueprint.lua
@@ -11,6 +11,7 @@
 ---@field Physics ProjectileBlueprintPhysics
 
 ---@class ProjectileBlueprintDisplay
+---@field Mesh MeshBlueprint
 --- mesh to use as the display of this projectile
 ---@field MeshBlueprint FileName
 --- uniform scale to apply to mesh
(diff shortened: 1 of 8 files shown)


Garanas:
image

image

PR #4676, opened 2023-02-04
Add more PBR techniques

BlackYps:
Adds techniques for units with personal shields and adds shaders for the build process. Also adds functionality to use a "water" environment map that can be provided if the default environment map is non-mirrored to provide a mirrored version. This can be used for ships and other units on water so their reflections make more sense.


effects/mesh.fx
```diff
@@ -9272,22 +9272,65 @@ float4 PBR_UEF(NORMALMAPPED_VERTEX vertex,
     roughness += planeCockpitMask - specular.b * 3;
     roughness = saturate(1 - roughness);
 
-    float4 color = PBR_PS(vertex, albedo.rgb, metallic, roughness, ao, hiDefShadows);
+    float4 color = PBR_PS(vertex, albedo.rgb, metallic, roughness, normal, hiDefShadows, .04, ao);
```
Garanas, line 9275: should `0.04` be `facingSpecular`?

BlackYps: I'm not sure what you mean. Yes, this parameter is for facingSpecular. It's only here because we want to set the ao, so we need to also specify facingSpecular because that argument comes first.

BlackYps:
I found two problems that existed before but become more obvious with these changes: 
Aeon units with personal shields use the UEF technique. I am not sure where this is defined.
The Aeon build vertex shader calculates the shader texture incorrectly. Near the end of the build animation it covers the whole unit in shadow. I investigated but I couldn't find an obvious error in the vertex shader.

effects/mesh.fx
Garanas: I overlooked - this is a different function. My mistake 😄

Garanas:
> Aeon units with personal shields use the UEF technique. I am not sure where this is defined.

These units are shipped with an additional blueprint file, see for example:

- Blueprint name: https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UAL0202/UAL0202_unit.bp#L76
- Mesh blueprint: https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UAL0202/UAL0202_personalshield_mesh.bp

That is a mesh blueprint like any other, and in this case it refers to the `PhaseShield` shader.

Garanas:
> The Aeon build vertex shader calculates the shader texture incorrectly. Near the end of the build animation it covers the whole unit in shadow. I investigated but I couldn't find an obvious error in the vertex shader.

I don't entirely understand, can you make it visual for me?

BlackYps:
PhaseShield is the uef shader. That explains it. Looks like an easy fix

BlackYps:
SCFrame_Sat_Feb_04_161626_2023_00000
Near the end of the build animation you can see a shrinking circle of light and the rest of the unit is treates as being in shadow. No idea why this happens. I found an ugly hack to disable the self-shadowing, which is an improvement. It's way less noticeable that shadows on the unit appear only once it has been completed than the bug in the picture.
If you can find a nicer solution then I am open for it though.

BlackYps:
It's quite impressive what we can do with the game.
Previous:
SCFrame_Sat_Feb_04_162218_2023_00002

Now:
SCFrame_Sat_Feb_04_162130_2023_00001

BlackYps:
Alright, from my view this PR is ready

BlackYps:
I noticed that cybran looked a bit too "shadeless" so I fixed that by abusing the ambient occlusion

Garanas:
@BlackYps shall I merge this, then you can continue on a new branch

BlackYps:
Yes, I think it's ready now for real this time

BlackYps:
Oh, the czar shader is missing, but yes, let's continue this in a new PR instead of having endless scope creep

Garanas (approved):

PR #4680, opened 2023-02-06
Revert a few decisions made the past year

Garanas:
The following fields of the Unit class become optional:

- `SoundEntity` 
- `EngineFlags` 

The following fields of the Unit class are removed:

- `Audio`
- `MovementEffects`


Garanas:
After these changes the average unit has 32 to 34 fields - just barely over the threshold that causes the table to grow to hold up to 63 fields. If we can reduce so that it always remains below 31 fields for the base game then we can save megabytes worth of memory.

PR #4682, opened 2023-02-09
Fix/ai exp formers

relent0r:
Description of changes
This PR changes the base AI experimental air and land platoonformers to no longer require some dubious conditions. 
The original single land experimental had a requirement that the enemies needed to have less and 1 SML and 2 TML else it would not form and you would need to wait until a second experimental unit was built.
The original single air experimental had a requirement that the enemies needed to have less than 3 SAMs and 8 ASF else it would not form and you would need to wait until a second experimental unit was built.

I get why it might seem like a good idea. But the AI will not use the experimentals defensively and they will sit in the base doing nothing until there are multiple. While this will cause a bit of mass donation for experimentals against higher level players it is a better scenario than how it currently work for the other player types.
In the next iteration I'll be looking to shift this sort of logic to the platoon behaviour itself so that the AI can defend its base likely leveraging the future conflict lines as a tether if it is too risky to attack the enemy.
Small upside is less function calls in the conditions too as the previous ones were performing full map scans looking for specific unit categories on every condition monitor loop.

I've made a small adjustment to the instance count so that only 1 single former will operate at a time. So that when/if the AI gets to the point of building multiple experimental it can still use the group former. That should satisfy the players that just want to give the AI massive eco cheats and have giant experimental battles.

There is also one slight fix to the factory upgrades for expansions where if the AI has upraded all base factories to T3 it would stop upgrading expansion bases factories from T1 > T2. Thanks to Grandpa Sawyer for pointing this one out.

Test setup for the changes
Start a skirmish game with the base AI. Spawn in 1x SML and 2x TMLs for the enemy and make sure the AI has them on radar, then spawn in a single land experimental. They should form and attack the enemy.
For the air experimental similar thing, spawn in 3x SAMS and 8X ASF again making sure they are on radar, then spawn in the air experimental. It should form and attack the enemy.

You can also do it the hard way an observe AI games, but that is quite time consuming as these are late game events that require certain circumstances.


Garanas:
Tested it - working great. Experimentals are made and used!

Garanas (approved):

PR #4685, opened 2023-02-10
Add underwater vision to engis

Spikey84:
Adds water vision to each engi the size of their build range. This will make underwater mexes visible when attempting to build them.

### DIFF
--- a/units/UAL0105/UAL0105_unit.bp
+++ b/units/UAL0105/UAL0105_unit.bp
@@ -239,7 +239,7 @@ UnitBlueprint {
     },
     Intel = {
         VisionRadius = 18,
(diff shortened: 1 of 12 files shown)


Spikey84:
It seems the intel values need to be double the build range values, so i doubled all values.

Garanas:
Seems fine, but units that do not have the water vision set in the blueprint get a water vision of 10 by default. Shouldn't they already be able to see the extractors?

Spikey84:
Oh thats strange, is this new to faf dev or always been a thing?

Garanas:
Always been a thing.

Garanas:
Closing this pull request as all engineers have water vision by default.

PR #4686, opened 2023-02-12
Re-implement veterancy system

### LINKED ISSUE #4135: Refactor the syncing of experience of units
Description
-----------

The implementation of the experience (veterancy) system is done via the Sync table. Because of this it includes a lot of additional baggage, which includes:

- 1) A complicated (and expensive) [indexing approach](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L66) for syncing the data with the UI
- 2) A lot of table and field allocations with regard to the `UnitData` table

For 1, note that this is the meta table of `self.Sync` as set [here](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L169). As a consequence, all the index calls run via these functions. And there are a lot of those!

Course of action
------------------

We'll take a similar approach to https://github.com/FAForever/fa/issues/4134, where we use the recently found `Unit:GetStat` and `Unit:SetStat` functions. These allow for direct communication with the UI, without using the sync. As a result we'll need to make the following changes:

- 1) Instead of storing the intermediate status in `self.Sync`, we store it in separate values that we keep track of in `self` 
- 2) We initialize all required values during `OnCreate`, unless the unit does not support keeping track of experience. This includes the amount of experience and the amount required for the next level
- 3) We initialize the stats by calling `unit:GetStat` during OnCreate
- 4) We update all functions related to experience, including:
- - `Unit:CalculateVeterancyLevel` 
- - `Unit:CalculateVeterancyLevelAfterTransfer`
- - `Unit:SetVeterancy`
- 5) We update the UI to use `UserUnit:GetStat` when we want to view the current experience values

Note that 1 and 2 should closely resemble how it works in https://github.com/FAForever/fa/issues/4134, and that 3 is required to prevent an engine bug. We need to try and get the stat (as that prepares the statistic in the engine) before we can set it (which crashes if it tries to read a value that does not exist).

Test plan
---------

The functionality of the experience system should remain the same. Technically you should be able to play a replay without desyncing, as the computation of experience is not adjusted (just the information we send to the UI is). In practice this is impossible due to floating point inaccuracy.

Sanity checklist:
- There should virtually be no more calls to `self.Sync` in the unit class
- At no point can the game do a `Unit:SetStat` on a value that has not been prepared using `Unit:GetStat`

Learning goals
--------------

Tackle large refactoring that spans across the sim to the UI.

Garanas:
Closes #4135 .

Completely re-implements the veterancy system, while keeping all the existing functionality. This doesn't only solve the performance problems as described by #4135, it also fixes a significant memory leak 😄 !

In the old implementation a unit `x` keeps a strong reference to every unit that dealt damage to `x`. A unit usually deals damage to multiple units and receives damage from multiple units. This mimics a direction graph of strong references. And because the references are strong the garbage collector will wait with collecting old instances of units until all the active references are gone. But because usually there are always some units leftover after an engagement there will always be some active refeference. Especially when there is an ACU involved 😄 !

To visualize it, say you have this situation:

image

Because both Mantis deal damage to each other they both generate a strong reference to the instance table of the unit other. Regardless of who survives, the instance table of the unit that died is never cleaned up.

And a more complicated example:

image

There is a cycle in this directional graph. If any of these units survive, the instance table of _all the other units_ remain in memory too because of that strong reference!

And the spill of the directional graph is usually the ACU: it takes damage from a lot of sources, and therefore it kept a strong reference to all those sources. And they usually do not get released when the ACU dies because the it usually dealt damage to units that are still alive after the ACU is destroyed.

Long story short: you end up with a really large directional graph with a lot of strongly connected components. And as long as there is 1 active reference to one of those components the entire graph (and all its instance reference it represents) remain in memory.

### DIFF
--- a/lua/defaultcomponents.lua
+++ b/lua/defaultcomponents.lua
@@ -716,5 +716,16 @@ VeterancyComponent = ClassSimple {
         self.VetExperience = 0
         self.VetLevel = 0
         self:AddVetExperience(self.Blueprint.VetThresholds[math.min(level, 5)] or 0, true)
-    end
+    end,
+
+    ---@deprecated
+    ---@param self Unit | VeterancyComponent
+    GetVeterancyValue = function(self)
+        local fractionComplete = self:GetFractionComplete()
+        local unitMass = self:GetTotalMassCost()
+        local vetMult = self.Blueprint.VeteranImportanceMult or 1
+        local cargoMass = self.cargoMass or 0
+        -- Allow units to count for more or less than their real mass if needed
+        return fractionComplete * unitMass * vetMult + cargoMass
+    end,
 }
--- a/lua/defaultcomponents.lua
+++ b/lua/defaultcomponents.lua
@@ -524,7 +524,6 @@ VeterancyComponent = ClassSimple {
     OnCreate = function(self)
         local blueprint = self.Blueprint
 
-        LOG("OnCreate")
         -- these fields are always required
         self.VetDamageTaken = 0
         self.VetDamage = {}
@@ -629,6 +628,7 @@ VeterancyComponent = ClassSimple {
         end
     end,
 
+    ---@param self Unit | VeterancyComponent
     AddVetLevel = function(self)
         local blueprint = self.Blueprint
         if not blueprint.VetEnabled then
@@ -677,4 +677,44 @@ VeterancyComponent = ClassSimple {
         self:DoUnitCallbacks('OnVeteran')
         self.Brain:OnBrainUnitVeterancyLevel(self, nextLevel)
     end,
+
+    -- kept for backwards compatibility with mods, but should really not be used anymore
+
+    ---@deprecated
+    ---@param self Unit | VeterancyComponent
+    ---@param massKilled number
+    ---@param noLimit boolean
+    CalculateVeterancyLevel = function(self, massKilled, noLimit)
+        self.VetExperience = 0
+        self.VetLevel = 0
+        self:AddVetExperience(massKilled, noLimit)
+    end,
+
+    ---@deprecated
+    ---@param self Unit | VeterancyComponent
+    ---@param massKilled number
+    ---@param noLimit boolean
+    CalculateVeterancyLevelAfterTransfer = function(self, massKilled, noLimit)
+        self.VetExperience = 0
+        self.VetLevel = 0
+        self:AddVetExperience(massKilled, noLimit)
+    end,
+
+    ---@deprecated
+    ---@param self Unit | VeterancyComponent
+    ---@param level number
+    SetVeterancy = function(self, level)
+        self.VetExperience = 0
+        self.VetLevel = 0
+        self:AddVetExperience(self.Blueprint.VetThresholds[math.min(level, 5)] or 0, true)
+    end,
+
+    ---@deprecated
+    ---@param self Unit | VeterancyComponent
+    ---@param level number
+    SetVeteranLevel = function(self, level)
+        self.VetExperience = 0
+        self.VetLevel = 0
+        self:AddVetExperience(self.Blueprint.VetThresholds[math.min(level, 5)] or 0, true)
+    end
 }
--- a/lua/defaultcomponents.lua
+++ b/lua/defaultcomponents.lua
@@ -21,8 +21,7 @@ ShieldEffectsComponent = ClassSimple {
     OnShieldEnabled = function(self)
         self.ShieldEffectsBag:Destroy()
         for _, v in self.ShieldEffects do
-            self.ShieldEffectsBag:Add(CreateAttachedEmitter(self, self.ShieldEffectsBone, self.Army, v):ScaleEmitter(self
-                .ShieldEffectsScale))
+            self.ShieldEffectsBag:Add(CreateAttachedEmitter(self, self.ShieldEffectsBone, self.Army, v):ScaleEmitter(self.ShieldEffectsScale))
         end
     end,
 
@@ -700,7 +699,6 @@ VeterancyComponent = ClassSimple {
         self:AddVetExperience(massKilled, noLimit)
     end,
 
-    ---@deprecated
     ---@param self Unit | VeterancyComponent
     ---@param level number
     SetVeterancy = function(self, level)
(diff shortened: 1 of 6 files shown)


lua/defaultcomponents.lua
```diff
@@ -19,7 +21,8 @@ ShieldEffectsComponent = ClassSimple {
     OnShieldEnabled = function(self)
         self.ShieldEffectsBag:Destroy()
         for _, v in self.ShieldEffects do
-            self.ShieldEffectsBag:Add(CreateAttachedEmitter(self, self.ShieldEffectsBone, self.Army, v):ScaleEmitter(self.ShieldEffectsScale))
+            self.ShieldEffectsBag:Add(CreateAttachedEmitter(self, self.ShieldEffectsBone, self.Army, v):ScaleEmitter(self
+                .ShieldEffectsScale))
```
speed2CZ, line 25: This is very weird line split

lua/defaultcomponents.lua
```diff
@@ -490,3 +493,239 @@ TreadComponent = ClassSimple {
...
+    CalculateVeterancyLevelAfterTransfer = function(self, massKilled, noLimit)
+        self.VetExperience = 0
+        self.VetLevel = 0
+        self:AddVetExperience(massKilled, noLimit)
+    end,
+
+    ---@deprecated
+    ---@param self Unit | VeterancyComponent
+    ---@param level number
+    SetVeterancy = function(self, level)
```
speed2CZ, line 706: Why is this deprecated?

Garanas:
I've done a small analysis.

We'll start with looking at the old code, which had a function such as this:

```lua
    CalculateVeterancyLevel = function(self, massKilled, noLimit)
        if not noLimit then 
            -- Limit the veterancy gain from one kill to one level worth
            massKilled = math.min(massKilled, self.Sync.myValue or self.Sync.manualVeterancy[self.Sync.VeteranLevel + 1])
        end

        -- Total up the mass the unit has killed overall, and store it
        self.Sync.totalMassKilled = math.floor(self.Sync.totalMassKilled + massKilled)

        -- Calculate veterancy level. By default killing your own mass value (Build cost mass * 2 by default) grants a level
        if self.Sync.myValue then
            local newVetLevel = math.min(math.floor(self.Sync.totalMassKilled / self.Sync.myValue), 5)

            -- Bail if our veterancy hasn't increased
            if newVetLevel == self.Sync.VeteranLevel then return end

            -- Update our recorded veterancy level
            self.Sync.VeteranLevel = newVetLevel
        else
            if self.Sync.totalMassKilled - self.Sync.manualVeterancy[self.Sync.VeteranLevel + 1] >= 0 then
                self.Sync.VeteranLevel = self.Sync.VeteranLevel + 1
            else
                return
            end
        end

        self:SetVeteranLevel(self.Sync.VeteranLevel)
    end,
    ```
    
Note all the calls to `self.Sync`, which is a special table: it has a `index` function in its meta table! Therefore each table access is not just a hash key lookup, but it also [this](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L68-L71):

```lua
    __index = function(t, key)
        local id = rawget(t, 'id')
        return UnitData[id].Data[key]
    end,
```

That makes all the code written there a lot more expensive than usual, let alone when you create new entries as that runs [this](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L73-L100):

```lua
    __newindex = function(t, key, val)
         -- globals to locals
        local rawget = rawget
        local UnitData = UnitData

        local id = rawget(t, 'id')
        local army = rawget(t, 'army')
        local unitData = UnitData[id]
        if not unitData then
            unitData = {
                OwnerArmy = army,
                Data = {},
            }
            UnitData[id] = unitData
        end
        unitData.Data[key] = val

        local focus = GetFocusArmy()
        if army == focus or focus == -1 then -- Let observers get unit data
            local SyncUnitData = Sync.UnitData
            unitData = SyncUnitData[id]
            if not unitData then
                unitData = {}
                SyncUnitData[id] = unitData
            end
            unitData[key] = val
        end
    end,
```

And as the sync table resets each tick, all that logic runs essentially each time you end up managing the veterancy state of a unit. Long story short: these functions were expensive. But we're not there yet.

When you add values to the sync of a unit they end up in the `UnitData` table as defined in [simsync.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/SimSync.lua#L19). This table is synced each tick to the UI to a table with the same name in [usersync.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/UserSync.lua#L14). This table is updated by merging in the changes as done [here](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/UserSync.lua#L91-L99):

```lua
    if not table.empty(Sync.UnitData) then
        UnitData = table.merged(UnitData,Sync.UnitData)
    end
```

That means that every tick when _something_ is changed in that table it creates an entire copy of it! As a result of that we end up creating and re-creating a lot of tables, almost every tick. That means that the old implementation:

- Has to deal with all the quirky logic when interacting with the sync
- Creates the same table up to three times: once in the sim-side of the sync, once when receiving it in the user side and then another time when processing it
- Every unit ends up having an entry in this table, and all units can make changes as they are registered (created), unregistered (destroyed) or gain veterancy (kills something)

And the best part is: the _entire_ units table is focus army specific, which means it has to recreate the table from scratch when you switch focus army, see also the [comment in simsync.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/SimSync.lua#L17-L19):

```
-- UnitData that has been synced. We keep a separate copy of this so when we change
-- focus army we can resync the data.
```

All of that on top of the memory hoarding as described in the original topic 😄 !

lua/defaultcomponents.lua
Garanas: I suppose this is one that is ok-isch, but it doesn't quite work as you'd want it to.

speed2CZ: This one is used by the mission scripts to set the veterancy directly to 1-5 levels. How does it work now?

Garanas: The same way - they get vetted as usual

PR #4698, opened 2023-02-17
Fix recall wording and UI bugs

Hdt80bro:
* Fixes the poorly worded recall voting terminology
* Fixes the way the tooltip is handled on the recall buttons (in particular, how the hit test works for disabled buttons in general)
* Removes the extraneous PNG files
* Minor fixes, such as defeat now cancels voting and currently building SCU's don't get teleport effects.


lua/ScenarioFramework.lua
```diff
@@ -934,27 +934,39 @@ end
...
----@param killUnits boolean
+---@param killUnits? boolean
 function FakeTeleportUnits(units, killUnits)
     IssueStop(units)
     IssueClearCommands(units)
     for _, unit in units do
-        unit.CanBeKilled = false
-        unit:PlayTeleportChargeEffects(unit:GetPosition(), unit:GetOrientation())
-        unit:PlayUnitSound('GateCharge')
+        if not unit:IsDead() then
```
Garanas, line 942: Better to use `if not IsDestroyed(unit) then`

Garanas: Same for all other uses of `IsDead()` or `.Dead` with code related to recall

PR #4701, opened 2023-02-17
Refactor TauntManager

4z0t:


Garanas:
Where is this manager used?

4z0t:
@Garanas it is used in most of the coop missions and some maps.

Garanas (approved):
Appears to be just inlining it into a class, no logical changes

PR #4705, opened 2023-02-18
Adjust reclaim values of mobile units based on tech tier

Tagada14:
Reduce the amount of reclaim that units and structures leave behind based on their tech level.

Tech 1 - 0.9 * 0.9 (81%)
Tech 2 - 0.9 * 0.8 (72%)
Tech 3 - 0.9 * 0.7 (63%)
Experimental - 0.9 * 0.6 (54%)

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1547,24 +1547,20 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent) {
         local layer = self.Layer
 
         -- Reduce the mass value based on the tech tier
-        local mass_tech_mult = 1
-        
-        for _, v in bp.Categories do
-            if v == 'TECH1' then
+        -- by default we reduce the mass value 2 times by 90% for a total of 81%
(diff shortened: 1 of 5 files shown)


lua/sim/Unit.lua
```diff
@@ -1848,6 +1848,29 @@ Unit = Class(moho.unit_methods) {
         local pos = self:GetPosition()
         local layer = self.Layer
 
+        -- Reduce the mass value based on the tech tier
+        local mass_tech_mult = 1
+        
+        for _, v in bp.Categories do
```
Garanas, line 1854: You can use the `self.Blueprint.TechCategory` field instead, which is one of these values: ` "TECH1" | "TECH2" | "TECH3" | "EXPERIMENTAL"`

MLGTASTICa:
Shouldn't it be considered to do the opposite , have later tier units leave more mass behind and starting tiers less ? This way a early steamroll doesn't go out of control too much if the opponent reclaims all destroyed units.

MrRowey (approved):

PR #4713, opened 2023-02-19
fixing platform animation

MadMaxFAF:
new anim


MadMaxFAF:
@Garanas this is ready now

PR #4714, opened 2023-02-19
Fix terrain alignment issues

Garanas:
A unit now only orientates to the terrain if it wants to flatten the skirt, with the exception of the player-buildable walls. Units are also no longer orientated when they are build on essentially flat ground.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -78,7 +78,6 @@ StructureUnit = ClassUnit(Unit) {
         end
 
         -- check for terrain orientation
-        local bp = self.Blueprint
         if not (
                 physicsBlueprint.AltitudeToTerrain or
                 physicsBlueprint.StandUpright
@@ -87,8 +86,8 @@ StructureUnit = ClassUnit(Unit) {
             -- rotate structure to match terrain gradient
             local a1, a2 = TerrainUtils.GetTerrainSlopeAngles(
                 self:GetPosition(),
-                bp.Footprint.SizeX or physicsBlueprint.SkirtSizeX,
-                bp.Footprint.SizeZ or physicsBlueprint.SkirtSizeZ
+                blueprint.Footprint.SizeX or physicsBlueprint.SkirtSizeX,
+                blueprint.Footprint.SizeZ or physicsBlueprint.SkirtSizeZ
             )
 
             -- do not orientate structures that are on essentially flat ground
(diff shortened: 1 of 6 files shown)


lua/defaultunits.lua
```diff
@@ -67,35 +69,39 @@ StructureUnit = ClassUnit(Unit) {
...
                 self:GetPosition(),
-                bp.Footprint.SizeX or bp.Physics.SkirtSizeX,
-                bp.Footprint.SizeZ or bp.Physics.SkirtSizeZ
+                blueprint.Footprint.SizeX or physicsBlueprint.SkirtSizeX,
+                blueprint.Footprint.SizeZ or physicsBlueprint.SkirtSizeZ
             )
 
-            self:SetOrientation(EulerToQuaternion(-1 * a1, a2, 0), true)
+            -- do not orientate structures that are on essentially flat ground
+            if MathAbs(a1) > 0.02 or MathAbs(a2) > 0.02 then
```
BlackYps, line 94: This can lead to problems, because the flattened terrain is still at an angle.
grafik

BlackYps:
I think the only other units that don't want to flatten the skirt are T1 pd and T1 aa right? I'm not sure if this is really an improvement, because they are now floating again.
grafik

BlackYps:
Maybe it's worthwhile to explore reducing the angle by which we rotate the structures. This way we can prevent the large terrain deformations that look ugly and are bad for pathfinding, while preventing the appearance of buildings being stuck onto the terrain like a pin.
We then have to flatten the terrain and rotate the building by the same amount.

CyrilPaulus:
I was wondering if instead of rotating the structure, the base of the model could be extended into the ground to compensate for the terrain not being flat.

In the last picture, the "feet" of the UEF turret could extend underground so the turret could still be kept upright but there wouldn't be any visible gap between it and the terrain?

The height of the turret would maybe have to be adjusted so all of its normal model features are visible and only the "foundation" part would go underground.

I guess it would maybe require too much modeling work ?

Garanas:
@BlackYps fixed with #4724

I don't see any particular issues with the current implementation where we slightly tilt the structure. It is barely noticeable in practice and it fixes all the terrain deformation issues we've had up to this point.

@CyrilPaulus not only would it take quite a bit of modelling work, there's also a few edge cases. One example is a structure that can be built on land and water. On top of that we'd increase the size of the repository by adding in all those meshes.

BlackYps:
@Garanas I guess it is a personal preference of aesthetics. Your solution is not objectively flawed

PR #4716, opened 2023-02-20
Update build_templates.lua

The-Balthazar:
Fix for templates created with a structure with a non-standard-centre-offset-to-grid as the first unit having broken offsets and being basically unbuildable.
Structures this affects includes strat defences and T2+ torpedo launchers.


Garanas:
image

Garanas (approved):

PR #4719, opened 2023-02-21
Fix weapon mode switching causing firing delay

RabidPope:
Some weapons that switch firing modes have a delayed activation. Keeping weapon aa mode on helps to fix this.


Garanas:
I'll give this some attention after the patch of today.

Garanas:
Is this ready for a review?

RabidPope:
This is ready for review. Is there a way to set that somewhere?

Garanas:
A pull request van be in draft mode, as chosen when creating the pull request or when switching to it here:

image

Then once you're ready you can remove it from draft mode. In this case the pull request was never in draft mode 😄 !

Garanas:
Interesting bug: the projectiles can be seen through the fog of war 😄 

image

Garanas:
@RabidPope looked into your changes, they appear to be fine. Noticed a small twitch when disabling or enabling a turret. This can be fixed by 'copying' the pitch of the gun that was previously trying to fire. You removed those lines for a reason however, why was that?

Garanas:
Looking into it deeper: I do not understand the approach we've taken in the past to allow the weapon to switch. It seems to me that there's a better approach

RabidPope:
@Garanas I removed those lines because they did not seem to help with the twitching turret. By all means if you can use a better approach, it would be very desirable.

Garanas:
@RabidPope turned it into a separate branch so that we do not pollute this one too much, you can find it here:

- https://github.com/FAForever/fa/pull/4765/files#diff-c5a7bbf020feab2e5215fd2e4fb3468a3e307c9250ec65a0403d024574951921

RabidPope:
I tried out the changes on the skyslammer. The turret is still twitchy. The siren's turret does not seem to be twitchy and don't think it was.

Garanas:
I'll look at the sky slammer later today.

Garanas:
https://user-images.githubusercontent.com/15778155/222921727-d124f728-a12c-44fe-8c10-b069922967f0.mp4

@RabidPope I think this is it, could you check and confirm that the guns feel natural now?

RabidPope:
Thanx. Yep that fixed it. Good job!

Garanas:
I'm not sure if we can apply the same logic to the Cybran T3 MMA. Unlike the other two, I recall the Cybran T3 MMA has an actual laser pointer

Garanas:
Great work @RabidPope 👍

PR #4724, opened 2023-02-23
Fix issues with terrain alignment of structures

Garanas:
Excludes non-land based structures and the orientation is only not applied when the structure is on perfectly flat ground.


BlackYps:
these are missing the T1 AA

PR #4729, opened 2023-02-23
Removal of Teleport Modification

MrRowey:
Removed the Teleport Modification lines from units that cant even teleport


Hdt80bro:
Do any mods rely on these values for when they teleport units that aren't able to normally in the base game?

Garanas:
@Hdt80bro is correct, it is not used by the base game but there are mods that use it (BlackOps / Survival Mayhem). It doesn't seem worth removing these

PR #4733, opened 2023-02-25
Update build_templates.lua

The-Balthazar:
Fixed incorrect axes on template save offset. Also renamed `s1Yoffset` to `s1Zoffset`.
Template co-ords are Z, X, not X, Z.
This would only come up on units that need a different offset for each axis.

### DIFF
--- a/lua/ui/game/build_templates.lua
+++ b/lua/ui/game/build_templates.lua
@@ -20,12 +20,12 @@ function CreateBuildTemplate()
     if next(template) then
         local str1bp = __blueprints[ template[3][1] ]
         local s1Xoffset = TemplateAxisOffset(str1bp, 'SizeX')
-        local s1Yoffset = TemplateAxisOffset(str1bp, 'SizeZ')
+        local s1Zoffset = TemplateAxisOffset(str1bp, 'SizeZ')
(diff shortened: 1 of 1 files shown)


ageekhere:
Some ideas on optimization and s1Yoffset was left in the if 

```
function CreateBuildTemplate()
    GenerateBuildTemplateFromSelection()
    local template = GetActiveBuildTemplate()
    ClearBuildTemplates()
    if #template > 0 then --remove next as it is an expensive funcation to call and can be avoided by checking length of array directly
        local str1bp = __blueprints[template[3][1]]
        local s1Xoffset, s1Zoffset = TemplateAxisOffset(str1bp, 'SizeX'), TemplateAxisOffset(str1bp, 'SizeZ')
        if s1Xoffset ~= 0 or s1Zoffset ~= 0 then -- is s1Yoffset meant to be s1Zoffset? maybe 'and' could be used here?
            for i, str in ipairs(template) do --ipairs loops are generally faster than the table.getn function.
                str[3], str[4] = str[3] + s1Zoffset, str[4] + s1Xoffset
            end
        end
        AddTemplate(template)
    end
end
```

The-Balthazar:
Turns out it was `AtCenterIn` thatz Z, X, not templates.

PR #4747, opened 2023-02-26
Some changes for the Shard

Basilisk3:
**Display:**
UniformScale: 0.06 --> 0.09	

The new shard is now about half the length of a Beacon Class, instead of almost only a quarter. This makes it possible for it to be hit more consistently. The size of the selection box is adjusted accordingly, so it still makes sense with the new size of the unit.

**Physics:**
Speed: 8 --> 7

The speed is reduced, so it does not drive in front of frigates and get destroyed as often. It is still the fastest naval unit in the game. The other variables relating to speed are adjusted accordingly.

**Economy:**
Energy: 1200 --> 1100
Mass: 120 --> 110
Buildtime: 800 --> 550

To make up for this, it gets a slight reduction in price and the unusually bad build time to cost ratio gets reduced. It now costs the same as 2 tech1 MAAs and has the same build time to cost ratio as the Beacon Class. This makes the unit more viable to be mixed in with frigates.


MrRowey:
This Requires Approval from the Balance team @Tagada14

Tagada14:
We will discuss this PR during a balance meeting next week.

Spikey84:
We discussed the idea of Shard change and came up with the ones mentioned in #4981. So we are going to close this one.

PR #4748, opened 2023-02-26
Add support for small state machines for AI

Garanas:


relent0r:
Jip are you able to provide a little blurb on how an AI dev would initialize a state machine from the skirmish AI? I'm only used to getting platoons to do things via the platoon former manager.

Garanas:
@relent0r can you show me where the current platoon AI is applied to a platoon?

relent0r:
The current entry point is here.
https://github.com/FAForever/fa/blob/17be322ef4b980476b2e036975cb7c46570ff61f/lua/sim/PlatoonFormManager.lua#L123

So we assume that the Builder has passed all the requirements. It hits this, sets a few bits. Then checks if it can form the platoon based on the template provided and the units available using the CanFormPlatoon. Then it uses the FormPlatoon function and does a SetAIPlan based on the function in the platoontemplate's Plan property, it will fork a function within the platoon.lua file.
Then if there are any addons or behaviors it does that afterwards.

ageekhere:
Some suggestions

```
        self.Identifier = string.format("Default - %d", GetUniqueIdentifier()) --removed self.Identifier, it may be unnecessary? 
	--move local to top
	--use pairs
	ToSync = function(self, key, value)
        local tick = GetGameTick()
        local syncState = self.StateSync
		local sync = Sync
        local syncStateMachines = sync.StateMachines or {}
        if syncState.Tick < tick then
            for k,v in pairs(syncState) do 
                syncState[k] = nil
            end
            syncState.Tick = tick
        end

    end,
```

Garanas:
@ageekhere we're not at the phase to optimize yet, first rule of improving the performance of any snippet of code is to have it functional first 😉

Garanas:
Closed in favor of the one introduced by #4858

PR #4764, opened 2023-03-02
Range Bots Improvements

RabidPope:
Changes for the hoplite and mongoose.
Increased turret traversal and decreased turn speed so the units don't stop firing when making sharp turns.
The mongoose had a 2 second firing delay for some reason. Changed this to 0.1.
The hoplite turret seems to be unable to stay on target unless yaw speed is increased significantly higher than turn speed.
Also slightly increased hoplite projectile velocity as even very slow units such as coms can easily dodge.


Garanas:
These are balance changes, they should point towards `deploy/fafbeta`

> The mongoose had a 2 second firing delay for some reason. Changed this to 0.1.

This is because of the animation, where the gun is starting to rotate faster and faster. Lowering this to `0.1` is undesired in my opinion

RabidPope:
> These are balance changes, they should point towards `deploy/fafbeta`
> 
> > The mongoose had a 2 second firing delay for some reason. Changed this to 0.1.
> 
> This is because of the animation, where the gun is starting to rotate faster and faster. Lowering this to `0.1` is undesired in my opinion

@Garanas I did realise this. It's just that 2 whole seconds of delay when first entering range seems a bit extreme. It would be kind of cool if the gun fires slowly at first and then picks up speed over that 2 seconds, but I am not sure if that can be done.

Garanas:
You may want to rebase the branch onto `deploy/fafbeta` too. Just pointing to `deploy/fafbeta` introduces a heap of commits 😄

PR #4765, opened 2023-03-02
Alternative implementation of cruiser 

Garanas:


Garanas:
Also noticed an interesting bug with the class / state system at this point:

- https://github.com/FAForever/fa/blob/feature/cruiser-aa-ground-gun/lua/sim/defaultweapons.lua#L728

Where it calls the function `OnGotTarget` of the weapon class, but it should be the `OnGotTarget` of whatever base class the state has. I don't know how to approach that yet 😄

PR #4766, opened 2023-03-02
Range bot improvements

RabidPope:
Changes for the hoplite and mongoose.
Increased turret traversal and decreased turn speed so the units don't stop firing when making sharp turns.
The mongoose has a 2 second firing delay for some reason.
The hoplite turret seems to be unable to stay on target unless yaw speed is increased significantly higher than turn speed.
Also slightly increased hoplite projectile velocity as even very slow units such as coms can easily dodge.


Garanas:
Better 👍 , thank you for rebasing it

Garanas:
@Tagada14 all yours

RabidPope:
@Garanas @Tagada14 Would it be reasonable to reduce the mongoose weapon charge time?

Garanas:
You'd need the animation to match. Reducing it to `0.1` is not desireable, but lowering it from `2` to `1` and then speeding up the animation seems fine to me.

Tagada14:
We are discussing proposed changes, they look pretty good to me.

RabidPope:
@Tagada14 So any feedback on if this is desirable?

MrRowey:
@RabidPope were active getting thought the backlog this will be brought up on this Sundays meeting Sorry for this take so long

MrRowey:
@RabidPope we agree with this but we would prefer the `MuzzleVelocityRandom = 0.5,` for the mongoose

units/DEL0204/DEL0204_unit.bp
```diff
@@ -328,7 +328,7 @@ UnitBlueprint {
             MuzzleSalvoDelay = 0.3,
             MuzzleSalvoSize = 4,
             MuzzleVelocity = 12.8,
-            MuzzleVelocityRandomness = 0.5,
+            MuzzleVelocityRandom = 0.5,
```
Garanas, line 331: The name of the field is different. Is this a typo?

RabidPope: Just changed it

RabidPope: Its "random" on live and "randomness" on beta. Do I just keep it as randomness?

ComradeStryker:
Should the ACU not also get this change?
They suffer from the same issue.

PR #4770, opened 2023-03-03
Cheatspawn menu overhaul

The-Balthazar:
Porting across my debug spawn menu changes from here https://github.com/The-Balthazar/BrewLAN/commit/23479f430ee98cd27795c3f091a93dd24a7a221d


Garanas:
@The-Balthazar there's still a few issues with this, do you have time to look into it?

The-Balthazar:
What are the issues?

Garanas (approved):

lua/SimCallbacks.lua
```diff
@@ -420,6 +420,8 @@ local function SetWorldCameraToUnitIconAngle(location, zoom)
 end
 
 Callbacks.ClearSpawnedMeshes = function()
+    if not CheatsEnabled() then return end
```
The-Balthazar, line 423: Unless there's a way to spawn mesh entities that get included in SpawnedMeshes without cheating already enabled, this just stops it looping over an empty table.

Garanas: I'm aware - but it is an interaction from a cheat menu. Least we can do is let people know

PR #4787, opened 2023-03-05
Clear `[0]`(and others) indexes in save files

4z0t:


Garanas:
Those are odd, @relent0r any idea whether this would have an impact?

relent0r:
It looks intentional at first glance. Since most of lua doesn't support zero indexes. The only lua based code I see which is looking at the indexes is here.

https://github.com/FAForever/fa/blob/b5d9e27e6f0bfaad8204682a4f94ebc7197073ca/lua/AI/OpAI/BaseOpAI.lua#L439
and some functions in Scenario Utilities

But then I don't see why it would want a zero index number. I can't see an example of it but my only thoughts would be that these may be or may previously have been passed to the C engine to do something which is why they required a zero index.

4z0t:
I think this came from C and that's why it is 0 indexed, I don't see any actual intention here. Am I correct that it allocates hash part, causing usage of more memory then could be used in array part?

relent0r:
That would make sense re : originally coming from C. I remember they didn't implement lua until the E3 demo so its quite possible some stuff was previously in C and just yanked across.

We'll just need to figure out which campaign missions are using these builders so we can log and confirm functionality.

4z0t:
Any progress of this? idk if we can test any performance gain, but I can tell that it makes it less confusing to look at.

Garanas:
I'll merge it after the upcoming hotfix

Garanas:
@relent0r I'm merging this in, if you find any odd behavior then this this pull request should be the first thing on your mind

PR #4789, opened 2023-03-05
Forward ACU kills to the user layer

rigomate:
Currently the user layer does not know who performed a kill on an ACU.
This information however is not supposed to be hidden, as it is clearly a UI related thing.

This information actually used to be available, until this big change broke this unintentionally:
https://github.com/FAForever/fa/pull/4016

This leads to a side effect, that e.g the Supreme Scoreboard Mod assumes that every kill was a ctrl+k, even if it was a normal kill.
Most of the streamers use this mod, and rely on this information, especially if there is a lot of going on. Like in 8vs8 games.

Using this change the Supreme Scoreboard Mod can also be fixed very easily to show the kills once again correctly.

Please look at the changes and tell me if fits the behavior of existing code.

I basically copied the existing Sync.Teamkill and used its interface.

### DIFF
--- a/lua/UserSync.lua
+++ b/lua/UserSync.lua
@@ -415,6 +415,11 @@ function OnSync()
         import("/lua/ui/dialogs/hotstats.lua").scoreData = Sync.ScoreAccum
     end
 
+    if Sync.AcuKill then
+        local victim, instigator = Sync.AcuKill.victim, Sync.AcuKill.instigator
+        import("/lua/ui/game/gameresult.lua").DoGameResultKills(victim, instigator)
+    end
+
     -- Game <-> server communications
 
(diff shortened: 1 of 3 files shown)


Garanas:
That was fast 👍 , I'll have time to look into it later this week.

Garanas:
@rigomate you can find an example usage in [0c4ee6c](https://github.com/FAForever/fa/pull/4789/commits/0c4ee6cfa6c8bfae5fc81376c853d6f4943bd758). I've generalized it because there is a general notion to communicate more events. I'm merging this in, that would allow you to fix the Supreme Scoreboard.

For the future I'd like to have more events, perhaps you're interested in shaping the architecture for it and implementing it? I'll be there to help, of course

Garanas:
Yes 👍 , you can do that via the game options. See also `options.lua`

PR #4790, opened 2023-03-09
An attempt to optimize DAA0206_script.lua 

ageekhere:
This is my first pull request attempt at optimizing lua. I have tried to cache classes and optimize the code. 

Feedback welcomed.

### DIFF
--- a/units/DAA0206/DAA0206_script.lua
+++ b/units/DAA0206/DAA0206_script.lua
@@ -4,20 +4,16 @@
 -- Summary : Aeon Guided Missile Script
 -- Copyright © 2007 Gas Powered Games, Inc.  All rights reserved.
 ------------------------------------------------------------------------------
-
 local AAirUnits = import("/lua/aeonunits.lua")
 local DefaultWeapons = import("/lua/sim/defaultweapons.lua")
--- Kept for Mod Backwards Compatablity
-local EffectTemplate = import('/lua/EffectTemplates.lua')
-local EffectUtils = import('/lua/effectutilities.lua')
-local weapon = ClassWeapon(DefaultWeapons.DefaultProjectileWeapon) {} --cached
+local weapon = ClassWeapon(DefaultWeapons.DefaultProjectileWeapon) {}
 
 local function f_OnRunOutOfFuel(self)
     self:Kill()
 end
 
 local function f_projectileFired(self)
-    local playSound = self.PlayUnitSound --maybe could be cached global?
+    local playSound = self.PlayUnitSound
     weapon.IdleState.Main = nil
     playSound(self, 'Killed')
     playSound(self, 'Destroyed')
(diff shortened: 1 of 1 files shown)


MrRowey:
@ageekhere the imports that were at the bottom for mod support are Thier as they are loaded last as lua files load top to bottom so by having them at the bottom it saves losing them at the beginning of Thier not being used

MrRowey:
Also have you ensured that the changes made the unit still work the same as before 🙂

ageekhere:
> @ageekhere the imports that were at the bottom for mod support are Thier as they are loaded last as lua files load top to bottom so by having them at the bottom it saves losing them at the beginning of Thier not being used

Hmm I thought because the backwards Compatibility imports are always loaded that it does not really matter if they are at the beginning or the end as they are always imported.
Maybe a condition check could be added to check if theses Backwards Compatibility imports should be used or not, if not they can be moved to the bottom.

> Also have you ensured that the changes made the unit still work the same as before 🙂
I have tested it and it seems to be working as before, though it would be good to have others check it as I am still new at this.

MrRowey:
We move unused import to the bottom for Modi g support as that fine is currently not using them in the faf version, but a mod may use is which is why we add it to the bottom as then a mode hooks the lua file if the unit it added to the bottom if it.

ageekhere:
> Also are you using the Visual Studio code Intelligence plugin we have which dose help.

Not yet, was just using notepad++ will set it up

Garanas:
@ageekhere could you describe why you turned the functions into local functions?

speed2CZ:
If you cache the weapon, won't then all the unit's use the same instance of the weapon?

ageekhere:
> @ageekhere could you describe why you turned the functions into local functions?

f_OnRunOutOfFuel and f_projectileFired are local so they are only available within the scope of the DAA0206 class and also the functions do not need to be created on each call (from my understanding).

I read that global functions require a table lookup to access them, while local ones can be accessed directly so by making DAA0206 local can give a performance boost.

I am still new to this so I am still learning.

I think there could still be some improvements to the code like maybe caching the PlayUnitSound as a global variable rather than in f_projectileFired however i was not sure if that was a good idea.

Garanas:
> I am still new to this so I am still learning.

That is why I ask 😉 

The class is created only once, but instantiated from many times throughout the game. The `ClassUnit` keyword references to the class creation for units, as defined in `class.lua`. It is performed as the file is loaded the first time when the first instance of this unit is created. The functions in the class body are not re-created for each instance. The class acts as a metatable for the instance, and therefore it is merely referenced.

Long story short: outlining the classes like this won't change the situation.

I'll look into the other changes this weekend and give you some more info on what they do. That will help you learn for future changes

Garanas:
@ageekhere I'm going to close this for now

PR #4796, opened 2023-03-10
Air Staging build time changes

Basilisk3:
**Economy:**
Buildtime of all Air Staging Facilities: 450 --> 350

FAF changed the tech level of air staging facilities to tier 1, however their build time remained very high (450), even though at 175 mass they are not particularly expensive. To put this into perspective: A tech 1 land factory costs 240 mass, but a tech 1 engineer still builds it 30 seconds faster than an air staging facility. With these changes, air staging facilities now have a build time to mass cost ratio of 2/1.


Tagada14:
Good catch

Tagada14 (approved):

PR #4801, opened 2023-03-12
Refactor `statusbar.lua` to improve performance and readability

Garanas:
Improves annotations while improving the performance of the interface, such as:

- https://github.com/FAForever/fa/wiki/Performance-and-hash-operations

### DIFF
--- a/lua/maui/statusbar.lua
+++ b/lua/maui/statusbar.lua
@@ -45,7 +45,7 @@ StatusBar = ClassUI(Bitmap) {
                         if not self._stretch then
                             self._bar:SetUV(0, 1 - rangePercent, 1, 1)
                         end
-                        return MathFloor(self.Top() + (rangePercent * (self.Bottom() - self.Top())))
+                        return (self.Top() + (rangePercent * (self.Bottom() - self.Top()))) ^ 0
                     end
                 )
             else
@@ -56,7 +56,7 @@ StatusBar = ClassUI(Bitmap) {
                         if not self._stretch then
                             self._bar:SetUV(0, 0, 1, rangePercent)
                         end
-                        return MathFloor(self.Bottom() - (rangePercent * (self.Bottom() - self.Top())))
+                        return (self.Bottom() - (rangePercent * (self.Bottom() - self.Top()))) ^ 0
                     end
                 )
             end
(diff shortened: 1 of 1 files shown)


ageekhere:
Can math.floor be replaced with integer division for more performance?

Something like
```
return MathFloor(self.Top() + (rangePercent * (self.Bottom() - self.Top()))) --old
 
return (self.Bottom() - (rangePercent * (self.Bottom() - self.Top()))) // 1 --new
```

lua/maui/statusbar.lua
```diff
@@ -1,98 +1,163 @@
...
+        if minSlidePercent == 0 then
+            return (value - min) / (range)
         else
-            return math.floor(((self._value() - self._rangeMin) / (self._range)) / self._minSlidePercent) * self._minSlidePercent
+            return MathFloor(((value - min) / (range)) / minSlidePercent) * minSlidePercent
         end
     end,
-    
+
+    --- Contraints the parameter to the min / max values of the status bar
```
Askaholic, line 145: maybe?

```suggestion
    --- Constrains the parameter to the min / max values of the status bar
```

Hdt80bro:
@ageekhere Our version of Lua is based on Lua 5.0, so we don't have floor division. This isn't a spot we need the hyperoptimization, but since LuaPlus adds bitwise operators, we could use the XOR operator with 0 `a ^ 0` to coerce to an integer (but only for positive numbers which isn't always guaranteed) to achieve the same effect - there's actually a few other optimizations we could use if this code was very performance-critical, but since it's not, it's probably better to leave the code still readable and somewhat unoptimized.

lua/maui/statusbar.lua
```diff
@@ -1,98 +1,163 @@
...
-        self._value:Set(self:_Constrain(value))
+        local min = self._rangeMin
+        local max = self._rangeMax
+        local lazy = self._value
+
+        if value < min then
+            value = min
+        end
+
+        if value > max then
```
Hdt80bro, line 105: `elseif`

lua/maui/statusbar.lua
```diff
@@ -1,98 +1,163 @@
...
+            value = max
+        end
+
+        lazy:SetValue(value)
```
Hdt80bro, line 109: the `self._value` field doesn't need to be stored up above, I'd prefer
```lua
self._value:SetValue(value)
```

Hdt80bro (commented):
For readability, I generally prefer to use my horizontal space before I start adding vertical space ;-)

PR #4805, opened 2023-03-14
Fix bug where orientations are applied multiple times

Garanas:
image


Garanas:
Not sure what happened, but I managed to copy #4800 into this pull request

PR #4808, opened 2023-03-14
Adding Omni E Cost to Novax Satalite

MrRowey:
In the current format this give the Novax Satalite Free Standard radar with no operational cost. where as the Omni will Consumes 350 E to keep operational.

Also the Formatted the Script file

### DIFF
--- a/units/XEA0002/XEA0002_script.lua
+++ b/units/XEA0002/XEA0002_script.lua
@@ -12,19 +12,24 @@ local TOrbitalDeathLaserBeamWeapon = import("/lua/terranweapons.lua").TOrbitalDe
 XEA0002 = Class(TAirUnit) {
     DestroyNoFallRandomChance = 1.1,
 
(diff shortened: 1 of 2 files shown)


Spikey84:
Seems like if we want to add this it should be more than 350 e drain.

Hdt80bro:
Considering how much smaller the Omni radius is than an Omni Suite and that radar isn't part of the maintenance cost, 350 e makes sense unless there's specific gameplay aspects to adding the energy maintenance that require it to be higher (such as the advantage gained from having uncontested omni coverage in enemy territory needing to be cost balanced - this is actually the only reason I can think of making the Novax's omni cost anything in the first place as all other units with intelligence seem to do just fine without one).

Tagada14:
I don't see why we should want that honestly. I don't think Novax needs a nerf and this will lead to weird things like turning off OMNI for all but one Novaxes if they are in 1 place. It would make more sense if it would cost E to fire the weapon like an arty.

PR #4810, opened 2023-03-14
Refactoring LazyVars

Garanas:
Initial setup to refactor LazyVars so that they re-use memory instead of allocating new memory all the time.

### DIFF
--- a/lua/lazyvar.lua
+++ b/lua/lazyvar.lua
@@ -16,22 +16,36 @@ local WeakKeyMeta = { __mode = 'k' }
 ---@alias Lazy<T> T | LazyVar<T> | fun(): T
 
 ---@class LazyVar<T> : Destroyable, OnDirtyListener, function
----@field [1] any           # Cached result of what we represent
----@field busy? boolean
----@field trace string
----@field compute function
----@field uses table<LazyVar, boolean>
----@field used_by table<LazyVar, boolean>
 ---@field OnDirty? function
+---@field [1] any                           # Cached result of what we represent
+---@field [2] boolean                       # Flag whether the lazy var is busy
+---@field [3] string                        # Optional trace when `ExtendedErrorMessages` is set
+---@field [4] function                      # Optional compute value when lazy var represents a function
+---@field [5] table<LazyVar, boolean>       # Table of lazy vars that we use
+---@field [6] table<LazyVar, boolean>       # Table of lazy vars that use us
+---@field [7] LazyVar?       # Lazy var that is dirty
+---@field [8] LazyVar?       # Lazy var that is dirty
+---@field [9] LazyVar?       # Lazy var that is dirty
+---@field [10] LazyVar?      # Lazy var that is dirty
+---@field [11] LazyVar?      # Lazy var that is dirty
+---@field [12] LazyVar?      # Lazy var that is dirty
+---@field [13] LazyVar?      # Lazy var that is dirty
+---@field [14] LazyVar?      # Lazy var that is dirty
+---@field [15] LazyVar?      # Lazy var that is dirty
+---@field [16] LazyVar?      # Lazy var that is dirty
+---@field [17] LazyVar?      # Lazy var that is dirty
+---@field [18] LazyVar?      # Lazy var that is dirty
+---@field [19] LazyVar?      # Lazy var that is dirty
+---@field [20] LazyVar?      # Lazy var that is dirty
 LazyVarMetaTable = {
     __call = function(self)
         local value = self[1]
         if value == nil then
 
             -- track circular dependencies and error them out
-            if self.busy then
-                local trace = self.trace
-                if self.compute then
+            if self[2] then
+                local trace = self[3]
+                if self[4] then
                     trace = trace or "[Set lazyvar.ExtendedErrorMessages for extra trace info]"
                 else
                     trace = trace or ""
@@ -40,10 +54,10 @@ LazyVarMetaTable = {
             end
 
             -- clean up where we're used and what we use
-            local uses = self.uses
+            local uses = self[5]
             if not TableEmpty(uses) then
                 for use in uses do
-                    use.used_by[self] = nil
+                    use[6][self] = nil
                 end
 
                 for k, v in uses do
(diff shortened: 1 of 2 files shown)


Hdt80bro:
Looks like this is mostly just moving the functions into the table. If you're going to remove the annotation hack I did, [these lines](https://github.com/FAForever/fa/pull/4810/files#diff-400e5297b564eef68495418887792b9916dd8d86033ff35cb08c069416aaa4e6R23-R24) describing it should go as well. If we're not worried about stack size and want to remove the `do ... end` blocks for readability like you did [here](https://github.com/FAForever/fa/pull/4810/files#diff-400e5297b564eef68495418887792b9916dd8d86033ff35cb08c069416aaa4e6L235-L239), there's a lot of other places we can do that in this file as well.

Garanas:
This is a mere setup, I couldn't think properly in the old setup. Next steps is casually applying the changes we described. I'll look into your suggestions

Hdt80bro:
`OnDirty` is actually just an event listener and doesn't do much anyway - the `used_by` list is the one that's set automatically from context when used in the computation of another lazyvar and the table that's getting replaced here.

Hdt80bro:
Oh you're right, it dirties the lazyvars from `used_by` directly and only adds stuff to the list if it has an `OnDirty` callback. I wonder if integrating one of those tables instead would be better memory-wise and instead we reuse the same dirtied table for all lazyvars to track `OnDirty` callbacks.

lua/ui/game/worldview.lua
```diff
@@ -186,14 +186,14 @@ local origBottom = false
 
 function Expand()
     if viewRight then
```
4z0t, line 188: Didn't know that this gross exists...

Hdt80bro: I'd recommend encapsulating the old `compute` field into an understandable getter method, just something as simple as
```lua
function LazyVarMetaTable:GetComputeFunction()
    return self[4]
end
```
so the magic numbers don't bleed over

Garanas: @Hdt80bro that is smart 👍 , I'll do that

Garanas:
> Oh you're right, it dirties the lazyvars from `used_by` directly and only adds stuff to the list if it has an `OnDirty` callback. I wonder if integrating one of those tables instead would be better memory-wise and instead we reuse the same dirtied table for all lazyvars to track `OnDirty` callbacks.

I'm not sure how well that would work with the idea of the keys being weak

Hdt80bro:
If we empty the table anyway, it doesn't matter

Garanas:
I was more thinkng of the `OnDirty` key being weak too, and since there is usually no other reference to the function it may ditch the function

lua/lazyvar.lua
```diff
@@ -13,232 +9,268 @@ local setmetatable = setmetatable
...
+
+    --- Lazy variable represents a function, the result is cached
+    ---@param self LazyVar
+    ---@param func function
+    SetFunction = function(self, func)
+        -- do not allow nils
+        if func == nil then
+            error("You are attempting to set a LazyVar's evaluation function to nil, don't do that!")
+            return
         end
```
4z0t, line 138: Use `assert`

lua/lazyvar.lua
```diff
@@ -13,232 +9,268 @@ local setmetatable = setmetatable
...
+    ---@param self LazyVar
+    ---@param onDirtyList LazyVar[]
+    ---@return number
+    SetDirty = function(self, onDirtyList, head)
+        if self[1] ~= nil then
+
+            if self.OnDirty then
+                onDirtyList[head] = self
+                head = head + 1
+                LOG("OnDirty")
```
4z0t, line 118: Logging

4z0t (changes requested):

Hdt80bro:
The `OnDirty` key? I'm not sure I entirely understand, but it sounds like you're talking about `used_by` and `uses` having weak keys (not values), which isn't something I had considered. However, as long only the keys are weak, then the other lazyvar indices in them can be collected just fine if move to the lazyvar while `OnDirty` is kept (because the string is used too much to be collected).

lua/lazyvar.lua
Hdt80bro: It is expected to be an unlikely branch, so using the cheaper `if` statement should be faster.

4z0t: I would say that assert us better here from readable perspective and performance as well.

Garanas: Is assert an engine call?

Hdt80bro:
I actually don't think strings get collected at all, so if the values are strong, then it should be fine.

lua/lazyvar.lua
Hdt80bro: I agree that it'd be more readable, but unfortunately the file is being hyperoptimized due to how much it gets used so that's not a priority. By theory, I don't agree that all of the extra setup required for a global call, function call setup on the Lua end, the C <-> Lua interop, and cleanup would be more performant than a simple `if` statement that bakes the comparison into the jump. If it actually is noticeable in practice though, I think those would be good benchmarks to have.

The other interesting thing to consider is that we don't use `assert` at all in our game code, so it seems that it'd go against the implicit conventions we have.

lua/ui/game/worldview.lua
```diff
@@ -186,14 +186,14 @@ local origBottom = false
...
-        origRightLeft = viewRight.Left.compute
-        origRightRight = viewRight.Right.compute
-        origRightTop = viewRight.Top.compute
-        origRightBottom = viewRight.Bottom.compute
-        origLeft = viewLeft.Left.compute
```
Garanas, line 193: Can someone help explain to me what is happening here, why does this code exist?

Hdt80bro: It makes sure that the left and right views are setup to the frame for the correct number of views visible. It does this by retaining the original layout functions applied to the views (which appear to be just the stock `return parent.Left()` functions) when the UI is show and applying the layouts on the views (half the root frame for dual view, full otherwise). It then reverts when the UI is hidden, hiding the world views. I'm actually not sure when `GetFrame(parentForFrame:GetRootFrame():GetTargetHead())` wouldn't just be `parentForFrame:GetRootFrame()` so it appears like a logically sound but redundant precaution.

Garanas:
@Hdt80bro / @4z0t I'll be merging this later today. The code is a lot more difficult to understand, but I don't feel like it requires any future changes. Especially after the changes of 4z0t where it now also supports the value `false`

PR #4811, opened 2023-03-15
Initial setup for a reclaim tracking system for AIs

Garanas:
Introduces a basic data structure to keep track of reclaim. The data structure mimics the iMAP configuration of the map. 

As reclaim is changed the data structure receives a signal to update the cell. The cell update is deferred by a few ticks to catch any other changes. Every five ticks (0.5 seconds) the data structure updates all changed cells. With four AIs and a lot of fighting on Open Palms the update of cells take less than a millisecond to complete. 

https://user-images.githubusercontent.com/15778155/225286939-ffdbaf06-3513-4096-a80a-7e42a477e688.mp4

A brain can register itself to receive a signal when a cell is updated. Updates are received via:

```lua
    ---@param self AIBrain
    ---@param grid AIGridReclaim
    ---@param cell AIGridReclaimCell
    OnReclaimUpdate = function(self, grid, cell)
        -- do things with it
    end,
```

Via this the brain can trigger other changes. It contains the reclaim grid and the cell that has been updated. A cell has the following fields:

```lua
---@class AIGridReclaimCell : AIGridCell
---@field TotalMass number
---@field TotalEnergy number
---@field TotalTime number
---@field ReclaimCount number
---@field Reclaim table<EntityId, Prop>
```

Note that this is a so-called `hot path` or `hot spot`, see also;

- https://en.wikipedia.org/wiki/Hot_spot_(computer_programming)

The callback works for small updates, but large computations should be deferred.

### DIFF
--- a/lua/AI/Grid.lua
+++ b/lua/AI/Grid.lua
@@ -1,0 +1,113 @@
+
+local tl = { 0, 0, 0 }
+local tr = { 0, 0, 0 }
+local bl = { 0, 0, 0 }
+local br = { 0, 0, 0 }
+
+---@class AIGridCell
+---@field Identifier string
+---@field X number
+---@field Z number
+
+---@class AIGrid
+---@field Cells table[][]
+---@field CellCount number
+---@field CellSize number
+Grid = ClassSimple {
+
+    ---@param self AIGrid
+    __init = function(self)
+        local cellCount = 16
+        local mx = ScenarioInfo.size[1]
+        local mz = ScenarioInfo.size[2]
+
+        -- smaller maps have 8x8 iMAP setup
+        if mx == mz and mx == 256 then
+            cellCount = 8
+        end
+
+        local cells = {}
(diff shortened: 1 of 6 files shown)


relent0r:
I did a proof of concept for an engineer behaviour based on this.

Logic was.
The engineer would get its current cell x,z then look for reclaim, in that square. 
If none found it would expand the search cell by cell until it found a one with reclaim.
It would then assign itself to that cell in the intel grid and move towards it until it was less than the cell size (e.g 32 units for a 10km map), then it would attack move to the cell position and either wait 30 seconds or until there was no more reclaim in the cell according to the intel grid that is being updated by the callback.

Worked really well as far as attack move based reclaim methods work and proved the functionality of the reclaimgrid.

image

relent0r:
Just thinking about it. If I want to keep things consistent I should be inheriting the Grid class into the intel framework for the initial grid generation rather than doing the same thing in another class.

PR #4827, opened 2023-03-22
BaseTemplates.lua memory lazy evaluation

4z0t:
A magick file that contains 40k+ rows of data. And it is being unused during regular games without AI... 
The solution was to determine what parts of the file are the same and then use lazy evaluation to allocate data when needed. Attempt of using `lazyimport` wasn't successful because some AI mods were using that file, but it wasnt imported causing bugs.
~~Please dont ask anything about formatting.~~


Garanas:
> And it is being unused during regular games without AI...

The file is not loaded during non-ai games, all the imports were inlined

Garanas:
@relent0r I don't know how to confirm these changes, do you have a suggestion?

4z0t:
> The file is not loaded during non-ai games, all the imports were inlined

Whats the point then trying to optimize it?

relent0r:
So the AI reference these by using a named key for the type then a faction index that correlates to the table index.

I'll be honest I don't understand what 4z0t is doing here, looks like black magic. Perhaps Sprouto could give better input than me.

The testing that I think we'll want is from mods that go and manipulate the basetemplates file. From memory Draguns SCTA one does and likely others that add new structures.

4z0t:
I tested SCTA and it didnt error.

Garanas:
Closed this in favor of #4864 . In practice about 70% of the file was not referenced. We decided to just remove the unreferenced bits. As a consequence there are no similar base templates anymore

PR #4828, opened 2023-03-23
Change default explosions

BlackYps:
The default explosions have this strange fireball texture that look very odd and low quality.

I changed the texture and emitter settings for these explosions. Now they are still far from perfect, but at least they somewhat blend in with the rest of the effects and they don't stick out like a sore thumb anymore.


BlackYps:
Comparison of the old and new explosions. I recommend trying it out in an actual game to get a better feel for it.

https://user-images.githubusercontent.com/52536103/227250108-97bc5d0e-7f34-401e-9cb5-e3c527f3c0da.mp4

https://user-images.githubusercontent.com/52536103/227250252-c820db09-30d3-4cf2-8e51-1c90b5ffdfae.mp4

relent0r:
As someone who preferred the original scfa explosions these look great! The previous ones always looked like they were added from an older game. 

I noticed the Cybran Exp gunship seems to be missing its explosions now tho.

Garanas:
I like it 👍 , I'll have to look at a few more units over the next couple of days

BlackYps:
What do you mean with the original scfa explosions? Are they different?
The Soul Ripper uses the same effect blueprint, so the explosions changed to fire effects there as well. We could change it back if we think that the explosions look better than the fire. The same thing happens on the hoplite, but I think the new effects look better there. For the soul ripper I am undecided.

relent0r:
Yea the original ones are closer to what you've done now (but not as good obviously). I guess my description of them is more of a hot white energy explosion rather than the fireball type effects they replaced them with in faf etc. But each to their own.

On mine machine the soul ripper has an initial small death explosion but no effects as it kind of crumples to the ground. Maybe my graphics settings are impacting it. The monkey lord by comparison has great ones.

BlackYps:
Oh wow, you are right, the original explosions are very different. I always thought that they were always like this. I can't believe that somebody greenlighted to change the explosions to the ugly fireball version we had for so long!

Ah, you meant the death explosion, not the weapons firing. It's still there, but less noticable. The soulripper just uses the default explosions and the come in only three sizes. The soulripper is at an awkward size for this as the medium explosion is kind of too small for it. It's more noticable now as I tuned down the size of the fire in the medium explosion. It should probably have its own custom explosion as it is an experimental, but that is out of scope of this PR.

Garanas:
Just looked at the explosions in Steam FA - they are a lot better then what we have in production at the moment. No idea when this changed, I recall them being like this since I started playing. Definitely some grounds to gain here

Garanas:
@BlackYps I think a lot of units can do with custom explosions. Out of scope of this pull request as you state, but in general there's more to gain there

Garanas:
@BlackYps how did you change these emitters?

Garanas (approved):

Garanas:
@BlackYps I've merged this in because I think it is great on its own. But I'm sure we can achieve a lot more

BlackYps:
I dug around in the git history and apparently the explosions were changed in 2014.

I used the emitter tool as described here: https://wiki.faforever.com/en/Modding/Modding_Emitters
It's actually surprisingly easy to work with. The only thing that is a bit tedious is when you want to scale the whole thing, because there is no global scale and you have to change like four different values for this instead. It could actually be worthwhile to try to find someone to play around with it. In contrast to the PBR texture stuff this basically requires no prior knowledge, just some time and maybe an artistic eye.

Yes, we can indeed do a lot when we utilize the emitters more. RK4000 showed with his explosions mod that a lot is possible.

PR #4830, opened 2023-03-24
Migrate default path generation to Nav Mesh

relent0r:
Description of changes
This PR migrates all the default AI pathing to the NavMesh equivalent. While the code side is relatively small it is a major change from the AI perspective and it removes the path marker restrictions from maps. Assuming the NavMesh can auto generate the layers then the AI will work on the map. 
This PR has place holders for the pathing. I will convert it to the threat based equivalent when it is ready so this should not be merged in as is due to it not taking imap threat into account.

While doing this conversion I noticed there are alot of issues in the default AI code in relation to handling unpathable locations where the error handling is just missing. I'll be looking to resolve this when refactoring the individual platoon/behavior functions. 
My conclusion is that the default skirmish AI only has half implemented pathing.

Test setup for the changes
There are not special setup requirements. Simply start a skirmish AI game using the default AI's, pick a map (any map) and off you go.
Ideally we are looking to see that the AI's platoons and behaviours that use pathing (not all of them do..actually a surprising number don't). Engineers usually all use the pathing functions. Transports do. AttackForceAI platoons do and others.
As we refactor the existing platoon and behaviour functions I'll be adding pathing support to those that don't currenlt have it.


Garanas:
Are these functions used by other AIs that may not have the navigational mesh, such as Uveso AI?

Garanas:
If no other AI in active development uses these then I think it is fine to refurbish them like this

relent0r:
Closing as its a duplicate. See 4879 using the most recent codebase.

PR #4832, opened 2023-03-25
AI Reclaim Platoon leveraging ReclaimGrid

relent0r:
### Description of changes

This PR adds a reclaim platoon for the default AI leveraging the reclaim grid. Currently a draft to look at what metrics might be missing from the reclaim grid and how to improve the basic concepts of AI reclaiming.
Currently it consist of a build condition and a platoon function.

Known issues.
- [ ] The AI will try to search outside the playable area, this can be resolved by having a min and max cell number against the grid parent class. (platoon is using placeholders that will not work on maps like fields of isis)
- [x] The ToWorldSpace function is only returning x and z. This requires more effort to get the 'actual' position. Ideally this function will return a vec3 position rather than what it does now.
- [x] The AI will assign multiple engineers to reclaim from the same cell when it is not required. Ideally there will be a method of recording which engineers are currently assigned to which cells. Because the reclaim grid is shared then it can't be done against it. Needs a per brain solution. I can make one but would like Jips opinion if he has a better option.
- [x] Engineers are not staying safe while moving to positions. Needs more logic for running from enemies, reclaiming enemy engineers etc
- [x] Engineers are not opportunistically building extractors as mass points that are close by.
- [x] Engineers will not use manually reclaim, only attack move.
- [ ] Loop is time limited to avoid strange issues rather than reclaim based. Pros n Cons.
- [ ] Engineers don't understand if reclaim is unreachable beyond if the cell is pathable to.
- [x] Engineers are not intelligent about where they are looking, they are just using a spiral from where they are now. I'm open to ways of improving that search methodology. It looks wrong but I'm not sure how to make it better.
- [x] Engineers are not grabbing reclaim on the way to the cell position, just moving.

### Test setup for the changes
Start an AI skirmish with the default AI. If the AI has reclaim around its base it will begin assigning engineers to reclaim. Use the reclaim grid debug tools dont by Jip to keep an eye on reclaim data.


Garanas:
@relent0r As discussed: new queries for the reclaim grid! They are all local to a position in grid space. You can use `ToCellIndices` to convert from world space to grid space. The radius is the same radius that you would give to iMAP-related functions. There is no predicate required, the functions work with the mass value of a cell. That is what matters most when we talk about reclaim.

- `MaximumInRadius`: given a radius and a position, retrieves the most valuable cell
- `FilterInRadius`: given a radius and a position, retrieves all cells that meet the threshold
- `FilterAndSortInRadius`: given a radius and a position, retrieves all cells that meet the threshold and sorts them from high value to low value

The debug thread is extended. It constantly searches for the maximum cell in a radius of 1. The cell that has the most value is drawn as blue. It also constantly searches for cells in a radius of 1 with a value of 500 or more. Those are drawn from white to grey (most to least valuable) with two lines, instead of one.

image

image

Garanas:
@relent0r I've updated your platoon function. I've added comments specifically for you. You can find them by searching for `@Relent0r` . If you have questions specific for me then I suggest to respond with a comment that says `@Jip` - I can respond to that accordingly again with `@Relent0r`

Garanas:
@relent0r I've updated the description of the pull request to use Markdown checkboxes. The syntax is: `- [ ]` for unchecked and `- [x]` for checked

relent0r:
Small note, there is still a requirement for mix/max cell properties on the grid to remain in playable area but the functionality is generally complete.

relent0r:
@Garanas are you able to give a review when you have time.

lua/editor/MiscBuildConditions.lua
```diff
@@ -357,5 +357,37 @@ function WaterMassMarkersPresent(aiBrain)
...
+    end
+
+    -- no need to reclaim when we have too much mass
+    if aiBrain:GetEconomyStoredRatio("MASS") > 0.7 then
+        return false
+    end
+
+    -- no need to reclaim when there's nothing around us to reclaim
+    local bx, bz = gridReclaim:ToGridSpace(manager.Location[1], manager.Location[3])
+    local maximumCell = gridReclaim:MaximumInRadius(bx, bz, 3)
```
relent0r, line 384: Set the radius dynamically else you may have engineers ranging too far on some maps in the early game and not far enough in the late game. e.g on a 20km map 3 rings might mean that you have a reclaim engineer venturing off a long way to get some reclaim when it would have been better utilized in the base in those first 5 minutes. Tho its very map dependant.

Garanas:
@relent0r I've merged these changes in, the current solution is not perfect but there's too many changes here that I'd like to use in other changes too.

One issue I also noticed is this situation:

image

Instead of reclaiming, the AI prefers to (constantly) re-build a tech 1 extractor. There's more value to gain here when it would just reclaim and delay the construction of the extractor

relent0r:
> 

For that particular engineer, its goal is to catch mass points that are out in the field rather than close to the base as otherwise you'll see engineers out there reclaiming and leaving masspoints uncapped while an engineer from the main base spends a few minutes trundling out to build it. We need to cut down on the masspoint candidates offered to the engineer (right now its all points on the map) before adding more logic to the decision making process. Hopefully we can leverage navmesh for this (its got other applications like raiding) to help with this as I'd like to implement pathing as well.

Yea there are certainly a few things that will pop out over testing. But as they say perfect is the enemy of finished :) Lucky we haven't tuned the AI yet as this will create a butterfly effect due to the economy swings it'll create.

PR #4839, opened 2023-03-27
AI Intel Framework Implementation

relent0r:
Description of changes
This PR adds an IntelFramework class to the AI. It is responsible for the AI understanding the Intel layer of the game. It is designed to assist the AI in making decisions based on data and metrics gathered at game start and during the game.

Currently it is contains the following information.
- An intel grid based on the Grid class implementation. This provides information to the AI on which the current intel status of each imap cell on the map providing information to the scouting platoons and so they can most efficiently cover the map.
- Team Count
- Ally Count
- Enemy Count
- Team mass point share
- Player mass point share
- Ally and Enemy start positions if available
- Self Start Position (vector3)
- PlayableArea table

To do
- [ ] Calculate self share mass points that are within easy access of the main base position. This will be used to calculate the factories that the AI can support within the first 5 minutes of the game and assist in initial engineer counts.
- [ ] Calculate easily accessible reclaim to assist in initial engineer counts. Will use ReclaimGrid to fetch this.
- [ ] Create query/update functions for scouts to callback on for updating the IntelGrid data
- [ ] Create build conditions that leverage the data for builders
- [ ] Provide data for shifting play styles.
- [ ] more to come

Test setup for the changes
This is currently a data only object so does not have test yet beyond confirming that the data is populated. Currently it will dump the data content of the IntelFramework on startup.


lua/platoon.lua
```diff
@@ -1165,6 +1165,67 @@ Platoon = Class(moho.platoon_methods) {
         end
     end,
 
+    ---## Function: LandScoutingAI
+    --- Handles sending land scouts to important locations.
+    ---@param self Platoon
+    ---@return nil
+    LandGridScoutingAI = function(self)
```
relent0r, line 1172: @Garanas can you sanity check this function for me.
Issue is that the CreateUnitDestroyedTrigger doesn't seem to match the scout deaths.
I'm getting a case where the grid assignments will get full across the grid, which indicates that the scouts are assigning themselves but when they die they are not always unassigning themselves.
I watched a game on a pure land map. Logged when a scout was assigned vs unassigned and after counting the scouts left on the map it didn't match up. But I can't see why.

In the game I watched.
I had 17 assignments and 11 unassignments with 5 grid cells currently assigned.

relent0r:
I'm closing this PR as it is very aged and alot of what it does has already been super seeded.  Will recreate when the time comes as the focus should be the new managers.

PR #4842, opened 2023-03-27
Veterancy value changes for T3 game enders

Basilisk3:
**T3 Artilleries:**
Mass required for one level of veterancy: 1.25*mass cost --> mass cost


MrRowey:
This Could Be Simplified to 
```lua
VeteranMasMult = 16500
```
its one line and reduces the number of tables needed

PR #4853, opened 2023-03-30
Annotate and refactor the builder manager of AIs

Garanas:
Improves the annotations of the builder manager and refactors the file to make it more intuitive to understand for new contributors.

- [x] Document `BuilderManager.lua`

The entire flow was undocumented. All functions have correct annotations and additional comments where necessary.

- [x] Pass location, location type and radius to base class

These properties were used by all instances.

- [x] Deprecate various unused functions

These merely clutter the code base

Improve performance of:
- [x] `SortBuilderList`: No longer creates a new table and uses the built-in sorting algorithm
- [x] `GetLocationCoords `: Returns the location of the manager instead of a newly-allocated table
- [x] `GetHighestBuilder`: Significant reduction in table operations and removes the `possibleBuilders` local in favor of re-using tables

### DIFF
--- a/lua/aibrains/base-ai.lua
+++ b/lua/aibrains/base-ai.lua
@@ -128,7 +128,6 @@ AIBrain = Class(StandardBrain) {
 
         self.UnitBuiltTriggerList = {}
         self.FactoryAssistList = {}
-        self.DelayEqualBuildPlattons = {}
     end,
 
     ---@param self AIBrain
--- a/lua/aibrains/base-ai.lua
+++ b/lua/aibrains/base-ai.lua
@@ -81,6 +81,7 @@ local StandardBrain = import("/lua/aibrain.lua").AIBrain
 ---@field TacticalBases? table
 ---@field targetoveride boolean
 ---@field Team number The team this brain's army belongs to. Note that games with unlocked teams behave like free-for-alls.
+---@field DelayEqualBuildPlattons table<string, number>     # Used to delay builders, the key is the builder name and the number is the game time in seconds that the builder remains delayed
 AIBrain = Class(StandardBrain) {
     ---@param self AIBrain
     ---@param planName string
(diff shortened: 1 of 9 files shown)


Garanas:
I tried to test these changes with other AIs (RNG / Uveso) but the vault versions have not yet adapted to the new brain structure. So they don't work at this point, can't use them to test the changes and see if they conflict.

relent0r:
> I tried to test these changes with other AIs (RNG / Uveso) but the vault versions have not yet adapted to the new brain structure. So they don't work at this point, can't use them to test the changes and see if they conflict.

You've still got a copy of that version of rng that we quickly converted don't you?

Garanas:
It got overwritten by the vault 😄 !

relent0r:
ah yea. I can test this weekend. btw its timestamp — 03/20/2023 9:24 PM in our chat history.

relent0r:
Gave it a quick test n got this.
warning: Error running lua script: g:\code\fa\lua\sim\buildermanager.lua(620): attempt to call method `IsPlattonBuildDelayed' (a nil value)
         stack traceback:
         	g:\code\fa\lua\sim\buildermanager.lua(620): in function `GetHighestBuilder'
         	g:\code\fa\lua\sim\engineermanager.lua(765): in function <g:\code\fa\lua\sim\engineermanager.lua:747>
         	(tail call): ?
         	g:\code\fa\lua\sim\engineermanager.lua(732): in function <g:\code\fa\lua\sim\engineermanager.lua:729>
         	
But this could well be my mod hooking the engineer manager rather than any issue with your commit. I'll investigate further later.

Garanas:
That is your AI hooking the manager. We should add the function back in but mark it as deprecated.

relent0r (approved):
Nice documentation and detail on how some of the threads operate.

PR #4857, opened 2023-04-02
Add new event when ACU enters transporter.

rigomate:
As seen in the discord suggestion channel, this could be a handy event for casters.
Whenever an ACU is put into a transport, there is likely some action going to take place.
So getting notified about this when casting could put the caster in the middle of the action.

The event is only created for replays, as it needs to be hidden from the enemy team.

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2230,6 +2230,23 @@ BaseTransport = ClassSimple {
             end
         end
 
+        --any ACU : Cybran, Aeon, Seraphim, Uef
+        if SessionIsReplay()  and (unit.UnitId == 'url0001' or unit.UnitId == 'ual0001' or unit.UnitId == 'xsl0001' or unit.UnitId == 'uel0001') then
+            -- prepare sync
+            local sync = Sync
+            local events = sync.Events or { }
+            sync.Events = events
+            local AcuEntersTransporter = events.ACUEntersTransporter or { }
+            events.ACUEntersTransporter = AcuEntersTransporter
+
+            -- sync the event
+            table.insert(AcuEntersTransporter, {
+                Timestamp = GetGameTimeSeconds(),
+                ACUArmyID = unit.Army
+            })
+        end
+
+
         unit:OnAttachedToTransport(self, attachBone)
     end,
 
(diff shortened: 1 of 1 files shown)


lua/defaultunits.lua
```diff
@@ -2230,6 +2230,23 @@ BaseTransport = ClassSimple {
             end
         end
 
+        --any ACU : Cybran, Aeon, Seraphim, Uef
+        if SessionIsReplay()  and (unit.UnitId == 'url0001' or unit.UnitId == 'ual0001' or unit.UnitId == 'xsl0001' or unit.UnitId == 'uel0001') then
```
4z0t, line 2234: Use category matching.

lua/defaultunits.lua
```diff
@@ -2230,6 +2230,23 @@ BaseTransport = ClassSimple {
...
+            -- prepare sync
+            local sync = Sync
+            local events = sync.Events or { }
+            sync.Events = events
+            local AcuEntersTransporter = events.ACUEntersTransporter or { }
```
4z0t, line 2239: no need for this shinanigans.

4z0t (changes requested):
You'd also have to put some changes in `UserSync.lua`

lua/defaultunits.lua
Garanas: What do you mean, no need? 😄 

This pattern prevents several table operations

4z0t: How often do you put your acu in a transport? 🙃

Garanas: Not that often, but it is good practice. Ideally at some point we'd create an interface that does it for you but today is not that day yet

lua/defaultunits.lua
rigomate: Thanks for the pointer, I was sure there would be something better to this.

rigomate:
> You'd also have to put some changes in `UserSync.lua`

Can you elaborate why that would be necessary?
I could just hook into this with a mod to get access to the information.
Like here:
https://github.com/rigomate/SupremeScoreBoard/blob/17055f312a8c652eda145946ddeca90482dd9f8e/hook/lua/UserSync.lua#L39

Thanks!

rigomate:
> 



> Observer is better than hooking, and we want to move towards using observers.

Ok, I will look into that, and modify the pull request.
Is there maybe a pointer for me for an existing pattern?

MrRowey:
Closing this PR for now as it has gone over a year with out any progress can be re-opened if needed

PR #4858, opened 2023-04-04
Refactor engineer manager

Garanas:
- [x] Remove economy triggers from engineer manager


Garanas:
The `ComputeVectors` function, now available in `NavUtils.lua`. The function can be use to compute rally points, defense points, attack vectors and more

https://user-images.githubusercontent.com/15778155/233721642-b6ad6540-3244-45e0-953c-528a6652b108.mp4

Garanas:
Using the new function (renamed to `GetDirections`) we can generate rally point markers, marked in red:

image

Garanas:
@relent0r introduced a basic silo behavior. You can enable the behavior through this hotkey:

image

All selected units are assigned to the platoon that manages (tactical) silo's. You can find the file here:

- lua/aibrains/platoons/platoon-silo.lua

This resembles a very basic platoon: the moment the silo is done building it triggers the platoon to start scanning for targets. It will scan for targets and pick one randomly. The scanning thread stops when it runs out of ammo, just to be triggered again when the silo is done building the next missile

The behavior is very basic, it could be improved in a thousand ways:

- Keep track of the same target until it is destroyed
- Keep track of TMD or shields by using the missile intercept events
- Keep track of other (allied) launchers and combine their volleys

But the idea was to show an initial basic platoon! I'll create another platoon that resembles the `HuntAI` behavior of the original platoon functions

Garanas:
Adjusted the silo behavior so that it supports multiple launchers:

https://user-images.githubusercontent.com/15778155/234113762-e3ce55bd-08e5-41cd-9273-f9a98bc92ef1.mp4

PR #4859, opened 2023-04-04
AI - Introduce scripted build out for ACU.

relent0r:
Description of changes

This PR adds a scripted build out capability for the ACU. Adds a fine control to the acu's initial build that will limit resource stalls in the first few minutes without affecting initial structure builds. It will also queue structure builds within the ACU's build radius during the build out so that the initial engineers will focus on expansion.
There is more potential for giving the AI personalities different build qualities here. Currently untapped.

The script is adaptable depending on the following.
Map size
Number of mass points within build range, outside build range but within the core base radius.
The personality type
If a hydrocarbon is present within 65 units

The function is used by the AI personalities that do NOT use the 'Default Initial ACU Builders' builder group.
This means that the Adaptive AI is that one that will use this. 
The following AIs definitely DON'T use it.
Tech
Turtle
Normal
Easy

The method it is implemented is via the intial ACU builder groups.
For example
the builder group Land Rush Initial ACU Builders has two builders, one is the standard initial builder and the second is used when prebuilt units are enabled.
The PlatoonTemplate for the initial ACU builders has been changed to use a the specific scripted build function.

The platoon function will effectively disable the ACU from taking on anything else until it is finished. It does this by setting the Combat and Initializing flag on the unit. This will stop the engineer manager from assigning it jobs and it will stop the commander thread from interrupting it during the build. If these are not in place then the platoon can easily be disbanded half way through.
I feel there may be some type of failback needed so that the ACU can never get stuck in an initializing state for too long. Perhaps a timeout thread that would remove the flags and disband the platoon if the build out took more than say 5 minutes.

This function also has a reliance on an early engineer building a hydro should one exist within 65 units of the spawn point. The acu can still complete the build out without it but it will still walk to the hydro marker and wait for an engineer for about 15 seconds. The hydro builders need to be updated to take this into account so they will build it as early as possible.

Test Setup
Start an AI game on a map of your choice selecting the adaptive ai (a couple of the others like rush will also run with this). The ACU should follow the build order function correctly with no errors.
Try multiple map sizes and types to see the adaptive methods it uses. e.g a map with no mass points within range vs one with many vs one with just a few.
Maps with starting hydros and maps without.
Maps with the spawn point in water (yes they exist). Note in this scenario everything else will probably break but not the BO.

ToDo, thoughts.
- [x] Path checking on mass points. Its possible on certain maps that distant markers won't be pathable.
- [x] hydro competition. On certain maps where two players start close to each other with only a single hydro. This is an untested scenario should may have issue.
- [x] What happens when enemy units come within range during the build out. This needs validation as the commander thread might force the ACU out of its build out without resetting flags. Might need changes to the commander thread for this scenario.
- [x] Shift reused code into sub functions to improve readability of the platoon function.


relent0r:
@Garanas can you take a look at this PR. I've tested it across a bunch of maps.
The only concern is if it fails during build out there is nothing to false the Initializing flag. I was thinking about forking a thread that would act as a timeout so should something go wrong the ACU will still go back into the engineer pool.

relent0r:
Two examples of build outs. One with not close mass points and a hydro and one with 4 mass points and no hydro.
https://youtu.be/EPAzOtYfDwE
https://youtu.be/CodAUT5qOQc

Garanas:
The pull request description mentions that there should be no warnings or errors, but I get these on Regor VI Highlands after the initial factory is finished:

```
WARNING: No buildLocation or whatToBuild during ACU initialization
WARNING: No buildLocation or whatToBuild during ACU initialization
WARNING: No buildLocation or whatToBuild during ACU initialization
WARNING: No buildLocation or whatToBuild during ACU initialization
```

Do these count of warnings or errors?

relent0r:
> The pull request description mentions that there should be no warnings or errors, but I get these on Regor VI Highlands after the initial factory is finished:
> 
> ```
> WARNING: No buildLocation or whatToBuild during ACU initialization
> WARNING: No buildLocation or whatToBuild during ACU initialization
> WARNING: No buildLocation or whatToBuild during ACU initialization
> WARNING: No buildLocation or whatToBuild during ACU initialization
> ```
> 
> Do these count of warnings or errors?

That warning is meant to indicate that the function to get a build location failed a reason. To avoid it breaking the build out on certain maps it would just dump a warning and move on to the next item. Improvements to the GetBuildLocation function will likely help decrease these. Though sometimes its related to wanting adjacency and not being able to find an adjacent spot within the acu's build range.

The warning is probably better served as a debug log rather than an in game log since its non fatal.

PR #4867, opened 2023-04-12
Recover missile progress when gifting a unit

Garanas:
Relates to: https://github.com/FAForever/FA-Binary-Patches/pull/15

Allows a (strategic) missile launcher to recover the progression of a missile when the unit is gifted.

### DIFF
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -110,6 +110,7 @@ function TransferUnitsOwnership(units, toArmy, captured)
 
         -- B E F O R E
         local orientation = unit:GetOrientation()
+        local workprogress = unit:GetWorkProgress()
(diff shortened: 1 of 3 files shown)


lua/SimUtils.lua
```diff
@@ -225,6 +226,10 @@ function TransferUnitsOwnership(units, toArmy, captured)
             newUnit:GiveTacticalSiloAmmo(numTacMsl - newUnit:GetTacticalSiloAmmoCount())
         end
 
+        if newUnit.Blueprint.CategoriesHash["SILO"] then
```
Strogoo, line 229: Does it work with tmls too? I tested it with nukes/smds while making a patch, but it seems like the "blocks" offset is the same for all silos, so it should be fine

Garanas: I have not tried that yet - I'll give it a spin

PR #4872, opened 2023-04-14
Feat: AI add initial transport file.

relent0r:
Description of changes
This PR reimplements the AI transport functionality with a set of utilities gifted by Sprouto from the LOUD mod. It is well tested and proven to function alot better than the default.

Implementation requirements.
- [x] Replace PlatoonGenerateSafePathToLOUD with NavMesh equivalent (still WIP). Discuss with Jip.
- [x] Change out get marker calls with cached marker equivalents
- [x] Validate Transport Pool functionality
- [x] Refactor default ReturnToBaseAI function to support transports correctly at a minimum
- [ ] Add visual debug tooling for transports
- [ ] Implement PlatoonCallForHelpAI for transports. Or decide on if we will support it. Its still required for ondamage drops. I haven't looked at the LOUD implementation yet.
- [x] Validate UEF gunship transport functionality (never was used in the default transport utilities)
- [ ] Discuss brain flags and decide if they are appropriately named etc
- [x] Create AIFindClosestBuilderManagerPosition for default AI
- [x] Create PlayableArea check function

Test Setup

nothing to test yet.


lua/sim/FactoryBuilderManager.lua
```diff
@@ -473,7 +473,9 @@ FactoryBuilderManager = Class(BuilderManager) {
                 factory.Trash:Destroy()
 				return self:FactoryDestroyed(factory)
 			end
-		end
+		elseif EntityCategoryContains(categories.TRANSPORTFOCUS - categories.uea0203, finishedUnit ) and self.Brain.NeedTransports then
+            self.Brain.NeedTransports = nil
```
relent0r, line 477: @Garanas Thoughts on this approach for clearing the brain flag for needing transports.

Garanas: It depends, is the need satisfied after creating a single unit? Or would the platoon that requires transports then reset the flag, which would create another transport?

relent0r: It may or may not be satisfied. But yes the platoons will query the transport pool and if there are still more transports needed they would reset the flag. Originally I thought counts would be better, but after thinking about it its really the same thing, you either need more transports or you don't.

The hard part so far is getting the platoons to make better decisions on when they need a transport vs when its not worth it due to the likelyhood of it getting shot down for the distance they want to travel.

relent0r:
Please note there are a couple of things that still need to be implemented for advanced functionality like the ability for transports to request air support and quickly drop units when being attacked, as well as handling fuel refits post drop off(less obvious on the smaller maps that faf uses).

PR #4876, opened 2023-04-18
Add path visualisation for navigational mesh

Garanas:
Adds ghosting of paths that are created. The feature is only enabled when the corresponding UI window is. The current implementation does **not** work for replays.

image


Garanas:
https://user-images.githubusercontent.com/15778155/232977045-1458c8db-229b-43bd-b7cb-319b2b852b02.mp4

PR #4879, opened 2023-04-18
Migrating pathing for default AI to navmesh

relent0r:
Description of changes
This PR converts all existing pathing logic in the default AI. This will impact 3rd party AI that leverage default AI functions. The most likely place that it would impact them is engineer pathing and transport pathing as people often don't create their own functions for this. I'll notify the AI devs about it. The biggest issue they will have is that they may not generate the navmesh on startup.

The main goal of this PR is to provide better map compatibility compared to what we have now where the AI will not function on most 3rd party maps since they don't have pathing markers.

Future updates will include logic to handle pathing failures due to threat etc. But that is more likely to be implemented in the state machine logic rather than the platoon logic.

Testing,
Run some AI games across various map types with and without default AI markers and validate that the platoons / units are getting paths correctly.
Currently the NavMesh does not generate naval areas so on maps with no markers the AI won't build naval expansions.


Garanas:
Are these functions used by other AIs?

relent0r:
> Are these functions used by other AIs?

I had a look across the various AI's and came to the conclusion that they don't. At least not directly.
SCTA, Swarm, RNG, Sorian use their own versions. Uveso hooks the default and replaces it for his AI. M27/M28/DilliDalli use their own stuff.
Given that we will likely replace alot of the functions with state machine logic this isn't a MUST merge PR. But is a nice to have for users that don't understand that AI's don't support most 3rd party maps. But if we have concerns of the impact to other AI (or maybe custom campaigns that might) then thats ok.

The transport platoons are not used by anyone else that I could see other AI's use the default transport functions which is why I've left them alone and used the ones in the new transport lua.

Though I did notice one problem in relation to the factory manager change I made that needs the aiBrain.NeedTransports to nil. This causes problems with other AI because they don't expect it to get reset to nil during the game.

PR #4881, opened 2023-04-20
Cast shadows on all props

BlackYps:
Some rocks were using a shader that didn't calculate shadows, making them look odd when they received shadows from units. So I fixed it.

Screenshot 2023-04-20 110005
Screenshot 2023-04-20 105856


Garanas:
That is way better 👍

Garanas (approved):

PR #4882, opened 2023-04-20
Initial draft implementation for AI map presence awareness

Garanas:


relent0r:
Have been testing this with a platoon to assist decision making and so far its been looking great. I'm still yet to do ffa based testing to see how the grid looks. Was going to test Vya-3 Protectorate in a 4 way ffa to test.

Garanas:
I think you're responding to the wrong pull request. The one that is merged in is from #4930

PR #4885, opened 2023-04-24
build sort priority of crab eggs

MadMaxFAF:
gives crab eggs the same buildsortpriority as the unit it is for thanks to evildrew for pointing this out


Garanas:
Great find! This pull request is about the order in the construction menu:

image

image

PR #4900, opened 2023-05-05
Texture quality experiments

BlackYps:
Experimental changes to increase texture quality. In the water shader we can play around with mipmaps to bring more consistent results. It's especially noticeable with the refractions on close to medium zoom levels or when you tilt the camera a lot and observe the sea surface.
For the terrain we can change the texture scale to improve the texture resolution on close zoom levels.

### DIFF
--- a/effects/water2.fx
+++ b/effects/water2.fx
@@ -333,17 +333,28 @@ float4 HighFidelityPS( VS_OUTPUT inV,
     // for the land cutout, we multiply by a large number and then saturate
     float mask = saturate(backGroundPixels.a * 255);
 
-    // calculate the normal we will be using for the water surface
-    float4 W0 = tex2D( NormalSampler0, inV.mLayer0 );
+    // Calculate the normal we will be using for the water surface.
+	// We bias the mipmaps for higher texture quality
+    float4 W0bias = tex2Dbias( NormalSampler0, float4(inV.mLayer0, 0, -2));
+	float4 W1bias = tex2Dbias( NormalSampler1, float4(inV.mLayer1, 0, -2));
+	float4 W2bias = tex2Dbias( NormalSampler2, float4(inV.mLayer2, 0, -2));
+	float4 W3bias = tex2Dbias( NormalSampler3, float4(inV.mLayer3, 0, -2));
+
+	// Calculate the normal we will be using for the foam.
+	// We don't change the mipmaps here because that would create 
+	// too drastic visual changes from the original shader
+	float4 W0 = tex2D( NormalSampler0, inV.mLayer0 );
 	float4 W1 = tex2D( NormalSampler1, inV.mLayer1 );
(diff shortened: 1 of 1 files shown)


Garanas:
Without:

image

With:

image

Without:

https://user-images.githubusercontent.com/15778155/236520666-358a73d9-90f9-4c0c-9013-2691f6194e15.mp4

https://user-images.githubusercontent.com/15778155/236520672-6f87b7e9-9f3b-46d0-b6be-e980c43d21d9.mp4

With:

https://user-images.githubusercontent.com/15778155/236520774-7f92977c-5cfa-4e94-bc78-eee4d511820a.mp4

https://user-images.githubusercontent.com/15778155/236520786-2d584ecc-06b9-49a7-a598-42a1afa6dbc6.mp4

Garanas:
It is a bit hard to see properly, but the seton's clutch screenshot shows some of the difference for the water shader. @BlackYps what map(s) did you used to test this?

BlackYps:
I tested it on Seton's clutch mainly. Also on dual gap reloaded, but I am not sure if I tested the final result there as well.
My test setup was mainly to spawn the tempest, dive it and then looking at the refractions when slowly zooming out.
Originally the size or intensity of the distortions seemed to change in a weird way on close zoom levels.
You could also notice a quality degration when tilting the camera and observing the sea close to the horizon. The small waves got smoothed out too quickly

BlackYps:
showcasing the decal change
Screenshot 2023-05-05 123748
Screenshot 2023-05-05 123846

Garanas:
With:

image
image
image
image

Without:

image
image
image
image

Maps:

- Archsimkats Valley (x2)
- Zerg Canyon
- Williamson's Bridge

The decal changes drastically reduce the visibility of normal decals on all maps. We can't do that.

Garanas:
It does look better for map-wide normals 😄 !

BlackYps:
I was gonna say that we can fix the visibility by playing around with the normal intensity increase of the decals, but I noticed a problem: It also affects the tarmacs and especially the UEF one has to completely replace the normals of the ground, otherwise it looks bad. I guess the way to go would be to do the adjustments directly in the files that we want to affect.

Garanas:
Let's do that in a separate pull request then. That way we can combine all the changes and review them as a whole.

BlackYps:
I'll also separated the water and terrain changes, this makes it less convoluted.
Showcase of the water changes, before and after:
Screenshot 2023-05-06 151928
Screenshot 2023-05-06 151719
Notice how the water looks more detailed in the distance. The refractions on the Tempest look more consistent now as well, before the refractions seem to be less strong on the parts of the ship that are further away from the camera.

Screenshot 2023-05-06 151306
Screenshot 2023-05-06 151208
This one shows how the refraction scale decreases in shallower water. The depth is always measured to the sea floor, not to the unit, so it's not perfect but still an improvement.
All pictures are from flooded tabula rasa

PR #3585, opened 2021-12-04
Terrain shader experiments

Garanas:
Various experiments to the terrain shader for improved fidelity. Currently it tries to implement:
 - https://www.gamedeveloper.com/programming/advanced-terrain-texture-splatting
 - https://blog.selfshadow.com/publications/blending-in-detail/
 - https://www.ronja-tutorials.com/post/010-triplanar-mapping/
 
 Sources:
  - https://github.com/bgolus/Normal-Mapping-for-a-Triplanar-Shader
  - https://catlikecoding.com/unity/tutorials/advanced-rendering/triplanar-mapping/
  - https://www.youtube.com/watch?v=ueC2qGzWrgQ
  - https://www.youtube.com/watch?v=V3wghbZ-Vh4
  - https://learnopengl.com/PBR/Theory
  
  Instruction counts:
   - https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx9-graphics-reference-asm-ps-instructions-ps-2-x
   - https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/dx9-graphics-reference-asm-vs-instructions-vs-2-x

PR #4902, opened 2023-05-06
Add new terrain shaders

BlackYps:
Continued from #4900
~~I used the rock scaling idea to increase texture quality when being zoomed in. This way the texel density of the ground now somewhat matches the texel density on units and props, making it look much better.
I also fixed some important typos in the shader. Due to the different normal interpolation and mask ranges this isn't a drop in replacement for existing maps, but it looks promising for future maps.~~

~~I noticed that in #3585 there were some commits that added different albedo blending and a shadow map, these could be further improvements.~~

This adds two new terrain shaders. One that is basically behaving like the TTerrainXP one, but we use the channels of the upper albedo to provide a shadow map, our own terrain base normals, and a channel to tune the albedo like the "overlay" blend mode. This means 0.5 keeps the albedo as is, greater values lighten it and smaller values darken it.

The other one uses the same improvements, but on top of that uses advanced texture splatting based on heightmap textures and a PBR implementation based on roughness maps for the reflections.

Both also come with improved light absorption and scattering calculations for the water. To get good results these maps should use the PBR shader for units, because the mesh shader needs to do equal computations on the units for the effect to look believable.


BlackYps:
Some comparison pics. I had to cheat a bit and adjust the mask scaling for the screenshots so the texture layers are the same. But this is not relevant as we can adjust the map generator to the new shader.
Screenshot 2023-05-06 234318
Screenshot 2023-05-06 234338
Screenshot 2023-05-07 000034
Screenshot 2023-05-07 000054

BlackYps:
Showcasing the reflections and different texture splatting of the PBR shader
Screenshot 2023-05-12 165609
Screenshot 2023-05-12 163715

BlackYps:
A comparison of the water improvements
Screenshot 2023-05-17 004351
Screenshot 2023-05-17 003754
By manipulating the color ramp of the water it is even possible to fake a glow effect
Screenshot 2023-05-17 124717

PR #4905, opened 2023-05-07
Add initial filtering implementations

### LINKED ISSUE #4891: Cybran Land Scout Deselection
**Describe the feature**
The Seraphim Selen (T1 Land Scout) has a cloaking ability. 
And when this ability is active, the unit is unable to be selected (with a large group of other units,) to prevent it from moving.
However, the Cybran Mole, which also has the same ability, does not have the same feature.

**Expected behavior**
Cybran Mole to have a Deselection feature, same as the Seraphim Combat Scout has.

**Screenshots**
Selecting all units on screen:
Screenshot 2023-04-29 085027

Result: Selen and Mole are both Cloaked, yet Mole isn't deselected.
Screenshot 2023-04-29 085038

**Additional context**
Holding Shift (or not) when selecting a Selen, regardless, will always be deselected if another unit is selected with it.

### LINKED ISSUE #4892: Scathis Deselection
**Describe the feature**
The Seraphim T1 land scout has a cloaking ability, and when this ability is active, the unit is unable to be selected with other land units.
This is to prevent accidental move orders. 

How many times have you given an accidental order to a Scathis? I have - Many times.
The Scathis should have a similar feature to prevent it from moving.

---

**Expected behavior**
When selecting all ground or combat units, the Scathis should be deselected - similar to the Seraphim Selen.

---

**Screenshots**
Selecting all units on Screen:
Screenshot 2023-04-29 091701

Result: Selen is not selected, but Scathis is.
Screenshot 2023-04-29 091709

---

**Additional context**
Similar to this post.
https://github.com/FAForever/fa/issues/4891

Garanas:
Closes #4892 and #4891


lua/ui/game/gamemain.lua
```diff
@@ -1090,3 +919,436 @@ SendChat = function()
...
+function DeselectLockedOut(selection, count)
+    -- try and filter locked out units
+    local head = 1
+    for k = 1, count do
+        local userUnit = selection[k]
+        if not userUnit.LockedOutOfSelection then
+            selection[head] = userUnit
+            head = head + 1
+        end
+    end
```
lL1l1, line 1082: I guess all of this is to preserve the table as an array? Simply doing `if locked then t[i] = nil end return t, table.getn(t)` leaves holes and requires a table.getn call.

lL1l1 (commented):
Since it's not written here, according to https://github.com/FAForever/FA-Binary-Patches/issues/30, this is on hold (is in draft form) because "It would likely break any UI mod that relies on how the selection works."

The PR format could be improved by splitting it up between implementing various filters and the actual new filtering. Though my biggest pain was that moving the entire `OnSelectionChanged` function around caused a giant diff with for just a few changes.

PR #4911, opened 2023-05-11
Janus Speed Fix

MrRowey:
From what i can see upping the multiplier from 0.67 back to 1.0 in the on target if statement prevents the slow speed when bombs being dropped or when the booming has been interrupted with ground fire

### DIFF
--- a/units/DEA0202/DEA0202_Script.lua
+++ b/units/DEA0202/DEA0202_Script.lua
@@ -1,12 +1,8 @@
---****************************************************************************
---**
---**  File     :  /cdimage/units/DEA0202/DEA0202_script.lua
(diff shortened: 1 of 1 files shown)


Garanas:
I think you can also just remove the line in question.

Garanas:
Let's just remove all of the code when targets are gained / lost. We now still edit properties of the Janus, we just removed the most notable property (the speed)

MrRowey:
I removed the `OnGotTarget` & OnLostTarget` and this kind of breck the Janus a tad as the the vial cue that mobs have drops breaks and the Janus itself dose really sharp turns once the bombs have been dropped trying to drop the bombs

Garanas:
Closing this until we have a better solution that takes into account fuel 😄

PR #4915, opened 2023-05-13
Add experimental approach to reduce decal flickering

Garanas:
A common nuisance among map authors is the flickering of decals. It turns out that there are console commands to influence the clipping behavior. Requires the extended graphics to be enabled.

Current behavior:

https://github.com/FAForever/fa/assets/15778155/1dc54d44-7a8b-40ee-acf3-9c6fbdcea74f

New behavior:

https://github.com/FAForever/fa/assets/15778155/0620efda-2f52-48b1-8cc7-6f4bf184be65

This doesn't work for all maps. As an example: Corona Hourglass is still affected by this behavior

### DIFF
--- a/lua/ui/game/gamemain.lua
+++ b/lua/ui/game/gamemain.lua
@@ -200,14 +200,15 @@ function CreateUI(isReplay)
             WaitSeconds(1.0)
 
             if Prefs.GetFromCurrentProfile('options.level_of_detail') == 2 then
-                -- allow meshes and effects to be seen from further away
-                ConExecute("cam_SetLOD WorldCamera 0.65")
+                ConExecute("cam_SetLOD WorldCamera 0.70")
             end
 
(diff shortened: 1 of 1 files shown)


Garanas:
When testing on a map with 13K decals that always render (called `Mavys Marsh`) this pull request does reduce the framerate below 60. Given, it is an extreme example.

With these changes it renders all 13K decals when completely zoomed out, and it reduces the fps to 30. It is usually around 70 on my machine, where the setting culls about 2000 decals - rendering only 11K decals when completely zoomed out.

image

Garanas:
I've done some testing while playing the game. A lot generated maps have these flickering issues. With these console commands they appear to be fixed. As an example:

Without:

https://github.com/FAForever/fa/assets/15778155/146fbcb7-92c1-4846-b2db-338fab2cc1d5

With:

https://github.com/FAForever/fa/assets/15778155/a9acb86f-e83b-48ab-a9e7-baf8211a85ac

PR #4919, opened 2023-05-14
Take into account playable area for skirmish maps

Garanas:
Introduces basic support for the playable area of a map. It does not re-generate the navigational mesh when the playable area is changed. The playable area is ignored for a non-skirmish (read: campaign) map. It takes a few assumptions:

- The code assumes the playable area is set immediately in the map script, with no delays. This is usually the case
- The code assumes that the navigational mesh is generated _after_ the map scripts have run. This is the case when you generate the navigational mesh in the recent introduced `OnBeginSession` callback of the brain. See also https://github.com/FAForever/fa/blob/nav-mesh/playable-area/lua/aibrains/easy-ai.lua#L52-L70

As an example of the result:

image

### DIFF
--- a/lua/sim/NavGenerator.lua
+++ b/lua/sim/NavGenerator.lua
@@ -1150,17 +1150,6 @@ local function GenerateMarkerMetadata()
     end
 end
 
----@param mapSize number
----@param compressionTreeSize number
----@param compressionThreshold number
-local function GenerateWithWater(mapSize, compressionTreeSize, compressionThreshold)
-
-end
-
-local function GenerateWithNoWater()
-
-end
-
 --- Generates a navigational mesh based on the heightmap
 function Generate()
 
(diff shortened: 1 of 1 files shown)


relent0r:
@Garanas When it comes to developers using the navmesh, is there a supporting query they can use on a position?

At the moment there is no standard for determining if a position is in the playable area or not. Just alot of adhoc methods, would be good to just have a utility we can call to simplify all that code.

Garanas:
Yes, what would you like that query to do?

relent0r:
> Yes, what would you like that query to do?

Basically this. But in reverse, return true. So you could say 'if navutils.InPlayableArea(position) then go do stuff end`
            if position[1] - playableArea[1] <= 8 or position[1] >= playableArea[3] - 8 or position[2] - playableArea[2] <= 8 or position[2] >= playableArea[4] - 8 then
                return false
            end

Garanas:
I'll add it in tomorrow!

PR #4921, opened 2023-05-17
Introduce various transport and carrier callbacks

Garanas:
Introduces various callbacks and platoon events that were missing, including:

- `OnTransportAttach`
- `OnTransportDetach`
- `OnTransportAborted` 
- `OnTransportOrdered`
- `OnAttachedKilled`
- `OnStartTransportLoading`
- `OnStopTransportLoading` 

Additional, adds the following missing events to the new platoon template:

- `OnStorageChange`
- `OnAddToStorage`
- `OnRemoveFromStorage`

They are all annotated. The comments describe when you can expect the event to trigger.


Garanas:
@relent0r would it be interesting to know what units are attached to a transport?

PR #4923, opened 2023-05-17
Factor out the campaign elements of the base AI

Garanas:
Separates the campaign AI from the skirmish AI. All campaign AI related logic can now be found in `/lua/aibrains/campaign-ai.lua` . As a result the remaining logic that remains in `/lua/aibrains/base-ai.lua` is entirely related to the skirmish AI.


Garanas:
Tested on various campaign missions including the unpublished Tight Spot and the Seraphim missions. It appears all is good.

PR #4930, opened 2023-05-18
Introduce grid that allows an AI to understand map presence

Garanas:
Introduces a grid-based data structure that allows an AI to understand his presence on a map. The grid takes into account whether it can path to a given location. The grid uses the `Hover` layer to determine pathability. A small peak:

image

This is from the perspective of the blue player. Green means it is considered safe. Red means it is considered hostile. Yellow means it is considered contested - it doesn't necessarily belong to either player.

The initial phase of the data structure is based on what mass points have structures on top or around it. This takes into account intel - if the AI is unaware of structures then it won't take them into account. The second phase is a randomized breath first search. The search is randomized to prevent an initial order (top left to bottom right) to influence the results. As a consequence the results are always slightly different, but that is fine. The last phase determines which cells are contested. It does so by scanning the neighbors - if one neighbor is considered hostile then the cell is considered contested.

- [x] Add debugging UI

A little animation that shows the process:

https://github.com/FAForever/fa/assets/15778155/f0affea0-ebce-4b1a-a433-184e6f9f5431

https://github.com/FAForever/fa/assets/15778155/0b218c58-93db-444b-a05d-9b3861309d02

### DIFF
--- a/engine/Sim/CAiBrain.lua
+++ b/engine/Sim/CAiBrain.lua
@@ -366,6 +366,7 @@ end
 ---@param position Vector
 ---@param radius number
 ---@param alliance AllianceType
+---@return Unit[]
 function CAiBrain:GetUnitsAroundPoint(category, position, radius, alliance)
 end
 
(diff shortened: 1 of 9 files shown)


Garanas:
Introduced a single query: `GetInferredStatus` which retrieves the inferred status of a given position.

PR #4931, opened 2023-05-18
Improve sun reflections

BlackYps:
Currently the sun reflection can be seen on the entire water surface. This is implausible, because units on the water should block out the sun reflection where the unit reflection is visible. This PR fixes that.

Screenshot 2023-05-18 223454
Screenshot 2023-05-18 223544


Garanas:
It is subtle 😄 ! Looking good, great work

PR #4932, opened 2023-05-19
Add flag to remap shader names during blueprint loading

Garanas:
Allows us to dynamically remap the shader names. The code expects the flag `applyAdvancedShaders` to be defined and set to `true` in the scenario file of a map. For debugging purposes it now works for all maps.

This approach takes the same route that we introduced for custom strategic icons. It also has the same flaw: it loads the preference file from disk, but the version on disk may not match what the player expects. This is because we preload the game during the lobby. As we make adjustments (change the map, change the ui mods) then the preference file on disk is not updated. It is updated when we launch the game. But at that point the game is already preloaded.

The only solution to this is to no longer preload the game while waiting in the lobby, like we used to do before last year when we introduced it. See also: https://github.com/FAForever/fa/pull/4406. The consequence is that we have loading times again.

### DIFF
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -95,6 +95,17 @@ local function LoadPreGameData()
         end
     )
 
+    reprsl(preGameData)
+    local scenario = DiskFindFiles(preGameData.CurrentMapDir, "*_scenario.lua")
+    reprsl(scenario)
+    local info = {
+        STRING = function(value)
+            return value
+        end,
+     }
+    doscript(scenario[1], info)
+    reprsl(info)
+
     -- tell us if something went wrong
     if not ok then
         WARN("Blueprints.lua - Preferences file is locked or corrupt. Skipping pre game data.")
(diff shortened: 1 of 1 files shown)


Garanas:
As an example output:

```
INFO: ura0107: Insect -> PBR_Insect
INFO: xec1908: Unit -> PBR_Unit
INFO: xec1908: Unit -> PBR_Unit
INFO: urc1901: Insect -> PBR_Insect
INFO: urc1901: Insect -> PBR_Insect
INFO: ual0202: Aeon -> PBR_Aeon
INFO: ual0202: Aeon -> PBR_Aeon
INFO: xeb0204: Unit -> PBR_Unit
INFO: xeb0204: Unit -> PBR_Unit
INFO: uab1301: Aeon -> PBR_Aeon
INFO: uab1301: Aeon -> PBR_Aeon
INFO: urs0201: Insect -> PBR_Insect_Navy
INFO: urs0201: Insect -> PBR_Insect_Navy
INFO: ueb4201: Unit -> PBR_Unit
INFO: ueb4201: Unit -> PBR_Unit
INFO: xal0305: Aeon -> PBR_Aeon
INFO: xal0305: Aeon -> PBR_Aeon
INFO: xac1401: Aeon -> PBR_Aeon
INFO: xac1401: Aeon -> PBR_Aeon
INFO: xab1401: Aeon -> PBR_Aeon
INFO: xab1401: Aeon -> PBR_Aeon
```

After discussing it with @BlackYps , it may be better to create an extensive lookup table instead of trying to generalize the problem like we are now.

BlackYps:
> This approach takes the same route that we introduced for custom strategic icons. It also has the same flaw: it loads the preference file from disk, but the version on disk may not match what the player expects. This is because we preload the game during the lobby. As we make adjustments (change the map, change the ui mods) then the preference file on disk is not updated. It is updated when we launch the game. But at that point the game is already preloaded.

Interesting, so this is the reason why changing an icon mod is only applied after the next game start?

Garanas:
Yes, that is why.

BlackYps:
I think in this case it is not relevant that the preferences file might be outdated. The shader loading is controlled by the scenario file of the map and that is static.

Garanas:
It does matter, because the blueprint process is unaware of the map by its own. We pass it to the loading process through the preference file.

BlackYps:
So if I change the map in the lobby, it will use the shader settings of the map that I started the lobby with?

Garanas:
Yes, that would be the case.

Hdt80bro:
Can't prefetch session can be restarted when a setting is changed?

Hdt80bro:
Is that something that `SavePreferences()` would fix?

BlackYps:
I'll provide a list how the shaders should be renamed to make the creation of the lookup table easier

PR #4934, opened 2023-05-20
Add event message for moderation

Garanas:
Adds a new callback to signify events. These events can then be read by the parser of the client so that they end up in the chat history of a replay.

See also: https://github.com/FAForever/faf-java-commons/pull/102


Garanas:
Closed in favor of #6002

PR #4935, opened 2023-05-20
Buff to the torpedo DPS of amphibious tanks

Basilisk3:
Right now, ACUs can stand inside an army of submerged amphibious tanks and just shrug off with regen for the most part. These changes buff their torpedoes, so they are at least somewhat useful. Note that even with this considerable buff, they still have low DPS as to not interfere with naval/submarine balance too much. 

**url0203 Amphibious Tank**
Damage: 15 --> 20 
DPS: 7.5 --> 10

**xsl0303 Siege Tank**
Damage: 30 --> 40
DPS:  15 --> 20

The Brick does not get a buff, because it has torpedo defense and a lot of health.

### DIFF
--- a/units/URL0203/URL0203_unit.bp
+++ b/units/URL0203/URL0203_unit.bp
@@ -353,7 +353,7 @@ UnitBlueprint {
             },
             BallisticArc = 'RULEUBA_None',
             CollideFriendly = false,
-            Damage = 15,
+            Damage = 20,
             DamageType = 'Normal',
             DisplayName = 'Meson Torpedo',
             FireTargetLayerCapsTable = {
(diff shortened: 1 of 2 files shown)


Spikey84:
We like this change. But we want to go a step further. Would you like to add the changes in #5012? If not one of us can make a new PR. Also whats your FAF name and Discord name?

Basilisk3:
I will test the proposed changes, they make sense to me but may be a bit on the strong side in my opinion.

I don't use discord and mostly communicate through GitHub. If I have time, I will pick up some of the issues.

MrRowey:
> I don't use discord and mostly communicate through GitHub. If I have time, I will pick up some of the issues.

What's your FAF username so we're able to add that when we do the main patchnotes release.

Basilisk3:
> > I don't use discord and mostly communicate through GitHub. If I have time, I will pick up some of the issues.
> 
> What's your FAF username so we're able to add that when we do the main patchnotes release.

You can use Basilisk3

Basilisk3:
> > What's your FAF username so we're able to add that when we do the main patchnotes release.
> 
> You can use Basilisk3

PR #4937, opened 2023-05-20
Tech 1 Sonar buff

Basilisk3:
At the same sensor range as tech 1 radars, tech 1 sonars were not very useful even with their cheaper price. This is due to naval units being much faster than land units. This change tries to alleviate this issue.

**Intel:**
SonarRadius of all tech 1 Sonar Systems: 115 --> 130


MrRowey:
We discussed the idea of Sonar change and came up with the ones mentioned in https://github.com/FAForever/fa/issues/4986. So we are going to close this one.

PR #4960, opened 2023-05-22
Update mod information for FAF

Hdt80bro:
Updates the documentation found in `MODS.LUA`, including old unimplemented features / bad documentation, new features FAF added, and touches on some of the interoperability with the FAF client.

### DIFF
--- a/lua/MODS.LUA
+++ b/lua/MODS.LUA
@@ -1,254 +1,364 @@
---[[ ------------------------------------------------------------------------------------------------------------------
-Mods
+--[[
+----------------------------------------------------------------------------------------------------
+FAF Mods
+----------------------------------------------------------------------------------------------------
 
-All mods must be in subdirectories somewhere under /mods. Each mod has a
-mod_info.lua file which contains various bits of info:
+FAF mods are special kinds of Supreme Commander mods in the FAF vault designed to work within the
+FAF framework. They differ from regular Supreme Commander mods, i.e. those that directly replace
+and hook original .scd files on startup. In a sense, FAF itself is a Supreme Commander mod that
+coordinates the participation of other mods intended to be used inside of it once the game launches.
+In this light, there are certain technical limitations and restrictions that are necessary to be
+able to flexibly support the modding scene, but beyond these basic managerial requirements, they are
+pretty much free to do whatever they like with the game and the FAF framework.
 
-    name = "Happy mod"                          -- Name to use for this mod
-    version = 123                               -- Version number to use for this mod
-    copyright = "Copyright © 2006, Someone"     -- Optional copyright info
+A couple (but not all) of the restrictions include:
+ * If you upload the mod to the vault, you MUST follow the vault rules:
+   https://wiki.faforever.com/en/Vault-Rules
+    * In particular, make sure you own all the content of the mod you are uploading. Require other
+      mods when available to avoid copying others' work.
+ * Mods are not able to modify the lobby or provide game options (since mods are only fully loaded
+   after the game launches).
+ * All of a mod's files are mounted behind the mod directory to avoid conflicts. Outside of lua
+   scripting (which can always hook original and FAF files), this means that files are never
+   replaced (in practice, since the location a resource such as a texture points to can be changed
+   by a lua file, this only really affects shaders, which are accessed by the game engine directly)
+
+
+----------
+Mod info file
+----------
+
+Each mod has a `mod_info.lua` file which contains various bits of info:
+
+    -- Required name to use for this mod. This is what is displayed in the Lobby, after undergoing
+	-- title-casing. Do NOT add version information here - keep this the exactly same when you
+    -- update what you consider to be the same mod as the FAF client will uses this to auto-update
+    -- mods (based on the most recent `uid` - the version number does not matter and is only used
+    -- for display purposes).
+    name = "Happy mod"
+  
+    -- Optional integer version to use for this mod (please only increment) to indicate to humans
+    -- the most current version. Decimals are not supported. If updating the mod, make sure to
+    -- change the `uid` as that is what the FAF client actually uses and leave the name exactly the
+    -- same.
+    version = 4
+
+    -- Optional copyright info. If absent, the mod author is presumed to maintain full ownership of
+    -- the mod. Invisible in-game but shown in the FAF client.
+    copyright = "Copyright © 2023, Someone"
+
+    -- Optional text displayed in the mods manager and FAF client
     description = "A long description of happy mod and why it will make you happy"
-    author = "Joe"                              -- Optional author info
-    url = "http://www.gaspoweredgames.com"      -- Optional URL to anywhere author likes
 
+    -- Required name(s) of author(s). Shows up as `"UNKNOWN"` if left blank. Will be begin to be
+    -- truncated in-game if over 20 characters. Make sure this reflects who is uploading the mod to
+    -- the vault (if applicable) as well as all contributors whose work is present.
+    author = "Joe"
+    
+    -- Optional URL to anywhere author likes. Will show up in the mods manager if valid.
+    url = "https://www.faforever.com"
+    
+    -- Optional URL to anywhere author likes. Will show up in the mods manager if valid.
+    github = "https://github.com/FAForever"
+
+    -- Uniquely identify this mod. This defaults to the mod's name, but that's not a very safe way
+	-- to keep mods differentiated. I'd recommend using a GUID or Guaranteed Unique ID.
+    -- Every new version of a mod should get a new GUID so that mods which rely on the old version
+    -- can select it appropriately, and so that the FAF client can keep track.
+    -- The following webpages allow you to generate a GUID online (as of  April 2023)
+    --   http://www.somacon.com/p113.php
+    --   http://www.famkruithof.net/uuid/uuidgen
     uid = B019FBEE-E411-11DB-AFAE-13D355D89593
-        Uniquely identify this mod. This defaults to the mods name, but that's
-        not a very safe way to keep mods differentiated. I'd recommend using a GUID
-        or Guaranteed Unique ID. Also, every new version of a mod should get a new
-        GUID so that mods which rely on the old version can select it appropriately.
-        This following webpages allow you to generate a guaranteed unique ID on line (as of 4/6/07)
-        http://www.somacon.com/p113.php
-        http://www.famkruithof.net/uuid/uuidgen
 
+    -- Flag for whether to list this mod in the mod manager window, where the player can select it.
+    -- An example where you would set it to false is library mod supplying functions, textures, and
+    -- props for use by other mods. Those mods indicate that they require this one, and it is
+    -- automatically enabled when they are.
     selectable = true
-        Flag for whether to list this mod in the mod manager window, where the
-        player can select it. Examples where you would set it to false are:
 
-        1.  A mod containing just custom maps, along with textures and props used
-            on those maps. The mod will be automatically enabled when one of its
-            maps is used.
-
-        2.  A mod supplying textures and props for use by other mods. Those mods
-            indicate that they require this one, and it is automatically enabled
-            when they are.
-
+    -- Setting enabled to false causes a mod to never be loaded. Provides an easy way to disable a
+    -- mod during development. The default is true.
     enabled = true
-        Setting enabled to false causes a mod to never be loaded. Provides
-        an easy way to disable a mod during development. The default is true.
 
+    -- A simple form of preventing conflicting mods. If a mod has `exclusive`, then no other mod can
+    -- be active. The default is false. Note that this precludes the use of `conflicts`.
     exclusive = false
-        A simple form of preventing conflicting mods. Only one 'exclusive' mod
-        can be active at a time. The default is false. Note that this precludes
-        the use of requires/conflicts
 
+    -- Set to true to indicate that this mod only affects the user interface. If this flag is false
+    -- (the default), all players in a multiplayer game must have it in order for anyone to use it.
+    -- If true, one player can have it active independently of the other players.
     ui_only = false
-        Set true to indicate that this mod only affects the user interface. If
-        this flag is false (the default), all players in a multiplayer game must
-        have it in order for anyone to use it. If true, one player can have it
-        active independently of the other players. (Games on GPGNet may restrict
-        this further.)
 
-        TODO: If ui_only is set true, the mod's files should not even be accessible
-        from within the sim, but we need the VFS stuff implemented properly first.
-
+    -- Path of the icon to use for this mod in the mods manager window. The default is
+    -- `"mod_icon.dds"` and in the same directory as the `mod_info.lua` file.
+    -- Note that this is unrelated to the `mod_icons.lua` file explained below.
     icon = "mod_icon.dds"
-        Name of icon to use for this mod in the mod selection window. The default is
-        mod_icon.dds in the same directory as the mod_info.lua file.
 
+	-- Optionally indicates that this mod only works if another mod is also present. Does not affect
+    -- mod loading order.
+    -- It might be nice if you add a comment after each UID to denote the name of the mod so mere
+    -- mortals can maintain your list :)
     requires = {
-        -- Optionally indicates that this mod only works if another mod is also present.
-        -- It might be nice if you add a comment after each uid to denote the name of the mod
-        -- so mere mortals can maintain your list :)
         "fec58b30-0036-4b9e-9995-fe2d6fe4c6e9",     -- Chris' Whacky Mod
         "18e391bc-67aa-49fb-8a5f-34efec2d1c2c",     -- Jeffs Cool Mod
         "61d2cb50-08ac-46ba-b261-3c3e3372aef0",     -- Teds Dumb Mod
     }
 
-    conflicts = {
-        Indicate any other mods that this mod is known to conflict with; the game
-        will refuse to enable both of them at the same time.
-        Same format as 'requires'.
-    }
-
-    before = {
-        List of other mod names. If this mod happens to be active at the same time
-        as any of the named other mods, it will be applied before them.
-        Same format as 'requires'.
-    }
-
-    after = {
-        List of other mod names. If this mod happens to be active at the same time
-        as any of the named other mods, it will be applied after them. If you do
-        not supply an 'after' list, the 'requires' list will be used in its place.
-        Same format as 'requires'.
-    }
-
+	-- Optional table which will map the required UID's to friendly names so the user of the mod can
+	-- determine what they are missing.
     requiresNames = {
-        -- an optional table which will map the required uids to friendly names so the user of the mod can determine what they are missing
         ["fec58b30-0036-4b9e-9995-fe2d6fe4c6e9"] = "Chris' Whacky Mod v123",
         ["18e391bc-67aa-49fb-8a5f-34efec2d1c2c"] = "Jeff's Cool Mod",
         ["61d2cb50-08ac-46ba-b261-3c3e3372aef0"] = "Ted's Dumb Mod",
     }
+    
+	-- Optionally indicates any other mods that this mod is known to conflict with; the game will
+	-- refuse to enable both of them at the same time. You do not need to add old versions of your
+    -- mod here (unless you'd like to be extra careful); the client determines mods with the same
+    -- name to be the same mod and auto-updates them. Same format as `requires`.
+    conflicts = {
+    }
 
-Fields 'id' and 'location' also get filled in during initialization, to the full
-path to the mod_info file and the directory containing it.
+	-- Optional list of other mod UID's. If this mod happens to be active at the same time as any of
+	-- the named other mods, it will be applied before them. Same format as `requires`.
+    before = {
+    }
 
-
-
-Adding extra maps
-
-   Any extra maps in a mod will automatically show up on the available maps list on
-   the single- and multi-player skirmish screen, as long as they have an appropriate
-   Configuration set up.
-
-   If you select a map from a mod for play, that mod (and any other mods it depends on)
-   will automatically be active for that session. This allows you to define custom
-   units or props just for a particular set of maps.
-
-
- Adding custom units
-
-   If an active mod has any .bp files defining units, they will be automatically
-   included during blueprint loading.
-
-   If a mod contains a blueprint matching the ID of an existing unit (e.g. uel0001)
-   then the mod's blueprint will replace the original. If the "Merge=true" flag is
-   set in the mod blueprint, it will be merged with the original can can override
-   some fields while leaving others intact.
+	-- Optional list of other mod UID's. If this mod happens to be active at the same time as any of
+	-- the named other mods, it will be applied after them. If you do not supply an `after` list,
+	-- the `requires` list will be used in its place (if present). Same format as `requires`.
+    after = {
+    }
 
 
- Balance changes
+`location` also gets filled in before initialization (in case you want to use it for relative paths)
+to the full path to the mod_info file and the directory containing it. It is also set afterwards as
+well, so changing this has no affect.
 
-   There are two ways to make large-scale balance changes.
+There are a couple other fields that a mod can specify - it's best to leave them alone.
 
-   First, you can define one or more .bp files with "Merge=true", that override
-   particular fields for many units. This is how the campaign balance mod works.
+----------
+Adding custom units
+----------
 
-   Second, you can define a function <TODO>, which is called on each blueprint
-   as it is loaded. This function can then manipulate the blueprint in arbitrary
-   ways.
+If an active mod has any .bp files defining units, they will be automatically included during
+blueprint loading.
 
-   First, a mod can simply add a new blueprint file that defines a new blueprint.
-
-   Second, a mod can contain a blueprint with the same ID as an existing blueprint.
-   In this case it will completely override the original blueprint. Note that in
-   order to replace an original non-unit blueprint, the mod must set the "BlueprintId"
-   field to name the blueprint to be replaced. Otherwise the BlueprintId is defaulted
-   off the source file name. (Units don't have this problem because the BlueprintId is
-   shortened and doesn't include the original path).
-
-   Finally, a mod can define a ModBlueprints() function which manipulates the
-   original_blueprints table in arbitrary ways. [How/when exactly is this done?]
+If a mod contains a blueprint matching the ID of an existing unit (e.g. `"uel0001"`) then the mod's
+blueprint will replace the original. If the `Merge = true` flag is set in the mod blueprint, it will
+be merged with the original can can override some fields while leaving others intact.
 
 
- Custom skins
+----------
+Balance changes
+----------
 
-   A mod can add a ??? file to indicate that it contains a custom skin. The skin
-   will automatically be made available on the list of available skins.
-   TODO: need help from CBlackwell to work out and document custom skin system.
+There are two ways to make balance changes.
 
+First, a mod can contain a blueprint with the same ID as an existing blueprint. In this case it will
+completely override the original blueprint. Note that in order to replace an original non-unit
+blueprint, the mod must set the "BlueprintId" field to name the blueprint to be replaced. Otherwise,
+the BlueprintId is defaulted off the source file name. (Units don't have this problem because the
+BlueprintId is shortened and doesn't include the original path). Setting `Merge = true` while doing
+this will make it so that only the new fields in the modded blueprint are overridden, and the rest
+from the original blueprint are kept (possibly also changed from other mods).
 
- Other changes
+Second, a mod can hook the `/lua/system/Blueprints.lua` file and override the `ModBlueprints` to
+modify existing blueprints in an organized manner. Be sure to call the old function like this, so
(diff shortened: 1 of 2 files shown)


Garanas:
@The-Balthazar would you have the time to review this pull request?

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
+    github = "https://github.com/FAForever"
+
+    -- Uniquely identify this mod. This defaults to the mod's name, but that's not a very safe way
+	-- to keep mods differentiated. I'd recommend using a GUID or Guaranteed Unique ID.
+    -- Every new version of a mod should get a new GUID so that mods which rely on the old version
+    -- can select it appropriately, and so that the FAF client can keep track.
+    -- The following webpages allow you to generate a GUID online (as of  April 2023)
+    --   http://www.somacon.com/p113.php
+    --   http://www.famkruithof.net/uuid/uuidgen
     uid = B019FBEE-E411-11DB-AFAE-13D355D89593
```
The-Balthazar, line 71: The uid value should be in quotation marks.

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
-        1.  A mod containing just custom maps, along with textures and props used
-            on those maps. The mod will be automatically enabled when one of its
-            maps is used.
-
-        2.  A mod supplying textures and props for use by other mods. Those mods
-            indicate that they require this one, and it is automatically enabled
-            when they are.
 
+    -- Setting enabled to false causes a mod to never be loaded. Provides an easy way to disable a
+    -- mod during development. The default is true.
```
The-Balthazar, line 80: Another use for this value being false is for defining dummy mods that contain other mods, since the game supports nested mods even if the client doesn't.

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
     enabled = true
-        Setting enabled to false causes a mod to never be loaded. Provides
-        an easy way to disable a mod during development. The default is true.
 
+    -- A simple form of preventing conflicting mods. If a mod has `exclusive`, then no other mod can
+    -- be active. The default is false. Note that this precludes the use of `conflicts`.
```
The-Balthazar, line 84: This is somewhat an aside, but I've never seen a good reason to have this `true`.

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
-        have it in order for anyone to use it. If true, one player can have it
-        active independently of the other players. (Games on GPGNet may restrict
-        this further.)
-
-        TODO: If ui_only is set true, the mod's files should not even be accessible
-        from within the sim, but we need the VFS stuff implemented properly first.
 
+    -- Path of the icon to use for this mod in the mods manager window. The default is
+    -- `"mod_icon.dds"` and in the same directory as the `mod_info.lua` file.
+    -- Note that this is unrelated to the `mod_icons.lua` file explained below.
```
The-Balthazar, line 94: Unless it's changed since I last checked, a dds wont load in the client, and it's relative to root, not to the mods folder.

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
+to the full path to the mod_info file and the directory containing it. It is also set afterwards as
+well, so changing this has no affect.
+
+There are a couple other fields that a mod can specify - it's best to leave them alone.
+
+----------
+Adding custom units
+----------
+
+If an active mod has any .bp files defining units, they will be automatically included during
```
The-Balthazar, line 144: active *sim* mod

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
-   units or props just for a particular set of maps.
+First, a mod can contain a blueprint with the same ID as an existing blueprint. In this case it will
+completely override the original blueprint. Note that in order to replace an original non-unit
+blueprint, the mod must set the "BlueprintId" field to name the blueprint to be replaced. Otherwise,
+the BlueprintId is defaulted off the source file name. (Units don't have this problem because the
+BlueprintId is shortened and doesn't include the original path). Setting `Merge = true` while doing
+this will make it so that only the new fields in the modded blueprint are overridden, and the rest
+from the original blueprint are kept (possibly also changed from other mods).
 
+Second, a mod can hook the `/lua/system/Blueprints.lua` file and override the `ModBlueprints` to
```
The-Balthazar, line 166: It's a bit structurally weird to explain a thing that can be done with a hook before explaining how to actually hook a file.

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
 
-   Finally, a mod can define a ModBlueprints() function which manipulates the
-   original_blueprints table in arbitrary ways. [How/when exactly is this done?]
 
+----------
+Strategic Icon Replacement
+----------
+
+A UI mod can contain a `mod_icons.lua` file to specify strategic icons to replace for particular
+units. This is one of the few times a UI mod is allowed to modify things in the simulation
```
The-Balthazar, line 217: It's probably more accurate to say that the sim loads them from the UI mod, rather than the UI actually having any power in the situation.

The-Balthazar:
I probably should have marked all that as a review. I wasn't actually planning to go through the whole of it right now.

Hdt80bro:
Oh that's quite alright. It's good to have any sort of feedback - some of what you commented was old stuff I wasn't sure what to do with, but now I have an idea of what to change. Thanks!

lua/MODS.LUA
Hdt80bro: Good catch - it actually defaults to `/textures/ui/common/dialogs/mod-manager/generic-icon_bmp.dds` not a file in the mod folder as well. Not that the FAF client cares, of course.

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
+ * If you upload the mod to the vault, you MUST follow the vault rules:
+   https://wiki.faforever.com/en/Vault-Rules
+    * In particular, make sure you own all the content of the mod you are uploading. Require other
+      mods when available to avoid copying others' work.
+ * Mods are not able to modify the lobby or provide game options (since mods are only fully loaded
+   after the game launches).
+ * All of a mod's files are mounted behind the mod directory to avoid conflicts. Outside of lua
+   scripting (which can always hook original and FAF files), this means that files are never
+   replaced (in practice, since the location a resource such as a texture points to can be changed
+   by a lua file, this only really affects shaders, which are accessed by the game engine directly)
```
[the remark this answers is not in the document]
Hdt80bro: Mmh, but that stems from how blueprints are loaded, not because the files are overwritten

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
+able to flexibly support the modding scene, but beyond these basic managerial requirements, they are
+pretty much free to do whatever they like with the game and the FAF framework.
+
+A couple (but not all) of the restrictions include:
+ * If you upload the mod to the vault, you MUST follow the vault rules:
+   https://wiki.faforever.com/en/Vault-Rules
+    * In particular, make sure you own all the content of the mod you are uploading. Require other
+      mods when available to avoid copying others' work.
+ * Mods are not able to modify the lobby or provide game options (since mods are only fully loaded
+   after the game launches).
```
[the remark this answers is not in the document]
Hdt80bro: That's what I thought, but when I checked the code, it only looked like it would work for maps.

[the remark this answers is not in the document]
Hdt80bro: Oh we *do* [import](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/lobby/lobby.lua#L89) all lobby options of *all* mods in the user's mod folder (regardless of if they're used). That is good document.

lua/MODS.LUA
```diff
@@ -1,306 +1,419 @@
...
+    -- A simple form of preventing conflicting mods. If a mod has `exclusive`, then no other mod can
+    -- be active. The default is false. Note that this precludes the use of `conflicts`.
     exclusive = false
-        A simple form of preventing conflicting mods. Only one 'exclusive' mod
-        can be active at a time. The default is false. Note that this precludes
-        the use of requires/conflicts
 
+    -- Set to true to indicate that this mod only affects the user interface. If this flag is false
+    -- (the default), all players in a multiplayer game must have it in order for anyone to use it.
+    -- If true, one player can have it active independently of the other players.
```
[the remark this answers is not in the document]
Hdt80bro: That's fair - though it doesn't even prevent access to sim files, the UI simply doesn't use them.

lua/MODS.LUA
Hdt80bro: Wouldn't making them unselectable be better?

lua/MODS.LUA
Hdt80bro: I can change the wording to be passive, but the UI mod does have that power - probably too much. I've already found a way for a UI mod to exploit this.

lua/MODS.LUA
Garanas: It can be a `png` or other 'normal' file formats, the game can read those

lua/MODS.LUA
Garanas: I'd like to know more about that exploit 😄

lua/MODS.LUA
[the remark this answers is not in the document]
The-Balthazar: If the point of the statement is avoiding conflicts, then that's not an important destinction.

lua/MODS.LUA
The-Balthazar: Setting them to unselectable would allow things outside of the game, such as the client, to enable them, which would defeat the whole point and also load all blueprints from all contained mods, since they are technically also within the container mod.

lua/MODS.LUA
The-Balthazar: `png` or `jpg` are what I'd recommend for being readable by both the client and the game.

lua/MODS.LUA
The-Balthazar: Did we not do a good enough job sandboxing?

lua/MODS.LUA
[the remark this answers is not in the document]
Hdt80bro: Filename conflicts and replacement, not logical conflicts in general. I'll add the distinction.

lua/MODS.LUA
```diff
@@ -137,41 +144,80 @@ well, so changing this has no affect.
...
+    end
+
+
 ----------
 Adding custom units
 ----------
 
-If an active mod has any .bp files defining units, they will be automatically included during
-blueprint loading.
+If an active sim mod has any .bp files defining units, they will be automatically included during
```
The-Balthazar, line 195: You could change this to
> If an active sim mod has any .bp files,

And it'll still be true and technically more accurate, since it'll load non-unit bps too.

lua/MODS.LUA
```diff
@@ -137,41 +144,80 @@ well, so changing this has no affect.
...
+blueprint loading. The blueprint id will be determined from the source file name if not specified.
 
-If a mod contains a blueprint matching the ID of an existing unit (e.g. `"uel0001"`) then the mod's
-blueprint will replace the original. If the `Merge = true` flag is set in the mod blueprint, it will
-be merged with the original can can override some fields while leaving others intact.
+If the blueprint contains a `BlueprintId` field matching the ID of an existing unit
+(e.g. `BlueprintId = "uel0001"`) then the mod's blueprint will replace the original. If the
+`Merge = true` flag is set in the mod blueprint, it will be merged with the original and can
+override some fields while leaving others intact (possibly also changed by other mods).
```
The-Balthazar, line 201: This would be better split into two clauses, with the `BlueprintId` part being on the end of
> The blueprint id will be determined from the source file name if not specified.

stating how the ID's specified outside of sourcing it from the filename, and the second part being that it'll merge or replace if the ID, however it evaluated it, matches an existing blueprints ID, since this currently implies that you can only merge or replace if you define `BlueprintId`.
It might also be a good place to mention that you can have multiple blueprints in a single `.bp` file if they specify `BlueprintId`. (Technically I think one/the first can get away without, but that's asking for trouble.)

lua/MODS.LUA
Hdt80bro: Hmm, it's in the section on units, but that's fine I guess

lua/MODS.LUA
Hdt80bro: That's a good call.

I'd ideally prefer to completely include all information that describes how things work, but I suspect the bit about having multiple blueprints per file might be something better left out.

The-Balthazar: Having multiple blueprints in a single file is very useful for doing a lot of small similar changes to several units that isn't worth setting up a `blueprints.lua` for, like for example:
```lua
UnitBlueprint {
    Merge = true,
    BlueprintId="uab5101",	--Aeon--
    BuildIconSortPriority = 150,
}
UnitBlueprint {
    Merge = true,
    BlueprintId="xsb5101",	--Seraphim--
    BuildIconSortPriority = 150,
}
UnitBlueprint {
    Merge = true,
    BlueprintId="ueb5101",	--UEF--
    BuildIconSortPriority = 150,
}
UnitBlueprint {
    Merge = true,
    BlueprintId="urb5101",	--Cybran--
    BuildIconSortPriority = 150,
}
```
The part that's not worth mentioning is that you can functionally get away with not defining `BlueprintId` on one of them and have that one evaluate from the filename.

lua/MODS.LUA
The-Balthazar: Saying "bp files defining units" implies that other blueprints function differently. It's like saying "I used to drink." and following up with "I still do, but I used to as well." Technically not wrong, but has implications that are wrong.

Hdt80bro: It sounds like you'd rather have this be a section on blueprints in general instead of units. That's fine.

lua/MODS.LUA
Hdt80bro: I can see that, I'll update it

lua/MODS.LUA
```diff
@@ -1,306 +1,506 @@
...
+   https://wiki.faforever.com/en/Vault-Rules
+    * In particular, make sure you have the right to distribute the content of the mod you are
+      uploading. Require other mods when available to avoid copying others' work.
+ * Mods are not able to modify the lobby (since mods are only fully loaded after the game launches).
+    * However, you can provide AI lobby options (see below).
+ * All of a mod's files are mounted behind the mod directory to avoid naming conflicts or accidental
+   file replacement. Outside of lua scripting (which can always hook original and FAF files), this
+   means that files are never replaced 
+    * In practice, since the location a resource such as a texture points to can be changed by a lua
+      file, this only really affects shaders, which are accessed by the game engine directly and
```
The-Balthazar, line 25: Aren't sounds in a similar situation to shaders?

lua/MODS.LUA
```diff
@@ -1,306 +1,506 @@
...
+----------
+Mod info file
+----------
+
+Each mod has a `mod_info.lua` file which contains various bits of info:
+
+    -- Required name to use for this mod. This is what is displayed in the Lobby, after undergoing
+	-- title-casing. Do NOT add version information here - keep this the exactly same when you
+    -- update what you consider to be the same mod as the FAF client will uses this to auto-update
+    -- mods (based on the most recent `uid` - the version number does not matter and is only used
```
The-Balthazar, line 40: Is this accurate? How does it determine which is the most recent uid if not by version number?

lua/MODS.LUA
```diff
@@ -1,306 +1,506 @@
...
+    -- for display purposes).
+    name = "Happy mod"
+  
+    -- Optional integer version to use for this mod (please only increment) to indicate to humans
+    -- the most current version. Decimals are not supported. If updating the mod, make sure to
+    -- change the `uid` as that is what the FAF client actually uses and leave the name exactly the
+    -- same.
+    version = 4
+
+    -- Optional copyright info. If absent, the mod author is presumed to maintain full ownership of
```
The-Balthazar, line 50: Except as defined in the FAF vault license.

lua/MODS.LUA
```diff
@@ -1,306 +1,506 @@
...
-        Uniquely identify this mod. This defaults to the mods name, but that's
-        not a very safe way to keep mods differentiated. I'd recommend using a GUID
-        or Guaranteed Unique ID. Also, every new version of a mod should get a new
-        GUID so that mods which rely on the old version can select it appropriately.
-        This following webpages allow you to generate a guaranteed unique ID on line (as of 4/6/07)
-        http://www.somacon.com/p113.php
-        http://www.famkruithof.net/uuid/uuidgen
 
+    -- Required name(s) of author(s). Shows up as `"UNKNOWN"` if left blank. Will be begin to be
+    -- truncated in-game if over 20 characters. Make sure this reflects who is uploading the mod to
```
The-Balthazar, line 58: Do we really want to be encouraging the uploader to put their name in the author field if they aren't the author?

lua/MODS.LUA
```diff
@@ -1,306 +1,506 @@
...
-            on those maps. The mod will be automatically enabled when one of its
-            maps is used.
-
-        2.  A mod supplying textures and props for use by other mods. Those mods
-            indicate that they require this one, and it is automatically enabled
-            when they are.
 
+    -- Setting enabled to false causes a mod to never be loaded. Provides an easy way to disable a
+    -- mod during development or for dummy mod packs that nest other mods inside directly (please
+    -- use the requires mechanism instead however). The default is true.
```
The-Balthazar, line 85: Using the requires mechanism isn't necessarily an appropriate alternative to mod packs if you're using packs to avoid uploading like 40 separate mods, which is what I use it for.

lua/MODS.LUA
```diff
@@ -1,306 +1,506 @@
...
-   Finally, a mod can define a ModBlueprints() function which manipulates the
-   original_blueprints table in arbitrary ways. [How/when exactly is this done?]
+----------
+Blueprints
+----------
+
+If an active sim mod has any .bp files, they will be automatically be run during blueprint loading
+to register all blueprint definitions inside. The ID of the blueprint can be specified by the
+`BlueprintId` field, but if absent, it will default to the source filename (lowercased) without the
+suffix (e.g. `"_unit.bp"`). For multiple blueprints in a file, be sure to specify each
```
The-Balthazar, line 198: Since this is now generic to bp files rather than units, this should specify that units default to lowercase filename sans `_unit.bp`, while all other blueprints default to lowercase filepath including `_proj.bp` ect..

lua/MODS.LUA
Hdt80bro: No, the sound banks are still referenced in Lua like textures.

lua/MODS.LUA
Hdt80bro: The vault uses upload time

lua/MODS.LUA
Hdt80bro: Don't they become responsible for the uploaded mod at that point?

lua/MODS.LUA
```diff
@@ -1,306 +1,506 @@
...
+    author = "Joe"
+    
+    -- Optional URL to anywhere author likes. Will show up in the mods manager if valid.
+    url = "https://www.faforever.com"
+    
+    -- Optional URL to anywhere author likes. Will show up in the mods manager if valid.
+    github = "https://github.com/FAForever"
+
+    -- Uniquely identify this mod. This defaults to the mod's name, but that's not a very safe way
+	-- to keep mods differentiated. I'd recommend using a GUID or Guaranteed Unique ID.
```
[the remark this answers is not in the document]
Hdt80bro: Yup, another leftover

lua/MODS.LUA
Hdt80bro: The reality is that it doesn't care what the ending is, it truncates everything from `_` to `.bp`

lua/MODS.LUA
Hdt80bro: Would a good way to say that be it must be compatible with https://content.faforever.com/tos/vault-license.html and defaults to that if absent?

lua/MODS.LUA
Hdt80bro: Oh, it does only call `SetShortId` for `UnitBlueprint` and `SetBackwardsCompatId` for everything else. That's a good catch.

lua/MODS.LUA
Hdt80bro: Mmh, that's an interesting perspective. I'll remove that bit.

lua/MODS.LUA
The-Balthazar: Compared to what though for things not in the vault? Does the vault just assume it's more up to date than local files?

lua/MODS.LUA
The-Balthazar: It's an 'author' field, not a 'who is responsible' or a 'care of' field. The vault tracks the uploader separately, and if the uploader hasn't actually contributed to the content then they aren't an author, regardles of if they have permission to upload or not.
I delegated uploading my main mods to someone else. They were not an author of the content and I would not have been happy if they put their name in the author field or if they were encouraged to do so.

lua/MODS.LUA
The-Balthazar: "If absent, mod author is presumed to reserve all rights except as required by ~~"
This is just documentation after all, not an actual part of any license or rights agreement, so we can't assume or state that it's anything more than would be required.

lua/MODS.LUA
The-Balthazar: They aren't mounted inside the mod dir though, they're mounted to `/sounds/`

lua/MODS.LUA
Hdt80bro: My understanding is that it takes the UID of the local mod and looks up the uploaded time of that UID to compare - I'm not sure what happens when there's a conflict between a local unuploaded mod and when hosting with a mod of the same name that's in the vault. Sounds like a weird edge case, though I suspect it just leaves the unuploaded one alone and would update the in the vault. Perhaps @Sheikah45 can offer some insight or tell me if I'm totally off base.

lua/MODS.LUA
Hdt80bro: I'll remove it then.

lua/MODS.LUA
Hdt80bro: I'll remove that bit then

lua/MODS.LUA
Hdt80bro: Ah, sound mountpoints. So those wouldn't be untouchable like shaders and could produce conflicts then (when they do end up working correctly) - I'll add that. I didn't realize only maps could mount movies, that's good to know.

lua/MODS.LUA
Sheikah45: It doesn't use uploaded time. FAF track versions of the mods separately. Each new upload increments the version by one of the mod. We use UID to find the mod and look up its most recent uploaded version and install that. It will overwrite any non matching local mods.

Hdt80bro: So it's something like this @Sheikah45 ?
* A mod author uploads a mod named "Mod Name" to the vault with a particular UID and FAF attaches an internal version number to it - the version in the mod info is cosmetic.
* A player downloads the mod.
* Later, the mod author updates the mod by uploading a mod with the same "Mod Name" name but with a different UID. FAF recognizes the mod as being a newer version of the same mod because it has the same name with a different UID and the internal version number is incremented - hopefully the mod author thought to change the mod info version.
* The player then hosts with the "Mod Name" mod.
* FAF then looks up the most recent UID of the mod named "Mod Name" based on the internal version, finds that this new mod is more recent than the one in local storage, and auto-updates the old mod.

Sheikah45: Correct

The-Balthazar: If the `mod_info` `version` is entirely cosmetic, why is it restricted to integers?

Sheikah45: Ah nevermind just double checked and we do use the version as defined in the lua. But the uid is still used to query the mods so trumps anything else.

The-Balthazar: Why is this marked as resolved when it still says the version number doesn't matter after it was established that it does? @Sheikah45 could you clarify specifically how it is used?

lua/MODS.LUA
```diff
@@ -1,306 +1,507 @@
...
-    uid = B019FBEE-E411-11DB-AFAE-13D355D89593
-        Uniquely identify this mod. This defaults to the mods name, but that's
-        not a very safe way to keep mods differentiated. I'd recommend using a GUID
-        or Guaranteed Unique ID. Also, every new version of a mod should get a new
-        GUID so that mods which rely on the old version can select it appropriately.
-        This following webpages allow you to generate a guaranteed unique ID on line (as of 4/6/07)
-        http://www.somacon.com/p113.php
-        http://www.famkruithof.net/uuid/uuidgen
 
+    -- Required name(s) of author(s). Shows up as `"UNKNOWN"` if left blank. Will be begin to be
```
The-Balthazar, line 58: Typo "Will be begin to be" should probably be "Will be"

lua/MODS.LUA
```diff
@@ -1,306 +1,507 @@
...
-        this flag is false (the default), all players in a multiplayer game must
-        have it in order for anyone to use it. If true, one player can have it
-        active independently of the other players. (Games on GPGNet may restrict
-        this further.)
-
-        TODO: If ui_only is set true, the mod's files should not even be accessible
-        from within the sim, but we need the VFS stuff implemented properly first.
 
+    -- Path of the icon to use for this mod in the mods manager window. The game defaults to
+    -- `"/textures/ui/common/dialogs/mod-manager/generic-icon_bmp.dds"`. The FAF client does cannot
```
The-Balthazar, line 100: Typo "does cannot"

lua/MODS.LUA
```diff
@@ -1,306 +1,507 @@
...
+        "61d2cb50-08ac-46ba-b261-3c3e3372aef0",     -- Ted's Dumb Mod
     }
 
+    -- Optional table which will map the required UID's to friendly names so the user of the mod can
+    -- determine what they are missing.
+    requiresNames = {
+        ["fec58b30-0036-4b9e-9995-fe2d6fe4c6e9"] = "Chris' Whacky Mod", -- v123
+        ["18e391bc-67aa-49fb-8a5f-34efec2d1c2c"] = "Jeff's Cool Mod",
+        ["61d2cb50-08ac-46ba-b261-3c3e3372aef0"] = "Ted's Dumb Mod",
+    }
```
The-Balthazar, line 121: Is `requires` required if `requiresNames` is defined with the same -and more- data? Seems redundant.

lua/MODS.LUA
```diff
@@ -1,306 +1,507 @@
...
+        },
+    },
+
+
+----------
+Strategic Icon Replacement
+----------
+
+A UI mod can contain a `mod_icons.lua` file to specify strategic icons to replace for particular
+units. This is one of the few times a UI mod will be able to have an affect things in the simulation
```
The-Balthazar, line 305: Typo; "to have an affect things", should be either "to affect things" or "to have an effect on things"

lua/MODS.LUA
Sheikah45: It is used with an existence check if that version of the mod exists in the vault already and to name the zip folder

The-Balthazar: Does that mean it's used to indicate which is newer rather than a timestamp?

Sheikah45: Yes

lua/MODS.LUA
Hdt80bro: Currently, yes. The intention is that `requires` is the array, and `requiresNames` is the look up for the name of the UID. The same could be done with `before`, `after`, and `conflicts`. However, combining required data with optional data is a little weird, particularly because `requiresNames` can be partial. I don't think we can change this at this point though.

lua/MODS.LUA
```diff
@@ -1,306 +1,506 @@
...
+FAF framework. They differ from regular Supreme Commander mods, i.e. those that directly replace
+and hook original .scd files on startup. In a sense, FAF itself is a Supreme Commander mod that
+coordinates the participation of other mods intended to be used inside of it once the game launches.
+In this light, there are certain technical limitations and restrictions that are necessary to be
+able to flexibly support the modding scene, but beyond these basic managerial requirements, they are
+pretty much free to do whatever they like with the game and the FAF framework.
+
+Several (but not all) of the restrictions include:
+ * If you upload the mod to the vault, you MUST follow the vault rules:
+   https://wiki.faforever.com/en/Vault-Rules
```
The-Balthazar, line 16: Dead link. It's currently `https://wiki.faforever.com/en/Development/Vault/Vault-Rules`, although I don't like the redundancy of `Vault/Vault`, so I'm going to see about getting it changed to `https://wiki.faforever.com/en/Development/Vault/Rules` and also inquire about redirects, 'cause this is a change that has broken a lot of links.

Should also mention that they need to follow the FAF rules as well as the vault rules, `https://faforever.com/rules`, which prohibit reclaim macros and autoclickers.

The-Balthazar (changes requested):

lua/MODS.LUA
The-Balthazar: It's now `https://wiki.faforever.com/en/Development/Vault/Rules`

The-Balthazar (approved):

PR #4962, opened 2023-05-23
T3MAA hp buff

Basilisk3:
**Defense**
Hitpoints of all T3 MAA units: +400

These units have very low amount of hitpoints for their mass cost. For example, the Bouncer has 600 fewer hitpoints than a Loyalist (a unit that is considered to be very squishy) while costing 320 more mass. This makes sniping them very easy when using them offensively (their intended role). 

Note, that this does not help them deal with tech 3 bombers, because they still die in one pass. This is still a problem, but I am unsure if the increasing their hitpoints by that much is a good idea gameplay wise.


RabidPope:
Was thinking of this, but I would go for a cost decrease instead. Makes them more spammable.

Tagada14:
Will discuss it next meeting

Spikey84:
We discussed and decided not to make this change right now. We think its interesting and that a change to at least the Cybran MAA may be desired, but we are going to wait and do that a later time.

PR #4963, opened 2023-05-23
Cybran red channel

BlackYps:
Fixes the long outstanding issue that most cybran units don't have values greater 0 in the red channel in their specTeam texture.
This finally allows us to tweak the metalness of cybran units individually. For most units I recreated the values that they had in the file I gave out for testing the shader. Some units got tweaks on top of that. The monkeylord, the naval factory and most civilian buildings. We could still improve the looks by adjusting the values more for each unit, but that takes a lot of time. At least now that the textures are in the repo and filled with sensible values, it will be much easier to tweak them.

Monkeylord before and after manual tweaking of the texture. The difference is subtle.
Screenshot 2023-05-24 003211
Screenshot 2023-05-24 002715

The texture changes also affect the current shader. I made some adjustments so the impact should be minimal. Here is a comparison of the naval factory using the old shader.
Screenshot 2023-05-24 001943
Screenshot 2023-05-24 002212


Garanas:
With:
image

Without:
image

With:
image

Without:
image

PR #4964, opened 2023-05-24
Apply water color to wreckages, props, shield and unit shaders

BlackYps:
The currently used shaders could do with some water effects as well. This implementation is more simple as I have to accomodate for the way the water ramp is applied to the terrain and I don't want to change that for existing maps. But it works well enough.
Screenshot 2023-05-24 203709
Screenshot 2023-05-24 203348

### DIFF
--- a/effects/mesh.fx
+++ b/effects/mesh.fx
@@ -3484,6 +3484,13 @@ float4 NormalMappedInsectPS_02( NORMALMAPPED_VERTEX vertex, uniform bool hiDefSh
     return float4( color, alpha);
 }
 
+float shieldWaterAbsorption(float depth) {
+    float factor = 1.0;
+    if (depth < 0) factor = 0.6;
+    factor *= 1 - tex1D(WaterRampSampler, (-depth / (surfaceElevation - abyssElevation))).w;
+    return factor;
+}
+
 /// ShieldPS
 ///
 ///
@@ -3527,6 +3534,7 @@ float4 ShieldPS( EFFECT_VERTEX vertex ) : COLOR
 
     // Add in our alpha channel to mask UV pinching at the top of the sphere
     color.a *= colorMask.a;// * 0.75;
+    color.a *= shieldWaterAbsorption(vertex.depth.x);
 
     return color;
 }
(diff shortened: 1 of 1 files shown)


Garanas:
This is great, look at this Megalith:

image

Garanas:
It also looks great when the Megalith moves out of the water, great effect

Garanas:
image

image

I think the effect is great. It does make some units hard to see (like the selected percivals), but for the competitive players you can toggle the strategic icons to always render. That way you can always see them.

Garanas:
@BlackYps What are your thoughts on applying it to props and shields too?

image

Garanas:
image

You can spawn in props via the cheat menu by clicking on the little wrench

BlackYps:
I need to be able to access the height of the mesh, but I think that should also be possible for shields and props. I'll look into it

BlackYps:
Props actually use various different shaders. It's a bit hard to track down, but I will manage. Shields however are surprisingly complicated. They use different magic with the color or alpha values per faction, so they react totally different to simply adding a color. On top of that the shield isn't affected by the water refraction. My first thought was that it depends on the render stages settings, but I randomly played around with them and found no effects. Do you know how or for what they are used?

Garanas:
> My first thought was that it depends on the render stages settings, but I randomly played around with them and found no effects. Do you know how or for what they are used?

I don't know.

BlackYps:
I think for now I wouldn't alter the shields. It's not straightforward at all and we would first have to define what behaviour we want of the shield when it is underwater

BlackYps:
I got all props now. Should have taken a screenshot, but i flooded the prop map to easily see if any props were missing. I think this PR is ready now.

BlackYps:
I realized it's actually not that hard to change the shields. We just need to reduce the alpha value. I also realized that when the camera is inside the shield dome, the UEF shield was actually the only one that you could see. All the other ones fell victim to backface culling, so I adjusted that as well.
Screenshot 2023-05-28 230411
Screenshot 2023-05-28 230348

BlackYps:
On the shield boat it is still hard to see, but that is because the shield is way more visible at the equator and that one is way lower than the water surface for some reason. I think this could be changed, but should probably be addressed by the balance team, because it affects the area the shield covers underwater.
Screenshot 2023-05-28 231443

Garanas (approved):

PR #4969, opened 2023-05-25
Remove Pointless Random map button in game Options

MrRowey:


Garanas:
What makes it a pointless button?

MrRowey:
> What makes it a pointless button?

85% of the time you select the map you want from the client side and then if you wanted to change it you look through the map list the amount of times I've hit random map by accident when I've met to hit okay as well . So if alot of people use it then sure we can keep it but majority of game select the map from the client and would probably be better to have a. Random maps button Thier than in the game options

Garanas:
Then the question is: how do we find out if people use this?

MrRowey:
Well the easiest way would be to remove it and see how many complain or we do a forum post as see what the outcome of that is?

Garanas:
It is not quite the user friendly approach to just remove it.

Garanas:
I'm going to close this pull request. Given that the button is in a sub menu it is quite difficult to accidentally hit it. There's also nothing to replace it with: we don't need space for some new feature at the moment.

PR #4972, opened 2023-05-27
Pbr terrain improvement

BlackYps:
Instead of layering two textures at different scales we now sample at the same scale but with two different rotations. Then we blend between them with a relatively sharp transition to make the visual texture repetition go away.

Screenshot 2023-05-26 211430
Screenshot 2023-05-27 140856
Screenshot 2023-05-27 141439
Screenshot 2023-05-27 141228

### DIFF
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -1862,14 +1862,14 @@ float4 TerrainPBRAlbedoPS ( VS_OUTPUT inV) : COLOR
 
     float cameraFractionNear = 1;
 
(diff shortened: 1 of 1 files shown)


Garanas:
There's a tiling / square-like pattern when you're zoomed out:

image

BlackYps:
It looks like that is because the textures in the atlas are bleeding into each other

Garanas:
image

Ah, I see now

BlackYps:
We can fix the bleeding once and for all if we have some texture space to spare. If we repeat each texture once, so we have a 2x2 layout of each texture and then only sample the middle, we have enough padding. Of course this doubles the texture size

BlackYps:
I'll take the atlas bleeding issues to a separate pull request

Garanas (approved):

Issue #4980, opened 2023-05-28
Balance idea: Cooper Adjustment


Spikey84:
## Intended changes

Cooper:
```
- Formation change to clump them more
- Increase range by 5
- Decrease hp, mass, e, and build time by 10%
```

## Thoughts on the changes

Coopers will be getting a more tight formation so they cover eachother better with their torp defense. Additionally they get multiple stat changes to get an overall buff.

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

Fichom:
Can you elaborate on what "tighter formation is"? I'm guessing it's how close they are when you move them. That sounds great, but the problem is you usually move them in a group of other units, so that wouldn't affect the grouping? Or am I wrong?

MrRowey:
> Can you elaborate on what "tighter formation is"? I'm guessing it's how close they are when you move them. That sounds great, but the problem is you usually move them in a group of other units, so that wouldn't affect the grouping? Or am I wrong?

yes and if all goes as plan you will be able to try the changes as of next week on fafbeta

PR #4988, opened 2023-05-28
Cooper Adjustments

MrRowey:
Completes #4980


units/XES0102/XES0102_unit.bp
```diff
@@ -128,13 +128,13 @@ UnitBlueprint {
...
-        BuildCostMass = 900,
-        BuildTime = 3600,
+        BuildCostEnergy = 6480,
+        BuildCostMass = 810,
+        BuildTime = 3240,
     },
     Footprint = {
-        SizeX = 2,
-        SizeZ = 6,
+        SizeX = 3,
```
Tagada14, line 136: Is the increase in the X dimension on purpose?

MrRowey: yes

Tagada14: Why was it necessary?

Garanas:
@MrRowey can you add a screenshot of the old formation and the new formation, with just coopers?

Issue #4981, opened 2023-05-28
Balance idea: Shard


Spikey84:
## Intended changes

Shard
```
Decrease the Mass to Built Time ratio to 5
Increase damage by 25%
```

## Thoughts on the changes

Simple buff to make them more viable. 

Idea for a buff was inspired by [Basilisk3](https://github.com/FAForever/fa/pull/4747).

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

PR #4989, opened 2023-05-28
Shard Adjustment

MrRowey:
Completes #4981

### DIFF
--- a/units/UAS0102/UAS0102_unit.bp
+++ b/units/UAS0102/UAS0102_unit.bp
@@ -217,7 +217,7 @@ UnitBlueprint {
             BallisticArc = 'RULEUBA_None',
             CannotAttackGround = true,
             CollideFriendly = false,
-            Damage = 7,
+            Damage = 7.5,
             DamageType = 'Normal',
             DisplayName = 'Sonic Pulse Battery',
             FireTargetLayerCapsTable = {
--- a/units/UAS0102/UAS0102_unit.bp
+++ b/units/UAS0102/UAS0102_unit.bp
@@ -217,7 +217,7 @@ UnitBlueprint {
             BallisticArc = 'RULEUBA_None',
             CannotAttackGround = true,
             CollideFriendly = false,
-            Damage = 7.2,
+            Damage = 7,
             DamageType = 'Normal',
             DisplayName = 'Sonic Pulse Battery',
             FireTargetLayerCapsTable = {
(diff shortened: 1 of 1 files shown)


Basilisk3:
I don't know a single unit that does x.y amounts of damage. It is better if the values are rounded to the nearest integer, because the UI does not indicate if a unit has 0.x hp, for example. 

Also, the unit is comically fast (a **third** faster than frigates, which are already extremely fast) and tiny, which results in them driving in front of your other naval units and dying. There is no reason for the unit to be kept in this state, size increases and speed nerfs have already happened to other naval support units, for similar reasons. 

I can understand, if you did not like the numbers I proposed in my pull request but touching the unit's speed and size are mandatory in my book.

Hdt80bro:
Perhaps these should be brought to the balance team's attention
https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UEL0203/UEL0203_unit.bp#L304
https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UEA0103/UEA0103_unit.bp#L263

Garanas:
@MrRowey / @Spikey84 fractional numbers does not work for damage. The engine will round them, not sure in which direction but it doesn't matter. You'll have to make a choice between 7 or 8. @Hdt80bro found a few other cases, best to make a choice there too and round them in the blueprint

Garanas:
Apparently we were wrong @Hdt80bro , the test setup was bad.

```lua
ForkThread(
    function()
        WaitSeconds(1.0)
        LOG("Doing damage")
        local val = 0.2
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
        Damage(nil, self:GetPosition(), self, val, 'Normal')
        WaitTicks(1)
    end
)
````

This does about 4 damage over time, meaning the numbers are fractional but what we get to see (in the UI) is not

Tagada14 (approved):

PR #5003, opened 2023-06-01
Ai/state machine01

relent0r:
Description of changes
This PR converts introduces an adaptive raiding state machine and engineer reclaim state machine.

It adds an inject function into the platoon.lua starting line 4901. This is temporary pending @Garanas doing something better.
Adds platoon templates for state machines and updates builders.
Adds an adaptive specific brain lua and updates the index.lua file.
Adds a utility function for checking threat then units to get correct threat number. Note does not get a position of said threat, may be temporary.

TODO:
- [ ] Rewrite platoon inject method.
- [ ] Add transport support for state machines
- [ ] flesh out state machines further
- [x] Fix economy metrics on the brain (errors right now)
- [ ] Expand state machine usage for reclaiming to all reclaim builders once validated.

Testing,
Run some AI games using the adaptive AI across various map types and validate that the platoons / units are running correctly. Currently alot of logging is included to show state machine switching and error states.


Garanas:
A few comments while looking into your code!

Always place a `return` statement after switching a state, see also [beee7e9](https://github.com/FAForever/fa/pull/5003/commits/beee7e91b99a34b4c330875d7ef5b3e1c2bf9dff). The reason is a bit technical, but it is very possible that switching the state does **not** stop the execution of the current thread. As a result you may still be running code of the previous state after switching. Therefore always place a return statement after switching a state

I've introduced an alternative to `CanPathTo` called `CanPathToCell` which returns true if the given (iMAP) cell contains the label of the origin _somewhere_ in the cell. When we spoke over voice you mentioned edge cases where a bit of reclaim is technically on unpathable terrain. This should alleviate that, but it may also introduce inaccuracies around plateau's. It is not a perfect solution

I noticed that when you run many platoons it quickly clutters up the logs. I've introduced a reference tag that each unit of a given platoon is assigned as a name. The same reference tag is used in the log. You can then filter the log on the reference tag to find out what that specific state machine is printing! This should help with debugging, in the future we'll have better tooling but this is the best we can get for now

image

Garanas:
Fixed the problem with the economy component, you can now use the UI for it to display the current values:

image

image

Garanas:
With regards to the separate brain instances: the goal is that every brain has an entirely separate brain instance. Trying to work with the base brain is going to be difficult, that is where the economy metric error originated from as an example.

At the moment it is a lot of duplicated code, we'll tackle that in the future

relent0r:
@Garanas 
Do you have thoughts on how the initial platoon creation should be handled?
Line 4900 in the platoon file. I'm not a fan of how I'm doing it here. But I'm also unsure how else I could do it.

Garanas:
I'll think about that next 👍

PR #5020, opened 2023-06-05
Split Aeon Gun Range Upgrade

### LINKED ISSUE #5014: Balance idea: Aeon Gun Range Upgrade Adjustment
## Intended changes

```
- Split the gun range upgrade into 2 upgrades
- The first upgrade gives the range of the other gun coms. Cost should be 50% to 75% the current cost.
- The second upgrade gives the range of the current gun com. Cost should be high enough that it requires a t2 pgen or the a very large amount of t1.
```

## Thoughts on the changes

The Aeon gun range upgrade is quite powerful with the current cost. We have split it to allow it to still be strong later but not oppressive in the t1/early t2 phase of the game. This along with the OC nerf should make the Aeon gun com far less oppressive while still being viable. 

MrRowey:
Closes #5014 

Base Range
E/M/BT Cost -30%
Range 30

Advanced Range
E Cost +33%
M/BT  +50%
New Range 35

### DIFF
--- a/units/UAL0001/UAL0001_script.lua
+++ b/units/UAL0001/UAL0001_script.lua
@@ -213,11 +213,11 @@ UAL0001 = ClassUnit(ACUUnit) {
         -- Crysalis Beam
         elseif enh == 'CrysalisBeam' then
             local wep = self:GetWeaponByLabel('RightDisruptor')
-            wep:ChangeMaxRadius(bp.NewMaxRadius or 44)
+            wep:ChangeMaxRadius(bp.NewMaxRadius or 30)
             local oc = self:GetWeaponByLabel('OverCharge')
-            oc:ChangeMaxRadius(bp.NewMaxRadius or 44)
+            oc:ChangeMaxRadius(bp.NewMaxRadius or 30)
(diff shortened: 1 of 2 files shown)


units/UAL0001/UAL0001_script.lua
```diff
@@ -226,6 +226,22 @@ UAL0001 = ClassUnit(ACUUnit) {
             oc:ChangeMaxRadius(bpDisrupt or 22)
             local aoc = self:GetWeaponByLabel('AutoOverCharge')
             aoc:ChangeMaxRadius(bpDisrupt or 22)
+        -- Advanced Cryslised Beam
+        elseif enh == 'AdvancedCrysalisBeam' then
+            local wep = self:GetWeaponByLabel('RightDisruptor')
+            wep:ChangeMaxRadius(bp.NewMaxRadius or 44)
```
Tagada14, line 232: The back-up value should be 30 for the normal one and 35 for advanced

MrRowey: Yhe i need to go back to as 44 is jsut the default since base game, this is why it in a draft state

Tagada14 (approved):

Issue #4976, opened 2023-05-28
Building explosions should flatten trees


BlackYps:
**Describe the feature**
When volatile buildings die they deal damage. Currently trees are pretty unfazed by this. Interestingly they sometimes catch fire, but in general they just keep standing there. It would be pretty cool if the explosion flattened the trees. This would also help people to get a feeling how far the damage radius reaches on different volatile buildings, but this is just a sidenote.

Additional sidenote: It would be even cooler if we spawned a sonic wave that grows to the size of the explosion radius, but this might be complicated to do, because then we probably have to sync the speed of the shockwave with the trees falling and suddenly everything gets 10 times as complicated. So as a baseline I think the trees simply falling is enough.

Garanas:
We're already working on your suggestions 👍 

> Additional sidenote: It would be even cooler if we spawned a sonic wave that grows to the size of the explosion radius, but this might be complicated to do, because then we probably have to sync the speed of the shockwave with the trees falling and suddenly everything gets 10 times as complicated. So as a baseline I think the trees simply falling is enough.

The ACU gate-in animation does this, it isn't as difficult as it looks like but it does require careful tweaking

> When volatile buildings die they deal damage. Currently trees are pretty unfazed by this. Interestingly they sometimes catch fire, but in general they just keep standing there. It would be pretty cool if the explosion flattened the trees. This would also help people to get a feeling how far the damage radius reaches on different volatile buildings, but this is just a sidenote.

I agree, see also: 

- https://github.com/FAForever/fa/pull/4977


There's a lot we can do to improve the death threads (explosion) of units in general.

Garanas:
![image](https://github.com/FAForever/fa/assets/15778155/85bc2b8f-ac9a-494c-b2f4-3ffb49ca18ca)

PR #5033, opened 2023-06-10
Allow death weapons to knock over trees

Garanas:
Closes https://github.com/FAForever/fa/issues/4976

This does not implement the wave as suggested by the issue. It requires quite a bit more work, but this pull request on its own achieves the described effect.

image

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2133,6 +2133,7 @@ AirUnit = ClassUnit(MobileUnit) {
         elseif self.DeathCrashDamage > 0 then -- It was completely absorbed by a shield!
             local deathWep = self.deathWep -- Use a local copy for speed and easy reading
             DamageArea(self, self:GetPosition(), deathWep.DamageRadius, self.DeathCrashDamage, deathWep.DamageType, deathWep.DamageFriendly)
+            DamageArea(self, self:GetPosition(), deathWep.DamageRadius, 1, 'TreeForce', false)
         end
 
         if with == 'Water' then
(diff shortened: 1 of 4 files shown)


BlackYps:
Very nice! Two things I noticed:
1) For higher tech units the death damage is instantly applied once the unit is dead in spite of the animation first playing some small explosions before the big one that acutally spawns the wreck. This is technically not the fault of this PR, but I think it should be changed, so the damage matches the visuals of the animation.

2) It doesn't seem to work for air units. Many air units have aoe crash damage, but the trees don't get knocked over. It's especially apparent with air experimentals. Not sure why this is the case.

Garanas:
> For higher tech units the death damage is instantly applied once the unit is dead in spite of the animation first playing some small explosions before the big one that acutally spawns the wreck. This is technically not the fault of this PR, but I think it should be changed, so the damage matches the visuals of the animation.

We can look into that in future pull requests. This would for example change how a compact volatile formation explodes. Which is fine, but it is better suited as a separate issue and we may need to discuss that with the balance team too

> It doesn't seem to work for air units. Many air units have aoe crash damage, but the trees don't get knocked over. It's especially apparent with air experimentals. Not sure why this is the case.

Air units are special snowflakes. I'll see what I can find

BlackYps:
> We can look into that in future pull requests. This would for example change how a compact volatile formation explodes. Which is fine, but it is better suited as a separate issue and we may need to discuss that with the balance team too

Yeah, that one is definitely out of scope of this pull request

BlackYps (approved):
Looks good now

Issue #5039, opened 2023-06-11
Emerge/Submerge toggle buttons do not work correctly


EvildrewFAF:
![image](https://github.com/FAForever/fa/assets/46637108/15c7a82c-0fd4-48d5-ba4e-e280827901d2)
If you click on the icon while submerging/emerging (double click) you can be confused by the symbol indicating that you can emerge and already emerged sub, same for a submerged sub looking like it is surfaced based on the icon. It can cause confusion....

Either the button should be disabled or not appear temporarily after clicking it during the emerge/submerge process until it is complete.

Garanas:
The UI layer doesn't know when the sequence is complete. We can try and temporarily disable it for a second?

EvildrewFAF:
Emerging/Submerging takes about 10 seconds but that is just a number, might be possible to put it in the OnMotionVertEventChange / OnLayerChange function in the scripts? Could be possible to use the logic behind upgrading your ACU with a TML launcher and removing it to make the icon appear/disappear?

Garanas:
The suggestion works

PR #5041, opened 2023-06-11
Remove the dive button while diving

Garanas:
Closes https://github.com/FAForever/fa/issues/5039

I'm not sure if this is what we should do, but it does work. The dive feature of a unit is temporarily disabled while diving.

https://github.com/FAForever/fa/assets/15778155/8603ffa2-89f1-482c-9e62-b2a3a0798b40

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -2064,7 +2064,25 @@ SubUnit = ClassUnit(MobileUnit) {
         self.Trash:Add(self.SoundEntity)
         Warp(self.SoundEntity, self:GetPosition())
         self.SoundEntity:AttachTo(self,-1)
+
+
     end,
+
+    ---@param self Unit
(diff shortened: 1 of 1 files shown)


EvildrewFAF:
Does the shortcut “d” still change it while the icon is not visible?

Garanas:
> Is it possible to make the sub not cancel move orders when toggling to dive/surface?

You issue an order and that resets the existing command queue. Not sure if it is possible

RabidPope:
> > Is it possible to make the sub not cancel move orders when toggling to dive/surface?
> 
> You issue an order and that resets the existing command queue. Not sure if it is possible

Ok. 
Is there a way to add the order to a queue with shift? So you can click a place and then surface.

Garanas:
@RabidPope 

image

It is already there

PR #5068, opened 2023-06-13
update IsPlattonDelay logic to work correctly in BuilderManager

relent0r:
This PR updates the BuilderManager logic to handle the builder delay logic.

Updates the IsPlattonBuildDelayed to no longer update the brain table for the delay. This was causing any builder that passed through the builder manager to update.

It shifts the delay update to the builder return logic statement. This means that only the builder that is selected will be updated.

Testing 
In order to validate this logic you will need to 
add logs to line 299 that will tell you when a builder fails the delay check.
add logs to line 226 and 234 to log when the delay is added.

The delay is triggered when a builder gets triggered multiple times. The goal of the delay is to stop a situation where an engineer gets a job and starts moving to the builder location, then another engineer gets the same job before the first could start the construction. This trys to avoid situations where multiple builders or multiple instances of the builder will trigger because the conditions are right for the seconds it takes for the engineer to move to the build location.

in draft so I can quickly test this tonight.

As a side note. There are only 3 AI using this functionality at the moment(incase we want to consider fixing the typo in the name).

### DIFF
--- a/lua/sim/BuilderManager.lua
+++ b/lua/sim/BuilderManager.lua
@@ -219,22 +219,23 @@ BuilderManager = ClassSimple {
             end
         end
 
-        -- only one candidate
+        -- only one builder found
+        local candidate
         if candidateNext == 2 then
-            if candidates[1].DelayEqualBuildPlattons then
-                local delay = candidates[1].DelayEqualBuildPlattons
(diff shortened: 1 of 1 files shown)


Garanas:
Great find @relent0r , it now appears to work as usual again

image

PR #5075, opened 2023-06-14
Buff Amphib Torps

### LINKED ISSUE #5012: Balance idea: Amphibious Land Unit Torp Change
## Intended changes

```
30 dps for bricks
7.7 -> 15 dps Wagner
15 -> 30 dps for Othuum
15.4 -> 10 seconds for torp defense fire cycle
```

## Thoughts on the changes 

Changes based on [this](https://github.com/FAForever/fa/pull/4935) Pr by [Basilisk3](https://github.com/Basilisk3). We went a bit farther as to make these units more viable underwater. 

MrRowey:
Close #5012


Spikey84 (approved):

Spikey84:
I don't think the exact method was discussed. Damage makes the most sense to me.

Issue #5045, opened 2023-06-12
Balance idea: Torp Bombers Adjustments 


MrRowey:
## Intended changes
All of the Tech 2 Torpedo Bombers to have the same number of projectiles.

```
- Set all projectile count to 3
```

This is just a small adjustment to the torps since the addition of  #5000 

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

PR #5082, opened 2023-06-15
Set Salvo Size to 3 across the board

MrRowey:
#5045


Garanas:
By increasing the salvo size you also increase the total damage output. You'll need to tweak the damage values too, unless that is intended

MrRowey:
> By increasing the salvo size you also increase the total damage output. You'll need to tweak the damage values too, unless that is intended

oh, interesting I just thought that it would split the damage across the 3 so damage is the value of each torp then?

Garanas:
Yes, it should be. You can test it in-game

Garanas:
@The-Balthazar can renaming the label of the weapon cause issues for mods?

The-Balthazar:
It would break any mod that's trying to hook the units weapon script directly, or that's trying to refer to the weapons by label. Modifying them with a bp merge would be unaffected since that would use table index. While I think it's unlikely that a mod would do those things, it's not outside of the realms of possiblity, and I see no actual benefit in changing the labels.

Hdt80bro:
@The-Balthazar have you noticed a difference with changing `EffectiveRadius`? The engine uses the value somehow, but I'm not sure where. Just noticed that this PR changes that.

PR #5084, opened 2023-06-15
Add new TransportRequested flag for 3rd party AI compatiblity support.

relent0r:
This PR makes the following changes.
Updates the new transport logic to leverage a flag called TransportRequested when more transports are needed.
Previously it used the default NeedTransports flag but this has been found to cause issues with 3rd party AI that are still using the old transport logic due to a callback updating the flag while the old transport code is in a time wait loop.
I've updated all the corresponding functions and build conditions to match and marked the originals as deprecated.

It also fixes a small issue with the FindNearestBaseIdentifier in the base-ai class so that the position search will still work if no factory manager is available (this was already corrected in the easy-ai but not brought across).

It also makes a small update to the TechMain and TurtleMain AI base template to use the new initial bo function. This is to avoid resource stalls early game. The original Default Initial ACU builder group has been renamed to 'Easy Initial ACU Builders' and it used for the Easy and Normal AI.

Testing requirements.
In order to test the transport change. A game can be run with a 3rd party AI that uses the default transport logic. It will no longer error while units are waiting for transports that are being built at factories.

### DIFF
--- a/lua/AI/AIBaseTemplates/ChallengeMain.lua
+++ b/lua/AI/AIBaseTemplates/ChallengeMain.lua
@@ -35,7 +35,7 @@ BaseBuilderTemplate {
         'ExtractorUpgrades',
 
         -- ACU Builders
-        'Default Initial ACU Builders',
+        'Easy Initial ACU Builders',
         'ACUBuilders',
         'ACUUpgrades',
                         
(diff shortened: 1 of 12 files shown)


relent0r:
I just realised there is still a logic gap here for supporting 3rd party AI.

relent0r:
I'm going to approach the other changes in a different PR as this resolves errors in the current develop branch.

PR #5085, opened 2023-06-16
Add notify messages for the new enhancements

Hdt80bro:
Adds default notify messages for the Advanced Quantum Disruptor and Cybran Nano-Repair System. In order to make them work with the existing preferences file, it is merged with the default table to fill in missing keys (rather than it being either all default or all from the preferences file).


MrRowey:
Will this make it easier now and work when adding more stuff to notify if they are set in the defaultmessages ?

Hdt80bro:
Yes - in fact, given how notify gets its messages, it's necessary to change to add new messages.

lua/ui/notify/customiser.lua
```diff
@@ -35,7 +35,19 @@ local notifyOptions = LazyVar.Create()
 
 function InitMessages()
     local prefsMessages = UTF.UnescapeTable(Prefs.GetFromCurrentProfile(NotifyMessages))
-    if prefsMessages and not table.empty(prefsMessages) then
+    if not table.empty(prefsMessages) then
```
Garanas, line 38: this is technically not equivalent. You'll still want to check if `prefsMessages` itself is not a falsy value (nil)

Hdt80bro: `table.empty` returns false when the value isn't a table, so a nil check is redundant.

Garanas: Where does it do that? I thought we were replacing `table.empty` with an assembly version (on faf develop)

Hdt80bro: It still works like that for `table.empty` even if it doesn't for `table.size` now - but plenty of code rely on both behaving that way anyway for us to be able to change that.

speed2CZ: > `table.empty` returns false when the value isn't a table, so a nil check is redundant.

This will break the condition, as it will run the `if` condition with non existent table.

Hdt80bro: I tested it a second time and it appears you are correct. The PR in fafdevelop does, in fact, change the behavior of the function. This is what should really be fixed instead as many other uses of the function rely on the behavior I described, but in the interest of getting this PR through, I'll change it here.

MrRowey (approved):

PR #5087, opened 2023-06-16
Fix normal AI key on index selection. Remove unused entries.

relent0r:
Pre-Patch release validation.
This PR fixes the normal AI loading as the key was set incorrectly for the brain index load. 

It also removes the medium and easy aix variants from the index list as they dont exist to avoid any future confusion.

The easy AI still needs to be fixed before this PR goes in.

### DIFF
--- a/lua/aibrains/index.lua
+++ b/lua/aibrains/index.lua
@@ -6,16 +6,16 @@ keyToBrain = {
     default = import("/lua/aibrains/base-ai.lua").AIBrain,
     campaign = import("/lua/aibrains/campaign-ai.lua").AIBrain,
 
-    -- base AIs
+    -- base AI
     easy = import("/lua/aibrains/easy-ai.lua").AIBrain,
-
     tech = import("/lua/aibrains/base-ai.lua").AIBrain,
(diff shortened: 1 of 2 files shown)


relent0r:
@Garanas 
Are you able to take a look at the easy AI. I tried just setting the brain file back to base-ai.lua but the easy AI was still failing with the error

warning: ...elent0r\fa\lua\aibrains\templates\base\easy-main.lua(1): access to nonexistent global variable AIBaseTemplate

I'm unsure how you are initializing this base template when the easy-ai.lua brain is not being used.

Garanas:
@relent0r Did you remove the other `easy` entry in the table?

relent0r:
> @relent0r Did you remove the other `easy` entry in the table?

Yes, the template load appears to be happening here in the siminit

        -- import base templates, builder group templates and builder templates
        for k,file in DiskFindFiles('/lua/aibrains/templates/', '*.lua') do
            import(file)
        end
        
      Its expecting an AIBaseTemplate class that doesn't exist. I can comment out that for loop to make it work but I never really delved into what your doing with the easy AI in regards to this functionality.

Garanas:
Ah, I see now. I'll look into it

PR #5088, opened 2023-06-17
Quicken inevitable voting results

Hdt80bro:
Currently, the recall vote only finishes early when all team members have voted, even if the outcome is guaranteed to succeed or fail. This PR changes that to close the vote as soon as it gains enough support or no votes. Also renames some of the internals from `Accept`, `Veto` to `Yes`, `No` and should handle teammate deaths during the vote more gracefully.


lua/sim/Recall.lua
```diff
@@ -306,6 +307,9 @@ function SetRecallVote(data)
                 end
                 return
             end
+            if ally.RecallVote then
+                yesVotes = yesVotes + 1
```
Garanas, line 311: Why does it add one at two places?

Hdt80bro: This one adds up all the votes from allies, the one in the check is for the potential vote.

Issue #5006, opened 2023-06-02
Shields do not protect from ACU explosions


Basilisk3:
**Describe the bug**
ACU explosions can destroy units through shields, even though they only do 2500 damage. A low damage explosion like this should not be able to kill units that are protected by shields that sometimes offer over 20000hp of protection. 

This is also very undesirable gameplay wise, as it enables players to teleport into clumps of drones or tech 3 engineers around high value targets, and destroy them all at the same time. 

**There is no counterplay to this, it is impossible to prevent. This is highly abusable and very easy to do.**

**Expected behavior**
ACU explosions should be blocked by shields, if the shield bubble has enough hp left.

**Additional context**
Inspired by a recent thread on the FAF forums.

My guess is, that this is because ACU explosions are handled similarly to those of stationary nukes internally.

Garanas:
I tried to look it up but I can't find how these explosions are made at the moment. Your explanation sounds legit.

This is a balance issue, therefore @Spikey84

ComradeStryker:
I've also noticed this. 
When a player disconnects (in TMM), even if they are under a shield, everything around them blows up with them. 
Radars, engineers, units, etc.

I believe this is because all projectiles - and consequently - their damage and explosions go through allied shields without nullification since they are not an enemy damage source.

Tagada14:
Would love to have this fixed if anyone has the know-how to pick this up

Garanas:
This is possible 😄 . The problem is that the 'average nuke' doesn't deal normal area damage but it deals 'ring damage'. And ring damage ignores shields. If we make the nuke deal normal area damage (and make those overlap) then it can take into account shields.

https://github.com/FAForever/fa/assets/15778155/329c82b2-a911-474f-b155-22049ce95ee6

![image](https://github.com/FAForever/fa/assets/15778155/6fd65853-6a86-450c-aa4e-8e8cc0d43c2a)

Garanas:
A Sanctuary 😃 

![image](https://github.com/FAForever/fa/assets/15778155/958a67c9-fd02-44f6-9b88-2b3756b5bada)

ComradeStryker:
That's actually cool.

Basilisk3:
**Amazing.** This is a big deal, and makes much more sense as well.

Tagada14:
Will this effect normal nukes? They should still ignore shields.

Garanas:
No, they'll do what they do now

lL1l1:
[...]

PR #5091, opened 2023-06-17
Give engineers and drones resistance against ACU explosions 

Basilisk3:
**1. Adds the ArmorType “Engineer”**

**2. All engineers**
ArmorType: “Normal” --> “Engineer”

- ACU explosions no longer lead to the crippling loss of all build power in the base.
- Especially relevant in light of the recent connection issues. Gives the team which loses a player to a DC more of a fighting chance.
- Also related to (and partially fixes) #5006

This commit tackles all engineers. I will make another commit for drones, which have the same problem and will have a damage resistance of 100%.

**Edit:** Added the changes to drones here to prevent conflicts.

### DIFF
--- a/lua/armordefinition.lua
+++ b/lua/armordefinition.lua
@@ -33,6 +33,7 @@
 ---| "ASF"
 ---| "Commander"
 ---| "Default"
+---| "Engineer"
 ---| "Experimental"
 ---| "ExperimentalStructure"
 ---| "Light"
@@ -108,5 +109,13 @@ armordefinition = {
         -- Armor Definition
         'Normal 1.0',
         'TacticalMissile 0.55',
+    },
+    {
+        -- Armor Type Name
+        'Engineer',
+
+        -- Armor Definition
+        'Normal 1.0',
+        'Deathnuke 0.032',
     },
 }
--- a/lua/armordefinition.lua
+++ b/lua/armordefinition.lua
@@ -33,6 +33,7 @@
 ---| "ASF"
 ---| "Commander"
 ---| "Default"
+---| "Drone"
 ---| "Engineer"
 ---| "Experimental"
 ---| "ExperimentalStructure"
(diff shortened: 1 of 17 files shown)


MrRowey:
i don't see this getting implemented but I will mention it in the meeting

Hdt80bro:
There are better ways to do this than with armor types - would also allow for this to be a game option.

Basilisk3:
What other way would you propose? I would be happy to try it out, I am going to have another look tomorrow. Having a game option just for this seems like overkill, though.

I do believe this could possibly salvage a lot of games from being rehosts. It also alleviates the issue of buildpower in the form of engies dying through shields.

Hdt80bro:
This means that enemy engineers also don't take any damage - better to intercept the damage and only apply it when needed. As this only makes sense with partial & full share, it needs to be disabled in other victory conditions - another reason armor types won't work.

Though I'm still not sure why engineers need this - the death nuke is part of the game; if you parked your ACU in your base and it got sniped, blowing up everything around it is exactly what should happen.

Solutions to disconnects should be handled separately (e.g. making all units transfer without death weapons), not lumped in with full share.

Basilisk3:
Enemy engies not dying is part of the reason this commit exists. It is so prevent another ACU from teleporting into heavily shielded bases and destroying all build power. So intercepting the damage and applying it when needed, is not required.

The solution you propose is highly abusable and because instead of dying to another player and losing all engineers, a player could just pull the plug on his internet connection to prevent them from dying. I do not think there should be exceptions, either they all live, or they all get destroyed. This commit more or less just puts them into the same category as buildings, which also take almost no damage.

At the very least, drones should be immune to it.

Hdt80bro:
I think we have a different understanding of the game and that's fine.

MrRowey:
I'm closing this PR as it, not something the Balance team wants to do

VaelophisNyx:
damn this looked pretty cool. Shame the balance tyrants won't let it through.

PR #5093, opened 2023-06-18
Add optimized PBR terrain shader - Terrain101

BlackYps:
There is actually a lot in here and the diff is horrible to read, so I will explain what I did here.
First I had to get rid of the atlas bleeding that we identified. It was caused by the textures for different layers being in the same image. When we sample a high mipmap level pretty close to the edge, then we get bleeding, because the texture sampler linearly interpolates adjacent pixels. So now we only sample the area inside of the blue lines and repeat the texture to achieve that. This "wastes" some texture space and requires us to double the image size, but we get rid of the bleeding and as it is only this singular texture where we need to apply all these atlas sampling shenanigans, I think we can afford it. It makes the creation of this atlas texture quite tedious, so a future map editor should definitely support generating this atlas map automatically. Luckily, it should be rather easy to do in code.
Screenshot 2023-06-19 221759
Next up we have a lot of optimizations and small bugfixes where I accidentally did some incorrect texture sampling.
We also have some refactoring to pass the memory constraints of the ps_2_a shader model. We need to use this instead of ps_3_0 as we lose the distance fog if we use the latter. So the game theoretically supports ps_3_0, but it will forever be the forbidden fruit, as we have no control over the distance fog calculations that seem to happen entirely in the engine.
The most notable optimization is that we save some of the texture scales in the vertex shader as each scale is a vector of four floats, but there is only one float of interesting information in it.
I made the normals more prominent where layers are partially blended as low opaacity albedo maps seemed more prominent than low opacity normal maps. The power of the blending is just eyeballed and there is no particular reason that I used 0.5, but it seems like an improvement.
Finally, there is a new shader variant, called Terrain101, that is basically the full PBR implementation, but without the additional sampling of textures at a 30° angle. I haven't tested the performance yet, but it should run a lot faster, potentially in the same ballpark as all the other shaders.


effects/terrain.fx
```diff
@@ -599,6 +601,11 @@ VS_OUTPUT TerrainVS( position_t p : POSITION0, uniform bool shadowed)
 {
     VS_OUTPUT result;
 
+    // We do this to be able to save some memory registers in the pixel shader as the Tiles only have
```
Garanas, line 604: I think the comment should be at the struct definition

PR #5100, opened 2023-06-19
Economy Structures HP changes

Ruan7053:
#5098

PR #5102, opened 2023-06-19
Revert "Economy Structures HP changes"

MrRowey:
Reverts FAForever/fa#5100


Garanas (approved):
Wrong target branch 😄

Issue #5097, opened 2023-06-18
Balance idea: Missile Speed


MrRowey:
## Intended changes
Increasing the Speed of the Missile of Mobile Missile Launchers (MML) & Spearhead

```
- 1.5x Speed Increase
- Lower the Acceleration
- Test that they can still hit targets
```

_Various thoughts on the reasoning of these changes_

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

Term-AZ:
ill get this one

Term-AZ:
> 

yeah will do

MrRowey:
@Term-AZ just realised that you edited the projectiles them self not the weapon on the units will revert the merged changes and could you re-submit your PR

BlackYps:
The unit BP doesn't have these variables. It's all in the projectile files

MrRowey:
Maybne the 1.5x speed was too slow as it didn't see any difference maybe it needs to be 1.75

PR #5120, opened 2023-06-21
Proof of concept missile speed increase

BlackYps:
For #5097 
This sets the missile speed to 20 (from 12) and acceleration to 9 (from 3). I only made a proof of concept, so this only works for the UEF T2 MML. The turn rate is only set once in the beginning. I think with this approach we don't even need a thread, but I am no expert.
@Garanas mentioned that we could adjust the turn rate of the missiles when they acquire the aeon TMD flare as a target. We should do that, because with the fairly low turn rate that the missiles have now the can't really follow the flare.
Comparison of the missile trajectories:
old
Screenshot 2023-06-21 020635
new
Screenshot 2023-06-21 020308

I included the changes to the launcher to increase the missile fire rate for easier testing. This is not actually a proposed change.


Garanas:
The arc is quite a bit lower. Can it still hit units that are on a (high) plateau in comparison to the mobile missile launcher? Old (gpg-era) maps usually do not have such cliffs, but a lot of custom maps do have excessive cliffs.

BlackYps:
I didn't test this on any maps. This is the highest we can get when we use a constant turn rate. It's more to explore the range of possibilities we have. We could adjust the turnrates during flight, to make it more like it was before, where the missiles would climb quite high and then come down at a steeper angle.
I didn't want to take away from @Term-AZ too much, I just figured that I might as well share what I have found out about the missiles so far.

MrRowey:
Closed in favor of #5137

Issue #5052, opened 2023-06-12
Balance idea: Vision Range Buff


MrRowey:
This will be a big undertaking this will be to increase the vision range of all Mobile units with some exceptions.

Unit Exceptions
```
- Scouts
- Carriers & Atlantis
- T1 Arty
```

Changes
```
- Increase the vision by +15% (Rounded Up)
```

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

PR #5121, opened 2023-06-21
Global vision range buff

Blodir:
Completes https://github.com/FAForever/fa/issues/5052

### DIFF
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -762,6 +762,12 @@ function PreModBlueprints(all_bps)
             }
         end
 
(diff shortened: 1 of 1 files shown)


MrRowey (approved):

lua/system/Blueprints.lua
```diff
@@ -762,6 +762,12 @@ function PreModBlueprints(all_bps)
             }
         end
 
+        local bpIsAirScout =  bp.CategoriesHash.SCOUT and bp.CategoriesHash.AIR
```
Garanas, line 765: it is better to add this to `blueprints-unit.lua`, see also:

- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/system/blueprints-units.lua

PR #5122, opened 2023-06-21
Move visioning change to unit Bp File

MrRowey:


Hdt80bro:
It'd be better to put this in the actual unit blueprints I think

Garanas:
Yes, but it is a lot of work to do that 😄 !

What would be ideal is to start using that blueprint read / write tool that @The-Balthazar wrote, so that we can start migrating a lot of this to just into the blueprints

MrRowey:
@Garanas & @Hdt80bro are we merging this in or not ?

Hdt80bro:
The code is already there - this PR just moves it around. We can do more blueprint work in the future.

MrRowey:
ill close it for now

PR #5125, opened 2023-06-22
Right click appends to selection set 

clyfordv:
Right clicking a control group avatar appends currently selected units to the group.


4z0t:
Use modifiers check instead of keycode to a magical number

PR #5128, opened 2023-06-23
Torrent Change

### LINKED ISSUE #4983: Balance idea: Torrent Rework
## Intended changes

```
- 7.5k hp
- 4k mass
- 40k E
- Increase salvo size x2
 - Make less accurate (exact number up for discussion)
- 200 damage
-  3 AOE
- Firing pattern similar to Scathis
```

## Thoughts on the changes

This change totally reworks the role of the Torrent from a expensive battleship level unit to a cheap AOE missile ship. This should make Torrents viable in more situations. 

Spikey84:
Attempt 3 of this PR.

Closes #4983.

@Garanas @The-Balthazar I am struggling to make the weapon have randomness. Any help would be appreciated. We want an effect similar to the scathis.


The-Balthazar:
Firing randomness is effectively meaningless with a tracking projectile.

Garanas:
Yes, missiles are fixed to the center of the unit here:

- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Projectile.lua#L132-L136

PR #5149, opened 2023-06-25
First attempt at fixing the save-load sequence

Garanas:


Garanas:
The cause is the navigational mesh

PR #5151, opened 2023-06-25
Update XRB2308_unit.bp

MrRowey:


units/XRB2308/XRB2308_unit.bp
```diff
@@ -255,14 +254,8 @@ UnitBlueprint {
             RangeCategory = 'UWRC_AntiNavy',
             RateOfFire = 0.25,
             TargetPriorities = {
-                'SUBMERSIBLE',
```
Garanas, line 258: These priorities are quite different. Is this intended?

MrRowey: I Copied over what was on the actual unit that dose the shooting. this one is just the one that is used on top of the water when being built I was just equalising the files

Garanas (approved):

PR #5153, opened 2023-06-25
Fix for mods that replace default platoon template files

relent0r:
This PR reverts platoon templates back to where they were prior to the base ai work being started. It then created new template files for any that have been changed and move the new template into them.

This should resolve issues where mods replace existing platoon templates.

To Test : 
Run a game of nomads with AI added. It AI should now start building its base normally.

Issues noticed with nomads specifically.
There is a duplicate of the template called `T2MassHuntersCategory`
The platoon template for T4ExperimentalSea in nomads is invalid and it will overwrite the default which will break the platoon former once one gets built.

### DIFF
--- a/lua/AI/PlatoonTemplates/AirPlatoonTemplates.lua
+++ b/lua/AI/PlatoonTemplates/AirPlatoonTemplates.lua
@@ -7,6 +7,7 @@
 --**  Copyright © 2005 Gas Powered Games, Inc.  All rights reserved.
 --****************************************************************************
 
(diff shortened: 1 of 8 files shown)


relent0r:
ok no I just tested it and there is still a problem related to it trying to get a buildingTypes error during the initial bo.

relent0r:
Have added some more logic to the initial build order that will fall back to the default basetemplates if the factionindex is > 4 so that new factions will try to use the build efficient one.

PR #5162, opened 2023-06-29
Fix shield visibility on land maps

BlackYps:
I tested it on funeral plains and it works as expected.


BlackYps:
I guess this should actually be included in a hotfix, but there seems to be no dedicated milestone for that

Garanas:
Yes, we'll gather them all up and try to do a hotfix this weekend

PR #5166, opened 2023-07-01
Reintroduce upper albedo scale as shader switch

BlackYps:
Reintroduces the scale of the upper albedo as a switch for which shader to use. To my understanding the AlbedoTile variables basically store how often the texture is supposed to show up on the map. If we check for < 1.0 that means the texture would not even fit on the map. Hopefully nobody set his texture scales to these nonsensical values, so we don't break any existing maps. I am pretty confident in this, because in the ozonex editor you would have to set the scale to greater than 256 and the editor doesn't allow this.
This different value in the editor is because the scale in the editor is basically "how big should your texture be in ogrids"

I am not sure if I agree with the addition of the Terrain001NormalsPS. Yes, it makes the layer masks more consistent, but it prevents us from using this shader as a drop-in replacement for existing maps.
Depending on our goals we should maybe offer both variants? I.e. one that uses Terrain001NormalsPS and one that uses TerrainNormalsXP.

### DIFF
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -395,13 +395,12 @@ float3 ApplyWaterColor( float depth, float3  inColor)
     return lerp( inColor.xyz, wcolor.xyz, wcolor.w );
 }
 
-float3 ApplyWaterColorExponentially(VS_OUTPUT inV, float waterDepth, float3 color) {
+float3 ApplyWaterColorExponentially(float3 viewDirection, float waterDepth, float3 color) {
     float4 waterColor = tex1D(WaterRampSampler, waterDepth);
-    float3 v = normalize(-inV.mViewDirection);
     float3 up = float3(0,1,0);
     // To simplify, we assume that the light enters vertically into the water,
     // this is the length that the light travels underwater back to the camera
-    float oneOverCosV = 1 / max(dot(up, v), 0.0001);
+    float oneOverCosV = 1 / max(dot(up, normalize(viewDirection)), 0.0001);
     // light gets absorbed exponentially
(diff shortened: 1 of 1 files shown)


Garanas:
Could you turn the switch into a separate, well-named function that returns a boolean? It would make it all easier to read when you go through the code

effects/terrain.fx
```diff
@@ -1284,7 +1308,14 @@ float4 DecalsPS( VS_OUTPUT inV, uniform bool inShadows) : COLOR
 
     float waterDepth = tex2Dproj(UtilitySamplerC, inV.mTexWT * TerrainScale).g;
 
-    float3 color = CalculateLighting(normal, inV.mTexWT.xyz, decalAlbedo.xyz, decalSpec.r, waterDepth, inV.mShadow, inShadows).xyz;
+    float3 color;
+    if (IsExperimentalShader()) {
+        float roughness = 1 - decalSpec.r;
+        color = PBR(inV, decalAlbedo, normal, roughness);
```
Garanas, line 1314: This is tricky - it enforces the PBR computations 🤔

BlackYps: True.
As far as I know and have tested all decals seem to have a specular amount of 0, which makes the PBR shader basically identical to the calculateLighting one. So it shouldn't cause issues, but then it begs the question if we shouldn't just leave DecalsPS untouched and switch to exponential water coloration in calculateLighting.

BlackYps: actually now that I think about it, I think this is what we should do

PR #5169, opened 2023-07-01
Format blueprints

Garanas:


BlackYps:
I get that we like to omit descriptions on trivial pull requests, but can we get a short description on ones that change thousands of lines?

Garanas:
Oh, forgot I only discussed it in the balance team channel.

We're using a tool from @The-Balthazar that enables us to rebuild or reading and then writing out the blueprints. This involves a certain formatting, that is what we did with this pull request. Now we can use the same process to start removing blueprint editing we do on the fly (like in `blueprints.lua` and the like) and start writing out the values to the blueprints themselves. This has a few advantages:

- Tools that rely on the blueprints are more accurate
- It becomes easier to see the real values for a developer
- Less unexpected side effects
- The changes we make during the blueprint loading enforce values that may break or cause unintended side effects for mods

This pull request only includes the formatting changes. Future pull requests will start pulling calculations from the various `blueprints.lua` files and into the actual blueprints

maudlin27:
I'm guessing the errors I have been having are linked to this - this causes M28 (and I presume M27 although I've not yet tested it) to crash the game as it looks like it is removing the OVERLAYANTINAVY category from units.  For example, the fatboy used to have this category and now doesnt.  When M28 references this category, it causes an error because it is referencing a nil value.

PR #5170, opened 2023-07-01
Fixed Advanced Regen Aura

### LINKED ISSUE #4439: Incorrect `RegenPerSecond` values for Restoration Field enhancements
**Describe the bug**
The Steam version has the values for the Seraphim commander enhancements
```lua
        RegenAura.RegenPerSecond = 0.0055555555
AdvancedRegenAura.RegenPerSecond = 0.01111111111
```
Sometime in FAF's prehistory, it got changed to
```lua
        RegenAura.RegenPerSecond = 0.02
AdvancedRegenAura.RegenPerSecond = 0.01111111111
```
Where the `RegenPerSecond` value of the Restoration Field is higher than that of the Advanced Restoration Field, meaning it's been stronger for the past decade. This was partially mitigated because tech level ceilings were also introduced, but that still means that units that have a certain maximum health based on the floor and ceiling of the enhancement will have more regen with the regular one. In particular,
 * The unit must have low enough maximum health so that the ceiling of the Restoration Field doesn't clamp the regen value to be smaller than regen value calculated with the smaller scalar of the Advanced Restoration Field.
 * The unit must have enough maximum health so that the regen gained from the Restoration Field is still greater than the floor of the Advanced Restoration Field (15 hp/s for all tech levels).

In practice, the actual value of `RegenPerSecond` doesn't usually matter since most units' maximum health multiplied by the `RegenPerSecond` value is almost always higher than the tech level's ceiling, but there are several units where this isn't true (recall that FAF also changed the enhancement from a `RegenPercent` affect type to a `Regen` type while hardcoding the `Mult` value in that case to scale by the unit's maximum health). For example:
 * The Rhino has 1,900 health. 1,900 * 0.02 = 38, much higher than the T2 ceiling of 15. For the advanced version, 1,900 * 0.0111 = 21.111, which is less than 38, but the T2 ceiling of 50 lets it actually be that value, so the advanced version is stronger in this case.
 * The Suthanus has 925 health. 925 * 0.02 = 18.5 which is less than the T3 ceiling of 25. For the advanced version, 925 * 0.0111 = 10.2777, but the floor is 15. Thus, it has just enough health that the floor of the Advanced Restoration Field don't raise the final result above the Restoration Field, and not enough health to make the ceilings matter.

Note that the restoration fields also increase the maximum health of the unit (by x1.1 for regular and x1.2 for advanced), but the calculation doesn't use this result (if it did, the Advanced Restoration Field *still* wouldn't be stronger, as 0.0111 * 1.2 = 0.01333 < 0.022 = 0.02 * 1.1).

**To Reproduce**
Take a Seraphim commander and give it the Restoration Field.
Now take, say, the Seraphim T3 Mobile Heavy Artillery Suthanus (it meets the criteria above). Should look like this:
image
Upgrade the commander to the Advanced Restoration Field. Should look like this now:
image
Note the decrease from +18/s to +15/s.

**Expected behavior**
*The Advanced Restoration Field should always be stronger than the Restoration Field*

MrRowey:
Closes #4439

Small fix to the Seraphim ACU Advance Regen Aura. it has a lower regen amount compared to the base one so this just doubles the regen from the base one so 
`Base RegenAura = 0.02`
`Advanced Regen Aura = 0.04`


Garanas:
What does the value represent, % of health regenerated per second?

Tagada14:
AFAIK yes

MrRowey:
> What does the value represent, % of health regenerated per second?

See the closing tag hdt dose a whole explanation 😄

Garanas:
What are the values of various units with this in practice?

MrRowey:
> What are the values of various units with this in practice?

Base RegenAura
image
image
image

New Advanced Regen Aura
image
image
image

Old Advanced Regen
image
image
image

@Garanas

MrRowey:
This has been Increases by 260% the same as the BaseRegen Was done in FAF At Some Point 
BaseRegen was `0.0055555555` > `0.02` = +260%
AdvRegen is now  `0.01111111111` > `0.04` = +260%

MrRowey:
Dropped the AdvRegen down to 0.35 so it's still powerful but not over the top with the t3 unit being over 100 regen

Hdt80bro:
In the buff code, `Regen.Mult` is hijacked to be taken as a percentage of the target unit's health. It's something I plan to fix.

Hdt80bro (approved):

PR #5171, opened 2023-07-02
Lod1 fixes aeon

MadMaxFAF:
lod 1 fixes for;

uaa0203
ual0001
xal0305
ual0401
dalk003


MadMaxFAF:
this is ready just waiting for wiki gen to run

Garanas:
It has run 👍 , but only on fafbeta at the moment. I'll pull in the change shortly

PR #5172, opened 2023-07-02
Enable switching between linear and exponential water absorption

BlackYps:
This enables switching between linear and exponential water absorption based on the value of lightMultiplier.
 
The upcoming terrain shaders will use exponential water absorption. The challenge now is to use the right absorption in the mesh shader. Originally we thought about having a map script change the unit shaders accordingly, but this gets messy because we have a lot of different shader techniques that would need to be changed and there are also prop and wreckage shaders. All of these would need a second variant, so we can switch them out. All of this only became necessary because the mesh shader doesn't know which terrain shader is being used.
Instead I came up with a different idea. There is in fact another variable that the map author can use to relay this information. The lightMultiplier. Its existence is not strictly necessary, because its value could just internally be multiplied with the sun and ambient color before saving the map file. Both the gpg and ozonex editor only allow setting this value in the 0 to 2 range. But as it is a float it could have any other value. If we set this value to e.g. 2.5 when using a terrain shader that uses exponential water absorption, then we check if this value is above the threshold in the mesh shader and switch our water color calculations accordingly. The elegance of this is that we don't even have to touch the lighting calculations, because 2.5 is still a reasonable enough value to work with. Our future map visualizer/editor would make sure to internally divide the values that get set for the sun and ambient color by 2.5 so the user doesn't even have to worry about this.

### DIFF
--- a/effects/mesh.fx
+++ b/effects/mesh.fx
@@ -304,6 +304,7 @@ struct VERTEXNORMAL_VERTEX
     float4 position : POSITION0;
     float3 normal : TEXCOORD3;
     float4 texcoord0 : TEXCOORD0;
+    float3 viewDirection : TEXCOORD5;
     float4 shadow : TEXCOORD2;
     float4 color : COLOR0;
     float4 material : TEXCOORD1;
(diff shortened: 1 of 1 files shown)


BlackYps:
The diff is an absolute mess. I don't know why it does that. It seems like git doesn't handle rearranging of code well at all :(

Sheikah45:
Just chiming in to say that magic values can be very annoying and dangerous as code ages. It might be convenient know but can really turn into a headache.

BlackYps:
I am aware, but I don't see an alternative in this case.

Sheikah45:
You mentioned using a map script or some flag in the map file to select the other shaders which I think would be preferable even if it is more work initially.

My reason being that it also uncoupled the techniques so you can adjust them independently if needed. And the selection method is more obvious.

I mostly say this from my experience with mapgen and trying to minimize code by using magic numbers causing a lot of headaches down the road. And that was in an environment where backwards compatibility was less strict too.

BlackYps:
hm, good input. I'll have to think about how to do it without producing enormous code duplication.

BlackYps:
What we could do is to just use a boolean value to control the water rendering. Then we would need one technique for the linear variant and one for the exponential variant. Four factions times four shader techniques (normal, building, stunned and personal shield) is 16, plus wreckage and three prop shaders is 20. Times three, because we need high, medium and low fidelity versions of each shader is 60 techniques at minimum that we would add to the file and that doesn't even include nomads and TA shaders. This translates to roughly 1500 lines of code.
It's doable, but also a lot of boilerplate to get rid of a single magic value. What do you think @Garanas ?

BlackYps:
git has tools to improve the diff. This one improves the situation a lot `git diff d6aa368a a115c6c6 --patience --color-moved`. On top of that I recommend to follow the changes commit by commit, then it should be easier to follow what happened

BlackYps:
I noticed that intellij produces much nicer diffs by default than VS Code does when viewing commits

PR #5200, opened 2023-07-07
Projectiles Part 3

MrRowey:


projectiles/CAAMissileNanite03/CAAMissileNanite03_script.lua
```diff
@@ -17,6 +21,9 @@ CAAMissileNanite03 = ClassProjectile(CAAMissileNaniteProjectile) {
         self:ChangeZigZagFrequency(2)
     end,
 
+    ---@param self CAAMissileNanite03
+    ---@param TargetType string
+    ---@param TargetEntity Unit
```
Garanas, line 26: can also be nil

MrRowey: image

PR #5212, opened 2023-07-10
T2 Bomber speed nerf

Basilisk3:
**T2 Bombers**
MaxAirspeed = 15 --> 14.25

T2 Bombers were as fast as their counter (interceptors) which meant that, in some scenarios, interceptors would never be able to catch them. This nerf increases the skip-cap of using T2 Bombers and enables better counter play for the defending player.

Edit: Removed the changes to transports, as similar changes have been merged with #5423.


MrRowey:
Transports and T1 Inits have been massively changed in the beta balance I suggest you check that out first.

neur0Grec0:
I think it's acceptable that T2 Bomber and T2 Transports are faster than interceptors.

Aren't the T2 bombers fighter bombers? If they cost more and are a tech level higher it makes sense that a t2 fighter bomber is faster than a t1 interceptor.
Additionally you look at the mass economy of t1 interceptors vs t2 bombers you can build 8 t1 fighters for the cost of 1 t1 bomber, which will easily defeat the t2 bomber in the air. I'm unsure why t1 interceptors should be buffed here. T1 interceptors are already good enough against t2 bombers and beat them in air combat. This will make them even worse in air combat and a worse trade against interceptors when they engage.
With 8 interceptors you should be able to cover a larger area and catch a t2 bomber in the air, even if the t2 bomber is slightly faster.

There's also other things to take into account, this will make T2 bombers significantly worse around the board. Worse vs AA, worse vs Navy etc.

I also don't see why T2 transports need to be nerfed. T1 artillery drops isn't a broken strategy, just counter with some land units or shoot it down it AA or planes.  T2 transports are quite expensive and you can built at least 6 interceptors for the cost of 1 t2 transport which will shoot down a T2 transport easily, and the transporting units will also loose all the mass being transported.

I don't see why this is a necessary balance, I think it just makes interesting and dynamic plays with transports marginally less viable.

Basilisk3:
> I think it's acceptable that T2 Bombers and T2 Transports are faster than interceptors. Aren't the T2 bombers fighter bombers? If they cost more and are a tech level higher it makes sense that a t2 fighter bomber is faster than a t1 interceptor. 

T2 Bombers and Transports are not faster than interceptors.

> Additionally you look at the mass economy of t1 interceptors vs t2 bombers you can build 8 t1 fighters for the cost of 1 t1 bomber, which will easily defeat the t2 bomber in the air. I'm unsure why t1 interceptors should be buffed here. 

Just looking at the mass cost simplifies the current situation too much for several reasons. For the record, this paragraph mostly applies to Janus.
- T2 Bombers have a very densely packed amount of anti-air DPS (also true for Nothas)
- It is very easy to dump a lot of resources into them and spam them (true for all of them)
- They have very high alpha damage

These points result in them performing well against interceptors, considering they also have a second functionality.

> T1 interceptors are already good enough against t2 bombers and beat them handily in air combat. This will make them even worse in air combat and a worse trade against interceptors when they engage. 

This pull request does not aim to change the interceptor vs T2 Bomber dynamic in air engagements. 

The changes do alter their engagements slightly, but this is not the main point. I am trying to address issues when doing the exact opposite - changing the dynamic between the two unit types when they are **NOT** engaged. Right now, a T2 Bomber can NEVER be caught by an interceptor as soon as it flies past it and is out of weapon range. 

This is not intuitive and reason why T2 Transports already got a speed nerf to be slower than interceptors, in past patches. 

> There's also other things to take into account, this will make T2 bombers significantly worse around the board. Worse vs AA, worse vs Navy etc.

It won't make them significantly worse, for that, the stat changes are too small and the units too strong.

> 1 artillery drops isn't a broken strategy, just counter with some land units or shoot it down it AA or planes. 

I disagree, artillery drops are not particularly hard to pull off but extremely effective. A transport full of Zthuees can kill your entire base in seconds. Considering their strength, increasing the amount of thought you have to put into using them effectively and increasing their skill cap in general, is important for the health of the balance. 

This pull request aims to reward good scouting and counter play, while still punishing bad scouting and awareness.

> T2 transports are quite expensive and you can built at least 6 interceptors for the cost of 1 t2 transport which will shoot down a T2 transport easily, and the transporting units will also loose all the mass being transported.

T2 Transports are really not that expensive and in some scenarios you can even reuse them.

PR #5217, opened 2023-07-13
Drone Order

ComradeStryker:
Grants Shoulder drones the Patrol order, command.


Garanas:
You pointed to `deploy/fafdevelop`, while the base is from `deploy/fafbeta`. If you point it to `deploy/fafbeta` then all these commits will go away 😄

PR #5228, opened 2023-07-14
Re-implementation of spread attack

Garanas:
An initial basic implementation of spread attack. The goal of the implementation is to reduce the number of commands we generate to a minimum. 

Various problems:
- [x] Problems arise when there are less units than orders
- [x] Problems with reclaim orders on props
- [x] Problems with build orders (these are special snowflakes)
- [x] Problems with introducing some redundancy (only pick one move command, but may want multiple attack commands for redundancy

And the usual problems such as keeping a tab on performance, etc.

Requires the following assembly patch by @Strogoo :
- https://github.com/FAForever/FA-Binary-Patches/pull/17
- https://github.com/FAForever/FA-Binary-Patches/pull/22

You can download the last version here of it in the pull request. Based on the performance tab of the task manager I can fairly guarantee that it is at least not a bitcoin miner 😄 

With this pull request we can distribute the following orders:

- Move orders
- Attack orders (on props, units or ground)
- Aggressive move orders
- Assist orders
- Guard orders
- Capture orders 
- Reclaim orders (on props or units)
- Patrol orders
- Sacrifice orders
- Build orders


Garanas:
https://github.com/FAForever/fa/assets/15778155/534d34c2-de41-4aae-8e3d-272e3ed988b8

https://github.com/FAForever/fa/assets/15778155/c986fc93-6c70-4dc7-b0b1-4375bdfc7767

Garanas:
https://github.com/FAForever/fa/assets/15778155/aefb6919-b457-4274-ace1-eb7e26548f45

https://github.com/FAForever/fa/assets/15778155/fbbe3b9c-af8d-449b-be5c-06d2702a7e54

PR #5230, opened 2023-07-15
Map shaders

Garanas:
@BlackYps has been doing a lot of work surrounding shaders the recent months, with absolutely exciting results. Some terrain related work had to be undone right before the patch because we missed an edge case where it could impact existing maps. To understand this I've tried to use the initial shader (`Terrain001`) and made a few tweaks or improvements.

While doing so I now better understand certain choices.

### A switch

We need some 'switch' because the engine does certain computations on the CPU that we can not interfere with. Therefore we need to 'embed' our changes in certain parts of the pipeline. The pipeline consists of:

- (1) Pre-computing the bi cubic normals
- (2) Pre-computing the stratum normals on top of (1) (on the cpu)
- (3) Blending of the stratum normals
- (4) Blending of normal decals on top of (3)
- (5) Blending of the stratum albedo / lighting, using the result of (4)
- (6) Blending of albedo decals on top of (5), using the result of (4)

The steps we control are all but (2). And the second step on its own is interesting, the result of `Stratum0NormalSampler` is this for example:

image

What is interesting is that the normals of the stratum mask are already blended in with the normals of the terrain. This is unusual, and even though I've done a lot of work in [#3 ](https://github.com/FAForever/fa/pull/3585) I didn't notice this crucial missunderstanding on my end 😄 

As a result we **need** the switch to be able to embed our normals into step (1). This also gives us an interesting side effect: using the console command `ren_bicubicnormals` we can 'skip' (1) and use the CPU-based interpolations. As a result we effectively remove the map-wide normals and anything else we may end up doing in (1). This could be a toggle in the game options

### Performance

The game ends up doing a lot of CPU-based processing on the frame buffers. It may explain why computers with high resolution displays perceive a higher CPU load and are therefore more likely to stutter when moving the camera. As an example: I perceive no stutters when I play at a resolution of 800x600, but at 2560x1440 my CPU can not always keep up, the render thread throttles and I perceive stutters.

**not finished writing yet**


Garanas:
@BlackYps we'll need your toggle as described in:
- https://github.com/FAForever/fa/pull/5166

BlackYps:
Nice writeup. The mentioned pull request is almost ready, it just needs your input how we want the normal pass to behave

Garanas:
@BlackYps Here's a map that you can use to test these changes:

[auralian_-_the_core.v0017.zip](https://github.com/FAForever/fa/files/12064581/auralian_-_the_core.v0017.zip)

image

image

image

BlackYps:
How much are the code changes related to what you wrote here? I see you put in the shader switch again, but there is also a lot of commented out code and some other changes. I am not sure which of these are just testing around and which of these are supposed to be merged in at some point

Garanas:
I (temporarily) commented out the PBR shader that you wrote because it increases the reload time with several seconds. I'll uncomment that again when we're done here. Not sure why it takes so much time to load that shader, maybe because it uses pixel shader 3.0 ?

BlackYps:
The reload time increase could very well be because it uses pixel shader 3.0. I refactored that to use shader level 2 again in #5093 anyway. I just haven't pushed the latest changes yet, because reverting some of the shader changes during the patch deployment introduced conflicts. Once #5166 is merged, I should be able to easily rebase.

RabidPope:
Those cliffs look really good!

Garanas:
'Functional' version of the shaders, with the exception of the normals being switched up:

[auralian_-_the_core.v0017.zip](https://github.com/FAForever/fa/files/12373352/auralian_-_the_core.v0017.zip)

PR #5235, opened 2023-07-16
Fixes, and updates for the coop AI's BaseManager

Dhomie:
This pull request is aimed to fix some long present issues, and also add some AI control for TMLs and SMLs.

The following issues were fixed:
- Inconsistent structure upgrading, sometimes AI factories fail to upgrade to the desired tech level.
- BaseManager not decrementing its total Engineer count if the Engineers were either captured, or reclaimed (wrong death callbacks)
- BaseManager not being able to build ACU/sACU enhancements that have any prerequisites.
- Changed the non-existant 'FighterBombers' child type to the proper 'CombatFighters' for the ReactiveAI OpAI type.

I've added a better method for the AI to grab TMLs and SMLs, without any limit, and also added an AI functionality for SMLs.


relent0r:
Hi @Dhomie I'm just starting a review on this. Doing some validation on your changes here and wanting to confirm a few things. Nice work btw, the Campaign AI is tricky to comprehend.

For the CombatFighter changes am I right to assume that you were taking a look at the _save.lua files to validate what was a valid childtype and what wasn't? Specifically the AirAttacks_save.lua file.

That unit destroyed trigger callback is a little odd. Not what you've done but the .Subtracted field check. Based on what I'm reading it won't be set until the unit is destroyed rather than when the callback is set on the unit. Wondering if that means the callback is potentially being created multiple times on those engineers.

I haven't delved into the enhancement changes yet. I've always been puzzled why some things use that SimUnitEnhancements table and other things use a GetEnhancements function for a common utilities lua file. I'll do some testing this weekend. Is there a particular map where the Enhancements will be used more than others?

Dhomie:
Hi @relent0r

I've went over the many _save.lua files actually, and cleaned them up, along with their _EditorFunctions.lua files.
It's actually available as a mod on the vault, I've been using that to test changes and such, however I didn't include them yet, only the things that wouldn't really overwrite, or otherwise mess things up for existing coop missions.
Here's a GitHub link for reference: https://github.com/Dhomie/FAF-Campaign-Experiments

Anyhow, I'll try to explain how the platoons work exactly in a nutshell, so please bear with me on this.

GPG probably tried to standardize things even for SC1, but never quite managed to finish their campaign framework, not even for SCFA. For SC1, their method was to have all the platoons they wanted the AI to build inside the map's save.lua file, which then would be loaded in from ScenarioUtilities.lua, done via `InitializeScenarioArmies()`, specifically `LoadArmyPBMBuilders(strArmy)`.

The names are very important here, because it determines whether to load a platoon builder as a PlatoonBuildManager (PBM for short) one, or an AttackManager (AM for short) one.

The PBM is what the AIs mostly use to actually create platoons that are built, but it's capable of forming platoons that don't require any construction, it can be done near a base, or it can pick any unit from the 'ArmyPool' on the map.

The AM however is a platoon forming thread at its core. In a nutshell, the AM is used to form the final 'Master' platoon, from a bunch of 'Child' PBM platoons. This is where the many _save.lua files in OpAI come in.
The BaseOpAI.lua practically loads in AM platoons the same way SC1 had them loaded in from the map's save.lua, except we can now load them in any time via map script, and with much less coding, again, GPG standardized this.
(This also allows for creating random platoon compositions.)

So, whenever we load in such an AM platoon, we basically load in a bunch of PBM platoons along with it, from the _save.lua files. Then the AI begins to build the PBM platoons, they get a dummy platoon function `DefaultOSBasePatrol(platoon)`, which only does something if the right marker chain exists, otherwise the platoon just waits until the AM platoon can be formed.

The BaseOpAI introduced the child types to actually allow to pick what units (platoon templates) the AI should build to assemble the final platoon. 'CombatFighters' in this case are the T2 Fighter/Bombers, and the T2 Aeon Swift Wind.
The ReactiveAI is yet another attempt by GPG to create a standardized platoon type that would respond to game ender threats by the player (they didn't account for coop of course).

Here's the following execution of it: `BaseInstance:AddReactiveAI('ExperimentalAir', 'AirRetaliation', 'Base_ExperimentalAir')`
This tells the AI to build air units to hunt down any air experimentals the player might have. Currently it builds a mixture of ASFs, and T1 Interceptors, because the 'FighterBomber' child type doesn't exist, it was named like that by GPG, but FAF changed it to 'CombatFighters'. Fixing that means the AI will now build those first instead of Interceptors, if possible.

Long explanation, I know, but you gotta understand the very foundation here to understand what changing something might result in.

As for the unit death triggers, it shouldn't be added multiple times. `BaseManagerEngineerPlatoonSplit(platoon)` is the platoon function for the engineer platoon, it reassigns each one into a new single unit engineer platoon. We don't disband these platoons at all.

As for the enhancements, I've stayed with the _if it works, don't change it_ approach, and used that table. If you have a better suggestion or actual application, I'm all ears.
At the moment none of the existing coop maps really have an ACU build an enhancement with any prerequisites, it's meant to be more of a fix for future missions if a mission maker wants to actually use that feature.

Consider joining the Coop Discord though, we usually share maps that are WIP there for testing.

Dhomie:
I've managed to set up a mission that can be used to test the enhancement changes, along with the SML and TML functions.
It plays out just fine with or without them, so the changes can be compared.
It's a reworked version of Cybran M6 from SC1.

[scca_coop_r06.remastered.zip](https://github.com/FAForever/fa/files/12113447/scca_coop_r06.remastered.zip)

speed2CZ (approved):
It's looking good

Garanas:
Urg, made a mistake 😞

PR #5240, opened 2023-07-18
Implement the assembly patch of 4zot on SetStat

Garanas:
Requires https://github.com/FAForever/FA-Binary-Patches/pull/21 or the game crashes

With thanks to @4z0t the use of `SetStat` is now safe, and returns a boolean that indicates whether it requires to be setup. The volatility of this function prevented us in the past from gathering all sorts of statistics, but now we can.

According to 4z0t the function relies on the use of a Mutex. We're not sure what the impact of that is, but as long as the access is short it shouldn't be too harmful.


engine/User/UserUnit.lua
```diff
@@ -117,11 +117,11 @@ end
...
----@generic T
+--- Retrieves a statistic that is assigned to a unit by the sim call `unit:UpdateUIStatistic`. Returns the default
+--- when the value does not exist
 ---@param name string
----@param defaultVal? T
----@return {Value: T}
+---@param defaultVal? number
+---@return {Value: number}
 function UserUnit:GetStat(name, defaultVal)
 end
```
HUSSAR-mtrela, line 127: add description of UserUnit:UpdateUIStatistic function

HUSSAR-mtrela (changes requested):

engine/User/UserUnit.lua
Garanas: This doesn't rely on an assembly patch to work however, we already use it for veterancy for example

lua/sim/Unit.lua
```diff
@@ -4764,6 +4764,27 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent) {
...
+    ---@param self Unit
+    ---@param key string
+    ---@param value number
+    UpdateStat = function(self, key, value)
+        -- With thanks to 4z0t the `SetStat` function no longer hard-crashes when the value doesn't exist. Instead, it returns 'true' 
+        -- when the stat doesn't exist. If it doesn't exist then we can use `GetStat` to initialize it. This makes no sense, therefore
+        -- we have this new function to hide the magic
+        local needsSetup = cUnit.SetStat(self, key, value)
+        if needsSetup then
+            cUnit.GetStat(self, key, value)
```
4z0t, line 4784: Btw, after GetStat, you have to call SetStat again

PR #5251, opened 2023-07-20
Tighter formation for the Shard

Basilisk3:
The Shard had a very spread out formation for a unit of its size, this change addresses this and enables it to deal more concentrated DPS when used in groups.


Garanas:
This should technically point to `deploy/fafbeta`

PR #5261, opened 2023-07-22
Update for Campaign AI Brain, and ScenarioPlatoonAI

Dhomie:
This pull request aims to add some QoL changes for transport attacks for coop, and also add some additional functions to the Campaign AI Brain so mission scripters could make more use of `aiutilities.lua` and `aiattackutilities.lua` (although I suppose some of those could be just included in `aibrain.lua` to avoid duplicate entries.).
In more detail:

- `LandAssaultWithTransports()` platoon AI functions can now grab transports from a desired platoon. Simply specify `BaseName` in the platoon data. Corresponding functions were also changed to work with this, like `TransportPool()`, that now can assign transports to the desired platoon via the same method. Previously it was just 1 universally shared platoon named `'TransportPool'`, which made the usage of transport attacks from 2 or more bases rather tedious.
This is all **OPTIONAL**, if no `BaseName` is specified, the universal pool is used instead, so this won't mess up any existing coop missions.

- Regarding the above functions, land units that couldn't be loaded by transports in time are currently reassigned to the `'ArmyPool'`, and due to the way the Campaign AI forms/builds platoons, practically these units are left sitting there permanently. These over time amass, and clutter up the base, so we simply self-destruct any units that couldn't be loaded.

- I fixed a slight issue with rally point markers for naval factories, if no `'Naval Rally Point'` marker exists on the map, (like on FA Mission 3, Red Flag)

- I added some enemy picking logic and its corresponding functions, so mission scripters can make use of `GetBestThreatTarget()` function, which requires said logic. I've been experimenting with making use of both the navigational mesh, and threat positions to dynamically pick attack positions for platoons. This could create some much needed variety to enemy AI pathing that were otherwise pre-created via marker chains. In case you require a demonstration, I have a sample map available, however that'll require a mod of mine as well that already features these changes, and some additions.


4z0t:
Could have referred transport logic from `Oxygen`...

Dhomie:
> Could have referred transport logic from `Oxygen`...

Initially I wanted to create transport pools for each base **by default**, by caching the platoon's origin base into it when it's formed initially, however that messed up missions that build transports from base `A`, while building land units at base `B`.
It also messed up transports that were assigned to `'TransportPool'` by map script (like Aeon M5).

So it ended up practically resembling the same method `Oxygen` uses, fair enough.

Issue #5263, opened 2023-07-23
Units Manager should not show dummy units 


HUSSAR-mtrela:
Units Manager shows dummy units (see purple squares) and throws errors and warnings in debug console

**To Reproduce**
- Open Units Manager in Lobby - Restrictions 
- Notice purple squares on right side of the screen

**Expected behavior**
Units Manager should not show dummy units and there should no be any errors and warning in debug console

![image](https://github.com/FAForever/fa/assets/11665978/0b853594-e30d-443b-be9b-29b84e234aca)

```
WARNING: UnitsAnalyzer /units/ues0401/ues0401_unit.bp - missing FACTION in bp.Categories
WARNING: UnitsAnalyzer /units/ues0401/ues0401_unit.bp - missing TYPE in bp.Categories
WARNING: UnitsAnalyzer /units/ues0401/ues0401_unit.bp - missing TECH in bp.Categories
...
WARNING: UnitsAnalyzer /units/xsc1902/xsc1902_unit.bp - missing FACTION in bp.Categories
WARNING: UnitsAnalyzer /units/xsc9010/xsc9010_unit.bp - missing FACTION in bp.Categories
WARNING: UnitsManager detected UNKNOWN faction without any LAND units
WARNING: UnitsManager detected UNKNOWN faction without any ACU units
WARNING: UnitsManager detected UNKNOWN faction without any SUPPORT units
WARNING: UnitsManager detected UNKNOWN faction without any NAVAL units
WARNING: UnitsManager detected UNKNOWN faction without any ECONOMIC units
WARNING: UnitsManager detected UNKNOWN faction without any SCU units
WARNING: UnitsManager detected UNKNOWN faction without any AIR units
WARNING: UnitsManager detected UNKNOWN faction without any CONSTRUCT units
WARNING: Error running HandleEvent script in CScriptObject at 1c1c0e00: a:\github\faf\fa\lua\ui\lobby\unitsanalyzer.lua(496): access to nonexistent global variable __blueprints
         stack traceback:
         	[C]: in function `error'
         	a:\github\faf\fa\lua\system\config.lua(54): in function <a:\github\faf\fa\lua\system\config.lua:53>
         	a:\github\faf\fa\lua\ui\lobby\unitsanalyzer.lua(496): in function `GetWeaponProjectile'
         	a:\github\faf\fa\lua\ui\lobby\unitsanalyzer.lua(583): in function `GetWeaponSpecs'
[...]

PR #5264, opened 2023-07-23
exclude dummy units in units manager

HUSSAR-mtrela:
fixes issue #5263 by excluding UNSPAWNABLE units that were shown as purple squares in Units Manager and fixed access to blueprints in Unit Manager

image

### DIFF
--- a/lua/ui/lobby/UnitsAnalyzer.lua
+++ b/lua/ui/lobby/UnitsAnalyzer.lua
@@ -52,6 +52,7 @@ CategoriesSkipped  = {
     ["NOFORMATION"] = true,
     ["UNSELECTABLE"] = true,
     ["UNTARGETABLE"] = true,
+    ["UNSPAWNABLE"] = true,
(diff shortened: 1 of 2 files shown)


Garanas:
You could also exclude the `DUMMYUNIT` category, which should end up using this class:
- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L5218

It is used by the Cybran build drones

HUSSAR-mtrela:
@Garanas as you requested, I excluded DUMMYUNIT category and INSIGNIFICANTUNIT since we do not need it in Units Manager

Garanas:
👍 , great work. I'll review this accordingly later this week

Garanas (approved):
Great work, fixes a lot of problems with minimal changes

ComradeStryker:
The new Mobile Factory is considered a Dummy unit.
Is that an exception?

image

Garanas:
No, it should be excluded too 😄

PR #4513, opened 2022-12-17
Blueprint cleanup draft

Hdt80bro:
Adds blueprint annotations to nearly every remaining blueprint field the game recognizes in some manner (including from individual unit scripts or unused engine functionality). Some, but not all, useless fields are marked with `unused`.

Some existing blueprint fields are interpreted as being incorrect and are fixed / removed:
* Identifiable typos (e.g. `PreferPrimaryWeaponTarget` instead of `PrefersPrimaryWeaponTarget`) are fixed. In particular, `Shader` has tended towards `ShaderName`.
* blatantly superfluous fields (e.g. `Accurate` or `MuzzleSalvoChargeDelay`) on a couple things (wide-spread superfluousness such as `ProjectilesPerShot` and `ConstructionBar` is left unaffected) are removed
* known bad structures (such as the mess that was the [T2 Seraphim Artillery](https://github.com/FAForever/fa/pull/4513/files#diff-935dc17faf0f4c4eb03c946f4e85b92a401121387048cf284e1a358a23e5e2d9)) are refactored
* a couple boolean fields are coerced from numbers to `true`

Note that some of these fields are still unused (e.g. `AttackReticleSize`), while others would affect balance and are commented out. These are:
* Mongoose, Heavy Fragmentation Grenade Launcher `MuzzleVelocityRandom` → 0.5
* Brick `BackUpDistance` → 5
* Ythotha `WaterVisionRadius` → 50

The first one is clearly game changing, while the second two make sense, but are still commented out.
A few Disintegrator Pulse Lasers had a `ChargeDelay` field (instead of `MuzzleChargeDelay`). Were these to be changed instead of removed, this would affect the following units' `MuzzleChargeDelay`:
* Dragon Fly, Disintegrator Pulse Laser → 1
* Cybran SCU, Disintegrator Pulse Laser → 0.5
* Wailer, Disintegrator Pulse Laser → 0.5

However, I opted to remove the field, along with `SalvoShotDelay`.

~~One of the few observable differences comes from the now correctly-formed contrail override for the Mercy:~~
MercyContrail
~~Also note that this makes them much more highly visible, which could be balance changing, but provides no more additional information to the player than they already have.~~
This change has been commented out

PR #5268, opened 2023-07-23
Fix various bugs with the unit restrictions menu

Garanas:
Closes https://github.com/FAForever/fa/issues/5263

Units with the categories `INSIGNIFICANTUNIT` and `DUMMYUNIT` are now ignored. Units with the `INSIGNIFICANTUNIT` category are ignored in the command graph. Units with the `DUMMYUNIT` category usually inherit from the [DummyUnit](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Unit.lua#L5218) class which has essentially none of the properties that you'd expect from a regular unit. It exists to reduce the impact of the unit to a minimum, used primarily by Cybran build drones as described in:

- https://forum.faforever.com/topic/2334/questions-about-performance-cybran-build-drones

Reverts a small patch from #4513 that assumes the `__blueprints` value is populated. This is not the case in the lobby

image

### DIFF
--- a/lua/system/Blueprints.lua
+++ b/lua/system/Blueprints.lua
@@ -891,6 +891,8 @@ local function SpawnMenuDummyChanges(all_bps)
                     Categories = {
                         'DRAGBUILD',
                         'UNSPAWNABLE',
+                        'INSIGNIFICANTUNIT',
+                        'DUMMYUNIT',
                     },
                     Display = {
                         BuildMeshBlueprint = '/meshes/game/nil_mesh',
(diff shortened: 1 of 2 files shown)


Garanas:
Ah, damnit - I completely forgot @HUSSAR-mtrela also did #5264

Garanas:
@HUSSAR-mtrela can you include the `DUMMYUNIT` category and the changes to `blueprints.lua` to your PR? I'll close this one

Issue #4823, opened 2023-03-19
Option to spawn in a specific number of units more easily


Basilisk3:
**Describe the feature**
When you want to test something with a specific number of units, it is easier to just punch in the exact number of units you want (for example, 100) instead of drag-spawning them on the screen. 

**Expected behavior**
The ability to spawn units via dragging was a very useful addition to the debug menu, but it would be great to have both options.

Basilisk3:
Closing this, as it has been completed in #5269.

PR #5269, opened 2023-07-24
add unit count input, fix army selector in debug/cheat window

HUSSAR-mtrela:
- [x] fixed layout and styling of observer selector with an icon at top of debug/cheat window
- [x] fixed styling of vertical scrollbar to match theme of debug/cheat window
- [x] improved readability of unit IDs in the list
- [x] improved of tooltips in unit list
- [x] added filters for Playable Units and Civilian units 
- [x] added unit tech level next to unit description
- [x] added unit faction icon next to unit description
- [x] added unit count input at bottom of debug/cheat window - issue:
 #4823

image


lua/ui/dialogs/createunit.lua
```diff
@@ -546,19 +554,19 @@ GetNameFilters = {
...
-                            return HasCat(unitID, 'SORTOTHER') or not (
-                                HasCat(unitID, 'SORTCONSTRUCTION') or
-                                HasCat(unitID, 'SORTECONOMY') or
-                                HasCat(unitID, 'SORTDEFENSE') or
-                                HasCat(unitID, 'SORTSTRATEGIC') or
-                                HasCat(unitID, 'SORTINTEL')
-                            )
-                        end,
-                    },
+                    -- { -- this sort does not affect unit list at all
```
Garanas, line 557: Could you remove the code instead of commenting it? No need to keep commented code around

lua/ui/dialogs/createunit.lua
```diff
@@ -975,6 +982,8 @@ function CreateDialog()
 
     local NumberInputFields = {
         units = {
+            -- creating input fields for count, vet, and rotation
+            {label='Count', name = 'Count',     default=1,   check=math.max, max=200},
```
Garanas, line 986: Should the max be 200? If you drag with anything beyond 20+ you'll easily end up with hundreds of units

HUSSAR-mtrela:
@Garanas made corrections based on your review and also I added faction icon and tech level to the list of units:
image

HUSSAR-mtrela:
@Garanas I made a few more improvements (see PR description) and I think this PR is good for final review.

Garanas:
I'll look at it this weekend 👍

speed2CZ:
Can we make the unit count and input field again?
Also didnt it used to space out the units, based on footprints, when spawning more than one?

Garanas:
> spacing units out based on footprint would be a new feature. Which direction should units be spaced out: from mouse click outwards (rings), in a square of with dimension of sqroot(unit count) or in some other way?

We had that, but it got removed in favor of the dragging the units. See also:
- https://github.com/FAForever/fa/pull/4770/files#diff-fd177b8a8078b12f4405de59c301724307f10d7999e2c835f2e0e5980354c155L389

We can try and add it back in, but it won't work well with dragging units 😄

lua/ui/dialogs/createunit.lua
The-Balthazar: I missed this discussion by a year, but it has function. It listed everything that would be sorted after the divider after intel sorted units. It's wrong to say it did nothing.

PR #5282, opened 2023-07-27
Fix tech 3 mobile anti air being at the front of a formation

Garanas:
Manually adds the `WEAKDIRECTIRE` category to the Bouncer and the Lightning tank, and uses the category to exclude units from the `DIRECTFIRE` category when computing the formation of a unit.

image

image


Garanas:
As suggested by the balance team 😄

PR #5287, opened 2023-07-28
Fix/ai acu threat values in blueprints

relent0r:
A previous PR ran the threat calculation on ACU's which produced overinflated surface threat values due to the overcharge weapons.
This PR reverts the values to a number that is between a standard ACU and a gun upgraded ACU. This will avoid the AI from refusing to engage with ACUs due to the massive amounts of threat they were generating.


Garanas:
This may be because of the formatting we did.

PR #5295, opened 2023-07-30
Stinger and C14 Star Lifter: Fix, so they no longer miss low height units

Basilisk3:
**Changes:**
Added a slight amount of homing to the Stinger and C14 Star Lifter
MuzzleVelocity 40 --> 35 to prevent the unit from missing low height units (Stinger only)
Health = 880 --> 920 (Stinger only)

Tries to fix the Stinger (and the C14 Star Lifter) missing units it really shouldn't miss. It also gains a slight buff to its hit points in line with its faction's theme.

### DIFF
--- a/projectiles/TDFRiot01/TDFRiot01_proj.bp
+++ b/projectiles/TDFRiot01/TDFRiot01_proj.bp
@@ -44,9 +44,6 @@ ProjectileBlueprint {
         DestroyOnWater = true,
         InitialSpeed = 13,
         UseGravity = false,
-        TrackTarget = true,
-        TurnRate = 25,
-        OnLostTargetLifetime = 0.5,
         VelocityAlign = true,
     },
 }
--- a/projectiles/TDFRiot01/TDFRiot01_proj.bp
+++ b/projectiles/TDFRiot01/TDFRiot01_proj.bp
@@ -44,6 +44,9 @@ ProjectileBlueprint {
         DestroyOnWater = true,
         InitialSpeed = 13,
         UseGravity = false,
+        TrackTarget = true,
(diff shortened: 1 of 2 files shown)


Basilisk3:
Ready for review.

Tagada14:
You need to be careful with the muzzle velocity, if you make it too high then the unit will miss land units that have small height.
@Garanas Correct me if I am wrong and this has been fixed btw.
If it wasn't then what happens is the projectile impacts the ground at the end of the tick and the game doesn't register the hit to the unit.

Garanas:
It is not - the bug is still in the game. When a projectile hits multiple targets in the same tick then it ends up processing the furthest (or last) target it hit, which could be the terrain.

Basilisk3:
I reduced the increase to 50 to be on the safe side. Is there a rough estimate at which values this bug starts to occur? After some tests, it appears to work fine at 50.

Tagada14:
It would be best to test on some small units to be on the safe side. I suggest trying it out with Aeon T1 artilleries since they are very short.

Basilisk3:
I tested it, and that bug occurs even at the muzzle velocity we have right now, which is 40 (though rarely). At lower values, it seems to never happen. At 60 the unit misses the Fervor most of the time. 

I will investigate adding soft tracking to it again.

Garanas:
An alternative solution is to increase the vertical height of the collision box of 'flat' units

Tagada14:
Please revert the hp changes for the Stinger, we would like to buff it in other ways.

Basilisk3:
@Tagada14 I reverted it.

Tagada14 (approved):

PR #5300, opened 2023-08-01
Use logarithmic values for score

maudlin27:
See the suggestion here:
https://forum.faforever.com/topic/6256/option-for-logarithmic-scaling-in-the-score-board?_=1690917013215

Having a logarithmic value for the score makes it easier to see how well players are performing throughout the game, since the score typically tends to increase closer to an exponentional value than a linear value as the game progresses.

This change uses logarithmic values to adjust how the graphs are displayed, and the y axis intervals, while still showing the original (pre-log) values.


maudlin27:
One point to flag - I wasn't sure what the label in row 371 related to (it's not the mouse-over on the graph itself), but I've added code to convert the figure from a logarithmic value to be safe.

Garanas:
@maudlin27 do you know what base it uses, is it 10?

image

Garanas (changes requested):
I end up with quite a few graphs that look like this:

image

Which doesn't use the top part. Usually running a game for 15+ minutes turns a graph into this. I like the idea, but the scale is off here

maudlin27:
Updated for the following:

- Moved logic adjusting the scaling into separate functions
- Switched from logarithmic scaling to using square root of the value - main reason is logarithmic would scale off too quickly, so the majority of the graph was taken up by the first minute of the score which is the least relevant part
- Adjusted y axis so most of the graph is used, and also made it 5% higher than the highest value so it's more likely the highest value labels appear within the graph
- Added commas for easier reading of large values

Illustrative example:
image

Garanas:
image

image

image

lua/ui/dialogs/hotstats.lua
```diff
@@ -852,7 +886,7 @@ function create_graph(parent,path,x1,y1,x2,y2)
                 infoText = false
             end
             if posX()>x1 and posX()<x2 and posY()>y1 and posY()<y2 then
-                local  value = tps_format((posX()-x1)/(x2-x1)*scoreInterval*data_nbr) .. " / " .. math.floor(((y2-posY())/factor))
+                local  value = tps_format((posX()-x1)/(x2-x1)*scoreInterval*data_nbr) .. " / " .. FormatNumber(math.floor(ReverseScaling((y2-posY())/factor))) --This is the value that gets shown when we hover the mouse over any point in the graph, so want to do exponentional to reverse the logarithmic so we see the actual value
```
Garanas, line 889: Can you update the comment here, it still mentions the exponentional / log approach

lua/ui/dialogs/hotstats.lua
```diff
@@ -476,6 +477,35 @@ function makeKMG(int)
     return (math.floor(int/1000000000)).."."..(math.floor((int/1000000000)-math.floor(int/1000000000))*10).." G"
 end
 
+function AdjustValueScale(val)
+    return math.sqrt(val)
+end
+
+function ReverseScaling(val)
+    --LOG('Val pre reverse='..val..'; expected val post reversal='..(val * val)..'; scaled value of this='..AdjustValueScale((val * val)))
```
Garanas, line 485: Please remove commented code

maudlin27:
Updated

Garanas:
Great work @maudlin27 !

PR #5303, opened 2023-08-03
T3 arty bug fix

RabidPope:


Garanas:
Question is however, what does the reset target accomplish? 😄

RabidPope:
Is it not to let the units keep finding better targets?

Garanas:
What if instead of having a fixed time of 5 seconds, we set it to reload time + 2.0 seconds?

RabidPope:
> What if instead of having a fixed time of 5 seconds, we set it to reload time + 2.0 seconds?

I have no idea what you mean by this. What would changing the reload time change?

Garanas:
Yes, it is not fixed to five seconds. But what if we'd make it dynamic?

RabidPope:
Ah. So unit reload time + 2 seconds? Will the timings sync correctly?

RabidPope:
@Garanas Seems to work. Hope it doesnt break anything else. Did some quick checks for that but maybe you could check also?

Garanas:
I'll see what I can find!

Garanas:
Done some testing - apologies that it took so long.

The code we're changing here triggers a lot when interceptors / ASFs are fighting. Not sure if changing the delay has a significant impact there, I'll be doing a few more tests

Garanas:
urg, file formatting 😄

lua/sim/defaultweapons.lua
```diff
@@ -837,12 +843,15 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
...
+            -- therefore the thread of this state is not destroyed
 
-            -- Check if there is a better target nearby            
-            self:ResetTarget()
-            
+            if not (IsDestroyed(unit) or IsDestroyed(self)) then
+                -- wait reload time + 2 seconds, then force the weapon to recheck its target 
+                WaitSeconds((1 / self.Blueprint.RateOfFire) + 3)
+                self:ResetTarget()
+            end
```
Garanas: This is the relevant bit

PR #5310, opened 2023-08-05
Finalize new capping behavior

Garanas:
The current capping / ringing behavior has its flaws and I hope that with this alternative we can overcome those flaws. To be specific:

- (1) The current implementation can cause serious slow downs, to the point where it halts the game
- (2) The current implementation does not apply immediately, but after five ticks
- (3) The current implementation always applies a guard or assist command on the unit you're trying to cap
- (4) The current implementation is difficult to customize using UI mods
- (5) The current implementation creates duplicated build orders that are difficult to remove individually
- (6) The current implementation does not take into account other queued structures

The new implementation relies on templates. These templates are pre-defined and adjusted on the fly. All the templates are defined in `lua/ui/game/commands/generated-templates-data.lua`. The description of the various fields are in the same file, I won't repeat them here as they may change as the feature matures.

In the next paragraph we assume that the hotkey to cycle is assigned to `tab`.

The new implementation works by hovering over a unit and cycle through the available templates using `tab`. All the available templates are filtered on whether they apply. The criteria depend on the unit you're hovering over, the units in your selection and whether these units are supposed to be able to construct the structures in the template.

https://github.com/FAForever/fa/assets/15778155/e133bd09-a65e-45d0-8e38-940e35b798f4

https://github.com/FAForever/fa/assets/15778155/9950a522-c6e2-4fb8-942a-e3fe7fa21b9c

The new implementation is 'natural' to the game: we use the template build feature by feeding it predefined templates that we change on the fly to create a new template. The new implementation is also trivial to mod: you add your templates to `lua/ui/game/commands/context-based-templates-data.lua` and you're done. There's no hooking involved. As a result people can create or remove templates as they see fit. See also the following example mod:

Previous pull requests:

- https://github.com/FAForever/fa/pull/5309
- https://github.com/FAForever/fa/pull/5308
- https://github.com/FAForever/fa/pull/5306
- https://github.com/FAForever/fa/pull/5304
- https://github.com/FAForever/fa/pull/5302


Garanas:
A few small additions:

- Instead of `TriggerOnEmptySpace`, we include:
- - `TriggerOnLand`
- - `TriggerOnWater`
- - `TriggerOnMassDeposit`
- - `TriggerOnHydroDeposit`

Where the last two allows us to quickly queue an extractor or hydro over different tech levels

PR #5312, opened 2023-08-06
Change depth preview to better show unpathable water

gordenwunderlich:
- Instead of always showing 4 dots the depth indicator will show dots at a fixed interval, so number of dots~depth
- The dot at navy required depth (1.5) will have a different color
Old: 
i
New:
image
- When hovering over too shallow water the final dot will turn red
image

Things to look at:
- Maximum number of dots is currently fixed at 8, not sure if that is enough/too many (possible to compute max water depth on game launch?)
- Red dot becomes hard to see due to transparency increasing when near the surface
- Instead of duplicate SetMesh calls (lots of identical code), maybe use a local var to hold texture (additional variable, higher complexity code, can't be easily done for the red dot)
- Might be smarter to change the color of the last dot instead of having 2 meshes prepared
- looks a bit weird in build mode. Building not placeable but indicator showing sufficient depth in the center (maybe detect the four corners instead and show which corner is to shallow. Limiting to one corner to not increase number of prepared dots required?)


Garanas:
This looks like a great start. Your last suggestion is good, but difficult to implement accurately. I'd encourage you to try that in a separate pull request once this one is complete 😃 

You're right that there is duplicated code, but nothing that appears harmful in my opinion.

> Red dot becomes hard to see due to transparency increasing when near the surface

This is introduced by the Lua code. It was intended to create a 'soft' transition. Feel free to adjust the parameters or behavior to make it work better.

Garanas:
In case it is unclear - we spoke on Discord about this feature. My name there is `Jip`

gordenwunderlich:
Ok fiddled around a bit and reduced the minimum depth for the indicator. This seems more natural to me and it was a bit confusing when testing water and the indicator disappeared "randomly". The red is much more visible than the yellow indicator so the difference between indicator disappearing (not deep enough) and turning yellow (deep enough) was hard to see.

@Garanas
what do you think about these two points
- Maximum number of dots is currently fixed at 8, not sure if that is enough/too many (possible to compute max water depth on game launch?)
- Might be smarter to change the color of the last dot instead of having 2 meshes prepared

1. might cause issues on maps with very deep water
2. is more of a technical question

Garanas:
> * Maximum number of dots is currently fixed at 8, not sure if that is enough/too many (possible to compute max water depth on game launch?)

We can't compute the maximum water depth at game launch. I don't recall the height between dots, but what if we go as far as a depth of 25? Anything beyond is not passable for amphibious units, for example

gordenwunderlich:
ok, the distance between dots is 0.75 (so 2 dots hit the 1.5 depth needed for navy), so we'd need 31 dots for depth 25 (dot 32 would appear at 25.5). If we take less the dots should just stop appearing at higher depth, which is probably also a good solution compared to increasing the dots from 8 -> 31, when they are rarely ever needed.

Garanas:
I've done some testing, apologies that it took so long. In general I think these changes are great. I noticed that it isn't 100% representative of the truth, I'm not sure where the difference is. The navigational mesh is also not 100% accurate. As an example:

image

Take the location of the submarine. Submarines have the same minimal depth. Yet, the submarine can go there but neither of water-based units can.

Garanas:
@gordenwunderlich would you have time to make a change this week? Then we can include it for the next patch, the 2nd of September.

What if we just show two points:

- The sphere at a maximum depth 1.5, being yellow if deeper otherwise red
- The sphere at the ocean floor

And when the depth is 1.5 or less we just show one sphere that is red.

Any other intermediate point feels meaningless. The tool was originally developed to help understand the offset of orders for units on the ocean floor. Especially useful when trying to micro reclaim with SACUs or engineers.

gordenwunderlich:
> I've done some testing, apologies that it took so long. In general I think these changes are great. I noticed that it isn't 100% representative of the truth, I'm not sure where the difference is. The navigational mesh is also not 100% accurate. As an example:
> 
> image
> 
> Take the location of the submarine. Submarines have the same minimal depth. Yet, the submarine can go there but neither of water-based units can.

It's not 100% accurate, but without the actual navmesh I think that's impossible.
the reason here is probably that the frigate is larger than the sub. It's not exactly the same spot, but the same navmesh layout and the frigate can fit if you click in the center of the square.
image

image
once you click closer to the land it backs off further than it could.
It won't tell you were your ships can move, but it should help you find untraversable bits.
(Is this your spot?^^)
image


> What if we just show two points:
> 
>     * The sphere at a maximum depth 1.5, being yellow if deeper otherwise red
> 
>     * The sphere at the ocean floor
It _could_ help you gauge the depth of amphibious units to see if you can groundfire them, but that doesn't seem too relevant.
And if a mod has navy units with different depths the feature breaks anyway, so might as well keep it to what's relevant in the "base game".

looks a bit lonely now :(
image
also looks a bit less intuitive on what it's trying to communicate
image
Would be nice to get some feedback from people unfamiliar with it.

ComradeStryker:
> @ComradeStryker You use this feature, what do you think of the suggested change?

I don't really see an issue with the current iteration. It works well, and gets the information needed across really easily.
That being said, there have been, at times, a few instances in which I glanced at the marker, and thought the water was deeper than it actually was.

Being able to see a color rather than a line of yellow, would probably be an improvement.
If it's red, it's shallow, if it's blue, it's deep, or whatever the proposed change was.

I think just seeing yellow all the time, eventually you will overlook it, because it blends in or you just get used to it...
like I did. 
So It may be good to change.

---

~ Stryker

gordenwunderlich:
@ComradeStryker 
I'm wondering which of these is more clear (especially how you think newer players will perceive this)
Both version have the red dot to indicate water too shallow for navy
> * image
But the new one only shows one intermediate dot (at the 1.5 depth required for navy)
> image
While the other one has multiple dots in between to form more of a line towards the ocean floor (and has a special mark at 1.5 depth)
> image

@Garanas
Should the description of the option also get an update? I think it should mention that it now shows where navy definitely can't go.

ComradeStryker:
So the yellow dots are still there, but there is now a red dot to indicate an area where navy is not able to be built,
and a blue dot to indicate where navy is able to be built?

Yeah, that seems like a huge improvement to indicate water depth.

As for "more user-friendly", well... 
Perhaps there could be a switch to swap between the two modes?
Let the players decide if they like the current one or the new reworked version.
If that's possible to do, that may be the way to go.

But as I mentioned before, the current one works well for what it was made to do, but that's it.
If the player is remotely distracted, the information may not be clear at a (quick) glance, in the current iteration.

---

That all being said... I'd really have to see this new version to fully understand it.
Is there any way I can test it for myself?
That way I can provide more accurate feedback.

---

~ Stryker

gordenwunderlich:
@ComradeStryker
to test it you currently need to download the repo and launch the game via the dev launcher.

Garanas:
@ComradeStryker You should have a dev environment setup. You did previous work on the repository. You can check out on this branch

Garanas:
I've decided to just go with it for now. We're free to change it up again in the future, to include additional bubbles for example. The ability to have a rough estimation as to whether water is deep enough is more important to me at this moment.

gordenwunderlich:
> because it still measures the depth at the mouse cursor even though the ball stays at the center of the structure

I can see that I fix that.
meanwhile I made a bit of a middle ground, by making the first two dots fixed at .5 and 1.5 and showing up to 3 additional dots based on the terrain difference.
The function for calculating the dots position is a bit more complex now. But A few basic math operations should slow it down much right? (position[2] - k*0.5 - (k-1)*0.5 - (k-1)*(k-2) * 0.05 * (position[2] - elevation))
image
(deepest part of the void)

image
(shallow water)

image
(medium deep water)

gordenwunderlich:
@Garanas
after looking at it this part
if info[1] == 'build' then
                position[1] = (position[1] ^ 0) + 0.5
                position[3] = (position[3] ^ 0) + 0.5
            end
sets the x,z positions to the center of the building, but how would I get the y position at that point?
since it's not predictable like the x or z coordinates

Garanas:
You don't. The UI layer is not aware of the heightmap except of the location of the mouse

BlackYps:
Maybe it is better then to not artifically set the x and z position of the dot, but to still have it follow the mouse.
But all of this discussion should probably move to a new PR

PR #5322, opened 2023-08-16
Move the distribute orders feature to a separate file

Garanas:
This turns it into a generic function that scripts and AI can use too.


Garanas:
@maudlin27 / @relent0r this may interest you. You can of course program it out in the AI, but this allows for a trivial way to spread commands over units.

Note that the function assumes that all units share the same command queue.

PR #5327, opened 2023-08-16
Double the Valiant's anti-air damage

Basilisk3:
For all intents and purposes, the Valiant’s anti-air solution was almost useless. This PR aims to make it more useful, while still keeping it slightly worse than the Salem’s, in accordance with the themes of both factions (stronger cruiser AA, weaker frigate/destroyer AA versus weaker cruiser AA, stronger frigate/destroyer AA).

### DIFF
--- a/units/UES0201/UES0201_unit.bp
+++ b/units/UES0201/UES0201_unit.bp
@@ -319,7 +319,7 @@ UnitBlueprint{
             BallisticArc = "RULEUBA_None",
             CannotAttackGround = true,
             CollideFriendly = false,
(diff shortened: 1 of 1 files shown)


Tagada14:
UEF already has the best naval AA capabilities with its combo of cruiser and shield boat. Cybran on the other hand has a weaker AA on the cruiser and thus has better AA on other ships (Frigate, Destroyer). Overall the AA on these ships is mostly to deal with scouts and random bombers. We don't believe there is a need to buff it.

PR #5335, opened 2023-08-18
Navigational mesh - Rectangle maps

maudlin27:
Fixes issue where navmesh wouldnt generate properly for non-square maps, like M2 of the Seraphim campaign

Previously the navigational mesh would compress locations based on the map size, but was only taking the x map size.  On a rectangular map where x is smaller than y (i.e. the actual map size, not just the playable area), this would lead to labels being generated for only part of the map.

The following is what the land grid looked like on M2 of the Seraphim campaign (which is a rectangular map):

image

This is the result after this change on the same map/replay:
image

Testing on a couple of other maps (using different colouring for different plateaus) it still looks ok.
#
image

image

One issue to note on dual gap is it's still seeing locations outside the playable area as pathable (I'd thought the issue had been patched out a while ago), but checking on the current FAF develop it has the same issue so isn't caused by these changes.

### DIFF
--- a/lua/sim/NavGenerator.lua
+++ b/lua/sim/NavGenerator.lua
@@ -1433,7 +1433,7 @@ function Generate()
     NavLayerData = Shared.CreateEmptyNavLayerData()
 
     ---@type number
-    local MapSize = ScenarioInfo.size[1]
+    local MapSize = math.max(ScenarioInfo.size[1], ScenarioInfo.size[2])
 
     ---@type number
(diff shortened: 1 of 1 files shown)


Garanas:
@maudlin27 If you use the `OnBeginSession` brain function to generate the navigational mesh then it works as expected:

image

See also: https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/aibrains/base-ai.lua#L131-L147

Garanas:
Note that `OnBeginSession` is a function that only exists in FAForever

PR #5336, opened 2023-08-18
Fix Ythotha's Ion Storm not damaging allied units

4z0t:


Garanas:
What lines contain the actual changes?

lua/seraphimunits.lua
```diff
@@ -581,12 +600,16 @@ SEnergyBallUnit = ClassUnit(SHoverLandUnit) {
             local beamLifetime = bp.Weapon[1].BeamLifetime or 1
             local reaquireTime = bp.Weapon[1].RequireTime or 0.5
             local weapon = self:GetWeapon(1)
+            local bp = weapon.Blueprint
 
             self:ForkThread(self.LifeThread)
 
+            weapon:SetEnabled(false)
```
4z0t, line 607: Here.

lua/seraphimunits.lua
```diff
@@ -595,18 +618,20 @@ SEnergyBallUnit = ClassUnit(SHoverLandUnit) {
...
                 if target then
                     weapon:SetTargetEntity(target)
                 else
-                    weapon:SetTargetGround({location[1] + Random(-20, 20), location[2], location[3] + Random(-20, 20)})
+                    weapon:SetTargetGround { location[1] + Random(-20, 20), location[2], location[3] + Random(-20, 20) }
                 end
+
                 -- Wait a tick to let the target update awesomely.
-                WaitTicks(2)
+                WaitTicks(1)
```
4z0t, line 631: Here

units/XSL0402/XSL0402_unit.bp
```diff
@@ -192,7 +192,7 @@ UnitBlueprint{
                 Land = "Land|Water|Seabed",
                 Water = "Land|Water|Seabed",
             },
-            FiringTolerance = 1,
+            FiringTolerance = 0,
```
4z0t, line 195: Here

4z0t:
@Garanas awaiting review.

Garanas:
I'll review it somewhere this week, there's no rush

4z0t:
@Garanas so?

Garanas:
Looks functionally fine but it turns the storm into an absolute beast. Up to the balance team to decide

4z0t:
Up to balance team to decide whether should it damage allied units? Or whether should it damage less?

Garanas:
All of it. This changes balance, that is not my call.

4z0t:
How did you decide whether it is balance change when you tried to fix it previously?

Garanas:
Previously I turned it back into what I think was the old behavior. You're turning it into an absolute beast, even stronger then what you find in Steam FA. That's how I decided

4z0t:
Does this mean we have double standards dedicated to balance changes? You are allowed and others cant?

Garanas:
Yep, sure - if you want to call that a double standards then go for it. Do take into account that I discuss almost everything with the balance team, or at least give them the opportunity to respond to it.

4z0t:
The old behaviour was storm that randomly damages allied and enemy units. Current behaviour obviously prefers enemy.

Garanas:
[August 21, 2022](https://github.com/FAForever/fa/releases/tag/3742)

https://github.com/FAForever/fa/assets/15778155/dc89d82f-dd6e-4998-8d9a-49952313aa4d

Current on FAF Develop:

https://github.com/FAForever/fa/assets/15778155/64c0039b-0213-4f7c-8f53-5cc537b83e13

Your suggested fix:

https://github.com/FAForever/fa/assets/15778155/a9c54fba-27ae-4366-829f-85e896f03371

I agree with you that what we have now (because of https://github.com/FAForever/fa/pull/5291) is still weaker than the damage output of 2022. But the original damage output is still way less than the damage output of the changes you're suggesting.

The commit of 2022 was chosen randomly. A release far enough back in time so that https://github.com/FAForever/fa/pull/4863 wasn't part of the release. That pull request caused the original bug where the storm would ignore all allied units.

Now please, take this up with the balance team. And I'd appreciate it if you'd apologize for 'calling me out to have double standards'. It was uncalled for, and you know it

4z0t:
Yeah, sorry for that. Offensive from my side. But I don't understand why you test it without enemy units at all. You clearly see it misses shots, thats why it does less damage to allied units.

Garanas:
It doesn't matter whether it misses or not. What matters is that these changes almost triple the damage output to allied units. That is a considerable change in comparison to August 2022. It is up to the balance team to decide whether that is desirable

lua/seraphimunits.lua
lL1l1, line 631: This logically should change the fire rate of the storm from every 4 ticks to 3 ticks, but there's something weird with `WaitSeconds(reaquireTime)` where `reaquireTime = 0` waiting 1 tick that's causing that to not change. I think you could get rid of this part and just set reaquireTime to a minimum of 0.1 seconds?
Also `reaquireTime` should be spelled `reacquireTime` and `RequireTime` in the BP should be `ReacquireTime`.

lua/seraphimunits.lua
```diff
@@ -595,18 +618,20 @@ SEnergyBallUnit = ClassUnit(SHoverLandUnit) {
                     end
                 end
 
+                ---@type Unit
                 local target = table.random(filteredUnits)
+
                 if target then
```
lL1l1, line 624:
```suggestion
                if target and (not IsAlly(self.Army, target.Army) or Random(1,5) <= 1) then
```
This could moderate ally damage while retaining visual effects. Some experimentation would be needed to determine the most balanced chance to hit allies. A storm has 152000 potential damage (2000 dmg, 4 tick waits, 30 second lifetime). I did 9 tests of 3 ythotha standing next to each other in a row, and killing the one on the right got me an average of 28700 damage to allies per storm, so for this suggestion I'll use a 1/5 chance.
In addition, there is currently an issue that the storm mostly only damages allies right next to the minimum range, so to retain that feature you could reduce the chance as range increase, although I don't think that's a good mechanic (it's hard to notice as a player), so I didn't try.

lL1l1 (changes requested):

MrRowey:
This is what me, sladow, arch and blodir agree on with the ion storms 

 I like the current ion storm because it makes thematic sense. It has no brain and is just lashing out at anything nearby. Blodir thinks it's nice to make it more consistent but it should match the animation by having more lashes in more directions (instead of making it homing) so rng plays less of a factor.

PR #5343, opened 2023-08-21
Document terrain shaders and streamline how they interpret properties

Garanas:
You can test the changes using `Auralian - The Core v19`. Also fixes some compression problems with the map, the game somehow interprets the values different and as a result the map looks different too 😄 

image

image

image

### DIFF
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -2033,53 +2033,116 @@ technique TerrainPBR <
     }
 }
 
+// ----------------------------------------------------------------------------
+//#region Terrain001 and Terrain002
+
+// Layer| Albedo stratum                                               | Normal stratum
+//      | R           | G             | B            | A               | R             | G             | B             | A            |
+//  ----            ---             ---             ---              ---             ---             ---             ---            ---
+// | L  | R             G               B              unused          | X               Y               Z               unused       |
+//  ----            ---             ---             ---              ---             ---             ---             ---            ---
+// | S0 | R             G               B              unused          | X               Y               Z               unused       |
+// | S1 | R             G               B              unused          | X               Y               Z               unused       |
+// | S2 | R             G               B              unused          | X               Y               Z               unused       |
+// | S3 | R             G               B              unused          | X               Y               Z               unused       |
+//  ----            ---             ---            ---               ---             ---             ---             ---            ---
+// | S4 | R             G               B              unused          | X               Y               Z               unused       |
+// | S5 | R             G               B              unused          | X               Y               Z               unused       |
+// | S6 | R             G               B              unused          | X               Y               Z               unused       |
+// | S7 | R             G               B              unused          | X               Y               Z               unused       |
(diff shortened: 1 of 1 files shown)


effects/terrain.fx
```diff
@@ -2089,25 +2152,29 @@ float4 Terrain001AlbedoPS ( VS_OUTPUT inV) : COLOR
...
-    float dotSunNormal = dot(SunDirection, normal);
-
-    float mapShadow = 1 - utility.z; // 1 where sun is, 0 where shadow is
-    float shadow = tex2D(ShadowSampler, inV.mShadow.xy).g; // 1 where sun is, 0 where shadow is
-    shadow = shadow * mapShadow;
+    // compute lighting
+    float dotSunNormal = max(0, dot(SunDirection, normal));
 
+    // combine everything
     float3 light = SunColor * saturate(dotSunNormal) * shadow + SunAmbience;
```
BlackYps, line 2171: You are clamping the dotproduct twice here

effects/terrain.fx
```diff
@@ -2284,17 +2359,17 @@ float4 Terrain003AlbedoPS ( VS_OUTPUT inV) : COLOR
...
     float3 r = reflect(normalize(inV.mViewDirection), normal);
-    float3 specular = pow(saturate(dot(r, SunDirection)),20) * mapSpecular * stratum7Albedo.w * shadow * SunColor;
+    float3 specular = pow(saturate(dot(r, SunDirection)), 20) * mapSpecular * stratum7Albedo.w * shadow * SunColor;
 
     // compute lighting
-    float dotSunNormal = dot(SunDirection, normal);
+    float dotSunNormal = max(0, dot(SunDirection, normal));
 
     // combine everything
     float3 light = SunColor * saturate(dotSunNormal) * shadow + SunAmbience;
```
BlackYps, line 2368: same here

effects/terrain.fx
```diff
@@ -2272,8 +2346,9 @@ float4 Terrain003AlbedoPS ( VS_OUTPUT inV) : COLOR
     albedo = lerp(albedo, stratum3Albedo, mask0.w);
     albedo = lerp(albedo, stratum4Albedo, mask1.x);
     albedo = lerp(albedo, stratum5Albedo, mask1.y);
-    // albedo = lerp(albedo, stratum6Albedo, mask1.z);
```
BlackYps, line 2275: you are not using stratum6, but you are also not documenting that in the overview table. If you start using it, you will run into the 16 texture limit that the shader version imposes. To prevent that I suggest that you move the specular slot from the Stratum7normal slot to the free channel of the upperAlbedo. This way you don't have to sample the Stratum7normal at all and gain a texture slot in the shader that you can then use to sample stratum6albedo.

Without that the shader doesn't compile because we exceed the 16 texture limit.

BlackYps:
> Also fixes some compression problems with the map, the game somehow interprets the values different and as a result the map looks different too

Can you elaborate on this?

PR #5346, opened 2023-08-22
Add tracking to the Vulthoo + small fix

Basilisk3:
**Changes:**
Added a slight amount of homing to the Vulthoo 
MuzzleVelocity 40 --> 35 to prevent the unit from missing low height units

Tries to fix the Vulthoo missing units it really shouldn't miss.


Garanas:
@Tagada14 / @Spikey84 is it okay if we include this with the next developers patch, set for the 26th?

Tagada14:
I prefer to have it together with the Stinger changes (which also buffs its HP which is deserved but would fit a Balance patch more). I don't think this is really high priority as Vulthus don't have too many issues.

Garanas:
@Basilisk3 Can you rebase to `deploy/fafbeta`?

Basilisk3:
@Garanas Yes. I will tackle the other gunships too. It is especially bad with the Jester, it misses the Fervor more than it manages to hit it. As that one is urgent, I will point it at fafdevelop instead. The other commits will be pointed at the balance branch.

PR #5348, opened 2023-08-23
Jester: Fix, so it no longer misses low height units

Basilisk3:
**Changes**
MuzzleVelocity 55 --> 35 

High values make units miss/shoot through low height units. It is especially bad with the Jester, it misses a Fervor more than it manages to hit it.

Pointed at fafdevelop, as it is relatively urgent and has little implications balance wise (it has homing).


Tagada14 (approved):

Garanas:
This branch is pointing to `deploy/fafdevelop`, is that intentional?

Tagada14:
Yes, Jester has the biggest issue so we wanted to have it roll out faster.

PR #5350, opened 2023-08-23
Wailer: Fix, so it no longer misses low height units

Basilisk3:
**Changes:**
Added a slight amount of homing to the Wailer's air to ground weapon
MuzzleVelocity 40 --> 35 to prevent the unit from missing low height units


Basilisk3:
@Garanas Yeah, I am only doing this for the gunships. It is pretty jarring, watching a broadsword struggle to hit a moving Titan.

By the way, I pointed the PR about the Jester to fafdevelop, so if you want, you can merge it for the game patch in two days. The Jester - Fervor interaction is very broken right now, over half of the Jester's don't register. That PR fixes that.

Tagada14 (approved):

PR #5359, opened 2023-08-25
Enable the PBR unit shader on all maps

BlackYps:
Rolls out the long-awaited PBR unit shaders on all maps.

Originally this was part of #5172 but I separated it into two parts


BlackYps:
I strongly recommend using a tool that uses the patience git diff algorithm to view the changes, the default used by github produces a very ugly diff. One way would be to use intellij IDEA.

BlackYps:
Switching the shader for navy units gets rid of the need to use a script like in #4932, the map simply needs to add the `water` environment map.
The shader is the same as the regular unit shader, but passes the `water` environment map.
As the environment name falls back to the default environment if it is not present, this is a transparent change.

Garanas:
Closed #4932 , do you (@BlackYps ) have time to walk through the changes by sharing your screen? I do not have Intellij IDEA installed

BlackYps:
Yeah, we can do that

Garanas (approved):
Reviewed it together and all is good

PR #5369, opened 2023-08-28
Add a fuzziness factor to where tactical missiles end up when firing on a unit

Garanas:
Adjusts where missiles end up shooting when they are firing at a unit. The initial implementation works, but because of how tactical missiles are scripted they'll end up flying in circles for minutes when they miss. There's no "generic" missile component, each missile kind of has its own behavior (to some degree, a lot are copy-pastes of each other). To fix that we'd need to create a generic component that we can tweak on a per-missile basis. But that is too much work.

Unrelated but related, sometimes missiles 'fly off' because they're (initially) moving away from the target. That is because a lot of the (copy-pasted) scripts assume they always go closer to the target, while that may not be true in practice. The component could fix that problem too.

Too much work for these few days, requires tinkering with the balance team. 

https://github.com/FAForever/fa/assets/15778155/ba7a3917-9239-4394-bdb1-939ee8ef0155

https://github.com/FAForever/fa/assets/15778155/4ede7fcc-759c-42bb-80cd-0d340de7a877

https://github.com/FAForever/fa/assets/15778155/3fbe6630-7181-493b-ae81-3033c6eea636

https://github.com/FAForever/fa/assets/15778155/d5cd8add-5b4e-4b42-b38c-65213660ef0c

https://github.com/FAForever/fa/assets/15778155/045af158-3e9d-4657-9ea6-08bf63d8dbe1


msgerbs:
This seems like it would make tactical missiles devastating against groups of units since it would be more likely to hit different units instead of just all blasting a single tank that died already.

Garanas:
It doesn't feel that much stronger, see also:

_new behavior of this pr_

https://github.com/FAForever/fa/assets/15778155/9706b77f-39f4-4b9d-9610-d3e4a1bc05bc

_current behavior on release branch_

https://github.com/FAForever/fa/assets/15778155/b257b94e-3196-4a8b-9d5b-b81baa191935

You're right in that it hits more units on average. It is a balance issue, discussed with the balance team 😄

Garanas:
@clyfordv this works very nicely with your changes on the tactical missiles 👍

PR #5370, opened 2023-08-29
Multiple pod orders

clyfordv:
Can select multiple pods/pod platforms when multiples of their counterpart is selected. Stock behavior remains unchanged.

### DIFF
--- a/lua/ui/game/orders.lua
+++ b/lua/ui/game/orders.lua
@@ -636,7 +636,7 @@ local function DroneInit(self, selection)
     local lastMode = nil
     if self._pod and next(self._pod) then
         for _, pod in self._pod do
-            if not lastMode then
+            if lastMode == nil then
                 lastMode = pod:IsAutoMode()
             elseif lastMode ~= pod:IsAutoMode() then
                 mixed = true
@@ -1046,8 +1046,8 @@ local defaultOrdersTable = {
     RULEUCC_Repair = {              helpText = "repair",            bitmapId = 'repair',                preferredSlot = 14, behavior = StandardOrderBehavior},
     RULEUCC_Dock = {                helpText = "dock",              bitmapId = 'dock',                  preferredSlot = 14, behavior = DockOrderBehavior},
 
-    DroneL = {                      helpText = "drone",             bitmapId = 'unload02',              preferredSlot = 13, behavior = DroneBehavior,               initialStateFunc = DroneInit},
-    DroneR = {                      helpText = "drone",             bitmapId = 'unload02',              preferredSlot = 13, behavior = DroneBehavior,               initialStateFunc = DroneInit},
+    DroneL = {                      helpText = "drone",             bitmapId = 'unload02',              preferredSlot = 10, behavior = DroneBehavior,               initialStateFunc = DroneInit},
+    DroneR = {                      helpText = "drone",             bitmapId = 'unload02',              preferredSlot = 11, behavior = DroneBehavior,               initialStateFunc = DroneInit},
 
     -- Unit toggle rules
     RULEUTC_ShieldToggle = {        helpText = "toggle_shield",     bitmapId = 'shield',                preferredSlot = 8,  behavior = ScriptButtonOrderBehavior,   initialStateFunc = ScriptButtonInitFunction, extraInfo = 0},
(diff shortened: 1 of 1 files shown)


lua/ui/game/orders.lua
```diff
@@ -1251,28 +1280,37 @@ local function CreateAltOrders(availableOrders, availableToggles, units)
...
-            assitingUnitList['DroneL'] = assistingUnits[1]
-        end
-        if assistingUnits[2] then
-            table.insert(availableOrders, 'DroneR')
-            assitingUnitList['DroneR'] = assistingUnits[2]
+            podUnits['DroneL'] = assistingUnits
+        end
+
+        
+        if next(assistingUnits) then
```
Garanas, line 1300: Please use `table.empty` instead of `next` to check if a table is empty

Garanas (commented):
It appears to be all good. Looked into some of the more common UI mods and none of them touch this aspect of the code. Perhaps @Hdt80bro or @4z0t could also give it a view, to be certain?

Also tested it with BlackOps. The mod pack has various drone carriers. This also works fine there.

lua/ui/game/orders.lua
```diff
@@ -608,22 +608,51 @@ end
 
 local function DroneBehavior(self, modifiers)
     if modifiers.Left then
-        SelectUnits({self._unit})
+        SelectUnits(self._unit)
     end
 
     if modifiers.Right then
+        if self._mixedIcon then
```
4z0t, line 615: why is it here?

lua/ui/game/orders.lua
```diff
@@ -1251,28 +1280,37 @@ local function CreateAltOrders(availableOrders, availableToggles, units)
     -- to determine where they go by using preferred slots
     AddAbilityButtons(standardOrdersTable, availableOrders, units)
 
-    local assitingUnitList = {}
+    local assistingUnitList = {}
     local podUnits = {}
-    if not table.empty(units) and (EntityCategoryFilterDown(categories.PODSTAGINGPLATFORM, units) or EntityCategoryFilterDown(categories.POD, units)) then
+    if not table.empty(units) and (next(EntityCategoryFilterDown(categories.PODSTAGINGPLATFORM, units)) or next(EntityCategoryFilterDown(categories.POD, units))) then
```
4z0t, line 1285: use separate named variables for readability

lua/ui/game/orders.lua
```diff
@@ -608,22 +608,51 @@ end
...
+            elseif lastMode ~= pod:IsAutoMode() then
+                mixed = true
+                break
+            end
+        end
+    end
+
+    if mixed then
+        self:SetCheck(false)
+        self._mixedIcon = Bitmap(self, UIUtil.UIFile('/game/orders-panel/question-mark_bmp.dds'))
```
4z0t, line 650: create icon and hide if not needed. no need for create/destroy

4z0t (changes requested):

lua/ui/game/orders.lua
clyfordv: This is taken straight from ScriptButtonOrderBehavior. If the unit states are mixed (and displaying a question mark) when the state is changed, destroy the question mark (because all the pods will have the same state now).

lua/ui/game/orders.lua
clyfordv: Also taken from from ScriptBitInitFunction--question mark will only be created once (when the other alt orders are) and will only be hidden/destroyed once (because the unit states can't be "unmixed" once selected). Didn't see a reason to break consistency with ScriptBit.

4z0t: I don't like copy-pasting and I'd suggest to make it as separate class to utilize. I'll approve, but would be much better to have no code dupolication.

lua/ui/game/orders.lua
```diff
@@ -1251,28 +1280,37 @@ local function CreateAltOrders(availableOrders, availableToggles, units)
     -- to determine where they go by using preferred slots
     AddAbilityButtons(standardOrdersTable, availableOrders, units)
 
-    local assitingUnitList = {}
+    local assistingUnitList = {}
     local podUnits = {}
-    if not table.empty(units) and (EntityCategoryFilterDown(categories.PODSTAGINGPLATFORM, units) or EntityCategoryFilterDown(categories.POD, units)) then
-        local PodStagingPlatforms = EntityCategoryFilterDown(categories.PODSTAGINGPLATFORM, units)
-        local Pods = EntityCategoryFilterDown(categories.POD, units)
+    local PodStagingPlatforms = EntityCategoryFilterDown(categories.PODSTAGINGPLATFORM, units)
```
4z0t, line 1285: camelCase for variables.

4z0t (changes requested):

lua/ui/game/orders.lua
Garanas: For the purpose of this change I think it is fine. If we'd introduce new files then the standard would be higher, like @clyfordv already did with #5378

PR #5371, opened 2023-08-30
Fix light particles through fog and destroying invulnerable props

Garanas:
See also:

- https://forum.faforever.com/topic/6467/no-privacy-for-mex-points

There's `CreateLightParticle` and `CreateLightParticleIntel`, where the last one takes into account intel 😄


msgerbs:
On line 222:

`local CreateLightParticle = CreateLightParticle`

This seems like this line is either not needed or should be changed to also be `CreateLightParticleIntel`.

Garanas:
You're correct, I missed that 👍

Garanas:
Fixed with: https://github.com/FAForever/fa/commit/bbbad47eecee98345bb1af417485e670e23d99b9

PR #5378, opened 2023-08-31
Sync strike

clyfordv:
Select a group of TML, hold ctrl, and click to give a launch order. Only launchers with no other orders queued will be included. TMLs between different factions don't work great when they're grouped together.


Garanas:
My first impression is that it all seems fine. Thank you for writing the pull request. We'll take this with the 4th developers iteration.

I'll give it a test once the release of the third developers iteration is done.

Garanas:
@clyfordv is this ready for testing?

clyfordv:
1. There are some additional features I've planned on adding, but the base architecture for doing so is there and not expected to change.
2. The flight time values haven't been updated since the tactical missile componentization, but don't differ significantly.

To test right now, select multiple launchers, hold ctrl, and give a launch command.

Garanas:
Let's first try to merge the current feature set before we extend on it

Garanas:
Would you mind if I make some edits / add annotations as I see fit?

lua/ui/game/syncstrike.lua
```diff
@@ -0,0 +1,221 @@
...
..., 27, 27, 27, 27, 27, 27, 27, 27, 27, 28, 29, 31, 31, 32, 32, 33, 34, 36, 37, 38, 38, 38, 39, 41, 42, 43, 43, 44, 45, 46, 47, 48, 49, 50, 50, 50, 53, 56, 57, 58, 59, 60, 61, 62, 80, 98, 99, 100, 100, 100, 102, 105, 106, 107, 107, 107, 112, 118, 118, 118, 118, 119, 119, 120, 120, 120, 121, 122, 122, 123, 122, 122, 123, 125, 126, 127, 127, 128, 129, 130, 130, 131, 131, 132, 133, 134, 135, 136, 136, 136, 137, 138, 137, 137, 138, 139, 139, 140, 140, 141, 142, 143, 144, 145, 145, 146, 146, 147, 147, 148, 148, 149, 150, 151, 152, 153, 154, 155, 155, 155, 156, 157, 157, 158, 158, 159, 160, 161, 162, 163, 163, 164, 165, 166, 167, 168, 168, 169, 169, 170, 170, 171, 172, 173, 173, 174, 175, 177, 177, 178, 178, 179, 179, 180, 181, 182, 183, 184, 185, 186, 186, 187, 188, 189, 189, 189, 190, 192, 193, 194, 194, 194, 195, 196, 196, 197, 198, 200, 200, 200, 201, 203, 203, 204, 204, 205, 206, 208, 208, 209, 210, 211, 211, 212, 213, 214, 214, 214, 215, 217, 218, 220, 219, 219, 220, 222, 222, 223, 224, 225, 226, 228, 228, 228, 229, 230, 231, 233, 233, 234, 234, 235, 236, 237, 237, 237, 238, 240, 241, 242, 242, 243, 244, 245, 245, 246, 247, 248, 248, 249, 250, 252, 252, 253, 253, 254, 255, 257, 258},
```
Garanas, line 10: Is there a reason to keep this commented entry around?

lua/ui/game/syncstrike.lua
```diff
@@ -0,0 +1,221 @@
+local AddBeatFunction = import("/lua/ui/game/gamemain.lua").AddBeatFunction
+local RemoveBeatFunction = import("/lua/ui/game/gamemain.lua").RemoveBeatFunction
+local FireState = import("/lua/game.lua").FireState
+
+--Upvalues for performance(?)
+local GameTick = GameTick
+local VDist2 = VDist2
```
Garanas, line 7: What you're doing here is changing the scope in which Lua can find the functions. The usual scope is the global scope. The Lua intrinsic `GETGLOBAL` is used to retrieve such values. Under the hood, `GETGLOBAL` is not much different from `GETTABLE` and they both involve table operations. Table operations are usually slower because of hashing (the key, name of the function) and because it is accessing 'some random memory'

By turning it into an upvalue Lua uses the intrinsic `GETUPVALUE` which is different from `GETGLOBAL` and `GETTABLE` - it is no table operation, but an O(1) lookup.

An interesting wiki entry from our neighbors:

- https://github.com/beyond-all-reason/springrts_engine_wiki_mirror/wiki/Lua_Performance

And the paper that defines Lua 5:

- https://www.lua.org/doc/jucs05.pdf

In particular chapters 4 and 5

lua/ui/game/syncstrike.lua
clyfordv: No, that's a holdover from the first round of datalogging. There's been some funny behavior around those numbers, but I'm working on a more robust datalogger (that works alongside sync strike) that will make updating them easier and more accurate.

lua/ui/game/syncstrike.lua
clyfordv: Thank you for concise explanation--this was as I suspected, but didn't want to cargo cult without stating some uncertainty.

Garanas:
@clyfordv how would this work with the new tactical missile trajectory?

clyfordv:
Load/launch times should stay the same. Initial testing suggests trajectories have a very similar flight time, so the as is implementation should be fine. However, now that the TML trajectories are deterministic, it would be best to just to rework it so the values are calculated automatically (once) instead of a using a lookup table.

Pokute:
Interesting addition that I think is a fine addition.

The TMLs get stuck if a sync strike is ordered out of range. They won't be able to fire at all.

Rarely the TMLs wouldn't fire, even with loaded missiles.

When targeting close, the sync strike isn't well synced. Firing without sync fire has the missiles impact within smaller time duration.

Ordering the launch of a sync strike when 0 missiles are loaded also makes TMLs stuck.

Garanas:
@clyfordv I'd love to include this; what is the status quo at the moment?

clyfordv:
Still need to incorporate some of Pokute's feedback and put the finishing touches on the UI (which is pretty slick if I may say so).

Primary roadblock is that some of the recent tactical missile changes have jacked up the missile trajectories (see comment here: https://github.com/FAForever/fa/pull/5525) and I can't do the flight time data until that's settled.

Garanas:
the missile changes are settled in if you ask me. At best we may increase their fire rate and reduce the damage at this point.

clyfordv:
The short range trajectories feel unfinished, where (at least for UEF, haven't looked at the others too much) the missiles tend to overshoot (the old short range trajectories were not great, pretty jerky, but we should strive to improve on them).

I feel like there's also some balance implications for, say, the flapjack effectively having its minimum range (against mobile targets) increased, along with the effectiveness of TML against close-in experimentals.

I'm happy to take a crack at some short range corrections if you'd like.

Garanas:
The minimum range increase was discussed with and accepted by the balance team. You're free to adjust some of the trajectories to make them work better up close 😃 . Ideally the arc on long range targets remains roughly the same

clyfordv:
Will do. See new UI/UX for sync strikes:

1. Launch command + hold CTRL to initialize sync strike
2. Launch command + hold CTRL + click existing sync strike to add launchers
3. Click sync strike order icon to release strike

Garanas:
Is this the final before? Once we go to the release branch there's no going back

clyfordv:
Can't confidently say that, no. There's a lot of new UI and backend machinery that needs a shakedown before I can say yes.

Garanas:
It certainly does look fancy 😃 

https://github.com/FAForever/fa/assets/15778155/817065ec-aff7-47c4-8a62-8af3a4d87961

MrRowey:
> It certainly does look fancy 😃
> 
>  sync-strike.mp4

TML stips just got a UI helping hand

MrRowey:
what was the hold up on this one ?

clyfordv:
1. Feature ready but needs testing/shakedown before being unleashed
2. Firing data tables are out of date

Garanas:
I'm also not entirely sure whether we should introduce this. The feature is nice, but it's not exactly maintainer friendly. If we edit the trajectories (which happened a lot the past year) then we also need to re-evaluate the timing tables. Which is easily forgotten 😃

Hdt80bro:
How difficult would it be to generate the timing tables at runtime?

Garanas:
You'd need to:

- (1) Launch a missile on an interval over the range of the missile launcher.
- (2) Launch multiple missiles per step of the interval.
- (3) Compute the average number of ticks that it takes to reach a target at a given step of the interval.

And then update the values in the UI, as all of this logic is in the UI at the moment 😃 !

MrRowey:
@clyfordv any ETA on this PR ?

MrRowey:
Going to close this PR in favor of a Mod instead

PR #5383, opened 2023-09-03
improved options dialog with searching options

HUSSAR-mtrela:
- improved scrolling performance in gird control
- improved default padding in tooltips
- ignoring DEV files in root of the repo
- fixed UI layout/clipping issues in options dialog
- replaced tabs with collapsible list of options (see screenshot 1)
- add missing tooltips for a few options
- add searching/filtering of all options by their names or values (see screenshot 2 and 3)
- add possibility of creating new group of options, e.g. mods can hook the [lua/options/options.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/options/options.lua#L167) file and add new group of options inside of the [options](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/options/options.lua#L167) table such as [my_mod_options](https://github.com/FAForever/fa/blob/hussar-game-options/lua/ui/dialogs/options.lua#L262) (see screenshot 4)

Screenshot 1 - the new options dialog with collapsible list of options:
image

Screenshot 2 - the new options dialog with searching by option's name:
image

Screenshot 3 - the new options dialog with searching by option's value:
image

Screenshot 4 - the new options dialog with custom mod options:
image

Screenshot 5 - the old options dialog:
image


lua/maui/grid.lua
```diff
@@ -252,19 +252,26 @@ Grid = ClassUI(Group) {
...
+                    if  (col >= scrollHorz) and (col < scrollHorz + self._visible["Horz"]()) and
+                        (row >= scrollVert) and (row < scrollVert + self._visible["Vert"]()) then
                         control:SetHidden(false)
                         local column = col
                         local rowumn = row
                         local horzPad = math.max(0, (self._itemWidth - control.Width()) / 2)
                         local vertPad = math.max(0, (self._itemHeight - control.Height()) / 2)
-                        control.Left:Set(function() return math.floor(((column - self._top["Horz"]) * self._itemWidth) + self.Left() + horzPad) end)
-                        control.Top:Set(function() return math.floor(((rowumn - self._top["Vert"]) * self._itemHeight) + self.Top() + vertPad) end)
+                        control.Left:Set(function() return math.floor(((column - scrollHorz) * self._itemWidth) + self.Left() + horzPad) end)
```
Garanas, line 273: These closures are likely the most expensive part of this function, not that you have to fix that but just pointing it out

Garanas (commented):
Looking great! It appears we're not 'properly' using the background texture, and instead apply black squares on top of it.

image

Would it be possible to properly use the background texture, put the search at the top and shorten the list with 2 or 3 elements?

Garanas:
Looking at the current options menu:

image

@HUSSAR-mtrela would it be possible to use the texture space accordingly by having the search + game options in the background area that is designed to hold the options? I don't know by heart what the dialog texture looks like exactly, I'll try to find it

HUSSAR-mtrela:
@Garanas the background texture has semi-transparent region where tab controls used to be and the same semi-transparent region above Reset/OK/Cancel/Apply buttons (see 2nd screenshot below) that is why I added that black bitmap to fill those regions. Otherwise search box would not be very visible and there would be a lot of wasted space that we can now use for displaying options. I have just fixed the alignment of the black fill with the background texture (see my screenshot below)

Latest Option Dialog with fixed alignment of fill bitmap:
image

Original Texture without tabs and options list:
image

By they way, I added an option to open this dialog via key binding
image

HUSSAR-mtrela:
@Garanas If you like I can replace the styling of the Option window (which does not changes based on selected faction) with styling of the Keybinding window which matches currently selected faction.

BlackYps:
I have to say, visually it looks worse than what we had before. Was there a reason to replace the scroll bar and the borders around the individual options?
It also looks like you tried to squeeze every inch of vertical space out of it. The black area now almost touches the buttons, which doesn't look good

MrRowey:
@HUSSAR-mtrela this can be updated to use the tab option as I agree with Blackyps that it dose not look as visually good.

MrRowey:
Closing for now Can be reopened if updated with the if updated with the feedback

PR #5385, opened 2023-09-03
Implement alternative of `IssueBuildMobile`

Garanas:
Implements the changes of https://github.com/FAForever/FA-Binary-Patches/pull/29 and applies it to the Ring Extractor, Distribute Orders and Copy Orders features.

The assembly change fixes the performance problem where each build order would create a separate built preview. Built previews are transparent and therefore expensive on your framerate. With the assembly change we can assign 1 build order to a group of engineers, instead of assigning a separate build order to each engineer.

### DIFF
--- a/engine/Sim.lua
+++ b/engine/Sim.lua
@@ -748,14 +748,24 @@ end
 function IssueBuildFactory(units, blueprintID, count)
 end
 
---- Orders a group of units to build a unit, each unit is assigned the closest building.
---- Takes some time to process (at least 3 ticks).
+--- Orders a group of units to build a unit, the nearest unit is given the order
+--- Takes some time to apply (at least 3 ticks).
 ---@param units Unit[]
 ---@param position Vector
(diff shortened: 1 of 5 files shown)


engine/Sim.lua
```diff
@@ -758,6 +758,16 @@ end
 function IssueBuildMobile(units, position, blueprintID, table)
 end
 
+--- Orders a group of units to start building another unit by blueprint id
+--- Takes some time to apply (at least 3 ticks).
+---@param units Unit[]
+---@param position Vector
+---@param blueprintID string
+---@param table number[] # A list of alternative build locations, similar to AiBrain.BuildStructure. Doesn't appear to function properly
+---@return SimCommand
```
4z0t, line 767: they all return nil

Garanas: Included in https://github.com/FAForever/fa/pull/5386

PR #5386, opened 2023-09-03
Implement the ringing behavior for fabricators

Garanas:
Implements the ringing of extractors with fabricators. Uses the same old logic as before, therefore players will not have to learn anything new. This is now safe for performance with thanks to https://github.com/FAForever/FA-Binary-Patches/pull/29

There's two different ways to ring extractors:

_Ringing only the inner fabricators, those that have the most adjacency_

image

_Ring with all fabricators that have some adjacency_

image

You can choose which through the game options:

image


4z0t:
Can you just revert other ringing stuff then too?

4z0t (changes requested):
Return as it was before.

Garanas:
Please stick to what this pull request is about. We can discuss whether the rest should return through Discord

PR #5396, opened 2023-09-05
Change depth indicator to multiple dots again and update description

gordenwunderlich:
new look
image
First two dots are at fixed depth (0.5 and 1.5) up to 3 dots appear afterwards and their distance scales based on water depth.
New description
image

### DIFF
--- a/lua/ui/game/cursor/depth.lua
+++ b/lua/ui/game/cursor/depth.lua
@@ -6,7 +6,8 @@ local meshSphere = '/env/Common/Props/sphere_lod0.scm'
 
 local MeshOnTerrain = nil
 local MeshOnTerrainRed = nil
-local MeshInBetween = nil
+local MeshesInBetween = { }
+local MaxMeshesInbetweenCount = 5
 local MeshFadeDistance = 300
 
 local Trash = TrashBag()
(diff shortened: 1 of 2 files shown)


Garanas:
Can you add that it is an estimated guess and not a ground truth to the description?

gordenwunderlich:
The thing is, The ground truth it's showing, is that where it's red you definitely 100% can't go.
This does of course not mean that you can go everywhere else, but I'm struggling to fit this into a description that's not 4 lines longer than the current one. Then nobody's gonna bother to read it or won't understand what it's trying to say.
Do you maybe have an idea?

Garanas:
That it may contain [type I errors](https://en.wikipedia.org/wiki/Type_I_and_type_II_errors) 😄 

What about:

```
When enabled, adds visual elements to visualize the offset between the water surface and the ocean floor at the position of the cursor. A separate indicator turns red when naval units are guaranteed to not be able to path at the location due to lack of depth.
```

gordenwunderlich:
We can take that. Conveys the same message as mine, while ominously hinting at the "guaranteed". Not sure if people will catch on to what that means, but at least they have a chance.

PR #5401, opened 2023-09-06
Lobby: Increase the number of options for setting the UnitCap

Basilisk3:
Adds the options to increase the UnitCap to:
- 1750
- 2000

In some games, 1500 just isn't enough.

### DIFF
--- a/lua/ui/lobby/lobbyOptions.lua
+++ b/lua/ui/lobby/lobbyOptions.lua
@@ -31,7 +31,7 @@
 ---@field Share 'FullShare' | 'ShareUntilDeath' | 'PartialShare' | 'TransferToKiller' | 'Defectors' | 'CivilianDeserter'
 ---@field ShareUnitCap 'none' | 'allies' | 'all'
 ---@field Timeouts '0' | '3'| '-1'
----@field UnitCap '125' | '250' | '375' | '500' | '625' | '750' | '875' | '1000' | '1250' | '1500'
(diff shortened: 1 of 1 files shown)


Garanas:
Even though you're correct that this is technically possible I would not encourage it. It is like when FAForever introduced being able to play up to 8v8 - yes, it is 'possible' but the game just breaks quite quickly if you'd actually do it. The same applies to this change - if everyone in a 6v6 makes 2000 units then you're quickly done playing.

Instead, what if we make it the default to share the unit cap with allies when a player dies? That already exists in the game and we wouldn't push the total number of units then one can theoretically make at this point in a single game.

Basilisk3:
I disagree with the points you make. The incentive to make lots of units actually decreases the more players you have. There is essentially no reason to ever make 1500 or more units in a 6v6, unless it is with full share, which is played with a shared unit cap anyway. From the replays I have seen, most players do not even come close to the standard unit cap of 1000 in these games. 

This change is not going to interfere with the regular game in any way in over 99% of cases. This option is specifically for 2v2s through 4v4s, on large maps with lots of reclaim and expansions, or modded games. Examples could be Seton's Clutch or The Ditch.

Also slowing down the game on purpose, for example by making 2000 Labs, is bannable anyway. If a player is worried about someone possibly abusing it, it can just be left at the default settings, which are lower.

Garanas:
See the comment in https://github.com/FAForever/fa/pull/5402#issuecomment-1722434818

PR #5402, opened 2023-09-06
Lobby: Change the default UnitCap to 1500

Basilisk3:
With the game running much faster now, I think we can finally increase the default UnitCap to 1500. A default UnitCap of 1000 can cause problems, especially when the host forgets to change it to 1500 on large maps. It is an artificial limitation, that is not required to this extent anymore. If a lower UnitCap is desired, it can still be selected in the settings.

### DIFF
--- a/lua/ui/lobby/lobbyOptions.lua
+++ b/lua/ui/lobby/lobbyOptions.lua
@@ -268,7 +268,7 @@ globalOpts = {
         },
     },
     {
-        default = 8,
+        default = 10,
         label = "<LOC lobui_0102>Unit Cap",
         help = "<LOC lobui_0103>Set the maximum number of units that can be in play",
         key = 'UnitCap',


Garanas:
With a similar reasoning in https://github.com/FAForever/fa/pull/5401 - I do not agree with this change. Yes, you're correct but in practice 1500 units per player is already too much. This is the maximum, and by default we should move far away from the theoretical boundary and encourage players to have roughly a thousand units each in total.

As an example, I have a Ryzen 5 3600. This is a 'budget' CPU, and often not even available anymore. With the most recent patch I can run between 6000 to 7000 units before slowing down. If I'd play a 4v4 and the default settings (1000 units per player) then that is 8000 units in total. That is already too much for my CPU to handle, let when each player can make up to 1500 units (12.000 total).

Basilisk3:
I also disagree with the points you make here. Just because the option to make more units exists, does not mean it has to be used. Increasing the unit cap does not automatically result in players altering their playstyle to make more units, it just removes the cap on playing the game regularly, until a certain point. This is about removing artificial limitations that, in this day and age, are just a nuisance and can, and will, break games. 

Essentially, this aims to save the couple percent of interesting, longer games with which a low unit cap would have interfered with otherwise. If the game slows down, because the players play for a long time, and there are a lot of units on the field, so be it. This change does not alter a player's behavior, until the default unit cap would have been reached. There are plenty of maps, where hundreds, or even thousands of engineers are built, just to reclaim. There is a reason maps like Seton's Clutch are always played with a cap of 1500 (or even more when an allied player dies), though here it is due to air units.

I have even seen 1v1 games, which are hindered by a unit cap of 1000. It is why it was changed, (or at least attempted to be changed) for the 1v1 matchmaker in #3293. I have also seen tournament games being hosted with a unit cap of 1000, because the host forgot to change the settings. Yes, this is the fault of the host, but since it is the default setting, this is bound to happen.

At the very least, the default should be 1250. I would never want to play a game with a unit cap as low as 1000. Yes, it will usually be fine, but in the 10% of games where it isn't, it is a real handicap and very annoying.

Garanas:
I understand your sentiment, but I don't think this is the correct approach. The same applies to https://github.com/FAForever/fa/pull/5401 . I'm going to close both of them, meanwhile I'll work on an alternative that better represents the limitations while also giving people more control over the unit cap and the consequences of increasing it.

Last year August we made these changes to track the performance of a player based on the current sim rate and compare it to the current unit count.

- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/system/performance.lua

At the time I got stuck on some lobby code issues. Last week because of your PRs I picked it up again and at least introduced a visualizer of the data:

- https://github.com/FAForever/fa/pull/5427

image

The vertical axes is the number of units. The horizontal axis is the game speed. Note that there's a bug: the horizontal axis is reverted.

The idea is that instead of defining a unit cap per player, you instead define a unit cap for each team. Everyone in the team then consumes from the same unit cap. As a result it is easy to understand the theoretical maximum: if the unit cap per team is 3000 and there are two teams then in the worst case the game will have 6000 units. We can represent that visually by creating a horizontal line in the graph. Inspecting the graph can help the host and other players in the lobby understand whether that specific player will slow down the game when it reaches its thresholds.

Speaking of the lobby, ideally the CPU score is entirely abandoned and we instead introduce a color:

- Green if the player is well above +0 at the theoretical maximum
- Yellow if the player is at +0 or +1 at the theoretical maximum
- Orange if the player is at -1 or -2 at the theoretical maximum
- Red if the player is below -2 at the theoretical maximum

By hovering over the color you can see the graph of that player and based on that information you can decide to acknowledge or ignore the color.

Basilisk3:
@Garanas Making the unit cap shared between all members of the team does sound like a good idea, but I would have to try it in real games at least a couple of times, to find possible issues with it. If it is set too low, there could certainly be conflicts sparked because of this, for example, when one player uses up most of the cap for whatever reason. As outlined in my previous posts, I do think that the default unit cap is just too low in some scenarios. So if this change helps to alleviate this issue a bit, then I am all for it.

One potential problem I can see in #5427, is that some unit types eat up much more performance than others. Hover tanks that constantly bump into each other are going to be more taxing than regular units. This would mean that, for example on naval maps, the performance is probably going to be worse than indicated by the UI. Especially on Seton's, there can be a huge amount of hover-spam.

Garanas:
> One potential problem I can see in #5427, is that some unit types eat up much more performance than others. Hover tanks that constantly bump into each other are going to be more taxing than regular units. This would mean that, for example on naval maps, the performance is probably going to be worse than indicated by the UI. Especially on Seton's, there can be a huge amount of hover-spam.

Yes, it is an approximation. But based on my own observations usually one composition of 4000 units is not far off from another composition of 4000 units. It's not perfect, but a lot better than the CPU score now 😃 

We do keep track of a separate score for three types of scenario's:

- (1) Campaign content
- (2) Games with AIs
- (3) Regular, PvP games

Content of types (1) and (2) will weigh down the simulation a lot more than (3)

Issue #5280, opened 2023-07-27
Mods Manager is overlapping mod info over the search box


HUSSAR-mtrela:
**Describe the bug**
Mods Manager is overlapping mod info over the search box in **some window heights** 

**To Reproduce**
- Open Mods Manager
- Change to compact mod list
- Observer search box

NOTE if this does not occur when testing, close Mods Manager and increase height of FA window by 50 pixels and then repeat above instructions.

**Expected behavior**
Mods Manager is not overlapping mod info over the search box. 

**Screenshots**
![image](https://github.com/FAForever/fa/assets/11665978/b1d0fd7a-41e5-4059-81c7-a9652aa26c50)

HUSSAR-mtrela:
I will work on this issue this week

Garanas:
There's also a similar problem with the description of mods being too long. Please make sure you have some UI scaling enabled (at 125%, for example) to also test the UI scaling aspect of it

lL1l1:
Fixed by #5405

PR #5405, opened 2023-09-07
fixed layout in mods manager and links dialog

HUSSAR-mtrela:
fixed default parent for the links dialog
fixed layout and text clipping in mods manager  (issue #5280)
fixed tooltips in mods manager
replaced custom scrolling control with existing Grid control in mods manager (less code but better performance)


Garanas:
These are great improvements @HUSSAR-mtrela , really fan of the tooltip when hovering over a mod.

image

image

PR #5407, opened 2023-09-08
Add `GetPositionsInRadius` and generate naval markers

Garanas:
https://github.com/FAForever/fa/assets/15778155/f07051ef-021d-4ca6-94f5-edbfbacebc4e


Garanas:
Basic generation of naval area's. May contain duplicates. The number of naval area's is also quite high, which is likely not desirable 😄 

image

image

image

image

image

image

BlackYps:
What's a naval area?

BlackYps:
the locations actually look decent, but it shouldn't try to fill them all with factories. If it just takes them as possibilities then it looks fine

relent0r (approved):
I did testing on this for both the positions in radius and it worked as expected. The size parameter is a great addition and allows testing for positions depending on the size of a naval unit that is being used (great for frigates vs battleships for example).

Also tested the naval marker generation on a map with no markers. Confirmed that AI would detect the naval markers and build a naval base happily.

The default AI may need some adjusting. Right now it looks to use the following for setting its BaseType property.
Scenario.MasterChain._MASTERCHAIN_.Markers[baseName].type

I believe you overwrite the marker cache rather than that scenario table.

I didn't actually notice this until now so possibly this has caused other issues int he past for things that leverage that basetype property (but I never noticed any impact)

Garanas:
> The default AI may need some adjusting. Right now it looks to use the following for setting its BaseType property. Scenario.MasterChain._MASTERCHAIN_.Markers[baseName].type
> 
> I believe you overwrite the marker cache rather than that scenario table.

Definitely something that we should fix 👍

Sc0lex:
@Garanas Did you happen to try this with Four-Corners? I noticed that the AI stopped building shipyards (and various T3+) on that map within the last few weeks. Same AI settings on Sludge and AI builds as expected. Attached log for reference. I see 'debug: Overwriting 0 markers of type Naval Area in cache!' for all brains on Four-Corners, but the # of markers overwritten is non-zero on Sludge.
[game_20858811.log](https://github.com/FAForever/fa/files/12733122/game_20858811.log)

Garanas:
That sounds like a bug 👍 , I'll see what I can find

relent0r:
I double checked this and indeed no naval area markers are being generated on this particular map.

Issue #2854, opened 2019-09-28
make unit counter number more visible


shalkya:
![image](https://user-images.githubusercontent.com/11618427/65817534-8033b980-e208-11e9-8d0a-a21fcfacb697.png)
either by changing the color or by adding contrast. It's hard to read when background is white (usually air units)

HUSSAR-mtrela:
this issue is fixed in above PR.
![image](https://github.com/FAForever/fa/assets/11665978/091d13df-d5ba-4ea4-9498-35df0a7a8f05)

4z0t:
https://github.com/4z0t/FAF-UI-Mods/pull/36

PR #5413, opened 2023-09-10
add semitransparent background for unit counter

HUSSAR-mtrela:
Adds semitransparent background for unit counter to improve visibility of displaying text which fixes  issue #2854

The new SetColorShadow() works better than internal SetDropShadow() for displaying text, e.g. white unit count over bright blue background of air units

Units counter with new semitransparent background:
image

this work similar to existing semitransparent background used in displaying keybinding in construction menu:
image

Note above screenshots where taken with a mod that increase size of strategic icons so this is why they appear to overlap with unit counter. @Garanas should we add an option to configure placement of strategic icons on left/right side of the unit icon? Also, I could add options to configure color of unit counter and background if you want?

### DIFF
--- a/lua/maui/text.lua
+++ b/lua/maui/text.lua
@@ -91,6 +91,13 @@ Text = ClassUI(moho.text_methods, Control) {
         end
     end,
 
+    --- creates semitransprent shadow behind this text control with optional styling parameters
+    --- this improves visibility of the displayed tex and it works better than SetDropShadow(true) function
+    SetColorShadow = function(self, color, left, top, right, bottom, targetDepth, shadowDepth)
+        self.Shadow = import('/lua/ui/uiutil.lua').CreateBackground(self, color, left, top, right, bottom, targetDepth, shadowDepth)
+        return self.Shadow
+    end,
+
     OnDestroy = function(self)
         if self._font then
             if self._font._family then
--- a/lua/ui/uiutil.lua
+++ b/lua/ui/uiutil.lua
@@ -552,6 +552,29 @@ function CreateText(parent, label, pointSize, font, dropshadow)
     return text
 end
 
+--- Returns background bitmap behind specified target control with optional styling parameters
+--- for example creating semitransparent background behind text control or an icon
+---@param target Control
+---@param color? LazyVarColor
+---@param left? number
+---@param top? number
+---@param right? number
+---@param bottom? number
+---@param targetDepth? number
+---@param bitmapDepth? number
+---@return Bitmap
+function CreateBackground(target, color, left, top, right, bottom, targetDepth, bitmapDepth)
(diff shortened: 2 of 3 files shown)


Garanas:
> @Garanas should we add an option to configure placement of strategic icons on left/right side of the unit icon? Also, I could add options to configure color of unit counter and background if you want?

I don't see the need to customize it further, the overlap is minimal. It'd be better I think if we draw the strategic icon on top of the background

HUSSAR-mtrela:
@Garanas I checked implementation of construction.lua and there are a lot of duplicated code and low performance code so I will make a few other improvements in this PR. Please do not merge it for now.

Garanas:
@HUSSAR-mtrela could you elaborate as to what you'd like to improve, and how you'd want to do that?

HUSSAR-mtrela:
@Garanas I'm working on this kind of improvements:
- move logic for getting icons and basic info about units to utility functions for reuse by other UI
- improved performance by caching common data that is generated for each unit and build option
- use classes to encapsulate some UI elements of construction.lua so we can re-use in avatar.lua
- add a few options to configure appearance of units in construction menu

above improvements should help us later do overhaul of the whole construction like you are planning with the lobby.

Garanas:
@HUSSAR-mtrela  Your work may have a conflict with https://github.com/FAForever/fa/pull/5484

lua/maui/text.lua
```diff
@@ -91,6 +91,13 @@ Text = ClassUI(moho.text_methods, Control) {
         end
     end,
 
+    --- creates semitransprent shadow behind this text control with optional styling parameters
+    --- this improves visibility of the displayed tex and it works better than SetDropShadow(true) function
+    SetColorShadow = function(self, color, left, top, right, bottom, targetDepth, shadowDepth)
```
4z0t, line 96: It doesn't make sense to present here.
Create an inherited class with background behind the text.

lua/ui/uiutil.lua
```diff
@@ -552,6 +552,29 @@ function CreateText(parent, label, pointSize, font, dropshadow)
...
+---@param target Control
+---@param color? LazyVarColor
+---@param left? number
+---@param top? number
+---@param right? number
+---@param bottom? number
+---@param targetDepth? number
+---@param bitmapDepth? number
+---@return Bitmap
+function CreateBackground(target, color, left, top, right, bottom, targetDepth, bitmapDepth)
```
4z0t, line 566: Create a class. Use Layouter.

4z0t (changes requested):
Don't stick to old style.

MrRowey:
Closing PR as it is not currly necessary and change isn't flowing new style as well as not being active for almost 10 months

PR #5422, opened 2023-09-10
Resource deposit grid for AIs

### LINKED ISSUE #5226: AI - Grid based mass point search
**Describe the feature**
Currently if the AI wants to find a mass point within a certain range it needs to search all the masspoints on the map and check the distances to find the ones it needs.
This means on a 20km map with say 200 mass markers in order to find one that is 60 units away you must loop through all 200 even though you know 90% will be too far away.

The desire would be to be able to search by grids and determine mass points that way, it would cut down the number of mass markers returned by a factor of 10 on large maps and allow the AI to perform checks in order to redirect units without having to loop through all mass markers on the map.

for example

GetMassMarkersInGridRadius(movementLayer, position, radius, status)

movementLayer = ignore under water maspoints if the movement layer is land and vice versa.
position = starting position
radius = radius to look around in units
status = All, Ally, Enemy, allow the return of only mass points within the enemy side of the map or the ally side depending on map control.

The function would search grid squares from the position moving out until the the grid squares distance from the unit beyond beyond the defined radius, it could then return just those mass markers. For large maps with high mass point counts this would decrease the compute time for many queries the AI makes for markers.

Garanas:
Closes #5226 

Introduces the resource deposit grid with two queries:

- `GetResourcesWithinDistance` Returns all deposits that are within a given pathing distance to the start location
- `GetResourcesWithinRadius` Returns all deposits that are within a given radius, ignoring other conditions

Visual examples of `GetResourcesWithinDistance`:

https://github.com/FAForever/fa/assets/15778155/18e5194f-93c1-4433-9769-2463f2d37ace

https://github.com/FAForever/fa/assets/15778155/52fdc48c-e0d2-41c6-aab0-9e6895f9bc14

### DIFF
--- a/lua/sim/NavUtils.lua
+++ b/lua/sim/NavUtils.lua
@@ -1295,3 +1295,11 @@ end
 function IsInBuildableArea(origin)
     return IsInPlayableArea(origin, 8)
 end
+
+--- Converts a world distance into grid distance
+---@param distance number
+---@return number
+function WorldDistanceToGridDistance(distance)
+    local sizeOfCell = NavGenerator.SizeOfCell()
+    return math.ceil(distance / sizeOfCell) + 1
+end
--- a/lua/sim/NavUtils.lua
+++ b/lua/sim/NavUtils.lua
@@ -1299,7 +1299,7 @@ end
 --- Converts a world distance into grid distance
 ---@param distance number
 ---@return number
-function WorldDistanceToGridDistance(distance)
+function ToGridDistance(distance)
     local sizeOfCell = NavGenerator.SizeOfCell()
-    return math.ceil(distance / sizeOfCell) + 1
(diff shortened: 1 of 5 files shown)


lua/sim/NavUtils.lua
```diff
@@ -631,8 +639,7 @@ function GetPositionsInRadius(layer, position, thresholdDistance, thresholdSize,
         return nil, 'OutsideMap'
     end
 
-    local sizeOfcell = NavGenerator.SizeOfCell()
-    local distanceInCells = math.ceil(0.5 * thresholdDistance / sizeOfcell) + 1
+    local distanceInCells = ToGridDistance(thresholdDistance)
```
Garanas, line 642: @relent0r this is the change, see if it works better for you now

PR #5423, opened 2023-09-10
Decrease the MaxAirspeed of all T2 Transports from 14.3 to 13.5

### LINKED ISSUE #5414: Balance idea: T2 transports
## Intended changes

T2 transports (all): movement speed: 14.3 -> 13.5

## Thoughts on the changes

Nerf primarily intended for arty drops

Basilisk3:
Closes #5414
Related to #5212


Basilisk3:
In my opinion, T2 Bombers need to be slightly slower than Interceptors as well.

Garanas:
Meets the requirements of the issue, no need to keep this lingering

PR #5428, opened 2023-09-11
Tactical Missile Component Rework

clyfordv:
Parameterizes and componentizes tactical missiles. Currently limited to the following projectiles for testing:
Aloha
Flapjack
Spearhead

Flight parameters are now:

- `LaunchTicks`: how many ticks we'll spend in a low-turn-rate launch phase. 2 for mobile launchers, 6-8 for vertical launched projectiles.
- `LaunchTurnRate`: turn rate in aforementioned phase. Set to a low value (6-8) to prevent jerkiness as soon as the projectile is created.
- `HeightDistanceFactor`: Range to the target will be divided by this number when calculating the maximum height of the trajectory. Set to a higher number for a flatter flight path. ~5 is a decent number.
- `MinHeight`: If the calculated highest point of the trajectory is lower than this number, it will be bumped up to reach this altitude above the launch point. Prevents low angle trajectory weirdness. Minimum range/2 is a decent starting point.
- `FinalBoostAngle`: Angle in degrees the missile will aim to be at when it reaches the highest point of its trajectory. 0 degrees (horizontal) is a reasonable choice, can be tweaked to achieve a more parabolic-esque flight path.

The new SemiBallisticComponent in DefaultProjectiles.lua is not terribly well optimized yet in terms of redundant information gathering function calls (target:GetPosition, projectile:GetVelocity, etc.). This can be addressed with an update function at the beginning of each flight phase, when the time comes.

### DIFF
--- a/lua/sim/DefaultProjectiles.lua
+++ b/lua/sim/DefaultProjectiles.lua
@@ -296,8 +296,10 @@ TacticalMissileComponent = ClassSimple(SemiBallisticComponent) {
 
         -- wait until we've allegedly hit our target, then turn tracking off
         -- (in case we miss, so we don't fly in circles forever)
-        WaitTicks(glideTime * 10)
-        self:TrackTarget(false)
+        WaitTicks((glideTime+1) * 10)
+        if not self:BeenDestroyed() then
+            self:TrackTarget(false)
+        end
     end,
 
 }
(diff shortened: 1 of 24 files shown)


Garanas:
This is looking very promising 👍

clyfordv:
All missiles for all factions updated and tested.

PR #5433, opened 2023-09-12
Tactical missile component rework

clyfordv:
Parameterizes and componentizes tactical missiles. All tactical missile projectiles for all factions are updated.

Flight parameters are now:

- `LaunchTicks` how many ticks we'll spend in a low-turn-rate launch phase. 2 for mobile launchers, 6-8 for vertical launched projectiles.
- `LaunchTurnRate` turn rate in aforementioned phase. Set to a low value (6-8) to prevent jerkiness as soon as the projectile is created.
- `HeightDistanceFactor` Range to the target will be divided by this number when calculating the maximum height of the trajectory. Set to a higher number for a flatter flight path. ~5 is a decent number.
- `MinHeight` If the calculated highest point of the trajectory is lower than this number, it will be bumped up to reach this altitude above the launch point. Prevents low angle trajectory weirdness. Minimum range/2 is a decent starting point.
- `FinalBoostAngle` Angle in degrees the missile will aim to be at when it reaches the highest point of its trajectory. 0 degrees (horizontal) is a reasonable choice, can be tweaked to achieve a more parabolic-esque flight path.

The new SemiBallisticComponent is not terribly well optimized yet in terms of redundant information gathering function calls (target:GetPosition, projectile:GetVelocity, etc.). This can be addressed with an update function at the beginning of each flight phase, when the time comes.


Garanas:
Would you mind if I add in some of the missing annotations?

Garanas:
And that are the annotations 😄 , it is looking great. I'll review it again tomorrow and I'll try to contact the balance team about this.

Garanas:
For clarity, this also fixes the bug reported by Deribus where tactical missiles can fly off into nowhere as mobile missile launchers move around. The missile may still take a long trajectory, but it always ends up at the target

Garanas:
@clyfordv without doing any dev work on this question, how would you imagine we mix and match the changes in this pull request with these changes:

- https://github.com/FAForever/fa/pull/5369

clyfordv:
> how would you imagine we mix and match the changes in #5369 

I believe TrackTarget, TargetPosition, etc. are target/target ground agnostic, so the component should (should) take whatever OnTrackTargetGround outputs without any modification. MoveThread disables target tracking just after impact time to prevent circling (now added to zig zaggers).

Garanas:
Great work @clyfordv , can't wait to release this in November 😃

PR #5435, opened 2023-09-14
Draft implementation to test `GetFractionComplete`

Garanas:
Requires the following binary patch: https://github.com/FAForever/FA-Binary-Patches/pull/40

The UI can now retrieve the build fraction that a unit that is being constructed is at. Previously the fraction was based on the hit points of the unit. This works, as long as the unit is not being damaged during construction.

Separately, also updates the work progress of the unit in question with the build fraction. This adds another layer of information to the player.

image

### DIFF
--- a/lua/defaultunits.lua
+++ b/lua/defaultunits.lua
@@ -210,6 +210,15 @@ StructureUnit = ClassUnit(Unit) {
             self:RotateTowardsEnemy()
         end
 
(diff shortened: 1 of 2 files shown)


Garanas:
@4z0t is the assembly change live on `deploy/fafdevelop`?

4z0t:
No idea.

PR #5436, opened 2023-09-15
Table allocation free versions of sim commands

Garanas:
This idea started with https://github.com/FAForever/fa/issues/5398 where we introduce alternatives to global functions that work for just a single unit, instead of a group of units. The goal is to prevent the allocation of a table with a single array entry in it when we need to apply the function to a single unit.

We can apply the same approach to the simulation. We introduce alternatives that work on a single unit and re-use a single table to efficiently manage memory without the developer being aware of it.

### DIFF
--- a/engine/Sim.lua
+++ b/engine/Sim.lua
@@ -773,7 +773,7 @@ end
 function IssueCapture(units, target)
 end
 
---- Clears out all commands issued on the group of units, this happens immediately
+--- Clears out all commands issued on the group of units, this happens immediately. See `IssueToUnitClearCommands` when you want to computationally efficiently apply it to a single unit 
 ---@param units Unit[]
 ---@return SimCommand
 function IssueClearCommands(units)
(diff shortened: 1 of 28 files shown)


Garanas:
A quick-and-dirty benchmark:

```lua
local start = GetSystemTimeSecondsOnlyForProfileUse()

for k = 1, 1000 do
    local arr = {self}
end

local diff = (GetSystemTimeSecondsOnlyForProfileUse() - start) * 1000
LOG("New table: " .. tostring(diff)) -- 0.12 ms

local start = GetSystemTimeSecondsOnlyForProfileUse()

local cached = { } 
for k = 1, 1000 do
    cached[1] = self
end

local diff = (GetSystemTimeSecondsOnlyForProfileUse() - start) * 1000
LOG("Reused table: " .. tostring(diff)) -- 0.013 ms
```

Where `self` is a unit. See also: https://github.com/FAForever/fa/pull/5436/commits/9c54351353c382714518645bdd6b168770be177f#diff-74bdc025a96f1e8f643036fed35e40e09ad92bcb5d5ac51af065e1ad7f63d25bR298-R317

Garanas:
@Hdt80bro would you have time to go over the changes, see if I did a hiccup anywhere?

PR #5438, opened 2023-09-15
Cycle templates hotbuild

gordenwunderlich:
Just gonna put this out here since changing the default hotkey layouts takes more time than I thought.

So this simply adds a hotkey which does the same thing as the hotbuild template key (G), just replacing the default template feature with the new cycle template feature, so the new feature can seamlessly replace the old templates without abandoning the hotbuild key.
image

Currently the cycle templates function only has a text preview and not an image preview like the old hotbuild templates key, but that can be added in the future


Garanas:
Anyone who uses the Hotbuild feature that can give their opinion on this implementation? 😃

PR #5442, opened 2023-09-16
Fix small details of the Trident Class

Garanas:
Adds better targeting bones that are well within the collision box of the unit.


Garanas:
https://github.com/FAForever/fa/assets/15778155/79eff4c2-835b-44bd-9df1-5d5b8ebb1593

PR #5454, opened 2023-09-18
Fix various issues with SMLs and TMLs

### LINKED ISSUE #5211: Aeon/Cybran T3 Strategic Missile Submarine & Sera T3 Battleship Nuke Launch Orders Queue
Both the Aeon & Cybran T3 Strategic Missile Submarine (uas0304 / urs0304) as well as the Sera T3 Battleship (xss0302) will not process any additional orders via Shift after being given a launch order, meaning that they'll fire one nuke & then not move to or fire at a different location. The UEF T3 sub nor any of the nuke launch structures have this issue. This behavior is the same in FAF & FAF Beta.

FAF Beta:
https://replay.faforever.com/20346455
[game_20346455.log](https://github.com/FAForever/fa/files/12001913/game_20346455.log)

FAF:
https://replay.faforever.com/20346513
[game_20346513.log](https://github.com/FAForever/fa/files/12001919/game_20346513.log)

In a sandbox on a large water map such as Gentleman's Reef, spawn the 3 nuke subs & the Sera battleship individually in each of the corners as well as a Paragon by your ACU (you're not going to nuke that, right?). Select the 3 subs & the battleship then right-click on the enable missile auto build order. Next, select one at a time & give it multiple launch orders or a combo of launch & move orders.

If you observe the UEF sub, after it fires its nuke, you'll see it move to a proper location to fire the next nuke. After that nuke is launched, then it will move into position to fire the next nuke & so on. This is the expected behavior as when you give the launch structures multiple launch orders, it will fire a nuke at one location, then at the next, then at each following location until it's fired at all the targets you gave.

When observing the Aeon/Cybran subs & Sera battleship, you'll see them move into position to fire the 1st nuke & then fire when ready, but then they'll just sit there as if the 1st nuke hasn't been launched yet, but they'll also not fire at the same location again as if they have.

Garanas:
Closes #5211 

This bug took a long time to figure out. It was reported at:

- https://forum.faforever.com/topic/6515/sera-mmls-are-broken
- https://discord.com/channels/197033481883222026/1144676891018666088

There's actually two bugs.

The first bug is that the condition that we changed in `lua/sim/defaultweapons.lua` does not check whether the unit was actually holding fire to begin with. All units with counted projectiles or with a reload animation were able to effectively skip that bit of logic at all moments. As a result they would be able to fire again, immediately one after another!

The second bug was because of the (long) `RackSalvoReloadTime` and the unit thinking it lost its target. As a result the unit would 'skip' a firing cycle and it would fire twice as slow as intended!

The third bug is related to (un)packing animations and dsitributing orders. Due to the unpacking animation you'd end up in the 'wrong weapon state' after clearing the command queue. There's now an exception where if you try to fire during the packing state while the weapon still thinks it is unpacking then it continues unpacking. This happens when you clear the orders and then immediately queue a new launch order. As a result all TMLs and all SMLs can now be distributed more reliably

And last not but least - there was a bug with manual fire weapons that immediate finish their weapon cycle being unable to proceed in the command queue. The units would remain stuck on the launch order. See the video below. As a result any other after the launch order would effectively be ignored.

All manual fire weapons should have _some_ delay in them, regardless of what type of delay that is. For convenience I choose a charge delay as that also gives you a short window (of three seconds) to distribute the launch orders

Visuals of the bugs that are fixed:

_Bug with Seraphim MML_

https://github.com/FAForever/fa/assets/15778155/e0b334d1-9697-463b-b969-ce23cf0841b6

_Bug with queued launch orders_

https://github.com/FAForever/fa/assets/15778155/02cd8035-4793-417a-88d4-3152f79027a1

_Bug with distributing launch orders_

https://github.com/FAForever/fa/assets/15778155/d62b8ad7-34e9-47a5-805e-bcf675e3c5b8


msgerbs:
Why did this unit even have a `RackSalvoReloadTime` at all? Isn't that meant for units with a salvo of multiple shots that has to reload between salvos?

Garanas:
I suspect it had this for years, maybe even in the Steam distribution:

image

But the real problem was triggered by the launcher having a reload animation. There are only two units in the game that have this.

msgerbs:
From a quick search it seems there are other units that have a `RackSalvoReloadTime` which also have a `RackSalvoSize` of 1. In fact I see that the few units with a `RackSalvoSize` of > 1 all have a `RackSalvoReloadTime` of zero. What does `RackSalvoReloadTime` do?

Garanas:
It is a separate reload time:

- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/defaultweapons.lua#L1062-L1100

After a discussion with Sprouto (of Loud) we think it may exist to mimic the time it takes for animations to finish. One way or another, it is a strange field that can cause strange behavior such as a unit being able to miss its firing cycle.

Garanas:
I think this change also fixes the bug where a strategic missile sub can launch two missiles, one immediately after the other. That unit has counted projectiles (the other condition, besides the reload time)

Garanas:
After testing, there's more involved to this bug than the fixes proposed so far. I've not yet fully mapped the how's and why's at this point

Garanas:
Long story short - there's really gameplay-affecting bugs that could cause a weapon to 'miss' its firing cycle. That is another problem to tackle for another time.

Garanas:
@ChessBerry this also fixes some issues you reported surrounding distributing launch orders 😃

PR #5455, opened 2023-09-18
Use Set and GetCreator to track external factory

clyfordv:
Creates a "select external factory button" with carrier:SetCreator(carrier.ExternalFactory). Rudimentary demo, no art on the button, both the changes here and the pod staging pre-processing at the beginning of CreateAltOrders would likely best be moved to their own functions with a callback system.


Garanas:
That's one way to do it 😄 , well found! I'll ask Fichom if he's available to make an icon for us

Garanas:
Is this part of #5484 ?

clyfordv:
It is, used some of the same backend elements

Garanas:
Does that mean we can close this pull request?

PR #5457, opened 2023-09-19
Scu rebalance

MrRowey:
Changes to the UEF,CYBRAN,Sera SACUS i did a while back


RabidPope:
Would it not make sense to reduce the cost of standard acu and add that cost to ras upgrade?

MrRowey:
> Would it not make sense to reduce the cost of standard acu and add that cost to ras upgrade?

That's an interesting approach.

PR #5458, opened 2023-09-19
Marr (Aeon stationary flak) buff

Basilisk3:
**Changes:**
MuzzleVelocity 35 --> 40

This unit has always been much worse than the other stationary flaks. Essentially it is a worse version of the Air Cleaner, with a much worse firing cycle, lower AOE, lower DPS and lower HP. In exchange for these disadvantages, it got 0.5 less FiringRandomnes, which does not help the unit much. 

This PR aims to buff the unit, while also making it more unique, in line with the theme of its faction. With this change it still has the disadvantage of having low damage output, but is now able to hit fast moving targets more reliably and at a greater range, than other flaks.


Garanas:
@Basilisk3 to remove the intermediate commits (sorry!) you can rebase this branch to the latest `deploy/fafbeta` of the main repository

Basilisk3:
@Garanas No worries, I remade my PRs. How do I keep this from happening in the future? Would protecting my branches help?

Garanas:
No, it is because of how FAF is setup. We have two development branches and when we update one with the content of the other then usually this happens. It is quite annoying 😃

PR #5461, opened 2023-09-21
Re-implementation of the lobby

Garanas:
This pull request represents the first baby steps towards a new implementation of the lobby. The goal is not just to re-implement the current lobby but also to document everything that is related to the lobby. 

Throughout this re-implementation we'll take into account various patterns:

- Use of classes to encapsulate distinct UI elements
- Use of `__init` to create the UI elements and position them
- Use of `__post_init` to attach the logic to the UI elements
- Use of callbacks to allow any function to attach their own logic to an event
- Use of extensive logging (that you can turn off with a switch) to see exactly what is happening at any moment

The use of callbacks may sound like a contradiction but it setup in such a way that is intuitive to debug and intuitive to write. Take as an example `lua/ui/lobby/lobby-communication.lua`. There are various engine events. Instead of overwriting them any UI element can attach behavior to those events and react to them accordingly. 

_more to come_


Garanas:
Note that at this moment there's nothing to see yet. We can discover lobbies over LAN (same machine) and we can host a lobby. Beyond that there's nothing implemented yet, at this moment I'm struggling to understand where the existing lobby actually starts to populate the UI 😃

Garanas:
This is the design that we'll be following:

- https://www.figma.com/file/JzKkkbFOCyjGB9lSSwgzAb/FAF?type=design&node-id=511-1205&mode=design

Basilisk3:
> This is the design that we'll be following:
> 
>     * https://www.figma.com/file/JzKkkbFOCyjGB9lSSwgzAb/FAF?type=design&node-id=511-1205&mode=design

Are you looking for feedback on this design, or is it just a mock-up? I think it looks pretty good so far, and it's great the lobby is finally being improved, but I would maybe change 1-2 smaller things.

Garanas:
You're welcome to provide feedback. The design was made last year by someone else, called Bobka on Discord

HUSSAR-mtrela:
@Garanas this is very ambitious project to modernize 7K+ lines of code in lobby and I might help with some UI if you want. Do you plan to re-implement just the lobby or are you planning to do the same for all lobby dialogs? (mods manager, units manager etc.)

Garanas:
@HUSSAR-mtrela someone has to do it 😄 , I do not want to undo your work on the map selection / mod selection dialogs. I'll try to re-use most of that as best as I can. At the moment it is a one-man job as I try to understand the foundations. Once that is done I'll give you a ping or two, see where we can cooperate on the new implementation

At the moment I'm still figuring out the foundations. Through LAN you can see games that are being hosted and you can join them!

image

Once joined however nothing happens, which makes sense because even giving a player a slot in the lobby is something that is communicated through the host / peer. But, I can see in the logs that everything 'works' and that I am connected to the lobby.

Garanas:
One interesting find is that you can connect to multiple lobbies in the same game instance 🤔 , not sure what the use of that is but it is possible it appears

Basilisk3:
> You're welcome to provide feedback. The design was made last year by someone else, called Bobka on Discord

Okay, for the sake of simplicity, I am just going to post all my thoughts on the design here. As soon as the new lobby is in a semi usable state, I will test it and post feedback again.

- Having to click to see who is in observers is not a good idea, as that information is pretty important and should be displayed at all times (having to choose between either seeing the chat or observers is odd too, if that is the case)
- Right now lots of options are missing, like moving to observer, patchnotes or game options
- Line spacing in chat is too high
- Small nitpick: The way ratings and game count are displayed is a bit odd. Maybe instead of R1800 it should be something like R: 1800. R1800 looks like it is supposed to be read as one “word”.
- (Maybe showing a user's avatar would be a nice addition, considering they are given out as tournament prices. I think they are pretty underused, and maybe it gives some players more of an incentive to participate in them)

MrRowey:
> * art
I agree with some of the nitpick you mentioned also remember like jip stated this design was treated over a year ago which has predated some new functionality in the lobby like the patch notes button :)

Basilisk3:
I disagree that having to click to see observers and a lot of missing buttons are nitpicks. If too much is missing, and too many smaller things need to be changed, you eventually end up redesigning the whole thing, because you start to put more and more layers on top of something that was designed without these things in mind. This is the way we ended up with our current lobby.

PR #5464, opened 2023-09-24
Process balance suggestions on tech 3 MAA

### LINKED ISSUE #5453: Balance idea: T3MAA
## Intended changes

Non Sera T3 maa
* Mass cost 800 > 600
* Energy cost 10.000 > 7.500
* Build time 4.000 > 3.000

Sera T3 maa
* Mass cost 960 > 720
* Energy cost 12.800 > 9.600
* Build time 4.800 > 3.600

Non Cybran T3 maa
* HP -300 (each faction has a slight difference in hp)

## Thoughts on the changes

Straight buff to T3MAA. The direction allows players to get them out earlier. Cybran hp is relatively stronger to make up for the lack of mobile shields.

Garanas:
Closes #5453


units/DRLK001/DRLK001_unit.bp
```diff
@@ -37,8 +37,8 @@ UnitBlueprint{
         AirThreatLevel = 33,
         ArmorType = "Normal",
         EconomyThreatLevel = 0,
-        Health = 1900,
```
Blodir, line 40: It's a bit confusingly stated, but (exceptionally) cyb hp should stay at 1900

Blodir (changes requested):

Issue #5418, opened 2023-09-10
Balance idea: Vulthoo


Blodir:
## Intended changes

Vulthoo
Mass: 300 -> 500 (+66%)
Energy: 6000 -> 10000 (+66%)
Buildtime: 2000 > 3333 (+66%)
Movespeed: 12 > 11
HP: 1080 > 1800 (+66%)
DPS: 80 > 100 (+20%)
Mesh scale: +10% ish

## Thoughts on the changes

Reinforces a more unique identity for the Vulthoo as a tech 2.5 gunship. Makes up for Sera lack of t3 gunship.

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

PR #5466, opened 2023-09-24
Process balance suggestions on Vulthoo

Garanas:
Closes https://github.com/FAForever/fa/issues/5418

Interpret the damage increase as 20%, meaning all projectiles now deal `19.2` damage instead of `16`.

Increased the collision sphere to match the increase of the size of the mesh.


units/UEA0203/UEA0203_unit.bp
```diff
@@ -19,7 +19,7 @@ UnitBlueprint {
         KTurn = 0.8,
         KTurnDamping = 1.5,
         LiftFactor = 7,
-        MaxAirspeed = 12,
+        MaxAirspeed = 11,
```
Blodir, line 22: Wrong blueprint (this is stinger, not vulthoo)

Blodir (changes requested):
The movespeed change is on the wrong blueprint. Otherwise ok.

PR #5467, opened 2023-09-24
Process balance suggestions on Stinger

### LINKED ISSUE #5417: Balance idea: Stinger
## Intended changes

Stinger
Mass: 240 -> 192 (-20%)
Energy: 4800 -> 3840 (same mass/e ratio as vulthoo)
Buildtime: 1600 -> 1280
Movespeed: 12 -> 13.5
HP: 880 -> 700 (approx. -20%)
DPS: 66.6 -> 53 (approx. -20%)

## Thoughts on the changes

Supports the intended identity of the stringer, which is to be a fast cheap raider and a hybrid transport.

Garanas:
Closes #5417


units/UEA0203/UEA0203_unit.bp
```diff
@@ -271,7 +271,7 @@ UnitBlueprint {
             FiringRandomness = 0.2,
             FiringTolerance = 2,
             Label = "Turret01",
-            MaxRadius = 22,
+            MaxRadius = 18,
```
Blodir, line 274: This is the range right? Damage should be 20->16, no changes to range

Blodir (changes requested):

PR #5472, opened 2023-09-24
Atlantis: Increase its BackUpDistance from 15 to 30

Basilisk3:
**Changes:**
Atlantis: BackUpDistance 15 --> 30

The Atlantis had a rather low BackUpDistance of 15. This meant, that in order to retreat slightly, it usually had to turn by 180 degrees and then drive forwards. This made the unit feel clunky to micro.


Garanas:
Calling it, without the increase it is barely useable. The increase may be useful in general for other experimentals that can walk backwards

PR #5474, opened 2023-09-26
Separate files for each type of projectile

Garanas:
The objective of these changes is to enhance the ease of hooking the individual classes. Some time ago, in the modding channel on Discord, there was a discussion where someone expressed difficulty in hooking the AIBrain components before they became part of the AIBrain class. This issue arises because both the components and the main AIBrain class reside within the same [file](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/aibrain.lua).

When you hook a file, your code is appended to the end of the file. Consequently, it becomes impossible to hook the AIBrainHQComponent before it has already been applied to the AIBrain class.

Similarly, it becomes challenging to hook various classes (e.g., projectiles, units) before they are already applied to another class before you can hook it. To address this challenge, one would have to redefine a class, which is not ideal for compatibility. A noteworthy example of this can be found in the [RKS Explosion mod](https://github.com/RK4000/RK-s-Explosions/blob/master/hook/lua/defaultunits.lua), where this (bad, but inevitable) practice is applied.


Garanas:
Note that this pull request is an example of what it would look like 😃

PR #5481, opened 2023-09-29
AI -  Add transport return to base functionality on low fuel/health 

### LINKED ISSUE #5478: AI Transport functionality missing rtb when low fuel/health
**Describe the bug**
In the transportutilities.lua module there is a function called ProcessAirUnits. This will make a transport return to base if it is low on fuel or health. The transport then returns and lands, and will either slowly refuel or an engineer will come and fix it. But we have not created the function of that yet so it does nothing and causes multiple issues with transport functionality.

**To Reproduce**
Start an AI game, when a transport takes an engineer to a location if it is damaged below 80% while enroute it will not depart nor will the engineer do what it was supposed to do.

**Expected behavior**
Transports will return for refuel or repairs, or the logic will handle this situation without causing transports hanging due to damage checks being true.
image

**Additional context**
Requires an RTB function specifically for transports
Requires a flag be set on the transport that repairs are required
Requires an engineer builder that will repair transports that are flagged as needing repair
Needs to stay out of the transport pool until it is repaired (I think currently that would go back in as soon as their health is above a certain percentage, dont quote me on that)

relent0r:
closes #5478

This PR fixes an issue with the new transport module where it would not do anything with transports that were low on fuel or health.

When the transport module was implemented there was an air refit thread that was not implemented due to the differences in FAF and LOUD.
In LOUD the transports would return to base and land close to a staging platform which provides an refuel/repair aura. This is not available in FAF.

I've added a return to base thread that will trigger when a transport is low on fuel or very low on health.
In the case of fuel it will return to base and wait until its fuel is above a threshold then will go back into the pool.

If the transport is less than 30% health it will kill itself when it is within 35 units of the return position.


relent0r:
p.s I'm having some stability issues with the dev branch at the moment so couldn't delve too far into adding more logic into the return to base loop.

Garanas:
I think the solution to just kill the transport is fine, repairing transports is highly unusual in FAF and it barely pays off economically

relent0r:
Have updated so that the transport won't kill itself until it is 35 units from the return position.

PR #5482, opened 2023-09-29
AI - State Machine fixes

relent0r:
This PR adds some state machine fixes and changes as per the following. Wanting to merge this in now so that other branches can be tested better.

- Adds small platoon template for lab hunter early raiders.

- Adds another land scout former. This was due to an issue where the default would take all the scouts and leave nothing for the platoons. Now there is only a single instance land scout former and the multi instance requires more than one scout in the army pool.

- Adds functionality for reclaim engineers to disable the builder former for a period of time if threat is too high in the reclaim areas multiple times. This will be better fleshed out in the future. The current implementation stops the engineer manager being thrashed by the platoon being picked up then the engineer logic exiting due to no available reclaim locations.

- Fixes an error with the raid platoon where it would not be checking pathing to raid positions correctly.

- Add a return to base state machine.

- Adds a guard state machine, this is currently a 1 to 1 match for the platoon function. Except it will attack units correctly and will search for other engineers to protect if its guardee goes into water and it can't follow. Will be improved more.


lua/editor/MiscBuildConditions.lua
```diff
@@ -373,6 +373,15 @@ end
 ---@return true | nil
 function ReclaimAvailableInGrid(aiBrain, locationType, mapSearch)
     -- this condition won't work without a reference to the reclaim grid
+    if aiBrain.ReclaimFailCounter > 10 then
```
Garanas, line 376: This should probably be a separate condition that you check first (because it is so cheap to check)

relent0r: ah yup fair call. Actually the bit I don't like about that condition is that its making changes to the brain, I'd rather keep builder conditions as read only functions.

relent0r: Have separated out. I will likely seek to have the brain changes managed by a brain thread before the release and have that condition just checking an enabled flag.

Garanas: I agree that conditions should be read-only, let's tackle that in another pull request however!

PR #5484, opened 2023-09-29
External factory platforms can pass build orders and show the queue of their attached factories

clyfordv:
Mobile factories (fatboy, carriers, etc.):
1. Have their vanilla build queues visible
2. Can send build orders to their child factory through that queue
3. Will in turn show the construction queue of their child factory
4. Synchronize pause and repeat build button clicks (and vice versa)
5. Can do all the above with any number of mobile factories selected*
6. ... and better yet, without interrupting their current order queue (including while moving!)

*Normal factory homogeneity rules apply--only the units every selected factory can build will be shown. As in the stock game, construction queues won't show up when more than one unit is selected.


lua/defaultcomponents.lua
```diff
@@ -849,22 +849,24 @@ ExternalFactoryComponent = ClassSimple {
...
         self.Trash:Add(self.ExternalFactory)
     end,
 
+    --- Comment these out for now until a more permanent solution is decided on
     ---@param self Unit | ExternalFactoryComponent
     OnPaused = function(self)
-        if self.ExternalFactory then
-            self.ExternalFactory:SetPaused(true)
-        end
+    --    if self.ExternalFactory then
```
Garanas, line 860: You can remove the `OnPaused` and `OnUnpaused` functions all together

lua/system/blueprints-units.lua
```diff
@@ -559,11 +559,11 @@ function PostProcessUnitWithExternalFactory(allBlueprints, unit)
         efBlueprint.SelectionCenterOffsetZ = 0.35 * unit.SelectionSizeZ
 
         -- remove properties of the seed unit
-        unit.CategoriesHash['FACTORY'] = nil
-        unit.CategoriesHash['CONSTRUCTION'] = nil
+        -- unit.CategoriesHash['FACTORY'] = nil
```
Garanas, line 562: Feel free to remove commented code too

lua/ui/game/construction.lua
```diff
@@ -1444,8 +1444,24 @@ function OnClickHandler(button, modifiers)
                     import("/lua/ui/game/commandmode.lua").StartCommandMode(buildCmd, {name = item.id})
                 else
                     -- If the item to build can move, it must be built by a factory
-                    -- TODO - what about mobile factories?
-                    IssueBlueprintCommand("UNITCOMMAND_BuildFactory", item.id, count)
+                    -- Mobile factories: we check for platforms (the attached units can be given orders as normal)
+                    -- If we've got platforms, we take our selected units (minus the platforms), then add the
+                    -- external factories to that list, then give orders with 
+                    -- IssueBlueprintCommandToUnits (which can give orders to an arbitrary list of units)
```
Garanas, line 1450: Great use of a function that exists only for weeks now 👍

Garanas (approved):
After a test or two I can only commend you for absolutely great work 👍 . The implementation you propose works seamlessly with what one would expect. I'll keep trying to break it over the coming days 😃 , see if we can find and attend to some edge cases

Garanas:
@clyfordv I've merged in these changes as that makes it easier for people to tweak. There's the toggle texture that is still in development / to be discussed, we can tweak that in future pull requests

clyfordv:
Definitely. In terms of the icon, I look into the future on this and I see a couple of things:
1. Toggle buttons turns into a "set rally point button", with a new style of icon (something like a flag, hopefully much less controversial)
2. Hitting that button activates a custom command mode, keeps the same selection visuals, but likewise shunts all orders into the rally point queue of the external factory unit (to be seen if that's possible without selecting the exFac unit). Right click resets the rally point, leading to...
3. An unset rally point follows its parent (actually: gets reset to a default position) on a beat/second interval, so a moving factory with an unset rally point will leave a line of units on the map.
4. Carriers with rally points auto-deploy constructed aircraft, while carriers with unset rally points put them in storage (let me know what you think of that one, putting the auto-deploy on the deploy button instead is an option).

PR #5487, opened 2023-09-30
Raised Platform visualisation for cheat spawn menu

The-Balthazar:
Since the `ShowRaisedPlatforms` command appears to do nothing, this adds a toggle option to the cheat spawn menu to outline raised platforms with beams.
image


Garanas:
Works as expected, well done 👍

Issue #5476, opened 2023-09-28
Spawn markers have incorrect names


relent0r:
**Describe the bug**
I think I take responsibility for this as I was asked to validate it and didn't validate the data all the way through.

When Spawn markers are created they are using the key which comes out as Army_X and this will be attached to the name.
BUT
When GetMarkersByType is used and no cache exist then it will do this.

    for k, marker in AllMarkers do
        if marker.type == type then
            -- mod support syntax
            marker.Name = k
            marker.Size = marker.size
            marker.Resource = marker.resource
            marker.Type = marker.type
            marker.Orientation = marker.orientation
            marker.Position = marker.position
            marker.Color = marker.color
            ms[n] = marker
            n = n + 1
        end
    end

It looks like we are overwriting the marker.Name with the index key instead of keeping the original name. So the Name for Spawn markers is no longer ARMY_X but just X

**To Reproduce**
Start a map and log the Spawn markers that are generated. Then have an AI request the Spawn markers, the resulting table will have invalid names causing AI to fail the check.

**Expected behavior**
This type of query will return the correct markers.

        local tempMarkers = AIGetMarkerLocations(aiBrain, 'Spawn')
        for k, v in tempMarkers do
            if string.sub(v.Name, 1, 5) == 'ARMY_' then
                table.insert(markerList, {Position = v.Position, Name = v.Name})
            end
        end

relent0r:
Obviously the alternative is to just remove that string check since you don't need it. I'm not sure how many other AI do a similar check. I know the default has a bunch of these sorts of string checks due to the usage of marker type 'Blank Marker'

relent0r:
I had a look over the codebase for this to see what the ramifications of removing the string check are.

These are the locations where we are referencing start positions.
aiutilities.lua
92 - AIGetStartLocations - References the marker chain so not impacted.

106 - AIGetSortedScoutingLocations - Uses the ScenarioUtils.GetMarker function to query the ARMY_ string.

345 - AIGetMarkerLocations - Uses new method but is checking for string.

504 - AIGetMarkerLeastUnits - Uses new method but is checking for string.

613 - AIFindStartLocationNeedsEngineer - Uses new method but is checking for string.

3094 - AIFindFurthestStartLocationNeedsEngineer - Uses new method but is checking for string.

The AI brains uses the ScenarioUtils.GetMarker function for initial threat assignment and building scout locations.

[...]

PR #5488, opened 2023-09-30
Try to preserve the original name field of a marker

### LINKED ISSUE #5510: AddBuilderManagers will not take NavMesh generated markers
**Describe the bug**

This is in regards to this report 
https://forum.faforever.com/topic/6538/ai-sea-gameplay-problem/5
This is the AddBuilderManagers function as it stands right now.
`    
AddBuilderManagers = function(self, position, radius, baseName, useCenter)

        local baseLayer = 'Land'
        position[2] = GetTerrainHeight( position[1], position[3] )
        if GetSurfaceHeight( position[1], position[3] ) > position[2] then
            position[2] = GetSurfaceHeight( position[1], position[3] )
            baseLayer = 'Water'
        end

        self.BuilderManagers[baseName] = {
            FactoryManager = FactoryManager.CreateFactoryBuilderManager(self, baseName, position, radius, useCenter),
            PlatoonFormManager = PlatoonFormManager.CreatePlatoonFormManager(self, baseName, position, radius, useCenter),
            EngineerManager = EngineerManager.CreateEngineerManager(self, baseName, position, radius),
            BuilderHandles = {},
            Position = position,
            BaseType = Scenario.MasterChain._MASTERCHAIN_.Markers[baseName].type or 'MAIN',
            Layer = baseLayer,
        }
        self.NumBases = self.NumBases + 1
    end,
`

This particular line here
`BaseType = Scenario.MasterChain._MASTERCHAIN_.Markers[baseName].type or 'MAIN',`

So the function is passed the name of the marker, which is the .Name property.
It then looks this up in the masterchain. Which won't contain the NavMesh markers.

So the BaseType will always be 'MAIN' since the marker doesn't exist in the marker chain.

**To Reproduce**
Run any AI game that uses generated markers and the bases will not be named correctly.

**Expected behavior**
The BaseType will use the type property of the marker used to create the base.

**Additional context**
I looked at a few options, but the best one is to have a helper function in the marker utilities module that will lookup the marker for a marker name and return it. Then update the AddBuilderManagers function that is against the default AI brains to perform this check while leaving the base functionality as is.

The other option is to modify the functions and pass the entire marker through rather than just the name and position. This is possible and makes sense but would negatively impact 3rd party AI that leverage these functions. It would require getting functions like `AIFindNavalAreaNeedsEngineer` to return the marker itself rather than just the position and name.

Garanas:
Closes https://github.com/FAForever/fa/issues/5476
Closes #5510


Garanas:
@relent0r would this suffice?

relent0r:
> @relent0r would this suffice?

The only issue I see and its not an issue in normal usage due to the cache.
Line 131 is checking for marker.type of 'Spawn' but the marker.type == 'Blank Marker'.

info:  - Name: ARMY_5
info:  - Position: table: 14D7C230
info:  -    1: 119.5
info:  -    2: 20.093799591064
info:  -    3: 519.5
info:  -    type: VECTOR3
info:  - RallyPoints: table: 26DC91E0
info:  - Size: 25
info:  - color: ff800080
info:  - orientation: table: 14D7CF78
info:  -    1: 0
info:  -    2: 0
info:  -    3: 0
info:  -    type: VECTOR3
info:  - prop: /env/common/props/markers/M_Blank_prop.bp
info:  - size: 25
info:  - type: Blank Marker

One thing I'm noticing though is we need to do work in the code due to the lower vs upper case we are using. 
e.g

function AIGetMarkerLocations(aiBrain, markerType)
    local markerList = {}
    if markerType == 'Start Location' then
        local tempMarkers = AIGetMarkerLocations(aiBrain, 'Spawn')
        for k, v in tempMarkers do
                table.insert(markerList, {Position = v.Position, Name = v.Name})
            end
        end
    else
        local markers = import("/lua/sim/markerutilities.lua").GetMarkersByType(markerType)
        if markers then
            for k, v in markers do
                table.insert(markerList, {Position = v.position, Name = k})
            end
        end
    end

    return markerList
end

Ideally I don't need to actually create the table anymore and can just pass the markers straight through. But I just need to figure out how to make sure its consistent.

relent0r:
Perhaps this is a case where having both lower and upper case will make life easier :)

Garanas:
@relent0r how's this solution? In general, we should no longer rely on the marker chain

PR #5489, opened 2023-09-30
Remove global state surrounding roll off for units

Garanas:
With thanks to https://github.com/FAForever/fa/issues/5436 we can now efficiently assign some orders to individual units. That removes the need for the global field `RolloffUnitTable`. Instead of re-computing the roll off point upon finishing we instead cache the roll of position in the factory that is building the unit. That removes the need for the global field `RolloffPositionTable`. 

We now roll off the unit the direction it is facing in. There's no separate spin at the end of the build sequence like we had before, which looked bad especially for large units

https://github.com/FAForever/fa/assets/15778155/a6c922d7-b7e8-4527-b419-915e128e6398

### DIFF
--- a/lua/DefaultUnits/Factory.lua
+++ b/lua/DefaultUnits/Factory.lua
@@ -1,15 +1,12 @@
 
 local StructureUnit = import("/lua/defaultunits/structure.lua").StructureUnit
-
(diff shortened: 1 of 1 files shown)


Garanas:
@Hdt80bro you'll be fond of this one, no more spinning!

PR #5490, opened 2023-09-30
Add missing emission on maps without water

BlackYps:
I forgot that we still need to add the emission, even if we don't do any water processing. This lead to inadvertantly turning of the emission on maps without water.

Screenshot 2023-10-01 000509

Screenshot 2023-10-01 000653


Garanas:
Great find 👍

PR #5491, opened 2023-09-30
Tune Cybran albedo

BlackYps:
There was feedback that the cybran units look too bland. Adding a bit of the red specular channel brings back some of the albedo variation. This works because originally the specular red channel added sunreflection, essentially making the albedo brighter. We have already encountered this on the Aeon units, where adding the specular red channel to the albedo greatly homogenized their look. Here we use this trick to bring back some detail that got lost.
It also makes metal parts look brighter, which is a nice touch as they were a bit too dark before.
Screenshot 2023-10-01 001715
Screenshot 2023-10-01 001535
Screenshot 2023-10-01 001538


Garanas:
image

image

image

Garanas:
Maybe we can convince @MadMaxFAF to look at the textures of the spiderbot, see if he can make it look similar to the Megalith

image

BlackYps:
Yeah, the monkeylord could use a revisiting of its textures

PR #5495, opened 2023-10-01
No longer force the fire state of mobile units to ground attack

Garanas:
When a unit is built by a factory it inherits the fire state of that factory. However, when a mobile unit is created the fire state was overwritten to always be the attack ground fire state

All factories default to the attack ground fire state. A user can now switch that to the return fire state or the no fire state and units will inherit that accordingly.

### DIFF
--- a/lua/DefaultUnits/Mobile.lua
+++ b/lua/DefaultUnits/Mobile.lua
@@ -3,7 +3,6 @@ local EffectUtil = import("/lua/effectutilities.lua")
 
 local Unit = import("/lua/sim/unit.lua").Unit
 local TreadComponent = import("/lua/defaultcomponents.lua").TreadComponent
-local FireState = import("/lua/game.lua").FireState
 
 ---@class MobileUnit : Unit, TreadComponent
 ---@field MovementEffectsBag TrashBag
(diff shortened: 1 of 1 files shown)


Garanas:
Note that the game behaves different when you attack move units depending on their fire state. Therefore forcing it to always be ground fire is a bit odd

Uveso:
Just as side note, we fored the firemode to default ground attack in 2017 to make groundattack easier and spare some apm.
So you can alt + right click for attack move and return fire, or just normal attack for goundfire.

Now we are back to the need of change the firestate, instead of just holding down alt key.
I am not sure if this was indended.

(This is just an info, i don't want do discuss this!)

Garanas:
Factories still have their fire mode set to ground attack upon creation. Units inherit that fire state. The default behavior (where the user does not edit the fire state of the factories) is the same to what we have now, at least that is the intention 😃 

The goal of these changes is to give players the control to do different if they choose to do so. The fire state does impact the behavior of the unit. I recall that @Hdt80bro had written some details about that

Garanas:
I found it after he found it on the AI Discord, see also: https://github.com/FAForever/fa/pull/4474#issuecomment-1339742523

Uveso:
You don't need to convince me, i am ok with it.
As i said, it was just an info :)

Garanas:
I'm not trying to convince you of anything, just adding more info 😃

PR #5497, opened 2023-10-03
Carriers with external factories can auto deploy constructed units

clyfordv:
What it says on the tin--carriers now have an order button to toggle putting units in storage or deploying them to the rally point.

Currently a separate order button, the current checkbox.lua only supports one toggle state so a derived checkbox will be needed + integrated with orders.lua (in development).

Included is the base infrastructure for blueprints to override any element of an order button (behavior, preferred position, etc.), not just bitmap and help_text.


lua/system/blueprints-units.lua
```diff
@@ -558,6 +558,23 @@ function PostProcessUnitWithExternalFactory(allBlueprints, unit)
         efBlueprint.SelectionSizeZ = 0.25 * unit.SelectionSizeZ
         efBlueprint.SelectionCenterOffsetZ = 0.35 * unit.SelectionSizeZ
 
+        -- add order overrides to carriers
+        if unit.CategoriesHash['CARRIER'] then
```
Garanas, line 562: Best to check for both `CARRIER` and `EXTERNALFACTORY`, as this only applies to that combination

clyfordv: Is it necessary to re-check, given that this is all in a function that's called only when `unit.CategoriesHash['EXTERNALFACTORY'] == true` to begin with?

Garanas: Oh woops, that was not visible in the diff 😃 . No, then we're fine!

Garanas:
The Atlantis requires an additional check:

image

It can at the moment deploy automatically while underwater. Unless we want that (it is experimental after all) and we're just lacking a splash effect for when the plane emerges 😃

Garanas:
I tried to make the unit 'eject' properly by attaching it to a dummy projectile. But it appears that the unit does not inherit the velocity of the projectile and just stalls to 0 when the dummy projectile is destroyed
https://github.com/FAForever/fa/assets/15778155/3fe66e02-469e-4c17-9b68-e6f864d2831b

clyfordv:
I'll experiment a little. Oddly enough the Atlantis releasing units from underwater appears to be stock behavior (when it has no more storage capacity).

Garanas:
That's odd 🤔 , it seems fine to keep it if the Atlantis indeed has no additional storage at that moment

Garanas:
@clyfordv I'll merge this in for now, we can make more improvements in future pull requests

PR #5498, opened 2023-10-03
Tweak trajectory of the tactical missiles

Garanas:
_Seraphim tech 2 cruiser_

https://github.com/FAForever/fa/assets/15778155/c537dfe3-68e9-4375-a901-490d2c7ae101

_Seraphim tech 3 carrier_

https://github.com/FAForever/fa/assets/15778155/0fca05c5-1ac9-41fd-bd78-b79222e59366

_UEF tech 2 cruiser_

https://github.com/FAForever/fa/assets/15778155/761a0d16-2fbc-4785-8ec4-f98f3f5cbbf5

_Aeon tech 3 missile ship_

https://github.com/FAForever/fa/assets/15778155/f5e9f6e0-6be6-48ae-b724-f44bcbc7d185

PR #5501, opened 2023-10-03
Add a bit of a fudge factor to the arc of tactical missiles

Garanas:
As mentioned in https://github.com/FAForever/fa/pull/5498 , a draft example of how such a range factor can work out in practice

https://github.com/FAForever/fa/assets/15778155/9adefce3-4193-49e8-acda-76e299efb38f


Pokute:
Ooh. Is it possible to change the launch angle too? Just a few degrees so they wouldn't clump together so much.

Garanas:
I'm sure that would be possible, but they are launched vertically for this particular unit; they are silo tubes after all 😃

Garanas:
@clyfordv what are your thoughts on these changes?

PR #5502, opened 2023-10-04
Randomizes max height and launch ticks

clyfordv:
Didn't see that you had already implemented a version of this, but had put it together in a few minutes so oh well! This one is a little simpler, which helps prevent parameter overload for anyone who comes along later, but otherwise I'm agnostic.


Garanas:
We'll go with what we have now; luckily one doesn't need to use all parameters 😃

PR #5503, opened 2023-10-05
Fix issues surrounding the Ahwassa

Garanas:
The LOD1 of the Ahwassa was out of touch with the LOD0, as shown in the videos below. On top of that, the Ahwassa has a separate trail effect but it does not trigger anymore after the unit had landed. All trails would no longer trigger after the unit landed. To fix this we re-create the movement effects of a unit as it changes layers (Land -> Air, Land -> Water, etc). This also fixes contrails missing of other air units

_new LOD_

https://github.com/FAForever/fa/assets/15778155/ce674d87-3d23-4c42-a61e-966e91f6ff21

_old LOD_

https://github.com/FAForever/fa/assets/15778155/3f58b3ed-f8a8-4a10-ae0f-4d75a7c94cdc

_trail effect of the Ahwassa_

https://github.com/FAForever/fa/assets/15778155/64a990ac-5ff2-4832-a14a-2590c40adf0b


Garanas:
There are a lot of Seraphim units with poor LOD1s:

image

BlackYps:
Good work on the Ahwassa! Problems with textures are relatively easy to fix, provided the uvs stay the same. Then we can just downscale and reuse the lod0 texture. Model changes are more complicated

Garanas:
This was achieved by using the `decimate` operator in Blender, it works ok-isch for large units and it preserves the UVs quite decently

PR #5507, opened 2023-10-06
Translate untranslated English -> Simplified Chinese.

HaoJun0823:
Hello.

Some Chinese players told me that some content in the game was not translated into Chinese, so I found this place based on the official website prompts.

I translated the content that has not been translated. I think it is not a perfect Chinese version, but I think it is enough for Chinese players to read.

I will actively address your feedback and provide suggestions.

Best regrads.


Garanas:
Cross-checked a few references, it seems all good. Thank you for your time to improve the translations 😃

PR #5512, opened 2023-10-07
Re-introduce mobile factory to CZAR

Garanas:
Re-introduces the mobile factory aspect of the CZAR. Still requires a bit of tuning, especially the idea of @BlackYps to have the ability to determine the location of the selection area is useful for the CZAR.


Garanas:
With thanks to @clyfordv we have a toggle for carriers to allow air planes to immediately take off. This pull request also makes that look better by toying with the build bone and the roll off order. The behavior is more similar to when units would take off through deploying them

https://github.com/FAForever/fa/assets/15778155/abd999fb-7c6b-4a4f-9882-c51e54821c3f

PR #5518, opened 2023-10-10
Unlocking projectile defending weaponry

Garanas:
Due to the `DesiredShooterCap` usually being set to 1 a single weapon can aim and fire at a single projectile. This works fine in general. Except that a weapon can lock onto a projectile while it is unable to fire due to a reload time. As a result other weapons capability of firing at the weapon ignore the projectile and because of that behavior the projectile in question may hit its target.

This looks and feels a bit silly.

This pull request introduces two new blueprint fields for weapons:

- (1) `TargetResetWhenReady`: When a weapon reaches the `ready to fire` state but does not immediately fire then the target of the weapon is reset after five ticks
- (2) `DisableWhileReloading`: When a weapon is done firing it will be disabled for the duration of the reload time

Where (1) is a less aggressive version of (2). With (1) we essentially tell a weapon to temporarily unlock the projectile, allowing another TMD to lock on the projectile that may be able to fire immediately. With (2) we completely disable the weapon for the reload duration, disallowing it to lock all together. The consequence of (2) is that the weapon doesn't track the target and therefore 'jumps' towards the next missile. This is okay for weapons that do not visually track, such as a torpedo defense.

As a consequence both torpedo and tactical defenses become a lot more efficient. As an example, these five TMDs have been surviving the relentless fire from five Seraphim cruisers:

https://github.com/FAForever/fa/assets/15778155/4394458f-2478-41db-957f-0c2ee523136e

And these six Seraphim tech 1 submarines have been surviving the relentless fire from a single Seraphim tech 3 submarine:

https://github.com/FAForever/fa/assets/15778155/6f2e4181-55ee-4034-946d-dafb5fb39270

The statistics of the weapons did not change. What changed is that weapons no longer needlessly keep a lock on a projectile while they are unable to fire at it due to reloading time.


Garanas:
There's some changes of https://github.com/FAForever/fa/pull/5515 in this pull request because I apparently do not read the command line when I try to switch branches

Garanas:
https://github.com/FAForever/fa/assets/15778155/f9d33098-4c86-42cd-a915-8d16dc526fe0

Garanas:
Seraphim missiles now have a `FxImpactNone` effect, giving a little explosion when the missile runs out of lifetime


https://github.com/FAForever/fa/assets/15778155/3282b010-b4a0-411b-9481-229b54c33a0a

PR #5519, opened 2023-10-10
Behavior toggles for built units

### LINKED ISSUE #5496: Automatically enable / disable properties of units that are produced by factories
Description
-----------

This feature is inspired by https://github.com/FAForever/fa/pull/5495

When a unit is created there are properties from the factory that the unit can inherit. There are two examples that I am aware of:

- A unit can inherit the fire state of the factory
- A unit can inherit the control group of the factory

In this issue I propose we extend this feature through toggles. The reason for toggles as that they can be adjusted by the UI, and therefore we can update the toggle to match what is in the preference file of a user. A similar approach is done in https://github.com/FAForever/fa/issues/4524 by @haifron 

Once the toggle exists we can interpret the toggle during the `OnStopBuild` event and update the properties of the unit that we're building. As an example of a few properties that may be interesting:

- (1) Automatically enable or disable the intel of a unit if it consumes maintenance (Cybran Mole, ASF, Strategic Bomber, ...)
- (2) Automatically enable or disable the construction of (strategic) missiles (strategic submarines, Seraphim battleship)
- (3) Automatically enable or disable the shield of a unit if it consumes maintenance

And I'm sure we can think of other examples in the (near) future. The goal of these changes is to reduce the 'initial maintenance' or the 'oh shit, I forgot' type of issues that players can perceive

clyfordv:
Adds a menu to factories that allows toggling stealth/shields etc. for units produced by that factory.

Uses a SimCallback from orders.lua to update a stat on the factory, then applies that in OnStopBeingBuilt of the unit under construction.

Processing for what toggles are available/which factories have which toggles is pushed up as much as possible to blueprints-units. This allows the relevant info to be easily accessible on both sides of the sim/ui divide, and minimizes the processing time when initializing the order popup (likely negligible, but it's the principle of the thing). Also makes overrides (manual or otherwise) much easier to track and apply.

Along with the main changes, it externalizes the pre-check functions in orders.lua (DronePreCheck, ExternalFactoryPreCheck, OnBuildTogglePreCheck) as that section of AddAltOrders was getting unwieldy. Also tightens up the visuals of the OnBuildToggle and Firestate selection popups (the stock width and height didn't look good with the round order buttons, and I wanted the firestate popup to be consistent).

To-do:
1. Remove hardcoded toggle behavior from remaining units (jamming, cloak, always off etc.) so they can inherit from the factory
2. Add construction specific tooltip overrides so the purpose of the popup menu is clear
3. Waterproof the SimCallback to prevent unforeseen stat setting shenanigans
4. As a toe-dip into non-script-bit OnBuilt behavior toggles, add a "constructed engineers assist factory" toggle alongside the existing options.


lua/SimCallbacks.lua
```diff
@@ -88,6 +88,21 @@ end
 Callbacks.EmptyCallback = function(data, units)
 end
 
+Callbacks.SetStatByCallback = function(data, units)
```
Garanas, line 91: I'm not sure about this callback, could you elaborate why it is required?

clyfordv: I couldn't find a method to set stats from the UI. Is there one?

Garanas: No, there's not. It is a sim -> UI type of communication, not the other way around. I'm not certain, but are toggles limited to just 8 toggles? We can't have more?

clyfordv: They are limited to eight unfortunately. After some experimentation and strife*, the only spare one (SpecialToggle) doesn't actually update a value, it can only calls a function (not useless, but not good as an address bit). Also, flipping a toggle bit causes the current selection to flicker (for whatever reason) which messes with some of the order button init stuff. It's also slow.

*OnScriptBitSet returns n, but if you want to get the same value with GetScriptBit you need to ask for n+1..!

I think the SimCallback solution is very robust--it's fast, smooth, reliable, and extensible. If the SimCallback is an issue we could put a go-list for the stats, but even a no-list (to protect veterancy, etc.) would do the job because updating a stat won't do anything if the sim doesn't look for it.

Garanas: Thank you for the elaboration! I'll let your findings sink in for a bit. I'm not a huge fan of sim callbacks because they burden the network and they need to be written in such a way that the change they are abused is minimal.

clyfordv: Do ScriptBitToggles have a different interaction with the network? Or does each SimCallback need to be sent out to each peer, not just the stat change?

clyfordv:
Done:
X. Remove hardcoded toggle behavior from remaining units (jamming, cloak, always off etc.) so they can inherit from the factory
X. Waterproof the SimCallback to prevent unforeseen stat setting shenanigans

To-do:
1. Tooltips
2. Re-work auto-deploy to use a StatToggle instead of the weapon toggle so the weapon toggle button doesn't show up (another advantage of the StatToggle system is that it solves a problem where auto-deploy required a little bit of a hacky workaround for the weapon toggle/Transport order).

clyfordv:
Most recent changes:

1. Auto-deploy now uses the StatToggle infrastructure (this also fixes the "set attribute on boolean" issue encountered before)
2. Maintenance consumption is properly applied to OnBuiltStatToggle units, whether they're created in a carrier or not

to-do:
- in the future, use merge instead of rebase to avoid jacking up comment history

lua/ui/game/orders.lua
```diff
@@ -609,10 +613,67 @@ local function ScriptButtonInitFunction(control, unitList, toggleCheck)
...
+end
+
+local function StatToggleInitFunction(control, unitList, subCheck)
+    local result = nil
+    local mixed = false
+    for i, v in unitList do
+        local thisUnitStatus = (v:GetStat(control._data.statToggle, 0).Value == 1 and true) or false
+        if result == nil then
+            result = thisUnitStatus
+        else
```
4z0t, line 661: use `elseif`

lua/ui/game/orders.lua
```diff
@@ -722,6 +783,117 @@ local function CreateBorder(parent)
...
+        end
+
+        -- see if we're mixed
+        local result = nil
+        local mixed = false
+        for i, v in selected do
+            local thisUnitStatus = v:GetStat(statToggle.stat, 0).Value
+            if result == nil then
+                result = thisUnitStatus
+            else
```
4z0t, line 810: use `elseif`

lua/ui/game/orders.lua
```diff
@@ -722,6 +783,117 @@ local function CreateBorder(parent)
...
+local function OnBuildToggleBehavior(self, modifiers)
+    if not self._OnButtonToggle then
+        self._OnButtonToggle = function(self, stat, state)
+            SimCallback( { Func="SetStatByCallback", Args= {[stat] = state}}, true )
+        end
+    end
+    
+    if self._popup then
+        self._popup:Destroy()
+        self._popup = nil
```
4z0t, line 878: return here instead of nesting.

Garanas:
@clyfordv I'm going to push this pull request into the new year. It's a bit too close to the release at the moment

Garanas:
@clyfordv I may be a bit lost, but where is the menu you're referring to?

Garanas:
I'm not sure; I'll check and confirm 😃

clyfordv:
Menu is in the same spot as the firestate popup (maybe I should highlight that it is a popup, and it isn't just swinging in the wind whenever the order grid is present).

MrRowey:
@clyfordv Whats the status on this ?

BlackYps:

PR #5523, opened 2023-10-10
Mute scoreboard background ambient if music is off

Pokute:
Currently there is no way to set volume for ambient sounds. (I did try `SetVolume("Ambient"/"AmbientTest", 0)`). Now at least won't play the ambient sound in scoreboard if music is off.


Garanas:
Seems fine enough 😃

PR #5524, opened 2023-10-11
Scoreboard graph optimization and cleanup

Pokute:
For some weird reason the scoreboard performance is significantly better if the bitmaps are stored and accessed through a temporary variable. Also moved the mouse hover information code to start so it is visible without having to wait full graph generation.

Following are scoreboard comparison videos of the same 5min 5vs5 AI game. The graph generation doesn't change much, but UI responsivity is significantly better:

Viewing the embedded 4k videos might not work. Re-encoding them into into 720p with worse quality results in file sizes too large to upload here. They should be viewable when saved using a videoplayer that supports AV1.

Old: `https://github.com/FAForever/fa/assets/5271115/a9fa6464-3288-4d32-9336-e99e82313ded`

https://github.com/FAForever/fa/assets/5271115/a9fa6464-3288-4d32-9336-e99e82313ded

New: `https://github.com/FAForever/fa/assets/5271115/a6e8918e-4cf2-4a87-ae65-71de59c26128`

https://github.com/FAForever/fa/assets/5271115/a6e8918e-4cf2-4a87-ae65-71de59c26128

### DIFF
--- a/lua/ui/dialogs/hotstats.lua
+++ b/lua/ui/dialogs/hotstats.lua
@@ -778,14 +778,64 @@ function create_graph(parent,path,x1,y1,x2,y2)
         inc_periode=((data_nbr)/(x2-x1)) -- give the increment on the screen between each periode (i.e. between each saved)
         if inc_periode<1 then inc_periode=1 end -- can not be <1 => use the whole screen
         local nbr=0 -- couting the number of iterancy done
+
+        -- for the windows under the mouse on the background
+        local hoverInfo = Group(grp)
+        hoverInfo.Left:Set(0)
+        hoverInfo.Top:Set(0)
+        hoverInfo.Right:Set(50)
+        hoverInfo.Bottom:Set(30)
+        hoverInfo:DisableHitTest(true)
+        local infoPopupbg = Bitmap(hoverInfo) -- the borders of the windows
+        infoPopupbg:SetSolidColor('white')
+        infoPopupbg:SetAlpha(.6)
+        infoPopupbg.Depth:Set(function() return hoverInfo.Depth()+1 end)
+        infoPopupbg.Width:Set(function() return hoverInfo.Width()+2 end)
+        infoPopupbg.Height:Set(function() return hoverInfo.Height()+2 end)
+        infoPopupbg.Left:Set(function() return hoverInfo.Left()-1 end)
+        infoPopupbg.Bottom:Set(function() return hoverInfo.Bottom() + 2 end)
+        local infoPopup = Bitmap(infoPopupbg)
+        infoPopup:SetSolidColor('black')
+        infoPopup:SetAlpha(.6)
+        infoPopup.Depth:Set(function() return hoverInfo.Depth()+2 end)
+        infoPopup.Width:Set(function() return infoPopupbg.Width() - 4 end)
+        infoPopup.Height:Set(function() return infoPopupbg.Height() - 4 end)
+        infoPopup.Left:Set(function() return infoPopupbg.Left() + 2 end)
+        infoPopup.Bottom:Set(function() return infoPopupbg.Bottom() - 2 end)
+        local infoText = UIUtil.CreateText(hoverInfo, '', 14, UIUtil.titleFont)
+        infoText:SetColor("white")
+        infoText:DisableHitTest()
+        infoText.Left:Set(function() return hoverInfo.Left() + 4 end)
+        infoText.Bottom:Set(function() return hoverInfo.Bottom() - 4 end)
+        infoText.Depth:Set(function() return hoverInfo.Depth()+3 end)
+                --displays the value when the mouse is over a graph
+        bg.HandleEvent = function(self, event)
+            -- return
+            local posX = function() return event.MouseX end -- - bg.Left() end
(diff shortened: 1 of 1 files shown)


Garanas:
Nothing strange about this! Forged Alliance ships with a Lua interpreter. Unlike a compiler, the interpreter literally executes what is written. No questions asked, no edits made. Take this snippet:

```lua
while (x<(parent.Left()+ x1 + nbr*dist) and x<(x2+parent.Left()))  do
    for name,data in line do
        graph[data.index][x-x1]=Bitmap(grp)
        graph[data.index][x-x1].Left:Set(parent.Left() +x)
        if data.y_factor != 0 then
        local yn=parent.Top() +data.y+data.y_factor/math.abs(data.y_factor)*size*((math.abs(data.y_factor)+1))
            if data.y_factor<0 and (yn+size)<data.yb then yn=data.yb-size end
            if data.y_factor>0 and (yn-size)>data.yb then yn=data.yb-size end
            graph[data.index][x-x1].Top:Set(yn)
        else
            graph[data.index][x-x1].Top:Set(parent.Top() +data.y-size)
        end
        graph[data.index][x-x1].Right:Set(graph[data.index][x-x1].Left() +size)
        graph[data.index][x-x1].Bottom:Set(parent.Top() +data.y)
        graph[data.index][x-x1]:SetSolidColor(data.color)
        graph[data.index][x-x1]:Depth(bg.Depth()+5)
        data.y=data.y+data.y_factor
    end

    -- (...)

end
```

Let us start simple and zoom in on this specific line:

```
graph[data.index][x-x1]=Bitmap(grp)
```

When you turn this into Lua bytecode it will look something like this:

```
MOVE
GETTABLE
GETTABLE
GETUPVALUE
MOVE
FUNCTION
SETTABLE
```

What matters here are the `GETTABLE` operations: they are expensive! These `GETTABLE` operations are repeated for each line that contains `graph[data.index][x-x1]` because we're dealing with an interpreter. What you did here is 'cache' the result once and re-use it, which is what a compiler would usually do for us.

As a result, by caching the value you're effectively culling a lot of table operations. Table operations involve a hash check and various other memory operations (pointer work, etc) which can easily miss the cache too. 

What you did here is what we've been doing to the code base for the past two years. And what the contributors at LOUD have been doing to the code base for a lot longer than that 😃 

For a good read, see the following paper:

- https://www.lua.org/doc/jucs05.pdf

And specifically chapter 4 and 7 and figure 5.

lua/ui/dialogs/hotstats.lua
```diff
@@ -801,23 +851,25 @@ function create_graph(parent,path,x1,y1,x2,y2)
...
             while (x<(parent.Left()+ x1 + nbr*dist) and x<(x2+parent.Left()))  do
                 for name,data in line do
-                    graph[data.index][x-x1]=Bitmap(grp)
-                    graph[data.index][x-x1].Left:Set(parent.Left() +x)
+                    local bitmap = Bitmap(grp)
+                    graph[data.index][x-x1] = bitmap
+                    bitmaps = bitmaps + 1
+                    bitmap.Left:Set(parent.Left() +x)
                     if data.y_factor != 0 then
                     local yn=parent.Top() +data.y+data.y_factor/math.abs(data.y_factor)*size*((math.abs(data.y_factor)+1))
```
Garanas, line 861: You can likely speed it up more by defining `local MathAbs = math.abs` at the top of the file and then using that instead of `math.abs`. Why this works is because `MathAbs` would be an `GETUPVALUE` bytecode, where as `math.abs` is a `GETGLOBAL` and then a `GETTABLE` bytecode.

Upvalues are stored in an array, as part of the closure of the function. They're cheaper to access than a `GETTABLE`, and a lot cheaper to access than both a `GETGLOBAL` and a `GETTABLE`

Then, we can take it further!

Next to the upvalue scope you also have the local scope. The local scope is part of the stack and is therefore by far the easiest to access. Right before the while loop you can add:

```
local Bitmap = Bitmap
local MathAbs = MathAbs
```

There are no other global-like values in this specific while loop. This will change their scope from global (Bitmap) or upvalue (MathAbs) to local which is even faster to access. 

Looking through this gigantic function, it can probably use a few more of these optimalisations. But let's keep those for a separate pull request and first focus on merging in these changes 😃

Garanas:
And yes - the embedded 4k video does not work on my system 😃

Pokute:
The mysterious part is the speed up after the graph has been generated. Outside those loops. I guess it's something about better memory layout for displaying those tens of thousands of bitmaps.

Garanas:
@Pokute if we merge this, will that create conflicts with your current work on performance?

Pokute:
I updated the pull request with another commit on top. I don't see much point in having these two commits merged separately. The code is mostly code cleanups, except for the innermost loops of graph renderings which are completely re-implemented.

lua/ui/dialogs/hotstats.lua
Pokute: Thanks for the info. The code is now so fast that further optimizations don't make much sense.

Garanas:
Great work @Pokute ! The graph is significantly faster and the thicker line looks better in general 😃

PR #5525, opened 2023-10-11
Processing UEF tactical missiles

Garanas:
Similar to #5515, but then for UEF tactical missiles. This pull request contains a wide range of changes. It implements the changes made in #5433 and applies uef-specific features to uef tactical missiles. 

## Gameplay related changes

UEF tactical missiles will try to fly in a large arc. In the following screenshot, red are UEF tactical missiles and purple are Seraphim tactical missiles. Unlike other factions, UEF tactical missiles will hammer the target from above and this makes them, on average, more effective against the Aeon TMD.

image

The UEF mobile missile launchers (`url0111` and `xel0307`) have an increased damage radius from `0.5` to `1.0`. The cruiser has a decreased damage radius from `3` to `2`. This was done to mimic the explosions / graphics of the missiles. 

## Graphics related changes

UEF tactical missiles now have  `OnKilled` (when shot down) and an `OnImpactNone` (when caught by Aeon TMD, when lifetime runs out) effects. They'll also let lose of a little bit of debris when they end up in either state. The debris represents fragment(s) of the missiles dropping down onto the planet.

https://github.com/FAForever/fa/assets/15778155/d4f1d9c2-deab-42c5-99fc-98607d304c58

The explosions have been increased / decreased in size where applicable to match the damage radius of the weapon.


clyfordv:
For the UEF (and looks like Aeon) missiles, is there a reason for setting LaunchTicks so high (16-24 instead of ~4-6 for the others)? It tends to make them drastically overshoot at short range.

units/UEL0001/UEL0001_unit.bp
```diff
@@ -1071,7 +1071,7 @@ UnitBlueprint{
             RackRecoilDistance = 0,
             RackReloadTimeout = 10,
             RackSalvoChargeTime = 0,
-            RackSalvoReloadTime = 30,
+            RackSalvoReloadTime = 0,
```
lL1l1, line 1074: This buffs billy nuke reload from 30 to 5 seconds!

PR #5526, opened 2023-10-11
Fatboy: Increase its BackUpDistance from 15 to 30

Basilisk3:
**Changes:**
Fatboy: BackUpDistance 15 --> 30

Same reasoning as in #5472.


Garanas:
Calling it 😃

PR #5527, opened 2023-10-11
Decrease the cost of the UEF ACU's Shield Field upgrade and increase its hitpoints

Basilisk3:
**Changes:**
ShieldMaxHealth: 9000 --> 11000
BuildCostEnergy:  112500 --> 101800
BuildCostMass: 2300 --> 2100
BuildTime: 2100 --> 1900

Splitting the two shield upgrades was a good decision, though with its current stats, the Bubble Shield upgrade is not really competitive with the Personal Shield, resulting in it almost never being picked over of the personal shield. These changes enable the player to build it slightly earlier in a match, making it more attractive.


Pokute:
Need more context: Why these exact values? How it compares to the personal shield? How it compares to Aeon ACU shield and maybe SACU shields?

Pokute (approved):
The value changes themselves are simple enough that and match the pull request description.

Basilisk3:
The values were chosen so the consumption figures for the upgrade stay around the same (in this case, they are almost identical to the personal shield upgrade). For example, both the shield field and personal shield will drain 535 energy while upgrading them with a tech 1 ACU. I think this is more intuitive. The value changes amount to around a 10 % buff. I will not post detailed comparisons to the other shield upgrades here because 1) There is only one shield field upgrade for ACUs, and comparing stats between the two upgrades is not all that useful and 2) The balance team knows the stats, and which upgrades are good anyway.

ComradeStryker:
If I may give my input:

The Bubble Shield's SHP-to-Mass ratio is still absurdly low when you compare it to the Personal Shield's ratio.
Bubble Shield (Current): 9,000 HP for 2,300 (3.913)
Bubble Shield (Suggested): 9,000 HP for 2,100 (4.285)
Personal Shield: 19,000 HP for 2,300 (8.261)

Even with the very slight decrease in mass and build time, the tradeoff between Bubble and Personal Shield is still far too great.
Personal Shield will still take precedence over Bubble Shield.
For 200 mass, you more than double the HP of your Shield.
No one would pass that up. Saving 200 mass is not worth losing out on 10,000 HP.

If I may offer, a more realistic approach would be to have Bubble Shield at around 12,000 SHP for a Mass cost of 1,500.
This would grant it a SHP-to-Mass ratio of 8.0. 
Slightly worse than the Personal Shield's ratio and stats (7,000 less HP), but still enough to make it a lucrative upgrade when compared to the Personal Shield.
These stats keep the Shield upgrade competitive with the Aeon ACU's shield upgrade as it will be more expensive, regardless.

---

~ Stryker

Basilisk3:
@ComradeStryker I was conservative with this PR because I need to know the Balance Team's opinion about this first. If the balance team agrees with a larger buff, I am happy to tweak the numbers further. That said, I will do more testing and most likely increase its hit points some more. 

Your suggestion of 12000 hit points at a mass cost of 1500 would be too much though. That is cheaper than Seraphim nano repair, while providing a strong and relatively large shield, that has 60 hp/s regen, and a recharge time of 90 seconds.

ComradeStryker:
That is fair.
The stats in my example are only a reference to the same ratio the current Personal Shield has.
And as Bubble and Personal shield are two different uses, it would make sense the costs and ratio would not be the same.

But, if I may also say, that the point of the Bubble shield is to protect the units underneath, and at a current HP of 9,000, it doesn't do much other than offer a few seconds of extra protection.
I've conversed with quite a number of players since the change and we all agreed that the Shield HP could use a drastic uplift in HP, at least.
Looking forward to that.

---

:)

---

~ Stryker

Basilisk3:
>  That is cheaper than Seraphim nano repair, while providing a strong and relatively large shield, that has 60 hp/s regen, and a recharge time of 90 seconds.

For the sake of context, the Personal Shield has 30 hp/s of regen (which is a pretty low amount) and needs 140 seconds to recharge its shield after a full collapse. These are big differences, so adding too much hp to the field or making it too cheap, will quickly result in it becoming overpowered.

MrRowey:
The balance team has a different approach here #5590

MrRowey:
@Tagada14 thoughts on this one as the one you proposed has been done

MrRowey:
Closing PR in favor of the one Balance Team wanted

Issue #5529, opened 2023-10-12
Terrain shader naming scheme


BlackYps:
This issue is to discuss naming conventions for new map shaders.
There is a load of different features that we can make available to mappers, but they might not want to use all of these at once. For example some things require more work than others if you want to transfer an existing map to a new shader.
We started out with a numeric scheme which sounds reasonable as we need all the names to be the same length to make switching them easier.
We can extend the naming scheme as follows. Note that not all permutations are necessarily available or even planned, but I think having some naming rules makes it easier to later select a specific shader than just incrementing the number every time we create a new terrain shader.

We have three digits to work with, the first digit shows the main category
- Terrain0XX for shaders that are pretty regular
- Terrain1XX for shaders using roughness maps
- Terrain2XX for shaders using advanced splatting
- Terrain3XX for shaders using roughness and advanced splatting

The second digit gives info about some technical properties of the shader with the range being divided into two halfes:
- TerrainX0X to TerrainX4X for shaders with half mask range
- TerrainX5X to TerrainX9X for shaders using the full mask range

For this we can implement a boolean toggle in the shader to avoid code duplication.

We can then use individual digits for different properties, for example:
- TerrainX1X and TerrainX6X for shaders using biplanar mapping?
- TerrainX2X and TerrainX7X for shaders using additional rotated sampling? (This is for breaking up the texture repetition)
- TerrainX3X and TerrainX8X for shaders using biplanar mapping and additional rotated sampling?

This leaves us with the third digit to have enough room to create multiple shader variations of the same category if needed.
For example we could variate in which layer the biplanar mapping is applied or on how many layers.

In this setup TerrainX3X and TerrainX8X are still unused and could be used for a category that I haven't thought of yet

Garanas:
I've got nothing to add; this seems like a great suggestion to me.

PR #5533, opened 2023-10-12
Terrain advancements

BlackYps:
I had to partially revert the fix that got rid of the water coloration on steep cliffs. When only checking for the terrain height we can get artifacts in shallow water when zooming out far because of the terrain tesselation. Then random segments of the water disappear. So as a compromise we only get rid of the start of the coloration. This is enough for reasonably steep cliffs while it minimizes the tesselation artifacts to a level that is not distracting.
Screenshot 2023-10-12 234552

The advanced shaders now use the previously unused channel in our mapwide texture for the water depth. This enables us to put the mapwide albedo "decal" into a normal slot that was already free, but we couldn't access it because of the limit of how many different textures we can sample. Now that we got rid of the need to sample the water depth texture that the engine provides, we can sample that normal slot.

The areas underwater now get their specularity reduced instead of their roughness increased, which is more physically accurate.

According to the naming scheme in #5529 I added shaders that utilize the full mask range and I added a shader that uses the pbr workflow with roughness maps without implementing the advanced splatting, which makes it much easier to use this shader with the current map editor.

The map generator will soon provide tools to easily generate the needed mapwide and pbr texture in https://github.com/FAForever/Neroxis-Map-Generator/pull/365

### DIFF
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -391,28 +391,31 @@ float ComputeShadow( float4 vShadowCoord )
 // apply the water color
 float3 ApplyWaterColor(float terrainHeight, float depth, float3 inColor)
 {
-    if (depth > 0.1 || terrainHeight <= WaterElevation) {
+    if (depth > 0) {
+        // With this extra check we get rid of unwanted coloration on steep cliffs when zoomed in,
+        // but we prevent that terrain tesselation swallows too much of the water when zoomed out
+        float opacity = saturate(smoothstep(10, 200, CameraPosition.y - WaterElevation) + step(terrainHeight, WaterElevation));
         float4 wcolor = tex1D(WaterRampSampler, depth);
-        inColor = lerp( inColor.xyz, wcolor.xyz, wcolor.w );
+        inColor = lerp(inColor.xyz, wcolor.xyz, wcolor.w * opacity);
     }
     return inColor;
 }
 
-float3 ApplyWaterColorExponentially(float3 viewDirection, float terrainHeight, float waterDepth, float3 color) {
-    // With this extra check we get rid of unwanted coloration on reasonably steep cliffs,
-    // but we prevent that terrain tesselation swallows too much of the water when zoomed out
-    if (waterDepth > 0.1 || terrainHeight <= WaterElevation) {
+float3 ApplyWaterColorExponentially(float3 viewDirection, float terrainHeight, float waterDepth, float3 color)
+{
+    if (waterDepth > 0) {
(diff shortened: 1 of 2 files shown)


Garanas:
Do you want these in for the 28th?

BlackYps:
There is no rush. You can also postpone it to later if that is easier for you

BlackYps:
Oh, the water fix should make it in, but I could extract that into a separate PR

Garanas:
It's all fine to me; do you have a (generated) map or two that I can use to review your changes?

BlackYps:
Do you think you will find time to review this before the release?

Garanas:
Noticed the effect you mentioned in the screenshot; it appears to be not symmetrical:

image

image

I noticed this before too. Any idea why this is?

With regards to the sharp transitions:

image

Isn't it maybe better to revert to what we had before? At least then it visually appears smooth

BlackYps:
It's probably because the heightmap is not completely symmetrical. It's a gpg map after all that originally also had asymmetric resource markers.

If we revert the water changes we get this again, which is a side effect of the new exponential water absorption. Unfortunately there seems to be no perfect solution. I think the best solution is to do it like in this PR and to expect map makers to not make insanely steep cliffs.
image

Garanas:
Ah yes; I forgot about those. And is a solution where we take into account the camera position viable?

BlackYps:
What are you thinking of?

Garanas:
Tesselation only shows up as you're zoomed out. The rough edge / transition this pull request introduces only shows up when you're zoomed in. If I'm not mistaken then we have the position of the camera. Maybe we could use the Y-value as a linear interpolation to get best of both worlds

BlackYps:
Ah yeah, that sounds like it could work

BlackYps:
But how do I interpolate a boolean decision?

BlackYps:
That was quite tricky to pull off, but I did it!

Garanas:
I think that the grass should be less shiny but that can be controlled by the input if I'm not mistaken. Note that maps that do not use the new shaders remain unaffected.

image

image

image

image

Garanas:
And with thanks to the last bit of work the water ramp artifacts on steep cliffs are less noticeable:

image

image

image

Garanas:
@BlackYps also a huge fan of using the new water textures 😃 , the only thing the map is missing at this point are waves

BlackYps:
The map is excessively shiny because I forgot to adjust the mask of the highest texture layer. That one adjusts the roughness, so it is indeed controllable by the map.

PR #5534, opened 2023-10-13
AI - state machine fixes and guardrails

relent0r:
This PR does the following.

Removes double ups of BuildStructure commands for the initial ACU build out when a position in unavailable. This avoids the ACU function from failing.
Adds a platoon merge for state machines.
Add platoon merging to state raid and attack platoons to assist in them forever being too weak to attack positions.
Adds guardrails to various state machines to decrease dead loop issues where a result is invalid for various reasons caused it to constantly switch states.
Fix issues with state machines that use GetPlatoonUnits as this now returns two objects instead of one.

Also adds more logic to filtering of allied bases so that AI are less likely to build expansions on team mates expansions, be it AI or human.

Testing.
Run AI games using default random ai selection and observe results.


relent0r:
Moving back to draft. Found issue related to the guard state machine and this particular piece of code in the engineerbuildai.

        if cons.BuildStructures[1] == 'T1Resource' or cons.BuildStructures[1] == 'T2Resource' or cons.BuildStructures[1] == 'T3Resource' then
            relative = true
            closeToBuilder = eng
            local guards = eng:GetGuards()
            for k,v in guards do
                if not v.Dead and v.PlatoonHandle and aiBrain:PlatoonExists(v.PlatoonHandle) then
                    v.PlatoonHandle:PlatoonDisband()
                end
            end
        end

lua/aibrains/platoons/platoon-adaptive-raid.lua
```diff
@@ -194,6 +195,7 @@ AIPlatoonAdaptiveRaidBehavior = Class(AIPlatoon) {
             local destination = self.LocationToRaid
             if not destination then
                 self:LogWarning(string.format('no destination to navigate to'))
+                WaitTicks(20)
```
Garanas, line 198: There's a wait statement when platoons change state, there should be no need to wait here

relent0r: I'm doing this intentionally as it indicates something usually went wrong with the raid location and time a delay before it researches may resolve it.

Issue #5399, opened 2023-09-05
Seraphim Sniper Switching Modes Fires Wrong Projectile


lL1l1:
**Describe the bug**
Switching modes during a reload makes the sniper fire the wrong projectile. It reloads and move likes the new mode, but still fires the old mode's projectile.

**To Reproduce**
Spawn a sniper and a nano SACU as a target.
After the sniper shoots its rapid shot, switch modes.
Notice how it shoots a rapid shot but reloads and moves slowly like with a powerful shot.
Switch modes during the long reload, and it will then fire a powerful shot but reload and move quickly.

**Expected behavior**
After shooting a rapid shot and switching modes, it should shoot a powerful shot, slow down, and reload for a long time.
After shooting a powerful shot and switching modes, it should shoot a rapid shot, speed up, and reload quickly.

K-ETFreeman:
If you press switch modes in short-reload state, it first finishes the short reload, then shoots rapid shot, then starts to prepare for a long shot
If you press witch modes in long-reload state, it will still moves slowly, and after heavy shot, he switches to "rapid fire" state
Everything is working as intended, it finishes the current shot then switches mode
Otherwise you will be able to get free short extra dps boost by switching the mode from rapid to heavy fire, which is probably not what balance team wants, i suppose

lL1l1:
> first finishes the short reload, then shoots rapid shot, then starts to prepare for a long shot

If it was "preparing" for a long shot, then switching modes would force the sniper to reload, regardless if it is currently reloading or not. This would mean that the correct behavior would be:
- Not in combat: Pressing switch modes forces a reload of the new mode.
- In combat: Rapid shot -> 3 second reload -> Realize you are switching modes (cannot shoot, or else you ought to reload 3 seconds) -> 15 second reload -> powerful shot
> it finishes the current shot then switches mode 

Since it fires the rapid shot and then does a 15 second reload to "prepare" without doing a 3 second reload, you are incorrect that it "finishes the current shot".
> Otherwise you will be able to get free short extra dps boost by switching the mode from rapid to heavy fire

You can still get a dps boost by starting combat in snipe mode and switching to rapid mode during the first reload:
Powerful shot -> 15 seconds reload -> powerful shot -> 3 seconds reload -> rapid shots/3 second reload
So it's clearly intended that it is a "reload" and not a "preparation". This means that the reload should be an effect of the shot fired before, which does not currently always happen because the firing mode is not updated at the proper time.
Here's a new example I found:
[...]

PR #5537, opened 2023-10-14
Rework sera sniper mode logic

lL1l1:
Now it is more responsive and logical, resolving #5399, and reworking #5196.
Previously it saved the next firing mode's BP and immediately applied only its fire rate to the weapon, and only switched the rest of the stats immediately after firing or if the switch modes button was pressed when the weapon was fully reloaded. This lead to unexpected behavior from the player's point of view.
Now it switches stats immediately, except for movement speed which is synced with the reload time, since the reload and movement reduction are both drawbacks of sniper mode shots, and without syncing you could ignore the movement penalty by toggling snipe mode off then on in between every shot.
In addition, it is easily possible to reduce micro needed to use sniper mode by moving the `ScheduleMovementChange` function from `SetSniperMode` to the MainGun's `RackSalvoFiringState Main` function, so that there is no need to micro sniper mode snipers out of combat for movement speed (the movement penalty would only apply during the long reload).

### DIFF
--- a/units/XSL0305/XSL0305_script.lua
+++ b/units/XSL0305/XSL0305_script.lua
@@ -52,7 +52,6 @@ XSL0305 = ClassUnit(SLandUnit) {
     ---@param mode boolean
     ScheduleMovementChange = function(self, delaySeconds, mode)
         if self.isSniperMoveMode == mode then return end
-        LOG("sniper move mode " .. tostring(self.isSniperMoveMode) .. "   changing mode " .. tostring(self.isChangingMoveMode))
         local switchToMode = function(mode)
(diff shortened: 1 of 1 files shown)


Pokute:
It works as advertised. Great work!
Code could use a little bit of clarification so while I read through it I made this PR: https://github.com/lL1l1/fa/pull/2

Pokute:
Oh yeah, the tooltip for the sniper mode toggle, and maybe the sniper itself could use some changes. They don't mention the speed drop at all.

Garanas:
I'll try to review this in the next two days 😄 !

Garanas:
Great work @lL1l1 👍

PR #5540, opened 2023-10-14
Assist to copy the command queue of an engineer

Garanas:
A small feature for the list of the 'assist to' features. With this change you can assist an engineer but instead of applying an assist order you copy and apply the command queue of that engineer to your selection.

This is similar to the [Copy orders](https://wiki.faforever.com/en/Play/Game/Hotkeys/OrdersQueueManipulation#copy-orders) feature except that this does not require a hotkey 😃

### DIFF
--- a/lua/options/options.lua
+++ b/lua/options/options.lua
@@ -510,6 +510,20 @@ options = {
             },
 
             {
+                title = "<LOC ASSIST_TO_COPY_COMMAND_QUEUE>Assist to copy command queue",
+                key = 'assist_to_copy_command_queue',
+                type = 'toggle',
+                default = 'Off',
+                custom = {
+                    states = {
+                        { text = "<LOC _Off>Off", key = 'Off' },
+                        { text = "<LOC _ASSIST_TO_COPY_ENGINEERS>Only engineers",
+                            key = 'OnlyEngineers' },
+                    },
+                },
+            },
+
+            {
                 title = "<LOC OPTIONS_0287>Factories Default to Repeat Build",
                 key = 'repeatbuild',
                 type = 'toggle',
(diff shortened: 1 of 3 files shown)


Garanas:
At the moment it requires the use of a `control` key. This enables you to still issue a regular assist order (by not holding `control`)

Pokute:
You want to make the earlygame super busy too? Now I have to create every unbuilt engineers' build order on my ACU and then copy it to the completed engie and move on to creating next engineer's build orders on the ACU.

I like it.

Pokute (approved):
I like the functionality. It works for copying orders (except factory reclaim) and I it doesn't copy build orders that can't be made. Cybran engies won't copy orders to build Aeon buildings, sparkys won't copy orders to build power gens etc.

There are three smaller issues I found:
* It gives the message "copied 5 orders" even if it couldn't copy any of them. Although checking which orders it could copy and which orders it couldn't would be difficult.
* having a group of engies and copying the orders wildly between them can result in missing building animation and progress bar on the engie (had buildspeed and cybran builder drones of all the engies), but I only got this once for many tries so might've not been due to this feature.
* I would like to append the orders too, so shift-ctrl-assist would append the copied orders after existing orders.

Pokute:
Also, didn't realize at first that I had to enable the feature from options.

Garanas:
We'll take your suggestions in with the next pull request on this feature 👍

Garanas:
> I would like to append the orders too, so shift-ctrl-assist would append the copied orders after existing orders.

This one is tricky 🤔 , I'm not sure how to fix that. I can reproduce it quite easily by having a tech 1 engineer copy the queue of a tech 3 engineer

Garanas:
https://github.com/FAForever/fa/assets/15778155/2a7e764e-c79a-4d15-9577-4ead4478cf09

PR #5541, opened 2023-10-14
Make it easier to start building extractors while zoomed out

Garanas:
When using context based templates you can start creating extractors by hovering over a deposit. The radius used to be fixed to 1.5 (extractors) and 3.0 (hydrocarbon plants). It now includes a factor based on the zoom of your camera. This makes it easier to cycle through extractors and / or a hydrocarbon plant based templates while zoomed out.


Pokute (approved):
Works as advertised.

PR #5545, opened 2023-10-15
Process the balance changes to walls

Garanas:
In addition to https://github.com/FAForever/fa/issues/5514 and because of https://github.com/FAForever/FA-Binary-Patches/pull/48 where we tackled the order of collisions for projectiles.

For all walls:

- Hitpoints: Reduced from `X` to `500`
- Energy cost: Increased from `10` to `20`
- Mass cost: Increased from `2` to `3`
- Build time: Reduced from `20` to `15`


Blodir:
Changed milestone to balance iteration II

PR #5551, opened 2023-10-15
T2 Bomber speed nerf

Basilisk3:
**T2 Bombers**
MaxAirspeed = 15 --> 14.25

T2 Bombers were as fast as their counter (interceptors) which meant that, in some scenarios, interceptors would never be able to catch them. This nerf increases the skip-cap of using T2 Bombers and enables better counter play for the defending player.


MrRowey:
Neededs rebasing to fafdevelop

PR #5554, opened 2023-10-16
Fix: Enable the Sparky to be buffed by the RegenAura-upgrades

Basilisk3:
This was not the case previously.

### DIFF
--- a/units/XSL0001/XSL0001_unit.bp
+++ b/units/XSL0001/XSL0001_unit.bp
@@ -284,7 +284,7 @@ UnitBlueprint{
             RegenFloor = 15,
             RegenPerSecond = 0.035,
             Slot = "LCH",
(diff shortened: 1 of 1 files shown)


Garanas:
What stops us from enabling all mobile units?

Basilisk3:
> What stops us from enabling all mobile units?

Correct me if I'm wrong, but doesn't `BUILTBYTIER3FACTORY` already mean that? Ignoring units like drones, all mobile units that aren't T4s can be built by the T3 factories. It just doesn't work for a couple of units, at least that is my understanding of it.

Basilisk3:
@Garanas Enabling all mobile units would mean ACUs get buffed, inlcuding your own. So we would need to introduce exceptions, which probably isn't worth it. This PR is ready to be merged.

Garanas:
You're right; it is likely setup this way to exclude ACUs.

PR #5556, opened 2023-10-16
Apply `NOSPLASHDAMAGE` category to sera MML missiles.

lL1l1:
Prevents missiles from killing eachother, like other faction MML (#5203).


Garanas:
Woops - merged into the wrong branch 😃 . It should be either `deploy/fafdevelop or `deploy/fafbeta`, my mistake for not spotting it

PR #5558, opened 2023-10-17
Fire Beetle Target Acquisition and Chase Behavior

clyfordv:
Stationary fire beetles and those with attack orders will blow up as usual.


Garanas:
Closed in favor of https://github.com/FAForever/fa/pull/5628

PR #5560, opened 2023-10-17
Increases size of mongoose grenade strategic icon

clyfordv:
Microfix that allows the grenades to be distinguished from the gatling gun bullets in strategic view.

image


Garanas:
This is a great idea; perhaps we could apply this to other units too in future PRs. Take as an example the Seraphim tech 3 Siege Tank and the Cybran tech 3 Assault Bot

clyfordv:
Yes, this was a quickie but any unit shooting multiple types of projectiles with varying fire rates is a good target for improvement.

PR #5562, opened 2023-10-18
Sniper mode description mentions slowdown.

lL1l1:


Garanas:
Are there no translations in other languages? If so; those should either be updated too or removed to prevent them being out of sync

lL1l1:
I would need help translating to other languages.

Garanas:
I understand; it was more of a question on how to process these type of changes

PR #5564, opened 2023-10-20
Fix aeon bomber getting stuck flipping

lL1l1:
Copies sera bomber values. I noticed you have to f10 -> "restart game" to have the `SizeX/Y/Z` changes apply.
Here are two pictures with tests of ~100 bombers.
Before (all the bombers flying low are stuck flipping over trying to aim at the move waypoint):
OldAeonBomber
After (the bombers don't fly nearly as low nearly as often, and they correct their trajectory+height quickly):
NewAeonBomber

### DIFF
--- a/units/UAA0103/UAA0103_unit.bp
+++ b/units/UAA0103/UAA0103_unit.bp
@@ -178,9 +178,9 @@ UnitBlueprint{
     SelectionSizeZ = 0.45,
     SelectionThickness = 0.24,
     SizeSphere = 1.6,
-    SizeX = 2,
-    SizeY = 0.25,
-    SizeZ = 0.625,
+    SizeX = 1.4,
+    SizeY = 0.2,
+    SizeZ = 0.8,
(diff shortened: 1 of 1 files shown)


Garanas:
That's great! But I completely lack the understanding why these changes would impact the behavior of the bomber in such a way; do you have any idea why?

lL1l1:
I didn't find any relevant references in the lua files; really I just saw that sera bomber worked and started copying values. 
My hypothesis is that, in the engine, banking uses the size to turn, but then that behavior disables its upwards thrust because it is a winged aircraft. But that doesn't explain why it only flips out when there are enemy targets nearby, so maybe it's also some conflicting behavior from an attack pattern. Just speculation; I have no experience with the engine or flight sim math.

Garanas:
Confirmed 👍 , great fix! What name would you like to end up in the contributors log?

lL1l1:
IL1I1 will be fine.

PR #5574, opened 2023-10-24
Nav mesh/performance 01

Garanas:
Closes https://github.com/FAForever/fa/issues/4314

Re-implements the data structure of the navigational mesh. The idea originates partially from using the optimalisation 'structures of arrays' instead of 'arrays of structures'. The other part originates from how one can map a heap into both a tree and an array. The same can be applied to quad trees. See also:

- https://en.wikipedia.org/wiki/AoS_and_SoA
- https://en.wikipedia.org/wiki/Heap_(data_structure)

Also implements the optimalisation as described in #4313 and the general good-practice of writing high performing code. The results are quite impressive! On average the navigational mesh generates 30% to 40% faster. The memory footprint is not significantly adjusted.

The great part about these changes is that the performance improvements will 'leak through' to any other logic that interacts with the navigational mesh, including all of the utility functions in [navutils.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/NavUtils.lua). The interface remains unchanged; if your code does not directly interact with the navigational mesh itself then no changes are required.

_note that you're highly discouraged to interact with the navigational mesh directly. You're better off using the interface in [navutils.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/NavUtils.lua) or add to that interface with a pull request_


Garanas:
_Strip mine (20x20, land only)_


image

Garanas:
_Sung island (20x20)_

image

Garanas:
_Seton's Clutch (FAF version)_

image

Garanas:
Separate but also a nice addition; the rendering of the layers now also show unpathable terrain as red:

image

Small change but it makes it a lot easier to analyse the map 😃

PR #5575, opened 2023-10-25
Fix Various Weapon Tooltips and Display Weapon Costs

lL1l1:
#### UI Changes:
Make sure to enable the "display weapon armaments in build menu" game option.
Display of salvo size and missile costs (cost is a darker color to be distinct, but not distracting):
image
Beam weapons display damage of the full beam life, and previously identical weapons are renamed so they split properly:
image
Shields display correct radius:
image
Display weapon charging costs:
image
and other misc. changes.
#### Gameplay changes:
Torrent's front missile launcher min range was increased from 40 to 60. It can't hit reliably at that close range, and it seems to have been left unincreased by accident according to the commit history.
There is a bug fix for mobile TML and billy fire rates being unlimited through manual re-groundfiring.

There is still a bug with the reload formula being incorrect sometimes (e.g. bouncer shows 100 dps instead of 210 and it doesn't round fire rate to game ticks), but I want to request these changes first.


Garanas:
These look like great changes!

> There is still a bug with the reload formula being incorrect sometimes (e.g. bouncer shows 100 dps instead of 210 and it doesn't round fire rate to game ticks), but I want to request these changes first.

Let's process these in a separate pull request. I also noticed https://github.com/FAForever/fa/issues/5576, is that related to what you describe here?

lL1l1:
> I also noticed #5576, is that related to what you describe here?

Somewhat, an understanding of the firing cycle would help fix that issue and create an accurate reload time formula.

units/UAL0001/UAL0001_unit.bp
```diff
@@ -896,9 +893,6 @@ UnitBlueprint{
             DamageRadius = 2.5,
             DamageType = "Overcharge",
             DisplayName = "AutoOvercharge Cannon",
-            EnergyChargeForFirstShot = false,
-            EnergyDrainPerSecond = 7500,
```
Garanas, line 900: Was there a reason to remove these fields? The overcharge toggle relies on them to exist

lL1l1: I thought all the logic for overcharge was done in overcharge.lua and OverchargeProjectile.lua, and I could simply remove the stats since they were displaying incorrectly as a weapon charge cost with the new tooltips.
It seems like only the EnergyRequired field is necessary, so I re-added that in a commit to this PR's branch just now.

Garanas: I'll include the changes in a hotfix. Where is the commit you're referring to?

lL1l1: https://github.com/lL1l1/fa/commit/45d203d95a712ce9ae3494fb938e31a1e7782d4f

Issue #4404, opened 2022-11-14
T3 artillery/ T3 bomber decals are same for every faction


semzer:
Hello. So I've encountered the problem that the t3 arty/bomber impact decals are the same for every faction. In the past every faction had their own faction specific decals.

You can join the game and spawn every faction mobile t3 arty to see it for yourself. They all have the same impact decals. Same for some experimental units like Ahwassa, its impact decal is same too.

I expect that every faction impact decals to be different and faction-specific, like they used to be.

I've attached a screenshot in which 3 different mobile artilleries are shooting the ground and the impact decals are the same for all of them (excluding seraphim mobile artillery, for some reason it wasnt able to shoot the ground)

This problem is also present in t3 bomber impact decals, stationary t3 heavy artillery impact decals, and ahwassa impact decals.

![bug](https://user-images.githubusercontent.com/75785672/201735132-026cc08b-628e-41b2-990f-7a337137c732.png)

Garanas:
I'll look into this 👍

semzer:
> I'll look into this 👍

Also, I'd look to point out that the decals are replaced with Fatboy's impact decals.

MrRowey:
@Garanas where would someone start to look at this ?

Garanas:
The projectile needs this defined in the `Display` section of the blueprint:

- https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/sim/Projectile.lua#L355

Where that references the decal to use, instead of one of the generic decals

PR #5578, opened 2023-10-25
ScorchSplats for T3 Mobile Artys

MrRowey:
Gave Each of the T3 Mobile Artys a Desytict ScorchSplat goes toward fixing #4404


Garanas:
After seeing the results I'm not sure if it improves the situation.

PR #5579, opened 2023-10-25
Introduce a coarse navigational grid

Garanas:
As an extension of #5574 we can now generate a coarse grid of the navigational mesh that we can use to navigate. The navigational mesh still exists - the coarse grid is generated on top of it.

For the average 10x10 map the setup generates in less than 0.4 seconds. For the average 20x20 map the setup generates in roughly 1.2 to 1.5 seconds. We can now use the coarse grid for navigation to speed up computations

### DIFF
--- a/engine/Sim/CPlatoon.lua
+++ b/engine/Sim/CPlatoon.lua
@@ -3,7 +3,7 @@
 
 ---@class PlatoonCommand : userdata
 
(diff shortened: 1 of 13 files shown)


Garanas:
image

image

image

image

image

image

image

image

Garanas:
image

image

PR #5581, opened 2023-10-25
Adds stat toggle infrastructure and switches AutoDeploy to use it

clyfordv:
- Adds the SimCallback and checks for the StatToggle boolean
- Switches AutoDeploy from ScriptBit to StatToggle
- DRYies out the carrier deploy/storage logic to all use an OnStopBuild function from ExternalFactoryComponent (not all external factories have storage, but all factories with storage are external factories)

### DIFF
--- a/units/UAA0310/UAA0310_script.lua
+++ b/units/UAA0310/UAA0310_script.lua
@@ -131,7 +131,8 @@ UAA0310 = ClassUnit(AirTransport, ExternalFactoryComponent) {
         end,
 
         OnStopBuild = function(self, unitBeingBuilt)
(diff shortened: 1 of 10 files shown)


units/UAA0310/UAA0310_script.lua
```diff
@@ -132,21 +131,7 @@ UAA0310 = ClassUnit(AirTransport, ExternalFactoryComponent) {
         end,
 
         OnStopBuild = function(self, unitBeingBuilt)
-            AirTransport.OnStopBuild(self, unitBeingBuilt)
```
Garanas, line 135: You still want to call the base function

Garanas:
Great work @clyfordv , I've add #5580 to this PR.

PR #5586, opened 2023-10-27
add option to show user defined templates instead of default onland/onwater templates

gordenwunderlich:
This adds an option to replace the on land / on water triggering templates with the users ingame templates.
The default templates for empty terrain are very basic (pd with walls, single aa, single torp launcher) and people often have some templates setup themselves that they still want to use preventing them from fully switching to the context based template system without first transferring their templates to the new format.
With this option (defaults to off) your own templates show whenever no other templates match (ignoring templates that match on empty terrain)
image


Garanas:
We'll include this with a hotfix. I'm wondering whether we should just default this to `enabled`

gordenwunderlich:
I defaulted to disabled to not break the currently running implementation for people using the feature, but nobody I talked to about the templates was using the water/land templates and only used the context triggering ones.

Also I also don't think the distinction between water and land templates is worth the additional effort of not being able to use the ingame tools to make the template.

If you have some good ideas for such templates then keeping it the way it is would be preferable I think.
If the templates are more like placeholders and you expect people to set up their own land/water templates, then I think defaulting to using the regular templates would be good.

MrRowey:
@gordenwunderlich Could you please update this PR with the Conflicx being fixed :)

gordenwunderlich:
@MrRowey Updated with the develop branch and I added an option to use both context base land/water templates and ingame created templates, for people that already made some templates using the new system or if people want to combine their ingame templates with special on water templates

MrRowey:
Can you provide me with a small description of the change overall so it can be added to the changlog

gordenwunderlich:
Something like:
Add option to use regular templates with the context based template hotkey.
image
You can choose between simply replacing the default land/water templates or merging them with your own templates

MrRowey (approved):

PR #5587, opened 2023-10-27
Fix error when deleting last template

gordenwunderlich:
Simple band aid solution, but I don't know what else this function is supposed to do / when it's called and if it's even possible for unitdata to be nil in other cases.


lL1l1 (approved):
It happens because when FormatData is called on line 391 the `"templates"` key does not exist
https://github.com/FAForever/fa/blob/7f2deeadbd8134dad8796ed2a57e7630d53c59e1/lua/ui/game/construction.lua#L391
That is because when the `build_templates` table is empty in game.prefs, which is what happens when you delete the last template, the `sortedOptions.templates` table is not created.
https://github.com/FAForever/fa/blob/7f2deeadbd8134dad8796ed2a57e7630d53c59e1/lua/ui/game/construction.lua#L2625-L2627

PR #5599, opened 2023-10-28
Draft: Reference enhancement stats directly through the (S)ACU's blueprints

Basilisk3:
This is not functional yet. No errors or crashes, but I cannot get the stats to display. So far I only tried it with the UEF ACU's Nano Repair upgrade.

Related to: #5566


Garanas:
I'm not fully certain, but I think @Hdt80bro did something similar with the tooltips. If you can't get it to work it may be worth asking him

KionX:
It is better to implement it similarly to LOC.

Garanas:
Could you elaborate? 😃

KionX:
Unit_Description_0006="Increases the speed at which the ACU repairs itself.\n\n+<VAR_Enhancements.DamageStabilization.NewHealth%d> Health\n+<VAR_Enhancements.DamageStabilization.NewRegenRate%d> Regen"

desc = VAR(bp, Unit_Description_0005)

Basilisk3:
Hi, thanks for the tip. Where exactly do I put this?

lL1l1:
Sorry I was incorrect before, the reason it isn't working at all is that the blueprint isn't passed in on :140.
Also you receive the enhancement's blueprint, not the unit bp.
It also seems quite annoying to create a var function, since the enhancements' effects aren't written with display in mind, so you still need to hardcode for weapon range and fire rate upgrades for example. I started on a working example in [my branch](https://github.com/FAForever/fa/compare/deploy/fafdevelop...lL1l1:fa:UI/auto-upgrade-descriptions).

Basilisk3:
That looks promising, I will close this in favour of your solution👍

Issue #5601, opened 2023-10-29
Balance idea: Rebalance Cruisers


Tagada14:
## Intended changes

Non-Sera Cruisers:
AA AoE Radius 0 > 1
Hp -600 
AA range +10

Sera Cruiser:
Hp -300
Sera 2nd AA range (45 ->70)
Sera Flak AA range (55 -> 60)

## Thoughts on the changes

We have decided to partially revert the changes to cruisers as they made them too vulnerable to T3 gunships. The higher hp also meant that sniping cruisers via non-air means became much harder which wasn't good for the gameplay. To compensate for the addition of AoE the cruisers don't get their full range back. Since the Sera cruiser already had AoE its hp is not fully reverted and its secondary (non-flak) cannon is getting a major buff. This should allow the Sera cruiser to deal with targets that are further away properly.

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

PR #5603, opened 2023-10-30
Rebalance Cruisers

MrRowey:
Changes for #5601


Tagada14:
What's up with the VS code file? I don't think that it should be here

MrRowey:
> What's up with the VS code file? I don't think that it should be here

Fixed didn't realise FA repo doesn't have .vscode settings in a gitignore file

PR #5615, opened 2023-11-01
Show the Backpack on SCU BubbleShield

MrRowey:
For consistent look keeping the Backpack bone showing when the SACU has the bubble shield make sense and shouldn't be removed


Garanas:
Great find 👍

Issue #5602, opened 2023-10-29
Overcharge is enabled even without having any energy storages


Basilisk3:
Potentially game-breaking, when the player tries to overcharge anyway.

PR #5618, opened 2023-11-01
Fix the energy requirements of Overcharge

Garanas:
Closes https://github.com/FAForever/fa/issues/5602


lL1l1:
Seraphim SACU XSL0301 also needs an energy requirement.

Garanas:
I'll add that one in with another commit directly to `deploy/fafdevelop`

PR #5626, opened 2023-11-02
Fix a bug with the navigational mesh

Garanas:
Interestingly enough the following is not the same:

```lua
local test = { }
for k = 1, 10 do
    test[k] = 1
end

for k = 5, 10 do
    test[k] = nil
end

LOG(table.getn(test)) -- 5
```

```lua
local test = { }
for k = 1, 10 do
    table.insert(test, 1)
end

for k = 5, 10 do
    test[k] = nil
end

LOG(table.getn(test)) -- 10
```


Garanas:
With thanks to @gordenwunderlich : https://www.lua.org/pil/19.1.html

Issue #5594, opened 2023-10-27
Balance idea: Adjust BT of factories


Tagada14:
## Intended changes

![image](https://github.com/FAForever/fa/assets/44676678/07a2ca89-b8d2-4ed0-93e4-77fb4b0e1f42)

## Thoughts on the changes

Standardizes build time across all factories to make more sense and be more balanced. No longer will the T2 -> T3 Support factory upgrade be ~4 times faster than the T1 -> T2 upgrade.

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

PR #5630, opened 2023-11-02
Adjust the build times of air factories

MrRowey:
Changes for #5594


Garanas (changes requested):
Some of the build time changes are not quite as intended; see also:

image

This is too fast. It should be 80 seconds

Garanas (approved):

PR #5631, opened 2023-11-02
Naval BT Changes

MrRowey:
Completes #5594


Garanas (changes requested):
I think the build time changes are based on the wrong build rate. As an example: 

image

This is too fast (should be 120 seconds)

MrRowey:
Ooo good catch i didnt take into consideration of the BP increases from the upgrade that T2 gives

Garanas (approved):

Issue #5625, opened 2023-11-02
3772: Engineers built from Cybran and UEF air factories do not automatically rotate towards the optimal rollout point


Basilisk3:
Instead, the engineers turn inside the factory and then move out themselves. This is significantly slower than them being turned instantly, like it is the case with Aeon engineers from Aeon air factories. This results in Aeon having better air factories than the other factions.

Garanas:
This has always been the case 😃

Basilisk3:
That is not ideal then, because like this some air factories are better than others for the basic task that is building engineers. I don't really see a reason why some factories shouldn't have this feature, as it gives them a disadvantage. This difference will add up over the course of the game, especially when considering that these are some of the earliest units being built and sometimes left on engineer production for 10+ minutes. This also results in some factories having fewer pathfinding problems than others, which is not ideal either. 

I think the automatic turning was a great feature added in FAF, and that all factories should have it enabled.

Garanas:
I agree with you; just pointing out that the behavior has always been like this

Issue #5627, opened 2023-11-02
Cybran, Seraphim and UEF Air Factories have a significantly slower roll off time than Land Factories, or the Aeon Air Factory


Basilisk3:
The Aeon Air Factory has exactly the same roll off time as the Land Factories, but the other Air Factories are significantly slower than either the Aeon Air Factory, or the Land Factories.

I noticed this when testing #5625.

Garanas:
You're right, but to fix this means the animation speed needs to be extremely fast. And that looks off 😃

Ejsstiil:
Isn't this kind-of intentional by design? I mean that every race indeed has a differences and it should - look for example the racial/national differences at the game Age of Empires for example. Or any other game, where one race has advantage in building, other has it cheaper, you got the idea.

If we slowly unify all races then all it remains is just a different graphics.

PR #5640, opened 2023-11-05
Normalize air factory roll off times

Basilisk3:
In this PR I try to normalize the roll off behaviour for all Air Factories. My aims are:

- All Air Factories must build engineers at the same rate. This is not the case right now, as described in #5627. In other words: Right now, some factories can build sooner after a prior engineer has been finished.
- No factory must have a significant advantage over the other. For example, one factory should not have a significantly more efficient roll off animation.

**Tasks:**
- [x] Evaluate the efficiency of the different roll off animations
- [x] Make the Seraphim Air Factory build engineers at the same rate as Land Factories and the Aeon Air factory
- [x] Add new roll off points where it makes sense, and improves pathfinding
- [x] Add Garanas' automatic rotation towards the optimal rollout point for the Seraphim Air Factories
- [ ] Add Garanas' automatic rotation towards the optimal rollout point for the UEF Air Factories
- [ ] Add Garanas' automatic rotation towards the optimal rollout point for the Cybran Air Factories
- [x] Unify the formatting for RollOffPoints in the .bp

Not possible without changing the animation:
- [ ] Make the UEF Air Factory build at the same rate as Land Factories and the Aeon Air factory
- [ ] Make the Cybran Air Factory build at the same rate as Land Factories and the Aeon Air factory

**Explanation of the changes:**
- Roll out points: The Aeon and Seraphim get an additional roll out point. They used to only be able to release engineers out of their sides, which, looking at their models, did not make much sense. Now engineers can move straight out of both factories, but can still choose to take one of the sides, if it makes sense for them. This makes pathfinding less complicated.
- Automatic rotation of engineers towards the optimal rollout point: I introduced this for the Seraphim Air Factory for now. 
- Build-Cooldowns: To bring the Seraphim Air Factory in line with the Aeon Air Factory, I decreased its cooldown by 4 ticks. From my testing, both factories behave exactly the same now. To bring the Cybran and UEF Air Factories in line with these two, the animations of both will have to be sped up, which is controversial and still up to debate.


Garanas:
This is looking good 😃 !

Basilisk3:
> This is looking good 😃 !

Thanks, I will try to implement your engineer-roation feature to the factories that don't have it yet, to close #5625. I hope that is okay.

Garanas:
Of course it is - looking forward to the end result!

Basilisk3:
Concerning the automatic engineer turning: I have managed to introduce it for the Seraphim Air Factory, but I am having problems with the UEF and Cybran Air Factories. Both will lower their platforms, then turn the engineer and let it move out correctly. But afterwards both factories won't raise their platforms anymore. 

If fixing this is more complicated, it would probably be better if this is done in another PR.

Garanas:
I'll see what I can find for you

Basilisk3:
I did some maintenance.

Garanas:
@Basilisk3 The current changes are a great start, great work 👍

PR #5642, opened 2023-11-05
Projectiles Annotation Work Part 3

MrRowey:
- Some Files have a everything deleted and are left with a blank files as they are unused but kept for shadow file for steam


Garanas:
Great work here, @MrRowey  👍

PR #5645, opened 2023-11-05
AI/Engineer State Machine

relent0r:
Description of changes

This draft PR is for implementing an engineer based state machine to pick up jobs from the engineer manager.
It is a wip with the following components.

Initialization.
Sets the Location type and engineer manager instance.

DecideWhatToDo
Legacy engineer manager task.
Takes a preallocated task or queries the engineer manager GetHighestBuilder
Sets platoon data and stores the engineer handle.

Note. Engineer manager is static. The defaulted taskfinished function would relocate the engineer to a different manager if it was closer to a new base. Happens on disband.

SetBuilderData
Sets data based on builder, requires base template  and construction data.

NavigateToTaskLocation
Handles navigation.

CheckForOtherTask
Handles checking for other task before going back into the normal engineer task.

Current issues.
Uncertainty on how to handle the queued construction. The default AI runs a buildstructure function on the build location in order to 'reserve' the build position.
It also does this as part of the navigation queue.
This means that pathing that is required must run through and then run a buildstructure command at the end before the engineer starts moving.
Ideal solution is to have the engineer manager understand reservation for build locations so that there is no requirement for this. This may be possible with Jips chunk based templates.


relent0r:
> image
> 
> There's something going wrong here 😃

Yer for some reason its picked up every sync my fork has done with the main repo over the last few months. Never used to do that.

Garanas:
@relent0r the drawing of #5647 is available in this branch now too

MrRowey:
@relent0r whats the update on this PR? Could the changes be moved to a clean pr without all the trailing commits?

PR #5653, opened 2023-11-08
Transport Quick Drop

clyfordv:
Adds a HorzMotionEvent check for air transports that, when they're coming in to drop off units, allows a little horizontal/vertical fudge factor for drop off position. This results in a crisp descent + immediate cargo release instead of the little drunken two-step they do currently.

Along with increasing the speed at which units are dropped off, it also effectively decouples the air controller parameters (KLift, KLiftDamping, LiftFactor, etc.) from drop time, allowing fast drops and a higher performance flight profile.

Base modification applied to all transports via the AirTransport class. Includes a demo on uea0107 (Courier) that shows a dramatic increase in the ability to fly over mountains.


Garanas:
This looks good; great find on that interaction with aborting the path finding! Perhaps we can also use this (in combination with a sim callback) to have units start assisting immediately once in range 🤔 ?

Can you process the other transports? Then we can merge all of that in and include it with the next and hopefully last hotfix 😃

Garanas:
Great work @clyfordv !

PR #5656, opened 2023-11-10
Allow Hotbuild to redirect orders to exfac units like the buttons do

gordenwunderlich:
https://github.com/FAForever/fa/assets/7266458/38e0c0c2-219e-415b-ba32-ba24ee881ef9

should work for both normal and alternate hotbuild (templates not yet)

Needs more testing since I just quickly copied the code from @clyfordv but I wanted to at least put this out there.


Garanas:
@4z0t can you take the time to review this? I generally do not use hotkeys

lua/keymap/hotbuild.lua
```diff
@@ -279,14 +279,22 @@ function factoryHotkey(units, count)
                     if IsKeyDown("Shift") then
                         count = 5
                     end
-                    IssueBlueprintCommand("UNITCOMMAND_BuildFactory", units, count)
+                    if exFacUnits then
```
4z0t, line 282: Copy pasted fragments of the code

lua/keymap/hotbuild.lua
```diff
@@ -416,15 +424,28 @@ function buildActionFactoryTemplate(modifier)
 
     local template = effectiveTemplates[cyclePos]
     local selectedTemplate = template.templateData
-
+    local exFacs = EntityCategoryFilterDown(categories.EXTERNALFACTORY, selection)
```
4z0t, line 427: move to separate function

4z0t (changes requested):
Follow [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

lua/keymap/hotbuild.lua
```diff
@@ -416,15 +424,28 @@ function buildActionFactoryTemplate(modifier)
...
+    local exFacUnits = nil
+    if not table.empty(exFacs) then
+        local exFacUnits = EntityCategoryFilterOut(categories.EXTERNALFACTORY, selection)
```
4z0t, line 430: redefinition of the local variable

lua/keymap/hotbuild.lua
```diff
@@ -416,15 +424,28 @@ function buildActionFactoryTemplate(modifier)
 
     local template = effectiveTemplates[cyclePos]
     local selectedTemplate = template.templateData
-
+    local exFacs = EntityCategoryFilterDown(categories.EXTERNALFACTORY, selection)
+    local exFacUnits = nil
+    if not table.empty(exFacs) then
```
4z0t, line 429: EntityCategoryFilterDown returns nil if no units were found, excess check here

lua/keymap/hotbuild.lua
4z0t, line 430: redefined local variable `exFacUnits`

gordenwunderlich:
@4z0t
I moved the filter code to a different function, that also fixes the issue of declaring the local twice.
Should I put the part with `if exFacUnits then` into a separate function as well? It feels kinda weird to hide the simple if behind another function call.

lua/keymap/hotbuild.lua
```diff
@@ -32,6 +32,21 @@ local worldview = import("/lua/ui/game/worldview.lua").viewLeft
 local oldHandleEvent = worldview.HandleEvent
 
 
+function translateExFacUnits(selection)
```
4z0t, line 35: Pascal case function name.

lua/keymap/hotbuild.lua
```diff
@@ -279,14 +294,22 @@ function factoryHotkey(units, count)
                     if IsKeyDown("Shift") then
                         count = 5
                     end
-                    IssueBlueprintCommand("UNITCOMMAND_BuildFactory", units, count)
+                    if exFacUnits then
+                        IssueBlueprintCommandToUnits(exFacUnits, "UNITCOMMAND_BuildFactory", units, count)
```
4z0t, line 298: Still the same code 4 times pasted.

4z0t (changes requested):

lua/keymap/hotbuild.lua
gordenwunderlich: Oh, I just copied the style of the other function names in that file.
should I update them all for correct casing?
Or should we keep them as is in case someone referenced them in a mod?

Garanas:
There's no need to write a separate function for that if statement. If you can apply the Pascal convention to the name of the function today then we can include it with the hotfix

lua/keymap/hotbuild.lua
gordenwunderlich: That's what I meant.
To me it feels more weird to have those effectively 2 lines of code (that also just call another function) behind another function call.

gordenwunderlich:
I changed the name of the new function, though all other functions in that file are still lowerCamelCase

Basilisk3:
Tested and works. The only thing that does appear to be missing, is that the automatic enabling of repeat build does not work. 

To remedy this, you can add `categories.EXTERNALFACTORY` and/or `categories.EXTERNALFACTORYUNIT` to line 684 in gamemain.lua. Though this will not solve this problem either, because the code requires you to _select_ the factory for it to change it to repeat build. So if you never select the external factory, it will not have repeat build enabled automatically.

MrRowey:
@gordenwunderlich can you look at fixing what Basalisk mentions?

Garanas (approved):

Garanas:
@MrRowey that can be fixed separately from this pull request.

PR #5666, opened 2023-11-12
Move classes of Cybran weapons to separate files

MrRowey:


Garanas:
Great work @MrRowey 👍

PR #5667, opened 2023-11-12
Automate Sera sniper mode micro

lL1l1:
This change makes the sniper mode slowdown only trigger during the reload, which is already possible manually by changing mode after the sniper shot.
#### Pros: 

- It's in theme with Supcom's design to remove tedious micro: clicking the sniper button back and forth when in range and after shooting isn't fun or in-depth like other micro tricks kept in the game.
- It would be much easier for all players to get full capabilities out of the unit.

#### Cons: 
- This allows bands of snipers roaming at max speed, but then shooting a very high damage volley whenever they spot anything, entirely without player attention. But that inattention can still be punished by abusing the overkill the snipers will inflict on fodder like land scouts or single targets.

### DIFF
--- a/units/XSL0305/XSL0305_script.lua
+++ b/units/XSL0305/XSL0305_script.lua
@@ -15,7 +15,6 @@ local SDFSihEnergyRifleSniperMode = SeraphimWeapons.SDFSniperShotSniperMode
 ---@class XSL0305 : SLandUnit
 ---@field TrashSniperFx TrashBag
 ---@field isSniperFiringMode boolean -- Whether the sniper mode is active for next shot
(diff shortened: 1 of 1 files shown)


Garanas:
@lL1l1 Do you have time to review the changes I made? Then we can merge this in

lL1l1 (commented):
Looks good. I'll avoid making new closures through local/anonymous functions in the future.

Side note: how do you want comments in the code annotations (`@field`, `@param`)? I noticed there are different styles with `# ` before the comment, or `-- `, or nothing. Sometimes tabs are used and sometimes they aren't.

Garanas:
I think the extension doesn't even require a comment symbol. Sometimes it glitches out, not understanding that the comment started. I personally use `#`, but there's no strict rules on that.

PR #5674, opened 2023-11-14
Introduce separate files for Seraphim unit classes

MrRowey:
Continuation of https://github.com/FAForever/fa/pull/5652


Garanas:
A lot of goodies in this one!

In particular, every time a factory build start building something it would essentially run two build sequences at once. Where the first is the default build sequence, used by all other factions, and the second a separate sequence only used by Seraphim factories. The default build sequence would run, but end up forking an empty function. That's a bit of a waste. Seraphim factories now use the same default build sequence, with their own effects embedded into it of course!

PR #5682, opened 2023-11-17
Apply primary damage of the Ahwassa after 2.5 seconds

Garanas:
https://github.com/FAForever/fa/assets/15778155/950aa288-9463-4f7d-b848-583f7fe9d1fb


ComradeStryker:
This is really cool, I love the new effects!
It's like a mini-Yolo nuke!

However, a tiny bug I noticed when testing it, is that this new 'bomb' does not apply damage to allied units and structures.

Garanas:
Did the old Ahwassa bomb deal damage to allied units?

ComradeStryker:
I'm certain it did. It's an AoE effect after all.

PR #5690, opened 2023-11-19
Updated all files to use CamelCase for EffectTemplates.lua

MrRowey:


Askaholic:
FYI this is `PascalCase` not `camelCase`.

Garanas:
@MrRowey What is the idea behind these changes?

MrRowey:
> @MrRowey What is the idea behind these changes?

It just removed Flags from our intellisense as some files are all Lowercase and some are using PascalCase this unifies them all under PascalCase and keeps the intellisense happy reducing the amount of warnings from it

Garanas:
As discussed privately, it is better to have lowercase imports as done via https://github.com/FAForever/fa/issues/4325. That is because of how importing a module works - it will lower the entire string (effectively creating a new string) if it can't find the module the first time

Issue #4886, opened 2023-04-25
Janus may not speed up again when bombing sequence is interrupted


Garanas:
**Describe the bug**

The Janus slows down and then speeds up as it finds targets. This is likely done so that the bombs cluster up more.

**To Reproduce**

Ground fire with a Janus and then interrupt the firing sequence. This isn't trivial to do. Then compare the movement speed with a Janus that didn't fire at all.

**Expected behavior**

For the Janus to always speed up as expected.

MrRowey:
i will add this to my todo list seems easy enough to fix :)

Issue #2407, opened 2018-03-09
Janus bombs track to units and buildings instead of dropping in a line


JaggedAppliance:
So this bug will just not die, it's still around and is as bad as ever. It also seems to be present in the scorcher (UEF T1 bomber).

CookieNoob:
it's not tracking.... the bombs are just launched in the direction of the estimated target position.
try to play around with these values:
https://github.com/FAForever/fa/blob/22737d8080277b0bc7b3752bab538c7d81578b83/units/DEA0202/DEA0202_unit.bp#L507-L512

JaggedAppliance:
Ok, thanks for the tip cookie. I'm aware it's not technically tracking but that word describes the behaviour well imo.

CookieNoob:
the important thing is that there is tracking in the game. I.e. anti air missiles track their targets, in the nomads mod there are several missiles that do that too (or did until I changed it recently)... so it's a bit misleading when you speak about tracking

MrRowey:
@Garanas this can be closed as it is not a bug or maybe just get the balance team to see if this is something they want to change or not

Garanas:
This is a known issue and should be solved by https://github.com/FAForever/fa/pull/3509.

MrRowey:
@Garanas this one fixed ?

Garanas:
No, still very much alive.

Ejsstiil:
Bombs actually does some kind of tracking, see the video, when I try to teleport my commander who was locked-on by enemy DEA, to the different locations during DEA's bomb run:

following explains that every bomb is locked to the target's position before it is ejected/spawned, and the target vector is adjusted for every upcoming next bomb.

https://github.com/FAForever/fa/assets/36369441/abd23159-835a-4d04-981d-fe488a750ab6

Next video shows when I increase the number of bombs to drop and make the bones fire sequentially, or - in other words - make the releasing much slower, the late bombs can go even backwards as they sticks to the target location:

https://github.com/FAForever/fa/assets/36369441/57f721da-9b97-4575-9af2-17923a2bbd8f

PR #5695, opened 2023-11-22
Unit rework - Janus DEA0202

Ejsstiil:
This PR is aimed to improve few things: to fix the bombs pointed backwards, un-welding the faces in the model, add new bones and to generate a new normal map for the newly flattened faces. And also create some emission effect to the grill of the engine exhaust.

### Bombs & Bones:

That was created as a balance stuff in this PR https://github.com/FAForever/fa/pull/1466 [`6b98110`](https://github.com/FAForever/fa/commit/6b98110bb94b741030625f4c7efaf9d7a5dee657#diff-5281625d8517ed57fd9066727315e3e3819455472d7120735c9721d7358e2a0cL479-L483) when the third bomb was introduced, probably in order to widen and shorten the carpeted area. The two old bombs were reattached to the contrail bones, which are located at the tips of the wings - but which are pointing backwards. Similarly, the added bomb was attached to the Exhaust bone in the mid, but which is also pointing backwards.
Now it looks like this:

Result | Right | Left
:---: | :---: | :---:
 | bones | bones
 . | New bones are attached to contrail ones, so when wings are rotated in aerial engagement, these bones will follow | *How many players actually know that fact about rotating wings? Not counting us, we can see it in the code.*

This keeps the mentioned balance stuff untouched, see the final bombarded area:

https://github.com/FAForever/fa/assets/36369441/253fe4a7-f23b-4e21-8f4f-c5eb3109d351

### Faces
When faces are split (which is something we want to do, UEF machinery supposed to look sharp) and then exported back to game, the original normalmap gets off:
Just the model | Model plus the new normal map
:---: | :---:
This is happens when just model is exported. Note that flat faces seems to be pushed inwards. *The selected units are the exported ones* | And this is the same model, but with new normal map.
flat-faces-original-n | flat-faces-new-normal

Thus the new normal map needs to be recreated:

### Normal map
I've spend a lot of time in order to generate best looking normal map as possible from the unit's assets available, like Albebo and SpecTeam, but without the need to repaint it by hand and usable for the all other units not just for Janus, so with some! processing I came with this:
Original | New | Channels 
:---: | :---: | :---:
normal-o | normal | channels
This is the original map that we know, but with inwards looking faces, when used on newly exported model | This is the new map.<br/>*Since shader only uses baa channels for compute normals, there is no point to duplicate g channel into r & b channel, beside to get back greyish look. Moreover it gives opportunity to utilize these channels to something meaningful.* | This is how the channels looks like, without r & b channel filled.

### Emission
Just for fun and for good-looking effect, even though hardly anyone will zoom to any unit during tough battles, and especially to the aerial units, which are hard to catch with the game camera in the mid-air, I added this emission which is just cool, right?
engine

And at last the whitespace reformat of the DEA0202_unit.bp, no functional changes made.

Ejs

Edit: typo and added link


MrRowey:
Nice work. if possible you could look at the 2 issues regarding the Jause in #2407 & #4886 one is about the bomb projectiles which may be fixed with this pr and one is after the bombs have been deployed.

Ejsstiil:
Thanks, yea I did noticed the bomb "tracking" of some kind, when I was shooting videos for this PR. But seemed acceptable to me as I was able to replicate it only under rare conditions - slowed game, teleporting bomber etc.
https://github.com/FAForever/fa/issues/2407#issuecomment-1824852821
Yea can take a look.

Ejsstiil:
> Nice work. if possible you could look at the 2 issues regarding the Jause in #2407 & #4886 one is about the bomb projectiles which may be fixed with this pr and one is after the bombs have been deployed.

This goes deeper than it seemed. For now, I won't mix them into this pull request, and I'll try to address the issue within another pull request.

Garanas:
This is great work @Ejsstiil , I'll even go as far as to use it as an example in our 'looking for volunteer' posts on Discord that we'll soon release 👍

PR #5698, opened 2023-11-25
Score graph start end fix

Pokute:
The end-of-game score graphs had buggy starts and ends. The start was actually the first saved datapoint at 00:30 instead of 00:00 and thus the scores never started from zero. The final graph segment between the last save datapoint and the end-of-game datapoint, which usually had a duration less than the default 30 seconds, had incorrect rendering.

Additionally the end-of-game datapoint was not taken into account when calculating the maximum values for graphs. Thus the graph could render outside the background.

Example of old graph:
image

Example of fixed graph:
image

This PR has a possible breaking change: Score history data now has an additional data point at 00:00, instead of the first one being at 00:30. If some mod uses this data, it might get confused.

### DIFF
--- a/lua/sim/score.lua
+++ b/lua/sim/score.lua
@@ -84,7 +84,6 @@ end
 
 local function ScoreHistoryThread()
     while not GameIsOver do 
-        WaitSeconds(scoreData.interval)
         local data = {}
         for index, brain in ArmyBrains do
             local Score = scoreData.current[index]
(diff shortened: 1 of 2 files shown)


Ejsstiil:
Nice catch, that looks interesting! Can you please do more tests? Ideally over a replay to see the same numbers on graphs? Not sure though if a replay shows the graphs at the end...

Pokute:
Game without patch:
image

Replay of that same game with patch:
image

Unfortunately the game length was 12:31, so the effect at right side is minimal.

Ejsstiil (approved):

PR #5700, opened 2023-11-26
UI Pulse rework - make it resilient to low framerate.

Pokute:
The old way could result in big and inconsistent drifts in alpha on framerate drops. The new code doesn't rely on what the alpha was during the previous frame.


Pokute:
> Could you hint me to a place where the pulse behavior is used?

Not many places. "IN TRANSIT" text during loading the map. I think the bottom-left buttons during in-game (like overcharge, launch and build tactical missiles). Also my incoming graph patch will use it - while the code has been there for a very long time, it has never really worked due to different issues.

PR #5702, opened 2023-11-28
Increase the Chariot's hitbox to fix interceptors missing it

Basilisk3:
**Changes**
SizeX: 1.8 --> 2.2
SizeY: 0.9 --> 1.3
SizeZ: 1.8 --> 2.2

This attempts to fix the bug where interceptors shoot in front of the Chariot and miss it. In my tests, these changes fixed this in most cases. They can still miss it when it starts and lands, even with the increased Y-size, although this is reduced with these values. Let me know if the increase in Y-size is too much in your opinion.

Pointed at fafbeta, because technically this is a balance change.


Garanas:
We're dropping `deploy/fafbeta` all together when the new year starts. Can you point it to `deploy/fafdevelop`? Then I can include this in a hotfix

PR #5703, opened 2023-11-29
UI - score graph player highlight, try #2

Pokute:
Created the previous PR too soon, closed it and couldn't seem to update it after forcing pushing, so this in PR attempt #2.

Highlights the active player's graph in the score screen. This one has Bessima, the green player highlighted:
image

Also has code cleanups.


Ejsstiil:
Does this highlighting change the original color of the player?

Pokute:
> Does this highlighting change the original color of the player?

Not really the color, but it applies the pulse effect to the opacity/alpha of the player's graph.

PR #5705, opened 2023-11-29
Fix bugs and improve effects of Aeon TMDs

Garanas:
Fix for the debug described on the [forums](https://forum.faforever.com/topic/6852/aeon-tmd-are-broken/3). Improves the effects and simultaneously reduces and improves the effectiveness of the defenses. They're improved in the sense that they can now more reliably defend against Seraphim tactical missiles (that are on average faster). They are reduced because a single flare can now only stop three missiles (instead of an infinite amount). 

https://github.com/FAForever/fa/assets/15778155/ac54f200-25c2-4ea8-9efe-3784906c216d

https://github.com/FAForever/fa/assets/15778155/bebd262c-f0cc-4e70-9ba1-e8e1615ad92b

The effects are introduced to make it easier to understand what is happening for the player. The beams are similar to the build beams of engineers. The beams are smaller and the direction of the matter is inversed (moving towards the flare, instead of towards the target).


Garanas:
https://github.com/FAForever/fa/assets/15778155/6fc28d03-c057-448e-85d9-6762021502cb

PR #5712, opened 2023-12-01
T2 Bomber speed nerf

Basilisk3:
MaxAirspeed: 15 --> 14.25

T2 Bombers were as fast as their counter (interceptors) which meant that, in some scenarios, interceptors would never be able to catch them. This nerf increases the skill-cap of using T2 Bombers and enables better counter play for the defending player.

Pointed at fafdevelop.

### DIFF
--- a/units/DEA0202/DEA0202_unit.bp
+++ b/units/DEA0202/DEA0202_unit.bp
@@ -20,7 +20,7 @@ UnitBlueprint{
         KTurn = 1,
         KTurnDamping = 1.5,
         LiftFactor = 10,
-        MaxAirspeed = 15,
+        MaxAirspeed = 14.25,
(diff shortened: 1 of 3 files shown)


MrRowey:
The Balance Team is currently discussing this and currently doesn't see a need for the changes, especially as the Janus just got a reworked.

MrRowey:
T2 fighter bombers are fins as they are currently especially and the Janus just recently got changes so we need to see how they impact first

Basilisk3:
I disagree. T2 Bombers being reasonably catchable by interceptors is a prerequisite of a well-balanced T2 air stage. If T2 Bombers are found to be too weak due to this change, they can be readjusted in another way.

PR #5713, opened 2023-12-01
Marr (Aeon stationary flak) buff

Basilisk3:
MuzzleVelocity: 35 --> 40

This unit has always been much worse than the other stationary flaks. Essentially it is a worse version of the Air Cleaner, as it has a much worse firing cycle, lower AOE, lower DPS and lower HP. In exchange for these disadvantages, it only gets 0.5 less FiringRandomnes, which does not help the unit much.

This PR aims to buff the unit, while also making it more unique, in line with the theme of its faction. With this change it still has the disadvantage of having low damage, but is now able to hit fast moving targets more reliably and at a greater range, than other flaks.

Pointed at fafdevelop.

### DIFF
--- a/units/UAB2204/UAB2204_unit.bp
+++ b/units/UAB2204/UAB2204_unit.bp
@@ -178,7 +178,7 @@ UnitBlueprint{
             MaxRadius = 50,
             MuzzleSalvoDelay = 0,
             MuzzleSalvoSize = 1,
-            MuzzleVelocity = 35,
+            MuzzleVelocity = 40,
             ProjectileId = "/projectiles/AAAFizz01/AAAFizz01_proj.bp",
             ProjectileLifetimeUsesMultiplier = 1.25,
             ProjectilesPerOnFire = 1,


MrRowey:
The Balance Team is currently discussing this and currently doesn't see a need for the changes.

MrRowey:
Unless you can provide some dat on how this is beneficial the flack is fin as it is

Basilisk3:
Stat wise, the Marr is significantly worse than all other flaks and because of this, it underperforms in-game. I am not sure why it makes sense for Aeon to just have a bad T2 flak, when every single other static AA or PD is good and has something unique going for it. As outlined in my OP, the Marr is just a bad version of the UEF flak.

MrRowey:
> Stat wise, the Marr is significantly worse than all other flaks and because of this, it underperforms in-game. I am not sure why it makes sense for Aeon to just have a bad T2 flak, when every single other static AA or PD is good and has something unique going for it. As outlined in my OP, the Marr is just a bad version of the UEF flak.

Can you provide side by side comparison then to back up the fact that it's worse ?

Basilisk3:
> Can you provide side by side comparison then to back up the fact that it's worse ?

Yes. 
https://faforever.github.io/spooky-db/#/UEB2204,UAB2204
https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UEB2204/UEB2204_unit.bp
https://github.com/FAForever/fa/blob/deploy/fafdevelop/units/UAB2204/UAB2204_unit.bp


|   | Air Cleaner | Marr |
| ------------- | ------------- | ------------- |
| HP  | 2590  | 2450  |
| DPS (in-game) | 156  | 136  |
| AoE  | 3.5  | 3.0  |
| Firing Randomness  | 2  | 1.5  |

PR #5715, opened 2023-12-01
Fix Deceiver Range Buff

lL1l1:
#5142 has no effect because the radar intel grid has a resolution of 4x4. This increases the range to 24, which is divisible by 4 so it has an effect.


RabidPope:
May also want to reduce visual indicator for range if that is possible somehow.

lL1l1:
Misclick

Garanas:
This is a great find; well done 👍

PR #5725, opened 2023-12-05
Perfect the reload time calculations

lL1l1:
The reload time now accounts for all engine quirks like rounding and minimum wait times by basically emulating the firing cycle, so the DPS display is perfectly accurate for all types of weapons. A few examples:
- The Pillar's 0.8 fire rate is correctly turned into a 1.2 second reload, and the DPS is updated accordingly.
- Tech 1 PD's DPS is displayed as 167 instead of 150 because their 3 fire rate turns into a 0.3 fire cooldown, not the 0.333... mathematical result.
- Apparently Mantis have 27 DPS (26.67 un-rounded), Thaams 25 (24.6), Strikers 24 (24), and aurora 24 DPS (23.53). Previously they were all shown as the same 24 dps. A +11% DPS difference!
- Spearhead's reload time is correctly displayed as 13.1 seconds instead of 10.0.
- Cybran T3 MAA's reload time is correctly calculated; It has a very complex firing cycle.
- Cougar and Lightning Tank reloads are corrected (+0.1 seconds because they use rack reload for their cycle instead of fire rate).

Small issue:
UEF and Cybran nuke subs display their TMLs' stats as a volley of 6 and 3 missiles every 60 and 30 seconds respectively because they have multiple racks to represent the individual tactical silos on the model.


lL1l1:
This can also be adopted in the two unit dbs and the lobby unit manager.

MrRowey:
Nice Work !

Garanas:
@The-Balthazar could I encourage you to review this? Sprouto of the LOUD community may also be interested in this, but I can't ping him 😞

The-Balthazar:
I'd already accounted for tick rounding in all my weapon calculations.

The-Balthazar:
One thing I do note, which is hard to actually take into account is this doesn't take into account projectiles that split and pass the whole amount of damage to all child projectiles. I have a list of all the vanilla projectiles that do this here: https://github.com/The-Balthazar/BrewWikiGen/blob/master/Components/Weapon.lua#L344

lL1l1:
> One thing I do note, which is hard to actually take into account is this doesn't take into account projectiles that split and pass the whole amount of damage to all child projectiles.

It does, see lines [571-575](https://github.com/FAForever/fa/pull/5725/files#diff-696cb30a1348da6464070aaa330a513ed8defd2cee61da5ae075d88450192d3cR571-R575):
```lua
local ProjectilePhysics = __blueprints[info.ProjectileId].Physics
while ProjectilePhysics do
    Damage = Damage * (ProjectilePhysics.Fragments or 1)
    ProjectilePhysics = __blueprints[string.lower(ProjectilePhysics.FragmentId or '')].Physics
end
```
What it doesn't account for is the SplitDamage table, which changes the damage and radius of cybran split missiles.

The-Balthazar:
I missed that. Although upon checking that relies on assuming that projectiles have been updated to define that in the blueprint. This notably excludes:
* [AANTorpedoCluster01](https://github.com/FAForever/fa/blob/deploy/fafdevelop/projectiles/AANTorpedoCluster01)
* [AIFGuidedMissile01](https://github.com/FAForever/fa/blob/deploy/fafdevelop/projectiles/AIFGuidedMissile01)
* [AIFQuanticCluster01](https://github.com/FAForever/fa/blob/deploy/fafdevelop/projectiles/AIFQuanticCluster01)
* [AIFQuanticCluster02](https://github.com/FAForever/fa/blob/deploy/fafdevelop/projectiles/AIFQuanticCluster02)

Although that's somewhat a separate issue, and also the last two of those are unused.

The-Balthazar:
I must've checked the wrong version of the second.
It's still worth updating the 3rd and 4th for the sake of any mods that use or reference it.

PR #5728, opened 2023-12-06
Link enhancement descriptions to bp stats

lL1l1:
This PR aims to create a VAR function to add bp stats into descriptions similar to the LOC format. Currently it uses `<VAR EnhancementStatName>`
I'm creating this as a draft to get opinions on what syntax to use for formatting the variable itself, because there are some bp stats that need some math to be done before being written down such as 
- `NewRateOfFire` (2 and 1.82), `RegenPerSecond` (0.035; used by regen aura) - needs to be formatted to percents.
- `NewMaxRadius` (30, 35, 40) - choose between hard coding subtraction of the specific unit's gun range or creating a way to reference the unit bp from the VAR function.
- `ShieldRegenRate`, `ShieldRechargeTime` - add a syntax for choosing to not use the prerequisite's stats, or else the aeon heavy shield will say +2 Regen +85 Recharge Time.

FYI LOC already uses the curly brackets `{}` for op codes so those aren't available.


lL1l1:
Functionality should be in a separate file so that both unit manager and unitviewdetail can use it.

MrRowey:
@lL1l1 whats the status on this ?

lL1l1:
I still want to do this, it's just not high priority.

BlackYps:

PR #5731, opened 2023-12-07
actually fix double veterancy and make wrecks copy enhancement costs

lL1l1:
Last fix didn't work because the `continue` was moved inside a new loop. The original method also doesn't work when enhancements are removed from a preset.
Now there shouldn't be any difference in wrecks or vet amounts for manually enhanced SACUs or preset SACUs or a mix of the two.


lL1l1:
I didn't intend for teleport to give wrecks mass, it would make teleport useless.

PR #5735, opened 2023-12-10
Small UI Fixes for Cybran Stealth

lL1l1:
- Fixes the toggle stealth button disappearing when you select a deceiver and a different stealth unit. Same issue occurs when selecting a personal shield unit with a shield structure.
- Stealth and cloak upgrades for ACU/SACU use separate togglecaps so they can be separately toggled.
- Stealth and cloak upgrades for ACU/SACU now use appropriate icons and help texts instead of only cloak:
Cloaking icon:
image
Personal stealth icon:
image
- Changes frequent warn to spew.


Garanas:
Didn't find the time to review this until now, great work!

PR #5738, opened 2023-12-10
Double the exponential absorption rate of water

BlackYps:
I noticed a problem with our new fancy water: the water depth for the terrain shader is limited to 1 maximum as we read that from a texture and the texture can't provide values greater than 1. With the exponential absorption that means we are stuck with ~90% absorption, which is a problem if someone wants to create really murky water. What I did is that we simply multiply the scale by two, then we can reach ~98% absorption, which should be enough.
If a map is supposed to have pretty clear water, then the abyss depth can be set to be significantly lower than the lowest terrain.

Maps using the linear water absorption are unaffected

### DIFF
--- a/effects/mesh.fx
+++ b/effects/mesh.fx
@@ -603,7 +603,8 @@ float3 ApplyWaterColor(float depth, float3 viewDirection, float3 color, float3 e
     if (surfaceElevation > 0) {
         // we need this switch to make it consistent with the terrain shader coloration
         if (IsExperimentalShader()) {
-            float scaledDepth = (-depth / (surfaceElevation - abyssElevation));
+            // We need to multiply by 2 to match the terrain shader.
+            float scaledDepth = (-depth / (surfaceElevation - abyssElevation)) * 2;
             float3 up = float3(0,1,0);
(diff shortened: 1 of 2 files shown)


Garanas:
From a gameplay perspective it is okay if units are capped at being 90% visible. That way there's always _something_ to see, if I'm not mistaken?

BlackYps:
Actually, the terrain is capped at 90% at the moment. The mesh shader is uncapped right now, as we don't use a texture to access the depth, but we calculate it from the world position. We definitely need to align both. The alternative would be to cap the mesh shader calculations as well.

I agree that most maps should be configured in a way that you can somewhat see units even in the deepest water. But this can be ensured by setting the abyss height appropriately. In most cases this would mean well below the deepest terrain elevation, but I see no problem with that. And then there are always the strategic icons as well.
Yes, most maps won't utilize this option to go to basically full absorption, but it would enable effects like shallow but murky water, so units stick out of the puddle, but their legs get swallowed. You have to keep in mind the scale of the units, so this scenario is not that unrealistic (speaking from a physical POV, not gameplay).

PR #5741, opened 2023-12-11
Fix Janus and Scorcher tracking bombs

### LINKED ISSUE #2407: Janus bombs track to units and buildings instead of dropping in a line

lL1l1:
Resolves #2407.
Greatly improves the bomb dropping pattern of the Janus and Scorcher by making them drop their entire salvo onto one location instead of constantly tracking their target and dropping each individual bomb onto the current predicted target location.
It does change balance for both bombers.
To test, turn on NoDamage in the console, spawn 10 bombers, and spawn a Harbinger as a target for you to micro. The Harbinger turns quickly so it can show off bomb dodging but isn't so fast that the bombing attempt fails like the Spirit. Notice how the Janus doesn't leave giant fiery rings when bombing a turning target.

https://github.com/FAForever/fa/assets/82986251/e17f1c57-8a12-4f36-84a0-49dd0ab74fce

https://github.com/FAForever/fa/assets/82986251/0d2cdb82-fa63-4ba8-82ad-f0d9d0ff3a15

The old "fix" in the weapon script didn't work because `data.usestore` was never set to true. Regardless, it would've had an incorrect bomb trajectory because targetPos was not adjusted for unit speed, while the ballistic acceleration calculation was.

### DIFF
--- a/lua/sim/defaultweapons.lua
+++ b/lua/sim/defaultweapons.lua
@@ -736,6 +736,9 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
             if unit.Dead then return end
             unit:SetBusy(false)
 
+            -- at this point salvo is always done so reset the data in case firing was interrupted
+            self.CurrentSalvoData = nil
+
             self:WaitForAndDestroyManips()
             local bp = self.Blueprint
             for _, rack in bp.RackBones do
--- a/lua/sim/defaultweapons.lua
+++ b/lua/sim/defaultweapons.lua
@@ -22,10 +22,9 @@ local CollisionBeam = import("/lua/sim/collisionbeam.lua").CollisionBeam
 local MathClamp = math.clamp
 
 ---@class WeaponSalvoData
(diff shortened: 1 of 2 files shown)


lua/sim/defaultweapons.lua
```diff
@@ -1089,6 +1097,7 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
                 end
             end
 
+            self.CurrentSalvoData = nil
```
Garanas, line 1100: It is not guaranteed that the code reaches this point, as an example when the target is lost. That is why the salvo data was also reset in the idle state

lL1l1: That is true

Garanas:
This is quite an interesting change, great work 👍 . I'll try to review it this week

Garanas:
Great work @lL1l1

PR #5742, opened 2023-12-11
Fix the Sparky and Deceiver being able to evade beam weapons

Basilisk3:
**Deceiver:**
SizeX: 0.7 --> 0.8
SizeZ: 0.8 --> 1.0

**Sparky**
SizeZ: 0.7 --> 0.8


Garanas:
What was your test setup?

Basilisk3:
This is extremely odd. When you spawn the unit first and then then PDs around it, they miss. But it does not happen the other way round. I cannot explain this.

Basilisk3:
There is always something new to be learned about this game, I guess. I revered the unwarranted changes.

Garanas:
> Contributor

@Hdt80bro any idea as to why this matters?

PR #5743, opened 2023-12-11
Eye of R-word target location cost rework

The-Balthazar:
The remote viewing script as a whole is a bit of a mess, but this attempts to address the clunky UX of trying to activate the ability when you have less than 10k energy in storage by being a drain event and a wait instead of a check and a take event.

~~Current issue with it though, is that this breaks the UI for the maintenance drain. (At least on the Steam version; the drain still happens, but it doesn't show on the unit.)~~


The-Balthazar:
This also makes it so that the 10k move/place cost is correctly attributed to the structure, and makes it so that if you re-target somewhere else before it's finished draining for the move/place it wont double charge you, and instead will just re-target the current drain.

Garanas:
Closed in favor of https://github.com/FAForever/fa/pull/5806 due to my rebase mistake 😃

PR #5745, opened 2023-12-12
Improve the build animation of Cybran naval factories

Garanas:
_current/old build animation_

https://github.com/FAForever/fa/assets/15778155/0ce320be-d552-4153-95ce-29000fd1b065

_new build animation_

https://github.com/FAForever/fa/assets/15778155/e40c9f29-1c4a-430a-9718-ed382b8ef189

With thanks to @The-Balthazar for creating the animation files.

The new build animation takes into account the size of the unit. The build beams are no longer aiming at the same location and they remain within the vicinity of the build arm so that they're not going all over the place. The build arm themselves are now also animated, just like the build arms of the Cybran air and land factories


The-Balthazar:
👍

PR #5633, opened 2023-11-04
Initial setup for the new chunk template concept

Garanas:
The chunk templates will eventually replace the base templates. It is similar, if not equal, to the templates used by the LOUD community. 

We use the navigational mesh to divide a map into squares. These 'chunks' templates represent what an AI could populate (in terms of structures) in one of those squares.

image

We can then create templates that are similar to build templates that fit in exactly one of those squares. As an example:

image

Each of those squares represents a chunk. A chunk of an AI base. We can create many variations of each size of a chunk to make sure one base is unlike the other. 

The idea is that the AI will claim additional chunks as it tries to build structures. This way the AI can expand the base and gain more ground.

We need chunks to prevent the AI from exhaustively searching for a place to build. An example are the current [base templates](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/basetemplates.lua). These are gigantic files with gigantic tables that the AI can use to find a build location that is 'nearby' of the intended build location. On average there will be a lot of negative tests before a valid build location is found.

PR #5750, opened 2023-12-15
Re-implementation of the AI - Part 2

Garanas:
One part of a long series of work on re-creating the foundation of an AI. Other chunks of work:

- #5508
- #5009
- #4858

And of course #5633 . There are more pull requests but this gives you a rough idea 😃 

## Engineer tasks

In this pull request, we are introducing a new feature we coin as "engineer tasks." Unlike traditional builders, these tasks are exclusively designed for engineers, focusing solely on the decision-making process of what to construct.

Conceptually, engineer tasks function as a dynamic build queue that can be queried and modified. Several key aspects surround these tasks:

### Strategic intent

Engineer tasks are generated by the brain, mirroring a player's strategic decision-making. In urgent situations, a base can autonomously generate tasks to address immediate concerns. The brain can articulate a 'strategic intent,' much like a player planning their moves. For example, if the goal is to establish four additional land factories in a specific navigational label (connected area), the associated task list might include:

- 10 tech 1 power generators
- 4 additional mass extractors OR upgrade a mass extractor
- 4 additional land factories

The factory building tasks can be allocated to specific bases with the corresponding navigational label. Economy-related tasks can be assigned to any base, with a preference for those farther from the front lines. This approach allows for the creation of task lists that articulate strategic intents, providing insights into the required resources and progress.

Engineers at each base execute these tasks, leading us to the second crucial concept: the tasks are persistent until they are completed.

### Persistent tasks

Engineer tasks exhibit persistence and are processed based on priority. When an engineer accepts a task, the task doesn't vanish from the task list. Instead, another engineer can take on the same task, collaborating rather than duplicating efforts. Unlike builders, which lack persistence and awareness of other instances, engineers naturally support each other in completing a task.

This persistence feature allows for efficient task management and resource allocation, contributing to a more organized and collaborative construction process. 

_more to come_


relent0r:
Very cool concept here Jip.
Thinking about it. With the strategic intent, what does the interpretation of this look like in your mind? Something that might Chp was working on which seemed like a really fascinating approach(and quite possibly too far fetched but I'm throwing it out there as I was super interested in the concept) and I'm thinking that its the interpretation interface that would be responsible for this. Is that you would define build power requirements and income requirements that AI will make happen. So I believe it is something where as long as the flexibility is there. This requires full blueprint interpretation.
A small scenario.
I need x air build power over x time. Depending on the values and whats available the AI may decide to make more air factories. But if the time value is lower it and it has a certain number of engineers available it may decide to assist an existing factory. Since the build time required would be over the limit.

Or I need this income. The AI will know that it can build mass on various points. But it may also decide that to achieve the required number due to distance or risk a mass fab is more economical to achieve it at the desired time frame.

The interesting thing about this type of setup is how it works with AIX multipliers. Since 6 factories for a 1.5 build/cheat AIX is not the same as 10 factories for a non AIX. 

The other interesting thing about it is how the AI's behavior would change depending on game balance changes or mods.

Garanas:
> Thinking about it. With the strategic intent, what does the interpretation of this look like in your mind? Something that Chp was working on which seemed like a really fascinating approach(and quite possibly too far fetched but I'm throwing it out there as I was super interested in the concept) and I'm thinking that its the interpretation interface that would be responsible for this. Is that you would define build power requirements and income requirements that AI will make happen. So I believe it is something where as long as the flexibility is there. This requires full blueprint interpretation.

This is exactly the idea!

There will be 'strategic intents' that you can write out (for the brain, but also on a base level for local issues) that can describe exactly that. For example, say the brain picks the intention to create five additional air factories. The code with the intent can then evaluate the status quo of the game (how much income do I have, how much do I need) and then generate tasks to reach the required income. Combine that with the factory tasks and there you have it!

Another idea: say the intent is to increase the economy by upgrading an extractor. The logic (that you wrote to perform the task) can not just take into account the economy status quo, but using the reclaim grid it could also take into account the reclaim! And as a result, it could say: we can build this, if we get engineers in that wreckage field near the base. Then order the engineers to go reclaim there. This can of course fail, but that's okay - it can fail for players too.

The dangerous factor here is mass of course - as you can't build that everywhere. But it doesn't need to be perfect, just good enough 😃

relent0r:
> > Thinking about it. With the strategic intent, what does the interpretation of this look like in your mind? Something that Chp was working on which seemed like a really fascinating approach(and quite possibly too far fetched but I'm throwing it out there as I was super interested in the concept) and I'm thinking that its the interpretation interface that would be responsible for this. Is that you would define build power requirements and income requirements that AI will make happen. So I believe it is something where as long as the flexibility is there. This requires full blueprint interpretation.
> 
> This is exactly the idea!
> 
> There will be 'strategic intents' that you can write out (for the brain, but also on a base level for local issues) that can describe exactly that. For example, say the brain picks the intention to create five additional air factories. The code with the intent can then evaluate the status quo of the game (how much income do I have, how much do I need) and then generate tasks to reach the required income. Combine that with the factory tasks and there you have it!
> 
> Another idea: say the intent is to increase the economy by upgrading an extractor. The logic (that you wrote to perform the task) can not just take into account the economy status quo, but using the reclaim grid it could also take into account the reclaim! And as a result, it could say: we can build this, if we get engineers in that wreckage field near the base. Then order the engineers to go reclaim there. This can of course fail, but that's okay - it can fail for players too.
> 
> The dangerous factor here is mass of course - as you can't build that everywhere. But it doesn't need to be perfect, just good enough 😃

I had a read through and it looks like a nice approach. I like the concept of the DefaultDistance. Also having assistees and builders, hopefully that might provide functionality(I didnt see anything that enabled this yet but its super useful especially for assistees) to pull an engineer off an assist task when its desired to switch to something else (e.g I'm assisting an upgrade but now our power is about to stall so I'll disconnect and look for a new builder).

Preferred chunk will be cool to, I tried doing a front and rear bias for my own AI's builders. It kind of worked but never quite. But I really like how this can potentially integrate the managers with modular strategy/tactical data sets.

Also really like the Deteriorating status for builders, since historically when the initiating engineer was killed you needed a special completion builder to get engineers to complete things but then your priorities and assistance counts were all off, where as now its effectively built in that engineers will complete things that are interrupted. Then having a task set to InProgress will stop the AI picking up multiple high value task just because the engineer hasn't arrived at the build location yet.

PR #5751, opened 2023-12-17
Do not include mexes under upgrade in "Select nearest onscreen lowest tech idle mass extractor"

4z0t:


Garanas:
I don't fully understand why this pull request is required. 

image

In this scenario the upgrading mass extractor is skipped just fine. In what scenario does it end up selecting the extractor that is upgrading?

PR #5752, opened 2023-12-17
Introduce a dithering effect as you zoom out

Garanas:
Prototype of a dithering-like effect for props.

https://github.com/FAForever/fa/assets/15778155/fb9bd357-2399-4a43-9783-cdf8527520aa


BlackYps:
I just tested it and the effect seems promising. Trees in a half-transparent state are prone to a lot of flickering, so we should try to keep the transition period as small as possible. But if it is too short then it will feel instant again if you zoom fast, so we need to do some experimenting.

Garanas:
Closing as I have no intentions to finalize this pull request.

PR #5756, opened 2023-12-17
Fix shields warping across the map

Garanas:
We can fix this by introducing a delay of one tick before we set the mesh of the shield. This is a visual change - functionally the shield is the same.

https://github.com/FAForever/fa/assets/15778155/aa49935a-d9fa-492d-80ab-07e285e042f3


Garanas:
Easier to review when you hide whitespace changes 😃 

image

PR #5758, opened 2023-12-18
Greatly improve Yolona Oss UX

lL1l1:
Greatly improves the controls and visual feedback for Yolona:
- Fixes being able to trigger the long 20 second cooldown by cancelling the launch order.
- Repurposes some firing emitters and the missile construction emitters to visualize the 20 second cooldown time:
  - The outer beams build up energy in the launcher.
  - The energy is shown in 3 effects, appearing from bottom to top as the charging progresses.
  - When the launcher is charged and the missile is built, there is a pulsing glow to indicate that it is ready to fire.
- Missile is raised according to build progress, not an arbitrary animation rate.

Bugfixes:
- Fixes the build effects appearing at the builder instead of the structure.
- Fixes the missile construction effects being offset from the center of the launcher.
- Removes unused T3 nuke animations, supposedly left behind from copying the T3 nuke to create the Yolona.


Garanas:
After a quick look - the changes make the unit and the effects feel more cohesive! I'll add a few videos later this week

Garanas:
https://github.com/FAForever/fa/assets/15778155/877a7507-3309-487b-86e8-cf30de325cdd

PR #3597, opened 2021-12-10
Show factory queue on rollover

4z0t:
Shows factory queue on top of unit info on rollover
fac queue4

Edit by Jip:
 - [x] Add game option to disable

PR #5761, opened 2023-12-21
Fix bug that can make build queue unresponsive

Garanas:
There are four cfunctions that allow you to interact with the build queue:

- (1) `ClearCurrentFactoryForQueueDisplay(): nil`: clear the internal build queue
- (2) `SetCurrentFactoryForQueueDisplay(userUnit) : UIBuildQueue`: define the internal build queue 
- (3) `DecreaseBuildCountInQueue(index, count):nil`: decrease count at index of internal build queue
- (4) `IncreaseBuildCountInQueue(index, count):nil`: increase count at index of internal build queue

The first problem is that they interact with some global object that is defined with (2). The functions (3) and (4) rely on that global object. With (1) the global object is reset. The jargon for this is 'global state' and it is considered a [bad practice](https://softwareengineering.stackexchange.com/questions/148108/why-is-global-state-so-evil).

The second problem is that (3) and (4) lack error handling. For example, the situation where you first call (1) and then (3) or (4) throws no errors. The jargon for this is a 'silent error' and it can be quite frustrating to work with.

The third problem is that you are unable to retrieve the current global state. There is no 'get' function.

The fourth problem is that you are unable to retrieve the build queue of a unit without also messing with the global state. There is no 'peek' function.

With the hooks as defined in `lua/userInit.lua` I try to help fix some of these problems. There is still a global object, but at least there are now safe to use functions that allow you to retrieve the global object and peek at the build queue of a given unit.

### DIFF
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -937,7 +937,7 @@ end
 
 ---
 ---@param unit UserUnit
----@return BuildQueue
+---@return UIBuildQueue
 function SetCurrentFactoryForQueueDisplay(unit)
 end
 
--- a/lua/ui/game/gamemain.lua
+++ b/lua/ui/game/gamemain.lua
@@ -711,7 +711,11 @@ function OnSelectionChanged(oldSelection, newSelection, added, removed)
     import("/lua/ui/game/unitview.lua").OnSelection(newSelection)
 end
 
+---@param newQueue UIBuildQueue
 function OnQueueChanged(newQueue)
+    -- update the Lua representation of the queue
+    UpdateCurrentFactoryForQueueDisplay(newQueue)
+
     if not gameUIHidden then
         import("/lua/ui/game/construction.lua").OnQueueChanged(newQueue)
     end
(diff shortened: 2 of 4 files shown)


Garanas:
Now, about the bug that we fixed with 13789a1. It was introduced by https://github.com/FAForever/fa/pull/3597 . That means the bug has been around for roughly two years. This particular bug only exists if the feature is enabled. It is by default enabled. It is called every frame.

image

The idea of #3597 is to show the build queue of a factory that you hover over. But - there is no 'peek' or 'get' function of any kind. The only way to retrieve the build queue was to set the build queue via (1). There was no way to properly reset the build queue back to the original build queue, there the build queue was cleared using (2). Any interaction with (3) and (4) the same frame would then silently fail. If your mouse is over another factory then (3) and (4) would consistently fail as the build queue is reset every frame. At this moment the build queue is unresponsive. This can happen when you use hotkeys.

It can occasionally happen if you use the mouse to interact with the build queue. The first click will usually fail, as when you move with the cursor the chance that it hovers over another factory at some point is high.

The idea that the bug occurs more late game than early game makes sense - there are more factories late game then there are early game. Therefore the chance that the bug occurs increases.

PR #5763, opened 2023-12-22
Code Formatting Units V3

MrRowey:


units/UEB1202/UEB1202_script.lua
```diff
@@ -10,36 +10,49 @@
...
-        if not self.AnimationManipulator then return end
-        self.AnimationManipulator:SetRate(0)
-        self.AnimationManipulator:Destroy()
-        self.AnimationManipulator = nil
+        local animManip = self.AnimationManipulator
+
+        if not animManip then return end
+        animManip:SetRate(0)
+        animManip:Destroy()
+        animManip = nil
```
Garanas, line 28: This should be `self.AnimationManipulator = nil`, or the field in the unit instance is never set to nil

units/UEB1301/UEB1301_script.lua
```diff
@@ -20,9 +25,9 @@ UEB1301 = ClassUnit(TEnergyCreationUnit) {
     ActiveState = State {
```
Garanas, line 25: Does the unit ever switch to this state?

PR #5772, opened 2023-12-24
Fix rapid fire ground to air flak

### LINKED ISSUE #5771: Seraphim & Aeon T2 AA turret rapid fires after update 3777 but with same damage per shot, resulting in skyrocketing DPS
**Describe the bug**
Previously before update 3777, Aeon's T2 AA turret could fire only once per 1 ticks. But post-update 3777 it rapid fires with the same damage, and in small enough number (20) could take down air experimentals much faster compared to other factions'.

Same behavior observed with Seraphim's T2 AA turret.

**To Reproduce**
Start a sandbox game in a map with civilian enemy, and either:  
- Put player-owned 20 Aeon T2 AA vs hostile civilian Soul Ripper, or better  
- Put one hostile civilian Aeon T2 AA vs player-owned CZAR in hold-fire, observe the damage dealt to CZAR per tick.

**Expected behavior**
Seraphim and Aeon T2 AA turret's DPS matches with DPS stated in the description.

**Screenshots**
-

**Additional context**
-

Garanas:
Closes #5771


The-Balthazar:
I am concerned that the fix to this is a blueprint change not a script change. That does not bode well for mod compatibility.

PR #5717, opened 2023-12-02
Fix multiple fire cycle bugs

lL1l1:
Resolves #5576.
I'm unsure about the RackSalvoFiringState OnLostTarget change. It would be overwriting https://github.com/FAForever/fa/pull/4863 and I haven't been able to break GC laser to fire forever or ythotha storm to not deal damage.
Also I used `MATH_IRound()` because it seems to round the same way as the game does. For example, Pillar has 0.8 fire rate, which turns into 12.5 ticks wait time, which rounds to 12 with `MATH_IRound()` and 13 with `math.floor(x*10+0.5)`. Testing shows a 12 tick wait time. The reasoning is that 0.8 in floating point is 0.800000011920928955078125, which gives a reciprocal of slightly below 1.25.

PR #5777, opened 2023-12-27
Fix weapons that fire after charging skipping fire rate

### LINKED ISSUE #5771: Seraphim & Aeon T2 AA turret rapid fires after update 3777 but with same damage per shot, resulting in skyrocketing DPS

lL1l1:
Fixes #5771, caused by #5717. The bug is similar to #5555. Even though #5772 shows that it can be fixed through bps (and some refactoring of effects, which is not included there), this fixes the weapon behavior for mod compatibility as mentioned on discord by @The-Balthazar.

### DIFF
--- a/lua/sim/defaultweapons.lua
+++ b/lua/sim/defaultweapons.lua
@@ -1177,7 +1177,9 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
             self:WaitForAndDestroyManips()
 
             local hasTarget = self:WeaponHasTarget()
-            local autoFire = not bp.ManualFire
+            -- Weapons that fire after charging will ignore the fire rate if we don't send them to the idle state
+            -- and if we send them to the fire ready state instead, they will ignore charge effects
+            local autoFire = not bp.ManualFire and not bp.RackSalvoFiresAfterCharge
 
             if hasTarget and bp.RackSalvoChargeTime > 0 and autoFire then
                 ChangeState(self, self.RackSalvoChargeState)


lL1l1:
Upon further investigation, there is a bug with double firing on the first shot before 5717, after 5717, and after this fix because the initial sequence is `IdleState.OnGotTarget -> Charge -> Firing -> IdleState -> IdleState.OnFire -> Charge -> repeat`. This causes aeon/sera static/mobile flak to fire when they get a target, even if they aren't aiming at the target, and without triggering the firing cooldown. In addition, it allows them to fire while on hold fire by giving an attack order.
#### How about we just remove `RackSalvoFiresAfterCharge`?
It doesn't fit into the logic of the firing cycle, and causes issues because of that. It's only use seems to be visual effects, which are relatively minor.
Units where `RackSalvoFiresAfterCharge = true`:
- UAA0310 (czar side flaks), UAL0401 (GC tractor claws): have a charge time of 0, so the value has no effect
- URA0104 (Cyb t2 transport): Has a charge time, but no associated effects
- XSB2204, XSL0205 (sera flaks): Has a charge time, and an effect, but it is misassigned to `FxChargeEffects` (copy paste from aeon) instead of `FxRackChargeMuzzleFlash` so it was never used.
- UAB2204, UAL0205 (aeon flaks): Charge time, script, and effect work, to create a muzzle flash-like effect one tick before the actual firing happens. The dotted lines are the charge effect.
Flak effect

##### Replication of effect but fixing force firing:
To replicate the effect correctly, we'd add `RackSalvoDelay` (trigger the effect 1 tick before all muzzles fire) and `PlayFxRackFiringSequence()` (the effect to use once per rack firing, instead of on every muzzle).
Alternatively, since aeon flaks only have 1 muzzle, you could abuse that fact with `MuzzleChargeDelay = 0.1` and overwriting `PlayFxMuzzleChargeSequence(muzzle)` to create effects on the charge effect bones.

Garanas:
We can remove it from units that we ship with FAF. But we can't remove it all together as there may be mods that rely on it.

Garanas:
@lL1l1 is this pull request complete on its own? If so, I'll merge it in and focus on the hotfix.

lL1l1:
Yes it is.

Garanas:
Included with the commit, that references this pull request.

Garanas:
@lL1l1 I'm fine with removing `RackSalvoFiresAfterCharge` if that means units behave more rationally.

PR #5779, opened 2023-12-27
Further tweaks to weapon details

lL1l1:
- Charge time was being counted even if it was less than the fire rate cooldown. This image is of the Ascendant (aeon mobile flak). It should be 0.5 reload time.
image
- Charge time should also be counted every rack firing, for example the cybran TML was saying it had 6.0 reload time, but in reality it is 10.0. Now it says 9.9, because inexplicably, after reloading, it waits 2 ticks from idle to idle.onfire instead of 1 tick.
- Avoid incorrectly saying a unit fires a salvo for units like mantis and blaze who have multiple racks but are only controlled by fire rate. Mantis showing a salvo:
image


Garanas:
Closed in favor of https://github.com/FAForever/fa/pull/5798

PR #5782, opened 2023-12-30
Fix issues with Aeon tactical missile defenses

Garanas:
People reported various issues specifically to the Aeon tactical missile defense (TMD). Other TMDs would not always respond to incoming threats because a nearby Aeon TMD was triggered, but out of reach of the missile to intercept it.

Issues that are fixed:

- (1) Missiles are now damaged by the flares instead of having their lifetime reduced. The damage is sufficient to kill any missile that can be intercepted by the other factions. 
- (2) The target check interval was set to 1.0 seconds for the Volcano due to a bug. It now has the same target check interval as any other anti-projectile weaponry (of 0.3 seconds).
- (3) The attack radius of the Volcano-like weapons was too large, which could cause it to lock on missiles that it can't possibly intercept.
- (4) The kill count of the Volcano would not update as missiles are intercepted.

The side effect of (1) is that missiles that are properly no longer trigger the `OnImpact` event, but the `OnKilled` event. Damage is only applied during the `OnImpact` event. This specifically matters to tactical-like missiles such as the Billy that would still explode. It also causes the Cybran missiles to split which creates all sorts of interesting effects.

The side effect of (2) is that the volcano can now properly intercept all tactical missiles. Specifically the Seraphim missiles are now easy targets for the Volcano. 

The side effect of (3) is that a lock would prevent other TMD to try and intercept a missile, even though the Volcano was unable to intercept the missile because the radius of the flare was too small. The attack radius was increased because I was unaware of (2).

The side effect of (4) is improved bragging rights.


Garanas:
https://github.com/FAForever/fa/assets/15778155/5f7587cd-c6bf-4ae9-9745-77788913f684

PR #5785, opened 2023-12-31
Fix Cybran nuke sub and SCU/ACU TML/Billy not impacting water

lL1l1:
It was reported on Discord that Billy was missing water targets, so I went to fix that but also found a few other missiles that can't impact water.

### DIFF
--- a/projectiles/SIFLaanseTacticalMissileCDR/SIFLaanseTacticalMissileCDR_script.lua
+++ b/projectiles/SIFLaanseTacticalMissileCDR/SIFLaanseTacticalMissileCDR_script.lua
@@ -44,6 +44,12 @@ SIFLaanseTacticalMissileCDR = ClassProjectile(SLaanseTacticalMissile, TacticalMi
         SLaanseTacticalMissile.OnCreate(self)
         self.MoveThread = self.Trash:Add(ForkThread(self.MovementThread,self))
     end,
+
+    ---@param self SIFLaanseTacticalMissileCDR
+    OnExitWater = function(self)
+        SLaanseTacticalMissile.OnExitWater(self)
+        self:SetDestroyOnWater(true)
+    end,
 }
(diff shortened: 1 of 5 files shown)


Garanas:
Similar to the Guass Cannon of the Valiant: https://github.com/FAForever/fa/commit/6333bdaac36110f89e3a021623d1c7edf801f838

Interesting how these issues managed to survive 13 years of gameplay.

lL1l1:
Nah, that's for Salem. In this case it's some functionality that we lost when refactoring the movement threads and cleaning up projectiles.

projectiles/SIFLaanseTacticalMissileCDR/SIFLaanseTacticalMissileCDR_script.lua
```diff
@@ -42,7 +42,20 @@ SIFLaanseTacticalMissileCDR = ClassProjectile(SLaanseTacticalMissile, TacticalMi
     ---@param self SIFLaanseTacticalMissileCDR
     OnCreate = function(self)
         SLaanseTacticalMissile.OnCreate(self)
-        self.MoveThread = self.Trash:Add(ForkThread(self.MovementThread,self))
+        self.MoveThread = self.Trash:Add(ForkThread(self.MovementThread, self))
+        self.Trash:Add(ForkThread(self.UpdateThread, self))
```
Garanas, line 46: What is the idea of the update thread?

Garanas (commented):
Projectiles have a second argument for the `OnCreate` function. It is a boolean that indicates whether the projectile starts underwater

lL1l1:
Using the `inWater` argument seems like a better solution. Not sure why that wasn't done in the original version of the script, but I couldn't find any bugs with using `inWater` to set water collision.

Garanas:
Maybe the original script was written before the `inWater` argument was introduced 😃

PR #5787, opened 2023-12-31
Fix Billy nuke bypassing shields

lL1l1:
Anything that uses the inner/outer damage model bypasses shields because it always uses either ring damage or the lua-side DamageArea that is made specifically for bypassing shields. This adds an exception for `Normal` damage type weapons without a time aspect to the area damage (only used for Yolona).
Considering mods, maybe it's better to add a separate class for inner/outer damage that doesn't bypass shields, since maybe there are nukes that use the normal damage type accidentally.


Garanas:
Closed in favor of https://github.com/FAForever/fa/pull/5789

PR #5788, opened 2024-01-01
Balance/improve new tactical missile behavior

### LINKED ISSUE #5683: Balance issues with new tactical missiles
### Seraphim
The increased Seraphim missile speed lets the Seraphim TML directly hit any TMD (including Aeon) and hit around 7 units behind a Cybran TMD (and probably more for other factions).
image
### Aeon
Aeon MML constantly overshoots and misses terribly, especially  near the minimum and maximum of its range.
15 Evensongs can barely kill this line of mass storages on +6 game speed:
image
Also, it may be worth looking into if the new aeon missiles have a slower travel time, which could impact their MML's effectiveness vs TMD.
### Cybran
Cybran missiles have terrible guidance after splitting (I think because they were made zigzaggers), leading to removing some gameplay elements and introducing RNG damage. In my opinion, the split missiles are meant to be reliable damage to make up for Cybran's 1 HP missiles, but now they are more like aesthetic fragments that happen to have a damage attached.
Specific Changes:
- Cybran TML can no longer critically damage TMD reliably. It used to be able to target a TMD and 2 split missiles would always hit the TMD for 933 damage (833 split dmg * 2 split hits * 0.56 special TMD armor). Cybran TMD has 950 hp for reference.
- Cybran TML can't reliably strike at targets behind TMD by using the split missiles' guidance; it's too random now. In my opinion it's a little too powerful for such an early game unit, and should only have short range guidance to kill TMD, but here is the defense for long range guidance nonetheless:
  - When grazing a TMD, other factions' 2 hp missiles would survive and deal 6000 damage (deleting wrecks). Cybran would have 3 split missiles and deal 2499 damage (destroying t2 pgens and t2 mex. Notably UEF t2 pgen is left at 1 hp).
  - At the cost of not destroying wrecks if grazing, cybran TMLs can fly directly over TMD to deal 1666 damage. This would cost 2 250 mass missiles to kill a mex and leave an around 200 mass wreck.
- Ironically, Cybran TML's split missiles are more accurate against targets far from the split than ones near the split location.
- Cybran MML can no longer chip away at TMD with 2x10 damage split missiles guaranteed every shot. It now can rarely hit full missiles on TMD if the split missiles *randomly* decide to fly over and behind the TMD. This doesn't make up for the missing chip damage since it's completely random.
A viper's split missiles completely missing their target:
image

- Cybran Strategic Submarine has completely ineffective split missiles, massively hindering its 1 HP missile compared to 3 HP aeon/uef. Considering 3 HP can fly over TMD, it should undoubtedly retain long range guidance on its split missiles. Before anyone mentions it, this does not balance out the stealth of the sub; that is balanced by the 9500 -> 10500 mass cost and 200 energy upkeep.

lL1l1:
Resolves #5683. You can view the commits individually to go faction by faction.
## Cybran:
- Missiles now split in a circular pattern, instead of the messy spread from before, and there are more options for how to split that are more precise than the old ones. TrackTarget and TurnRate don't need to be set in the component anymore as well. I noticed that this can also be used for mercy if desired. `SplitComponent`
-  `CIFSplit01`: Split missiles track their target instead of zigzagging into the ground. Their stats are also tweaked for the purpose of the unit changes.
- Per-unit changes to balance out Cybran 1 HP missiles:
  -  `CIF01`: Viper's split missiles spread out more and hit their target later, to act as decoys against TMD. In my opinion it becomes more effective against TMD, particularly in large numbers where more decoys let more full missiles through.
  -  `CIF03`: TML-4's old behavior is replicated. You can fly over or into a TMD with 2 split missiles, or graze the TMD with 3 split missiles at angles similar to the nerfed UEF TML here. This is done by making the split missiles spread out less so that they can quickly escape the edge of TMD, but making the main missile trajectory attack from above so that the TMD gets to shoot twice at a missile targeting it.
  - `CIF02`: The nuke sub's missiles were underpowered compared to 3 HP 2000 damage missiles from other subs because one TMD would reduce the damage to 1000. The child count is increased from 3 to 4 so that it can deal 1500 damage after passing one TMD, 75% of other subs instead of 50%. It also uses the same trajectory as the TML to help the TMD interaction be consistent between the two.

**Issue:** the split missiles can spawn camp the Volcano's flares, especially with Viper. They fly in circles and either collide with the flare or steal capacity from it. This can be fixed by reducing the maximum turn rate given to the split missile from 360 to lower than 120, but I'd prefer if the flare was able to modify the missile turn rate, make it turn upwards, or drastically decrease its lifespan (before Volcano fires again), because a lower turn rate makes it harder for the split missiles to hit the target if they are close to the ground. @Garanas

## UEF:
ACU and TML max speed reduced from 16 to 14. According to TMD range/reload, a TML should have max 16 speed. Accounting for turning time, it should be less than 16. 14 seems to be unable to hit TMD from my testing.

## Seraphim:
Their missiles now speed up in the middle of their flight and slow down at the end, before hitting TMD. Let's say it's the atmospheric resistance. This lets them be uniquely very fast but without breaking down balance at the victim's end. Since the ballistics math doesn't account for this changing max speed, I tested and made sure the missiles don't under/overshoot. The speed changing behavior is opt-in for compatibility.
Unit changes:
  -In general the missile acceleration is increased so that they can reach a higher max speed early.
  - SCU TML is made the same as the ACU's. It had no changes from before for some reason.
  -`SIF04/CDR/SCU`: TML, ACU, and now SCU missiles slow down to 13 speed outside of TMD range, guaranteeing for them to be destroyed when targeting TMD. They have a max speed of 20 so that they can't fly directly over TMD taking only one shot.
  -`SIF02/03`: Cruiser/Carrier missiles are slightly more effective by microing them towards a TMD, so that the missiles clump up and can actually hit the TMD.
  -`SIF01`: MML is barely changed because it already had a rather high velocity and doesn't have the fire rate or HP to let missiles clump up. The launchtick change just lets it target at close range without having the missile fly backwards.

## Aeon:
The missiles zigzag up to 50% less as they close in on their target in order to guarantee a hit, and the fuzzy targeting that they use is tweaked so that they don't overshoot at the upper part of the unit less and instead target the actual ground "in the vicinity of the unit".
Fuzzy targeting is also made to be used for ground fire as well, since groundfiring was sometimes far more accurate than attacking (Torrent). It is still more accurate because fuzziness takes into account the unit mesh size, but it's the best that can be done with the feature.
Evensong was very badly overshooting in some parts of its range because its missiles were going way too high for its 20 degree final boost angle. Now it can fire on a more direct line of fire, but in return it needs a 1 speed decrease to not randomly hit TMD in a 1v1.


Garanas:
You mentioned the work [here](https://github.com/FAForever/fa/issues/5683#issuecomment-1872528659), but I didn't fully perceive the consequences of them yet 😃 

> Fuzzy targeting is also made to be used for ground fire as well, since ground firing was sometimes far more accurate than attacking (Torrent). It is still more accurate because fuzziness takes into account the unit mesh size, but it's the best that can be done with the feature.

This was intentionally not the case. The primary reason for this is the Megalith:

https://github.com/FAForever/fa/assets/15778155/b100fd30-9c91-4ac4-b93e-0264ae2c2f4d

The secondary reason is that it is visually more interesting when tactical missiles consider the center of the unit and extend from there for their targeting. Yes, ground targeting makes it easier to hit the target. But normal units with non-missile based weapons do the same by default - why should missiles be different?

Note that for counted tactical missiles you always groundfire - you can't manually target a unit. For the missiles that cost you resources you always have the 'efficient' approach.

> The missiles zigzag up to 50% less as they close in on their target in order to guarantee a hit.

This is a good change, it makes the tactical missiles more reliable. 

> Missiles now split in a circular pattern, instead of the messy spread from before, and there are more options for how to split that are more precise than the old ones. TrackTarget and TurnRate don't need to be set in the component anymore as well. I noticed that this can also be used for mercy if desired. 

Great to see you found use of the split component. Call me nit-picky, but the old effect was visually more attractive, haha. I'm not going to fight you over this - whatever people prefer we'll do.

> Issue: the split missiles can spawn camp the Volcano's flares, especially with Viper. They fly in circles and either collide with the flare or steal capacity from it. This can be fixed by reducing the maximum turn rate given to the split missile from 360 to lower than 120, but I'd prefer if the flare was able to modify the missile turn rate, make it turn upwards,

It already does so, see [defaultantiprojectile.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/defaultantiprojectile.lua#L88). Maybe the collision triggers first, and it is overwritten by the missile script?

Garanas:
What if instead of decreasing the missile speed, we increase the fire rate of TMDs?

lua/seraphimprojectiles.lua
```diff
@@ -228,12 +231,37 @@ SLaanseTacticalMissile = ClassProjectile(SinglePolyTrailProjectile) {
...
+        if self.MinSpeed then
+            self.Trash:Add(self:ForkThread(self.SpeedUpdateThread, self.DistancePerSpeed, blueprintPhysics.MaxSpeed, self.MinSpeed))
+        end
+    end,
+
+    ---@param self SLaanseTacticalMissile
+    ---@param distancePerSpeed number
+    ---@param maxSpeed number
+    ---@param minSpeed number
+    SpeedUpdateThread = function(self, distancePerSpeed, maxSpeed, minSpeed)
```
Garanas, line 252: Would it be better to use the glide time and incorporate this behavior into the tactical missile component?

lua/sim/projectiles/aeon/AMissileSerpentineProjectile.lua
```diff
@@ -70,6 +70,22 @@ AMissileSerpentineProjectile = ClassProjectile(SingleCompositeEmitterProjectile,
         local blueprintPhysics = self.Blueprint.Physics
         local radius = 0.105 * (blueprintPhysics.MaxSpeed + blueprintPhysics.MaxSpeedRange)
         self:SetCollisionShape('Sphere', 0, 0, 0, radius)
+        self.Trash:Add(self:ForkThread(self.AccuracyThread, blueprintPhysics.MaxZigZag))
+    end,
+
+    ---@param self AMissileSerpentineProjectile
+    ---@param MaxZigZag number
+    AccuracyThread = function(self, MaxZigZag)
```
Garanas, line 78: Same here, it feels to be it is better to use the glide time and incorporate it into the tactical missile component.

lua/sim/projectiles/components/SplitComponent.lua
```diff
@@ -32,40 +32,60 @@ SplitComponent = ClassSimple {
...
-            local yVec = vy + MathCos(i * childConeSection) * spreadMultiplier +
-                (2 * spreadMultiplierRange * (Random() - 0.5))
-            local zVec = vz + MathCos(i * childConeSection) * spreadMultiplier +
-                (2 * spreadMultiplierRange * (Random() - 0.5))
+            local a = i * childConeSection + rotateOffsetRadRandom + (Random() - 0.5) * childRotateOffsetRange
+
+            local cosA = MathCos(a)
+            local sinA = MathSin(a)
+
+            -- Rotation into forwards-facing circle
```
Garanas, line 66: Very nice 😃

lL1l1:
I re-added fuzzy unit targeting with some improvements. One very interesting find was that offset unit hitboxes use the height of the terrain below the hitbox instead of the main unit bone. This causes my fuzzy targeting to partially miss hitboxes that are hanging off to the side of the unit on a slope, even though they are perfectly accurate against offset hitboxes on level ground.

image
90 missiles with 0 zigzag, 1 fuzziness, and 0 offset, they hit with 100% accuracy across the entire hitbox for megaliths on level ground. The ones aiming at the top miss because the hitbox is offset vertically and rotated based on the ground beneath it instead of the main unit bone.
Of course this is an extreme example, but at least it's good to know. For real units, I doubt it matters for even the most precise weapons.

lua/sim/Projectile.lua
```diff
@@ -156,21 +159,29 @@ Projectile = ClassProjectile(ProjectileMethods) {
             local mch = MathCos(heading)
             local msh = MathSin(heading)
 
+            local targetOrientation = Quaternion.new(unpack(target:GetOrientation()))
```
Garanas, line 162: This allocates two tables

lua/sim/Projectile.lua
```diff
@@ -156,21 +159,29 @@ Projectile = ClassProjectile(ProjectileMethods) {
...
+            -- correct the hitbox size and offset if it extends below the surface to prevent undershooting
+            if cy*2 < sy/2 then
+                sy = sy - (0.5 * sy + 2 * cy)
+                cy = sy/4
+            end
+            -- don't offset under the hitbox, we'll likely miss because projectiles come from above.
+            local dy = (Random() - 0.5) * (sy + offset * 0.5) + cy * 2 + offset * 0.25
+
+            local target = Quaternion.new(0, dx, dy, dz)
+            target = targetOrientation * target * targetOrientation:conj()
```
Garanas, line 183: This allocates a table

Garanas (commented):
Can you describe better what is being fixed with your last commit? It introduces several table allocations where we previously only had 1 table allocation (at the end, to create the target object).

lL1l1:
I hid my previous comment because it was caused by misunderstanding how Supcom uses quaternions (`y,z,x,w` instead of `w,x,y,z`).

> Can you describe better what is being fixed with your last commit?

1. The fuzzy targeting didn't account for 3d rotation, which is relevant with buildings and units on sloped terrain.
I remade it without allocating too many new tables. The math is done manually to avoid the extra operations in the first step with `dw = 0` and the second step in finding `dw` (the end value only needs `x,y,z`), and also because I saw HDT80bro's PR to remove that Quaternion library.
2. It wasn't using collision offsets correctly so missiles would miss or fly into the ground:
    - TMDs' hitboxes extend below the surface to allow torpedoes to hit them, but missiles are destroyed on water so they could never reach that part of the hitbox.
    - Some units have collision offsets that can make missiles miss, like the megalith.

Maybe the changes are overkill and not noticeable without 1 fuzziness, 0 offset, and 0 zigzag, but I think it's nice to have the correct behavior just in case someone does want to do that.

Garanas:
I understand the value now, and with the table allocations gone it's looking a lot better in my opinion!

Garanas:
@lL1l1 are you okay if I cherry-pick the changes to `OnTrackTargetGround` of your recent commits? Then we can include that with the hotfix

lL1l1:
Yeah sure, the changes in that area are finished.

lL1l1:
I re-implemented the speed and accuracy threads as a terminal phase for the tactical missile component. Basically some time/distance before the missile hits, it can slow down and change its zigzag amount.
The terminal speed could be improved to work with speedups.
You can opt out of the terminal zig zag change by setting it equal to the blueprint zigzag. For example for changing zigzag in your own script.

lL1l1:
I couldn't really figure out a way to make split missiles look nice when dying to flares: 
- the velocities got all messed up
- the missile could die to hitting the flare and the flare's kill thread in the same tick spawning split missiles twice
- the split missiles would just circle around the flare, even if it died, killing the next flares

Garanas:
@lL1l1 I think we've cherry-picked the majority of the suggested changes in this pull request. Unless I missed something, I think we can close this pull request 😃

lL1l1:
I still dislike how Cybran missiles behave, but yes, most of the changes are done and this pull request doesn't need merging.

PR #5790, opened 2024-01-01
Re-implement the missile redirect of the Loyalist

Garanas:
This was a tough one! Even with the [implementation of Equilibrium](https://github.com/FAForever/equilibrium/blob/master/equilibriumhook/lua/defaultantiprojectile.lua) anything in that direction was unreliable. For some odd reason the projectiles would spawn but half of them would immediately be destroyed again by the engine. The implementation of Equilibrium appears to be fighting the same behavior where it provides a 'retry' functionality.

If you can't beat them, join them. And that's what we do here - the missile redirect behavior is now part of a weapon that thinks it is a tactical missile defense. But right before it fires, we switch out the projectile blueprint and do all the logic to properly 'redirect' the missile. The new projectile is essentially identical to the old projectile and we even trigger the same movement thread of the tactical missile component! We skip the launch sequence of it.

https://github.com/FAForever/fa/assets/15778155/6f51378d-35ff-4ce1-b93d-cb69eae15668


Garanas:
Tennis!

https://github.com/FAForever/fa/assets/15778155/97b35dc8-2d9f-402c-9d3d-b8bb342d2513

Garanas:
@lL1l1 with regards to #5788 and Cybran missiles that split. I've adjusted the behavior a bit in this pull request while I was allowing them to be redirected. They now impact units just fine, assuming they do not get split right before their target.

Could you give it a view, see if it is sufficient for you?

Garanas:
https://github.com/FAForever/fa/assets/15778155/ba6ba002-d5c9-4f1e-ad35-c7451a206c7d

lL1l1:
> @lL1l1 with regards to #5788 and Cybran missiles that split. I've adjusted the behavior a bit in this pull request while I was allowing them to be redirected. They now impact units just fine, assuming they do not get split right before their target.
> 
> Could you give it a view, see if it is sufficient for you?

I prefer the more consistent behavior with my missiles.

PR #5795, opened 2024-01-04
Adjust collision size of air units when they are landed

Garanas:
Air units that are landed already had reduced vision. They now also have a reduced collision size. When air units are landed they change use their box-sized collision size. When they take off they use their sphere-sized collision size.

https://github.com/FAForever/fa/assets/15778155/baa34cbf-4496-4092-9649-e4ea13038255


Garanas:
The box-sized collision shape is also used by wrecks.

Garanas:
An interesting side effect: planes that are sinking also switch their collision shape! 

image

The-Balthazar:
Why are you inverting the effect of `RevertCollisionShape` for air units instead of using it to revert to their natural box shape? That's just going to cause unintended effects for anything that was already using the function.

The-Balthazar:
You're also supposed to half the size values given to `self:SetCollisionShape`. It uses them as radius instead of diameter, although that bug can be sidestepped by using revert to revert to box and setting to sphere instead of forcing it the other way round.

Garanas:
> Why are you inverting the effect of `RevertCollisionShape` for air units instead of using it to revert to their natural box shape? That's just going to cause unintended effects for anything that was already using the function.

I guess you could do that too. But it may be a bit unexpected to have an air unit that always has a sphere to 'revert' back to a box.

The-Balthazar:
Well they don't always have spheres. They start life as boxes and may never be spheres if they die while under construction. It's also changing the effect of an engine function, and that's never going to be good for compatibility. It'd be more unexpected to anyone that actually used the function before to suddenly have it stop working as expected and give probably the exact opposite of the previous effect. For no real gain. If you wanted to have a function to call to set the sphere it should be called `SetCollisionSphere` or something else new.

The-Balthazar:
While I was walking back from the dentist just now I thought of a good analogy; having revert take you to the sphere state is like telling a catapillar to revert into a butterfly.

Garanas:
@The-Balthazar alright - I see where you're coming from. How about now?

The-Balthazar:
I can't vouch for the specific size changes, but much better. I'd have replied earlier but I was in bed.

PR #5797, opened 2024-01-05
Working prototype of an alternative way for units to crash to the ground

Garanas:
_This is very much a draft at this moment_

How units crash is a bit of a web at the moment. There's the sinker projectile and the shield collider projectile. And then there's units with death animations. And all of that comes together with a lot of overhead, such as closures and threads and a lot of table operations.

I've been wondering: could we do this different? And the answer is: yes we can!

Features:

- [x] (1) A generic crash component
- [x] (2) A generic crash projectile
- [x] (3) All units that are ballistic can now bounce off shields
- [x] (4) All units can now technically crash
- [x] (5) All crashing units inherit their velocity when they start crashing

Fixes:

- [x] Hover units create wrecks that are on the ground now

Todo:

- [ ] (1) Integrate with death animations
- [ ] (2) Add functionality to create explosions
- [ ] (3) Add functionality to create debris
- [ ] (4) Apply death weapon damage to shields that it bounces off
- [ ] (4) Adapt the functionality of (2) and (3) so that you can customize it on a per-unit basis
- [ ] (5) Add functionality to manipulate the fall direction based on damage taken

Problems:

- [ ] What to do with units attached to a transport and the transport died?

### DIFF
--- a/effects/Entities/CrashProjectile/CrashAirProjectile_proj.bp
+++ b/effects/Entities/CrashProjectile/CrashAirProjectile_proj.bp
@@ -1,0 +1,12 @@
+ProjectileBlueprint {
+    Physics = {
+        CollideSurface = true,
+        CollideEntity = true,
+        DestroyOnWater = false,
+        RotationalVelocity = 60,
+        RotationalVelocityRange = 120,
+        UseGravity = true,
+        Lifetime = 100,
+        VelocityAlign = true,
+    },
+}
--- a/effects/Entities/CrashProjectile/CrashAirProjectile_script.lua
+++ b/effects/Entities/CrashProjectile/CrashAirProjectile_script.lua
@@ -1,0 +1,51 @@
+--**********************************************************************************
+--** Copyright (c) 2023 FAForever
+--**
+--** Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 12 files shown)


Garanas:
An example of the shield Bounce: 

https://github.com/FAForever/fa/assets/15778155/0196f12d-bc85-478c-ac8c-b1437c47ce9c

RabidPope:
Hi. If I am allowed to give some constructive criticism, the shield bounce does not look right. There seems to be a small delay before the bombers bounce back. Also they bounce back too far compared to normal.

Garanas:
> the shield bounce does not look right. There seems to be a small delay before the bombers bounce back. 

Yes, there's a single tick where it hiccups at the moment. We'll fix that in the future 😃 .

> Also they bounce back too far compared to normal.

That is adjustable with a parameter!

And of course, all feedback is welcome.

VaelophisNyx:
would this finally stop the death donut from feeling like it directly targets something expensive to fall on when it dies? (seriously I have never seen a Czar *not* fall on something valuable unless it gets shot down faaaar from anything it can land on)

MrRowey:
@Garanas It's a Shame this was not finished, would be nice to have.

BlackYps:

PR #5799, opened 2024-01-05
Rework/economy cost eye

Garanas:
_continuation of #5743_

The remote viewing script as a whole is a bit of a mess, but this attempts to address the clunky UX of trying to activate the ability when you have less than 10k energy in storage by being a drain event and a wait instead of a check and a take event.

~~Current issue with it though, is that this breaks the UI for the maintenance drain. (At least on the Steam version; the drain still happens, but it doesn't show on the unit.)~~


Garanas:
and I did it wrong 😃 !

PR #5773, opened 2023-12-25
Add UEF ACU silo hatch animation

lL1l1:
Adds an animation for the hatch with speed based off the reload time. Helps with #5770.

https://github.com/FAForever/fa/assets/82986251/361c0f5d-6ce8-4287-a345-3708bbcef161

Also removes the two `WeaponRepackTimeout` which aren't needed.

Issue #5770, opened 2023-12-24
Visualize the `RackReloadTimeout` blueprint value


Garanas:
This issue is related to #5769, but in general it is related to any strategic asset including the Billy of the UEF ACU.

The rack reload time is invisible to the user. It is unclear that the reason the weapon doesn't fire is because of a 'scripted' reload time. We'll need to find a way to visualize this.

A couple of observations:

- The `RackReloadTimeout` is primarily applied to counted weapons.
- We can not use the shield or work progress bars as they are usually already in use.
- We can not create a generic effect because of the diversity of the units (some are above water, some are underwater, etc).

Garanas:
One option would be to create an entity that is attached to the unit in question. The entity is only visible to the player and its allies. This can be done via:

- `function Entity:SetVizToAllies(type)`
- `function Entity:SetVizToEnemies(type)`
- `function Entity:SetVizToFocusPlayer(type)`
- `function Entity:SetVizToNeutrals(type)`

The entity is destroyed when the weapon is ready to fire again.

lL1l1:
`RackReloadTimeout` determines how long the unit has to wait in the idle state until the racks are considered reloaded. It's practically only for the visual effect of always firing from the first rack when entering combat. I think you mixed it up with `RackSalvoReloadTime`.
The linked dummy entity idea sounds useful, and it could also be utilized for build progress on naval nukes, which currently is concurring with the TML/battleship fire rate cooldown progress. To implement it, you could use the work progress bar, scale/offset it so it renders below the current bar, then set the progress to the minimum of the build or reload progress.

Garanas:
You're right, my mistake 😃 !

You're taking it a step further, where we introduce an additional unit (much like a dummy unit). That may work out just fine too. After all, Cybran build drones are units and they barely impact the simulation speed.

PR #5802, opened 2024-01-05
Add UEF ACU silo hatch animation

Garanas:
_Continuation of https://github.com/FAForever/fa/pull/5773_

Adds an animation for the hatch with speed based off the reload time. Helps with #5770.

https://github.com/FAForever/fa/assets/82986251/361c0f5d-6ce8-4287-a345-3708bbcef161

Also removes the two `WeaponRepackTimeout` which aren't needed.


Garanas:
@lL1l1 in general - please be aware of closures. A closure is created each time the `function` keyword is read by the interpreter. In your implementation, the function that is forked to close the hatch would create a closure each time a missile is launched.

PR #5808, opened 2024-01-06
Fix/ai basemanager typeset

relent0r:
This PR fixes an issue with the BaseType that is set on AI bases when they are created.

Due to the marker name being used instead of the type the AI could never understand that it already had Naval bases which caused it to constantly make new ones. On naval maps this results in too many naval expansions being created.

This will have also caused issues with some of the UnitCountBuildConditions that also relied on this property being a specific string rather than a marker name.

This was deceptively hard to track down :)

The ramification of this change is that the AI's naval expansion will be weaker now as its not tuned correctly. It won't ruin its economy and cause the game slow downs on naval maps.


relent0r:
Closing as I forgot to fix my local repo.

PR #5810, opened 2024-01-07
Rotate every other texture layer

BlackYps:
I discovered that rotating some texture layers makes the repeating texture pattern less visible.
Now not all textures are aligned in the same direction and the 30° rotation from the horizonal/vertical grid also seems to make it harder for the eye to notice repeating features.
The effect is not amazing, but it's almost free, so it still seems like a good deal.
The only question is, on which shaders do we want to apply that?

Before:
Screenshot 2024-01-07 015257

After:
Screenshot 2024-01-07 015310

### DIFF
--- a/changelog/snippets/graphics.5810.md
+++ b/changelog/snippets/graphics.5810.md
@@ -1,0 +1,3 @@
+- (#5810) Rotate every other texture layer in the Terrain301 shader
+
+By rotating every other texture layer by 30° we can make it harder to spot texture repetition. At the moment the Terrain301 shader is only used by the Sunset biome of the map generator. Hand-made maps don't use it yet, because there is no support in the map editor.
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -2707,6 +2707,9 @@ technique Terrain151 <
 float4 Terrain301NormalsPS ( VS_OUTPUT inV, uniform bool halfRange ) : COLOR
 {
     float4 position = TerrainScale * inV.mTexWT;
+    // 30° rotation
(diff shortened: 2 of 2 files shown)


Garanas:
@BlackYps what would be the behavior if we apply this to existing shaders, for example on the basic shader used on Seton's Clutch?

BlackYps:
well, it would basically be the same behaviour I guess. I'm not sure what your question is.
You only see an effect when you are at least somewhat zoomed in because of the texture blurring. So when you are zoomed out it doesn't change anything.
There could be a problem if a map uses a texture like floor tiles or something similar, because that would look odd. But I don't know if there even is a map that does this.

BlackYps:
What is your plan with this?

MrRowey:
could you provide some other images based of how this will look on other well know maps ?

BlackYps:
This PR doesn't affect any existing maps other than the sunset biome of the map generator. To my knowledge there are not other maps yet that use this particular shader

BlackYps:
If you wanted to see how this would look on existing maps to judge if this should also apply to other shaders I can of course set something up. In this case it would be good to know if you have particular maps in mind.

MrRowey:
ahh that fine didnt see where it specified that it was for the map generator

MrRowey (commented):
Just need New Snipit Adding then its good to go

MrRowey:
> If you wanted to see how this would look on existing maps to judge if this should also apply to other shaders I can of course set something up. In this case it would be good to know if you have particular maps in mind.

for this I think it something that you could ask the creative or matchmaker teams ?

MrRowey (approved):

PR #5811, opened 2024-01-07
Fixes for default templates and clarify template hotkey names

lL1l1:
- People are often confused by "cycle templates" and "build templates", so they are changed to "cycle context templates" and "cycle user templates" respectively.
- Standardize factory build queue template hotkeys based off the options menu text. Most importantly the hotkey to create new build queue templates now appears with the hotkeys to use build queue templates when you search.
- Omni could not be appended pgens to. Tech 2 radar could only have pgens appended because of the TML template.
- The TML template was appending t1 power generators to all tech 2 structures, on top of being a poor strategic decision.
- Placing t1 power next to energy storage is not often enough a good idea to require templates.


Garanas:
Removing templates from the base game is not the right approach. It will just confuse players (like myself) that got used to them existing

lL1l1:
Do you or others use the TML template or energy storage template? I can't see anyone ever using that TML template, and the energy storage template is a trap that shouldn't be encouraged building in the base game.
And we've removed base game functions before through hovering over deposits, which was far more useful than these templates.

lL1l1:
I'm still against e storage and tml power appending templates but this isn't really the point of this PR (fixing the templates).
T3 artillery template was applying to salvation, so I just added a salvation capping template as part of fixing that.

PR #5812, opened 2024-01-07
Add new track target ground logic

Garanas:
Logic changes are cherry-picked from #5788, all credits to @lL1l1

### DIFF
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -198,7 +198,7 @@ Projectile = ClassProjectile(ProjectileMethods) {
             local pos = self:GetCurrentTargetPosition()
 
             local physics = self.Blueprint.Physics
-            local fuzziness = 0
+            local fuzziness = physics.TrackTargetGroundFuzziness or 0.8
             local offset = physics.TrackTargetGroundOffset or 0
             local dx = (Random() - 0.5) * fuzziness * (1 + offset)
             local dz = (Random() - 0.5) * fuzziness * (1 + offset)
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -154,12 +154,10 @@ Projectile = ClassProjectile(ProjectileMethods) {
             local sx, sy, sz = unitBlueprint.SizeX or 1, unitBlueprint.SizeY or 1, unitBlueprint.SizeZ or 1
             local px, py, pz = target:GetPositionXYZ()
 
-            -- add offset to prefer the lower end of the hitbox
-            if cy == 0 then
-                cy = sy / 4
-            elseif cy * 2 < sy / 2 then
-                sy = sy - (0.5 * sy + 2 * cy)
-                cy = sy / 2
+            -- don't target the part of the hitbox below the surface
+            if cy < 0 then
+                sy = sy + cy
+                cy = 0
             end
 
             local physics = self.Blueprint.Physics
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -178,16 +178,20 @@ Projectile = ClassProjectile(ProjectileMethods) {
             -- Supreme Commander quaternions use y,z,x,w!
             local ty, tz, tx, tw = unpack(target:GetOrientation())
 
-            dw = -tx * dx - tz * dy - ty * dz
-            dx = tw * dx + tz * dz - ty * dy
-            dy = tw * dy + ty * dx - tx * dz
-            dz = tw * dz + tx * dy - tz * dx
+            -- compute the product in a single assignment to not have to use temporary, single-use variables.
+            dw, dx, dy, dz = 
+            -tx * dx - tz * dy - ty * dz,
+             tw * dx + tz * dz - ty * dy,
+             tw * dy + ty * dx - tx * dz,
+             tw * dz + tx * dy - tz * dx
 
             tx, tz, ty = -tx, -tz, -ty
-
-            dx = dw * tx + dx * tw + dy * ty - dz * tz
-            dy = dw * tz + dy * tw + dz * tx - dx * ty
-            dz = dw * ty + dz * tw + dx * tz - dy * tx
+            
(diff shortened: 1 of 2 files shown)


lua/sim/Projectile.lua
```diff
@@ -141,20 +141,26 @@ Projectile = ClassProjectile(ProjectileMethods) {
...
-            local heading = -1 * target:GetHeading() -- inverse heading because Supreme Commander :)
-            local mch = MathCos(heading)
-            local msh = MathSin(heading)
+            -- add offset to prefer the lower end of the hitbox
+            if cy == 0 then
+                cy = sy / 4
+            elseif cy * 2 < sy / 2 then
+                sy = sy - (0.5 * sy + 2 * cy)
+                cy = sy / 2
+            end
```
Garanas, line 163: I noticed the subtle preference here to target the lower end of a unit, is that there to fix a bug or to increase the accuracy of missiles @lL1l1 😃

lL1l1: It's not meant to bias the offset in any direction, only change the targeting volume to the above-surface part of the hitbox.
I think I made a mistake with the parentheses. It should be `sy = sy - 0.5 * sy + 2 * cy`
image
Also there's a mistake in `cy = sy / 2` where it should be `cy = sy / 4`

lua/sim/Projectile.lua
```diff
@@ -163,19 +169,41 @@ Projectile = ClassProjectile(ProjectileMethods) {
...
 
-            self:SetNewTargetGround(target)
+            local pos = { px + dx, py + dy, pz + dz }
+            DrawCircle(pos, 1, 'ffffff')
+            self:SetNewTargetGround(pos)
         else
             local pos = self:GetCurrentTargetPosition()
+
+            local physics = self.Blueprint.Physics
+            local fuzziness = 0
```
lL1l1, line 199:
```suggestion
            local fuzziness = physics.TrackTargetGroundFuzziness or 0.8
```

lua/sim/Projectile.lua
```diff
@@ -163,19 +169,41 @@ Projectile = ClassProjectile(ProjectileMethods) {
...
+            dw = -tx * dx - tz * dy - ty * dz
+            dx = tw * dx + tz * dz - ty * dy
+            dy = tw * dy + ty * dx - tx * dz
+            dz = tw * dz + tx * dy - tz * dx
+
+            tx, tz, ty = -tx, -tz, -ty
+
+            dx = dw * tx + dx * tw + dy * ty - dz * tz
+            dy = dw * tz + dy * tw + dz * tx - dx * ty
+            dz = dw * ty + dz * tw + dx * tz - dy * tx
```
lL1l1, line 190: These changes would break the code because variables are assigned new values in the middle of the computation. That is why the original code only used 1 assignment to change all the values.
```suggestion
            -- compute the product in a single assignment to not have to use temporary variables.
            dw, dx, dy, dz = 
            -tx * dx - tz * dy - ty * dz,
             tw * dx + tz * dz - ty * dy,
             tw * dy + ty * dx - tx * dz,
             tw * dz + tx * dy - tz * dx

            tx, tz, ty = -tx, -tz, -ty
            
            -- compute the product in a single assignment to not have to use temporary variables.
            dx, dy, dz = 
            dw * tx + dx * tw + dy * ty - dz * tz,
            dw * tz + dy * tw + dz * tx - dx * ty,
            dw * ty + dz * tw + dx * tz - dy * tx
```

lua/sim/Projectile.lua
```diff
@@ -163,19 +169,41 @@ Projectile = ClassProjectile(ProjectileMethods) {
...
 
-            self:SetNewTargetGround(target)
+            local pos = { px + dx, py + dy, pz + dz }
+            DrawCircle(pos, 1, 'ffffff')
```
lL1l1, line 193: Don't forget to remove the DrawCircle debug.

lua/sim/Projectile.lua
```diff
@@ -163,19 +169,41 @@ Projectile = ClassProjectile(ProjectileMethods) {
             sy = sy + offset
             sz = sz + offset
 
-            local dx = (Random() - 0.5) * fuzziness * sx
-            local dy = (Random() - 0.5) * fuzziness * sy
-            local dz = (Random() - 0.5) * fuzziness * sz
+            local dx = sx * (Random() - 0.5) * fuzziness
+            local dy = sy * (Random() - 0.5) * fuzziness + cy * 2
+            local dz = sz * (Random() - 0.5) * fuzziness + cz * 2
+            local dw
```
lL1l1, line 175: I placed `dw` in the rotation computation section because it is a temporary variable needed for the actual rotation and nothing else. Frankly you could replace `dw` with `(-tx * dx - tz * dy - ty * dz)` in the second multiplication and explain it with a comment, then you wouldn't need the variable at all.

lL1l1 (changes requested):

lua/sim/Projectile.lua
lL1l1: > It's not meant to bias the offset in any direction, only change the targeting volume to the above-surface part of the hitbox. The comments in the original code explained this.

I don't know how I made the mistake of thinking the offset is 2x or acted so weirdly.
```suggestion
            -- don't target the part of the hitbox below the surface
            if cy < 0 then
                sy = sy + cy
                cy = 0
            end
```

lL1l1: Unexpectedly megalith changes its collision box when it finishes building to values different from its blueprint. That's what I was compensating for in my testing process when writing the code.

lL1l1:
I fixed the megalith hitbox issue and put all the correct changes together into [2d1296e](https://github.com/lL1l1/fa/commit/2d1296e166a1d979b75205796eeb95eb2af58e6e).

lua/sim/Projectile.lua
Garanas: Ah, my mistake. That is a bit confusing, haha

PR #5823, opened 2024-01-13
Fix offsets in ground targeting related to Megalith

lL1l1:
Since I was testing on Megalith when writing the code, I got some illogical math for the offsets. This makes the math work for normal units and fixes Megalith's hitbox modification.
I also checked and found no other units that modify their hitbox incorrectly.


Garanas:
The Galactic Colossus changes hit box too while building, if I'm not mistaken

lL1l1:
Yeah I noticed, but it uses `RevertCollisionShape()` when it finishes.

PR #5824, opened 2024-01-13
Apofenas storage suggestion

Apofenas:
Remove storages from engineers and factories.
Increase amount of stored mass in Mass storage by 2.


MrRowey:
After Discussion, the balance team does not see this as a necessary change at this time but does appreciate the vision and direction this was aimed towards.

PR #5827, opened 2024-01-13
Allow SACUs to shoot ACUs while moving

Basilisk3:
- Adds the "SNIPEMODE" category to all SACUs
- Adds the "COMMAND" target priority to all SACUs


lL1l1 (changes requested):
- Defense structures should not be moved below tech 1 priority because it doesn't make sense for a later game T3.5 unit.
- The overcharge's defense structure priority should be moved above tech 1.
- The `COMMAND` priority could also be set above tech 3 in my opinion because of how large of a threat overcharge can be to SACUs that late in the game. Except for overcharge, where it should be above tech 1 because overcharge deals limited damage to ACUs.

Garanas:
What if we turn it into the following for the default weapon:

```lua
TargetPriorities = {
    "MOBILE EXPERIMENTAL",
    "COMMAND",
    "MOBILE TECH3",
    "MOBILE TECH2"
    "STRUCTURE * DEFENSE",
    "ALLUNITS",
},
```

The SACU will prefer experimental units and the ACU over regular units because of how much damage Overcharge can do to a SACU.

Tech 1 units are removed all together as it does not make sense that a SACU should take the time to rotate around and firing one shot at a tech 1 unit over just firing at what is in front of the SACU on that moment. 

The same can apply to tech 2 units, but I'm not sure where to draw the line as some tech 2 units can be a real threat to SACUs.

I'd also remove the 'no anti missile' line as that implies that tactical missile defenses are likely to survive even though they have low health. The sudden lack of tactical missile defenses can open up a base to future missile strikes.

And the following for the Overcharge on the Seraphim SACU:

```lua
TargetPriorities = {
    "MOBILE EXPERIMENTAL",
    "MOBILE TECH3",
    "ALLUNITS",
},
```

It doesn't make much sense to overcharge an ACU in my opinion, hence the category is removed. It also does not make sense to overcharge tech 1/2 units or structures over just overcharging what is in front of the SACU because it can cause the SACU to rotate and that may cost more DPS then the overcharge would accomplish all together.

lL1l1:
Not targeting TMD is a common theme across all units, I wouldn't change it for that reason as well as PD having a lot of DPS/range so it needs to be prioritized. Mentioning that, there could be a target priority for all surface units of Mobile directfire > PD > TMD > AA, but I'm not sure how expensive complex target priorities are.

For overcharge it doesn't make sense to exclude Tech 2, ACU, or PD because it aims in the same direction as the main gun, so usually it won't require enough turning to miss DPS.
-  main gun: 400 damage per 1 second
-  overcharge: 1250+ (units) or 800 (structures) or 400 (ACU) damage per 7.5 seconds
- 180 turret yaw rate, 150 body turn rate

The gun would have to turn for overcharge if it's chasing an ACU and sees a better target.
It would usually take less than a second for the SACU to retarget and target back while walking in a straight line. If it's turning, the user would have to micro the turn in the wrong direction for the 30 net yaw to last over a second. So you need more than a 400 damage difference for it to be worth retargeting.
Since PD cannot be killed by a single overcharge, and the main gun won't be targeting PD to finish it off while chasing an ACU, it's not worth targeting despite dealing 800 damage with splash multiplying that.
Tech 2 is over 1250 damage, so it would usually result in a lot more damage dealt, and enemy DPS would actually die, so it's worth putting over PD and ACUs.
```
TargetPriorities = {
    "MOBILE EXPERIMENTAL",
    "MOBILE TECH3",
    "MOBILE TECH2",
    "COMMAND",
    "(STRUCTURE * DEFENSE - ANTIMISSILE)",
    "ALLUNITS",
},
```
Not having the target ACU or PD priorities at all could result in the OC firing at economic units/low HP spam in the middle of the main gun turning.

Garanas:
@Basilisk3 can you remove the tech 1 categories line? After that it's fine with me. 

I also noticed that the categories of the Seraphim SACU are quite different.

PR #5828, opened 2024-01-14
Improve the descriptions of various units

Basilisk3:
WIP


Garanas:
@Basilisk3 is this still a work in progress? I'm fine with merging this and you making other changes in a different pull request

Basilisk3:
Yes, I am working on it currently.

Basilisk3:
This is ready for review now.

The-Balthazar:
While I agree that a centralised spreadsheet/database is a would be good overall, there's no reason to assume the kind of issues this is fixing would have been reflected in human-done translations to begin with.

Basilisk3:
I will go through Cybran next. It would also probably be good if we could standardize which version of english we use, right now it is a mishmash of British English and American English.

Basilisk3:
Technically, the folder is called "US" 😃. But I will make that decision once I have corrected the typos/outdated info first, I think those are more pressing at the moment.

The-Balthazar:
As a Brit it pains me to say that it should be US English to be consistent with the base game, and the folder structure. If we're doing UK/US English properly, then it should be in those folders.

PR #5829, opened 2024-01-15
Fix an error when the Salem makes landfall

Basilisk3:


Garanas:
Does it show any effects? I can't seem to find them

Basilisk3:
I noticed no changes visually, so this just fixes the error in the log. I assumed that maybe at some point there was an effect, or someone planned to introduce one.

PR #5830, opened 2024-01-15
Rework the FPS counter to be less obstructive and more modern

Basilisk3:
This PR orients itself on other, modern FPS counters, like the one Steam provides.

- The counter now takes up much less screen space
- The numbers do not constantly jump around anymore, as fixedFont is used instead of bodyFont
- Unnecessary decimal places have been removed
- The counter does not obstruct other UI-elements anymore

To access the FPS counter in-game, go to: Key Bindings --> Debug --> "Toggle the display of frames rendered per second"

### DIFF
--- a/lua/debug/uidebug.lua
+++ b/lua/debug/uidebug.lua
@@ -10,25 +10,31 @@ function ShowFPS()
         return
     end
     
-    fps = UIUtil.CreateText(GetFrame(0), '', 30)
-    fps:SetColor('ffffffff')
+    fps = UIUtil.CreateText(GetFrame(0), '', 12)
+    fps:SetColor('ffffff')
+    fps:SetFont(UIUtil.fixedFont, 12)
     fps:SetDropShadow(true)
     fps:DisableHitTest()
     fps.Depth:Set(GetFrame(0):GetTopmostDepth() + 1)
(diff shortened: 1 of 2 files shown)


Garanas:
https://github.com/FAForever/fa/assets/15778155/60537847-515a-42f1-b5a7-46738539cb1a

Basilisk3:
Thanks, some things noticed with these changes: 

- As σ fluctuates, the length of the output changes constantly, so the numbers jump around again
- Both numbers refresh too often, so they are hard to read

I noticed this, because I have my FPS unlocked and play with this command enabled automatically.

Garanas:
ah, that is why you limited the when you show the number. I'll add a limit to how often it updates too

Basilisk3:
As a point of reference, AMD's tool samples at a rate of 4 times a second. I think this is on the higher side, and some other tools only do twice a second.

Garanas:
@Basilisk3 can you re-introduce the code that I removed to reduce the update frequency of the text? Make sure it only slows down updating the text, and not the computations. What if we go for 4 times a second, like AMD does?

Basilisk3:
@Garanas Is this correct?

Garanas:
Now it is correct 😃 , do you see the difference?

PR #5832, opened 2024-01-18
Fix offmapping wrecks

lL1l1:
If an aircraft/transported unit creates a wreck outside the playable area, the wreck position is moved back inside.

image

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1704,7 +1704,7 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent) {
         end
 
         -- Create potentially offmap wrecks on-map. Exclude campaign maps that may do weird scripted things.
-        if IsHumanUnit(self) and not ScenarioInfo.CampaignMode then
+        if self.Brain.BrainType == 'Human' and (not ScenarioInfo.CampaignMode) then
             pos = GetNearestPlayablePoint(pos)
         end
 
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -39,6 +39,8 @@ local UpdateAssistersConsumptionCats = categories.REPAIR - categories.INSIGNIFIC
 
 local DefaultTerrainType = GetTerrainType(-1, -1)
 
+local GetNearestPlayablePoint = import("/lua/scenarioframework.lua").GetNearestPlayablePoint
+
 --- Structures that are reused for performance reasons
 --- Maps unit.techCategory to a number so we can do math on it for naval units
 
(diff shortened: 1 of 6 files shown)


lua/sim/Unit.lua
```diff
@@ -1700,6 +1702,11 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent) {
             energy = energy * 0.6
         end
 
+        -- Create potentially offmap wrecks on-map.
+        if EntityCategoryContains(categories.AIR, self) or self.killedInTransport then
```
Garanas, line 1706: What if we just apply it to all wrecks in general?

Garanas: And we probably want to only apply this behavior to skirmish maps, so that we exclude campaign maps.

lL1l1: > What if we just apply it to all wrecks in general?

I suppose there's no harm in applying it to all wrecks if you say so.

> only apply to skirmish maps

Wouldn't that break with scripted maps like survivals if they ever spawn wrecks offmap?
The ScenarioFramework.lua has a good idea of checking if the unit belongs to a human, I think it's the best option. Surely there aren't any scripted maps/campaigns that spawn human units offmap then kill them?

Garanas: You're right, let's start with human wrecks and see how we can tweak it from there in future pull requests. I'd still exclude campaign maps all together from this behavior. You can check the map type, for non-skirmish maps I'd not apply this behavior to any wreck:

- https://github.com/FAForever/faf-coop-maps/blob/master/FAF_Coop_Fort_Clarke_Assault/FAF_Coop_Fort_Clarke_Assault_scenario.lua#L5

lL1l1: I'll use ScenarioInfo.CampaignMode instead, it seems more correct because it's used in other places, and across my 788 FAF maps I could only find 20 that don't use the skirmish type, and those same maps' scripts use `ScenarioUtils.InitializeScenarioArmies()` which sets CampaignMode to true.

lua/sim/units/AirTransportUnit.lua
```diff
@@ -164,7 +164,7 @@ AirTransport = ClassUnit(AirUnit, BaseTransport) {
                 end
             end
             if not EntityCategoryContains(categories.COMMAND, unit) then
-                unit.killedInTransport = true
+                unit.willBeKilledByTransport = true
```
Garanas, line 167: with mod compatibility in mind - we can't suddenly change the name of this value

projectiles/Sinker/Sinker_script.lua
```diff
@@ -67,7 +67,8 @@ Sinker = ClassProjectile(Projectile) {
     ---@param targetType string
     ---@param targetEntity Prop|Unit
     OnImpact = function(self, targetType, targetEntity)
-        if targetType == 'Terrain' then
+        -- 'Underwater' is impacted when sinking off map for a long time.
+        if targetType == 'Terrain' or targetType == 'Underwater' then
```
Garanas, line 71: My guess is that the 'long time' is the moment where the projectile runs out of life time, and therefore it impacts being 'underwater'

Issue #4788, opened 2023-03-05
redundant entries for anti-torpedo weapons' `FireTargetLayerCapsTable`


haifron:
Description
-----------
Anti-torpedo weapons in the game have a variety of different values for this table, all being some subset of `Water|Seabed|Sub` (and the cooper's antitorp enthusiastically being able to attack `Land` too 😄).

#### list of the number of weapons with each variety of `FireTargetLayerCapsTable`
```Lua
19 'Seabed|Sub|Water'
1 'Sub|Water'
1 'Sub|Water|Seabed|Land'
1 'Water'
```
After some testing, I've concluded the only required layer is `Water` however, which the t3 aeon sonar's antitorp has.

Question
----------------
Do these excess entries cause the game to search more units/projectiles than it needs to? Like is there any point in making them all just `Water`.

Garanas:
These values are used to enable / disable targets as the unit switches layer. As an example: the destroyer disables the torpedo launchers as it walks on land. Submarines disable their deck guns as they submerge.

That they exist but can not be used has no impact on performance.

relent0r:
I would have thought they'd need seabed as well since thats the layer that things like the wagner switch to when they are under water.

haifron:
@relent0r I tested having each layer in the table (including all at once _except,_ `water`) with ship, sub, plane & amphibious launched torpedoes and the antitorp only ever killed torpedoes when `water` was included. So apparently not.

You can also test this yourself whenever with the T3 aeon sonar which only has `water`.

haifron:
> These values are used to enable / disable targets as the unit switches layer

@Garanas I'm actually referring to the values of the table, not the keys - the layers the weapon can fire _at_ not fire _from_. So in 
```Lua
FireTargetLayerCapsTable = {
    Water = 'Sub|Water|Seabed|Land',
},
```
I'm referring to `Seabed|Sub|Water` (not `Water`) - which are being used unless I'm mistaken. Regardless of the probably tiny performance change, if it's ok for me to change this just to make the blueprints more concise & precise then I'd probably like to do that anyway; the bps are mysterious enough as they are.

Garanas:
I miss understood - feel free to fix this for weapons where it makes no sense to be able to target a specific layer

PR #5834, opened 2024-01-18
Fix some issues in naval weapons

lL1l1:
- T3 sub got its DPS reduced in #5000. There are no patch notes or conversations mentioning it, and its behavior is buggy (5.1 second cycle with 4 second fire rate cooldown). I assume the changes were made to buff it because of new anti-torp effectiveness, but the bug with the firing cycle wasn't spotted?
Pre-Change: 1340 volley, 335 expected dps, 327 actual dps
Current: 1450 volley, 363 expected dps, 284 actual dps
After fix: 1450 volley, 363 expected dps, 363 actual dps
- Aeon anti-torps were targeting the same torpedo with two projectiles from a single weapon. Now it is split into two weapons with halved fire rate. I can put the fire rate back to what it should be if that is how it's supposed to work.
  - UEF/Cybran use flares on opposite sides of the boat so I assumed they are meant to fire together, and didn't split them.
- Aeon antitorps given similar minimum range to Seraphim destroyer. The T2 sub importantly had no minimum range, so its antitorps should now be a bit more effective.
- Cooper effectively had 52 water vision instead of 55 because of the intel grid, making it unable to see underwater within its full weapon range. Increased to 56.
- Valiant now correctly shows 275x2/4 damage/reload instead of 275/2.


lL1l1:
Fixes #4788.

lL1l1:
Yes

Garanas:
We're not going to fix these intel issues in this pull request, but I thought it was interesting to decipher how many units have inaccurate intel radius values 😃 . There are 440 in total, excluding mods of course

PR #5836, opened 2024-01-19
Setup to run tests on blueprint files

Garanas:
The visual representation of intel is... not quite the truth. Intel works in a grid-like fashion. The size of a cell for vision is 2 where as the size of a cell of all other intel is 4. As a result, certain intel values look visually bigger but they have no impact on your actual intel.

The grid can be seen using the console command `dbg Radar`. As a quick example, the walls in the following video have an radar radius of 28, 29, 30 and 31. For radar the size of a cell is 4. Therefore these all map to the same cells. Notice how inaccurate the radius is because of that.

https://github.com/FAForever/fa/assets/15778155/e4f770b8-7cb3-4d62-9d3f-cd4aa6d2fd8c

Occasionally players come to us with bugs related to intel - units that pop in/out of vision as you try to chase them. This effect is compounded by the incorrect/exaggerated visual ranges. This can cause frustrations. By properly aligning these values we can prevent a lot of these hiccups.

In combination with https://github.com/FAForever/FA-Binary-Patches/pull/61 intel should be more reliable than ever when chasing units.


Garanas:
@Askaholic would you have a suggestion on how to run a set of tests on each (unit) blueprint file?

ComradeStryker:
"(https://github.com/FAForever/fa/pull/5836) Revert of the 15% global vision range increase as introduced by game version 3761"

Some stealth ranges were increased to adjust for a global vision increase, was this also reverted?

Garanas:
No, that is not reversed. Consider it a buff for now 😃

PR #5842, opened 2024-01-20
Re-implement the billy: Prevent billy nuke detonating upon impacting flares

lL1l1:
Copies https://github.com/FAForever/fa/commit/8fb229dfa0891c813f184cc95810e67cfc6a72dc with the small refactoring needed for the new version.


Garanas:
Do you have a replay of where this happens? The flare should deal 30 damage to any tactical missiles that it collides with directly, that triggers before `OnImpact` does

lL1l1:
I was doing some testing on this branch and noticed the issue, but it seems to no longer happen, perhaps I just misremembered.

Garanas:
Make sure to do tests on the latest deploy/fafdevelop 😃 !

PR #5849, opened 2024-01-22
Salvation nerf

### LINKED ISSUE #5839: Balance idea: Salvation Nerf
## Intended changes

_A clear concise description of the changes_

After some discussion from the team, we agree that it needs a nerf but not as drastic as what is proposed in #5781 what we are proposing instead is the following changes

Edit new Proposed changes
```
- HP > 9k
- EnegeryRequired: 15k > 25k (Wasent updated when the Ecost per shot was increased)
- MuzzleChargeDelay: 0.15 > 0.2 (revert older changes)
- MuzzleVelocity: 150 > 98 (Revert older change)
- Min Range: 175 > 150 
```

## Thoughts on the changes

_Various thoughts on the reasoning behind these changes_

lL1l1:
Closes #5839.
I decided against reducing the visual scale and hitbox size because it is already quite small compared to other artilleries.
Top right is smaller in proportion to the footprint changes.
image


MrRowey (approved):

lL1l1:
@Garanas Since the template conditions use` - SIZE20` and `* SIZE20` for the t3 artillery and Salvation templates, the t3 artillery template automatically works with this change. Should the Salvation template be kept or removed in this case?

MrRowey:
@lL1l1 can you also add the following changes then we can merge for testing
```
- EnegeryRequired: 15k > 25k (Wasent updated when the Ecost per shot was increased)
- MuzzleChargeDelay: 0.15 > 0.2 (revert older changes)
- MuzzleVelocity: 150 > 98 (Revert older change)
```

lL1l1:
I didn't revert the `MuzzleVelocityReduceDistance` to 1750 with the muzzle velocity change because it caused it to fire at an abnormally low angle. Though that can be changed if you want it to be slightly better at shooting armies/uniqueness.

I noticed Salvation has 175 minimum range instead of 150 like all other arties, should that be changed?

#4441 didn't mention T4 arty getting their energy cost increased at all, so I wouldn't say it "wasn't updated when the Ecost per shot was increased". Should Mavor get the cost increase too? 
At a closer look, the efficiency of energy/shot is all over the place for different arties in general, and they're only balanced by energy/dps (a faulty metric in my opinion):
|Artillery|Charge Cost|Damage|RoF|E/Dmg|E/s|E/DPS|
|--|--|--|--|--|--|--|
|Salvation (nerfed)|25000|7524|0.32|3.32|8065|10.3|
|Mavor|30000|16000|0.125|1.88|3750|15|
|Emissary|17000|12000|0.05|1.42|850|28.3|
|Duke|16000|5500|0.1|2.91|1600|29.1|
|Hovatham|15000|5000|0.1|3.00|1500|30|
|Disruptor|14000|3700|0.13|3.78|1842|28.7|

_Fire rate tick rounding accounted for. Adjacency gives about 0.7x E/Dmg, 1.167x E/s, and 0.42x E/DPS_

MrRowey:
@lL1l1 After discussion for now can we just make the following changes and see how it performs if I need more nerfs we can do this.

```
- HP > 9k
- EnegeryRequired: 15k > 25k (Wasent updated when the Ecost per shot was increased)
- MuzzleChargeDelay: 0.15 > 0.2 (revert older changes)
- MuzzleVelocity: 150 > 98 (Revert older change)
- Min Range: 175 > 150 
```

lL1l1:
How is the DPS not a concern after some energy cost and HP nerfs?

MrRowey:
> Contributor

For now, were taking it back to as close to the 2016 balance and then once we get it there we can adjust it further if deemed necessary

lL1l1:
image
Textures could be improved for Salvation, they're quite blurry.

units/XAB2307/XAB2307_unit.bp
```diff
@@ -163,8 +163,8 @@ UnitBlueprint{
             DamageRadius = 2,
             DamageType = "Normal",
             DisplayName = "Rapid-Fire Artillery Cannon",
-            EnergyDrainPerSecond = 10000,
-            EnergyRequired = 15000,
+            EnergyDrainPerSecond = 17857,
```
MrRowey, line 166: keep this at 10k please

MrRowey (changes requested):

Basilisk3:
See my last comment in #5868. Unless there is another way of fixing this, its MuzzleVelocity should be increased and not decreased. Deceasing it only makes the bug worse. Though to be honest, I am not sure why this is changed to begin with. The Mavor and Scathis both have a MuzzleVelocity of 160, why should the Salvation only have 98?

units/XAB2307/XAB2307_unit.bp
lL1l1: That would break the firing cycle with adjacency buffs because the gun won't charge before the fire rate cooldown is finished, so it'll wait the fire rate cooldown twice.

MrRowey: if this is the case then best not to touch this and reset it back to the 10k & 15k.

lL1l1: That also can't be done because the 0.15 -> 0.2 muzzle charge delay change reduces the time the weapon has to charge by 1 tick, so it has to be changed to 15000 cost and 10714 drain.

units/XAB2307/XAB2307_unit.bp
```diff
@@ -163,7 +163,7 @@ UnitBlueprint{
             DamageRadius = 2,
             DamageType = "Normal",
             DisplayName = "Rapid-Fire Artillery Cannon",
-            EnergyDrainPerSecond = 10000,
+            EnergyDrainPerSecond = 10714,
```
MrRowey, line 166: Whats with the extra 714 Energy requirement?

MrRowey (changes requested):

units/XAB2307/XAB2307_unit.bp
lL1l1: > reset it back to the 10k & 15k.

> That also can't be done because the 0.15 -> 0.2 muzzle charge delay change reduces the time the weapon has to charge by 1 tick, so it has to be changed to 15000 cost and 10714 drain.

Also it is 714 energy drain, not requirement. It will still use 15000 energy per shot.

lL1l1: Same with the 25000 energy per shot and 17857 energy drain. The drain is only increased so that the charging can actually keep up with the reload time, and it has no relevant balance consequences outside of needing 2 energy storages to fill in between charges.

MrRowey (approved):

PR #5850, opened 2024-01-22
Allow custom AI descriptions and ratings

maudlin27:
Adds in logic to assign campaign AI personalities to AIs, and to change AI ratings from 0.
For now setup to work with M28AI, with comments/details for if other AIs wished to be added.
Has to be done via the lobby.lua rather than via the AI mod itself, as the AI mod does not load at this point, and hence it is not possible for the AI mod to change these values (and for the changes to then be reflected properly).

In-game example (with 4z0t's scorecard):
image

(note: For some reason when creating this pull request 2 other commits relating to a Sera category fix and merging a branch show, but the changed file only contains the file that this PR relates to, the "Allow custom AI descriptions and ratings"; let me know if I've overlooked something though)

Currently this is setup to apply to M28 (and M27 for ratings), but I'm happy to expand to cover other AIs if desired (and subject to suggestions from the other AI devs on what ratings to apply to their AI).


maudlin27:
Also to note that in the case of ratings, ideally these would vary based on the map size (5km is generally easier to beat than 20km due to the apm benefits AI have), but while in theory it's possible to import the specified map (using import(gameInfo.GameOptions.ScenarioFile).ScenarioInfo.size), map scripts often cause errors when importing, and I'm unsure how to resolve them.
For example, importing the Theta passage FAF version causes an error because it doesnt recognise string()
Unless there's a simple way of having it recognise all standard lua functions/operators (to cover maps that might use other similar functions/operators) then I don't think it'd be worth the risk to try and add this feature.

maudlin27:
Thanks to Hdt80bro for highlighting the part of the code that loads map info - I've incorporated this so the AI is now able to factor in if it is a skirmish map (when picking campaign style names), and adjust the rating based on the map size.
Also moved the rating aspect to a separate function to improve readability, particularly for if other AI are added to it

relent0r:
This is very cool. I'm still of the opinion that the iBaseRating should come from the AI mods aitypes data. 

Obviously I don't know jack about the lobby.lua but I can see that the aitypes data is loaded into it at the top level so a base rating defined in that should be available for use. This would remove the requirement to only support specific AI's with this method and allow the AI developer to change the rating over time as they improve (or degrade) the AI's capabilities. 

The result would be a robust rating system for AI that uses all the goodness that you've got here and removes the administrative overhead of needing to update it based on changes to 3rd party AI's. It would also allow someone like Maudlin to have m27 at one rating and then be able to change m28 to a different rating on his next release because he made some great improvement without having to wait for an official faf release.

Garanas:
Closed in favor of #5855

PR #5851, opened 2024-01-23
Revert "UEF ACU: Revert the option to have the teleporter and nuke upgrades simultaneously"

lL1l1:
Reverts FAForever/fa#5783
Now that the Billy is fixed and rebalanced for teleport, this can be reverted.


MrRowey:
@Garanas This was something the balance team decided on that they wanted at the time

MrRowey (approved):

Garanas:
@MrRowey and I think it is a really bad idea 😃

MrRowey:
Let's Leave it on develop for now and with the issues being fixed see how it goes it it still completely a bad thing then we can revert it.

Garanas:
@MrRowey it will now be part of the hotfix 😃 , to fix an issue with the Billy ironically.

PR #5852, opened 2024-01-23
Fix decimal places not showing up for the Damage stat of the Shard

Basilisk3:
The Shard deals 7.5 damage, but the in-game UI only shows 7, this PR fixes this.

### DIFF
--- a/loc/US/strings_db.lua
+++ b/loc/US/strings_db.lua
@@ -6754,7 +6754,7 @@ uvd_0011="Speed: %0.1f, Turning: %0.1f"
 uvd_0012="Speed: %0.1f, Reverse: %0.1f, Acceleration: %0.1f, Turning: %d"
 uvd_0013="Vision: %d, Underwater Vision: %d, Regen: %0.1f, Cap Cost: %0.1f"
 uvd_0014="Damage: %d - %d, Splash: %d - %d"
-uvd_0015="Damage: %d x%d, Splash: %.3g"
+uvd_0015="Damage: %.3g x%d, Splash: %.3g"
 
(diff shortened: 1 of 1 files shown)


loc/US/strings_db.lua
```diff
@@ -6754,7 +6754,7 @@ uvd_0011="Speed: %0.1f, Turning: %0.1f"
 uvd_0012="Speed: %0.1f, Reverse: %0.1f, Acceleration: %0.1f, Turning: %d"
 uvd_0013="Vision: %d, Underwater Vision: %d, Regen: %0.1f, Cap Cost: %0.1f"
 uvd_0014="Damage: %d - %d, Splash: %d - %d"
-uvd_0015="Damage: %d x%d, Splash: %.3g"
+uvd_0015="Damage: %.3g x%d, Splash: %.3g"
```
lL1l1, line 6757:
```suggestion
uvd_0015="Damage: %.8g x%d, Splash: %.3g"
```
Higher precision is needed for high damages such as Yolona to not show as exponents.

lL1l1 (changes requested):
Make sure to change the formatting in the other languages, the formatting for `uvd_0010` (the single shot damage display), and the default formatting in `unitviewdetail.lua`.

loc/US/strings_db.lua
Basilisk3: The Scathis appears to require this, so this is still good to have.

Basilisk3:
There are no translations for uvd_0009 and beyond, I think adding those is probably a change for a separate PR. If there is no translation, it just uses the english one.

PR #5854, opened 2024-01-24
More Robust Auto OC toggle

4z0t:
Make AutoOC as units stat, that doesn't rely on Sync table magic. ~~Can't actually tell if this one will fix auto oc not working, at least it must show it is turned off.~~ Fixed Auto OC not working too.


4z0t:
Turns out it is not UI issue. I'm investigating why Auto OC didn't work there.

4z0t:
How to reproduce the bug and what bug this is: sometimes UI shows that auto OC is on, but meanwhile ACU is not using it. This is *very* annoying bug that once decided my game. To reproduce it, enable auto OC and then start upgrade, while upgarde let ACU fire some targets. After upgrade you will find out that auto OC is gone.

Issue #5420, opened 2023-09-10
Balance idea: Telemazor


Blodir:
## Intended changes

Teleport (all)
Max range: 350
Increase teleport fx radius by 25%

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

Garanas:
And what about the LOD of the effects? Increasing that makes it visible sooner when zoomed out

Blodir:
Keep the LOD levels as they are for now, as there's no agreement about raising them.

PR #5475, opened 2023-09-27
Variable teleport cost

clyfordv:
Externalizes the teleport cost function, allows variable cost by distance. Includes a UI reticle for displaying variable cost on cursor (reticle now externalized and generalized).

Fully backwards compatible, new algorithm is opt-in on a per-unit basis with a boolean "UseVariableTeleportCosts" blueprint field.

In-depth explanation and photos here: https://forum.faforever.com/topic/6549/tech-demo-variable-teleport-costs?_=1695828399713

PR #5856, opened 2024-01-26
Add 350 teleport range limit

lL1l1:
Partially completes #5420 (needs +25% teleport fx radius).

Draft because I need help adding a teleport range overlay. It seems like it requires a binary patch because it relies on enums.
Aside from the overlay, it works as intended.

If we add a new overlay/weapon range category, I think it would also be useful to add some others:
- Separate Tactical Missile, T2 artillery, T3 artillery, T4 artillery + nukes, and secondary weapons overlays. They currently all use indirect fire, which can obscure your actual indirect fire units if you want to use the permanent overlay option or select a backpack TML unit.
- Separate SMD, shield, and TMD overlays.


Garanas:
I would personally rather go with https://github.com/FAForever/fa/pull/5475. That way the range is infinite, but large jumps simply consume significant amounts of energy.

Garanas:
> Separate SMD, shield, and TMD overlays.

This is an interesting idea, if we can make this work we should definitely do that 😃

lua/ui/game/rangeoverlayparams.lua
```diff
@@ -135,6 +135,18 @@ RangeOverlayParams = {
...
+        Label = '<LOC range_0014>Teleport Range',
+        Categories = categories.OVERLAYTELEPORT,
+        NormalColor = glowAllNormal..'7b61d8',
+        SelectColor = glowAllSelect..'8c75d0',
+        RolloverColor = glowAllOver..'9c86e9',
+        Inner = innerMilitary,
+        Outer = outerMilitary,
+        Type = 3,
+        Tooltip = "overlay_teleport",
+    },
```
4z0t, line 149: Game engine wont process this one.

4z0t:
> > Separate SMD, shield, and TMD overlays.
> 
> This is an interesting idea, if we can make this work we should definitely do that 😃 

Engine patch is required. Pretty complicated one.

MrRowey:
Going to close this in infavor of #5475 being implemented

VaelophisNyx:
should've gone with this one, tbh. Cost is no object. Range matters more and is actually counterable outside of dumb tactics like T2 PD walls around every single building.

VaelophisNyx:
additionally the engine change attached to this would've been a boon to modders who have figured out how to limit teleport range, but have to "show" it with a hacky fake stealth field range indicator

Garanas:
This particular assembly patch is not trivial to make. This was by far an easier, and more natural solution to the problem 😃

VaelophisNyx:
variable costs can't scale high enough to matter without making teleporation borderline useless; where a capped range would mean you can't just cross-map blink into the enemy base. It adds risk that minimal economy pressure does not.

VaelophisNyx:
Still should've gone with this. It was worth it, and is a highly in-demand feature for years now, even as far back as 2015. Variable cost has, as expected, not caused any change to teleporting strategies so far as I have seen.

Garanas:
It's been less than a day - we'll have to see what happens the next (few) week(s). Always welcome to post replays and everything 😃

PR #5857, opened 2024-01-26
Revert "Revert "UEF ACU: Revert the option to have the teleporter and nuke upgrades simultaneously""

Garanas:
Reverts FAForever/fa#5851

I'm veto'ing these changes as game lead. I cannot fathom how this change is interesting or competitively healthy. All I can see from it are meme-games where one side loses significant assets because they forgot to build TMD _all around their base_, instead of just the direction where you can expect the enemy. It will be completely counter intuitive to the average player.

The difference with the Cybran Telemazor is that with the Cybran Telemazor your ACU is suddenly the center of attention and the chances that you'll die is significant. The UEF Telebilly on the other hand would likely already have teleported out before the Billy even reaches its destination. 

Another difference with the Telemazor is that for the Telebilly you can evaluate whether it will succeed before teleporting. After all - TMDs are structures and therefore one scout will give you all the intel you need, regardless whether you have radar coverage. The defending party can't move the TMD around like one can with units. The same applies to shields that can meaningfully stop the Billy. With Telebilly you can teleport in exactly where you need to be, launch the Billy and teleport out again with guaranteed success and essentially guaranteed safety.


Garanas:
Both this branch and #5851 were pointing to `develop` instead of `deploy/fafdevelop` 😃

Basilisk3:
I completely agree with this decision. Removing the option of T3 + Billy actually made this upgrade significantly worse, in exchange for a meme strategy which was clearly not healthy for the game.

ComradeStryker:
:D

VaelophisNyx:
Consider: Teleporting (S)ACUs in general suck to deal with and shouldn't really be doable!

PR #5860, opened 2024-01-27
Cycle through templates based on highlighted command

Garanas:
With thanks to https://github.com/FAForever/FA-Binary-Patches/pull/65

https://github.com/FAForever/fa/assets/15778155/37e63128-74f5-4e23-8fe1-b98d90846cad


Garanas:
Some duplicated templates are removed.

PR #5862, opened 2024-01-27
Introduce additional lobby guards

Garanas:
An attempt to combat the following situation where a lobby is crashed by embedding (a large amount of) characters into afield that is supposed to be used for the number of games:

image


Garanas:
@4z0t any other sanity checks that we can/should do on the player data that you can think of?

4z0t:
Maybe match name with regex, but I think it is too excessive.

Garanas:
@Sheikah45 are there character limits for the client? We can make a regular expression to match with that

PR #5864, opened 2024-01-29
Fix the status of the minimap's pin feature not being saved into the game.prefs file

Basilisk3:
Before this PR, you had to re-pin the minimap every game, as its pin-status would not be saved into your game.prefs file.

### DIFF
--- a/lua/ui/game/minimap.lua
+++ b/lua/ui/game/minimap.lua
@@ -131,12 +131,19 @@ function CreateMinimap(parent)
         end
         frameCount = frameCount + 1
     end
-    controls.displayGroup.OnPinCheck = function(control, checked)
-        control:SetSizeLock(checked)
-        control:SetPositionLock(checked)
-        -- add save pin in prefs
-    end
+	controls.displayGroup.OnPinCheck = function(control, checked)
+		control:SetSizeLock(checked)
+		control:SetPositionLock(checked)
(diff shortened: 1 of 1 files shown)


lua/ui/game/minimap.lua
```diff
@@ -131,13 +131,20 @@ function CreateMinimap(parent)
...
-    controls.displayGroup.OnPinCheck = function(control, checked)
-        control:SetSizeLock(checked)
-        control:SetPositionLock(checked)
-        -- add save pin in prefs
-    end
+	controls.displayGroup.OnPinCheck = function(control, checked)
+		control:SetSizeLock(checked)
+		control:SetPositionLock(checked)
+		Prefs.SetToCurrentProfile('minimapPin', checked)
+		Prefs.SavePreferences()
```
Garanas, line 138: Please do not save the preferences to disk. It is not required - the game does that when it exits. 

Because of this pull request I took a closer look at the preference file. And it was wild:

- https://github.com/FAForever/fa/pull/5871

Garanas (changes requested):
Please do not save the preferences to disk manually.

lua/ui/game/minimap.lua
Basilisk3: Thanks for the help, and great to see my PR lead to this being discovered after all those years👍

PR #5868, opened 2024-01-31
Fix the Salvation's projectiles falling short of their target

Basilisk3:
This PR fixes the Salvation consistently undershooting. Best tested by removing its firing randomness and then aiming at a static target, like an unclaimed mass point. When leaving its firing randomness unchanged (like it is in-game), it does not undershoot every single time, but there is a clear pattern of its projectiles falling short of their target.

The balance impact of this is nearly zero, because when changing the MuzzleVelocity, the trajectory of the projectile is changed to compensate. 

**Changes:**
MuzzleVelocity: 150 --> 180

### DIFF
--- a/units/XAB2307/XAB2307_unit.bp
+++ b/units/XAB2307/XAB2307_unit.bp
@@ -177,7 +177,7 @@ UnitBlueprint{
             MuzzleChargeDelay = 0.15,
             MuzzleSalvoDelay = 0,
             MuzzleSalvoSize = 1,
-            MuzzleVelocity = 150,
+            MuzzleVelocity = 180,
             MuzzleVelocityReduceDistance = 4000,
             ProjectileId = "/projectiles/AIFFragmentationSensorShell01/AIFFragmentationSensorShell01_proj.bp",
             ProjectileLifetime = 150,


MrRowey:
Please See #5849 for planed Changes

Basilisk3:
@MrRowey At the very least, provide me a concrete reason for closing this and not a one-liner, please. This PR deals with a real issue that has been in the game for years, and it seems to me like it has not even been tested out in-game.

Are you aware that the MuzzleVelocity stat also changes the pitch of the Salvation's turret and in return the flight path of its shells? Why is a flatter flight path with a lower MuzzleVelocity something that is wanted? A low MuzzleVelocity/flight path makes units undershoot, so #5849 would make the issue worse and the Salvation would undershoot even more. This surely cannot be what you want?

Garanas:
I don't think this pull request conflicts with #5849 . Interesting enough however the pull request in question takes the opposite direction:

image

Does the same problem occur in that branch or do those values fix the problem too @Basilisk3 ?

lL1l1:
After #6142 reverting the balance changes, this can be looked at again.

clyfordv:
I have a suspicion: image

clyfordv:
I was close but it's actually this: 
image

Do we want FragmentImpactRadius as a blueprint (balance) value?

MrRowey:
Yes do that

MrRowey:
@clyfordv im reclosing this PR if you can make a new PR with that value available via the BP File

PR #5871, opened 2024-02-03
Tackle performance issues when accessing the preference file

Garanas:
The preference file can be accessed with `GetPreference`, updated with `SetPreference` and written to the disk with `SavePreferences`. These functions have some interesting side effects:

```lua
--- Retrieves a value in the memory-stored preference file. The value retrieved is a deep copy of what resides in the actual 
--- preference file. Therefore this function can be expensive to use directly - if you're not careful you may be allocating kilobytes worth of data!
---
--- You're encouraged to use `/lua/user/prefs.lua` to interact with the preference file.
---@param string string
---@param default any?
---@return any
function GetPreference(string, default)
end
```

```lua
--- Updates a value in the memory-stored preference file.
---
--- You're encouraged to use `/lua/user/prefs.lua` to interact with the preference file.
---@param key string
---@param obj any
function SetPreference(key, obj)
end
```

```lua
--- Writes the preferences to disk to make it persistent. This is an expensive operation. The 
--- game does this automatically when it exits, there should be no reason to call this manually.
---
--- You're encouraged to use `/lua/user/prefs.lua` to interact with the preference file.
function SavePreferences()
end
```

Todo:

- [x] Unbreak the 'all faction templates' from GazUI

```
WARNING: Error running lua command: c:\users\4z0t\git\fa\lua\keymap\hotbuild.lua(462): bad argument #1 to `gsub' (string expected, got boolean)
         stack traceback:
             [C]: in function `gsub'
             c:\users\4z0t\git\fa\lua\keymap\hotbuild.lua(462): in function `ConvertID'
             c:\users\4z0t\git\fa\lua\keymap\hotbuild.lua(474): in function <c:\users\4z0t\git\fa\lua\keymap\hotbuild.lua:432>
             c:\users\4z0t\git\fa\lua\keymap\hotbuild.lua(779): in function <c:\users\4z0t\git\fa\lua\keymap\hotbuild.lua:771>
             (tail call): ?
             c:\users\4z0t\git\fa\lua\keymap\hotbuild.lua(339): in function `buildAction'
             [string "import("/lua/keymap/hotbuild.lua").buildAct..."](1): in main chunk
             ```

### DIFF
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -366,7 +366,10 @@ end
 function GetOptions(key)
 end
 
+--- Retrieves a value in the memory-stored preference file. The value trieved is a deep copy of what resides in the actual 
+--- preference file. Therefore this function can be expensive to use directly.
 ---
+--- You're encouraged to use `/lua/user/prefs.lua` to interact with the preference file.
 ---@param string string
 ---@param default any?
 ---@return any
@@ -846,7 +849,10 @@ end
 function RestartSession()
 end
 
+--- Writes the preferences to disk to make it persistent. This is an expensive operation. The 
+--- game does this automatically when it exits, there should be no reason to call this manually.
 ---
+--- You're encouraged to use `/lua/user/prefs.lua` to interact with the preference file.
 function SavePreferences()
 end
 
@@ -1005,7 +1011,9 @@ end
 function SetPausedOfUnit(unit, pause)
 end
 
+--- Updates a value in the memory-stored preference file.
 ---
+--- You're encouraged to use `/lua/user/prefs.lua` to interact with the preference file.
 ---@param key string
 ---@param obj any
 function SetPreference(key, obj)
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -366,8 +366,9 @@ end
 function GetOptions(key)
 end
 
---- Retrieves a value in the memory-stored preference file. The value trieved is a deep copy of what resides in the actual 
---- preference file. Therefore this function can be expensive to use directly.
+--- Retrieves a value in the memory-stored preference file. The value retrieved is a deep copy of what resides in the actual 
+--- preference file. Therefore this function can be expensive to use directly - if you're not careful you may be allocating 
+--- kilobytes worth of data!
 ---
 --- You're encouraged to use `/lua/user/prefs.lua` to interact with the preference file.
 ---@param string string
@@ -1011,7 +1012,7 @@ end
 function SetPausedOfUnit(unit, pause)
 end
 
---- Updates a value in the memory-stored preference file.
+--- Updates a value in the preference file. Updating the preference file on disk is delayed until the application exits.
 ---
 --- You're encouraged to use `/lua/user/prefs.lua` to interact with the preference file.
 ---@param key string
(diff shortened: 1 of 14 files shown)

### CODE REVIEW

**Comment by Garanas** [comment]
The old implementation has some interesting consequences. As an example: if a user creates build templates then it can quickly consume your entire preference file. This would be fine if the table would be allocated once and then simply re-used. But that is not the case... in fact, it was re-created all over! When UI code retrieves a template it would call this function:

https://github.com/FAForever/fa/blob/d0d62d8adf0c9bf86c6143c9acf4c2cb428c6e0e/lua/ui/game/build_templates.lua#L69-L71

In the old implementation it would then use `GetPreference` to get the user template data. This would be fine if it was retrieving a constant value. But in reality it is retrieving a table that looks like this:

```lua
build_templates = {
    
    -- (...) other templates

    {
        templateData = {
            32,
            32,
            {
                'ueb0102',
                16257,
                0,
                0
            },
            {
                'ueb0102',
                16323,
                8,
                8
            },
            {
                'ueb0102',
                16477,
                16,
                16
            },
            {
                'ueb0102',
                16491,
                24,
                24
            },
            {
                'ueb0102',
                16543,
                16,
                0
            },
            {
                'ueb0102',
                16597,
                24,
                8
            },
            {
                'ueb0102',
                16611,
                0,
                16
            },
            {
                'ueb0102',
                16663,
                8,
                24
            },
            {
                false,
                16680,
                0,
                8
            },
            {
                'ueb1301',
                16835,
                8,
                0
            },
            {
                'ueb1301',
                16975,
                24,
                0
            },
            {
                'ueb1301',
                17156,
                24,
                16
            },
            {
                'ueb1301',
                17322,
                16,
                24
            },
            {
                'ueb1301',
                17517,
                0,
                24
            },
            {
                'ueb4202',
                17878,
                17,
                7
            },
            {
                'ueb1101',
                17997,
                11,
                19
            },
            {
                'ueb1101',
                18033,
                11,
                17
            },
            {
                'ueb1101',
                18044,
                11,
                15
            },
            {
                'ueb1101',
                18051,
                11,
                13
            },
            {
                'ueb1101',
                18074,
                9,
                13
            },
            {
                'ueb1101',
                18099,
                7,
                13
            },
            {
                'ueb1101',
                18205,
                13,
                5
            },
            {
                'ueb1101',
                18241,
                13,
                7
            },
            {
                'ueb1101',
                18253,
                13,
                9
            },
            {
                'ueb1101',
                18290,
                13,
                11
            },
            {
                'ueb1101',
                18306,
                17,
                11
            },
            {
                'ueb1101',
                18313,
                19,
                11
            }
        },
        name = 'Air Factory',
        icon = 'ueb0102'
    },

    -- (...) other templates

},
```

This is a single entry with _just_ 27 structures in it. It is a little over a kilobyte worth of data. Considering that some people even have entire wall-based paintings in their templates this can easily represent up to dozens, if not hundreds of kilobytes worth of data!

And in the old implementation all of this data would be **unique data**, recreated each time the user build templates are retrieved! As a few humble examples:

- (1) When the selection of a user changes

This is a subtle one, because when a unit in the selection dies then the selection changes too. During large battles, this would trigger constantly!

- (2) When the selection of a user changes and the 'All faction templates' setting is enabled

Note that this is _on top of_ (1). We'd allocate two instances of the user templates table when this option is enabled.

- (3) When the user is cycling through templates with Hotbuild

Each cycle would recreate the user templates table from scratch!

Garanas:
_reserved for future comment_

Garanas:
@4z0t / @gordenwunderlich would either if you be available to test the hotbuild alternative implementation for all-faction templates?

And yes - this is not 'do not repeat yourself' proof, but the goal of this PR is not to fix the code style issue. It is to fix a gigantic garbage collection problem.

PR #5874, opened 2024-02-04
Various navy tweaks to make the Tech 2 stage more interesting

Basilisk3:
This PR aims to fine-tune the previous balance changes to the Tech 1 and Tech 2 naval stages. The overall direction we are moving into is good, so these changes are relatively minor.

### <ins>Uashavoh </ins>
**AntiTorpedo RateOfFire: 0.26 --> 0.3**

This unit suffered due to the buffs to torpedoes (automatic retargeting) and torpedo defenses. This slight buff aims to bring the unit's performance roughly back in line with how it was before the these changes.

### <ins>Cooper </ins>
**Health: 1750 --> 2000**

The Cooper is a unit that is usually used relatively close to the frontlines. Unfortunately, its hit points are lower than even that of the weakest frigates, while costing around three times as much. This change gives it slightly more survivability, while still emphasizing the need to micro it properly and keep it behind your destroyers.

### <ins>Salem Class
**BackUpDistance: 5 --> 10**


lL1l1:
> DamageRadius: 1.0 --> 1.4

Important to note that this will allow groundfire to hit submarines.

Basilisk3:
That is intended, do you think that it would be too strong? I liked the idea of it not being almost completely defenseless without Coopers, if you have the time to micro, that is. I am still split on including this. Maybe something like 1.2 would make more sense, though then it would not be a very meaningful change.

lL1l1:
In terms of balance, subs are very fast and can easily dodge groundfire, but on the other hand our only experience is with the lower fire rate, lower DPS battleships.
It would definitely be controversial on the user experience part of things because you'll get the same non-obvious skill as battleship vs sub groundfire (I've seen multiple complaints about it from newer players) but on the more prevalent and more important UEF destroyer stage.

Up to balance team to decide how they want to deal with all that though, I only wanted to point out that major detail because you didn't mention it in the opening post, and instead phrased it as a mostly aesthetic change.

Basilisk3:
The Exodus can groundfire subs as well, but with the very high torpedo damage it has, it probably is not required. I am going to revert the change and investiage its torpedoes.

Basilisk3:
I think this is ready for review now.

MrRowey:
Asked team to discuss this

Basilisk3:
@MrRowey I kept the change to the Siren's `BackUpDistance` in place, because it would not make sense if the Salem has the increased distance and the Siren doesn't. I applied the rest of the feedback.

MrRowey:
> @MrRowey I kept the change to the Siren's `BackUpDistance` in place, because it would not make sense if the Salem has the increased distance and the Siren doesn't. I applied the rest of the feedback.

whats the backupdistance of all the other factions cruiers ?

Basilisk3:
> What is the `backupdistance` of all the other factions cruisers? For Salem it is different due to being able to keep the guns in the right direction shooting without it turning around and not shooting. for the cybran cruiser this is less important as its main role is air so if the backupdistance is different for all the other cruisers then keep the cybran one unchanged

The `BackUpDistance` stat is not very relevant for the other cruisers, but it is for the Siren, so it would make sense to increase it in my opinion. Especially because the Salem's and Siren's cannons function very similarly. I will ask in the Discord thread and see what the others think. Cruisers currently have their `BackUpDistance` set to `5`.

MrRowey (approved):

PR #5882, opened 2024-02-07
Fix Chrono Dampener visual effects not scaling with gun range

lL1l1:
Solution feels a bit hard-coded, but it's the simplest.


lL1l1:
Will go with a different solution.

Issue #1271, opened 2016-05-15
bug chrono dampner


shalkya:
game is 4755438

https://gist.github.com/shalkya/e318cc53baa8c23b5b4bdcf85a6664c8

it is running on the balance patch, but we didn't change anything to chrono dampner.
1st chrono fired, then no more.

aeoncleanse:
Hmmm...  I tested chrono in 3654 and with blackops, working fine both times...

shalkya:
yes i did sandbox too and couldn't replicate..
but i still wanted to post the bug.

shalkya:
OK i found the bug. OC messed up with the chrono dampner.
(reset of count down at each OC ?)

https://gist.github.com/shalkya/3b80265725db390bf8084a26de7cf353
4755513

test done in normal faf.

aeoncleanse:
I'll look closer another time, but I do know that OC temporarily disables all other weapons, which would indeed disrupt chrono fire.

aeoncleanse:
Hmm. Upon inspection, OC only disables the primary gun...

aeoncleanse:
Logging confirms that OC messes with the chrono sequence for some reason :/

PR #5883, opened 2024-02-08
Re-implement Chrono Dampener to properly use firing cycle + VFX fixes

### LINKED ISSUE #2410: need to fire chrono dampner on land to fire it afterward in water
can't be really more precise... i will try to show you different example : 
1st : be in water, get opponent land/navy around you =>chrono dampner doesn't work.
2nd : be on the ground, fire chrono dampner, get in water, chrono dampner keep on working.
3rd : lose the range on unit (land or navy) while being water, get back in range, still in water => chrono dampner doesn't work.

i guess it is triggered by the fact that it can't fire his primary weapon under water or something like that?
[game.uid.7630337.log](https://github.com/FAForever/fa/files/1814935/game.uid.7630337.log)
content.faforever.com/faf/vault/replay_vault/replay.php?id=7630337

lL1l1:
Fixes #2410.
To replicate the issue, start chrono with targeting of a land/sea surface unit, then walk underwater while staying within its 35 range. It will stay locked on and never check its own inability to fire underwater, so it'll keep firing underwater.
This is caused by its custom firing cycle that never goes to the OnFire event except in the initial target acquisition. The cycle:
`idle -> FireReady -(FireReady.OnFire)-> FiringState (repeat forever) -(FiringState.OnLostTarget)-> Idle`
### Changes
- RateOfFire actually determines the fire rate now.
- It can no longer fire underwater.
- Categories to stun are generated from the BP's stun buff definition instead of being coded in the weapon.
- Chrono uses its own range instead of relying completely on the primary weapon. Now the range upgrades adjust the range of Chrono, like they do with overcharge.
- The initial stun effects are fixed (target.InitialStunFxApplied was never reset) and the effect was changed to be a more noticeable flash, as if the enemy was suddenly hit by a burst of energy. The flash effect also reduces with distance since the stun is weaker far away.
- If you zoomed out or looked away while Chrono was firing, the large ring particle wouldn't emit and you'd never see the effect if you looked back at it. This is changed to always emit because Chrono is rather important.
- `AIR*MOBILE` is not a valid target restriction, so Chrono was uselessly firing at aircraft. The target priorities are fixed to only fire at t1-t3 mobile surface units.


Garanas:
The chrono dampener intentionally did not use the firing cycle. It was like this before I refactored it in https://github.com/FAForever/fa/pull/3918. I discussed this back then too and it would prevent multiple Aeon ACUs from permanently stunning enemy units because they have control over when the chrono dampener triggers.

lL1l1:
I prefer that it does use the fire cycle since tying it to the game tick has the potential to delay the activation for 4 seconds, making the weapon seem unresponsive and sluggish sometimes.
I added stunlock prevention by preventing new chrono stuns applying before the initial stunning chrono has reloaded. This allows it to be responsive when you walk into range, but it also harms multi-chrono strategies since one chrono far away from a unit can steal the ability to stun that unit from a chrono much closer, if the timings line up. But to be fair, seeing more than one chrono in a single fight is very rare.

Garanas:
I'll look into it for you this weekend 👍

lL1l1:
@MrRowey It should be a very minor change in terms of balance, so I didn't add do not merge. Doesn't change how often Chrono fires or how long it can stun, just when it fires.
https://github.com/FAForever/fa/blob/3096dffb7ecc2c5afaf14fbca6301360af669902/changelog/snippets/balance.5883.md?plain=1#L1-L5

If Aeon SACU get chrono with a longer cooldown, this PR would make creating that new Chrono weapon much easier.

MrRowey:
thats fine, Can i get a ping when merged so I can add to patchnotes

MrRowey:
@Garanas can you review this, please

MrRowey (approved):

MrRowey:
@lL1l1 can you fix the conflict and can we get this in before the new game patch

lL1l1:
Please make sure to test that overcharge doesn't mess with the firing sequence: #1271 
I only tested that the visuals work and that multiple chronos stunning the same unit didn't stunlock the unit and that the closest chrono took priority in stunning (since it stuns the longest).

lL1l1:
Tested with a bunch of Percies on Setons mid (has underwater areas near land), NoDamage enabled, and adv gun range vs no gun range upgrades.
- No errors
- Overcharge doesn't do anything to chrono
- Multiple ACU can't stunlock, and they don't interrupt eachother, as much as I could see, the closest ACU gets to stun the unit so stun time is maximized.
- Gun range upgrades scale Chrono's effects correctly.

Issue #5886, opened 2024-02-08
You can gain veterancy by destroying your own units


Basilisk3:
I have never seen this abused in-game, but it does not make sense either.

PR #5888, opened 2024-02-09
Fix/veterancy from allied units

Garanas:
Closes https://github.com/FAForever/fa/issues/5886

Veterancy is now only dispersed across units that are considered hostile to the instigator.


Garanas:
Woops, this also includes https://github.com/FAForever/fa/pull/5887

PR #5898, opened 2024-02-11
Rework the Solace and the projectile count of all Torpedo Bombers

Basilisk3:
**Changes:**

**Solace:**
Damage: 500 --> 400
RateOfFire: 0.08 --> 0.1 (required, so the unit does not drop its torpedoes behind or on top of ships after tight turns)
MaxRadius: 32 --> 26 (decreases the likelihood of the torpedoes impacting on terrain)
MuzzleSalvoSize: 4 --> 5
MuzzleSalvoDelay: 0.4 --> 0.2
BreakOffDistance: 42 --> 44 (decreases its effective DPS, to make up for the unit functioning properly now)

The unit does not drop behind or on top of ships anymore. The scenario, where it drops most of its torpedoes on land, should also be reduced with these changes.

**Uosioz:**
Projectile count: 15 --> 6

**Skimmer; Cormorant; Stork:**
Projectile count: 5 --> 4


Garanas:
This was intentional because the projectiles of the Solace split as they enter the water.

Basilisk3:
I am aware, this is mostly because I think this unit requires a similar treatment as the other Torpedo Bombers got in #5632👍. In fact, the changes in this PR might not even be enough.

The math behind it is as following (The Solace is roughly equivalent to 6 Tech 2 Torpedo bombers):

Number of projectiles for Tech 2 Torps (pre-patch): 6 Torps * 3 projectiles = 18 projectiles total
Number of projectiles for Tech 2 Torps (post-patch): 6 Torps * 5 projectiles = 30 projectiles total
Number of projectiles for the Solace: 1 Torp * 4 projectiles * 2 split = 8 projectiles

So for roughly the same mass spent, you get 30 projectiles for the Tech 2 Torpedo Bombers, but only 8 for the Solace. So this has been a significant indirect nerf to the Solace. The Solace has suffered quite a bit from the torpedo defense rework in general, this is another change which attempts to bring some of the unwanted effects on gameplay back into line.

Garanas:
I understand your reasoning. 

From a performance perspective, maybe it's better to reduce the salvo size for tech 2 torpedo bombers from 5 to 4 (and to 2 for Seraphim, because they split into 3 or 4 torpedo's). That way we level it out a bit in favor of the Solace, but still take into account the limitations of the game.

Basilisk3:
I agree, I will add it in this PR.

Garanas:
I think the PR is fine with this, I'll review it tomorrow!

Basilisk3:
Great👍

Garanas:
The Cybran torpedo bomber deals 10 less damage than the other tech 2 bombers. How would we fix that? 😃

Garanas:
I was toying around with the split code related to Seraphim torpedo's, we could for example have the following:

https://github.com/FAForever/fa/assets/15778155/041899a6-6203-4875-b288-d8483d6ea395

https://github.com/FAForever/fa/assets/15778155/a628e563-0b2d-4e64-bc6d-36603949014d

Garanas:
@Basilisk3 the last thing is to fix the damage of the Cybran torpedo bomber

Basilisk3:
> I was toying around with the split code related to Seraphim torpedo's, we could for example have the following:

👍

> @Basilisk3 the last thing is to fix the damage of the Cybran torpedo bomber

That is fixed now.

Garanas:
Great work @Basilisk3

Basilisk3:
Thanks, the Solace should behave much more consistently now.

PR #5899, opened 2024-02-12
Remove the dive button for units that do no not need it

Basilisk3:
A lot of submarines do not have a deck-gun, which makes surfacing them detrimental and a noob-trap. This PR removes the dive button for these units. Note, that Strategic Missile Submarines can launch missiles while submerged, surfacing them is not necessary.

### DIFF
--- a/lua/sim/units/SubUnit.lua
+++ b/lua/sim/units/SubUnit.lua
@@ -1,4 +1,3 @@
-
 local EffectTemplate = import("/lua/effecttemplates.lua")
 
 local Entity = import("/lua/sim/entity.lua").Entity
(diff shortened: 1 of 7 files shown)


Garanas:
@relent0r / @maudlin27 would this cause problems for AIs?

relent0r:
> @relent0r / @maudlin27 would this cause problems for AIs?

No impact from an AI perspective since they are just using TestCommandCaps('RULEUCC_Dive') to check if they need to dive.

No issue from an exiting factory perspective I assume?


Hurts my submarine captain role playing tho.

Garanas:
@Basilisk3 there's a bug. When you have other units in your selection that have the toggle to dive, the submarine that does not have the toggle can still dive/undive. As a result the units are stuck on the surface until your selection matches up again.

https://github.com/FAForever/fa/assets/15778155/40c39747-3ab8-49ef-9e62-3fdf729d610b

Basilisk3:
I will look into it.

Basilisk3:
I am pretty sure it is because of this line:

`MotionType = "RULEUMT_SurfacingSub",`

@Garanas Do you know where these are defined?

Garanas:
Then we'll revert this. It was a nice idea 😃

PR #5900, opened 2024-02-13
Improve Teleport Reticle Cost Display

lL1l1:
- Icons instead of text
- Display Energy cost/second
- Round time to 0.1 (0.01 does not fit in game ticks)

Before <br>  |After <br>   
|:--:|:--:|

### DIFF
--- a/lua/shared/teleport.lua
+++ b/lua/shared/teleport.lua
@@ -43,14 +43,14 @@ TeleportCostFunction = function(unit, location)
     local pos = unit:GetPosition()
 
     -- or, if queuing commands, use the position of the last teleport/move command
-    if IsKeyDown('Shift') then
-        local queue = unit:GetCommandQueue() --[[@as (UICommandInfo[])]]
-        if table.getn(queue) > 0 then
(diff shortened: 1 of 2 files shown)


lua/shared/teleport.lua
```diff
@@ -42,15 +42,17 @@ TeleportCostFunction = function(unit, location)
     -- use unit position by default
     local pos = unit:GetPosition()
 
-    -- or use the position of the last teleport/move command
-    local queue = unit:GetCommandQueue() --[[@as (UICommandInfo[])]]
-    if table.getn(queue) > 0 then
-        for k = 1, table.getn(queue) do
-            local command = queue[k]
+    -- or, if queuing commands, use the position of the last teleport/move command
+    if IsKeyDown('Shift') then
```
Garanas, line 46: This function is not available in the sim

lL1l1:
New color is the same as energy stored/storage color from the economy bars in the top left.

Garanas:
Could you add a little bit of padding between the icon and the text? Beyond that, I think this is a great improvement!

PR #5901, opened 2024-02-13
Buff UEF T3 MAA

lL1l1:
## The Issue

Currently the Cougar is the slowest and by far the lowest DPS T3 MAA despite UEF being the second most or most disadvantaged faction vs air.
### T3 MAA Comparison:
Stat|Sera|Cybran|Aeon|UEF
|--|--|-|-|-|
Damage|200x7|100x12|1200|100x12
DPS|275|235|200|176
Mass|720|600|600|600
DPS/Mass|0.382|0.392|0.333|0.293
Speed|3.4|3.6|3.5|3.3

HP is balanced. Turn rate isn't very relevant. Weapon range is somewhat balanced with projectile travel time.
In terms of the weapon effectiveness at retargeting: Sera (Very fast laser) > Cybran (many tracking missiles) > UEF (loses dps due to missing) > Redeemer (single missile)

## Proposed changes
```
Cougar:
Damage: 100x12 -> 100x14
Reload Time: 4.5 -> 4.3
Speed: 3.3 -> 3.5

Redeemer:
Speed: 3.5 -> 3.3
```
## Thoughts on the changes

The reload time reduction cancels out the 0.2 seconds to fire 2 extra shots.
The DPS is brought to a comparable level with other factions (176 -> 206).
The volley increase also allows 3 Cougars to kill a strategic bomber in 1 volley instead of requiring 4 MAAs.
The speed increase allows it to keep up with Parashields (also 3.5 speed), and better assist Titans. 
Swapping the speed with Redeemer maintains a semblance of faction diversity. Redeemer was chosen because it doesn't really need its speed due to asylums being far too quick (4.0) to keep up with, and harbingers being relatively slow at 3.0 speed.


Pokute:
You might want to add a range and time-to-kill (for t3 bomber) and stats to that table. Also alpha damage is really important against t3 bombers.

Ravagers have such long time to kill that most t3 bombers will easily drop bombs almost anywhere within the T3AA clump's range. Also the T3AAs probably grossly overkill their targets and often even get outran.

For example, against a line of 6 cybran T3AA, you can target a single bomb very back from the line successfully. On the flying 3 t3bombers over a line of 6 cybran t3aa loses 2, against 6 uef t3aa loses one.

Another point is that some of these units also shoot ground units, with ok efficiency.

As a summary, just looking at unit stats for those very different units is not enough to make stat changes - we need to look at in-game performance.

lL1l1:
> You might want to add a range and time-to-kill (for t3 bomber) and stats to that table. Also alpha damage is really important against t3 bombers.

Range is already varied among the MAA to compensate for travel time.

Stat|Sera|Cybran|Aeon|UEF
|--|--|-|-|-|
Range|58|62|64|60
Velocity|inf|15-80 (10 accel)|34-60 (6 accel)|100
Reaction time|unpacks, but 0|0|unpacks|1

> Ravagers have such long time to kill that most t3 bombers will easily drop bombs almost anywhere within the T3AA clump's range. Also the T3AAs probably grossly overkill their targets and often even get their projectiles outran.

Did you mean Bouncers instead of Ravagers? I haven't seen their projectiles get timed out too often, and after the new retargeting buffs it can be considered among the best T3 MAA because it deals with T1/T2 aircraft very well while having high DPS and an almost constant fire rate.

> some of these units also shoot ground units, with ok efficiency.

Bouncer is not very efficient vs ground. I think it misses too much and does little damage like the sky slammer. The Lightning Tank has its cost increased to add the ground weapon.

> As a summary, just looking at unit stats for those very different units is not enough to make stat changes - we need to look at in-game performance.

I do not think they are different enough to not allow stat analysis. Ground targeting is unrelated to the AA DPS (as we see with the DPS/mass stats) and projectile behavior is already compensated by range and retargeting ability. Really these are minor issues compared to the large DPS differences.

MrRowey (approved):

PR #5905, opened 2024-02-16
Increase teleport effect scale by 25%

### LINKED ISSUE #5420: Balance idea: Telemazor

lL1l1:
Closes #5420 alongside the variable teleport costs instead of a 350 max range.


Garanas:
What is the idea behind this change, to make it easier to spot?

lL1l1:
Ask @Blodir or the balance team about it.

I also guess it is to make it easier to spot, but I don't really see a huge difference, especially since the effect simply disappears entirely when out of view/zoomed too far out. Personally, I would remove the LOD limits (teleportation is basically a strategic weapon), and slightly increase the size of the cybran effects because they use dark/transparent particles which are harder to see than other factions.
Increasing the size of the UEF effect kinda messes up the visual of the effect covering the height of the ACU, so that's also why I would only increase Cybran's effect.

Garanas:
What if we instead improve the LOD?

MrRowey:
Could you provide the before % after image in the PR, please

lL1l1:
+25% Larger effects:

https://github.com/FAForever/fa/assets/82986251/af9fc616-c6c4-4afc-b84c-01a5ad167654

Current effects:

https://github.com/FAForever/fa/assets/82986251/3ca6d467-275a-4337-96b3-b76d81aa2835

+25% larger on top, current on bottom comparison:

https://github.com/FAForever/fa/assets/82986251/8658d3a0-ce73-4f5e-b403-b830956c02a7

#### Thoughts:
- Cybran's effect is very hard to see because it uses a lot of transparent distortion effects instead of flashy lights.
- Removing the LoD would be quite nice aesthetically since there is a pop-in effect when you zoom in on the effect as seen in the first video. Balance-wise, the non-cybran effects would be even easier to see without LoD, especially with +25% size making them larger than a factory.
- The teleport sound effects are disproportionate (`TeleportChargingAtDestination` in unit bp): 
  - Cybran uses the Monkeylord's laser continuous firing sound effect, so it is very loud and and can be heard easily across the map. One caveat is that the sound effect itself also has a LoD, so you need to zoom in to hear it, but once you zoom in enough to hear it once, it will play until the teleport finishes. You could say this balances out the worse visibility of the teleport effect.
  - Aeon and UEF use their ACU's reclaim sound loop for their teleport sound. Seraphim uses Ahwassa's movement sound loop for its teleport sound. These are way quieter and harder to discern than Cybran's teleport sound.

MrRowey:
It was mentioned about the LOD in the balance chats could you look into what would look better?

MrRowey (approved):

PR #5906, opened 2024-02-16
Allow Ctrl to bypass Alt to force attack move

### LINKED ISSUE #5876: Holding Ctrl + Alt should not show Attack Move cursor
Currently, holding Alt for attack move triggers when holding Ctrl+Alt as well. This is probably not useful to anyone and is a bit unexpected/annoying in my opinion.

Furthermore, some hotkeys kan not be bound with Alt modifier because it interferes with detecting hovered units. If I remember correctly this is the case for the copy orders hotkey for example. When Alt is held it doesn't detect the unit under the attack cursor and no orders are copied.

So this would at least free up Ctrl-Alt for use with those hotkeys. Ideally it should work for only Alt as well.

lL1l1:
Resolves #5876. This allows using Ctrl + Alt combinations for the mouse context hotkeys while having the "alt to force attack move" setting enabled.

Ideally we could get unit rollover info for hotkeys regardless of the selection threshold.


MrRowey (approved):

MrRowey:
@lL1l1  can you fix the confict

lL1l1:
Does open up an avenue to re-adding target firing TMLs to kill all 4 mex storages by holding Ctrl to not be forced into groundfire. "Ctrl" gives you direct control 😄.
I noticed while testing that static structures get the attack move behavior when holding alt. They shouldn't do this because they can't attack move. 
Also, it would be convenient to repurpose the Alt key so that holding it enters the launch tactical/nuke command when your entire selection is static tml/sml. Releasing it exits the command.

PR #5910, opened 2024-02-17
Improve path finding of the Blaze and Zthuee

Garanas:
Improve the path finding behavior of the Blaze by removing the `SlavedToBody = true` blueprint field and instead turn it into a turret. 

https://github.com/FAForever/fa/assets/15778155/2412f06e-a4e0-412f-9864-671564f386dc


Garanas:
The Zthuee now behaves similar to other tech 1 artillery units. It is unable to fire while rotating because the weapon is too slow in correcting the angle. That is intentional.

https://github.com/FAForever/fa/assets/15778155/627daf88-d6dd-4c54-9da9-0456f8510c43

Garanas:
The other two units that come to mind are the Aurora and the Riptide. Sadly, their weapons are attached to the body and it would look visually strange if we would apply the same fix:

image

image

On top of that, even if we'd fix the Aurora it would turn into an absolute beast. The 'hiccup' that the Aurora has can now be used to catch up to the Aurora more easily. If we'd change this, the Aurora would be extremely good at kiting.

PR #5915, opened 2024-02-19
Prevent tactical missiles from hitting TMD

### LINKED ISSUE #5903: Seraphim TML can circumvent TMD because of excessive missile speed

lL1l1:
Cherry picked from #5788.
- Resolves #5903: Seraphim missiles slow down as they approach the target through a new Terminal Phase in the tactical missile component. It allows setting a time or distance (for ease of use) from the target where the missile slows down or changes its zigzag.
- Torrent missiles no longer have a reduced zigzag when nearing the target, since it is balanced as an area bombardment ship.
- UEF TML, UEF ACU, and Aeon MML missiles are slower.


Garanas:
The speed decrease essentially reverts the speed changes introduced a while back. What if instead we increase the fire rate of TMD so that this is possible again? Practically it's the same

lL1l1:
Personally I don't feel that the speed changes are that aesthetically important, and this mostly changes specific TML missiles which are less frequently seen and more important for balance.
Increasing the fire rate of TMD would require thorough changes to fire rate and damage numbers of cruisers, torrent, MMLs, and Spearheads. It would also be a significant nerf to 1 HP missiles. *Especially* if balancing for 20 speed Seraphim TML.

Garanas:
But if all tactical missiles are faster, one solution would be to increase the fire rate of all TMDs to compensate to allow them to fire twice again in the give time frame?

Garanas:
Anyway; if people really want to slow down the tactical missiles again then I'm fine with that. I'll include this PR in the next release, please make sure you've tested it thoroughly to catch dem errors 😃

lua/sim/projectiles/components/TacticalMissileComponent.lua
```diff
@@ -22,8 +22,16 @@
 
 local SemiBallisticComponent = import("/lua/sim/projectiles/components/semiballisticcomponent.lua").SemiBallisticComponent
 
+local MathAbs = math.abs
+
 ---@class TacticalMissileComponent : SemiBallisticComponent
+---@field TerminalTime number       # How long before the terminal phase starts, in seconds. Default is 1/4 of the total glide time.
```
Garanas, line 28: In what scenario would one populate this field? A missile can take any amount of time to reach its target, without knowing the anticipated glide time it's difficult to determine a correct time. What do you think of just always using 1/4th of the total glide time?

lL1l1:
New commit:
- Fixes SCU missile speed being unchanged in the missile rework (13 instead of 20)

- The max speed and acceleration are increased on cruiser, carrier, and MML missiles so that the speed changing effect is prominent across the faction, not only TML.

lL1l1:
This is ready for release

Garanas:
It appears that at least UEF TMD can now occasionally miss the missile. Could that be because of the (increased) acceleration?

https://github.com/FAForever/fa/assets/15778155/e095956c-aac3-4193-be8c-c0dcf4935437

image

lL1l1:
I got the same results in testing as you, even if the different acceleration missiles approach the UEF TMD at the same velocity. Seems like something in the acceleration really messes up the TMD's targeting. Fortunately, there don't seem to be any issues with simply high velocity missiles, so I opted to increase the muzzle velocity while keeping the acceleration low so that seraphim missiles can still reach max speed easily. I think it makes sense since they are some kind of fired energy projectile instead of a more realistic (like UEF) self-propelled missile.
- UEF TMD no longer misses cruiser missiles.
- Since muzzle velocity is increased, launch ticks are decreased so that the missile doesn't fly higher than before.
- Naval missiles have increased launch turn rate to get rid of an ugly turn when they switch from launch to glide.
- TML/SACU/ACU missiles have further decreased launch ticks so that they can properly angle in close range.

lL1l1:
Also I noticed the Aeon and cybran TMLs fly very low in close range. The minimum range for TMLs got massively increased from 15 to 25, despite only Seraphim and UEF benefitting aesthetically from that (except ACU missiles which stayed short range). Do you think Aeon/Cybran TML should fly higher?

Garanas:
For Cybran I'm fine either way. 

The problem with Aeon missiles is that they need to zig zag. And zig zagging is tied to the turn rate. Therefore they need to have a relatively high turn rate. But because of that, they fly lower. At the time I did not find a way to make them have a decent trajectory **and** have the zig zag behavior. But I bet with more tweaking of the values such as muzzle velocity we can get somewhere.

PR #5918, opened 2024-02-19
Buff Continental

apollodeathstar:
As per discussion in https://discord.com/channels/197033481883222026/1208637035682725920

For balance team review
The Connie is currently a very underused unique UEF Unit as a T3 Transport. With a high mass cost and high build time.
you could get x4.2 T2 UEF transports for the cost of a single Connie. It has a low to nil usage in ranked matches.

As a fraction unique it should be more capable for its role of getting units to their reasonable destination and maybe making it home.

It should not be capable of fighting other aircraft for their roles, but it should be capable in its own role and a wanted upgrade over a T2 transport in a T3 environment.

-Reverts 2013 E change https://github.com/FAForever/fa/commit/3e78ef4dcc7ff95fb87842dce0d30b62cb873a88
&

Connie Target priority changes moved to https://github.com/FAForever/fa/pull/5921

-Changes

Shield energy cost reverted from 2013 change
Shield changed from 3k to 6k
Speed changed from 15 to 17.3 (just below T2 interceptor)
Numbers are as always flexible and open to changes.


Basilisk3:
> Shield energy cost reverted from 2013 change
Shield changed from 3k to 6k
Speed changed from 15 to 17.3 (just below T2 interceptor)

These changes are way too transformative, especially for a unit that is exclusive to only one faction. The Continental is already the obvious choice of transport later in the game, as the units it carries are significantly more expensive than the transport itself. When dropping Tech 3 units worth several thousands of mass, spending 1400 mass on the transport is not a big deal. If Tech 3 Air is available, you would always want to use a Continental instead of a Tech 2 transport to drop your Percivals (or your ACU). 

Attempting to down a 10500 HP transport flying at these speeds would also be very challenging and frustrating.

The maintenance cost does seem excessive, so reducing that does make sense to me.

zqh9843:
I strongly agree with the e drain reduction. But even 250e/s for a mere 3k shield is still a bit too high IMHO.
If more overall health is OP, why don't we swap the shield for health so the overall HP stays the same?
hp: 4.5k->3k
shield: 3k->4.5k
shield regeneration: 30->45
rebuild time: 35->40
This will buff continental's longevity just a little bit between short encounters and improve its cargo protection, without making it very challenging and frustrating to deal with since overall HP is the same.
Remember that Continental is not only more prone to putting more eggs into the same basket but also suffers from higher BT/slot compared to 2 T2 transports resulting in an overall less efficiency which is one of the reasons they aren't widely used atm.

Garanas:
@apollodeathstar Please split the pull request into balance related changes (shield, consumption, speed) and non-balance related changes (target priorities). Then we can merge in the target priorities separately

apollodeathstar:
> @apollodeathstar Please split the pull request into balance related changes (shield, consumption, speed) and non-balance related changes (target priorities). Then we can merge in the target priorities separately

https://github.com/FAForever/fa/pull/5921

apollodeathstar:
> I strongly agree with the e drain reduction. But even 250e/s for a mere 3k shield is still a bit too high IMHO. If more overall health is OP, why don't we swap the shield for health so the overall HP stays the same? hp: 4.5k->3k shield: 3k->4.5k shield regeneration: 30->45 rebuild time: 35->40 This will buff continental's longevity just a little bit between short encounters and improve its cargo protection, without making it very challenging and frustrating to deal with since overall HP is the same. Remember that Continental is not only more prone to putting more eggs into the same basket but also suffers from higher BT/slot compared to 2 T2 transports.

I like this idea of trading its health for more shield

MrRowey:
@apollodeathstar can you please fix the conflicts

MrRowey:
Closing this PR as the balance team dose not see the need for buffing the Continental at this stage

PR #5921, opened 2024-02-20
Continental fix target priorities

apollodeathstar:
-Changes 
Connie Target priority changed to match that of gunships and other defensive AA units

-Changes
Connie Target priority changed to match that of gunships and other defensive AA units before these changes its AA would copy that of a ground SAM sight and focus bombers and ground attack air threats first where as its ground attack weapons would not prioritize ANTIAIR ground units that were a threat to it and would just target ALLUNITS.


units/XEA0306/XEA0306_unit.bp
```diff
@@ -339,7 +339,14 @@ UnitBlueprint{
             RackSlavedToTurret = false,
             RangeCategory = "UWRC_DirectFire",
             RateOfFire = 2.2,
-            TargetPriorities = { "ALLUNITS" },
+            TargetPriorities = {                
+                "ANTIAIR",
+                "GROUNDATTACK",
+                "TECH3 MOBILE",
```
Garanas, line 345: It seems better to me to directly go to 'ALLUNITS' after "ANTIAIR" and "GROUNDATTACK". These weapons do not deal sufficient damage to more complicated target priorities.

apollodeathstar: changed

PR #5922, opened 2024-02-20
Fix Aeon and UEF T2 transport slot/clamp count for "load into transports" command

lL1l1:
UEF was using Cybran's count. Aeon was based off the outdated description (bones causing this were fixed 2 years ago).


apollodeathstar:
I noticed the connie has this set as its transport slots   
```
Transport = {
        AirClass = true,
        Class1Capacity = 28,
        Class2AttachSize = 2,
        Class3AttachSize = 4,
        SlotsLarge = 6,
        SlotsMedium = 12,
        SlotsSmall = 24,
        TransportClass = 10,
``` 
Where as in other transports the SlotsSmall equal the Class1Capacity. When tested it did still loaded 28 units but could this be throwing errors or causing problems for AI etc?

PR #5923, opened 2024-02-20
Add the ability to load into transports by right clicking the transport button

lL1l1:
In general it's bad that we have commands completely inaccessible from the GUI.


Garanas:
Great idea 👍

PR #5924, opened 2024-02-20
Fix a rare error when the GC and ML make landfall

Basilisk3:
This is the same issue as fixed in #5829.

**How to reproduce this 100% of the time:** 
- Spawn a line of GCs via the Alt+F2 menu (the line must be both on land and in the water)
- Give a movement order to the units you spawned in

This is a very rare bug, but it can happen in real games too.


Garanas:
Can you post the error/warning that shows in the log?

Basilisk3:
> Can you post the error/warning that shows in the log?

`WARNING: *WARNING: No movement effect groups defined for unit 	ual0401	, Effect groups with bone lists must be defined to play movement effects. Add these to the Display.MovementEffects	Land	.Effects table in unit blueprint. `

Garanas:
I'm wondering what/when/how this happened. Was the blueprint always wrong, for example? Or did we change something that made this wrong?

Basilisk3:
These errors only started occurring semi-recently, I first noticed one case in #5829. As far as I can see, the blueprints have always been like this, but they did not error out.

Garanas:
That's strange, it would suggest that the way we interpret the blueprints has changed.

Garanas:
There's an issue with this pull request: all the units processed no longer deal footfall damage!

The code expects footfalls next to the `Effects` table and not inside it. We'll need to move the footfall information outside of it again as that's how mods and everything work too. 

Related code:

https://github.com/FAForever/fa/blob/9db1b491da2308fb4f7fe9b0f6d1af82772e57c2/lua/sim/Unit.lua#L2413-L2425

Basilisk3:
@Garanas I will revert it for the GC, ML and the Mega then.

PR #5925, opened 2024-02-21
Fix an error when Cybran air factories are destroyed while they are building a unit

### LINKED ISSUE #5796: Bug in effectutilities.lua
Bug 3778. Need to add a check for nil or on unit death..
```
warning: Error running lua script: ...a\faforever\gamedata\lua.nx2\lua\effectutilities.lua(345): attempt to loop over field `BuildEffectBones' (a nil value)
         stack traceback:
         	...a\faforever\gamedata\lua.nx2\lua\effectutilities.lua(345): in function <...a\faforever\gamedata\lua.nx2\lua\effectutilities.lua:341>
         	...ata\lua.nx2\lua\sim\units\cybran\cairfactoryunit.lua(62): in function <...ata\lua.nx2\lua\sim\units\cybran\cairfactoryunit.lua:56>
         	```

### PR DESCRIPTION
**Pull request by Basilisk3**
Closes #5796

### DIFF
--- a/lua/EffectUtilities.lua
+++ b/lua/EffectUtilities.lua
@@ -340,10 +340,10 @@ local UnitBuildEffects = {
 ---@param buildEffectsBag TrashBag
 function CreateCybranFactoryBuildEffects(builder, unitBeingBuilt, buildBones, buildEffectsBag)
     CreateCybranBuildBeamsOpti(builder, nil, unitBeingBuilt, buildEffectsBag, false)
-    WaitTicks(10)
+    WaitTicks(8)
 
     local builderArmy = builder.Army
-    while not unitBeingBuilt:BeenDestroyed() do
+    if not unitBeingBuilt:BeenDestroyed() then
         for _, bone in buildBones.BuildEffectBones do
             for _, effect in BuildEffects do
                 TrashBagAdd(buildEffectsBag, CreateAttachedEmitter(builder, bone, builderArmy, effect))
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: lua/EffectUtilities.lua [hunk 1]
```diff
@@ -340,11 +340,14 @@ local UnitBuildEffects = {
 ---@param buildEffectsBag TrashBag
 function CreateCybranFactoryBuildEffects(builder, unitBeingBuilt, buildBones, buildEffectsBag)
     CreateCybranBuildBeamsOpti(builder, nil, unitBeingBuilt, buildEffectsBag, false)
+    WaitTicks(10)
 
     local builderArmy = builder.Army
-    for _, bone in buildBones.BuildEffectBones do
-        for _, effect in BuildEffects do
-            TrashBagAdd(buildEffectsBag, CreateAttachedEmitter(builder, bone, builderArmy, effect))
+    while not unitBeingBuilt:BeenDestroyed() do
```
Garanas, line 346: I'm fairly certain that this should be an `if` statement and not a while loop.

lua/EffectUtilities.lua
```diff
@@ -340,11 +340,14 @@ local UnitBuildEffects = {
 ---@param buildEffectsBag TrashBag
 function CreateCybranFactoryBuildEffects(builder, unitBeingBuilt, buildBones, buildEffectsBag)
     CreateCybranBuildBeamsOpti(builder, nil, unitBeingBuilt, buildEffectsBag, false)
+    WaitTicks(10)
```
Garanas, line 343: Why would we wait here?

lua/EffectUtilities.lua
Basilisk3: 👍

lua/EffectUtilities.lua
Basilisk3: I reduced it to 8, unfortunately some sort of delay appears to be required. At 5 ticks, it still errors out sometimes.

Garanas:
I'm not certain whether this is the right direction. The additional wait is 'fixing' something else, but I feel like we do not fully understand what the wait fixes.

Garanas:
I don't think this is the correct direction to take. The error is not critical, I'll try another approach in a bit when I re-evaluate the Cybran build effects in general.

PR #5926, opened 2024-02-22
[Revision] Buff Continental 2.0

zqh9843:
Current maintenance cost is a bit too excessive for a 3k shield, 400e/s is like 500 mass investment in T3 Pgen that both you and your land teammate will have to seriously consider.
Also the shield and armor HP are swapped so the continental fares a little bit better in long term operations(less need for landing/repair) and protecting its passengers without making intercepting it a big challenge. (compared to 10500 overall hp in previous PR made by Apollo)
Reduce maintenance cost and swap Shield/Armor values.
Armor: 4500 -> 3000
E drain of shield: 400 -> 225
Shield: 3000 -> 4500
Shield Rebuild: 35s -> 40s
Shield Regen: 35/s -> 45/s

All those stats can be adjusted if necessary.
Hopefully with these buffs we will see more use of Continental.


apollodeathstar:
I like these changes, I think it could use a tiny speed buff to 16 to help set it more apart and justify the extra investment T3 air investment and give it a little suga.

MrRowey:
The Balance Team at this moment in time sees no need to buff the Continental

Issue #5919, opened 2024-02-19
Target Priorities are not standardized among unit types


apollodeathstar:
**Describe the bug**
Target Priorities are not standardized among unit types, possibly leading to either performance issues or less then optimal unit actions, effecting balance. 

**Expected behavior**
Units that share unit types have standardized targeting behavior at similar performance costs.

**Additional context**
Under guidance, I would attempt to standardize unit target priorities across unit types.

_This bug report will be read by people that have not experienced the bug themselves - this report is all they have to investigate. The more precise your report, the easier it becomes for us to help you._

apollodeathstar:
I also think it may be good to add a new target priority category specifically for T1-3 interceptors/ASF then def Air units will be able to focus their defense on their biggest threats the ASF trying to kill them. 

The same category could be added to interceptors/ASF so that they focus other ASF's first but that may have to be checked for performance due to the very large numbers that get amassed in games.

I don't know if there has been much meta play regarding using T1 scouts etc in order to cheese/spoof enemy ASF balls but it is something to think about as a player behind on ASF numbers could tech pump out a bunch of scouts in order to make up for lack of ASF numbers in a fight to have enemy ASF waste salvos and win the fight. It is something to consider though. 

This may only be a issue in late game ASF vs ASF engagements so even just adding a category for ASF may be enough to ward off such an issue.

apollodeathstar:
I agree that's the best way to go about it. splitting it into manageable categories will make it easier to submit and track and air units are a good place to start because it will be simpler then more complex interactions.

Is there a targetpriorities list somewhere that has the full list of possible targets? If not that is probably the first thing I can work on gathering from the various unit files.

Garanas:
Target priorities are just (combined) categories. One could consider this to be a list:

- https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Core/Categories.lua

But in general for air units `GROUNDATTACK` means it is a gunship, `BOMBER` means it is a bomber, `TRANSPORTATION` means it is a transport and `ANTIAIR` means it shoots the previous three categories 😃

Garanas:
> I also think it may be good to add a new target priority category specifically for T1-3 interceptors/ASF then def Air units will be able to focus their defense on their biggest threats the ASF trying to kill them.

This would be the `ANTIAIR` category

apollodeathstar:
[...]

PR #5928, opened 2024-02-22
Evaluate and rework the categories of UEF air units

Garanas:
Related to #5919 

Introduces the `WEAK...` categories to various units. Also removes the `NUKE` category from the strategic bomber as it is... not a strategic unit. 

~~What I found interesting is that there's very little to no distinction between regular bombers and torpedo bombers. I was thinking of introducing a `TORPEDO` category to the torpedo bomber to make it more distinguishable. Now, of course - mods that are unmaintained will not follow this practice and therefore it would only apply to units we ship with FAForever directly.~~

edit: apparently the `ANTINAVY` category means 'something that fires torpedo's according to torpedo weapons and the categories of naval units, as seen in https://github.com/FAForever/fa/pull/5929. As a result, introducing `TORPEDO` is not necessary because `ANTINAVY` essentially means that.

### DIFF
--- a/units/UEA0103/UEA0103_unit.bp
+++ b/units/UEA0103/UEA0103_unit.bp
@@ -54,6 +54,7 @@ UnitBlueprint{
         "PRODUCTSC1",
         "RECLAIMABLE",
         "SELECTABLE",
+        "SHOWATTACKRETICLE",
         "SNIPEMODE",
         "TECH1",
         "UEF",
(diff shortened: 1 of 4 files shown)


Garanas:
@relent0r / @maudlin27 please evaluate these category changes to make sure they do not mess with your AI 😃 !

relent0r:
No issues with this one.

p.s did not know that particular bomber had a nuke category!

PR #5929, opened 2024-02-22
Evaluate and rework the categories of UEF naval units

Garanas:
Related to https://github.com/FAForever/fa/issues/5919

Introduces the WEAK... categories to various units. Also removes the NUKE category from the strategic bomber as it is... not a strategic unit.

What I found interesting is that the `ANTINAVY` category actually means `anti submarine`. Weapons that are anti navy are torpedo-like weapons.

### DIFF
--- a/engine/Core/Categories.lua
+++ b/engine/Core/Categories.lua
@@ -13,12 +13,19 @@ categories = {
     ALLPROJECTILES = categoryValue,
     ALLUNITS = categoryValue,
     AMPHIBIOUS = categoryValue,
+    --- Used for units that can target air units
     ANTIAIR = categoryValue,
+    --- Used by support commander presets
     ANTIAIRPRESET = categoryValue,
+    --- Used for units that can target missiles
     ANTIMISSILE = categoryValue,
+    --- Used for units that can target submarines (torpedoes and depth charges)
     ANTINAVY = categoryValue,
+    --- Used for units that are specifically designed to target shields
     ANTISHIELD = categoryValue,
+    --- Unused, see `ANTINAVY` instead
     ANTISUB = categoryValue,
+    --- Used for units that can target torpedoes
     ANTITORPEDO = categoryValue,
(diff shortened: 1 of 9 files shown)


Garanas:
@relent0r / @maudlin27 please evaluate these category changes to make sure they do not mess with your AI 😃 !

Especially the removal of `ANTINAVY` from the Battlecruiser and the Destroyer. They still have the overlay and the `WEAKANTINAVY` categories, but they do not deserve the `ANTINAVY` category.

relent0r:
This one I don't love, its going to cause problems when trying to analyze if destroyers can deal with submarines or attack an underwater unit.

Garanas:
The destroyers of the UEF are definitely not made to fight with submarines. It's intentionally weak because of the Torpedo Boat.

Do you have snippets of code where you analyze what a unit can (not) target?

relent0r:
An example of the use case would be something like this.

PlatoonTemplate {
    Name = 'RNGAI Sea Hunt',
    Plan = 'StateMachineAIRNG',
    GlobalSquads = {
        { categories.MOBILE * categories.NAVAL * categories.ANTINAVY, 1, 10, 'Attack', 'GrowthFormation' }
    },
}

This is used for taking out under water mexes and dealing with submarines. With this change the AI would never form this platoon if only destroyers were available, even though a destroyer can tank enough damage to take out multiple submarines attacking its units without much trouble. 
But thats more of a balance debate than a functionality one. You could apply the same logic to frigates since they have weak aa, but then if you've got a group of frigates and someones using t1 bombers to destroy them your not going to pack up and go home as they can tank the damage and dish out enough to kill the bomber/s.


So for this PR I would need to add another category to a platoon former like this is all. Not the end of the world and not a major inconvenience.

Garanas:
@relent0r I've re-introduced the `ANTINAVY` categories. We can look into that separately in the future if the need arises

PR #5932, opened 2024-02-23
Refactor how the reclaim costs are computed

Garanas:
Relates to #5931 .

Refactors how the reclaim costs are computed, which then turned up an interesting 'bug' of the code: in practice it will never use the mass build cost but always the energy build cost of a unit because there are no units that cost more mass than they cost energy 😃

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -3932,7 +3932,7 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent) {
             -- find largest build cost value, this is always energy? :)
             local costs = buildEnergyCosts
             if buildMassCosts > buildEnergyCosts then
-                costs = buildEnergyCosts
+                costs = buildMassCosts
             end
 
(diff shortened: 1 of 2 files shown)


Garanas:
Note that this should not change the output of the function, it just makes it faster.

lua/sim/Unit.lua
```diff
@@ -1076,7 +1082,12 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent) {
 
     ---@param self Unit
     GetBuildRate = function(self)
-        return math.max(moho.unit_methods.GetBuildRate(self), 0.00001) -- Make sure we're never returning 0, this value will be used to divide with
+        local buildrate = cUnitGetBuildRate(self)
+        if buildrate < 0 then
+            buildrate = 0.00001
```
4z0t, line 1087: Value never used afterwards.

Garanas: You're right, fixed with https://github.com/FAForever/fa/commit/6c22f57bccbdcf32932f2f18c62130e8f630cbcf

PR #5933, opened 2024-02-23
Fix the collision boxes of various miscellaneous units to prevent the game from creating warnings for them at the start of the game

Basilisk3:


Garanas:
Great work 👍

PR #5935, opened 2024-02-24
T3 Cybran Gunship AA weak added

apollodeathstar:
"WEAKANTIAIR" added


Garanas:
Is this supposed to be in draft? 😃

apollodeathstar:
> Is this supposed to be in draft? 😃

It was, and a good thing as well because I had to fix them.

PR #5938, opened 2024-02-24
T3 UEF Connie Weak AA and DR added

apollodeathstar:
"WEAKANTIAIR",
        "WEAKDIRECTFIRE",
added


Garanas:
It should be `WEAKANTIAIR` and not `ANTIAIRWEAK` 😃

PR #5939, opened 2024-02-26
Improve the naming scheme of all UEF drones and update their descriptions

Basilisk3:
**ACU drone**
- The old name for the UEF ACU's second drone upgrade was `C-D2 Shoulder Drone`, which is wrong, both drones are identical, and called `C-D1 "Rover"`. The `C-D2 "Rover"` is/was the name of the UEF SACU's drone.

**SACU drone**
- The name of the drone itself is changed to `C-D3 "Rover-3"`, in accordance with the untis it can build.
- The name of the drone upgrade is changed to `"C-D3 Engineering Drone"`.
- The Tech Level of the drone is moved from `Tech1` to `Tech3`, in accordance with the untis it can build.
- The drone now gets its own description (`Unit_Description_0457`), instead of using the ACU drone's description.
- The unit and upgrade descriptions now correctly describe which structures it can build (it used to say it can only build Tech 1 units).

**Kennel drone**
- The Tech Level of the drone is moved from `Tech1` to `Tech2`, in accordance with the unit it is housed in.
- Its description now correctly states that it cannot initiate the contruction of structures.


lL1l1:
The descriptions for the kennel drone and SACU drone are incorrect about what structures they can build as well.

Basilisk3:
> The descriptions for the kennel drone and SACU drone are incorrect about what structures they can build as well.

Do you mean the SACU's drone should be called `C-D3 "Rover"` instead?

Garanas:
Looks great 👍 , changes make the game more consistent

Issue #5768, opened 2023-12-24
Not reproducible: Beam weapons can still miss the Mantis


Basilisk3:
Replay ID: 21471096
Timestamp: 9:15, watch Swatoslav's first Tech 2 PDs in his mid base.

I cannot reproduce this, even with gifted units like in the replay.

Basilisk3:
It appears that this is not exclusive to the mantis and that it can happen to other units as well.

Basilisk3:
I also remember seeing this happen to the shard. @Garanas Shall we just increase the Z-size of these units, and see if this still happens?

Garanas:
I'm not sure. If it is trivial to reproduce then yes, but for rare edge cases I don't think we should keep increasing the size of the collision boxes.

The Shard still has some spare room, you can increase that by one or more one tents!

Basilisk3:
It is not reproducible, but I have seen it multiple times now (at least with the Mantis). In case I see it in more replays, I will post those too.

Basilisk3:
This is probably related to this:

https://github.com/FAForever/fa/issues/5742#issuecomment-1858836306

Basilisk3:
A fix is in the pipeline: https://github.com/FAForever/FA-Binary-Patches/pull/143

PR #5941, opened 2024-02-27
Attempt to fix the Mantis and the Medusa being able to evade beam weapons

Basilisk3:
**Changes**
SizeZ of the Mantis: 0.9 --> 1.0
SizeZ of the Medusa: 1.0 --> 1.1

For further context see #5768


Garanas:
It appears fine still, especially in comparison to the other factions. 

image

Issue #5452, opened 2023-09-18
Balance idea: ASF


Blodir:
## Intended changes

Mass cost 350 > 450 (~28%)
E cost +28%
Bt cost +28%
HP +-1800 > 2300
Damage 400 > 500

## Thoughts on the changes

Reduce the size of ASF blobs in the lategame. Reduce ASF effectiveness vs interceptors (due to larger overkill). Bring ASF mass cost more in line with other T3 units.

_This is a balance idea from the balance team. It is not set in stone, it may be adjusted or rejected as the changes are tested._

PR #5944, opened 2024-02-28
Increase Cybran ASF stealth maintenance in accordance with previous cost increases

lL1l1:
#5452 increased ASF cost by ~28%, but missed the stealth maintenance cost, thus unintentionally making Cybran ASF 22% cheaper to stealth relative to their cost.


Garanas:
What if we go for `30` or `35` instead? Usually these are values are a multiple of `5`, at least initially before adjacency is applied

Tagada14:
I would go with 30 e/s

MrRowey:
like how this was only approved in the balance team and already merged.

PR #5945, opened 2024-02-28
Formations: Fix the Tech 3 Sub Hunter not being defined as “SubNaval”

Basilisk3:
This is a problem because the following line gets spammed into the log hundreds, even thousands of times:

`INFO: *FORMATION DEBUG: Unit xss0304 does not match any Naval categories.`

This dilutes the log and makes it less readable. Additionally, I cannot image this is good for the performance of the game. The more Sub Hunters you have, the more lines are printed into the log. 

**How to reproduce the issue:**
- Open the Moho Log and deselect everything except “Info”
- Clear the log
- Spawn in a number of Tech 3 Sub Hunters
- Select the Sub Hunters you spawned in and give them a move order
- Select a portion of the Sub Hunters and give them another move order
- You will now see that the log gets spammed by the above message

### DIFF
--- a/lua/formations.lua
+++ b/lua/formations.lua
@@ -449,7 +449,7 @@ local GrowthChevronBlock = {
 
 local LightAttackNaval = categories.LIGHTBOAT
 local FrigateNaval = categories.FRIGATE
-local SubNaval = categories.T1SUBMARINE + categories.T2SUBMARINE + (categories.NUKESUB * categories.ANTINAVY - categories.NUKE)
+local SubNaval = categories.T1SUBMARINE + categories.T2SUBMARINE + (categories.SERAPHIM * categories.TECH3 * categories.SUBMERSIBLE)
 local DestroyerNaval = categories.DESTROYER
 local CruiserNaval = categories.CRUISER
 local BattleshipNaval = categories.BATTLESHIP


Basilisk3:
`(categories.NUKESUB * categories.ANTINAVY - categories.NUKE)` 

I assume this definition was made to include the Yathsou in `SubNaval`, because it adds all units with the NUKESUB category and then removes all units with the NUKE category, which would only leave the Tech 3 Sub Hunter. At least that is what it did until 2022, because the `NUKESUB` category was removed from the Yathsou in #4031.

lua/formations.lua
```diff
@@ -449,7 +449,7 @@ local GrowthChevronBlock = {
 
 local LightAttackNaval = categories.LIGHTBOAT
 local FrigateNaval = categories.FRIGATE
-local SubNaval = categories.T1SUBMARINE + categories.T2SUBMARINE + (categories.NUKESUB * categories.ANTINAVY - categories.NUKE)
+local SubNaval = categories.T1SUBMARINE + categories.T2SUBMARINE + (categories.SERAPHIM * categories.TECH3 * categories.SUBMERSIBLE)
```
lL1l1, line 452:
```suggestion
local SubNaval = categories.T1SUBMARINE + categories.T2SUBMARINE + (categories.TECH3 * categories.SUBMERSIBLE * categories.NAVAL - categories.NUKE)
```
This would be more general-purpose for mod compatibility. `NAVAL` category excludes HARMS. 
This would still catch *any* submersible tech 3 units, e.g. a submerging battleship like Tempest but at tech 3, which is why I could see a `T3SUBMARINE` category being useful.

lL1l1 (changes requested):

Garanas:
Reminder to https://github.com/FAForever/fa/pull/5908 where we also need to apply it properly 😃

PR #5947, opened 2024-03-02
Make unbuilt structures and experimentals not spawn a wreck

### LINKED ISSUE #5837: Balance idea: No wrecks to spawn unless building completed
## Intended changes

This is based on these older issues (#2375). The team discussed and would like it to be that it doesn't leave any wrecks when a building or experiment is built unless it is completed.

- This should affect just the Building and Experimentals themself.

lL1l1:
Resolves #5837. T1-T3 units in factories still spawn a wreck above 50% completion progress.


MrRowey:
Can we get this in the next patch, please :)

PR #5948, opened 2024-03-02
Create capture cost reticle

lL1l1:
A capture cost reticle that displays the energy cost, energy rate, and time for the current selection's total buildrate when mousing over non-allied, capturable units.

Improvements: Would be nice to be able to directly check if a unit is capturable or if it has a capture rate modifier.
Code-wise, not sure why the teleport reticle was attached to the command mode data, but capture isn't always called with a command mode active (mousing over enemy units tries to capture them) so I can't do that anyway.


lua/ui/controls/reticles/capture.lua
```diff
@@ -0,0 +1,118 @@
...
+                end
+            end
+        end
+        if isEnemy and isCapturable then
+            if self:IsHidden() then
+                self:SetHidden(false)
+            end
+
+            local targetBpEconomy = targetBp.Economy
+            -- Mimic Unit.lua GetCaptureCosts calculations
```
Garanas, line 98: Best to create a shared function that computes the capture costs

lua/ui/controls/reticles/capture.lua
```diff
@@ -0,0 +1,118 @@
...
+
+    ---@param self CaptureReticle
+    SetLayout = function(self)
+        local selection = GetSelectedUnits()
+        local totalBuildRate = 0
+        for _, unit in selection do
+            totalBuildRate = totalBuildRate + unit:GetBuildRate()
+        end
+        self.selectionBuildRate = totalBuildRate
+        self.focusArmy = GetArmiesTable().focusArmy
```
Garanas, line 52: You can also use `GetFocusArmy()`

Garanas (commented):
Really exciting 😃 !

lL1l1:
How difficult would it be to include capturability in the rollover info?

----

> `SetCaptureTimeMultiplier`
>  Multiplies the time it takes to capture a unit, defaults to 1.0. Often used by campaign events.

Isn't this incorrect? It seems like it multiplies the capture time of a unit capturing others, not the capture time of a unit being captured.
https://github.com/FAForever/fa/blob/d38a6fa6127e34a571b424f7cae431c633112056/lua/sim/Unit.lua#L3979-L3981
To "multiply the time it takes to capture a unit" it would have to be `time = time * (target.CaptureTimeMultiplier or 1)`.

Garanas:
> How difficult would it be to include capturability in the rollover info?

I'm not certain, we'd need to ask @Hdt80bro or @4z0t about that 😃 

> Isn't this incorrect? It seems like it multiplies the capture time of a unit capturing others, not the capture time of a unit being captured.

It applies to the engineer that is capturing yes, that would be my mistake in the comment.

lL1l1:
What would be the best way to update the selection when it changes? The selection could change by certain UI actions or more commonly by units dying.

arch625:
Does it update the reticle to the new total leftover time remaining when you add additional build power later on onto the capture target?

lL1l1:
> Does it update the reticle to the new total leftover time remaining when you add additional build power later on onto the capture target?

It doesn't.
I looked into it since it's a good idea, but there's no way UI-side to get info about what units are currently capturing an enemy/neutral unit or the capture progress on that unit.

Garanas (approved):
Such a good improvement, apologies that it took this long to review. Looking at this it may also be interesting to have similar reticle's for the reclaim duration when in reclaim mode and for the total damage per second when in attack mode.

lL1l1:
I considered a reclaim cursor (it could be immediately useful analytical info) but an implementation would be incomplete:
- How do I get the mass/energy of a prop under the cursor when in reclaim mode? Reclaim time is `highestCost / (5 * buildrate)` as far as I know.
- Units are set to be unreclaimable/have adjusted reclaim values with scenario scripts (on the map Loki).
```lua
makeInvincible = function(unit)
	unit:SetDoNotTarget(true)
	unit:SetCanBeKilled(false)
	unit:SetCapturable(false)
	unit:SetReclaimable(false)
	unit:SetDoNotTarget(true)
	unit:SetMaxHealth(1)
	unit:SetHealth(nil,1)
	unit:SetRegenRate(1)
end

makelowmass = function(unit)
	unit:GetBlueprint().Wreckage.MassMult=0.05
end
```
I tried making a reclaim cursor with just blueprint values but it doesn't update when the Sim modifies the blueprint like this, and `SetReclaimable`'s effects are inaccessible with rollover info (same issue as `SetCapturable`).

A DPS cursor seems more appropriate for a UI mod because it's not very useful information and doesn't require any Sim code.

lL1l1:
https://github.com/lL1l1/fa/tree/UI%2Freclaim-reticle

Garanas:
Ah, you're right - these are valid issues. Especially when the (map) script adjusts the values then all bets are off.

PR #5127, opened 2023-06-22
Allow UEF drones reclaim during patrol/attack-move

4z0t:
https://discord.com/channels/197033481883222026/1121169759187255398

PR #5950, opened 2024-03-02
First draft of a rework for the UEF ACU's drone upgrades

Basilisk3:
These upgrades have been in dire need for a buff for a long time now. Instead of buffing them straight-up, I decided to rework the upgrades and change the role they occupy in-game. 

**I believe a rework is necessary because:** 

- If they were buffed and kept in their current role, they could become too strong on maps where quick expansions are important (e.g. Point of Reach). I assume this is also why the build time of the current upgrades is excessively high.
- Adding an additional 10 build power to your ACU is essentially useless in the mid to late stages of the game. With these changes you get 40 additional build power.

**Changes:**
Drone BuildRate: 5 --> 20 (x4)

Upgrade BuildCostEnergy: 2400 --> 10800 (x4.5)
Upgrade BuildCostMass: 120 --> 320 (x2.66)
Upgrade BuildTime: 400 --> 800 (x2)


Garanas:
To me the biggest problem with these drones is that they are almost instantly evaporated by any unit that can target them. In your proposed changes they become even more prone to a lone interceptor as it is worth 10 interceptors mass-wise.

Just as an idea: we could also make them cheaper, but limit their action radius. That way you can't use them to expand unless your ACU is nearby. They'd become a tool to quickly build extractors on top of cliffs, for example.

Basilisk3:
The economy changes are to the upgrade and not the drones themselves (I improved the formatting of my post). The economy values of the drones appear to be hardcoded, I was unable to change those at all.

lL1l1:
> The economy values of the drones appear to be hardcoded, I was unable to change those at all.

https://github.com/FAForever/fa/blob/f1c4d5ceab65b482102531e65890c0b671ede972/units/UEL0001/UEL0001_script.lua#L166

It's hardcoded as 1600 energy, 160 mass over 10 seconds.

Basilisk3:
👍I think they are reasonable for the time being.

Basilisk3:
I think this is ready to be reviewed now, feedback is welcome.

Garanas:
> These upgrades have been in dire need for a buff for a long time now. Instead of buffing them straight-up, I decided to rework the upgrades and change the role they occupy in-game.

What is the role that you had in mind with these changes?

Basilisk3:
A help for base building and rebuilding destroyed expansions. I experimented with giving them more health, but due to how small these units are, it did not make a lot of sense. I also evaluated their engineering suite, e.g. essentially turning them into flying field engineers. It ended up not really working out.

So this is just a straight-up buff for the time being, I am still open to suggestions.

MrRowey:
could you make it so it s field engine but if the acu is tech 2 then I get access to that suite of tools too as a field engine ? as other wise it kinda not worth having them  later on in the game and having shield

Basilisk3:
Unfortunately, I do not know how. 

Theoretically, we could combine and rearrange the upgrades to the left arm, as an additional option after the Tech 2 and Tech 3 upgrades. So: `Tech 2 --> Tech 3 --> Drone Upgrade`. And then give both drones a Tech 3 suite. I am not sure if that is a good idea though, it would make them appear pretty late in the game.

Garanas:
I think the suggestion of @MrRowey would be better discussed separately from these changes. Adjusting the tech suite of the drones is not trivial, requires some scripting and keeping track of the status quo of the ACU.

apollodeathstar:
anyway you can make them able to perform attack move/patrol based reclaim? I think that would buff the units a bit. then you could send them out to hard to reach reclaim and not have to individually click each reclaim spot. currently they cant do attack move or patrol reclaim. also it would be nice if the replacement drones kept the same assigned Ctrl + # squads.

Basilisk3:
> anyway you can make them able to perform attack move/patrol based reclaim?

That should already be possible with #5127.

MrRowey:
Closing this PR for now as Looking at SCU/ACU Upgrade is on the table to be looked at.

Issue #5891, opened 2024-02-10
Valiant Class: UEF Tech 2 Destroyer


Garanas:
# Valiant Class

![image](https://github.com/FAForever/fa/assets/15778155/bfd6fddd-abf8-4dc4-80ad-fcfb6321dded)

- Blueprint identifier: UES0201

## Lua script related issues

The radar dishes keep rotating even though the unit is sinking.

## Effect related issues

- [x] Tracks are correct (if applicable)
- [x] Footfalls are correct (if applicable)

## Texture related issues

- [x] Correct albedo intensity
- [x] Correct team color channel
- [x] Correct roughness channel
- [x] Correct metallic channel(s)
- [ ] Correct normals

## Mesh/bone related issues

- [x] Bones for target recoil exist
- [x] Bones for targeting exist (only relevant for large units)
- [x] LOD1 exists
- [ ] LOD2 exists (optionally)
- [ ] LOD3 exists (optionally)

## Relevant shaders

You can find all relevant shaders in [mesh.fx](/effects/mesh.fx).

- Mesh shader techniques: `Unit_HighFidelity` / `Unit_MedFidelity` / `Unit_LowFidelity`
- Shield shader techniques: `ShieldUEF_MedFidelity` / `ShieldUEF_LowFidelity`

### Relevant texture interpretations

| Mesh shaders | R         | G         | B        | A                                           |
| ------------ | --------- | --------- | -------- | ------------------------------------------- |
| Albedo       | R         | G         | B        | Unused                                      |
| Normals      | Unused    | Y         | Unused   | X                                           |
| Spec         | Roughness | Roughness | Emission | Team color / Metallic + roughness reduction |

PR #5953, opened 2024-03-03
New mesh and normal texture for the valiant

Ajamajan:
had to reexport mesh too since the mesh normals were broken

related issue: https://github.com/FAForever/fa/issues/5891

OLD
image

NEW
image


BlackYps:
It looks like the barrels of the main weapon are now flat. Is that correct?

Garanas:
Some more images:

image

image

image

BlackYps:
yeah, this barrel should be round in my opinion

Garanas:
I think the problem is that the barrel has 4 sides, where as the barrel of the frigate has 5 sides. I don't think it matters much because you need to be zoomed in quite a bit, but if it is not much effort then it also can't harm 😃

BlackYps:
It only having four sides is not a problem. Even with three sides you could interpolate the normal. You can even see in the old comparison screenshot that it was interpolated and thus appeared round before.
Luckily it is very easy to fix

Ajamajan:
I could change it if it REALLY REALLY bothered you guys. the fact that the smaller barrels on the AA turret have more subdivisions lets me think that 4 on the bigger cannon suggests it may be intended as a hard edge barrel. since i already re-exported the mesh anyway the fix shouldnt take too long. IF it is requested.

Garanas:
I do not personally see it as a (big) issue, the unit is already quite a bit better now than it used to be.

BlackYps:
I personally would appreciate it, if you changed it to a round barrel :3

Ajamajan:
i am getting mixed signals here..

BlackYps:
Idk, Jip says it is already good, I say it would be even better if the barrel was smooth shaded. That's not a contradiction to me.
Ultimately it is your decision if you want to edit the barrel again or not

Garanas:
I'm fine with what we have, but a smooth barrel would of course be better. I'll be merging this in because on its own this pull request is a large improvement. Feel free to make a separate pull request with the changes for a better barrel however 😃

jospoortvliet:
Looking super nice!!!

PR #5958, opened 2024-03-06
Fix SACUs having less effective range due to low projectile lifetimes

Basilisk3:
In the current branch, the projectile lifetime of three of the four SACUs is extremely low. This means, that at the edge of their effective ranges, their shots disappear in front of units before they can hit them. This is especially bad for the Seraphim SACU, as that one does not have an AOE upgrade (ignoring OC). The projectile lifetime for the Cybran SACU was excessive, so I reduced it to be more reasonable.

**Changes:**
ProjectileLifetimeUsesMultiplier: 1.15 --> 1.3 (Aeon; Seraphim; UEF)
ProjectileLifetimeUsesMultiplier: 2.4 --> 1.4 (Cybran) 

A higher multiplier is required for the Cybran SACU, as its projectile's muzzle velocity is lower.

Only relevant with the respective range-increasing upgrades.

### DIFF
--- a/units/UAL0301/UAL0301_unit.bp
+++ b/units/UAL0301/UAL0301_unit.bp
@@ -624,7 +624,7 @@ UnitBlueprint{
             MuzzleVelocity = 25,
             MuzzleVelocityReduceDistance = 5,
             ProjectileId = "/projectiles/ADFReactonCannon01/ADFReactonCannon01_proj.bp",
-            ProjectileLifetimeUsesMultiplier = 1.15,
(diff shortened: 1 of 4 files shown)


Basilisk3:
I believe in the case of the Aeon, Seraphim and UEF SACUs, this has been an issue for over 10 years, which is disheartening to think about😃

Garanas:
There's been a lot more findings the past few years that makes you wonder why it's been like that for ten years 😃

PR #5960, opened 2024-03-09
Rework the representation of the high-value tactical missiles

Garanas:
Based on feedback from [Discord](https://discord.com/channels/197033481883222026/1215210337352024184) and the [forums](https://forum.faforever.com/topic/7293/suggestion-for-tml-bigger-yellow-projectile-indicator/1).

High-value tactical missiles now have a strategic icon. The Billy has a separate strategic icon. Their effects also better represent the damage. Light particles are introduced so that the effect is visible at any zoom level. Sounds are adjusted so that they're more noticeable at more distance zoom levels.

https://github.com/FAForever/fa/assets/15778155/74b5c0fd-d63f-440e-ba0f-50a72820eae9


Basilisk3:
I think that looks much better than before👍

PR #5963, opened 2024-03-10
Improve the handling of all movable Sonars

Basilisk3:
**Explanation of the Changes**
When the Seraphim Tech 2 Sonar was made movable, it was given an extremely low TurnRate, which is the reason it always felt very sluggish to use. You can test this yourself, by issuing a movement order close to its current position at a 90-degree angle. Its movement essentially comes to a crawl, perhaps moving at a third of its normal movement speed. Tech 3 sonars do not suffer from this problem, as their TurnRate is five times higher.

In addition, all movable Sonars get a BackUpDistance, because it helps with pathfinding.

**All Tech 3 Sonars:**
BackUpDistance: 0 --> 10
MaxSpeedReverse: 0 --> 4.5

**Shou-esel (Seraphim Tech2 Sonar System)**
BackUpDistance: 0 --> 8
MaxAcceleration: 2.5 --> 2.8
MaxBrake: 2.5 --> 2.8
MaxSpeed: 2.5 --> 2.8
MaxSpeedReverse: 0 --> 2.8
TurnRate: 20 --> 60 (Tech 3 Sonars have 100)


MrRowey:
> This seems fine to me; it does feel a bit odd that the Seraphim sonar is so fast while it is underwater. It's not exactly aerodynamic in the end, like the submarines

I think speed was a balance decision as the OG didn't move

Garanas:
oh, maybe because I spawned it in submerged the speed multiplier doesn't apply

Garanas:
That was it, unit works fine 😃 

https://github.com/FAForever/fa/assets/15778155/a559e4f7-c5d0-4021-84a0-3d5019e961f6

PR #5964, opened 2024-03-11
Evaluate and rework the categories of cybran naval units

apollodeathstar:
WEAKANTIAIR added


Garanas:
I don't think frigates should have the `WEAKANTIAIR` categories simply because there is no dedicated anti air unit for tech 1 naval - they are the strongest (and only) anti air unit for that tech and layer

apollodeathstar:
updated

Garanas:
Oh, that is interesting. I was not aware it also did not have the `ANTIAIR` category either 😃

Garanas:
@relent0r would this be a problem for the AI, if this unit suddenly has the `ANTIAIR` category?

relent0r:
> @relent0r would this be a problem for the AI, if this unit suddenly has the `ANTIAIR` category?

Nah no impact, the naval side of the default AI hasn't changed so it won't be doing anything around categories.

With regards to my AI, I use the threat values from the blueprint defense table to know if a naval platoon has any capability there rather than categories.

Though on the flip side, it would probably be a good thing for the new factory manager we were putting together since I imagine naval factories would think they have no aa production capability unless they were aeon.

PR #5971, opened 2024-03-13
Add disconnection share and ACU recall options

lL1l1:
It can be game-ruining to lose players because of disconnects, blowing up all the engineers in their base or units at their front and destroying all their units in share until death games. Especially because of recent connection issues/inherent Supcom connection issues. This PR fixes that with two new settings:
- DC Share Conditions: Determines how units are shared in case of a disconnect. To prevent abuse in Assassination, one ACU for that player must be at full HP and shields for the disconnect share conditions to apply. Default: Same as share condition
Common use case: share until death players could use fullshare for unfortunate disconnects
- Recall Disconnected ACUs: Instead of exploding, an ACU can recall and leave everything intact. To prevent abuse, the ACU must be at full HP and shields. This works individually on all ACUs a player owns. Default: Off
Common use case: could be used in team matchmaker to prevent games from being completely lost due to an early disconnect.

The full HP/shields threshold can be adjusted rather easily. Another idea is to not have enemy combat units within a large range. This would really ensure that the ACU is out of danger and the player isn't just abusing the feature.
Maybe the Recall Disconnected ACUs feature could use a time limit after which the rule goes away?
There is also an idea to leave the ACU around on disconnect, and apply fullshare rules until the ACU is killed, and applying the share condition rules then.


Garanas:
What if instead of checking whether the unit is full health, we instead check if the unit received damage in the last 10 seconds?

We could keep track (for all units) in what game tick that unit received damage for the last time.

lL1l1:
Doesn't updating a value every time any unit takes damage just to check it once at the end of the game sound expensive? I'm not experienced, so this is more of a curiosity/learning question. I do agree it's better for ACUs, since regen/shields varies a lot between factions, and the feature shouldn't depend on faction or upgrade choice.

Garanas:
> Doesn't updating a value every time any unit takes damage just to check it once at the end of the game sound expensive?

Yes, as long as the complexity is O(1) then we're fine. There's a lot that happens when units take damage (health adjustments, veterancy) and that runs fine at the moment.

We could also use the same value to 'enhance' the CTRL + K ability - if a unit did not receive damage the past 10 seconds then it can be destroyed without triggering the death weapon for example.

Note that the '10 seconds' is arbitrarily chosen, can be any value.

apollodeathstar:
I feel the easiest choice is to just have the ACU be handed over along with the other units, if it was going to die it would still die anyway + more likely due the the "stall" that happens to units during the handover. This prevents abuse and a 99% health ACU getting shot by a single enemy tank on DC exploding and leading to the same outcome as now (which we are trying to avoid)

I would say either hand the ACU over or make the ACU neutral (civilian) and give it a "recall" CD timer of 30 seconds to 1 min after which it recalls off the field. (despawns)

lL1l1:
> I feel the easiest choice is to just have the ACU be handed over along with the other units

> they are still being punished for the DC'ed players DC if they do not get it via the loss in combat power.

In a lot of maps and at increasing rating the shared economy and units can be more beneficial than the control or commentary from another player, enough to overcome the loss of an ACU. It has also been brought up on Discord that sharing ACUs would be extremely powerful (two gun coms perfectly coordinated).

> make the ACU neutral (civilian)

This relies on the map having a civilian army, and the fallback is killing (exploding) the ACU according to the civilian desertion share condition.

> Imagine if you did not get half the Factories or mexs from your allies because they DC'ed. The ACU early game is a small army onto itself.

This does remind me that the ACU has a large resource storage in the beginning of the game.

How about when a player disconnects, share their ACU + army and apply the share condition 5 minutes after the game started and 2 minutes after disconnect, or if the disconnected ACU dies early? The 5 minutes after start sharing allows use of the initial resources.  The 2 minutes after disconnect lets the ACU explode due to natural game rules instead of using an ACU recall option, and gives time to move units away, use the ACU one last time, and get tech to stabilize vs enemy ACU.
I think this would be the most balanced way to avoid abuse while preserving benefits. In assassination fullshare, which is the matchmaker mode, it would be most balanced. In assassination partial/no share, it wouldn't be fair because of the unit loss, but those are custom games that can rehost.
Could add it as an option instead of the ACU recall setting. Maybe like: "ACU Share: Explode, If-safe Recall, Delayed Recall, Permanent". One of them can be set as ranked if it's considered worthy. Permanent share might as well unlock giving your ACU to allies too.

apollodeathstar:
I think having it share for something like 2 mins after disconnect is a good idea, after which it should just recall or explode. that is enough time for the player to compensate for the missing teammate and adjust their play. 

All I really care about doing is avoiding punishment for players in the early game who have a ally drop/dc and them suffering badly for it.  

Also the 2 mins timer means players will be unlikely to waste mass upgrading the ACU. I think that is a good way to handle it.

lL1l1:
> We could also use the same value to 'enhance' the CTRL + K ability - if a unit did not receive damage the past 10 seconds then it can be destroyed without triggering the death weapon for example.

I think this is too overpowered for reclaiming volatile buildings, but it could be nice to prevent (possibly accidental) griefing with ACUs: if the ACU hasn't take damage recently and there are no enemies near the explosion range, it wouldn't detonate. After all, there's 0 benefit to the ACU dying without dealing damage because it leaves no wreckage.

lL1l1:
The ACU sharing mechanic is functional but it still needs proper UI with LOC tags and a better "received units from [player]" message that tells you the time remaining for delayed recall and the share condition after disconnect for delayed recall/permanent ACU share.

clyfordv:
Could:
1. Put a timer on the transferred ACU, after which it becomes un-selectable.
2. Similarly, start a timer when the receiving player gives the ACU an order, then make it un-selectable after that timer expires.
3. As an addendum/combined with the above, give the ACU a move order back to its spawn point when it's transferred.

lL1l1:
> Put a timer on the transferred ACU

I did that. I'm writing a new announcement box so that it can 2 lines of text but I'm open to better ideas to show how long that ACU has left. Maybe a new tab or a timer in the avatars tab.

> after which it becomes un-selectable

Too out of the ordinary for how multiplayer skirmishes work, and it takes away control from the player for a key unit. It does make me wonder how the select/zoom to ACU hotkeys will work/break with multiple ACUs.

lua/sim/units/ACUUnit.lua
```diff
@@ -59,6 +62,19 @@ ACUUnit = ClassUnit(CommandUnit) {
...
     end,
 
+    ---@param self ACUUnit
+    ---@param instigator Unit
+    ---@param amount number
+    ---@param vector Vector
+    ---@param damageType DamageType
+    OnDamage = function(self, instigator, amount, vector, damageType)
+        if self.CanTakeDamage and damageType ~= "TreeForce" and damageType ~= "TreeFire" then
+            self.LastTickDamaged = GetGameTick()
```
Garanas, line 72: Won't this reference to `LastTickDamaged` be shared among all classes inheriting from this class? Because the value exists at the scope of the meta table. Not at the scope of the instance.

lua/ui/game/announcement.lua
```diff
@@ -125,62 +165,49 @@ function CreateAnnouncement(text, goalControl, secondaryText, onFinished)
...
+    Layouter(tl):TopLeftOf(parent):End()
+    Layouter(tm):CenteredAbove(parent):FillHorizontally(parent):End()
+    Layouter(tr):TopRightOf(parent):End()
+    Layouter(ml):CenteredLeftOf(parent):FillVertically(parent):End()
+    Layouter(mr):CenteredRightOf(parent):FillVertically(parent):End()
+    Layouter(bl):BottomLeftOf(parent):End()
+    Layouter(bm):CenteredBelow(parent):FillHorizontally(parent):End()
+    Layouter(br):BottomRightOf(parent):End()
+
+    return { tl, tm, tr, ml, mr, bl, bm, br }
```
Garanas, line 212: This is a different type of object then the original. It changes the interface. The original has hash-based references. Your implementation has index-based references. Are you sure that you can make this change without breaking UI mods?

Garanas (commented):
Some quick notes when purely reviewing the code.

lua/ui/game/announcement.lua
lL1l1: I didn't notice I changed the type, I think I just wanted to minimize table operations by creating the table as the last step. I'll change it to the original hash format because this function has been global in the file since GPG so anyone could've imported it just to create its specific border.

lua/sim/units/ACUUnit.lua
lL1l1: It won't because LastTickDamaged is a number and not a table. Tables would get shared at the meta table scope because the class system doesn't deep copy them, as we ran into when discussing enhancements.

lua/SimUtils.lua
```diff
@@ -62,15 +34,18 @@ local function TransferUnitsOwnershipDelayedWeapons(weapon)
...
 ---@param toArmy number
----@param captured boolean?
+---@param captured? boolean
+---@param noRestrictions? boolean
 ---@return Unit[]?
-function TransferUnitsOwnership(units, toArmy, captured)
+function TransferUnitsOwnership(units, toArmy, captured, noRestrictions)
     local toBrain = GetArmyBrain(toArmy)
-    if not toBrain or toBrain:IsDefeated() or not units or table.empty(units) then
+    if not toBrain or (not noRestrictions and toBrain:IsDefeated()) or not units or table.empty(units) then
```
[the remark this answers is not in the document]
lL1l1: How about multiple lines like this?
```lua
if not toBrain or (toBrain:IsDefeated() and not noRestrictions)
    or table.empty(units)
then
    return
end
```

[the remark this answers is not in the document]
lL1l1: Done in 6baa71b

Garanas:
Together with @lL1l1 we decided to merge the technical aspect of this pull request. The new functionality is not merged in. It needs some further discussion about what the behavior should be!

Garanas (approved):

PR #5977, opened 2024-03-14
Allow Selen to fire from transports

lL1l1:
image

1e93266 makes this possible compared to vanilla, because the Selen can fire in all directions in FAF.


MrRowey:
Let me test this and will forward to balance team

MrRowey:
Balance Team does not want to add the ability for selen's to shoot from transports.

PR #5978, opened 2024-03-14
Correctly make Ythotha turn towards its target

lL1l1:
Reverts #5916 because AttackAngle caused Ythotha to constantly rotate, which looked ugly. Instead, uses a reduced `SlavedToBodyArcRange` to turn towards the enemy for all weapons to fire, while not constantly turning to maintain an exact angle. 32 arc range instead of 45 to give weapons plenty of space to lead their target.
AttackAngle is more for weapons that can't all fire forwards, like Fatboy and battleships, so using it for Ythotha was incorrect in the first place.


lL1l1:
The constant walk animation triggering also makes the gun lose its aim and stop firing.

Garanas:
@relent0r this may also be related to what you reported about the Ythotha in the AI Discord.

@lL1l1 Great work, this is a lot better and fixes the problem 😃 !

PR #5990, opened 2024-03-17
Update XES0102_unit.bp "ANTITORPEDO", added to categories 

apollodeathstar:
"ANTITORPEDO", added to categories

### DIFF
--- a/units/XES0102/XES0102_unit.bp
+++ b/units/XES0102/XES0102_unit.bp
@@ -11,6 +11,7 @@ UnitBlueprint{
     BuildIconSortPriority = 20,
     Categories = {
         "ANTINAVY",
+        "ANTITORPEDO",
         "BUILTBYEXPERIMENTALSUB",
         "BUILTBYTIER2FACTORY",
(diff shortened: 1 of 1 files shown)


Garanas:
@apollodeathstar please remove the bits and pieces of the template that you do not use 😃 !

apollodeathstar:
> @apollodeathstar please remove the bits and pieces of the template that you do not use 😃 !
> 
> edit: the same applies to #5991 and #5992

I updated it, but I will say originally I was trying to use the template, but its a bit confusing as to where the submitters information is supposed to go. Can you update it with clear "write here/replace the text here" areas for each section. I tried 3 different methods and got it right only once.

lL1l1 (approved):
I didn't find any mentions of the ANTITORPEDO unit categories in the repository, M28AI, M27AI, RNGAI, or AI-Uveso. The category changes should just improve the consistent application of the category.
There are many other units with antitorps that need the category too.

lL1l1:
I cherry picked this to #6001 so that the changes are more consolidated.

PR #5992, opened 2024-03-17
Update UES0201_unit.bp   "ANTITORPEDO", added to Categories

apollodeathstar:
"ANTITORPEDO", added to Categories

PR #5991, opened 2024-03-17
Update XES0307_unit.bp     "ANTITORPEDO", added to Categories

apollodeathstar:
"ANTITORPEDO", added to Categories

## Checklist

- [ ] Changes are annotated, including comments where useful
- [ ] Changes are documented in the changelog for the next game version

PR #6001, opened 2024-03-18
Add `ANTITORPEDO` category to all units with antitorp weapons

lL1l1:
Adds `ANTITORPEDO` to units with antitorpedo weapons. Cherry picks #5990 #5991 #5992 to consolidate into one PR.
The regex string used to find antitorp weapons was: `ProjectileId ?= ?.*Antitorpedo`.

I didn't find any mentions of the ANTITORPEDO unit categories in the repository (only mentioned for projectile categories), M28AI, M27AI, RNGAI, or AI-Uveso. The category changes should just improve the consistent application of the category, making it useful for AI developers. For example, `M28UnitInfo.lua` uses
```lua
if oCurWeapon.RangeCategory == 'UWRC_Countermeasure' then
  --Target restriction should catch all cases, have but the most common labels as redundancy as well (doesnt cover mega? and fatboy? torpedo defence which have antitorpedoleft, left1, left2, right, etc. labels)
  if oCurWeapon.TargetRestrictOnlyAllow == "TORPEDO" or oCurWeapon.Label == 'AntiTorpedo' or oCurWeapon.Label == 'AntiTorpedo01' or oCurWeapon.Label == 'AntiTorpedoF' then
    oUnit[refbHasTorpedoDefence] = true
  else
```
which can be simplified and improved thanks to theses category changes. @maudlin27 

## Checklist

The category is already annotated in `categories.lua`.

### DIFF
--- a/units/UAS0103/UAS0103_unit.bp
+++ b/units/UAS0103/UAS0103_unit.bp
@@ -15,6 +15,7 @@ UnitBlueprint{
     BuildIconSortPriority = 40,
     Categories = {
         "AEON",
+        "ANTITORPEDO",
         "BUILTBYEXPERIMENTALSUB",
         "BUILTBYTIER1FACTORY",
         "BUILTBYTIER2FACTORY",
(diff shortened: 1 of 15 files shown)


Garanas (approved):
This looks great 👍 , I don't think Maudlin can fully drop the requirements because mods may be as or more inconsistent as we were up until now

PR #6004, opened 2024-03-19
add antinavy category to CZAR UAA0310

### LINKED ISSUE #5988: CZAR has no ANTINAVY category
**Describe the bug**
I was troubleshooting why the AI wouldn't detect that it could attack naval units with the depth charges and noticed there is no ANTINAVY category on it.

**To Reproduce**
Check categories in blueprint for CZAR

**Expected behavior**
Contains ANTINAVY category

**Additional context**
I'm unsure if its intentional to not have it with the category.

relent0r:

This PR adds the ANTINAVY category to the CZAR to match its sub threat capabilities.

Validation of not causing obvious issues

## Additional context

@Garanas I'm not sure if this can cause issues with existing formation logic etc. The default AI already excluded experimentals for its air based antinavy platoon templates so there are no issues there. 

fixes #5988

## Additional Information

This issue came about when creating an improved state machines for air based experimentals, it was found that the czar could not detect its anti naval capabilities using unit categories when flying over water so it could not be determined to attack naval units should it want to.

### DIFF
--- a/units/UAA0310/UAA0310_unit.bp
+++ b/units/UAA0310/UAA0310_unit.bp
@@ -51,6 +51,7 @@ UnitBlueprint{
         "AIR",
         "AIRSTAGINGPLATFORM",
         "ANTIAIR",
+        "ANTINAVY",
         "BUILTBYTIER3COMMANDER",
         "BUILTBYTIER3ENGINEER",
         "CANNOTUSEAIRSTAGING",


lL1l1:
According to discussion in #3952 and related pulls, using the category this way is incorrect, and you should use `OVERLAYANTINAVY` to identify the unit.
According to https://github.com/FAForever/fa/blob/2e38df98fd5207df678ad8ae42c37c5a8b1de2db/lua/system/blueprints-units.lua#L425-L500 you should also have `WEAKANTINAVY` available.
I think the correct solution for the issue would be to add the weak categories to all blueprints in the base game. The automatic weak categories can be for modded units.

relent0r:
> According to discussion in #3952 and related pulls, using the category this way is incorrect, and you should use `OVERLAYANTINAVY` to identify the unit. According to
> 
> https://github.com/FAForever/fa/blob/2e38df98fd5207df678ad8ae42c37c5a8b1de2db/lua/system/blueprints-units.lua#L425-L500
> you should also have `WEAKANTINAVY` available.
> I think the correct solution for the issue would be to add the weak categories to all blueprints in the base game. The automatic weak categories can be for modded units.

I see what you mean. I thought OVERLAY categories were related to UI stuff. I'm easy with whatever as long as I can query that its capable of antinavy without looping through all the weapons.

@Garanas If your in agreement I can change to whatever.

lL1l1:
The overlay categories are related to the UI in terms of loyalist, monkeylord, othuum, and sera sniper who have secondary direct fire weapons displayed by the indirect fire range overlay. I think the WEAK categories should be used for this.

Garanas:
You're right, my mistake. The anti navy capabilities are not the primary capabilities of the CZAR. Hence, it should have `WEAKANTINAVY`. The `OVERLAYANTINAVY` is used to create the overlay, see also [rangeoverlayparams.lua](https://github.com/FAForever/fa/blob/deploy/fafdevelop/lua/ui/game/rangeoverlayparams.lua#L80) and does not help determine whether the unit is good it.

PR #6005, opened 2024-03-19
Remove the 'BUILTBYEXPERIMENTALSUB' category from various units

Basilisk3:

Removes the `BUILTBYEXPERIMENTALSUB` category from various units.

The Tempest can still build these units just fine. Except units like the Seraphim cruiser of course, interestingly enough, this unit also had the category.

## Additional context

According to @The-Balthazar on Discord, this category is unused. 

## Checklist

### DIFF
--- a/changelog/3804.md
+++ b/changelog/3804.md
@@ -30,7 +30,7 @@
 
 - (#6004) Add missing categories to the CZAR
 
-- (#6004) Remove the `BUILTBYEXPERIMENTALSUB` category from all units that have it. This category was unused, the Tempest is not affected by this change in any way.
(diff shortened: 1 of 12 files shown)


The-Balthazar:
If the Tempest ever used `BUILTBYEXPERIMENTALSUB`, it was in the original beta. In SupCom1 it's got the same build categories as it does in FA. `BUILTBYEXPERIMENTALSUB` exists in a lot of mods, but because they uncritically copied the blueprints of the units this PR removes it from, not because they planned to use it; same reason non-Aeon FA units appear to have it, at least as far as my searches have gone.

PR #6006, opened 2024-03-19
Tweaks to the categories of various blueprints

Basilisk3:
- Add the `ANTIMISSILE` category to the Iavish (Seraphim Tech 3 Aircraft Carrier).
- Removed the `MASSFABRICATION` category from all ACUs, SACUs and the Paragon. They retain the `MASSPRODUCTION` category.

Is there any reason why SACUs might need the `MASSFABRICATION` category?

Edit: Apparently this category is not vital for these units.

## Checklist


Garanas:
Perhaps they originally were mobile mass fabricators. There have been numerous 'redesign' left overs in the code base 😃

Garanas:
> * Removed the `SNIPER` category from the Megalith.

This is an interesting one - it is actively used by AIs to identify the Megalith as an experimental that can act like a sniper. If I'm not mistaken then @maudlin27 does this

Basilisk3:
> > * Removed the `SNIPER` category from the Megalith.
> 
> This is an interesting one - it is actively used by AIs to identify the Megalith as an experimental that can act like a sniper. If I'm not mistaken then @maudlin27 does this

I see. If this causes trouble for the AI, then I will revert the change.

lL1l1:
The `SNIPER` category also makes t2 artillery prefer to target Megalith over lower range experimentals.

https://github.com/FAForever/fa/blob/31772f0d4dbac5ffdb79a4fdb3aaaa880c6a9f4c/units/UEB2303/UEB2303_unit.bp#L199-L217

PR #6007, opened 2024-03-19
Remove Siren `REPAIR` category

Hdt80bro:
Currently, the Siren Class cruiser has the `REPAIR` category - presumably because it's an air staging platform that can repair aircraft. What this actually does is let it repair other units like an engineer, despite having no visual indication or defined build rate - it defaults to 1 and is incredibly slow. If this was intended, it would have been more clear with an appreciable build rate and effect (maybe even an animation, unit description line, or unit ability), but given the context, it appears to be a simple oversight.

image
*Notice the orange progress bar and 1 power drain - the damaged cruiser used to be at 30% health but is now at 32%*

### DIFF
--- a/changelog/3804.md
+++ b/changelog/3804.md
@@ -4,6 +4,8 @@
 
 - (#6002) Fix a bug where the map would not reveal at the end of the game
 
+- (#6007) Fix Siren uselessly trying to repair units when assisting them
+
 ## Balance
 
 <!-- Remove header when empty -->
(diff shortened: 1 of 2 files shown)


lL1l1 (approved):
That's an old screenshot; units with an undefined buildrate are postprocessed to 0.

https://github.com/FAForever/fa/blob/2e38df98fd5207df678ad8ae42c37c5a8b1de2db/lua/system/blueprints-units.lua#L537-L541

Even with 0 buildrate the Siren still tries to repair, not moving into formation and displaying the work progress bar, so it's good to remove the category for that reason too.

PR #6009, opened 2024-03-20
Adjust 'BOMB' category usage and add torpedo/normal bomber selection hotkeys

### LINKED ISSUE #5298: Add two new hotkeys for different type of bombers
- Select all bombers excluding torpedo bombers
- Select all torpedo bombers

When using "Select All Bombers" I have to manually select one of them to use it for land or for sea.

lL1l1:
Closes #5298. Reduces the need to manually filter after using the select all bombers hotkey. Could be done with game.prefs editing instead, but it seems useful enough to be a default available to all players.
`BOMB` category and Suicide Weapon ability text removed from Loyalist because it no longer has the self-destructing charge ability.
`BOMB` category added to Mercy because it has to die to attack, just like the fire beetle.

Spawned all torpedo bombers and all bombers + mercy. Pressing the select all bombers (non-torpedo) hotkey selects all normal bombers. Pressing the select all bombers (torpedo) hotkey selects all torpedo bombers. Mercy does not get selected.

## Checklist


relent0r:
Nice change on the Mercy. Until now if I want to capture all MOBILE AIR BOMBER I've always had to exclude the Mercys unit id since from a category perspective you can't tell the difference.

Garanas:
@lL1l1 can you add the category and descriptions for the new hotkeys to the changelog? That way players do not have to go searching for them

lL1l1:
One thing I noticed is that this would make it more difficult to identify Loyalist's death weapon usage since there's no category for it anymore. `BOMB` is for suicide weapons, and `VOLATILE` is for units who's death effects should be stopped by reclaiming (this doesn't work though). `VOLATILE` category is also a separate definition from the ability text displayed in the UI, which just means a unit has a death weapon.

PR #6010, opened 2024-03-20
Re-add hotkey label for attack move

### LINKED ISSUE #5597: There is no hotkey label for the AttackMove order
There is no hotkey label for Attack move, even though it should be RMB.

Related PR from 2017: #1859

Here, immutableOrderKeys wants to apply a fix hotkey label for the AttackMove order on the bottom left of the ui. For example, the label for patrol changes as you change your hotkey for it. But since Attack move is always Alt+RMB immutableOrderKeys tries to label it as "RMB" permanently.

lL1l1:
Closes #5597. Alt + RMB is a special case of a keybind that needs to be manually assigned a text and color in the hotkey labels. This was removed in a refactoring a long time ago 545f16a, and this re-adds that.
image

## Checklist


lua/keymap/hotkeylabels.lua
```diff
@@ -113,6 +113,12 @@ function getKeyTables()
     local upgradeKey = false
     local orderKeys = {}
```
Garanas, line 114: I'm not too familiar with this file, but you can add it to the definition here too. The key `attack_move` is oddly specific, where does that reference to?

lL1l1: It's the same as in the `helpText` in this table of `OrderInfo`s

https://github.com/FAForever/fa/blob/31772f0d4dbac5ffdb79a4fdb3aaaa880c6a9f4c/lua/ui/game/orders.lua#L1133-L1174

because the label is retrieved with that as the key:
https://github.com/FAForever/fa/blob/31772f0d4dbac5ffdb79a4fdb3aaaa880c6a9f4c/lua/ui/game/orders.lua#L1323-L1326

PR #6012, opened 2024-03-20
CZAR categories - remove antinavy and add weakantinavy

relent0r:

This PR changes the previously added category of ANTINAVY to WEAKANTINAVY.

After some discussion I realised that non primary weapon types for units should have the weak variant of the category.

### DIFF
--- a/changelog/3804.md
+++ b/changelog/3804.md
@@ -28,7 +28,7 @@
 
 - (#6001) Add missing categories to units with Anti-Torpedo defenses
 
-- (#6004) Add missing categories to the CZAR
+- (#6012) Add missing categories to the CZAR
 
 - (#6008) Remove unecessary category from the Brick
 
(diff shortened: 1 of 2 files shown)


Garanas:
Relates to #5988

Garanas:
@relent0r could you add and/or update the entry in the changelog of the previous pull request? It is inside `/changelogs/3804.md`

relent0r:
> @relent0r could you add and/or update the entry in the changelog of the previous pull request? It is inside `/changelogs/3804.md`

You've just got it listed as Add missing categories to the CZAR, so it still applies the same. (tho didn't realise we were updating the change log when doing PRs so good to know)

Garanas:
That is very recent - since a week. It's an experiment to see if we can build up the changelog as we make the changes instead of me spending several hours crunching the notes together right before a release 😃

PR #6013, opened 2024-03-20
Fix exploit with the `Copy/Distribute orders` feature

Garanas:

One can copy the command queue of another unit and append it to the command queue of the units in your selection. In Discord people noticed that this effectively allowed you to copy attack commands that are assigned to units that you no longer have vision over. 

That exploit is now fixed.

image

Have a group of units constantly go in and out of (radar) vision. The orders should only be copied when the attack commands are visible to the player.

## Additional context

See the following [Discord thread](https://discord.com/channels/197033481883222026/1220006738552684697)

## Checklist


Pokute:
Does this affect only order copying/distributing or does this affect issuing attack commands in general?

If latter, then this will very likely affect negatively AI/single player/scenarios.

Garanas:
This only affects features that rely on the mapping provided in `lua/sim/commands/shared.lua`, which is only used by the copy/distribute queue orders if I'm not mistaken 😃

PR #6015, opened 2024-03-21
Fix a bug and extend the context based templates feature

Garanas:

Related to [this bug report on Discord](https://discord.com/channels/197033481883222026/1219563479581720577). The bug was that a tech 1 engineer can not create the storage template when trying to cap a tech 2 extractor because the tech 2 extractor would block the template as a tech 1 engineer can't build it. We fix that by trying to build the unit it upgrades from, or by trying to build the unit it originates from.

Next to that we extend the feature by allowing it to act like a color picker - if we're hovering over a unit or a build order and we can't find matching templates then we'll try to enter command mode with a build order for the given unit and/or blueprint order. This is inspired by the featured of @4z0t mentioned on the [forums](https://forum.faforever.com/topic/7351/ctrl-ui-mod).

https://github.com/FAForever/fa/assets/15778155/1b3270b0-dc49-4626-9e9e-b8f6d2efaab9

Spawning in structures and triggering the hotkey in various conditions.

## Checklist

### DIFF
--- a/changelog/3804.md
+++ b/changelog/3804.md
@@ -6,13 +6,17 @@
 
 - (#6007) Fix Siren uselessly trying to repair units when assisting them
 
+- (#6015) Fix a bug with capping extractors via the context-based templates feature
+
 ## Balance
 
 <!-- Remove header when empty -->
 
 ## Features
 
-<!-- Remove header when empty -->
+- (#6015) Extend the context-based templates feature
+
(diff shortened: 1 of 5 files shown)


Garanas:
With these changes you can do the following:

https://github.com/FAForever/fa/assets/15778155/b6773892-fb72-45b0-a1e5-cebfa79d1a09

PR #6021, opened 2024-03-27
Sync playable area in `UIScenarioInfo` table

### LINKED ISSUE #6019: Reticle uses the wrong fields to determine the map size
**Describe the bug**

The following snippet uses `PlayableAreaWidth` and `PlayableAreaHeight`:

https://github.com/FAForever/fa/blob/fc1fb38f5809a40025e5d910d9518bd62a41cfa0/lua/ui/controls/reticle.lua#L53-L55

Which is defined here:

https://github.com/FAForever/fa/blob/fc1fb38f5809a40025e5d910d9518bd62a41cfa0/lua/ui/game/score.lua#L522-L533

But that code only runs when the user uses the default scoreboard. Which is not the case for the majority. If it is not available, the script crashes when updating the costs.

**To Reproduce**

Start a game with an modded scoreboard.

lL1l1:

Fixes #6019. Also fixes the `UIReticle` class calculating on-map status out from the top left corner instead of inside the actual map rectangle.
Instead of having the scoreboard write the playable area info to the UI scenario info table, this is done in `userSync.lua`, making that playable area information easily available for anything to use.

Turned off all UI mods. Checked the scoreboard for the correct map size and then upgraded teleport and checked the map boundaries on the following maps: 
Map Prefix | `_Scenario.lua` `size =` | `_Save.lua` 'AREA_1' Rectangle | Notes
-|-|-|-
astrogap_fixed|`{1024,1024}`|`(0,0,1024,1024)`| Typical setup for modern user-made maps because the map editor puts in the area rectangle
Abrogation|`{1024,1024}`|`(128, 128, 896, 896)` | This map's playable area is offset from the top left corner, showing how the Reticle's on-map status used to be broken if you mouse over the bottom right part of the map. Typical setup for mapgen maps and other maps that don't have 20x20/10x10/5x5 sizes.
scmp_025|`{512, 512}`|`nil`| Map name is "Saltrock Colony", map_version 2. This older setup without a playable area is the reason for the `SimInit.lua` changes where we make sure to sync a new playable area at the start of the game.

After that, turned back on my UI mods and checked that my scoreboard was still displaying the correct map size; old UI mods using the old `score.lua`'s method of setting the PlayableAreaWidth/Height should still work.

## Checklist


Garanas:
That is interesting, usually these engine calls return a unique table on each call. But for this call the table is re-used

PR #4584, opened 2023-01-11
Integrate wonky structures

Garanas:
See also: https://github.com/FAForever/fa/pull/4507

When a structure is build it no longer generates a flat plane but a gradient between terrain heights at the outer tips of the build skirt. The structure is then slightly rotated to match the orientation of the terrain. This applies to all structure. On top of that Extractors and Hydrocarbon plants can now always be built.

With thanks to Balthazar for the Wonky Structures mod and the gradient terrain 'flattening'.

Various examples that would previously deform the terrain to the point where it seriously impacts gameplay:

image

image

image

PR #6022, opened 2024-03-28
Fix Seraphim Air Factories being harder to place than other factories

### LINKED ISSUE #5460: Seraphim air factories are harder to place than other air factories when terrain deformations are present
Seraphim air factories are considerably more picky when it comes to finding a valid place to build them. I only noticed this recently, perhaps this bug was introduced by the changes to terrain deformation through buildings.

**To Reproduce**
1. Go to the southern air slot on Seton's Clutch
2. Locate the topmost mex of the 4 core mexes (the one where the first air factory is usually built)
3. Drag build 3 air factories adjacently to the topmost mex (drag the buildings upwards)

You will see that all factions can build three air factories, except for the Seraphim, which can only build two.

**Expected behavior**
Seraphim air factories should behave like any other air factory.

Basilisk3:

`FlattenSkirt` needs to be set to `true`, as it is for all other Air Factories. Resolves #5460.

## Additinal Context

#4584 set this value to `false`.

## Checklist

### DIFF
--- a/changelog/3806.md
+++ b/changelog/3806.md
@@ -2,7 +2,7 @@
 
 ## Bug Fixes
 
-<!-- Remove header when empty -->
+- (#6022) Fix a bug, which made Seraphim Air Factories harder to place than other Air Factories.
 
 ## Balance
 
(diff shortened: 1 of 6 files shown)


Garanas:
@Hdt80bro or @4z0t could you look into assembly why this particular field impacts the ability to place the structure? It feels to me that it should not

PR #6024, opened 2024-03-30
Introduce `Deathnuke` armor for ASFs

Basilisk3:

All ASFs: `Deathnuke` armor introduced (ASFs take 25% less damage from ACU explosions)

## Additional context
Due to #5465, bombing ASFs with ACU explosions has become even more powerful, as they are now more expensive but remain under the threshold of 2500 hit points to survive the explosion.

## Checklist

### DIFF
--- a/changelog/3806.md
+++ b/changelog/3806.md
@@ -2,7 +2,7 @@
 
 ## Bug Fixes
 
-<!-- Remove header when empty -->
+- (#6022) Fix a bug, which made Seraphim Air Factories harder to place than other Air Factories.
 
 ## Balance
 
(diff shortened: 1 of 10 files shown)


Garanas:
Mentioned it to the balance team 👍

Garanas:
I'm a bit confused, given that damage of type `normal` has a value of 1.0. Doesn't a value of 0.75 mean that 75% of the damage is applied, instead of 75% resisted?

Basilisk3:
I improved my wording to clear up any confusion. They are supposed to take 25% less damage.

lL1l1:
This will need a rebase before merging.

PR #6028, opened 2024-03-31
Fix wrong splash radius decal appearing for certain units

lL1l1:

This decal is the one you get in when in groundfire mode or mousing over an enemy. Previously, only the first valid weapon would be used to display splash radius. This meant that some units wouldn't display their splash radius at all, or displayed a smaller radius than expected. Now all valid weapons are used to determine the decal's radius.
Overcharge weapons are added to the filter because this makes the overcharge radius show up for ACU/SACU constantly, which is unnecessary with the overcharge cursor. It would only be useful for groundfiring reclaim with auto OC. The SACU one isn't even always available.
Enhancement enabled weapons are added to the filter because they aren't always available, and it would show up as a 0.5 radius circle for Cybran ACU's laser.

The next improvement for this cursor could be checking unit enhancements, as there are many damage radius enhancements that would be nice to display.
OC could also be unfiltered when auto OC is enabled and storage is available.

Checked units with a changed displayed radius:
- Ythotha: 4 (EyeWeapon) -> 6 (RightArm)
- Mongoose: 0 (GatlingCannon) -> 2 (Grenade)
- Soul Ripper: 0 (Missile01) -> 3 (HeavyBolter)
- ACU/SACU: 0 (the main gun) ~~-> 2.5 (AutoOverCharge)~~

<details><summary>Console command to find these units</summary>

```lua
UI_Lua 
for i,v in __blueprints do
    if type(i) == 'number' or not v.CategoriesHash["SHOWATTACKRETICLE"] then continue end
    local weapons = v.Weapon
    if weapons then
        local once = true
        local fweps = {}

        for i, w in weapons do
            if w.ManualFire == false and w.WeaponCategory ~= 'Teleport' and w.WeaponCategory ~= 'Death' and w.WeaponCategory ~= 'Anti Air' and w.DamageType ~= 'Overcharge' and not w.EnabledByEnhancement then 
                table.insert(fweps, w)
            end
        end
        if table.getn(fweps) > 1 then
            for i, w in fweps do
                if once then 
                    LOG(string.format('%-7s %-12s %-20s %-20s', v.FactionCategory, v.TechCategory, LOC(v.Description), LOC(v.General.UnitName) or ''))
                    once = false
                end
                LOG(string.format('\t%-35s %10.1f', w.BlueprintId, w.DamageRadius))
            end
        end
    end
end
```

</details>

## Checklist


MrRowey (approved):

changelog/snippets/fix.6028.md
```diff
@@ -0,0 +1,3 @@
+- (#6028) Fix attack cursor splash radius decals appearing with the wrong size for units that display splash radius and have multiple weapons.
+
+In FAF, this affects the Ythotha, Mongoose, and Soul Ripper.
```
Garanas, line 3: What do you mean with `In FAF` ? Don't all these changes apply to FAF? 😄

lL1l1: I wanted to differentiate between modded units and vanilla units, but these changes also are only in FAF so that makes it nuanced.

PR #6033, opened 2024-03-31
Fix defensive and arty structure orientations on build.

Pokute:

Fix defensive structure orientations on build. Previously the orientation really worked right for pd1 and pd2.
* Now torpedo defenses, SAMs and TMDs are oriented towards nearest enemy or center. The Aeon TMD getting rotated like an idiot might clash with fragile senses of aesthetics.
* Previously t2 arty pointed only north or south towards closest enemy or center.
* Now all t2+ static arty points at center of map but limited to 90 degree steps in order to not affect pathing/terrain. This fixes the very long time for T3+ artys to rotate towards enemy depending on position. Previously for a newly built Mavor at north of the map, it already faces towards center while one built at the south of the map will require a minute for the Mavor to turn towards center.

image
image
image

Bug: HARMS still always points south after being built due to it's unique spawning code.

Built pds, tmds, static aa and arties on various positions of the map with some enemies to check how the buildings are placed.

## Checklist


lL1l1:
It's possible to spawn the unit's weapon(s) rotated towards the enemy by using `CAimManipulator:SetHeadingPitch`. I think that would look better than rotating the entire unit. The only issue is that if there are no enemies to target then the weapon will rotate towards the default heading/pitch. If that could be solved, it would be an interesting change on top of/replacing the unit rotation. Or you could apply it if you do find an enemy in range, since you're searching for them already anyway. i.e. rotate cardinally towards enemy, then rotate weapon directly towards enemy

Example (there is a visible enemy far away, so it doesn't rotate to default here):

https://github.com/FAForever/fa/assets/82986251/44ecd88b-e8e5-487b-bd40-37ed9b05b438

Inside `UEB2401_script.lua`'s `MainGun`
```lua
SetupTurret = function(self, bp)
    TIFArtilleryWeapon.SetupTurret(self, bp)
    local aimManip = self:GetAimManipulator()
    aimManip:SetHeadingPitch(90*math.pi/180, 0)
end,
```

Garanas:
I like the idea of the two-step approach: use the the 90 degrees angle for the rough approximation and fine-tune it by rotating the weapon (if applicable).

Basilisk3:
This looks promising👍

One thing to note: the `TurretYawSpeed` of TMDs is over five times greater than that of PDs, so they probably do not need to be turned in intervals of 45 degrees, if they need to be turned at all.

lL1l1:
With faster missiles the turn speed can matter, although it shouldn't be a significant issue after some recent speed reductions for TMLs.

Garanas:
Let's start small - these changes look promising and we can extend them further in a additional pull requests. I'll give it a review this week

Garanas:
This may sound strange, but in what situation would artillery not end up pointing to the center of the map even if we search for valid targets?

Why don't we just always orientate towards the center (in 90 degree steps or not)? It removes the need to search for things

lua/sim/units/StructureUnit.lua
```diff
@@ -41,8 +41,8 @@ local GetTarmac = import("/lua/tarmacs.lua").GetTarmacType
 
 -- compute once and store as upvalue for performance
 local StructureUnitRotateTowardsEnemiesLand = categories.STRUCTURE + categories.LAND + categories.NAVAL
-local StructureUnitRotateTowardsEnemiesArtillery = categories.ARTILLERY * (categories.TECH2 + categories.TECH3 + categories.EXPERIMENTAL)
-local StructureUnitOnStartBeingBuiltRotateBuildings = categories.STRUCTURE * (categories.DIRECTFIRE + categories.INDIRECTFIRE) * (categories.DEFENSE + (categories.ARTILLERY - (categories.TECH3 + categories.EXPERIMENTAL)))
+local StructureUnitOnStartBeingBuiltRotateBuildings = categories.STRUCTURE * ((categories.DIRECTFIRE + categories.INDIRECTFIRE + categories.ANTINAVY) + (categories.DEFENSE + categories.ARTILLERY))
```
lL1l1, line 44:
```suggestion
local StructureUnitOnStartBeingBuiltRotateBuildings = categories.STRUCTURE * (categories.DIRECTFIRE + categories.INDIRECTFIRE + categories.ANTINAVY + categories.DEFENSE + categories.ARTILLERY)
```

lua/sim/units/StructureUnit.lua
```diff
@@ -156,33 +156,36 @@ StructureUnit = ClassUnit(Unit) {
...
-                    radius = 1.1 * weapon.MaxRadius
+        -- find targets in range unless a static arty
+        if not EntityCategoryContains(StructureUnitRotateStaticArty, self) then
+            -- determine radius
+            local radius = 40
+            local weapons = self.Blueprint.Weapon
+            if weapons then
+                for k, weapon in weapons do
+                    if weapon.MaxRadius and weapon.MaxRadius > radius then
+                        radius = 1.1 * weapon.MaxRadius
```
lL1l1, line 167:
```suggestion
                    if weapon.MaxRadius and 1.1 * weapon.MaxRadius > radius then
                        radius = 1.1 * weapon.MaxRadius
```

lL1l1:
> This may sound strange, but in what situation would artillery not end up pointing to the center of the map even if we search for valid targets?
> 
> Why don't we just always orientate towards the center (in 90 degree steps or not)? It removes the need to search for things

The PR's code already does this for artillery. But answering the question,  naval maps have beaches that point towards the center of the map or are behind the base, so it doesn't always make sense to point towards the center for t2 arty at those beaches.

lua/sim/units/StructureUnit.lua
lL1l1: I'm not a fan of Volcano rotating.
```suggestion
local StructureUnitOnStartBeingBuiltRotateBuildings = categories.STRUCTURE * ((categories.DIRECTFIRE + categories.INDIRECTFIRE + categories.ANTINAVY) + (categories.DEFENSE - categories.uab4201 + categories.ARTILLERY))
```

lua/sim/units/StructureUnit.lua
```diff
@@ -206,8 +209,8 @@ StructureUnit = ClassUnit(Unit) {
         local rad = math.atan2(target.location[1] - pos[1], target.location[3] - pos[3])
         local degrees = rad * (180 / math.pi)
 
-        if EntityCategoryContains(StructureUnitRotateTowardsEnemiesArtillery, self) then
-            degrees = MathFloor((degrees + 90) / 180) * 180
+        if EntityCategoryContains(StructureUnitRotateStaticArty, self) then
+            degrees = MathFloor((degrees + 45) / 90) * 90 -- The static artys might have footprint that might affect pathing/terrain unless rotation is at 90 angle steps.
```
lL1l1, line 213: This would cause issues with modded structures that don't have square footprints. Usually the hitbox is inside the footprint no matter how it is rotated, but if the footprint isn't square then if you rotate the hitbox then it could easily stick outside the footprint.

lL1l1: This could be why it was north/south only before.
In addition, UEF's build effect is especially messed up for klinkhammers, but that should be addressed in a separate PR.

lL1l1:
8711699 HARMs now rotates towards the center of the map as well.

lua/sim/units/StructureUnit.lua
lL1l1: Units do not care about structure hitboxes it seems. Still, the footprint can't be rotated, so it makes sense to limit oblong footprints to 180 degree rotations.

lua/sim/units/StructureUnit.lua
```diff
@@ -207,15 +207,25 @@ StructureUnit = ClassUnit(Unit) {
...
 
-        local rotator = CreateRotator(self, 0, 'y', degrees, nil, nil)
-        rotator:SetPrecedence(1)
-        self.Trash:Add(rotator)
+        -- rotation quaternion {cos, 0, sin, 0}
+        local cos = math.cos(rad/2)
+        local sin = math.sin(rad/2)
+
+        -- quatRotate * quatOrient
+        local rhs1, rhs2, rhs3, rhs4 = unpack(self:GetOrientation())
```
Garanas, line 242: What happens when you gift a unit? Does it then get applied another time?

lL1l1: It doesn't get applied another time for `ChangeUnitArmy` but it does with `CreateUnitAtMouse` and `CreateUnitHPR`

lL1l1:
Old reasons to exempt arty from rotating:
- Pathing is handled by the unrotateable footprint, and structure hitboxes do not affect unit pathing.
- Terrain is handled separately by `self:FlattenSkirt` and `self:FlattenSkirtHorizontally`. The unit is just orientated onto that terrain after the fact.

Remaining reasons to exempt arty from rotating: 
- Adjacency visual effects
- Selection hitbox vs selection visual being especially ugly on t3/t4 arties.
- Doesn't fit UEF faction aesthetic, particularly on the large t3/t4 arty.

Since these reasons are more about aesthetics, should I change it to just checking the footprint size?
Going off footprint size I can also get rid of the arty entity category filtering entirely by checking min range and adding an overall max range for the target search radius.

Garanas (approved):
These are great improvements 👍

Garanas:
> Since these reasons are more about aesthetics, should I change it to just checking the footprint size?

I see where you're coming from. Let's look into that in the next pull request.

lua/sim/units/StructureUnit.lua
```diff
@@ -169,50 +182,73 @@ StructureUnit = ClassUnit(Unit) {
...
-        -- get direction vector, atanify it for angle
-        local rad = math.atan2(target.location[1] - pos[1], target.location[3] - pos[3])
-        local degrees = rad * (180 / math.pi)
+        -- get direction vector, atanify it for angle, then subtract our existing heading if we were spawned in angled
+        local rad = MathAtan2(target.location[1] - pos[1], target.location[3] - pos[3]) - self:GetHeading()
+
+        local stepSize
+        local bpFootprint = self.Blueprint.Footprint
+        if bpFootprint then
+            if bpFootprint.SizeX ~= bpFootprint.SizeZ then
```
Garanas, line 227: Are there (modded) units that match this criteria?

Garanas:
> Here's how it looks like when the artillery finishes and the weapon is rotated. I haven't found any way to rotate the weapon itself while under construction. I don't think the janky animation is worth it for the turn speed since changing pitch already takes longer than the max 45 degrees yaw change with the 90 degree orientations.

I agree, I think this is taking it too far. The goal of the feature is to make sure that the game is 'fair', not that your artillery can fire as quickly as possible.

PR #6040, opened 2024-04-01
Re-Open "Make veterancy requirements equal for presets and SACUs

lL1l1:

Re-opens #5730 except pointed at `fafdevelop` since `fafbeta` is no longer used, and this change wasn't explicitly denied, so I assume it got lost in that transition of no longer using `fafbeta`.

## Checklist

### DIFF
--- a/changelog/3807.md
+++ b/changelog/3807.md
@@ -6,7 +6,9 @@
 
 ## Balance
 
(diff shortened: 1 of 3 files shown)


MrRowey (approved):

MrRowey:
@lL1l1 please fix the conflict

PR #6048, opened 2024-04-04
AI - Increase distance limits for Naval Area marker generation

relent0r:
Description of changes

This PR increases the distance limits when the navmesh is generating Naval Area markers for the AI. This allows larger maps where the distance from water is greater than smaller maps allowing AI to still have a naval presence. 

Testing method
Start a 20km then 40km map and check the locations of the Naval Area markers using the Map Utilities window (see game key bindings under debug).

Notes
The decision to limit and decide which expansion should be used falls to the AI.

I would ideally like to be able to have another parameter for the GenerateForExpansion that can perform a sort by distance and allow the developer to select a Naval Area maximum count taking the closest first per marker type. That would minimize the number of markers generated and still provide flexibility for the various map types.

Known maps that had issues include Flooded Strip Mine, Liberiam Battles. There is also a 40km remake of The Wilderness which comes from the LOUD mod which was used for testing as this map featured navy but the initial distance to water for certain start positions was very high.

Testing :
Open map utilities window and set Naval Area markers as visible. Test with and without PR on a map like seraphim glaciers to see the difference with expansion markers. Then another map like flooded strip mine where the naval areas are further away to see the difference between the two.


relent0r:
@Garanas I had to pop this back into draft and it produced some crazy numbers of naval areas on certain maps.

Are you able to either assist or suggest how I could go about updating the GenerateForExpansion to support a count of markers that would sort the locations by distance then return just the closest count. I've got a good idea on how to do it but I'm not sure if I'll potentially add a big chunk of overhead to the marker generation.

Garanas:
> Are you able to either assist or suggest how I could go about updating the GenerateForExpansion to support a count of markers that would sort the locations by distance then return just the closest count. I've got a good idea on how to do it but I'm not sure if I'll potentially add a big chunk of overhead to the marker generation.

If it is done once at the start of the game then a tiny bit of overhead is fine!

PR #6054, opened 2024-04-06
Introduce `Deathnuke` armor for ASFs

Basilisk3:

All ASFs: `Deathnuke` armor introduced (ASFs take 25% less damage from ACU explosions)

## Additional context
Due to #5465, bombing ASFs with ACU explosions has become even more powerful, as they are now more expensive but remain under the threshold of 2500 hit points to survive the explosion.

## Checklist

### DIFF
--- a/lua/armordefinition.lua
+++ b/lua/armordefinition.lua
@@ -101,6 +101,7 @@ armordefinition = {
         'Normal 1.0',
         'CzarBeam 0.25',
         'OtheTacticalBomb 0.1',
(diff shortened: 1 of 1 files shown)


MrRowey:
- [ ] Team Approval

MrRowey:
Team Currenly doesn't see an issue with the way ASF are Right now and if Com-Bombs were more of a regular occurrence then would look into a different approach as we would like to try and stay away from Special Armor types.

PR #6060, opened 2024-04-07
Make shields absorb ACU explosions

### LINKED ISSUE #5006: Shields do not protect from ACU explosions

lL1l1:

Resolves #5006. Gives shields the ability to absorb ACU explosions. The absorption behavior can be easily configured in the new `ShieldAbsorptionValues.lua` file, similar to how armor types are set up.

Spawn power, a shield, and an engineer under the shield. Spawn an enemy/friendly ACU outside the shield and explode it. The shield should take 2500 damage and everything underneath will stay unharmed.

## Checklist


MrRowey (approved):

Garanas:
@lL1l1 this is quite an innovative solution, much better than what I would've made of it 😃 !

PR #6063, opened 2024-04-07
Example implementation of new tree group behaviour

BlackYps:
This is an initial implementation how tree group behaviour can be modified to allow the game to be smarter with attack move reclaim, moving us closer to something that feels like a proper area reclaim command.

Individual trees have 0 mass. The mass values of tree groups remain unchanged, but they have an additional bone in the mesh now, that triggers the spawning of a rock when the tree group breaks. At the moment this is only implemented for a single treegroup, the slightly darker pines you can see in the screenshot.

There are some advantages and some constraints to this implementation (the following paragraph assumes you have issued an attack move order in the general area):
As the game engine ignores reclaim targets that have only energy when you are mass stalling but have a full energy storage we effectively prevent engineers from reclaiming single trees in normal game scenarios. Tree groups do still get reclaimed. If a tree group is broken for whatever reason, the engi will only reclaim the rock which takes the same time as reclaiming the tree group. so no mass from the tree group is lost, only the energy that is stored in the trees, because you don't get these for free. So breaking tree groups is still a punishment, but it is not severe.

We could still implement some sort of multiplier on the mass and/or energy when the tree group brakes if we want this for gameplay reasons.

The only real limitation I have found so far is that we can't have the rock exist from the get go, it needs to spawn when the tree group brakes. But we could use a smaller rock than this example implementation and we could spawn it in a location that is hidden by a tree in the default camera angle, as to not break the illusion of a static game world.

grafik

## Checklist


Basilisk3:
This looks like an interesting concept, but I am against implementing this for several reasons:
- Spawning rocks because trees get trampled is not intuitive.
- Engineers being unable to reclaim broken trees in some scenarios is a deal-breaker gameplay wise. You attack-moved to reclaim those trees, so the engineers should not ignore them. There are also performance considerations with this.
>We could still implement some sort of multiplier on the mass and/or energy when the tree group brakes if we want this for gameplay reasons.
- This is already the case. When a tree group is broken, the broken trees gain a 2x multiplier to their mass and energy values.

BlackYps:
>     Spawning rocks because trees get trampled is not intuitive.

I need you to qualify the "not intuitive" a bit more. The whole game is not intuitive. The current implementation makes the rock spawning pretty obvious on purpose, for a refined version this would be a lot more subtle. Jip is already investigating a way to have the rock exist from the beginning on, so maybe we can fix this completely.

>    Engineers being unable to reclaim broken trees in some scenarios is a deal-breaker gameplay wise. You attack-moved to reclaim those trees, so your engineers should not ignore them. There are also performance considerations with this.

What performance considerations?
Most of the time you attack moved to get the rocks or wrecks between single trees. The single trees are only getting in the way. And the single trees will be reclaimed if you need energy.

>  When a tree group is broken, the broken trees gain a 2x multiplier to their mass and energy values.

Yes, I was mentioning this to say that we don't lose this feature, not to imply that this is a new feature.

Garanas:
> What performance considerations?
> Most of the time you attack moved to get the rocks or wrecks between single trees. The single trees are only getting in the way. And the single trees will be reclaimed if you need energy.

This is a valid point to consider: tree groups exist to reduce the number of draw calls by 10 to 20 folds. Tree groups break all the time, but usually trees still end up reclaimed because what is there else to do for an engineer that would otherwise be idle.

But usually people do not stall on energy, and therefore these trees would be ignored. Pushing the draw call, until the trees are hit by some splash damage or the like.

It's just something we need to evaluate carefully when we merge this in to experiment.

Basilisk3:
> I need you to qualify the "not intuitive" a bit more. The whole game is not intuitive. The current implementation makes the rock spawning pretty obvious on purpose, for a refined version this would be a lot more subtle. Jip is already investigating a way to have the rock exist from the beginning on, so maybe we can fix this completely.

While tree-breaking is definitely not the most intuitive part of the game, it is a mechanic that is still relatively easy to understand. You have groups, if something walks through them, they break up. 

With this implementation, we now have:
- Trees, which give you no mass, even though as a group, they would have provided considerable amounts of mass.
- Rocks appearing because trees get broken. 
- Trees that used to get reclaimed but now sometimes do not. 

We take something that is already not overly intuitive, make it more complex, and add exceptions to already understood rules where there had been none previously. This is less intuitive than what we have right now.

> What performance considerations? Most of the time you attack moved to get the rocks or wrecks between single trees. The single trees are only getting in the way. And the single trees will be reclaimed if you need energy.

What Jip said.

Garanas:
I think there is a miss understanding @Basilisk3: tree groups will still give mass. And the rock should be visible. It's just that the tree group breaks into individual trees _and_ a rock, where all the mass is in the rock and all the energy in the trees.

Basilisk3:
> I think there is a miss understanding @Basilisk3: tree groups will still give mass. And the rock should be visible. It's just that the tree group breaks into individual trees _and_ a rock, where all the mass is in the rock and all the energy in the trees.

That was my understanding as well. I am against singular trees not providing mass and spawning a rock to compensate, for the above-mentioned reasons.

BlackYps:
There are already trees in the game that give you no mass. Jip is working on a solution that the rock is there from the beginning.
So all in all the new behaviour is not really more complex than the old one.

Basilisk3:
> There are already trees in the game that give you no mass. Jip is working on a solution that the rock is there from the beginning. So all in all the new behaviour is not really more complex than the old one.

If the rock is already present from the beginning, that is a good improvement, I will definitely test this PR thoroughly once it is merged for testing.

Unfortunately, I am still against any solution, which entails certain props being ignored when attack moving. If you have a full energy bar, then for all intents and purposes it becomes impossible to remove large quantities of broken trees in a reasonable time frame. 

Just think of maps like Vya, Selkie or Seton's. I could see this pose a massive issue on these maps, not only because of the game's performance, but also because you now have potentially thousands of trees, which cannot be reclaimed. What if you have a player that mismanages his economy and overbuilds energy? This player will never be able to reclaim all those trees.

PR #6050, opened 2024-04-04
Implement the inspect library to replace recursive printing

Garanas:
## Description of the proposed changes

The existing recursive printing functions has always been lacking. The function `repr` that ships with the game can not deal with infinite loops. The function `reprs` that was made in #3861 is a bit verbose and it doesn't look or feel like a Lua table anymore. 

While looking into the language server after a discussion on Discord I noticed this library exists:

- https://github.com/kikito/inspect.lua/tree/master

I was not aware of this library at the time and it does a great job at inspecting a table. The license is MIT, therefore we can implement it in our repository as long as we keep the same license on the file.

I've implemented it which was a breeze, luckily it did not use new Lua syntax. I've also extended it with a bit of debugging information. For each table the memory reference and the flat, non-recursive size of the table is also printed in bytes. We use `debug.allocatedsize` for this which is a special function introduced by the moho engine. This value is important in my opinion as the memory footprint matters quite a bit in a 32 bit game.

As an example, this is the result of printing a unit table with a depth of 1 (the default depth):

```
INFO: { -- table: 1F385258 (1320 bytes)
INFO:   AmbientEffectThread = <thread 1>,
INFO:   AmbientExhaustEffectsBag = {...} -- table: 1F388028 (56 bytes),
INFO:   Army = 1,
INFO:   AutoModeEnabled = false,
INFO:   BeamExhaustEffectsBag = {...} -- table: 1F398190 (56 bytes),
INFO:   Blueprint = {...} -- table: 1CC3DFA0 (2600 bytes),
INFO:   Brain = {...} -- table: 1EEE4C08 (680 bytes),
INFO:   Buffs = {...} -- table: 1F3934D8 (120 bytes),
INFO:   DetectedByHooks = {...} -- table: 1F388A28 (48 bytes),
INFO:   EntityId = "1",
INFO:   EventCallbacks = {...} -- table: 1F3881B8 (40 bytes),
INFO:   IdleEffectsBag = {...} -- table: 1F3937F8 (56 bytes),
INFO:   Layer = "Land",
INFO:   MovementEffectsBag = {...} -- table: 1F393460 (56 bytes),
INFO:   OnBeingBuiltEffectsBag = {...} -- table: 1F393910 (56 bytes),
INFO:   PlayDeathAnimation = true,
INFO:   Repairers = {...} -- table: 1F3938E8 (40 bytes),
INFO:   Sync = {...} -- table: 1F385280 (120 bytes),
[...]

Issue #5946, opened 2024-02-29
Evaluate the current amount of logging performed by the game


Basilisk3:
I am by no means an expert on this, so if you feel like I am totally wrong, feel free to close this issue.

**Example:**
- The entire text of the mod.info file of a UI mod is written to the log

Garanas:
I do not have those info messages in my logs. Do they also occur without UI mod?

Basilisk3:
> I do not have info messages (1) and (2) in my logs, just once at the start. Do they also occur without UI mods?

Apparently (1) and (2) are caused by the `ZeP_MiniMapZoom` UI mod. My bad, I removed them.

lL1l1:
How about this for the mod info output?
(in `SessionInit.lua`)
```lua
LOG('Active mods in session:')
for _, mod in __active_mods do
    LOG(string.format('\t"%-30s v%02d (%-37s by %s', mod.name..'"', mod.version, mod.uid..')', mod.author))
end
```
It avoids arbitrary output from longer fields like the description. I have noticed a 1-2 second lag spike because of the large amount of text printed to the log, especially with mods including changelog fields in their mod.info (supreme scoreboard).
Example output:
```
INFO: Active mods in session:
INFO:         "Selection Deprioritizer"       v06 (bc7731b6-1416-11ea-8d71-362b9e155667) by partytime, changes by CodingSquirrel and Myxir
INFO:         "ZeP_MiniMapZoom"               v02 (5F593E02-FEF3-487C-916F-D6020903054B) by Ze_PilOt
INFO:         "Penguin's Icon Mod"            v02 (3ccc4eca-0cb1-11ec-ac65-a37599393411-02) by Emperor_Penguin
INFO:         "Ranges"                        v11 (71a26460-ec4e-11ea-adc1-0242ac120003) by Myxir, Anihilnine, BlackYps, Prophet
INFO:         "Better chat"                   v14 (better-chat-4z0t-v14)                 by 4z0t
INFO:         "Advanced Target Priorities Tweaked" v05 (afc472ed-ab3b-4c0a-a05d-e0ce966d08ee) by Strogo, Nomander
INFO:         "Notifications v5.2"            v05 (0faf3443-1122-633s-ya-V00000005002)   by Myxir
INFO:         "Share Condition Popup"         v01 (casdb00a-0ca5-11ec-bf8a-346f5a12eff26) by Emperor_Penguin
[...]

PR #6067, opened 2024-04-09
Remove usage of `repr` in production code

Garanas:

Related to https://github.com/FAForever/fa/issues/5946, #4872 and https://github.com/FAForever/fa/pull/6050

The use of `repr` is extremely dangerous from a computational perspective and should **only** be used when debugging. Due to the recursive nature it can try to create a string of many, many more tables than just the current table reference.

Also removed the cycle counting of the inspect library. The computation occurred but the output was not used.

Run several AI games.

## Additional context

@relent0r reported stutters since #6050 , this is because of the following line:

```lua
safecall("Unable to IssueTransportLoad units are "..repr(units), IssueTransportLoad, newunits, transport )
```

Because Lua evaluates eagerly the units were recursively turned into a string, regardless of whether the call succeeded or not. In FAF this also includes various references to the blueprint(s) and the brain. Therefore this can be an expensive operation!  It's a miracle that this previously did not generate (noticeable) stutters!

## Checklist


Garanas:
AI vs AI games are noticeable faster with the relevant code fixed 😃 !

PR #6073, opened 2024-04-12
Allow Mongoose and Hoplite to fire from transports

lL1l1:
Makes the small bp changes needed for the units to work in transports. Hoplite has its attachpoint bone parented to the Torso instead of the Turret bone so that it can rotate its turret while in a transport.

Loaded the units into transports and made sure they can fire in all directions. Since Hoplite's bones were changed, I also paid attention to its loading/unloading animations and found no new issues.
I also tried making a version where the Hoplite loaded upside-down into the transport, so that it wouldn't look as bad, but it had bone-snapping issues when loading, and would fall through the ground and die when unloading.

## Checklist

~~- [ ] Changes are annotated, including comments where useful~~ N/A


MrRowey (approved):

Garanas:
I'm not certain if I'm a fan of how the Hoplite behaves in a transport. Yes it shoots. But visually it looks odd:

- It is attached to the turret, but somehow it can manage to rotate the turret to aim
- When the turret aims while in the transport it self-intersects with the rest of the mesh
- It is attached, but appears to be hovering more than before

An alternative solution may be to adjust the (un)load animation, but I'm not fully certain how to make it look better even if we did. The way the Hoplite works makes it unnatural.

image

Melanol:
> I'm not certain if I'm a fan of how the Hoplite behaves in a transport. Yes it shoots. But visually it looks odd:
> 
> * It is attached to the turret, but somehow it can manage to rotate the turret to aim
> * When the turret aims while in the transport it self-intersects with the rest of the mesh
> * It is attached, but appears to have a larger distance between the transport and the Hoplite than before
> 
> An alternative solution may be to adjust the (un)load animation, but I'm not fully certain how to make it look better even if we did. The way the Hoplite works makes it unnatural.
> 
> image

In Nomads, they made Nomads labs turn upside down for their turrets to shoot down.

PR #6077, opened 2024-04-12
Improve/fix the Tempest's anti-torpedo weapon

Basilisk3:
- Instead of having one torpedo defense spread across two MuzzleBones, it now has two torpedo defenses, each with one MuzzleBone. Previously, both anti-torpedo projectiles would usually home in on the same torpedo. This change lessens this issue.
- MinRadius: 5 --> 10
Previously, the torpedo defenses attempted to home in on torpedoes, which could not be stopped anymore and were just about to hit the Tempest. This change mitigates this issue.
- RateOfFire: 1 --> 0.7
To make up for the increase effectiveness of the weapon.
- Remove unnecessary definitions from the Tempest's script file.

Spawning in a Barracuda with infinite health and have it attack the Tempest.

## Additional context
In its script file, `AntiTorpedo02` is referenced as a weapon, yet it did not exist until this PR.

## Checklist


MrRowey:
@Basilisk3 can you provide a visual comparison?

Basilisk3:
The anti-torpedo projectiles are still ejected from the same places, so visually, there is essentially no difference. The MuzzleBones themselves were not moved around, the projectiles of the torpedo defense just work like any other now.

MrRowey (approved):

PR #6079, opened 2024-04-13
Enable various structures to be placed more easily

### LINKED ISSUE #6032: What if certain structures were slightly easier to place?
## Intended changes

The idea is to make structures easier to place by having them accept higher variations in terrain. 

## How this can be achieved

Making units accept higher variations in terrain is trivial, and can be achieved via the `MaxGroundVariation` field under `Physics` in their blueprint. The default value is `1`.

**Units that should be evaluated:**

- All Tech 3 Power Generators (for air grids)
- Larger Game Enders like the Paragon or the Salvation, as they can be pretty difficult to place sometimes
- Shield Generators
- ...

## Testing

- `MaxGroundVariation = 1.1` is a reasonable and safe value, visually there is almost no difference.
- `MaxGroundVariation = 1.2` still reasonable, but fits some structures more than others.

## Additional context

Possible since #4584, as structures do not deform the terrain underneath them as much anymore.

Basilisk3:
The changes in this PR allow various units to accept higher variations in terrain. Resolves #6032.

- [x] All Tech 3 Power Generators
- [x] All Tech 3 Mass Fabricators
- [x] All Land and Air Factories
- [x] All Shield Structures
- [x] All Game Enders (not necessary for the Scathis)
- [x] All Tech 3 Artillieries

I will go with a value of `1.2` for Game Enders and `1.1` for all other units. If the `MaxGroundVariation` field is not present in the blueprint, it defaults to `1`.

- `MaxGroundVariation = 1.1` is a reasonable and safe value, visually there is almost no difference.
- `MaxGroundVariation = 1.2` still reasonable, but fits some structures more than others.
 

## Additional context
Possible since #4584, as structures do not deform the terrain underneath them as much anymore.

## Checklist


Basilisk3:
@Garanas is there any point in changing the MaxGroundVariation for HQs?

lL1l1:
Placing HQs in cheat mode should be the same as support factories and t1 factories.

Garanas:
> Possible since #4584, as structures do not deform the terrain underneath them as much anymore.

Note that this change also improved the situation, it did not make it worse.

I agree with @lL1l1 , best to make them match up. All structures of an upgrade chain should have the same value

Basilisk3:
👍

lL1l1 (approved):

PR #6083, opened 2024-04-15
platoon.lua annotations and warning fixes [lua/platoon.lua]

clyfordv:
Annotation and warning fixes for platoon.lua

files:
lua/platoon.lua

focus:
- annotation fixes (optional params, corrects improper types, etc.)
- nil checks (mostly for GetPlatoonPosition) in logic where appropriate
- streamlines several GetPlatoonPosition calls into local variables
- converts several `$type|false` local variable declaration and returns into `$type|nil`


lua/platoon.lua
```diff
@@ -1005,6 +1021,7 @@ Platoon = Class(moho.platoon_methods) {
             repeat
                 WaitSeconds(5)
                 platLoc = self:GetPlatoonPosition()
+                if not platLoc then break end ---@cast oldPlatPos Vector
```
relent0r, line 1024: If this validates to true your going to get another error further down at line 1070 I believe along the lines of self.Stop is nil assuming platLoc is nil because the platoon get destroyed. Might be better to just return.

lua/platoon.lua
```diff
@@ -2205,23 +2234,24 @@ Platoon = Class(moho.platoon_methods) {
...
         local relative
         local baseTmplList = {}
 
-        -- if we have nothing to build, disband!
-        if not cons.BuildStructures then
+        local platLoc = self:GetPlatoonPosition()
+        -- if we have nothing to build or no units, disband!
+        if not cons.BuildStructures or not platLoc then
             coroutine.yield(1)
             self:PlatoonDisband()
```
relent0r, line 2248: This could produce an error like self.PlatoonDisband is nil. Might be better to perform that platLoc check above it.

lua/platoon.lua
```diff
@@ -2205,23 +2234,24 @@ Platoon = Class(moho.platoon_methods) {
...
         local buildFunction
         local closeToBuilder
         local relative
         local baseTmplList = {}
 
-        -- if we have nothing to build, disband!
-        if not cons.BuildStructures then
+        local platLoc = self:GetPlatoonPosition()
+        -- if we have nothing to build or no units, disband!
+        if not cons.BuildStructures or not platLoc then
```
relent0r, line 2246: This could produce an error like self.PlatoonDisband is nil. Might be better to perform that platLoc check above it.

lua/platoon.lua
```diff
@@ -2237,10 +2267,10 @@ Platoon = Class(moho.platoon_methods) {
             buildFunction = AIBuildStructures.AIExecuteBuildStructure
             table.insert(baseTmplList, AIBuildStructures.AIBuildBaseTemplateFromLocation(baseTmpl, reference))
         elseif cons.Wall then
-            local pos = aiBrain:PBMGetLocationCoords(cons.LocationType) or cons.Position or self:GetPlatoonPosition()
+            local pos = aiBrain:PBMGetLocationCoords(cons.LocationType) or cons.Position or platLoc
             local radius = cons.LocationRadius or aiBrain.BuilderManagers[cons.LocationType].EngineerManager.Radius or 100
             relative = false
-            reference = AIUtils.GetLocationNeedingWalls(aiBrain, 200, 4, 'STRUCTURE - WALLS', cons.ThreatMin, cons.ThreatMax, cons.ThreatRings)
+            reference = AIUtils.GetLocationNeedingWalls(aiBrain, 200, 4, categories.STRUCTURE - categories.WALL, cons.ThreatMin, cons.ThreatMax, cons.ThreatRings)
```
relent0r, line 2273: Did you update GetLocationNeedingWalls in aiutilities? 
Its performing this ParseEntityCategory(unitCategory) at line 1052.

lua/platoon.lua
```diff
@@ -3308,6 +3345,10 @@ Platoon = Class(moho.platoon_methods) {
                 
                 WaitTicks(100)
                 platPos = self:GetPlatoonPosition()
+                if not platPos then
+                    self:PlatoonDisband() -- units are dead, disband anyway
```
relent0r, line 3349: If there is not platPos then this might error saying that self.PlatoonDisband is nil. Might be better just doing a return.

lua/platoon.lua
```diff
@@ -3866,14 +3907,14 @@ Platoon = Class(moho.platoon_methods) {
...
-            local newtarget = false
+            local newtarget
             if aiBrain.T4ThreatFound['Land'] or aiBrain.T4ThreatFound['Naval'] or aiBrain.T4ThreatFound['Structure'] then
                 newtarget = self:FindClosestUnit('Attack', 'Enemy', true, categories.EXPERIMENTAL * (categories.LAND + categories.NAVAL + categories.STRUCTURE + categories.ARTILLERY))
                 if newtarget then
                     target = newtarget
                 end
             elseif aiBrain.AirAttackPoints and not table.empty(aiBrain.AirAttackPoints) then
-                newtarget = AIUtils.AIFindAirAttackTargetInRangeSorian(aiBrain, self, 'Attack', atkPri, self.AirAttackPoints[1].Position)
+                newtarget = AIUtils.AIFindAirAttackTargetInRangeSorian(aiBrain, self, 'Attack', atkPri, aiBrain.AirAttackPoints[1].Position)
```
relent0r, line 3917: Funny thing about this. I couldn't actually find anything that is populating AirAttackPoints against the AI's brain. I'm assuming it came from Sorians AI, so I checked in that and while I can see it being defined its never populated. So I think you could just remove this if statement completely.

Looks like there are a few other functions related to this table that I forgot to pull out when I removed the Sorian AI.

relent0r:
Did testing of this in AI games and did not see any obvious errors except about that wall builder. It produces this, I did see you had another PR that was for the aiutilities but I haven't delved into it yet.

warning: Error running lua script: string expected but got userdata
         stack traceback:
         	[C]: in function `ParseEntityCategory'
         	g:\code\fa\lua\ai\aiutilities.lua(1052): in function `GetLocationNeedingWalls'
         	g:\code\fa\lua\platoon.lua(2274): in function <g:\code\fa\lua\platoon.lua:2192>
         	
Though tbh its got a zero priority so is effectively disabled, you wouldn't know there was a problem unless you specifically enabled it.

lua/platoon.lua
clyfordv: I did not. Would you say it's better to categorize the parameter earlier, like here, or keep it as a string and push the parsing down to the function?

lua/platoon.lua
clyfordv: I found a lot of logic that looks suspiciously unused but wanted to stay on the safe side of not hacking away anything important. Somebody with more experience in AI could probably go through here and do a good bit of trimming.

(But the only reference I found to AirAttackPoints was to the aibrain, which is why I changed it here.)

lua/platoon.lua
relent0r: I agree better to categorize it here, it saves the downstream function having to parse it.

lua/platoon.lua
relent0r: Thats fine. I'll put it on my list of things to do. Just keep as is for now and I'll do it in another PR once this is merged.

relent0r (approved):
Happy with changes.
Dependency on PR#6085 due to switching from string to category in wall builder function.

relent0r:
@clyfordv Can you add a snippets entry for this PR.
It would be 
/changelog/snippets/ai.6083.md
With just an entry of something like say
- (#6083)  annotations of platoon.lua and logic fixes

PR #6085, opened 2024-04-15
platoon.lua annotation and warning fixes [lua/AI/aiutilities.lua]

clyfordv:
Annotation and warning fixes for platoon.lua

files:
aiutilities.lua

focus:
- annotation additions and fixes
- convert `$type|boolean` return values to `$type|nil`
- optional params


relent0r:
What do the question marks mean in the annotations there @clyfordv?

clyfordv:
Optional parameters. Most of those marked as optional here are done so implicitly (read as: I found another piece of code somewhere that calls it with a bunch of `nil` values).

relent0r:
> Optional parameters. Most of those marked as optional here are done so implicitly (read as: I found another piece of code somewhere that calls it with a bunch of `nil` values).

Ah cheers. I'll take a look at this PR once I get home from work.

clyfordv:
Appreciate it. Happy to take any more questions or feedback, stylistic or mechanical.

relent0r (approved):
No issues with this PR.
Similar to the other, we just need a snippet added.

It would be
/changelog/snippets/ai.6085.md
With just an entry of something like say
`- (#6085)  annotations of aiutilities.lua and logic fixes`

PR #6090, opened 2024-04-16
Reduce the amount of logging performed by the game

Basilisk3:
Related to #5946.

This PR reduces the amount of logging performed by the game, which should have a positive impact on performance. Outside of the `fafdevelop` game type, the game no longer logs if a module is loaded, which caused huge amounts spam in the log previously.

## Checklist


lua/system/import.lua
```diff
@@ -119,8 +119,12 @@ function import(name, isLazy)
         return existing
     end
 
-    SPEW(string.format("%sLoading module: %s", string.rep("-> ", indent) or "", name))
-    indent = indent + 1
+    -- As this creates a lot of entries in your log very quickly, we only enable it on fafdevelop
+    local GameType = rawget(_G, "GameType")
```
Garanas, line 123: I'm fairly certain that this will always be `nil`, have you logged the value that this returns?

init_faf.lua
```diff
@@ -8,6 +8,8 @@
 -- imports fa_path to determine where it is installed
 dofile(InitFileDir .. '/../fa_path.lua')
 
+local GameType = rawget(_G, "GameType")
```
Garanas, line 11: What is the idea behind this local?

lua/system/import.lua
Basilisk3: It is indeed 0. I set up the global variables in init_faf.lua like this:
```
ClientVersion = tostring(ClientVersion)
GameVersion = tostring(GameVersion)
GameType = tostring(GameType)
```
And retreive them in import.lua. What did I do wrong?

Basilisk3: We could remove the logging every time we release a game version, and then reenable it with another PR. Would that be a feasible solution?

Garanas: I think we can remove the logging all together. It's always been there to some degree, but it's usually not my primary tool to fix an issue. What if we hide it behind a debug flag? We just define `local Debug = false` at the top of the file, and put the logging behind an if statement?

That way you can enable it when it's useful to your development session, but it's disabled by default.

Basilisk3: Sounds good, I will do that.

Basilisk3: I decided to reuse the existing local.

Garanas: The new condition also needs to be applied to `    indent = indent - 1`, at the bottom of the function

Basilisk3:
This line also gets spammed a lot: 
`debug: SND: Loaded WaveBank 'xyz'`

Perhaps something to investigate in the future.

PR #6093, opened 2024-04-18
Add entity + dummy Unit in between ExFac and ExFacUnit

clyfordv:
1. Re-adds the entity attached to the external factory base unit. This prevents "deploy" on the fatboy from detaching the external factory.
2. Adds another (newly defined) dummy unit in between the entity and the external factory. The dummy unit has the "CARRIER" category which prevents the external factory from being highlighted or selected in the worldview with the cursor, but still allows it to be selected with the exFac button.


apollodeathstar:
@clyfordv mild request : can make it so the assist command from the external factory when selected to assist the fatboy unit itself causes units produced by the external factory to surround/guard the fatboy like they do when a unit is selected and told to assist the fatboy body normally? right now the fatboy external factory can only have a set rally point that stays statically behind the fatboy and units get produced and left behind. This way you can have units produced by the factory stay with the fatboy as it advances/fights. 

Currently when you use the assist command from the external factory to the main body it simply cancels all its production queue.

clyfordv:
Currently the exFacs, like normal factories, can only assist other factories and can't have assist orders on units (you see it yourself, you can give assist commands to units but the engine eats it after a beat). The canceling the production queue is likely because the exFac thinks on some level that it's being given an assist order to another factory (even though the base unit has no production capability), which has the same result.

But, but, if there was another unit sitting around somewhere, related to the exFac, like a _dummy unit_, we could task that unit with tracking the assist target (or rally point entirely, which would probably be cleaner).

Garanas:
Let's first finalize these changes however to fix the bug 😃 !

clyfordv:
We're good to go. I was experimenting with trimming down the dummy units bp to the absolute minimum viable but that can happen later.

Garanas:
It is still possible to detach the unit that you're building:

image

Garanas:
but I'll take this pull request, it fixes the problem that matters the most

Garanas (approved):

clyfordv:
Is it known if detaching the unit was possible with the original original version, with just the entity? And do you reckon adding the carrier category to the exFac would mess anything else up? (because I think that would solve the problem)

Garanas:
No idea - we'd need to try and see what happens

PR #6096, opened 2024-04-18
Tree mass reclaim

BlackYps:
Farms said that only dead trees having no mass is not enough to properly test it, as team game maps with relevant forests often feature green trees, so I extended the change to all trees for testing purposes.

Btw, I don't really understand what "Changes are annotated" in the checklist is supposed to be...

## Checklist

### DIFF
--- a/env/Evergreen/Props/Trees/Brch01_prop.bp
+++ b/env/Evergreen/Props/Trees/Brch01_prop.bp
@@ -51,7 +51,7 @@ PropBlueprint {
     },
     Economy = {
         ReclaimEnergyMax = 3.3,
-        ReclaimMassMax = 0.66,
+        ReclaimMassMax = 0,
         ReclaimTime = 5,
     },
     Interface = {
--- a/env/Evergreen/Props/Trees/Brch01_s1_prop.bp
+++ b/env/Evergreen/Props/Trees/Brch01_s1_prop.bp
@@ -53,7 +53,7 @@ PropBlueprint {
     },
     Economy = {
         ReclaimEnergyMax = 3.3,
-        ReclaimMassMax = 0.66,
+        ReclaimMassMax = 0,
         ReclaimTime = 5,
     },
     Interface = {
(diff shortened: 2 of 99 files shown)


lL1l1:
I don't think it's useful to experiment with just 0 mass trees because it throws the mass/energy balance way off compared to normal games.

Garanas:
> Btw, I don't really understand what "Changes are annotated" in the checklist is supposed to be...

It means that when you write functions such as:

```lua
function IsAlly(army1, army2)
  -- (...)
end
```

That you [annotate](https://github.com/LuaLS/lua-language-server/wiki/EmmyLua-Annotations/85b3e724f9e40e9d9bc450577bd68b35a9fcb464) it like this:

```lua
---@param army1 Army
---@param army2 Army
---@return boolean
function IsAlly(army1, army2)
end
```

Which applies to [prop blueprints](https://github.com/FAForever/fa/blob/deploy/fafdevelop/engine/Core/Blueprints/PropBlueprint.lua) too, but that work is mostly completed.

Garanas:
> I don't think it's useful to experiment with just 0 mass trees because it throws the mass/energy balance way off compared to normal games.

I agree - removing it from individual trees is fine but this also removes it from tree groups which is not the intention in the end?

Garanas:
These changes are equivalent with just playing with the following mod:

image

Which does the following:

```lua
do
    local oldModBlueprints = ModBlueprints

    function ModBlueprints(all_bps)
        oldModBlueprints(all_bps)

        for k, prop in all_bps.Prop do
            if prop.ScriptModule == '/lua/proptree.lua' then
                prop.Economy.ReclaimMassMax = 0
            end
        end
    end
end
```

Farms can host a game or two with the mod to test out the feature.

BlackYps:
> removing it from individual trees is fine but this also removes it from tree groups which is not the intention in the end?

Yes, that is not the intention in the end. It's to temporarily test out the scenario of basically all tree groups have been broken. Farms described some potential gameplay problems that can arise if trees give only energy and you rely on the energy reclaim but you sometimes overflow as well. So this is intended to be an easy way to gather some insight how the game plays out in these scenarios. I opted to do this, as it is significantly less effort than doing the armature changes in all the tree groups. Especially because the blender importer is a bit bugged at the moment.

BlackYps:
I pinged you in discord, but actually keep this closed. It was based on a misunderstanding and is not needed

PR #6097, opened 2024-04-19
Draft setup for persistent tree groups

Garanas:

This is based on a suggestion of @clyfordv where when a tree group breaks (because of a collision, damage, etc) we do not delete the group and instead spawn in individual trees that are not reclaimable. That way we can still reclaim the tree group as a whole, and instead reduce the mass value as individual trees are destroyed.

<todo>

## Additional context

Ran into a strange bug where I am unable to change the mesh of a tree group after creating it. 

When the tree group 'breaks' we want to hide the mesh of the tree group itself. This allows the individual trees to take over, which in turn allows for interactions with units and damage.

Attempts:

- (1) Hide the mesh - this is not possible. Units are able to do this, but those are unit-specific functions.
- (2) Scale the mesh to 0 - this is not possible. The rendering bounding box appears to be used to determine if we're hovering over a tree (group). As a result if we reduce the scale to 0 then the tree group becomes unclickable.
- (3) Change the mesh - this appears to be buggy and is where I am now. I can change the mesh to a version with a shader that doesn't render any pixels on screen but only in the tick the prop is created. When I try it at a later moment it simply does not work and the game does not tell me why. I suspect it may be because of batching of the props, but then why does it work for individual trees as they take (fire) damage?

## Checklist

### DIFF
--- a/effects/mesh.fx
+++ b/effects/mesh.fx
@@ -2899,6 +2899,16 @@ float4 NormalMappedPS( NORMALMAPPED_VERTEX vertex,
     return float4( color.rgb, alpha );
 }
 
+/// InvisiblePS
+///
+/// Lighting using normal maps.
+/// Used by some props, mostly trees
+float4 InvisiblePS( NORMALMAPPED_VERTEX vertex) : COLOR0
+{
+    clip(1);
+    return float4(0, 0, 0, 0);
+}
+
 /// MapImagerPS0
 ///
 ///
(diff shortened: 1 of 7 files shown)


Garanas:
@Hdt80bro any idea why the mesh change no longer works 1 tick after creating a tree group?

clyfordv:
What if instead of hiding the mesh you just spawn an additional prop with the invisible shader? So parent tree group goes away, new invisible reclaim entity appears from the break.

Garanas:
Because you'd lose the reclaim order on the original tree group when we do that

clyfordv:
Okay what about the reverse--the tree group is always an invisible, but has a submesh/entity/something(?) that gets resized when it breaks (so the parent bounding box remains for clickability). Does that require an additional dummy parent entity?

Garanas:
I just tried that suggestion; I can't get it to work reliable. I just don't know how the engine determines what is and is not clickable and what the preferences are

PR #6105, opened 2024-04-21
Fix a typo in the script file of the Seraphim Tech 3 Mass Fabricator

Basilisk3:
Hopefully this also fixes this warning I sometimes see in replays, but I cannot promise it:

```warning: ToggleEnergyExcessUnitsThread: ...\gamedata\units.nx2\units\xsb1303\xsb1303_script.lua(35): attempt to call method `SetSpinDown' (a nil value)```

## Checklist

### DIFF
--- a/changelog/3810.md
+++ b/changelog/3810.md
@@ -2,7 +2,7 @@
 
 ## Bug Fixes
 
-<!-- Remove header when empty -->
+- (#6105) Fix a typo in the script file of the Seraphim Tech 3 Mass Fabricator.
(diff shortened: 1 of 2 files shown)


lL1l1 (approved):
That typo shouldn't do anything since it ends up as an array-index entry into the table of the class. The nil error could happen if the fabricator is automatically turned off at the same exact time that it dies and calls :destroy() on itself.
Taking a look let me catch the animation issue though.

PR #6107, opened 2024-04-22
Generalize Projectile behavior for `DamageToShields`

lL1l1:

Fixes ["New shield disruptor does not damage personal shields" (Discord)](https://discord.com/channels/197033481883222026/1227390701474545665).
Fixes Absolver damaging shields that are down, which makes no sense and prevents them from recharging.

Shoot both Pulsar and Absolver at Harbinger (personal shield), and ED5 (static shield).

## Checklist


Garanas:
> Fixes Absolver damaging shields that are down, which makes no sense and prevents them from recharging.

I'm not sure if this is a bug; I noticed it years ago too and back then it was coined a feature. I'm not able to retrieve the conversation about it however; it is somewhere in Discord 😃

lL1l1:
@Garanas What do you think is the best way to implement this? It doesn't feel right to apply it to every single weapon/projectile, so I was thinking of making it a weapon type + projectile type, a weapon type that modifies the projectile's script, or a component version of the two options.

lua/armordefinition.lua
```diff
@@ -28,8 +28,12 @@
 ---| "TacticalMissile"
 ---| "TreeFire"
 ---| "TreeForce"
+---| "Disintegrate" # Used by tree props
+---| "Force"        # Used by tree props
+---| "Fire"         # Used by tree props
 ---| "WallOverspill"
 ---| "TransportDamage" # Skips visual effects in OnKilled
+---| "FAF_AntiShield" # Only deals damage to shields
```
Garanas, line 36: I recall a discussion about this a while ago, but why do we prepend it with `FAF_`?

lua/sim/Projectile.lua
```diff
@@ -6,6 +6,8 @@
 ------------------------------------------------------------------
 
 local DefaultDamage = import("/lua/sim/defaultdamage.lua")
+local UnitDoTThread = DefaultDamage.UnitDoTThread
```
Garanas, line 9: Nice catch 😃

lua/armordefinition.lua
lL1l1: The prefix is just for avoiding collisions with mods, just like how we did with the Aeon ACU's advanced range upgrade in #5231. In particular because `AntiShield` could easily be something someone else came up with and used.

Garanas (approved):
Great work @lL1l1 !

PR #6108, opened 2024-04-23
Entity Tree replacement

clyfordv:
Replaces prop trees spawned by tree groups with entity versions. Trees get damaged/lit on fire/knocked over, but tree groups remain as reclaimable entities, and the entity tree doesn't interfere with the cursor when reclaiming the group.

Hiding the mesh of the tree group when it breaks is accomplished by moving it underground. Not my proudest moment, but extremely near to giving it an invisible shader, and reliable in that tree groups by their nature are only ever on flat areas.

Is functional and stable, but is critically missing unit collision. I was experimenting with handing all collision over to an attached dummy projectile (which would remove the need for a collision box on the tree itself). Right now you knock trees over with teleport blast/shoot em with the commander gun etc. to see how it works.


Garanas:
Closing as there's been little to no interest in this direction of improving the reclaim behavior.

PR #6109, opened 2024-04-23
Annotate shield assist mechanics and max jammer blips

lL1l1:
Describe how shield assistance works in annotations.
Annotate that jammer blips have a max of 255. Past 255, it overflows back to 0.

Tested modifying repair rate to change shield assist cost. Determined regen assist mechanics through looking at regen and consumption values on assisted shields tick by tick.

## Checklist


Garanas (approved):
Great work; always useful to document this behavior!

PR #6111, opened 2024-04-23
Add MassValueKilled stat for access by the UI

clyfordv:
Unlike veterancy, which is capped at 1 level/kill, MassValueKilled is a straight sum of the mass exp a unit would have gained with no level cap. Let me know if this a go, and I'll add it to https://github.com/FAForever/fa/pull/6110.


Garanas:
This is fine 👍

PR #6112, opened 2024-04-23
Condense internal storage display for carriers and multiple transports

clyfordv:
Turns the mess of icons in a fully loaded carrier into a list of single icons with numbers denoting how many units are loaded.
- Multiple carriers of different types display condensed cargo
- Multiple transports display condensed cargo
- Multiple unit icons can be deployed/unloaded as expected (that is, all at once) by clicking the button
- Selecting carriers + transports will display nothing
- Single transport behavior is unchanged

Needs fixing (likely before merge):
- Sort function is bad, a better version should be in `shared`


BlackYps:

PR #6113, opened 2024-04-23
Add adjustable delay to how fast QuickDrop fires

clyfordv:
Saw enough talk both ways that I'd figured I'd bang this out as an option. Adds a minimum time for QuickDrop to fire.

Reasons:
- Provides a balance knob
- Current drop looks a little too sharp (units fly off at an odd velocity, transport isn't settled)


lua/sim/units/AirTransportUnit.lua
```diff
@@ -68,9 +71,15 @@ AirTransport = ClassUnit(AirUnit, BaseTransport) {
                     return
                 end
 
-                -- Tell our navigator to abort the move
+                -- Tell our navigator to abort the move after a delay/if we haven't detached cargo in the interim
                 -- this has the effect of causing the next unload command to be executed immediately
-                navigator:AbortMove()
+                self.dropFlag = true
+                ForkThread(function()
```
lL1l1, line 77: This should be in a local function to avoid creating new closures.

lua/sim/units/AirTransportUnit.lua
```diff
@@ -68,9 +71,15 @@ AirTransport = ClassUnit(AirUnit, BaseTransport) {
...
+                    WaitTicks(QuickDropDelay)
```
lL1l1, line 78: Forking a thread waits 1 tick but then WaitTicks correctly waits 10 ticks from before the thread was forked. e.g.:
```
LOG(GetGameTick()) -- tick 1
ForkThread(function ()
LOG(GetGameTick()) -- tick 2
WaitTicks(10)
LOG(GetGameTick()) -- tick 11
```
I don't really understand it but it works as long as QuickDropDelay >= 2 (1 delay ends up the same as 2).

lua/sim/units/AirTransportUnit.lua
```diff
@@ -30,6 +30,9 @@ local HorzUnloadMargin = 2.5
 -- Factor by which to multiply TransportHoverHeight when determining if we can unload cargo
 local VertUnloadFactor = 1.5
 
+-- time delay after which quick drop fires (drops can be faster than this, but not slower)
+local QuickDropDelay = 10
```
lL1l1, line 34: Need to mention that the time delay is in ticks.
Maybe `MaximumUnloadWaitTicks` is better. Conveys ticks and matches the two variables above in using `Unload`.
Need to document what is a "quick drop" function.

lL1l1 (dismissed):

lua/sim/units/AirTransportUnit.lua
Garanas: Waiting ticks is a bit strange - waiting 0 and 1 ticks is effectively the same tick! When we pass in 1 tack then we effectively tell the game to perform this function at **the end of** the current tick. When we wait 2 ticks, we tell the game to queue it for the next tick. 

Effectively we also wait for the 'current' tick.

Garanas (commented):
I'm not sure if this pull request takes it in the right direction. I understand what it is trying to fix, but what if we'd just remove the quick drop functionality instead?

Since the transport changes loading and unloading was much faster than it was before to begin with. I may run a few simulations to verify how much quick drop actually makes everything drop faster.

clyfordv:
The original motive was not even to make transports drop faster, but to untangle drop time from transport `LiftFactor` and K factors, given that:
1. Transports are pretty particular about how close they need to be to their drop point before releasing cargo.
2. Adjusting flight characteristics so that transports don't (as an example) descend from mountains at a glacially slow pace caused them to oscillate around their drop point, taking even longer.

If #1 has been fixed (through an engine patch?) I could see quick drop no longer being necessary, but as long as it remains #2 is going to apply.

I didn't see a huge need for the addition of this delay, but it was raised as an issue on the discord (not a majority opinion, perhaps) so I figured I'd bang it out.

(And a minor point, but the visuals of the transports taking a moment to level out are much better, which means we could undo the bank factor changes that came with the original QuickDrop because dropping at a steep bank looked rather jarring.)

lua/sim/units/AirTransportUnit.lua
lL1l1: > Do you think that warrants a warning if the delay is too short? I don't see this number being changed too often.

I think a comment saying that the minimum is 2 ticks is sufficient.

> When we pass in 1 tack then we effectively tell the game to perform this function at the end of the current tick. When we wait 2 ticks, we tell the game to queue it for the next tick.

Is that why some waits have off by 1 error compensation?
https://github.com/FAForever/fa/blob/b29226f32bc6224377c0334b2b49e738ec7b2fe5/lua/EffectUtilitiesUEF.lua#L24-L29

https://github.com/FAForever/fa/blob/b29226f32bc6224377c0334b2b49e738ec7b2fe5/lua/sim/projectiles/components/TacticalMissileComponent.lua#L132-L135

Garanas:
_With these changes_ 

https://github.com/FAForever/fa/assets/15778155/664a83d7-a23f-4123-92af-452d399d1caf

_Without the quick drop feature_

https://github.com/FAForever/fa/assets/15778155/96d337da-0484-4ac9-bab4-5768b05a2b86

There's not much difference in my opinion. Of course some maps have rather extreme cliffs, but there's a limit to what we can (not) support.

BlackYps:

PR #6118, opened 2024-04-25
fix incorrect table being inserted into AI marker function

relent0r:
A very small change that corrects an error in a default AI marker function that was unused until now.

Validated that the marker table is being populated and returned

## Additional context
Noted by Grandpa Sawyer as he was wanting to use it for a campaign mission.

## Checklist

Jip I haven't updated the changelog yet, not sure if you still want to do the create file with name of next version thing still or not?


Garanas:
> Jip I haven't updated the changelog yet, not sure if you still want to do the create file with name of next version thing still or not?

When it is done - but until then, the current approach will have to do 😃 !

Garanas:
@relent0r could you add the changelog entry in the new setup? See also `/changelog/snippets` and then you'd need to create a `ai.6118.md` file, which contains just the entry that you introduced to `3810.md`.

lL1l1 (approved):

PR #6119, opened 2024-04-25
UnitviewDetail: do not display the Bouncer's targeting laser

Basilisk3:
The Bouncer's targeting laser is a purely aesthetical weapon and does not deal any damage.

The unit still functions normally. According to the wiki, `WeaponCategory` is a deprecated attribute and changing it will not affect the weapon.

## Checklist


lL1l1 (approved):
WeaponCategory is used for some AI threat calculation and UI things, but yes it's not read by the engine (doesn't show up in blueprint viewer).

PR #6124, opened 2024-04-26
Remove caching of the preference file

Garanas:

Fixes bugs that were introduced by #5871 . Caching the preference file is significant, but apparently the existing implementation introduces all sorts of buggy behavior.

Start a skirmish game, create a template and then start a new skirmish game in the same game session. The template should now still exist.

Try to change the resolution. The message to verify the changes should now pop up.

## Additional context

It was apparently a greedy change of me 😃 and we should probably hot fix this.

## Checklist

### DIFF
--- a/lua/user/prefs.lua
+++ b/lua/user/prefs.lua
@@ -20,61 +20,23 @@
 --** SOFTWARE.
 --******************************************************************************************************
 
-local optionsLogic = import("/lua/options/optionslogic.lua")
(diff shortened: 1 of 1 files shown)


lL1l1:
Are the bugs from caching the pref file engine bugs? Templates not saving on session end sounds like prefs not saving on session end, which may be hookable with lua `Exit.lua AddOnExitCallback`.

PR #6134, opened 2024-04-27
Update campaign AI functionalities

Dhomie:
The primary intent of this PR is to move from the thread-based method of processing units for the current campaign `BaseManager` to a callback-based one. 
The secondary intent was to do some cleanups, fixes (TMLs ignoring T2 units), code optimizations, and where I could, commenting, and annotations.

This affects some build condition files used by skirmish AIs, so @relent0r if you could review those, and tell me if anything conflicts, or the changes are wrong, let me know.
If anyone else's reviewing this, and got feedback, I'm all ears.

I've done plenty of testing to ensure this works as intended, since all the changes can be found in a mod in the vault: **Coop AI Adjustments**
You can test these changes either via the mod, or the dev enviroment.

GitHub link: https://github.com/Dhomie/FAF-Campaign-Experiments

## Additional context

This is going to need some context, so bear with me please.

First of all, Engineer platoons, and even structures now use several unit callbacks, to assign unit names, upgrades, etc. instead of checking every X tick if a unit's state has changed. This issue was most noticable with the current upgrade thread. An example how to break it: Put 152475213675 Hives next to a T1 Land Factory that should upgrade all the way to T3, and it'll remain stuck at T2, because the structure upgrade finishes before the upgrading thread can register that a structure upgrade is being built.
The same can happen to just Engineers building something, if it finishes before the currently assigned game ticks, the `BaseManager` won't register it.

The callbacks I've made use of are: `OnStartBuild, OnStopBeingBuilt, OnFailedToBuild`.
In case an upgrade order somehow doesn't get processed, I've added a fallback thread for the `BaseManager` that will re-order the upgrade.
Due to the reduction of overall checks thanks to the above, I've reduced game ticks needed to wait for certain tasks.

While doing all this, I've went over some campaign specific files, cleaned unnecessary stuff up, and added comments/explanations where possible.

## Checklist


relent0r:
At first glance I have no issues with the condition changes, looks like they are all just logic and performance improvements. We have debated some of this stuff before as it makes troubleshooting a little harder.

I'll need to take a longer look at the other changes as some of the stuff was quite interesting. I'd never considered doing adhoc unitbuilt callbacks for upgrading factories.
A word of warning based on my own recent experiences with callbacks that I got caught out by. I may not be relevant for campaign AI but if an engineer gets a build command then for some reason it has issueclearcommands sent the OnFailedToBuild callback will fire. That particular one caused(causes) a bunch of problems with the default skirmish AI due to how it adds things to the engineerbuildqueue then later on clears the commands.

Dhomie:
`OnFailedToBuild` does exactly what is expected of it in this case, lemme explain.

The main idea to move to build callbacks was to cache unit names on the Engineer (the `BaseManager` uses single-unit Engineer/ACU/sACU platoons for its bases).
All the structures, and special units `(ConditionalBuilds)` are in the `save.lua` file with all the necessary data, like unit IDs, and most importantly their names, ie. `'UNIT_126'` unless given a custom name, it's the naming method of the map editors.

So, when the Engineer finds something to build, AND the build order has been issued successfully, we cache that *something's* name on it, which is either a structure part of the base, or a `ConditionalBuild`.
If the Engineer begins building, and this cached name is available on the Engineer, ONLY then we assign all the relevant data on the unit being built, which `EngineerOnStartBuild()` handles. There shouldn't be a case where both a `BuildingUnitName` and `ConditionalBuildUnitName` is cached on the Engineer, it's either one or the other.

This is where `OnFailedToBuild` comes in, because we ONLY issue 1 build related order at a time for Engineers in the campaign enviroment.
What `OnFailedToBuild` does in this case, is that if for whatever reason (like the one you mentioned, the Engineer getting its orders cancelled) the build failed before it could even be started, all we do is set BOTH cached name types on the Engineer to `nil`, so it can either retry, or build something else.

lua/AI/AttackManager.lua
```diff
@@ -202,20 +265,25 @@ AttackManager = ClassSimple {
...
             if self:CheckAttackConditions(v) then
                 local combineList = {}
                 local platoonList = self.brain:GetPlatoonsList()
+				
+				-- Loop through all of the platoons the AI has, check if it's part of the attack force
+				-- The PBM platoons are the ones set to be part of the attack force, if they have 'AMPlatoons' platoon data set, this is handled by the "PBMFormPlatoons()" function
+				-- If it's part of it, check the PlatoonData for the name of the 'master' platoon it belongs to, and if it matches, insert this platoon to the combineList
+				-- These are defined in the lua/AI/OpAI *save.lua files, containing all of the default platoons
                 for j, platoon in platoonList do
                     if platoon:IsPartOfAttackForce() then
```
relent0r, line 288: This function is used twice in the entire codebase and all it does is check platoon.PartOfAttackForce.

lua/AI/AttackManager.lua
```diff
@@ -342,24 +445,28 @@ AttackManager = ClassSimple {
...
         return result
     end,
-
+	
+	--- SC1's use of attack vector data, mostly on the engine side, so I can't comment on it
+	---@param self AttackManager
     AttackManageAttackVectors = function(self)
         local enemyBrain = self.brain:GetCurrentEnemy()
         if enemyBrain then
             self.brain:SetUpAttackVectorsToArmy()
```
relent0r, line 458: If curious I explored this function a while back. You can pass a category to it to filter down the attack vectors that it generates, then there is another function called GetAttackVectors which will return a list of IMAP positions along with a look around radius (32 was what it returned for me which I believed was the imap block size) around the enemies base related to the units that had those categories.

lua/AI/OpAI/BaseManager.lua
```diff
@@ -158,6 +158,67 @@ local BuildingCounterDefaultValues = {
...
+--- Failsafe callback function when a structure upgrade is finished building
+--- Updates the ScenarioInfo.UnitNames table with the new unit, and upgrades further if needed
+---@param unit Unit
+function FailSafeUpgradeOnStopBeingBuilt(unit)
+	local aiBrain = unit.Brain
+	local bManager = aiBrain.BaseManagers[unit.BaseName]
+	
+	if bManager then
+		local armyIndex = aiBrain:GetArmyIndex()
+		ScenarioInfo.UnitNames[armyIndex][unit.UnitName] = unit
```
relent0r, line 204: I'm not familiar with this part of the campaign. Looks like its a way of keeping track of units that the AI is building/upgrading, the unit name is the position vectors?

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
```diff
@@ -523,12 +691,12 @@ function PermanentFactoryAssist(platoon)
 
         -- Determine the number of guards on all factories
         local high, highFac, low, lowFac
-        for _, v in facs do
+        for k, v in facs do
```
relent0r, line 694: Is this change required given you don't use the key anywhere? (or do you)

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
```diff
@@ -757,12 +914,10 @@ end
...
 function ExpansionPlatoonDestroyed(brain, platoon)
     local aiBrain = platoon:GetBrain()
-    local data = platoon.PlatoonData
-    local bManager = aiBrain.BaseManagers[data.BaseName]
-    local eBaseName = false
+    local bManager = aiBrain.BaseManagers[platoon.PlatoonData.BaseName]
 
     for num, eData in bManager.ExpansionBaseData do
-        if eData.BaseName == data.ExpansionBase then
+        if eData.BaseName == platoon.PlatoonData.ExpansionBase then
```
relent0r, line 920: Whats the benefit here? your doing an extra table lookup on each loop.

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
```diff
@@ -793,103 +948,94 @@ function BaseManagerEngineerThread(platoon)
...
+    end
+	
+	-- Grab the Engineer
+    local Engineer
+	
+    for _, v in platoon:GetPlatoonUnits() do
         if not v.Dead and EntityCategoryContains(categories.CONSTRUCTION, v) then
-            if not eng then
-                eng = v
+            if not Engineer then
```
relent0r, line 963: Standard naming convention for an engineer is usually eng. Not an issue really but for consistency. (or is it being defined at Engineer everywhere else in the campaign codebase?)

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
```diff
@@ -793,103 +948,94 @@ function BaseManagerEngineerThread(platoon)
...
-    platoon:MoveToLocation(tempPos, false)
+        until Engineer.Dead or Engineer:IsIdleState()
+		
+		-- Break out if we couldn't find a structure to build
+		if not StructureFound then
+			break
+		end
+	end
+	
+	platoon:MoveToLocation(aiBrain.BaseManagers[platoon.PlatoonData.BaseName]:GetPosition(), false)
```
relent0r, line 1038: I know your not changing anything here. But is this valid? e.g is this looking at an entity?

lua/aibrains/campaign-ai.lua
```diff
@@ -1284,45 +1284,42 @@ AIBrain = Class(StandardBrain) {
...
-                    and (not vp.LocationType or vp.LocationType == location.LocationType))
+					and numBuildOrders and numBuildOrders == 0
+					and (not vp.LocationType or vp.LocationType == location.LocationType))
                     or (((self:PBMHandleAvailable(vp)) and (not requireBuilding or not globalBuilder.RequiresConstruction))
-                    and (not vp.LocationType or vp.LocationType == location.LocationType)
-                    and self:PBMCheckBuildConditions(globalBuilder.BuildConditions, armyIndex)) then
+					and (not vp.LocationType or vp.LocationType == location.LocationType)
+					and self:PBMCheckBuildConditions(globalBuilder.BuildConditions, armyIndex)) then
                 local poolPlatoon = self:GetPlatoonUniquelyNamed('ArmyPool')
+				local ptnSize = personality:GetPlatoonSize()
```
relent0r, line 1293: Can you fix the tabing of these(I think there are two instaces of it), just made it a little harder to read.

relent0r:
@Dhomie I'm not seeing any big issues here.  The callbacks are the only major deviation, the rest looks like small improvements and some great documentation.
There are some elements of consistency where it seems like you had one view on things when you started and a different one by the time you finished.(like converting WaitSeconds to WaitTicks or localizing variables vs table lookups).

Your callback explanation makes sense. My views on these are from the skirmish AI's perspective where callbacks can fire for multiple reasons and require different resolution logic depending on the engineers status and location. I've been going through an interesting process of creating a statemachine for AI engineers which is where some of these recent learnings have come from. 

The skirmish AI uses a catchall function called processbuildcommand that most of these callbacks go to, its been modified quite a few times over the years and became a little detached from what the callbacks were trying to achieve so causes transient failures that are hard to track down.

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
relent0r: Actually ignore that. I didn't realize there was a GetPosition function defined in the BaseManager.lua module.

lua/AI/AttackManager.lua
Dhomie: There are 2 types of platoons, PBM, and AM.
A PBM platoon is the first part of it all, and they aren't necessarily set to be part of an AM platoon later on (ie. Engineer platoons and such).
Since GPG went with checking through **all** platoons, they probably came up with this method to filter them out.

If a platoon is part of the attack force, it means it's meant to be part of an AM platoon, otherwise they shouldn't be messed with. It's really a campaign-only thing for SCFA, and FAF.
In SC1 this had more impact because the PBM and the AM were what made up the skirmish AI.

lua/AI/AttackManager.lua
Dhomie: This was probably used for SC1's skirmish AI, so good to know.
I understood the idea around it, but due to it being on the engine side, not what happened exactly,
There are functions like `PickBestAttackVector()` and `GetNewAttackVectors()` and enemy picking logic in SC1's AI files.
They actually had some pathing and vector setup functioning, just mostly on the engine side, and relied on what I'm guessing is surface height checks for pathing.

Depending on the skirmish map, they were sometimes accurate, sometimes not.
This is something you have to observe in SC1 during a skirmish match.

lua/AI/OpAI/BaseManager.lua
Dhomie: `ScenarioInfo.UnitNames[ArmyIndex][UnitName]` is basically the method GPG came up with to globally keep track of existing units in a game instance. The name itself is what we assigned in the `save.lua`, it's unique, but the map editors usually use something like this by default: `'UNIT_196'`, then this entry is assigned some data, you can check such an entry in a `save.lua` file for the sake of an example.

The `BaseManager` populates some internal tables with data on what base to maintain from the map's `save.lua`.
The position is relevant for building, but that data is handled inside those internal tables.

Upgrades however require for the unit to exist, and comparisons with the provided unit type (ID specifically).
Structures like T3 factories are automatically converted to their T1 counterparts when we add a base template, or just any groups to the template. When the unit is created, we update the `ScenarioInfo.UnitNames` table accordingly, hence why unit names are important. The desired ID is saved internally in the BaseManager, with which we compare if the current unit is of the desired tech level.

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
Dhomie: I've done this long ago, just for the sake of consistency, it's not required.

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
Dhomie: Other than the removal of an unused local, `eBaseName`, not much, this again is something I touched a long time ago when I first started going through this.
I tried reducing the code amount, just didn't consider it might be bad for overall performance.

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
Dhomie: This could be changed to `engineer`, that's actually more commonly used in the file.
Variable naming conventions are not exactly consistent throughout the codebase, and that's a GPG thing.

lua/aibrains/campaign-ai.lua
Dhomie: I use space, and a tab size of 4 when viewing these.
If the segmenting looks off, try viewing it with those settings.

Dhomie:
I've reverted some of the unnecessary/wrong changes.

Engineer logic for the campaign enviroment is designed to be simple, and usually is linear in its task.
It's usually 1 type of task at a time that is expected from them, and we don't do anything too fancy, we build special platoons if something fancy is needed (ie. if reclaiming is needed, a set of patrol orders will suffice, or just an attack-move).

Structure upgrading logic still resembles the one GPG intended, which means upgrading as soon as possible, no need to check for eco values and such. An AIs economy is mostly set up by the mission's maker, alternatively we simply give the right enviroment for their eco to start out (ie. give some starting Engineers to patrol wreck areas).

@speed2CZ I'd like your review on this PR as well, since it features overall campaign related refactoring, and I'd like your opinion on what you'd approve of, and not.

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
Dhomie: Change reverted.

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
Dhomie: Change reverted, except for the unused local variable.

lua/AI/OpAI/BaseManagerPlatoonThreads.lua
relent0r: All good, I don't have a preference.

relent0r (approved):

relent0r:
@Dhomie I've approved this PR (I ran through some play testing on mind games and it seemed good). You can either wait for speed2 or I can merge in now.

speed2CZ:
I dont have time for a proper review, but the stuff I've tried worked fine, so its probably good.

Dhomie:
@relent0r It's all good from my end, you can merge whenever you're ready.

PR #6137, opened 2024-04-27
Feature/changelog snippets

Garanas:
## Description of the proposed changes

Implements a similar setup to https://github.com/PrefectHQ/prefect/issues/2311 to manage the changelog and prevent conflicts.

## Testing done on the proposed changes

I tried testing with [Act](https://github.com/nektos/act) to run the workflow locally but I've clearly not fully understood how to use the tooling yet.

PR #6138, opened 2024-04-27
Automate creation of the changelog

Garanas:

Continuation of #6137 

Updates the Github workflow file to create a changelog without the need to run the bash script on your system. The workflow removes the need to have a Bash interpreter available on your system. At the moment it outputs the following file:

[changelog.md](https://github.com/FAForever/fa/files/15140017/changelog.md)

Which is a great start!

Run the bash script locally and on Github via the workflow.


Garanas:
@Sheikah45 and/or @Brutus5000 would either of you be available to review the workflow file? In particular I do not fully understand why the script fails when it's not run in the faf-lua container. See also the failed runs.

Sheikah45:
The workflow does run in the lua container though. It is specified in the container line of the yaml as part of this PR. From the log messages it just looks like the container does not have bash installed in it.

Sheikah45:
Also why are you specifying bash? Why not just execute the file and allow the shebang to tell the system how to execute it

Sheikah45:
Also you are calling combine.sh as if it is in the root directory but it looks like the file lives in changelog/snippets so just ./combine.sh will result in file not found

Garanas:
@Sheikah45 processed your feedback 👍 

> Also you are calling combine.sh as if it is in the root directory but it looks like the file lives in changelog/snippets so just ./combine.sh will result in file not found

Note the working directory that is set!

Sheikah45:
Ah yeah I had missed that

PR #6139, opened 2024-04-28
Add workflow to verify the changelog snippets

Garanas:

Continuation of #6137  and #6138 .

Adds a verify stage so that the format of the snippets are verified.


Garanas:
It's interesting how one can keep running into this problem:

- https://stackoverflow.com/questions/40978921/how-to-add-chmod-permissions-to-file-in-git

PR #6140, opened 2024-04-28
Slightly buff the Vulthoo

Basilisk3:

**Vulthoo: T2 Gunship**
Damage (x2): 19.2 --> 20
MaxRadius (x2): 22 --> 24

## Motivation behind the changes
This unit underperforms slightly due to several factors: 
- It is less spammable than its counterparts.
- It has considerably less DPS/mass than the other Tech 2 gunships (`0.192`), with the Renegade has `0.222` and the Specter at `0.241`. 
- It has no AoE or alpha damage unlike the Specter and the Renegade, both of which also have more DPS/mass.
- It does not have stealth/jamming like the Wailer and Broadsword.
- It lacks the increased range of the Tech 3 gunships, which is important against mobile flak.
- It is more survivable than other Tech 2 gunships, but it still performs closer to its Tech 2 counterparts than its Tech 3 counterparts.

## Additional context

- I believe we should not use decimal places for stats like damage. Balance wise adding decimal places is usually not necessary and currently units do not display decimal places in their life bar.
- This fixes unitviewDetail displaying its damage as `19.200001`.
- The original issue (#5418) was unclear about how the damage should be changed: `DPS: 80 > 100 (+20%)`. This would actually be a 25% buff, which I think fits the unit better. Its current DPS is 96.

## Checklist


Basilisk3:
Note that currently all Tech 2 gunships sometimes have difficulties targeting larger, fast-moving units. You can test this easily by spawning in a Tech 2 gunship with infinite health, and give it an attack order on a retreating Carrier. Sometimes the gunships will trail behind the carrier but will not fire. This can be mitigated by giving them a higher MaxRadius.

lL1l1:
> Note that currently all Tech 2 gunships sometimes have difficulties targeting larger, fast-moving units. You can test this easily by spawning in a Tech 2 gunship with infinite health, and give it an attack order on a retreating Carrier. Sometimes the gunships will trail behind the carrier but will not fire. This can be mitigated by giving them a higher MaxRadius.

The best way to fix this is to adjust the `CirclingRadiusChangeMaxRatio` and `CirclingRadiusChangeMinRatio`. They determine how far away the gunship circles in proportion to the first weapon's max range. 
As a last resort, you can lightly adjust `KMove` (higher = move with more force) and `KMoveDamping` (lower = loses speed slower) to make it chase more aggressively. T2 gunships have 0.8 KMove, T3 gs 1, ASF/Inties 1.5, and transports 3.0.

By the way if you didn't know, the console command `NoDamage` can disable damage instead of having to edit infinite health in.

Basilisk3:
> The best way to fix this is to adjust the `CirclingRadiusChangeMaxRatio` and `CirclingRadiusChangeMinRatio`. They determine how far away the gunship circles in proportion to the first weapon's max range. 

That is good to know, thanks for pointing that out. I will test this and determine if it is something we should change. This would make them circle closer to all units, including mobile flaks, right? 

I will keep the range buff for the Vulthoo in place, since as a Tech 2.5 unit, it makes sense for it to have a higher MaxRadius.

MrRowey (approved):

PR #6143, opened 2024-04-30
Pull threads out of functions

Hdt80bro:
Several functions define anonymous thread functions like `ForkThread(function() ... end)` which are low-hanging fruit to adding customization points by naming them. These threads are something that, were a mod were to want to change, would require replacing the entire function in addition to the thread function. This is unnecessary on our part, and increases such a mod's stiffness - which I have seen cause at least some parts of one mod to cease to function. While this won't help with anything already written, it is possible to prevent future bit rot.

### DIFF
--- a/lua/SimPing.lua
+++ b/lua/SimPing.lua
@@ -8,14 +8,18 @@ MaxPingMarkers = 15
 --On first ping, send data to the user layer telling it the maximum allowable markers per army
 Sync.MaxPingMarkers = MaxPingMarkers
 
-function AnimatePingMesh(entity)
+---@param entity Entity
+---@param lifetime number
+function AnimatePingMesh(entity, lifetime)
     local time = 0
-    local ascending = true
-    while entity do
-        entity:SetScale(MATH_Lerp(math.sin(time), -.5, 0.5, .3, .5))
-        time = time + .3
-        WaitSeconds(.001)
+    while entity and lifetime > 0 do
+        entity:SetScale(MATH_Lerp(math.sin(time), -0.5, 0.5, 0.3, 0.5))
+        time = time + 0.3
+        WaitSeconds(0.1)
+
+        lifetime = lifetime - 0.1
     end
+    entity:Destroy()
 end
 
 function SpawnPing(data)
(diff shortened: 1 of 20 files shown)


lua/SimPing.lua
```diff
@@ -66,7 +66,10 @@ function SpawnPing(data)
             end
         end
 
-        ForkThread(function(owner) WaitSeconds(PingTimeout) PingsRemaining[owner] = PingsRemaining[owner] + 1 end, data.Owner)
+        ForkThread(function(owner)
```
Garanas, line 69: I think you missed this one; it's just formatted different

lua/defaultantiprojectile.lua
```diff
@@ -236,42 +236,13 @@ MissileRedirect = Class(Entity) {
...
-                    end
-
-                    -- aim at right below surface if unit is submerged
-                    enemyPos = enemyPos or projPos
-                    local surfaceHeight = GetSurfaceHeight(enemyPos[1], enemyPos[3]) - 0.02
-                    enemyPos[2] = math.max(surfaceHeight, enemyPos[2])
-
-                    proj:SetNewTargetGround(enemyPos)
-                end)
+                ForkThread(self.RedirectionThread, self, proj, self.Enemy)
```
Garanas, line 245: This is not functionally the same - by using the built in fork thread (proj:ForkThread) it also adds the thread to the trashbag of the projectile. The equivalent version would be:

```lua
ForkThread(self.RedirectionThread, self, proj, self.Enemy)
```

lua/defaultantiprojectile.lua
```diff
@@ -282,6 +253,41 @@ MissileRedirect = Class(Entity) {
             ChangeState(self, self.WaitingState)
         end,
 
+        RedirectionThread = function(self, proj, enemy)
+            local enemyPos = enemy and enemy:GetPosition()
```
Garanas, line 257: This may fail now! When a thread is forked is created it doesn't immediately execute. It is pushed to the back of the queue. By then the enemy may have been destroyed, and therefore the position may not be retrievable. It is better to pass the enemy and the position as separate parameters here.

lua/ui/game/gamemain.lua
```diff
@@ -624,9 +629,7 @@ function OnSelectionChanged(oldSelection, newSelection, added, removed)
         newSelection, changed = DeselectSelens(newSelection)
 
         if changed then
-            ForkThread(function()
-                SelectUnits(newSelection)
-            end)
+            ForkThread(SelectUnits, newSelection)
```
Garanas, line 632: Can you fork a c function? If I'm not mistaken then SelectUnits is not a Lua function

Garanas:
After running a few AI games it appears to be fine 👍

clyfordv:
Think we got a problem. I can't identify the reason but the changes to `DefaultProjectileWeapon.lua` throw this error when cheat spawning a unit that A. requires energy to fire and B. does not have `EnergyChargeForFirstShot` set to false (so Salvation, but not Mavor): 
image

Garanas:
Ah yes, it passes the wrong reference:

```lua
    -- Create an economy event for those weapons which require Energy to fire
    ---@param self DefaultProjectileWeapon
    StartEconomyDrain = function(self)
        if self.FirstShot then return end
        if self.unit:GetFractionComplete() ~= 1 then return end

        if not self.EconDrain and self.EnergyRequired and self.EnergyDrainPerSecond then
            local nrgReq = self:GetWeaponEnergyRequired()
            local nrgDrain = self:GetWeaponEnergyDrain()
            if nrgReq > 0 and nrgDrain > 0 then
                local time = nrgReq / nrgDrain
                if time < 0.1 then
                    time = 0.1
                end
                self.EconDrain = CreateEconomyEvent(self.unit, nrgReq, 0, time)
                self.FirstShot = true
                self.unit:ForkThread(self.EconomyDrainThread) -- <-- should be the weapon, not the unit
            end
        end
    end,

    ---@param self DefaultProjectileWeapon
    EconomyDrainThread = function(self)
        WaitFor(self.EconDrain)
        RemoveEconomyEvent(self.unit, self.EconDrain)
        self.EconDrain = nil
    end,
```

PR #6144, opened 2024-04-30
Fix gunships trailing behind quick units and not being able to engage them

Basilisk3:
Change the `CirclingRadiusChangeMaxRatio` and `CirclingRadiusChangeMinRatio` parameters of all relevant gunships to improve their behavior against large and fast units, naval units in particular.

**All Tech 2 and Tech 3 gunships, as well as the Jester:**
CirclingRadiusChangeMaxRatio 0.9 --> 0.7
CirclingRadiusChangeMinRatio 0.6 --> 0.5

**The SR:**
CirclingRadiusChangeMaxRatio 0.9 --> 0.8
CirclingRadiusChangeMinRatio 0.6 --> 0.5

## How to reproduce the problem:
1. Enable `NoDamage in the console`.
2. Spawn an Exodus Class Destroyer and 10 Vulthoos.
3. Order the Vulthoos to attack the Destroyer.
4. Move the Destroyer away.
5. Usually 3 to 6 gunships will follow the Destroyer without engaging it.

## Additional context
- The problem is very noticeable with naval units, but can also be observed with some land units, like the Titan, Loyalist, or the Harbinger. The issue is less prominent with slower units, but it still occurs.

## Test results
- Reducing the value of `CirclingRadiusChangeMaxRatio` alleviates the issue somewhat.
- A higher `MaxRadius` also helps.
- Tech 3 gunships also suffer from this, but it is less pronounced.
- Jesters can lose up to 70% of their DPS when engaging a retreating Mantis.
- There is a bug where some gunships are not engaging the Exodus, no matter what you change their stats to. This can only be fixed by giving them a manual move order and then another attack order.

## Checklist


Garanas:
Would be great to have this fixed 👍

PR #6154, opened 2024-05-04
Update Ythotha tread mark textures.

Fichom:
Updating the tread texture for Ythotha, as well as adjusting tread offset values in the unit blueprint.
image

Created a Ythoth and made it move to check if the treads are displayed correctly.

## Additional context
Tread marks don't show on all maps, probably because of terrain type.

## Checklist


Garanas (approved):
These look much better than what I made 👍

PR #6155, opened 2024-05-04
Draft version of veterancy effects for UEF and Cybran

Fichom:
Draft effects for veterancy for UEF and Cybran

https://github.com/FAForever/fa/assets/19254560/91d8556c-1234-4bd1-a663-27a92d29bccd

## Checklist


arch625:
This is cool, I'd love to see it added. I think the visual effect should be a bit more concentrated on the exact unit so when you have a larger army you can see more clearly which unit got the vet.

arch625:
In CNC3 I recall the unit just had the veterancy icon appear overhead when it reached a new rank, I think the same could work here, e.g. for UEF you just see the exact rank (⭐,⭐⭐, ⭐⭐⭐, etc.) appear above the unit with a little burst of color around it.

Garanas:
@Fichom the effect now spawns for UEF units when they gain veterancy. I've also scaled the effect based on the unit size, so that it's more concentrated (by the suggestion of @arch625 ). See also:

https://github.com/FAForever/fa/assets/15778155/b8f8b4cb-28dd-496c-ad63-0fbb98a58f31

From a testing perspective everything is the same - you can edit the effects table and it shows you how the effect would appear as if the unit gained veterancy.

BlackYps:

PR #6158, opened 2024-05-06
Remove the 'CommandCaps' field from various blueprints that do not need it

Basilisk3:
When these fields are set to `false`, they can be safely removed from the blueprint.

## Checklist


Basilisk3:
@Garanas I will remove the rest in another PR once this is merged. This one has become quite large already.

Garanas:
Do you use a tool to make these changes?

Basilisk3:
I search the unit folder for all instances I want to remove, and then replace them with nothing via the replace feature.

Garanas:
Fair enough 👍

Basilisk3:
CommandCaps are part of the .bp index of both unit-databases, but they do not show up in any calculations. Correct me if I am wrong, but this should not cause any issues.

MrRowey (approved):

PR #6159, opened 2024-05-07
Fix pause_unit_all and unpause_unit_all Hotkey Behavior

ChessBerry:
Fix the ToggleUnitPauseAll() and ToggleUnitUnpauseAll() functions and related hotkeys, pause_unit_all and unpause_unit_all, to actually behave as their name would suggest.

In the current version of fafdevelop, when having the construction or selection tab open, the ToggleUnitPauseAll() behaves identically to the ToggleUnitPause() function, pausing engineers in the selection only when all are currently unpaused, and the ToggleUnitUnpauseAll() function does the opposite of what it should _pausing_, instead of unpausing, everything in the selection. Both are fixed with this PR.

Pressing the pause button hotkeys with many different combinations of engineers, factories and other units, between all three UI tabs (construction, selection and enhancements).

## Additional context
I'm not sure since when these hotkeys have been bugged. 

I think I remember a time when pausing a unit via keyboard hotkey did pause/unpause the unit, but set the UI button "pause construction" to displaythe opposite of the actual unit state. It seems this was fixed at some point, and this may have broken these hotkeys unintentionally? I'm not sure, this is just a guess.

If this guess is correct though, it would also explain the if statement distinguishing between the construction and selection vs enhancements tabs, which seems useless in the current implementation. I haven't removed the if statements, because I'm not really sure what unintended consequences this could have, but we should check if this is possible.

## Checklist


Garanas:
@ChessBerry Can you add a changelog snippet as you find them in `/changelog/snippets`? That way this change will show up in the changelog.

This change will for sure confuse some people, haha.

ChessBerry:
@Garanas Done, I think.

PR #6149, opened 2024-05-02
Remove the 'BuildOnLayerCaps' field from the blueprints of all mobile units

Basilisk3:
## Description of the proposed changes
Closes #5913.

## Testing done on the proposed changes
All units I tested still functioned normally.

## Checklist
- [x] Changes are documented in the changelog for the next game version

PR #6161, opened 2024-05-09
Remove several unused fields from various blueprints

Basilisk3:
This PR removes the following fields from various blueprints:
`bp.General.TechLevel`
`bp.General.Classification`
`bp.General.Category`
`bp.General.UnitWeight`

## Additional context
According to an older Discord post by @The-Balthazar, these fields can be removed safely. 

## Checklist
- [x] Changes are documented in the changelog for the next game version (in #6149).

### DIFF
--- a/units/OPC1001/OPC1001_unit.bp
+++ b/units/OPC1001/OPC1001_unit.bp
@@ -47,7 +47,6 @@ UnitBlueprint {
             RULEUCC_Transport = false,
         },
         FactionName = 'Cybran',
-        TechLevel = 'RULEUTL_Basic',
     },
     Intel = {
         VisionRadius = 16,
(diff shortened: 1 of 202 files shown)


Garanas (commented):
To solidify https://github.com/FAForever/fa/pull/6171#pullrequestreview-2051301535, the field `Classification` is used by the Spooky unit viewer. I then had to fix that by 'guessing' the classification and it's not entirely correct yet:

- https://github.com/FAForever/spooky-db/pull/115

We already did a 'whoopsie' in https://github.com/FAForever/fa/pull/5169 , but point is that even though these values may not be used by the game they can still be read somewhere as they existed for eternity. Please verify if the unit viewers will be fine.

The-Balthazar:
I have always said that things relying on those fields are prone to the data being wrong because it's not actually used, and I think the fact that Spooky used to have to redefine the values of the Kennels is a prime example of that. But it's still important to make sure things aren't breaking.

Basilisk3:
@Garanas Yes, I can remake it at a later date.

PR #6162, opened 2024-05-09
Fix unit reclaim statistics in non-default unit layouts

PaletzTheWise:

There are three UI layouts one can cycle through using the alt-up and alt-down shortcuts.
 * unitview_mini - unit bars at the bottom and command bar on the left side. This is the default and works OK. OriginalMini 
 * unitview_right - unit bars at the bottom and command bar on the right side. This one reports lazyvars errors and does not display reclaim statistics in the unit tooltip.  [Right.log](https://github.com/FAForever/fa/files/15267435/Right.log) OriginalRight
 * unitview_left - vertical unit bars on the left side. This one also reports lazyvars errors and displays a white rectangle instead of reclaim statistics in the unit tooltip. [Left.log](https://github.com/FAForever/fa/files/15267437/Left.log) OriginalLeft

Root cause:
   * unitview_right is completely missing reclaim statistic UI positioning code.
   * unitview_left is missing the vast majority of reclaim statistic UI positioning code
   * Almost nobody uses the alternate layouts. Even I did so only because I hit the shortcut by accident and did not know how to go back.

Notable observations:
   * the unitview_x class selection is made when the game starts so in-game layout change (alt-up or alt-down) does not immediately apply the above characteristics. (This is also why the tooltip is not docking correctly until restart.)
   * The observer factory build queue tooltip probably works only in mini.

Changes:
  * Made cosmetic whitespace changes in unitview_mini for consistency.
  * Copied UI positioning code to unitview_left and unitview_right, made them look the same. The layout within the tooltip is the same for all layouts so this works.
    * In unitview_left there already was a line to set the group position, but the offset was different. It was the same as veterancy bar but the bar has text positioned above it so the position for the reclaim group from unitview_mini is, indeed, the correct one.
    * In unitview_left there was a line to set MassIcon color. This is not needed since we have an actual texture and, indeed, unitview_mini does not use it.
 * Removed duplicate unit tooltip Veterancy bar positioning code from unitview_left.lua, which was introduced in the same commit that introduced Reclaim statistics.

* For each layout
   * Selected the layout
   * Restarted the game
   * Verified the unit tooltip renders correctly for an engineer and a combat unit.
   * Cycled through layouts and verified the unit tooltip still renders correctly albeit in the wrong place for non-startup layouts.
   * Checked the logs and found no lazyvar issues.

FixedLeft
FixedMini
FixedRight

## Additional context

There probably should not be three copies of the same code but there are differences in the tooltip queue that need to be reconciled first.

## Checklist


lua/ui/game/layouts/unitview_left.lua
```diff
@@ -82,17 +84,27 @@ function SetLayout()
...
+
+    LayoutHelpers.FillParent(controls.ReclaimGroup.Debug, controls.ReclaimGroup)
+
+    controls.ReclaimGroup.Debug:SetSolidColor('00ffffff')
+    controls.ReclaimGroup.Debug.Depth:Set(-1000000)
 
-    LayoutHelpers.AtLeftTopIn(controls.ReclaimGroup, controls.bg, 192, 68)
-    LayoutHelpers.SetDimensions(controls.ReclaimGroup, 200, 200)
-    LayoutHelpers.FillParent(controls.ReclaimGroup.MassIcon, controls.ReclaimGroup)
-    controls.ReclaimGroup.MassIcon:SetSolidColor('ffffffff')
```
PaletzTheWise, line 94: SetSolidColor is not needed, because we have an actual texture.

Garanas:
@clyfordv given that you're working on similar files in https://github.com/FAForever/fa/pull/6148, would you be available to review this pull request?

clyfordv (approved):
Looks good to me. Test is good (rest of the surrounding non-standard layouts are still a mess of course).

P.S.--For future layout endeavors, check the Layouter. I'm just getting started with it myself, but aside from the performance improvements it makes apply multiple layout functions to the same element much cleaner. See an incomplete but functional example here: [lua/ui/game/layouts/construction_mini.lua](https://github.com/FAForever/fa/pull/6148/files#diff-fac90885c8508eebb88b85fd8e31b3457c270001f6bf4c44c3ba1549ccc29368R52)

PR #6164, opened 2024-05-10
Allow automatically loading constructed units into constructed transports

clyfordv:
Demo of two backend lua widgets for autoloading units in transports. Allows (for example) infinite construction of stingers carrying mech marines.

First widget: If a land factory is assisting an air factory, any non-engineer units from the land factory will be routed to assist the air factory instead of going to the land factories rally point.

Second widget: When an air factory finishes constructing a transport, it checks for the presence of any non-engineer land unit guards. This works independently of the first element; units can be manually ordered to assist a factory and will be loaded into transports in the same manner. Prototype only loads 1 unit per transport, but additional logic can check the size of the transport and load units as appropriate.

Stretch goals: give units assisting an air fac special formations, either a basic square or a more advanced arrangement where they group up with respect the air factory's build queue. Right now they play musical chairs as more units arrive/are loaded, which is workable but distracting and carries all the usual factory assist pathfinding issues.

https://github.com/FAForever/fa/assets/7411478/df2edc70-f0f5-42e0-b008-cca1eb489d9f

See discussion here: https://discord.com/channels/197033481883222026/1144605729911738439

### DIFF
--- a/lua/sim/commands/copy-queue.lua
+++ b/lua/sim/commands/copy-queue.lua
@@ -37,7 +37,7 @@ local dummyUnitTable = {}
 ---@param target Unit               # the unit that we read the queue of
 ---@param clearCommands boolean     # if true, copied orders are applied immediately
 ---@param doPrint boolean           # if true, prints the total distributed orders
-CopyOrders = function(units, target, clearCommands, doPrint)
+CopyOrders = function(units, target, clearCommands, queueOverride, doPrint)
 
     ---------------------------------------------------------------------------
     -- defensive programming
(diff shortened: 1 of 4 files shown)


Garanas:
About the feature: I like it, but I'm also feeling conflicted. I do certainly think it is interesting and novel and I also understand the value. While testing I noticed two problems:

- (1) At the moment only one unit attaches to the transport. That appears to be intentional though, given the Stinger example.
- (2) This only works when the land factory shares at least one blueprint that it can build with the air factory. When an Aeon land factory assists a UEF air factory then the assist order is cancelled. This is engine behavior and it makes sense from a code perspective but to the user it could be annoying.

> Stretch goals: give units assisting an air fac special formations, either a basic square or a more advanced arrangement where they group up with respect the air factory's build queue. Right now they play musical chairs as more units arrive/are loaded, which is workable but distracting and carries all the usual factory assist pathfinding issues.

While working on https://github.com/FAForever/fa/pull/5908 I noticed that units can have unique guard formations. The property is called `GuardFormationName` and it is in the AI table section. I assume links to a function name in `formations.lua`. I think you can experiment with that to create a better, alternative formation.

BlackYps:

PR #6166, opened 2024-05-10
Fix `AttackGroundTries` and Othuy Lightning Storm

lL1l1:

Fixes the ability of `AttackGroundTries` for bombers and fixes an off by one issue for other units. Annotates the field.
As a bonus, fixes lightning storm spawned from Ythotha too (cheat menu spawn and Ythotha spawn act differently).

## Checklist

### DIFF
--- a/lua/sim/weapons/DefaultProjectileWeapon.lua
+++ b/lua/sim/weapons/DefaultProjectileWeapon.lua
@@ -14,6 +14,11 @@ local UnitGetTargetEntity = UnitMethods.GetTargetEntity
 
 local MathClamp = math.clamp
 
+function LOGWeapon(self, string)
+    LOG(("T: %d WeaponId: %s %s"):format(GetGameTick(), tostring(self), string))
+    -- LOG(("T: %d HasTarget? %s"):format(GetGameTick(), self:WeaponHasTarget() and "true" or "false"))
+end
+
 ---@class WeaponSalvoData
 ---@field target? Unit | Prop   if absent, will use `targetPos` instead
 ---@field targetPos Vector      stores the last location upon which we dropped bombs for a target, or the ground fire location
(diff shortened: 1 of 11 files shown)


lua/sim/weapons/DefaultProjectileWeapon.lua
```diff
@@ -898,7 +899,7 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
 
                     -- wait reload time + 2 seconds, then force the weapon to recheck its target
                     WaitSeconds((1 / self.Blueprint.RateOfFire) + 3)
-                    self:ResetTarget()
+                    self:ResetTarget() -- this breaks AttackGroundTries for bombers
```
Garanas, line 902: That's a great find!

lL1l1:
The main change is the salvo retargeting feature.
An issue that was being solved before this PR is that salvos can be restarted from shot 0 if you cancel the firing through OnLostTarget in the middle of the FiringState.
Timeline:
1. Before this PR, HaltFireOrdered solved restarting salvos, but that causes AttackGroundTries to be off by one since the last shot isn't fired because it goes `ReadyState.OnFire -> FiringState.OnLostTarget` immediately (halt fire is ordered immediately) so the firing state doesn't get time to shoot.
2. In the state at 039dfe76b, this isn't possible, but an issue that pops up is that salvos will keep shooting even if the target is lost.
3. This makes me create the salvo retargeting feature which has to:
 - keep track of how much of a salvo has been fired across states
 - if there's an incomplete salvo, allow the weapon to restart the salvo immediately when it is allowed to fire instead of waiting for OnFire
 - To be honest I'm not sure if tracking targets in lua is required by the retargeting feature. I think I implemented that because before FAF's target resetting fix weapons were allowed to stay in FireReadyState forever. FAF's fix uses `ResetTarget` which calls `OnLostTarget` which changes the weapon state to idle/packing, unlike the vanilla behavior. That state change can happen constantly if the weapon has a target in tracking radius but not firing range, which I suppose can lead to unexpected animation or FX.

lL1l1:
I'm going to have to split the various changes in this PR.

PR #6169, opened 2024-05-11
Assisting engineers track new construction orders

clyfordv:
Additions:
1. When an engineer with assisting engineers starts a new construction in the middle of working on another, the assisting engineers will immediately switch to the new construction instead of staying on the old.
2. Engineers assisting UEF shoulder pods will now assist construction initated by the shoulder pods.
3. Assist orders for UEF shoulder pods are "sticky"--when the shoulder pod attaches to storage, the assist orders are transferred to the parent unit, and then transferred back when it detaches.

### DIFF
--- a/changelog/snippets/features.6169.md
+++ b/changelog/snippets/features.6169.md
@@ -1,0 +1,4 @@
+(#6169) Assisting engineers now track new construction
+- When an assisted engineer starts a new construction before the current one is complete, any assisting engineers will switch over immediately instead of completing the old
+- Engineers (and engineer pods) assisting UEF shoulder pods will assist with construction properly
+- Assist commands for UEF shoulder pods are "sticky"--when the shoulder pod attaches to its parent, the assist orders are transferred to the parent. When it detaches the assist orders are reapplied.
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -2719,7 +2719,6 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent) {
     --- the assisting unit to perform the same task as the unit it is assisting.
     ---@param self Unit
     CheckAssistFocus = function(self)
-
         --- Given engineer A that is assisting engineer B in doing some task. This function fixes the following situations:
         ---
         --- - (1) Engineer B is damaged. Engineer B starts the construction of a structure. Engineer A is repairing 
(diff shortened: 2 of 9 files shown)


clyfordv:
Looking for feedback on #3--I'm a little skeptical of the implementation, but in practice it's very smooth.

MostLostNoob:
> 1. When an engineer with assisting engineers starts a new construction in the middle of working on another, the assisting engineers will immediately switch to the new construction instead of staying on the old.

This seems like it could be troublesome as there will end up being orphaned unfinished structures since all the builders, even assisting ones, moved on to another project before the first/original was finished.

If ordering all assisting engineers to switch from the first/original project to the new project could be accomplished via hovering over the lead/main engineer & then using a key bind, then that might be a compromise as it will still be in the hands of the player if they want to change which project takes priority.

lL1l1 (changes requested):
Issues:
- Pods/engineers don't stop construction when the leader gets a stop order.
- Pods/engineers that are assisting other pods don't retarget when the main pod switches targets.
- Assist orders to pods don't go away when the sticky assist order is removed from the parent unit. Undock pod -> assist pod with engi -> dock pod -> remove assist order from parent -> undock pod (assist order will return, but logically it shouldn't)

> If ordering all assisting engineers to switch from the first/original project to the new project could be accomplished via hovering over the lead/main engineer & then using a key bind

I agree this would be a nice hotkey, similar to interrupt pathfinding. Currently you can accomplish the task of retargeting assisters by dragging the assist order and placing it back on top or by selecting all the engis and assisting the main one again, but of course these options don't work perfectly with multiple assists or engis close to each other assisting different units.

> Assist orders for UEF shoulder pods are "sticky"--when the shoulder pod attaches to storage, the assist orders are transferred to the parent unit, and then transferred back when it detaches.

It does look smooth in my testing, but is it necessary? It seems like it obfuscates assist orders overall and makes them harder to manage, since you visually lose track of the assist you placed on the drone for whatever special reason.

lL1l1:
Also assisting engineers don't cancel/retarget reclaim orders, even though they seemingly should since they immediately start assiting in reclaiming.

clyfordv:
- "Pods/engineers don't stop construction when the leader gets a stop order."
Will fix.

- "Pods/engineers that are assisting other pods don't retarget when the main pod switches targets."
Because the base assist functionality doesn't work with pods, we have to fake it with repair orders (like when a unit isn't at full health). Because `GetGuards` only retrieves units that have the guard order first in queue, fixing this would require doing a search through all units to see which ones fit the profile of "repair order" -> "assist pod order" (ouch). I didn't see a way around but let me know if you have any ideas.

- "Assist orders to pods don't go away when the sticky assist order is removed from the parent unit. Undock pod -> assist pod with engi -> dock pod -> remove assist order from parent -> undock pod (assist order will return, but logically it shouldn't)"
I wanted to avoid inspecting each potential assister individually for performance reasons, but this warrants it (do a check to make sure the cached assister is still guarding the parent unit). I'll try the changes then we can make the call.

The motivation for "sticky" pod orders is that A. I see people potentially getting a lot more use out of shoulder pods as independent actors if other engineers (but really mostly kennel pods) can assist them properly and B. it's a pain to send a shoulder pod off to build Template X, assign some kennel pods to assist, then have those assist orders disappear as soon as construction is complete and the shoulder pod redocks.

Another solution is giving the shoulder pod an assist order to the base station instead of the (silent) "AssistCommander" order they get now--that just prevents it from docking, and is arguably much cleaner.

clyfordv:
Accidentally did this on the main branch, closing and remaking.

PR #6171, opened 2024-05-11
Remove even more unnecessary fields from various blueprints (bp.Defense)

Basilisk3:
This PR removes the following fields from various blueprints:

`AirThreatLevel = 0,`
`RegenRate = 0,`
`SubThreatLevel = 0,`
`EconomyThreatLevel = 0,`
`SurfaceThreatLevel = 0`

## Checklist
- [x] Changes are documented in the changelog for the next game version (in #6149).


Garanas (commented):
The problem with removing fields that the game does not use is that external tools (such as websites that visualize the units) may not have safe guards in place to support missing values. In this pull request the regeneration rate would be such a candidate.

Specifically referring to:

- https://github.com/FAForever/spooky-db
- https://github.com/FAForever/UnitDB

Could you check and verify the code by searching for the blueprint field, and see if they properly allow the value to be missing?

A small example for the UnitDB is the fire rate that is missing on the Fire Beetle - it crashes the script.

Basilisk3:
> Specifically referring to:
> 
>     * https://github.com/FAForever/spooky-db
> 
>     * https://github.com/FAForever/UnitDB
> 
> 
> Could you check and verify the code by searching for the blueprint field, and see if they properly allow the value to be missing?
> 
> A small example for the UnitDB is the fire rate that is missing on the Fire Beetle - it crashes the script.

I am not familiar with either repo, so correct me if I am wrong but here are my findings:

The UnitDB has this check:
https://github.com/FAForever/UnitDB/blob/bbef3ca4c18145a04c43fc800b152813d8b74d12/www/res/scripts/functions.php#L1761C1-L1764C4

Spooky checks it too:
https://github.com/FAForever/spooky-db/blob/e19cc5f560889f0c175f6f84465590082a038679/app/views/unit.html#L13

Garanas (approved):

PR #6175, opened 2024-05-12
Rework UnitsTooltip to use Layouter to fix its UI scaling

### LINKED ISSUE #5424: Tooltip is clipped in Unit Manager
Tooltip is clipped in Unit Manager when hovering over an unit with long description

Open Unit Manager (Restrictions) in lobby options
Hover over an unit with long description, e.g. SML
Observer with of tooltip

image

lL1l1:

Fixes #5424 by rewriting unit tooltips using the Layouter. In addition, fixes localized empty strings appearing as empty parentheses in unit names (for example UEF mex in RU/US were "T3 Mass Exctractor ()"), and changes the buildtime display to just buildtime ticks instead of a convoluted MM:SS format.
100%: 
image
80%: 
image
150%:
image

Go to options -> interface -> UI Scale -> 150% and then open a skirmish -> options -> restrictions and mouse over the units in a circle to make sure all 4 variants of the tooltip work as expected (variants that move which corner of the tooltip is attached as to avoid going off the screen). Repeat for 80% and 100% scale.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6175.md
+++ b/changelog/snippets/fix.6175.md
@@ -1,0 +1,4 @@
+- (#6175) Rework the unit manager's unit tooltips using the Layouter with various fixes:
+  - Fix UI Scaling for unit tooltips in the unit manager.
+  - Display buildtime in ticks instead of MM:SS in the unit manager.
+  - Fix empty names appearing as empty parentheses in the unit manager.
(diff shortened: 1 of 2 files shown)


Garanas (approved):
This is looking much better 👍 

Old:

image

New:

image

PR #6177, opened 2024-05-12
Add support for generating the wiki with BrewWikiGen

Garanas:

- (1) Removes FA-Lua specific syntax from files that are shared with the generator script
- (2) Adds missing unit icons
- (3) Adds missing unit strategic icons
- (4) Creates a setup to generate the input required to generate the Wiki

## Checklist

- [x] Find a proper fix for e6b287f2775e292a98bdac617aac78d319dc11e8


Garanas:
With thanks to @The-Balthazar for fixing and extending the [wiki generator](https://github.com/The-Balthazar/BrewWikiGen) to support our setup here at FAForever!

PR #6179, opened 2024-05-13
Ilshy Turn Rate Buff

MrRowey:
Improve the microabilty of the ilshy

```
TurnRate: 45 > 90
```

### DIFF
--- a/changelog/snippets/balance.6179.md
+++ b/changelog/snippets/balance.6179.md
@@ -1,0 +1,1 @@
+- (#6179) Increase Ilshavoh's max turn rate from 75 to 90 and on the spot turn rate from 45 to 90.
--- a/units/XSL0202/XSL0202_unit.bp
+++ b/units/XSL0202/XSL0202_unit.bp
@@ -198,7 +198,7 @@ UnitBlueprint{
         MotionType = "RULEUMT_Land",
         RotateOnSpot = false,
         TurnRadius = 2,
-        TurnRate = 45,
+        TurnRate = 90,
     },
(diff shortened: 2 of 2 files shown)


Garanas:
Wouldn't this make it a very strong unit? It has more range, health and damage output than the average counter part - the only downside is that it's more difficult to micro.

If the idea is to make it better, what if instead of 90 we go with 60?

MrRowey:
https://github.com/FAForever/fa/assets/20344151/07416307-1074-44da-9eb9-2029be920b3a

As Farms Shows here how one when trying to micro/ turn the ilsy it looks complete odd and 2 is really slow

lL1l1:
Ilshie is weak vs ACU (teamgames), and strong vs other units (1v1). Other units can catch up with ilshies pretty easily due to the speed difference and fight for a long time so the turn rate doesn't have a huge impact. In stark contrast, ACUs are slow and their overcharge is a one shot kill so the turn rate can help a lot with avoiding them.
The actual difficult part of ilshies when fighting other units is that they're slow and expensive, so they're more prone to getting swarmed and bad at catching raids.

There's also the nuance that `TurnRadius = 2` for ilshie gives it 74.5 turn rate when the nav waypoint is further than 2 away (just set TurnRadius to 0 to see how slow 45 TurnRate actually is), so your gut instinct to go with a small +15 buff would result in 75 -> 90 turn rate and not 45 -> 60.

lL1l1 (approved):
From an aesthetic POV, nothing seems wrong with buffing the turn rate.

PR #6185, opened 2024-05-14
Salvation Undershoot Fix

clyfordv:
Issue:
Salvation undershoots, sub-fragment and sub-sub-fragment code is generally brittle and not blueprint dependent.

Fixes:
1. Adds a `FragmentRadius` blueprint value to the AIFFragmentationSensor projectile blueprints, and links the fragment initial velocity to the blueprint `InitialSpeed` value.
2. Calculates the ballistic acceleration of the (progressively slower) sub-fragments so they don't fall short of the target.

Tangentially related fixes:
- Fixes an error where the wrong reference was passed to the Econ thread in `DefaultProjectileWeapon.lua`

### DIFF
--- a/changelog/snippets/fix.6185.md
+++ b/changelog/snippets/fix.6185.md
@@ -1,0 +1,1 @@
+(#6185) Fix Salvation Undershooting - reworks Salvation projectile code to not drop short and use blueprint values for speed and fragment spread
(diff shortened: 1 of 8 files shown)


MrRowey (approved):

MrRowey:
@clyfordv make sure to make a changelog snipit

PR #6188, opened 2024-05-15
Feature - Add github workflows for wiki generation

relent0r:
This PR adds workflows for the generation of wiki content.

At a high level it does the following.

- Checks out the following repositories
  The-Balthazar/BrewWikiGen
  FAForever/fa - Sparse checkout
  FAForever/fa.wiki
- Installs the ImageMagick software
- Performs an image conversion of unit/stragetic icons from .dds to .png
- Copies the wiki Run.lua to replace the Brewlan one
- Performs the brewlan wiki generation via Run.lua
- Pushes the generated content to the wiki repostiory

It is currently in draft pending functional testing and some more inline documentation. Requires changing repository references to official ones for production implementation.

Testing without production impact requires updating the repositories used to forked variants

## Checklist


.github/workflows/wiki-generate-units.yaml
```diff
@@ -0,0 +1,90 @@
+name: FAF Wiki Generator for unit icons
+
+# not sure if we will use a trigger or if this will be manual? It could trigger on release
```
Garanas, line 3: We'll start with a manual dispatch. Once we're certain it works then we'll probably want to trigger it for each commit that ends up on `deploy/fafdevelop`

.github/workflows/wiki-generate-strategicicons.yaml
```diff
@@ -0,0 +1,91 @@
...
+            luaVersion: "5.4"
+
+        - name: Install Image Magick
+          run: |
+            sudo apt-get install -y imagemagick
+
+        # copy strategic and unit icons and convert them to PNGs. Assume this should be going to the wiki location not the fa repo
+        - name: Convert Strategic Icons
+          run: |
+            if [ ! -d "fa.wiki/icons/strategicicons" ]; then
```
Garanas, line 55: You should be able to just call the `.sh` file that is in the repository

.github/workflows/wiki-generate-strategicicons.yaml
```diff
@@ -0,0 +1,91 @@
...
+          with:
+            luaVersion: "5.4"
+
+        - name: Install Image Magick
+          run: |
+            sudo apt-get install -y imagemagick
+
+        # copy strategic and unit icons and convert them to PNGs. Assume this should be going to the wiki location not the fa repo
+        - name: Convert Strategic Icons
+          run: |
```
Garanas, line 54: When running a shell, make sure to specify what shell via the `shell` property. See also:

- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrunshell

.github/workflows/wiki-generate-units.yaml
```diff
@@ -0,0 +1,90 @@
...
+            luaVersion: "5.4"
+
+        - name: Install Image Magick
+          run: |
+            sudo apt-get install -y imagemagick
+
+        # copy strategic and unit icons and convert them to PNGs. Assume this should be going to the wiki location not the fa repo
+        - name: Convert Unit Icons
+          run: |
+            if [ ! -d "fa.wiki/icons/units" ]; then
```
Garanas, line 56: When updating the unit blueprints there should be no need to also generate the icons. The wiki gen will use those that are in the wiki already.

.github/workflows/wiki-generate-strategicicons.yaml
```diff
@@ -0,0 +1,91 @@
+name: FAF Wiki Generator for strategic icons
+
+# not sure if we will use a trigger or if this will be manual? It could trigger on release
```
Garanas, line 3: Since unit icons do not change much we'll likely only call this manually.

.github/workflows/wiki-generate-strategicicons.yaml
```diff
@@ -0,0 +1,91 @@
...
+              textures/ui/common/icons/units
+
+        - name: Checkout FAF Wiki Repository
+          uses: actions/checkout@v4
+          with:
+            repository: FAForever/fa.wiki
+            ref: master
+            path: ./fa.wiki
+
+        - name: Install Lua 5.4
```
Garanas, line 43: There should be no need for Lua 5.4 in this workflow. We do not want to run the generator; just update the icons that are in the wiki.

.github/workflows/wiki-generate-strategicicons.yaml
```diff
@@ -0,0 +1,91 @@
...
+
+
+
+jobs:
+
+  prepare-environment:
+      runs-on: ubuntu-latest
+
+      steps:
+        - name: Checkout Brewland Wikigen Repository
```
Garanas, line 17: There should be no need to checkout this repository. We only want to update the (strategic) unit icons.

.github/workflows/wiki-generate-strategicicons.yaml
```diff
@@ -0,0 +1,91 @@
...
+            fi
+
+            mogrify -path "fa.wiki/icons/strategicicons" -format png "fa/textures/ui/common/game/strategicicons/*.dds" -verbose
+
+        - name: Convert Strategic Icons
+          run: |
+            ls "fa/wiki"
+
+        - name: Replace run.lua # tbd
+          uses: canastro/copy-file-action@master
```
Garanas, line 67: You can just use the `mv` bash command to move files from one place to another, no need to install an action for this.

.github/workflows/wiki-generate-units.yaml
Garanas: Or alternatively, each night if it detects changes any changes to `deploy/fafdevelop`. We'll need to see how many minutes it will cost us to run this once we're done!

wiki/Run.lua
```diff
@@ -8,20 +8,52 @@
 --[[ ---------------------------------------------------------------------- ]]--
 local OutputDirectory = "D:/faf-development/fa.wiki/"
 local WikiGeneratorDirectory = "D:/faf-development/BrewWikiGen/"
+local FADirectory = "D:/faf-development/fa/"
+
+-- This section deals with overriding the OutputDirectory and WikiGeneratorDirectory if required
```
Garanas, line 13: Looks like a clean solution 😃 !

.github/workflows/wiki-generate-units.yaml
relent0r: I haven't tried it yet but we could try make it trigger on specific tags?

Garanas:
And on a different note to make testing easier: you can output the entire wiki as an artifact when the job is done. See also the non-wiki changelog generation script that adds the changelog as an artifact. That way you can download and verify the output.

.github/workflows/wiki-generate-strategicicons.yaml
relent0r: ah yes i wrote this before I realized you already had scripts for it :D was educational tho.

.github/workflows/wiki-generate-strategicicons.yaml
relent0r: ah is there a processing cost to using custom actions for things that you can otherwise easily do with shell?

.github/workflows/wiki-generate-units.yaml
Garanas: Yes, there are various triggers available including labels:

- https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#using-activity-types

But for now, for this pull request we'll focus on the manual dispatch. When and where we run the workflow can be a separate discussion from the workflow working properly 😃 !

.github/workflows/wiki-generate-strategicicons.yaml
Garanas: Yes - the cost to retrieve them and overhead that it may apply through the script. There's also a security concern, especially relevant for private repositories of corporations. This is why in the settings of a repository (or at an organization level) you can limit what external actions are allowed. 

In the end an external action can do anything, including just shipping your entire repository as a post to some URL that you do not control. Therefore it's not just a 'keep it simple' principle, but also a security concern to use actions that you may not trust.

image

.github/workflows/wiki-generate-units.yaml
```diff
@@ -0,0 +1,90 @@
...
+            ref: master
+            path: ./brew-wiki-gen
+
+        - name: Checkout FAF Repository # -png folder doesnt exist yet, confirm location.
+          uses: actions/checkout@v4
+          with:
+            repository: relent0r/fa
+            ref: feat/github-wf-icons
+            path: ./fa
+            sparse-checkout: |
```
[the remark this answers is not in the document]
The-Balthazar: It loads and uses unit and projectile blueprints, and it loads the lod0 models of units to scan for attach points for the sake of transports and landing pads. It doesn't touch dds files at all, and it also doesn't touch unit/projectile scripts.
It then also loads:
loc/'..Language..'/strings_db.lua (The generator is only around half set up for non-EN translations though)
lua/ui/help/unitdescription.lua
lua/ui/help/tooltips.lua
lua/sim/AdjacencyBuffs.lua
lua/system/Blueprints.lua

.github/workflows/wiki-generate-units.yaml
The-Balthazar: There's a non-0 chance that converting icons needs to gets done at most yearly.

.github/workflows/wiki-generate-units.yaml
relent0r: There is a note in the documentation here.
Note: This event will only trigger a workflow run if the workflow file is on the default branch.

Since this is a PR branch I have a feeling that I can't trigger the workflow for testing due to this not being on the default branch. At least I don't see the run workflow option.

wiki/Run.lua
```diff
@@ -8,20 +8,52 @@
...
+
+print("Directories set")
+print("Output Directory: " ..OutputDirectory)
+print("Wiki Generator Directory: " ..WikiGeneratorDirectory)
+print("FA Directory: " ..FADirectory)
 
 EnvironmentData = {
     name = 'Forged Alliance Forever',
     author = 'Gas Powered Games',
     version = '1.6.6',
```
The-Balthazar, line 47: Version should be the current dev branch FAF version, 1.6.6 is the Steam version, and the Author should us; `The FAF development team` or something.

.github/workflows/wiki-generate-units.yaml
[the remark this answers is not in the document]
relent0r: I've refined the sparse checkout with the above suggestions.

.github/workflows/wiki-generate-strategicicons.yaml
relent0r: Ah nice, thanks for pointing it out I wasn't aware of the security risk.

relent0r:
> And on a different note to make testing easier: you can output the entire wiki as an artifact when the job is done. See also the non-wiki changelog generation script that adds the changelog as an artifact. That way you can download and verify the output.

Thats good to know thanks. 

I've done updates based on the above feedback.

Outstanding is the question about how to get the current version and how I could put this into the commit message for the wiki. 

I've tested up to this point with the updates and it 'seems' good. I just need to update the triggers and repo references once we are happy to move forward.

wiki/icons-convert-strategic.sh
```diff
@@ -22,16 +22,26 @@
 #** SOFTWARE.
 #******************************************************************************************************
 
-if [ -d "wiki/generated" ]; then
-    if [ -d "wiki/generated/strategicicons" ]; then
-        rm -rf "wiki/generated/strategicicons"
+# Process named arguments
```
Garanas, line 25: The changes to this and `wiki/icons-convert-unit.sh` should be reverted. They expect the repository as their working directory, which you can define in the [script as a property](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#defaultsrunworking-directory). 

Then the script by default populates the `wiki/generated` folder which is intentional. It's not the correct position; but that's not the goal of the script. The goal of the script is to create/convert the assets. You can then use additional, separate shell commands to move those assets to where you need them to be. This approach allows you to re-use the script over different workflows. In this workflow we use it to populate the wiki, in the next workflow we may use it to populate the client (hint, hint to Sheikah if you happen to read this).

Note that the generated folder is also included in the gitignore and therefore you can even run the scripts local on your computer.

.github/workflows/wiki-generate-units.yaml
```diff
@@ -0,0 +1,90 @@
...
+          run: |
+            $GITHUB_WORKSPACE/fa/wiki/icons-convert-unit.sh -w fa.wiki -f fa
+
+        - name: Replace run.lua
+          run: |
+              sudo mv fa/wiki/Run.lua brew-wiki-gen/Run.lua
+
+        - name: Execute lua run
+          run: |
+            lua brew-wiki-gen/Run.lua --OutputDirectory="fa.wiki/" --WikiGeneratorDirectory="brew-wiki-gen/" --FADirectory="fa/"
```
Garanas, line 71: See also: https://github.com/FAForever/fa/pull/6188/files#r1603974326

.github/workflows/wiki-generate-units.yaml
```diff
@@ -0,0 +1,90 @@
...
+            ref: master
+            path: ./brew-wiki-gen
+
+        - name: Checkout FAF Repository # -png folder doesnt exist yet, confirm location.
+          uses: actions/checkout@v4
+          with:
+            repository: relent0r/fa
+            ref: feat/github-wf-icons
+            path: ./fa
+            sparse-checkout-cone-mode: |
```
Garanas, line 31: Please add a comment as to why we checkout these folders and not other folders

.github/workflows/wiki-generate-strategicicons.yaml
```diff
@@ -0,0 +1,59 @@
...
+            path: ./fa.wiki
+
+        - name: Install Image Magick
+          run: |
+            sudo apt-get install -y imagemagick
+
+        # copy strategic and unit icons and convert them to PNGs. Assume this should be going to the wiki location not the fa repo
+        - name: Convert Strategic Icons
+          run: |
+            $GITHUB_WORKSPACE/fa/wiki/icons-convert-strategic.sh -w fa.wiki -f fa
```
Garanas, line 42: This is too complicated; it should suffice to:

- (1) Set the working directory to the root of the `fa` repository
- (2) Call the script (`./wiki/icons-convert-strategic.sh`)
- (3) Move the created/converted assets with [mv](https://en.wikipedia.org/wiki/Mv_(Unix))

See also: https://github.com/FAForever/fa/pull/6188/files#r1603974326

wiki/icons-convert-strategic.sh
relent0r: Ah ok I didn't realize you could do that. I've reverted the argument code.

Based on testing I still need to make changes to these scripts though.
1. The version of magick that is installed in linux doesn't have a magick executable. Just a convert and mogrify one.
2. If the read command is at the end of the script it will fail the pipeline.

.github/workflows/wiki-generate-units.yaml
relent0r: How would you do this? The changes to the script don't have any impact if running locally since it will just use the defaults if the arguments are not specified. But given there are 3 separate locations that are being used (compared to the other where it is only interested in the current directory) I'm unsure how to avoid it without statically setting folder names in the script.

relent0r:
Moving out of draft as this is at a v1 point.

.github/workflows/wiki-generate-units.yaml
Garanas: You're right; it may be better to just use the inputs here.

PR #6192, opened 2024-05-15
Fatboy unit detach bug + exFac footprint fix

clyfordv:
Issues:
1. Ordering "deploy" on the Fatboy will detach the unit under construction
2. External factory dummy anchorUnits had a lingering footprint in the blueprint that blocked structure construction (minor, but aesthetically bad)

Fixes:
1. Add a check in the fatboy script to make sure the detaching unit is complete--this could be generalized, but the fatboy is the only unit with this issue, and I believe it also has specific factory attach bone requirements
2. Trim the anchor unit blueprint to remove footprint + additional unnecessary values

Afterthoughts:
1. Anchor dummy unit still necessary to prevent selecting the exFac from the worldview


units/ZXA0003/ZXA0003_unit.bp
```diff
@@ -30,33 +30,18 @@ UnitBlueprint{
     Economy = {
```
Garanas, line 30: Are the changes in this file related to the fix?

clyfordv: Aye, specifically `SizeX`, `SizeY`, and `SizeZ`--if they're not zero the dummy unit shows a structure footprint that blocks construction. The rest were just unnecessary fields.

changelog/snippets/fix.6192.md
```diff
@@ -0,0 +1 @@
+(#6192) Removes an erroneous exFac dummy unit structure footprint, and prevents the fatboy from detaching units under construction when "deploy" is ordered.
```
Garanas, line 1: Can you turn this into two lines, e.g.:

(#6192) Removes an erroneous structure footprint from the dummy unit of external factories

(#6192) Fix a bug where deploying when building a unit in the external factory would detachg the unit under construction 

This primarily applies to the Fatboy.

lL1l1 (commented):
The unit inside the mobile factory still blocks construction.

clyfordv:
I was unaware of that, but I'm guessing it has to do with the build area needing to be free etc. etc. We could chase that goose but I think the return would be minimal, I was mostly worried about the little square popping up (which looked scuffed).

Garanas (approved):

PR #6194, opened 2024-05-18
Fix AI rating for sub-1.0 ratings

maudlin27:
##Issue
If a sub-1.0 AiX modifier is chosen, the rating impact was modified by 1 / (AiX - 1) previously.  So for example, a 0.9 rating modifier would have 10 times the normal modifier applied as a negative amount, wheras a 0.6 modifier would have 4 times the normal modifier applied as a negative amount.  In other words, a 0.9 AiX would have a far worse rating than a 0.6 AiX.

This removes the code applying this - there is no need for such a modification.

Created a lobby and used 0.9 and 0.6 AiX modifiers.

For example, with this change, M28's ratings are (on a 5km map):
0.9 AiX: 521 rating
0.6 AiX: 156 rating

Without this change, M28's ratings are:
0.9 AiX: -3279 rating
0.6 AiX: -595 rating

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -493,14 +493,7 @@ function ComputeAIRating(gameOptions, aiLobbyProperties)
     local cheatBuildMultiplier = (tonumber(gameOptions.BuildMult) or 1.0) - 1.0
     local cheatResourceMultiplier = (tonumber(gameOptions.CheatMult) or 1.0) - 1.0
 
-    -- if they're smaller than 1.0 then the AI doesn't get better; it gets worse!
-    if cheatBuildMultiplier < 0 then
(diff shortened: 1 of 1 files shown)


Garanas:
I'm trying to think of the original reasoning of that snippet of code. Do you remember anything about it @maudlin27 ?

maudlin27:
No, my recollection was it was something you drafted and I thought 'ah because the numbers are negative it wouldn't work properly hence why Jip's added that', while I only tested with one negative number (probably -0.5) which gave a sensible result

maudlin27:
Another possibility is it was to try and reflect how a 50% reduction is equivalent to a 100% increase

Garanas:
Great find @maudlin27 , thank you for taking the time to look into it

PR #6200, opened 2024-05-20
commandmode.lua callback refactor

clyfordv:
Issues:
- `OnCommandIssued` is a long list of if else statements--aside from being generally unorganized it makes modifying the logic for different commands more difficult for modders.
- Minor detail: attack move commands give the default feedback blip instead of the attack feedback blip

Changes:
- Split the logic for the different `OnCommandIssued` sections into callback functions, and access those functions with a table keyed with the command type ('Repair', 'Reclaim', 'Attack' etc.).
- Modify the feedback blip for attack move to match its order line color
- Adds a SkipBlip field for the UserCommand table that allows callbacks to skip the feedback blip (like for build mobile, where a special blip is displayed corresponding to the mesh of the ordered blueprint).

Notes:
- Mod compatibility should remain, and hooking/modifying functions for the different order types should be easier


Garanas:
Great initiative 👍 

Would it be interesting to add the remaining callbacks, but have them be empty?

Garanas:
One interesting problem is that you can no longer hook these functions 'the usual way'. You'd also need to overwrite the definition in `OnCommandIssuedCallback` in order for the hook it to work.

One solution to that is to put them in a separate file. What are your thoughts on this?

clyfordv:
The hook would need to (with the current setup) look something like:

```lua
oldOnGuardIssued = OnGuardIssued

OnCommandIssuedCallback.OnGuardIssued = function(command)
    $NEW_LOGIC
    oldOnGuardIssued(command)
end
```
?

To me it isn't terribly different from hooking a class. I'm on the fence about putting it in a separate file for readability reasons, but the above doesn't seem so bad to me. We could also add a comment to the callback table indicating the proper hook format.

Garanas:
You're right; maybe I am a bit too anxious about it. Could you add a bit of documentation on how to hook these functions? Then we can merge this in.

PR #6205, opened 2024-05-21
Remove the 'CommandCaps' field from various blueprints that do not need it (part two)

Basilisk3:
When these fields are set to `false`, they can be safely removed from the blueprint.

## Checklist


Basilisk3:
@Garanas This one can be merged too now👍

Garanas:
I think these changes are also processed in https://github.com/FAForever/fa/pull/6271 , are you fine if I close this instead?

Basilisk3:
Yes, that is fine.

PR #6207, opened 2024-05-22
Fix free intel not being toggleable

lL1l1:

Adds 1 energy cost for spy plane jamming to make the jamming retoggling feature work with spy planes.

Spawned a radar for myself and a UEF spy plane on the enemy team. Toggled the full map vision cheat on then off to clear the spy plane's jammer blips. After some time, the blips return.

## Additional context

I looked into the `defaultcomponents.lua` `DisableUnitIntel` function and how the intel status is created in `blueprint-units.lua` and it seems like there is no way to create a free but toggleable intel since the table `AllIntel` isn't filled if all intel is free, but `DisableUnitIntel(disabler, intelType)` requires the `intelType` to be in `AllIntel`.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6207.md
+++ b/changelog/snippets/fix.6207.md
@@ -1,1 +1,3 @@
-- (#6207) Fix UEF spy plane's jammer blips never reappearing after being seen.
+- (#6207) Fix free intel not being toggleable, which lead to UEF spy plane's jammer blips never reappearing after being seen.
+
+- (#6207) Fix Cybran ACU and SACU stealth not being toggleable.
--- a/changelog/snippets/fix.6207.md
+++ b/changelog/snippets/fix.6207.md
@@ -1,0 +1,1 @@
+- (#6207) Fix UEF spy plane's jammer blips never reappearing after being seen.
--- a/engine/Core/Blueprints/UnitBlueprint.lua
+++ b/engine/Core/Blueprints/UnitBlueprint.lua
@@ -987,7 +987,6 @@
 ---@field AllIntel table<IntelType, boolean>
 ---@field AllIntelRecharging table<IntelType, boolean>
 ---@field AllIntelMaintenanceFree table<IntelType, boolean>
----@field AllIntelFromEnhancements table<IntelType, boolean>
 ---@field AllIntelDisabledByEvent table<IntelType, table<string, boolean>>
 
 ---@class UnitBlueprintIntel
--- a/lua/defaultcomponents.lua
+++ b/lua/defaultcomponents.lua
@@ -124,8 +124,11 @@ IntelComponent = ClassSimple {
                     allIntelFromEnhancements[intel] = true
                 end
 
-                allIntelDisabledByEvent[intel] = allIntelDisabledByEvent[intel] or {}
-                if not allIntelDisabledByEvent[intel][disabler] then
+                local allIntelDisablers = allIntelDisabledByEvent[intel]
+                if not allIntelDisablers then
+                    allIntelDisabledByEvent[intel] = {}
+                end
+                if not allIntelDisablers[disabler] then
                     allIntelDisabledByEvent[intel][disabler] = true
                     self:DisableIntel(intel)
                     self:OnIntelDisabled(intel)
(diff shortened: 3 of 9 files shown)


Garanas:
How much does the Cybran Spy Plane consume for its maintenance? We could use that value instead of `1`, as `1` looks odd

lL1l1:
The Cybran spy plane doesn't consume anything since stealth never needs to be toggled, but UEF frigates use 5 energy.

lL1l1:
What about allowing setting custom status from the bp? Currently it is completely overwritten by lua blueprint modification. This would allow placing Jamming in `AllIntel` and omni/radar into `AllIntelMaintenanceFree` manually.

Garanas:
I'm not in favor of adjusting the blueprint process related to intel. The process is complicated and difficult to verify given that it also has to deal with numerous units from mods. If we can get away with just giving a unit an energy maintenance of 5 then that would be my preferred solution here.

clyfordv (commented):
I think if we just add a check for a non-energy disabler here:
image

... and here:
image

It fixes the problem. The intel we want isn't being added to `allIntel`/being disabled because it doesn't consume energy, disabling it in cases where energy consumption isn't the issue feels pretty watertight.

lL1l1:
Yeah that's a better fix. Currently it is impossible to disable free intel entirely, which you might want to do for intel types other than jamming if you're modding some custom intel resource.

lL1l1:
I upvalued and optimized the intel logic quite a bit, it should work the same way as before. Tests:
- `EnableUnitIntel(disabler)` Cybran spy planes (`MaintenanceFreeIntel`) and Cybran ASF (`AllIntel`) landing in a carrier (carriers disable intel with the `"cargo"` disabler).
- `DisableUnitIntel(disabler)` Ctrl+k the Cybran spy planes and ASF
- `DisableUnitIntel(disabler, intel)` and `EnableUntilIntel(disabler, intel)` Toggle an ASF (`AllIntel`) and a UEF spy plane (`MaintenanceFreeIntel`)
- `DisableUnitIntel("Energy")` and `EnableUnitIntel("Energy")` Spawn some t3 pgens and a bunch of omnis to energy stall, then disable some of the omnis to stop energy stalling (`AllIntel`). While energy stalling, fly a spy plane near some civilians and see that they do show up on radar (`MaintenanceFreeIntel`).
- Enhancements: Spawn a cybran cloak preset SACU in enemy radar range. Turning off the cloaking makes it show up on radar, and turning it back on makes it disappear from enemy radar. Same for full map cheat vision from an enemy perspective.
  - I'm not sure why enhancements needed to be tracked in the first place. Is it for mod compatibility?

lL1l1:
By the way, @clyfordv you can link files and their line numbers as a preview inside a comment by going to the file page, clicking+shift clicking the line numbers to select a line/range of lines, and then getting a permalink from the options menu. I saw you having trouble with that in another comment too so I thought I'd mention it.
image
The preview is this:
https://github.com/FAForever/fa/blob/87e6fb0ed883e404a407d727bc3baddf95341018/lua/defaultcomponents.lua#L198-L212

clyfordv:
I ran all your tests, everything is good except for the cybran SACU:
1. Spawn an un-upgraded SACU
2. Upgrade to personal stealth
3. It still turns off when e stalling, but toggling it manually has no effect (energy consumption and stealth effect stays).

So looks like that enhancement section probably did something after all.

lL1l1:
All fixed. The enhancements section had nothing to do with that, the CCommandUnit script was just overriding the script bit behavior entirely instead of hooking, which broke the stealth toggle.

Garanas:
I'll be honest - these are scary changes because it's a complicated system. If you both trust the changes then feel free to approve and merge them. But please keep checking intel continuously once this is merged in.

clyfordv:
1. I asked Balthazar about the enhancementIntel and he said he had never seen it in his life.
2. There's a lot of green in the changed files, but I think most of it is formatting. Perhaps the formatting/performance improvements would be best separated from the bugfix.
3. I think we're good to go.

lL1l1:
@clyfordv If it's ready to go can you approve it so it can be merged?

clyfordv (approved):
Tested, good to go, only feedback is to separate future logic/bugfix changes from performance/formatting improvements for easier reviewability.

PR #6213, opened 2024-05-24
Revert change in default teleport time

lL1l1:
Reduces the default teleport delay set in #5907 from 15 seconds to 0 seconds to return old behavior for modded units. I noticed this with BlackOps ACUs, which used to have a lightning fast and very cheap 1.4 second teleport.

After the changes, the quick teleport time is back.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6213.md
+++ b/changelog/snippets/fix.6213.md
@@ -1,0 +1,1 @@
+- (#6213) Fix the default teleport delay being set to 15 seconds instead of 0, which caused some modded units to teleport much slower.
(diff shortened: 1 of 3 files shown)


MrRowey:
Can you please state how this will now effect the base balance if the game and surly the mod should be updated rather than reversing a new changes that the balance team requested.

lL1l1:
FAF ACUs and SACUs have a teleport delay defined in their blueprint, so they aren't affected by these changes.

MrRowey (approved):

PR #6215, opened 2024-05-25
Add IsAlly check to MobileUnit OnStartBuild

### LINKED ISSUE #2725: building opponent structures
if you are the same faction as your opponent you can build his structure if you ordered to build the same building at the same place in the fog of war (without you scouting his building).
This mostly happen with mexes.

clyfordv:
fixes #2725
Prevent mobile units (engineers) from building enemy buildings. This typically happens when two engineers of the same faction are given a build order to the same resource spot.

Spawn two engineers of a given faction, give them an order to build a mex/hydro on the same spot. Previously whichever engineer arrived second would assist with constructing the enemy structure; now they clear their command queue instead.

### DIFF
--- a/changelog/snippets/fix.6215.md
+++ b/changelog/snippets/fix.6215.md
@@ -1,0 +1,1 @@
+- (#6215) Fix an edge case where an engineer can assist an enemy construction task if they were given an order to build the same unit in the same location before the enemy task was started.
--- a/lua/sim/units/MobileUnit.lua
+++ b/lua/sim/units/MobileUnit.lua
@@ -126,6 +126,20 @@ MobileUnit = ClassUnit(Unit, TreadComponent) {
     end,
 
     ---@param self MobileUnit
+    ---@param built Unit
+    ---@param order string
+    ---@return boolean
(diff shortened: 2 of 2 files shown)


Garanas:
This usually doesn't that often, but if it does - would it be natural to cancel the current command queue? It means your engineer would effectively be idling.

I think this is great as a first version to fixing the problem. But what if we'd extend the copy orders feature to allow you to skip the first N orders? That way you can copy the orders of the engineer but skip the current order and the queue remains functional. This extension would not be available through the sim callback

MrRowey:
if your extending this like jip sudjest that fine when its ready for merge please add the snipit

MrRowey (approved):

clyfordv:
Extension is in the works

Garanas:
@clyfordv is it fine if I merge this in? You can make the improved version in a different branch, if you haven't already 😃

clyfordv:
We're good to go, I'll add the improvements later.

MrRowey:
> We're good to go, I'll add the improvements later.

Add Changes Snipit and ill merge in

PR #6220, opened 2024-05-30
Move build range overlay creation to post-mod blueprint loading

lL1l1:
## Issue
There are frequently modded games with build range multiplier mods, but the overlay is created before mod loading, so it's not possible to see the new range without UI mods.

- Moves the build range overlay to post mod blueprint loading so it is compatible with mods.
- Adds engineering pods to the overlay since it shouldn't lag and it is helpful to see their build range.
- Excludes engineer stations from the build range overlay since being able to see what they will auto-assist is useful for not getting them stuck on hard to assist things like SMD.
- Excludes engineers with air staging from the overlay just in case it ever happens.

- Enabled a 1.5x build range mod and made sure that the overlay updated for engineers and not for engineer stations.
- Checked that guard scan radius for engineer stations works within the exactly stated radius by disabling the blueprint overlay creation, using the default 25 build range overlay from the blueprint, and placing t1 pd inside/outside that overlay.
  - Larger build skirts (visible with `dbg grid`) like t1 land facs need to have 2 cells within range of the engineer station.
  - Also tested placing an SMD nearby outside the guard radius at various places and it has to clearly overlap the radius to be assisted automatically.
- Checked that build radius works for engineers with the `MaxBuildDistance+2` range by placing T1 PD 7/8 ogrids away from a t1 engi. It can build land factories even further away (8-9 ogrids away from the build skirt max directly vertically/horizontally, but not diagonally (7-8 max)).

## Checklist


lua/system/blueprints-units.lua
```diff
@@ -218,6 +218,28 @@ local function PostProcessUnit(unit)
         end
     end
 
+    -- Build range overlay
+    -- only for engineers, excluding insignificant units such as Cybran build drones or air staging that has its own radius set
+    if isEngineer and not (unit.CategoriesHash['INSIGNIFICANTUNIT'] or unit.CategoriesHash['AIRSTAGINGPLATFORM']) then
+        -- guarantee that the table exists
+        if not unit.AI then unit.AI = {} end
+
+        -- Engine allows building +2 range outside the max distance (or even more for large buildings)
```
Garanas, line 227: This depends on the footprint. I think they use a different type of query to check the distance,  instead of center (engineer) to center (structure) it is center (engineer) to bounding box (structure)

lL1l1: Maybe it checks that the centers of the builder and structure are within `MaxBuildDistance + SkirtDiagonal` (maybe with rounding of some type), since I didn't get consistent build range between placing the structure at a diagonal or at a straight line.

MrRowey:
is this complete @lL1l1

MrRowey (approved):

PR #6228, opened 2024-06-04
Fix several flaks consistently missing their first shot at specific angles

Basilisk3:
## Description of the bug
The bug is easily reproducible by spawning a flak and a hostile air unit north of the flak. Order the air unit to fly directly over the flak. You will observe that the flak fires as soon as it has completed charging up, even if it has not fully turned towards its target yet. This results in the first shot always going haywire.

As far as I am aware, this is another bug that has been in the game since release.

**Ascendant and Marr**
UseFiringSolutionInsteadOfAimBone: false --> true

**Sinnatha and Iashavoh**
RackSalvoChargeTime: 0.1 --> 0

The units function properly now.

## Additional context
- The Sinnatha and Iashavoh do not have a charge-up animation. I assume this field was simply copied over when the Seraphim faction was created.
- I also checked the Czar and the Seraphim cruiser, they are not affected by the bug.
- Changing `RackSalvoFiresAfterCharge` to `false` also fixes the issue, but has significant implications on the unit's firing cycle, which is why I avoided it here.

## Checklist


Garanas (approved):
Looks like a clean fix.

lL1l1:
The Aeon flaks still instantly fire, it's just that they no longer shoot where the weapon is aiming (they use the firing solution instead of the aim bone). It's especially noticeable with the aeon stationary flak shooting twice on the first shot. I'd rather look for a way to fix the effects so that they don't need charging to look good instead of making them ignore their turrets on the first shot.

The Seraphim flaks have a very tiny charge effect in their weapon file, but it is completely unused since Steam, so I guess it's just left over from before the Seraphim designs got finished and were just Aeon clones. 
Here's what it looks like if you enable it as a normal muzzle charge effect:

https://github.com/FAForever/fa/assets/82986251/810915b2-20ed-4bc7-94f8-39798256b763

I think it looks bad so it's fine to just remove the charge time.

Basilisk3:
> The Aeon flaks still instantly fire, it's just that they no longer shoot where the weapon is aiming (they use the firing solution instead of the aim bone). It's especially noticeable with the aeon stationary flak shooting twice on the first shot. I'd rather look for a way to fix the effects so that they don't need charging to look good instead of making them ignore their turrets on the first shot.

I agree it is not an optimal solution, but unfortunately changing the effects is not something I can do. Since I do not know if or when it will happen, I still think it makes sense to include this for the time being, as it does fix a long-standing issue. 

> The Seraphim flaks have a very tiny charge effect in their weapon file, but it is completely unused since Steam, so I guess it's just left over from before the Seraphim designs got finished and were just Aeon clones.

This is basically what I have been thinking as well.

lL1l1:
> Can you clarify what you mean by this? This does not happen on my screen.

A weapon charges up when an enemy gets into tracking range, which is something completely unrelated to the weapon rate of fire. So what happens is that the flak charges due to a target getting into range, fires because it is charged, enters the fire ready state, and then immediately charges and fires again because the engine tells it to fire because the fire rate cooldown was not triggered yet.

lL1l1:
I fixed the free shooting by making it use the muzzle charging instead of rack charging for creating visual effect.

PR #6232, opened 2024-06-07
Enable the Czar's depth charges to function again

Basilisk3:
## Description of the bug
Instead of homing in on their target, the depth charges simply drop to the bottom of the ocean after they hit the water. This renders them practically useless.

This PR enables the Czar's depth charges to track their target again. Once they have entered the water that is.

## Checklist


Garanas:
That would be my mistake 😃 !

PR #6233, opened 2024-06-08
Update and add moderator events

Garanas:

Moderator events are now delayed by 2 seconds by default. This is to prevent the generation of events as players are trying to disconnect. During local testing the events would show for the player leaving, but not for the players still in-game.

Adds a new moderator event: everything that is passed into `GpgNetSend` is also turned into a moderator event. This way we know what people are (not) sending to the server.

Run local multiplayer games.

## Additional context

Related to https://github.com/FAForever/faf-java-commons/issues/121

## Checklist

### DIFF
--- a/lua/ui/game/gameresult.lua
+++ b/lua/ui/game/gameresult.lua
@@ -38,28 +38,23 @@ function DoGameResult(armyIndex, result)
     local armies = GetArmiesTable().armiesTable
     announced[armyIndex] = true
 
-    if (not SessionIsReplay()) and (sessionInfo.Options.CheatsEnabled ~= 'true') then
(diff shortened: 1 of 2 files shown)


lua/ui/game/gameresult.lua
```diff
@@ -38,28 +38,23 @@ function DoGameResult(armyIndex, result)
     local armies = GetArmiesTable().armiesTable
     announced[armyIndex] = true
 
-    if (not SessionIsReplay()) and (sessionInfo.Options.CheatsEnabled ~= 'true') then
-        SetFocusArmy(-1)
```
Garanas, line 42: This was a bug - some old feature that is no longer relevant.

PR #6251, opened 2024-06-13
Add missing localizations to faf init file mounting whitelist

lL1l1:
## Issue
Origin: https://forum.faforever.com/topic/7748/game-crashed-right-after-start/6 Summary: If there is no localization file from FAF servers, then the game will crash on startup because Steam's Czech/Chinese/Polish localization files cannot be mounted due to the whitelist.

Adds other localization files available from the steam version to FAF's init files' whitelists.

Open steam -> right click supcom:FA -> properties -> language
Check that the language is on the whitelist or if it isn't switch to the language and then close the properties window to download the files and add the filename to the whitelist.

## Checklist


Garanas (approved):
Great fix for an oversight 👍

MrRowey (approved):

PR #6252, opened 2024-06-13
Fix issues with fire rate test related to rounding tie-breaking towards evens

lL1l1:

Fixes the fire rate test to account for banker's rounding (tie break towards even numbers), and corrects erroneous fire rate changes from #5845.

Set my ACU's fire rate to 4/0.8 and added a log statement when the projectile is created to make sure it fires every 2/12 ticks.

## Checklist


Garanas (approved):
Great find again 👍 , always a nice reality check to see how many little mistakes my contributions have at times haha.

PR #6253, opened 2024-06-13
Feature/reclaim cost adjustments

Garanas:

Props: from 50m/s -> 10m/s
Units: from 25m/s -> 20m/s

Tree groups behave the same. This is an experimental change.


Tagada14:
@Garanas Could you please revert this PR as the Area Commands are also being reverted.

PR #6257, opened 2024-06-14
Fix overcharge not being fully absorbed by shields

lL1l1:
- Sets shield absorption of overcharge damage type to 1.0 by default. If you want I can make a super simple PR with only this change (change `shieldabsorptionvalues` and remove the *2 OC structure shield damage).
- Reworks the structure of how the damage is calculated so that only one instance of damage from AoE is processed. Previously both the absorbed and unabsorbed portions of the AoE hit the shield. This removes the weird *2 for overcharge structure damage in the shield calculation. It also stops impact effects and spillover from triggering twice when a shield is hit by AoE.
  - The structure damage looked like it was being multiplied by 0.5 because usually the total damage from the double processed instances was: `[dmg * armor] * armor + [dmg - dmg * armor] * armor = dmg * armor`, while the OC math replaced both parts of the AoE damage (square brackets) with `structureDamage`, so you got `800 * armor + 800 * armor`, and armor was 0.25 so you got `2 * 0.25 * 800 = 800 * 0.5`.
  - The new code knows when the damage it is getting is from the absorbed AoE or from the remainder, so it can calculate the absorbed damage properly and discard the remainder. Basically `[dmg * armor] + [dmg - dmg * armor] * 0 = dmg * armor`.

There was an idea to not just set OC damage higher if it was below the structure/command damage, but it makes it weird since the 0.25 structure armor for OC starts taking effect which would require creating a new damage type for FAF overcharges which seems unnecessary.

## Testing done on the proposed changes:
Shoot a titan (personal shield) and an ED1 (static shield) with the main fire (direct damage, so it doesn't trigger `OnGetDamageAbsorption`) or overcharge (AoE damage, OC special case) of a Seraphim SACU and check that the damage is correct (800 for OC) and there are no errors.

## Checklist
- [x] Changes are annotated
- [ ] Changes are in the changelog for the next version

### DIFF
--- a/lua/ShieldAbsorptionValues.lua
+++ b/lua/ShieldAbsorptionValues.lua
@@ -28,6 +28,7 @@
 ---@type table<AbsorptionType, table<DamageType, number>>
 shieldAbsorptionValues = {
 	["Default"] = {
-		["Deathnuke"] = 1.0
+		["Deathnuke"] = 1.0,
+		["Overcharge"] = 1.0,
 	},
 }
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -521,14 +521,10 @@ Shield = ClassShield(moho.shield_methods, Entity) {
         -- damage correction for overcharge
         if dmgType == 'Overcharge' then
             local wep = instigator:GetWeaponByLabel('OverCharge')
-            local newAmount = amount
             if self.StaticShield then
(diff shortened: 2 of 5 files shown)


lL1l1:
Gonna split the pr, it has issues with throwing away incorrect damage instances.

PR #6258, opened 2024-06-14
Make shields absorb full overcharge damage

lL1l1:
## Description of the proposed changes:
Sets shield absorption of overcharge damage type to 1.0 by default. So that units do not take damage from OC when standing at the edge of a shield.

## Testing done on the proposed changes:
Shoot a UEF bubble shield or any static shield and they take 400/800 damage without letting through any damage to nearby units underneath.

## Additional Context
Cherry picked from #6257, further details on the armor multiplier comment there.

## Checklist
- [x] Changes are annotated
- [x] Changes are in the changelog for the next version

### DIFF
--- a/changelog/snippets/fix.6258.md
+++ b/changelog/snippets/fix.6258.md
@@ -1,0 +1,1 @@
+- (#6258) Fix overcharge hitting units standing at the edge of a shield.
--- a/lua/ShieldAbsorptionValues.lua
+++ b/lua/ShieldAbsorptionValues.lua
@@ -28,6 +28,7 @@
 ---@type table<AbsorptionType, table<DamageType, number>>
 shieldAbsorptionValues = {
 	["Default"] = {
-		["Deathnuke"] = 1.0
+		["Deathnuke"] = 1.0,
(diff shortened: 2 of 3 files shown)


MrRowey:
Wasn't it on purpose that oc did less dmg to stationary buildings ?

Garanas:
That's not what this pull request is changing. What it is changing (I think) is eventually using the armor multiplier so that increased (or decreased) damage on one unit type does not necessarily affect another.

lL1l1:
This doesn't change how much damage OC does to buildings, it fixes the bug where a unit standing at the edge of a shield will take 938 damage from the OC splash, even though it should be protected by the shield.

MrRowey:
> This doesn't change how much damage OC does to buildings, it fixes the bug where a unit standing at the edge of a shield will take 938 damage from the OC splash, even though it should be protected by the shield.

AHH gotcha, yh that's a nice fix 😎

MrRowey:
@lL1l1 is this ready?

lL1l1:
Yes

MrRowey (approved):

PR #6129, opened 2024-04-27
Use the internal unit statistics

Garanas:
## Description of the proposed changes

Remove the tracking of unit statistics. They are unnecessary as the engine tracks the same information. 

## Testing done on the proposed changes

Launch a game using 2badd4141ef4ee7a26dd6a8a8356311b25ffaffe and compare the results when you exit. You should see something like this in your logs:

```
INFO: 1	uel0001	0	->	nil -- killed
INFO: 1	uel0001	1	->	nil -- created
INFO: 1	uel0001	1	->	1   -- still alive
INFO: 1	uea0303	0	->	nil
INFO: 1	uea0303	8	->	8
INFO: 1	uea0303	3	->	3
INFO: 1	uel0401	0	->	nil
INFO: 1	uel0401	1	->	1
INFO: 1	uel0401	0	->	nil
INFO: 2	ual0001	0	->	nil
INFO: 2	ual0001	1	->	nil
INFO: 2	ual0001	1	->	1
INFO: 2	uel0001	1	->	1
INFO: 2	uel0001	0	->	nil
INFO: 2	uel0001	0	->	nil
INFO: 2	uea0303	1	->	1
INFO: 2	uea0303	0	->	nil
INFO: 2	uea0303	0	->	nil
INFO: 3	ual0001	1	->	1
INFO: 3	ual0001	0	->	nil
INFO: 3	ual0001	0	->	nil
INFO: 3	xsl0001	1	->	1
INFO: 3	xsl0001	0	->	nil
INFO: 3	xsl0001	0	->	nil
INFO: 3	ura0303	0	->	nil
INFO: 3	ura0303	7	->	7
INFO: 3	ura0303	2	->	2
INFO: 3	xsa0303	7	->	7
INFO: 3	xsa0303	0	->	nil
INFO: 3	xsa0303	0	->	nil
INFO: 3	url0402	0	->	nil
INFO: 3	url0402	1	->	1
INFO: 3	url0402	0	->	nil
INFO: 3	xrl0403	0	->	nil
INFO: 3	xrl0403	1	->	1
INFO: 3	xrl0403	0	->	nil
INFO: 3	xsl0401	2	->	2
INFO: 3	xsl0401	0	->	nil
INFO: 3	xsl0401	0	->	nil
INFO: 4	xsl0001	0	->	nil
INFO: 4	xsl0001	1	->	nil
INFO: 4	xsl0001	1	->	1
INFO: 4	ura0303	2	->	2
[...]

PR #6267, opened 2024-06-17
Rework the brain components

Garanas:
The brain components are now in separate files. This allows you to hook the component before it is being applied to the standard brain class.

Start a scenario and observe that all the functionality (energy management, jammers, hqs, etc) are still functional.

## Checklist


Garanas:
The internal `HQs` field of the brain was not referenced by any of the existing mods and is therefore safe to rename.

Garanas:
I won't touch the statistics module because of: https://github.com/FAForever/fa/pull/6129
And the same reasoning applies to the energy management module because of: https://github.com/FAForever/fa/pull/5822

PR #6271, opened 2024-06-18
Format and cleanup all unit blueprint files

Garanas:
## Description

Large cleanup of the blueprint files using the [BrewLan WikiGen](https://github.com/The-Balthazar/BrewWikiGen) by @The-Balthazar . Each commit represents a separate type of clean up.

PR #6272, opened 2024-06-19
Format the majority of the remaining blueprint files

Garanas:
Continuation of #6271

The only blueprint files that are not formatted are the emitter files. It would merge the key-based and value-based fields, creating a bit of a mess. Luckily we won't be doing much (if any) batch processing on those type of files.

With this we can now finally start batch processing file empowered by the [Brewlan WikiGen](https://github.com/The-Balthazar/BrewWikiGen) made by @The-Balthazar. This would empower us to do various edits to units, props and projectiles an mass and bake them right into the blueprint files.


Basilisk3:
@Garanas What do we do with the campaign-related and miscellaneous units? E.g.: https://github.com/FAForever/fa/blob/develop/units/XAC0101/XAC0101_unit.bp

Garanas:
We can include them if we want, but how the blueprints are now is also harmless. What would you prefer?

Basilisk3:
I never used them for anything. If the memory gains are negligible, then the only advantage I can think of is uniformity. There are a lot of these units, so maybe it is worth it. But only if the amount of time required is reasonable, of course👍

PR #6274, opened 2024-06-19
Batch process all props and bake in existing blueprint manipulation

Garanas:
Batch processes all props and bakes in existing blueprint manipulations, this involves:

- Bake in the invulnerability properties
- Bake in the (prevention of) obstruction properties
- Bake in the LOD properties

And something that would previously take forever to do:

- Bake in adjusted health properties

Practically this applies all the post processing in [blueprints-props.lua](https://github.com/FAForever/fa/blob/develop/lua/system/blueprints-props.lua) and bakes the values into the blueprints for everyone to see.


Garanas:
There's been no objections to these changes, I'd like to merge them in. Could someone review the first two commits? It contains the relevant logic. Everything else is part of the batch processing and just bakes in the values. 

Note to self for next time: the batch processing should have been a separate branch, like what was done with the Blinking Lights.

PR #6276, opened 2024-06-20
Make PR template default instead of requiring query

lL1l1:
The current template requires you to use a [query parameter](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/using-query-parameters-to-create-a-pull-request) to bring up the template, so it doesn't show up as default. For example: https://github.com/FAForever/fa/compare/develop...lL1l1:fa:Fix/attackgroundtries?expand=1&template=pull_request_template.md (notice the `&template=pull_request_template.md` at the end is needed).
This PR moves the template to the .github directory so it should appear as the default according to the [Github Documentation](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/creating-a-pull-request-template-for-your-repository#:~:text=In%20the%20file%20name%20field%3A) on PR templates.


MrRowey (approved):

MrRowey:
This should be fine as we only have the One Pull Request Template

PR #6279, opened 2024-06-20
Populate the blueprint field `AverageDensity` of units

Garanas:

Affects the behavior of units when they bump into each other. Units with more weight push units with less weight. As a result units with more health will receive less to no pushback from units with less health.

Units still try to stop for other units. When the engine decides for the unit to push on through the unit with the most health will have a much, much better time.

Start the game and spawn in various units with different health values and observe how they behave.

## Additional context

The current values are the same as used in this mod that was created to allow people to test the behavior changes:

image

## Checklist


Garanas:
There's been no objections to these changes. They're not harmful to the game and I think we can get a better understanding if we'd include them with the release. Then we can tweak the values and re-bake it as our understanding improves.

PR #6281, opened 2024-06-21
Fixed ExtractCloakMeshBlueprint which assumes Categories being called on a unit without Categories.

The-Balthazar:


MrRowey:
@The-Balthazar can you please make sure that there is a description provided so I can be understood at a late date 
https://github.com/FAForever/fa/blob/develop/.github/pull_request_template.md As the template shows you also require the snippet

lL1l1:
Reported on [Discord](https://discord.com/channels/197033481883222026/832710161847287819/1253539547917193267).

Bug was in the original implementation of the extraction function.

Tested by removing the categories table from UEL0001 and making sure the blueprint loading phase passes successfully.

lL1l1 (approved):

PR #6287, opened 2024-06-21
Fix intel state not being assigned

lL1l1:

Caused by #6273. Intel was not being put into the intel tables used for the intel state table because of a bad order of operations on some booleans.

`UI_Lua LOG('number' ~= 'table')` prints true, while `UI_Lua LOG(not 'number' == 'table') prints false. Enemy radar blips show up ingame.

## Checklist
~~- [ ] Changes are documented in the changelog for the next game version~~ Not necessary.


4z0t:
Because not is applied to string first. It has higher precedence than equal operator.

relent0r (approved):

PR #6288, opened 2024-06-21
Improve guards for `table.copy` to fix blueprint reloading crash

lL1l1:

Fixes in-game blueprint reloading crash caused by #6273 ([specific diff](https://github.com/FAForever/fa/pull/6273/files?diff=unified&w=0#diff-36fc4661c2ca0ef4424b14ce65a6831bb9c6b5b78cf99ae96d9051263f71c676L311-R311)) from `blueprints-units.lua:311` due to spawn dummies having `CommandCaps = "0"`. Instead of checking for nil, checks for the value to be a table.

The function could use a better annotation if possible.

The game no longer crashes when reloading blueprints during a session.

## Checklist
~~- [ ] Changes are documented in the changelog for the next game version~~ Not necessary.

### DIFF
--- a/lua/system/utils.lua
+++ b/lua/system/utils.lua
@@ -73,8 +73,10 @@ function table.getsize(t)
     return size
 end
 
---- table.copy(t) returns a shallow copy of t.
----@overload fun(t: Vector): Vector
+--- Returns a shallow copy of t
+---@generic T
+---@param t T
+---@return T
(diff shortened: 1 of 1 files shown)


Garanas:
What about:

```
--- table.copy(t) returns a shallow copy of t.
---@generic T
---@param t T
---@return T
```

PR #6291, opened 2024-06-23
Fix unit view clipping in replays for the mini layout

PaletzTheWise:
Before:
image
After:
image

I have changed the unit view positioning in replays for the mini layout to the same as the right layout, which is effectively an identical layout because they differ only in:
1) The orders control position but there is no orders control in replays.
2) The clipping issue itself, which I assume is not intended :)

In order to achieve that, I have added a condition that checks whether the orders control exists.

Additionally, I have moved imports in layouts to the top of the file since that is tradition, it avoids import duplication in some cases and it is theoretically faster (it saves up to 30% processing time of the PositionWindow() method, but that still amounts to nothing in the grand scheme of things).

No changes to non-replay positioning of the mini layout:
image

Replay testing:
 * Ran a replay
 * Switched to a specific player
 * For each layout
    * Verified unit view renders correctly without selection
    * Verified unit view renders correctly while a factory is selected
 * Switched to the observer full view
 * For each layout
     * Verified unit view renders correctly while a factory is selected

Skirmish testing:
 * Started a skirmish as a player
 * For each layout
    * Verified unit view renders correctly without selection
    * Verified unit view renders correctly while a factory is selected 

Skirmish observer testing:
 * Same as replay testing except the observer has the orders control so it looks the same as actually playing

## Additional context
I wondered if Gyle runs into this in his casts but he mostly selects combat units that do not have the extra bits at the top of the construction control.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6291.md
+++ b/changelog/snippets/fix.6291.md
@@ -1,0 +1,1 @@
+- (#6291) Fix unit tooltip clipping with contruction tech level selection or enhancement slot selection UI in replays on the default layout (mini).
--- a/lua/ui/game/layouts/unitview_left.lua
+++ b/lua/ui/game/layouts/unitview_left.lua
@@ -4,6 +4,8 @@ local LayoutHelpers = import("/lua/maui/layouthelpers.lua")
 local Prefs = import("/lua/user/prefs.lua")
 local options = Prefs.GetFromCurrentProfile('options')
 local NinePatch = import("/lua/ui/controls/ninepatch.lua").NinePatch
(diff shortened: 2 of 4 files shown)


MrRowey (approved):

MrRowey:
@PaletzTheWise Dose this scale properly when people use UI Scale ingame?

PaletzTheWise:
> @PaletzTheWise Dose this scale properly when people use UI Scale ingame?

Yes, I have checked all layouts in a replay on 80%, 125%, and 200%. Found no inconsistencies.

PR #6297, opened 2024-06-25
Disable area reclaim command

Garanas:

Disables the area reclaim command. Area ground attack command is polished in https://github.com/FAForever/fa/pull/6317


lL1l1:
Area attack has a consensus that it's a beneficial feature for both strategic depth and aesthetics, and it should be added.

Tagada14:
I would leave the area attack as well. I haven't seen any pushback against that change.
Could also revert the reclaim rate changes in this PR or should I make a separate PR for that?

Garanas:
@lL1l1 / @Tagada14 processed the feedback

Tagada14:
@Garanas  AFAIK it was decided to not release the Pulsar this patch either, are we moving that to a separate PR?

lL1l1:
@tagada14 Here's the reasoning to not release Pulsar: [Discord](https://discord.com/channels/197033481883222026/1248621733112578109/1255075225998196816)
Otherwise it's ready to go. The textures are the highest priority (Progress: https://github.com/FAForever/fa/pull/6319). Shield effects would be nice to have. Further balance feedback seems unlikely because of how rare FAF develop games are, combined with the useful situations for it being uncommon.

PR #6308, opened 2024-06-26
Fix sats being able to trigger nuke explosions

Basilisk3:
Instead of the Novax being able to destroy nukes, the nuke now destroys the Novax satellite the majority of air units.

- This fixes an abuse which made it possible to destroy nuke launchers and bases with a falling satellite wreck. The chances of this occurring are rare, but it is possible.
- Nukes destroy all other low-hp air units, and the Novax satellite is a very low hp unit, so this solution is more coherent with the internal logic of the game.
- Rebuilding the satellite is cheaper than building a nuke, even ignoring the launcher. This fixes this balance issue.

This PR does not remove the "simulated projectiles" part of the game. The satellite dies when impacting on the nuke, just like any other low-hp air unit.

## Checklist


MrRowey:
As discued in the balace channle thier is no abuse to such mechanic due to falling sataline has falling randomness. as well as their is no balance created issues cureenly with the satalite. Do not see the need for nukes ignoring the satellite as it litlery falling from space at immense speed hense the missiles detonate.

Basilisk3:
> As discued in the balace channle thier is no abuse to such mechanic due to falling sataline has falling randomness. as well as their is no balance created issues cureenly with the satalite. 

As I explained on Discord, both of these statements are wrong. Just because something is random, does not mean it cannot be abused.

Do not see the need for nukes ignoring the satellite as it litlery falling from space at immense speed hense the missiles detonate.

This is also wrong, the nuke does **not** ignore the satllite. The nuke destroys the satellite, in accordance with the game having simulated projectiles.

lL1l1 (commented):
It's up to balance team to decide if sacrificing a satellite is fair to block nukes. To individually fix the absurdly rare chance of blocking a nuke using the wreck you can check `self.Dead` and return false in the satellite's script.
You could do a more casual nerf to this mechanic by increasing the costs, especially buildtime, of satellite to make it more fair when it does get used. This would create a real opportunity cost in not being able to use the satellite for a long time because you end up with a 19.5k mass difference basically sitting around doing nothing on your side. The mass difference of course isn't as large because of the let's say extra t3 pgen the nuke user has to build, so you can consider adjusting the satellite energy cost too.
Here's the data for analysis:
| Unit | Mass Cost | Energy Cost | Buildtime | Notes
|--|--|--|--|--|
Novax Station | 36k | 512k | 44,800 | 300 bp
Satellite | 10k | 160k | 30,000 | Gains value slowly by being alive, can block nukes by being nearby when they are launched and then being microed to float right above them.
Nuke launcher | 16.5k | 412.5k | 25,000 | 1500 bp
Nuke Missile | 12k | 1,500k | 450,000
Yolona missile | | | | takes 2 smd to intercept
SMD | 7500 | 105k | 5841 | 1080 bp
SMD missile | 3.6k | 360k | 259,200

Basilisk3:
Thanks for your analysis. I read your post on discord. If you can control the trajectory of the wreck, then this is definitely something that should be fixed in my opinion.

Basilisk3:
I changed it so nukes still explode when colliding with satellites. The exploit is removed by preventing falling satellites from triggering nuke explosions. I prefer my initial solution, so this is a compromise.

@lL1l1 I can also move the changes over to the Defense Satellite's script file, as you suggested. The current solution would apply to all possible satellites, which I would prefer.

lL1l1 (approved):
I think it's fine to keep it in the main air unit script if it's only for falling satellites. Satellite units in general shouldn't be able to block nukes multiple times per life after all.

Basilisk3:
Thanks for taking the time👍

PR #6311, opened 2024-06-27
Move hidden crash damage out of Ahwassa and CZAR scripts

lL1l1:
## Issue
Ahwassa/CZAR deals 1000 damage in 5 radius due to a line in its script, which makes that information unknown to the user because it is not in the blueprint.

Removes the scripted damage and adds it to the blueprint. This results in:
- high damage area radius: 5 -> 10 (Ahwassa) / 15 (CZAR)
- high damage area damage through Seraphim T3 shields: 1000 (constant) + 2800 (reduced by shields) -> 3800 (reduced by shields)
  - The shield choice is relevant because shields can reduce crash damage by up to 20% of their max HP. That means the max crash damage reduction in normal games is `21k * 0.2 = 4200`. 3800 damage under shields keeps the ability to one shot SMDs with crash damage.
- high damage against unshielded targets: remains the same

Crashing these air T4s through shields deals expected damage.

## Additional context
Previous balance PR reducing crash damage: https://github.com/FAForever/fa/pull/4543

## Checklist

### DIFF
--- a/changelog/snippets/fix.6311.md
+++ b/changelog/snippets/fix.6311.md
@@ -1,0 +1,1 @@
+- (#6311) Ahwassa and CZAR dealt an extra 1000 crash damage due to a line hidden in their script, which doesn't show up with the 7000 crash damage listed in the UI. The 1000 damage is added to the 7000 damage to make it work in the UI. As a consequence, the radius where that extra damage is dealt is increased from 5 to 10/15.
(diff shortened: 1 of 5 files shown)


MrRowey (approved):
Nice Catch

Garanas:
It feels like to me that this is merged in a little bit fast. 

For example, unrelated to the idea behind the changes, the current implementation also changes the behavior how trees fall to the ground. This may not matter much gameplay-wise, but it's a small and harmless detail that matters for those that happen to see it 😄 !

MrRowey:
`It feels like to me that this is merged in a little bit fast.` tbh didn't see when the pr was made xD just saw the notification xD

lL1l1:
> For example, unrelated to the idea behind the changes, the current implementation also changes the behavior how trees fall to the ground.

The Unit.lua deathweapon already creates TreeForce damage. Does stacking TreeForce damage create a stronger effect?

PR #6316, opened 2024-07-01
Implement GetMousedOverWorldView

lL1l1:
## Description of the proposed changes

Adds a function that gets the moused over WorldView.

## Testing done on the proposed changes

Using the function returns the view as expected.

## Additional context

Will be used by area attack dragger in #6317.

## Checklist
- [x] Changes are annotated, including comments where useful
- [x] Changes are documented in the changelog for the next game version

PR #6317, opened 2024-07-01
Polish area attack dragger code and visuals

lL1l1:

Builds on top of/contains #6312 and #6316 (rebase this PR when those are merged in).
- Clean up area attack script with removing unused values and refactoring variable names.
- Make the area attack radial dragger red colored.
- Make the area attack dragger show up in the worldview it is clicked on (previously it only worked in one worldview).

Use the area attack order by clicking and dragging and make sure it looks good.

## Checklist


lua/sim/commands/area-attack-ground-order.lua
```diff
@@ -51,9 +45,8 @@ end
 
 ---@param units Unit[]
 ---@param target Vector
----@param doPrint boolean           # if true, prints information about the order
----@param radius? number
-function AreaAttackOrder(units, target, doPrint, radius)
```
Garanas, line 56: The printing was not used, but in general the printing exists as a form of feedback for the user that the command is processed. I'm fine with it (not) existing, but that's why all the other special commands (distribute/copy orders, assist-to side effects, etc) have this printing.

Garanas (approved):
The polishing that it needed 👍 !

PR #6331, opened 2024-07-08
Update Workflow File for doc Site

MrRowey:
Added a Pull Request Path so you can stop other Workflow tests that are not needed.


Garanas:
Was it the intention to also not run the test workflow for Lua scripts?

Garanas:
We don't want this workflow to trigger on pull requests, closing it for now

PR #6332, opened 2024-07-08
Fix various small issues and update the processing of patch notes

MrRowey:


docs/404.html
```diff
@@ -21,5 +21,6 @@
   <h1>404</h1>
 
   <p><strong>Page not found :(</strong></p>
+  <img src="../assets/uef-dance.gif">
```
Garanas, line 24: Nice touch 😄

PR #6333, opened 2024-07-08
Prevent TML from overflying TMD when the target is at a higher altitude

clyfordv:
See discussion here: https://forum.faforever.com/topic/7834/tml-height-needs-to-be-lowered/9

**Overview**
Fix to prevent TML from (more easily) overflying TMD when shooting at targets at higher elevation. Previously the height of the target was used to calculate the highest point of the arc; this uses the height of the launcher instead. Note that with this fix (and, implicitly, prior to the original projectile rework) it's still possible for missiles to overfly nearby TMD, but this narrows the window.

**Possible Issues**
The original intention behind using the height of the target was to eliminate an edge case where the target is so high that the missile doesn't arc properly. This reintroduces that (exceedingly rare--missiles fly very high relative to the terrain) possibility.

### DIFF
--- a/changelog/snippets/fix.6333.md
+++ b/changelog/snippets/fix.6333.md
@@ -1,0 +1,1 @@
+(#6333) Prevent TML from overflying TMD when firing at targets at higher elevation
--- a/lua/sim/projectiles/components/SemiBallisticComponent.lua
+++ b/lua/sim/projectiles/components/SemiBallisticComponent.lua
@@ -214,8 +214,7 @@ SemiBallisticComponent = ClassSimple {
     ---@return number
     OptimalMaxHeight = function(self)
         local horizDist = self:HorizontalDistanceToTarget()
-        local targetHeight = self:GetCurrentTargetPosition()[2]
-        local maxHeight = targetHeight + horizDist/(self.HeightDistanceFactor + self.HeightDistanceFactorRange * (2 * Random() - 1))
+        local maxHeight = self:GetPosition()[2] + horizDist/(self.HeightDistanceFactor + self.HeightDistanceFactorRange * (2 * Random() - 1))
         return maxHeight
     end,
 


lL1l1 (commented):
This doesn't fix UEF TML flying so high that zapper's beam doesn't reach them on the first shot:
image
and that buzzkill can miss the missile randomly (this image has the projectile lifetime increased, as timing out is also an issue):
image

Sera TMD has no issue with timing out or hitting because it has a homing projectile. Aeon TMD doesn't shoot the flare high enough to intercept the missile, but that is a good idea to set the muzzle velocity targeting the projectile altitude.

lL1l1:
Closing since it doesn't solve the original issue.

Basilisk3:
> This doesn't fix UEF TML flying so high that zapper's beam doesn't reach them on the first shot:

How did you make it fly so high? I cannot reproduce this.

lL1l1:
> How did you make it fly so high? I cannot reproduce this.

That height is the middle of the max range arc. In the old test I also targeted onto a higher surface to make the missile fly even higher, but that is not necessary to break TMD with UEF TML, it does so on a flat surface:
(Mavor for scale)
{6955FDFF-FD06-4E28-9F20-1C0243F372CC}

Some other distances:
{006B4F4E-8DF5-425B-9469-872348E071CF}

These are the zapper effects you should look for:
{440BB1D6-7A7E-4A3D-9301-6086E78DAA22}

It's weird that the zappers hit far more often when targeting the missile on the left side in my picture but not on the right.
{B17B32F4-D06B-4013-BB72-9FE5A5262747}

Basilisk3:
Edit: More testing required.

Basilisk3:
Do you happen to know how to increase the hitbox for the missiles? I also found that the Zapper really misses a lot.

lL1l1:
The hitbox is set based on the missile speed in faction-specific projectile scripts, for example:
https://github.com/FAForever/fa/blob/c3978e6a7bdf46e1ce7c4516ad77a8d76a77e31d/lua/terranprojectiles.lua#L342

You have to search all files for that pattern to change them all. It's this way because we didn't get around to refactoring projectiles after rowey left.

Basilisk3:
Alright, thanks for your help.

PR #6335, opened 2024-07-09
Fix platoon bug introduced in 6083

maudlin27:
Previous change meant :GetPlan() no longer worked on platoons

The pull request 6083 changed various parts of platoon.lua:
https://github.com/FAForever/fa/pull/6083/files

One of these was changing "if not self[plan] then return end" to:
"if not plan or self[plan] then return end"

I.e. it was aborting if self[plan] was not nil, when the intent was presumably to abort if it was nil

Without this change - tested on M27, it has an error that breaks the mod (since it relies on :GetPlan() which returns nil for its platoons); Also confirmed such an error didn't occur on the FAF release in April 2024 (before PR 6083).
With this change - re-ran a replay where M27 was broken and it works again.

### DIFF
--- a/changelog/snippets/fix.6335.md
+++ b/changelog/snippets/fix.6335.md
@@ -1,0 +1,1 @@
+AI - Fixed a bug with recording a platoon's plan and certain other platoon related logic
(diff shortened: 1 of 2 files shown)


MrRowey:
Remember to add the Snipit :)

MrRowey (approved):

PR #6338, opened 2024-07-09
Add Root pathing for Posts

MrRowey:


docs/_posts/2024-03-16-3801.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: post
 title: Game Version 3801
-permalink: /:year/:month/:day/:title:output_ext
+permalink: /fa/:year/:month/:day/:title:output_ext
```
Garanas, line 4: These should be more simple and only contain the game version. Otherwise we can't link to it from the in-game patch notes.

Garanas: As an example: `changelog/3810`

MrRowey: The issuse is for the _posts to work properly with jekyll the file name has to be Year-Month-Date-title.Extention so they are like 2024-07-31-3610.md

Garanas: What's the difference between a post and a page? We can just turn them into pages if posts don't work.

MrRowey: so if we do it like I first did the first entry will be the oldest version number, not the newest unless we Manually Link each post. so we have 3 Options.

A - This current approach with Posts and the naming convention declared by Jekyll
B - Switch to pages and have the ToC auto-populate them from Oldest to Newest
C - Manually Do the pages by setting the Parent and Child with the Nav Order Preset.

Garanas: I think it's fine to manually create the list of items. It's tough to setup in the first run, but it gives you complete control in the end.

MrRowey: ill update it to be a manual list then

PR #6340, opened 2024-07-09
1 bug hotbuild preview cycle skips selection

### LINKED ISSUE #6337: [Bug]: hotbuild preview cycle skips selection
### Describe the Bug

When using hotbuild hotkeys, after building the first, the selection will not reset to the first option in the cycle. Leads to wrong units being built.

### Reproduce the bug

1. Press hotbuild hotkey to place building
2. Place selected building.
3. Quickly press key again to select new building, hotbuild preview cycle does not restart with first item as expected, instead it jumps to next (wrong) building in the cycle preview

### Screenshots

_No response_

### Additional context

_No response_

ErikMekkes:

1. Moved `if not command.Clear then issuedOneCommand = true` signal above the return statement
2. Added additional `if command.Clear then endCommandMode(true)` call above callback return statement
These two interact with resetting the hotbuild cycle preview, and were missing for cases where the early callback return triggered.
Fixes #6337

Tested preview cycling by repeatedly placing factories to make sure the issue is resolved.
Tested repeatedly cycling other hotkeys to make sure the preview still works in other cases.
Played some coop maps as a minimal 'does it break anything else' check.
Tested the cheat menu to ensure it is not broken by hotbuild preview interaction.

## Additional context

there are some vague elements left from the original hotbuild integration:
unsure if there is other functionality where `EndCommandMode(true)` should NOT be called for callbacks that satisfy:
```if (OnCommandIssuedCallback[command.CommandType] and OnCommandIssuedCallback[command.CommandType](command)) or command.CommandType == 'None' and command.Clear```

unsure if the signal `if not command.Clear then issuedOneCommand = true` can cause issues for other functionality when set before callbacks that satisfy ```if (OnCommandIssuedCallback[command.CommandType] and OnCommandIssuedCallback[command.CommandType](command)) or command.CommandType == 'None'```

## Checklist


Garanas (approved):
I think you're spot on and found the bug. Can you add a [changelog snippet](https://github.com/FAForever/fa/tree/develop/changelog/snippets)?

ErikMekkes:
Added, assuming that is sufficient, not familiar with the template

MrRowey (approved):

PR #6341, opened 2024-07-10
Various improvements to unitviewDetail.lua.

Basilisk3:
1. Update the formatting for the 'Speed' and 'Vision' columns in unitviewDetail.lua, by introducing more precise rounding and the removal of trailing zeroes where possible.
   - Example: unitviewDetail now correctly shows the Harbinger as having a speed and reverse of `2.95` instead of `3.0`. Another Example would be the Othuum which is now shown as having a speed of `2.85`.

2. Enable unitviewDetail.lua to show the 'Scry' ability of the Eye of Rhianne and the 'Vision' ability of the Soothsayer.
   - UnitviewDetail now shows the radius, initial cost and upkeep of the Eye of Rhianne's 'Scry' ability.
   - UnitviewDetail now shows the maximum vision radius of the Soothsayer's 'Vision' ability.

3. Enable unitviewDetail.lua to show whether a unit can prioritize ACUs via the snipemode feature.

4. Enable unitviewDetail.lua to show the `ShieldRechargeTime` stat of personal shields and shield domes.

## Checklist


Basilisk3:
How about `Target Priorities: can prioritize ACU`? I added `Target Priorities:` to specify that the info concerns itself with target priorities; otherwise, this might be unclear for players who are not as familiar with the topic as we are. I think this is not optimal either though, because it's a bit long.

lL1l1 (approved):
I don't think there is really a shorter way to phrase `SNIPEMODE`'s behavior because it's a complex mechanic.

PR #6343, opened 2024-07-12
Make it possible to give HARMS to other players and fix them being destroyed when shared via full share

Basilisk3:
Remove the 'INSIGNIFICANTUNIT' category from XRB2309_unit.bp.
- This makes it possible to give HARMS to other players.
- This keeps them from getting destroyed when their original owner dies in a full share game.

Unfortunately, this makes two HARMS show up in the unit databases. One is the original HARMS (XRB2308), the other one (XRB2309) is used to make the HARMS targetable while it is under construction (#2472).

## Checklist


lL1l1 (approved):
Unit databases should figure out how to fix their aesthetic issue, this is an important change for gameplay.

PR #6352, opened 2024-07-15
Automagically verify URLs in documentation

Garanas:

Extends upon #6351

### DIFF
--- a/.github/workflows/docs-url-check.yaml
+++ b/.github/workflows/docs-url-check.yaml
@@ -42,4 +42,4 @@ jobs:
         uses: urlstechie/urlchecker-action@0.0.34
         with:
           subfolder: docs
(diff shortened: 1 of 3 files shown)


Garanas:
I'm not a huge fan of the setup of this Action. The repository is:

- https://github.com/urlstechie/urlchecker-action/tree/0.0.34

And one may notice that the source code is not in this repository. It is actually retrieved via the Docker image, found at:

- https://github.com/urlstechie/urlchecker-python/pkgs/container/urlchecker

And all the GitHub Action is a wrapper to call the Python in that Docker image. The source code of the actual tool can be found here:

- https://github.com/urlstechie/urlchecker-python

I understand the setup, but not having the source code in the GitHub Action itself feels a little sketchy 🤔 ?

PR #6358, opened 2024-07-17
Update Amphibious Vision

MrRowey:
## Checklist

### DIFF
--- a/units/URL0203/URL0203_unit.bp
+++ b/units/URL0203/URL0203_unit.bp
@@ -108,8 +108,8 @@ UnitBlueprint{
         UnitName = "<LOC url0203_name>Wagner",
     },
     Intel = {
-        VisionRadius = 22,
-        WaterVisionRadius = 20,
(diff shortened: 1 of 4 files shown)


units/URL0203/URL0203_unit.bp
```diff
@@ -108,8 +108,8 @@ UnitBlueprint{
         UnitName = "<LOC url0203_name>Wagner",
     },
     Intel = {
-        VisionRadius = 22,
-        WaterVisionRadius = 20,
+        VisionRadius = 20,
+        WaterVisionRadius = 18,
```
lL1l1, line 112: Imo the difference between water vision and normal vision is very small so they should be the same so that the UI looks better.

MrRowey: if this were the case then the Current would be the same. I was generally matching the same difference when -2 to the visions I applied that to water as well.

PR #6363, opened 2024-07-22
Transfer all unbuilt units on death

### LINKED ISSUE #6361: [Bug]: Part-complete buildings dont transfer in full share
### Describe the Bug

Part-constructed buildings don't transfer on death.

### Reproduce the bug

Example of how to reproduce - create a game with an AI teammate in full share.  Cheat in some t2 engineers and start building a T2 PD.  Ctrl-K your ACU.  The part-complete T2 PD will die instead of transferring over.

### Screenshots

_No response_

### Additional context

See this forum post discussing the issue:
https://forum.faforever.com/topic/7914/why-do-unfinished-units-buildings-ctrl-k-in-fs-games

Also this pull request (referenced in the above) which fixed the issue for experimentals:
https://github.com/FAForever/fa/pull/2530

One comment in the pull request was "This transfer is pretty complex and we don't want to transfer too many units this way. So it doesn't contain nukes/smd buildings, as their cost are too low compare to t4/arty." - I took this to mean that it could be a performance issue for not expanding the functionality.  However, Jip noted Strogoo and Keyser were members of the balance team at the time and thought it intentional for balance reasons for only experimental units to be transferred.

I've messaged Tagada (balance team lead) to see if it can be clarified whether it's intended from a balance team perspective that only experimentals should transfer in full-share on a teammate's death.  If the balance team doesn't object to expanding this, then I can look at changing this and quantifying the performance impact to assess if fixing what feels like a bug would have prohibitive performance implications.

lL1l1:

Resolves #6361.
- Changes the categories for structure transfer from T4s and T3 Arty to all units. 
- Adjusts the stats of the wreckage created when a unit rebuild fails so that the wreckage has the full mass of the failed unit.

Spawn a paragon and some t3 engineers (they dont shoot) then build any unfinished structure or unit from a factory and run the console command to transfer the units to the opposite of the current focused army:
```lua
SimLua 
local units = ArmyBrains[GetFocusArmy()]:GetListOfUnits(categories.ALLUNITS)
local SimUtils = import('/lua/SimUtils.lua')
local target = GetFocusArmy() == 2 and {1} or {2}
ForkThread(SimUtils.TransferUnfinishedUnitsAfterDeath, units, target, false)
```
All structures get shared properly, and unbuilt units in factories turn into wrecks of the same partial cost as the unbuilt unit.

Didn't test performance, I'm not sure how to.

## Checklist

### DIFF
--- a/changelog/snippets/features.6363.md
+++ b/changelog/snippets/features.6363.md
@@ -1,0 +1,2 @@
+- (#6363) When an army is defeated, transfer all unbuilt units instead of only Experimentals and T3 Arty.
+  - When a unit fails to rebuild because its build site was blocked (for example a unit in factory), the mass invested is returned as a wreck.
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -8,7 +8,7 @@
 
 local CreateWreckage = import("/lua/wreckage.lua").CreateWreckage
 
-local transferUnbuiltCategory = categories.EXPERIMENTAL + categories.TECH3 * categories.STRUCTURE * categories.ARTILLERY
+local transferUnbuiltCategory = categories.EXPERIMENTAL + categories.STRUCTURE
 local transferUnitsCategory = categories.ALLUNITS - categories.INSIGNIFICANTUNIT
 local buildersCategory = categories.ALLUNITS - categories.CONSTRUCTION - categories.ENGINEER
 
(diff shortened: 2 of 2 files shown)


MrRowey (approved):

Garanas:
@MrRowey the hotfix did not happen (yet), do we want this to be part of that?

MrRowey:
Yes this is fine as it more of a bug fix

PR #6382, opened 2024-07-31
Enable us to alter the styling of the github.io changelog-page

Basilisk3:
Adds an .scss file.


docs/_config.yml
```diff
@@ -6,3 +6,6 @@ nav_enabled: true
 search_enabled: false
 logo: "./assets/FAF_Logo.png"
 favlogo_ico: "./assets/favicon.ico"
+sass:
+  style: expanded
+  ./assets/css/custom.css
```
Garanas, line 11: The file is called `styling.css` and not `custom.css`. What is the goal, should the css apply to all pages? If so then perhaps call it `base.scss`. Note the additional `s` since we work with [sass](https://sass-css.org/).

Basilisk3: 👍

BlackYps: Where did you find the documentation to do it this way? I only found this:

"You can add custom schemes. If you want to add a scheme named foo (can be any name) just add a file _sass/color_schemes/foo.scss (replace foo by your scheme name) where you override theme variables to change colors, fonts, spacing, etc.
To use the custom color scheme, only set the color_scheme parameter in your site’s _config.yml file:
color_scheme: foo"

Garanas:
Closed in favor of: https://github.com/FAForever/fa/pull/6386

PR #6383, opened 2024-07-31
Enable the SR's Iridium Rocket Pack to deal damage in an area of effect

Basilisk3:
The SR has four air-to-ground weapons: two bolters, which are the more dominant weapons, and two rocket packs, which are often overlooked. In the current release, only the bolters have area of effect damage, the rocket packs have none at all. 

From a design perspective, this is suboptimal since the rockets have a decently large explosion effect. The Renegade uses the same projectile with similar effects and has an area of effect radius of 3. It is also not intuitive, because it could be reasonably assumed that the rocket packs perform similarly to either the Renegade or the other air-to-ground weapons of the SR. Balance wise, it results in the rocket packs underperforming against smaller targets, such as Titans.

**Soul Ripper: Experimental Gunship (URA0401)**:
   - Iridium Rocket Pack (x2):
      - DamageRadius: 0 --> 3
      - Damage: 190 --> 150
      - DPS: 285 --> 225

## Additional context
Since this does increase the effectiveness of the rocket packs, their DPS is reduced.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6383.md
+++ b/changelog/snippets/balance.6383.md
@@ -1,0 +1,7 @@
+- (#6383) The Soul Ripper's Iridium Rocket Packs gain the same area of effect radius as its primary weapons. Visually, these weapons appeared like they dealt damage in an area of effect, however, this was never the case. Due to being very inaccurate, this caused them to underperform against smaller units.
+
+    - Soul Ripper: Experimental Gunship (URA0401):
+    - Iridium Rocket Pack (x2):
+      - DamageRadius: 0 --> 3
+      - Damage: 190 --> 150
+      - DPS: 295 --> 225
--- a/changelog/snippets/balance.6383.md
+++ b/changelog/snippets/balance.6383.md
@@ -1,6 +1,6 @@
 - (#6383) The Soul Ripper's Iridium Rocket Packs gain the same area of effect radius as its primary weapons. Visually, these weapons appeared like they dealt damage in an area of effect, however, this was never the case. Due to being very inaccurate, this caused them to underperform against smaller units.
 
-    - Soul Ripper: Experimental Gunship (URA0401):
+  - Soul Ripper: Experimental Gunship (URA0401):
     - Iridium Rocket Pack (x2):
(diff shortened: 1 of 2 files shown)


lolaenter:
dont lower his  dps, wtf. sr is bad now will be worse then now, why dont you lower mass of harms  now that  it is usless

Basilisk3:
> dont lower his dps, wtf. sr is bad now will be worse then now,
The SR is not bad and this PR only changes its secondary guns, so the balance impact of this is minimal. Please take note of the addition of a damage radius.

> why dont you lower mass of harms now that it is usless
HARMS are not useless, and this PR is not about changing them.

lL1l1:
You could argue that the inaccurate and low damage radius weapon is meant to increase DPS against large targets like structures or experimentals, and giving it a splash radius would make it significantly better against small targets.
Old DPS vs small targets | Old DPS vs large targets | New DPS vs all targets
--|--|--
1858 | 2428 | 2308 (+24% small targets, -5% large targets)

Basilisk3:
Your calculation assumes that the rockets cannot hit smaller targets at all, which I think is a bit unrealistic. To test this, I disabled the SR's main weaponry and left only its rockets enabled. The rockets are inaccurate but they managed to hit their target (a Titan) with a likelihood of around 33% to 50% without splash.

When adjusting for this, I arrived at these values:

| Old DPS vs small targets (33% hit chance):  | Old DPS vs small targets (50% hit chance) | New DPS vs all targets |
| -------------------------------------------- |------------------------------------------- |----------------------- |
| 2046                                                           | 2143                                                        | 2308 (+12.8% or +7.7% small targets)      

The hit chance is closer to 33% than it is to 50%, so I think it's fair to assume that the DPS is roughly 11% higher against smaller targets and 5% lower against larger targets. I think this is reasonable, but if you disagree with this, I can still move the values around.

clyfordv (approved):
Tested, changes are as advertised. Synergizes with: https://github.com/FAForever/fa/pull/6435

Basilisk3:
Approved by the Balance Team.

PR #6390, opened 2024-08-04
Fix the navigational mesh not culling labels

Garanas:

Related to #5574 and #5579

In order to serialize the navigational mesh there can be no direct references to other leaves. This would turn it into a gigantic graph, which would cause the game to crash upon saving.

To fix this all references to a leaf are a number. This behaves similar to a pointer, where the number represents the index in a large list of leaves. When we loop over all the neighbors we therefore retrieve a number. And before we do any computations and/or table operations we should convert that number to the actual instance.

Which we did not do.

There's two underlying issues that should have prevented this, but it did not:

- One is related to https://github.com/FAForever/fa/blob/develop/lua/system/config.lua#L16-L18, where numbers do not error out when trying to perform a table operation on them.
- The other is related to the annotations - we override the actual value and that provided a fake sense of intellisense

Start a game on a generated map (seed: neroxis_map_generator_1.12.3_7zoovyk44bxky_aqeaefam) and observe that the number of labels is reduced drastically (from 6226 to 2).

_Behavior on the release branch (each isolated color is a separate area)_
image

_Behavior in this pull request_
image


relent0r:
I did a test against a few problem maps. This completely removes the stutter and actually improved (took longer to drop from +10 than previously) sim speed when AI are involved that use the navmesh.

PR #6392, opened 2024-08-04
Github.io changelog: increase the font-weight of the level 1 heading

Basilisk3:
As per the title. Currently, h2 has a higher weight than h1, which looks a bit odd.

### DIFF
--- a/docs/_sass/color_schemes/faf.scss
+++ b/docs/_sass/color_schemes/faf.scss
@@ -1,2 +1,6 @@
 $link-color: $blue-100;
 $body-text-color: $grey-dk-300;
+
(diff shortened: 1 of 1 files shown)


lL1l1 (approved):
Tested by modifying using inspect element.
Before:
image

After:
image

Basilisk3:
Looks like this change is overridden in base.scss.

Replacing the code added in this PR with the code below could fix this, but since it is not possible to test changes to the .scss via inspect element, I cannot be sure.

```
.custom-container h1,
.custom-container .text-alpha {
  font-weight: 600;
}
```

PR #6360, opened 2024-07-18
Fix various units being able to evade torpedoes due to their hitboxes being too shallow

Basilisk3:
## Description of the proposed changes
If a unit's hitbox is not submerged far enough into the water, torpedoes can randomly miss the unit by traveling underneath it. This was quite common for the Cooper and not something that should happen. A CollisionOffsetY of -0.375 appears to be the standard for most UEF naval units and resolves the issue. Since this is technically a nerf, the Cooper gets increased hitpoints to compensate.

**Cooper: T2 Torpedo Boat (XES0102)**
CollisionOffsetY: -0.25 --> -0.375
SizeY = 0.8 --> 0.925

**Bulwark: T2 Shield Boat (XES0205)**
CollisionOffsetY: -0.25 --> -0.375
SizeY: 1.2 --> 1.325

**Shard: T1 Air Attack Boat (UAS0102)**
CollisionOffsetY: -0.25 --> -0.375

**Siren Class: T2 Cruiser (URS0202)**
CollisionOffsetY: -0.29 --> -0.375
SizeY: 1.5 --> 1.6

**Tech 3 Sonars**
CollisionOffsetY: -0.25 --> -0.375 (all Tech 3 Sonars)
SizeY: 1.1 --> 1.2 (Aeon and UEF)
SizeY: 1.2 --> 1.3 (Cybran)

**Tech 2 Sonars**
CollisionOffsetY: -0.25 --> -0.375 (all Tech 2 Sonars)
SizeY: 1.0 --> 1.1 (UEF and Seraphim)
SizeX = 0.5 --> 0.7 (Aeon)
SizeY = 0.6 --> 0.8 (Aeon)
SizeZ = 0.5 --> 0.7 (Aeon)
SizeX = 0.5 --> 0.6 (Cybran)
SizeY = 0.6 --> 0.8 (Cybran)
SizeZ = 0.5 --> 0.6 (Cybran)

**Tech 1 Sonars**
CollisionOffsetY: -0.25 --> -0.375 (all Tech 1 Sonars)
SizeY: 0.6 --> 0.7 (Aeon and Cybran)
SizeY: 0.8 --> 0.9 (Seraphim)
SizeY: 0.9 --> 1.0 (UEF)

**Iathu-ioz: T3 Anti-Air Defense (XSB2304)**
CollisionOffsetY: -0.25 --> -0.375
SizeY: 0.9 --> 1.7

**Ialla: T1 Anti-Air Defense (XSB2104)**
[...]

PR #6395, opened 2024-08-06
Hitbox improvements for various amphibious and naval units

Basilisk3:
#6360 fixed various units being able to evade torpedoes due to a too shallow hitbox. This was done by increasing the part of the hitbox that is underwater, while the above-water hitbox was usually kept at the same size. To accomplish this, the `CollisionOffsetY` was changed from around `-0.25` to `-0.375`, and the `SizeY` of the hitbox was increased, so the above-water size of the hitbox remains unchanged.

Unfortunately, this made a targeting bug rear its head, which caused a few weapons to miss some of the units that were changed in the PR. This was unexpected since the above-water hitbox was usually kept exactly the same (e.g., the Cooper). I assume this is an engine bug.

**The goals of this PR:**
- Ensuring that lasers and other weapons no longer shoot over shorter units.
- Ensuring that torpedoes can hit units reliably.

## Stat changes
- Added targeting bones to a wide range of units, like the Shard, Cooper and various sonars.
- Increased the depth of various hitboxes to prevent torpedoes missing.
- Miscellaneous hitbox improvements.

## Unfixed units
- Vesper (lasers missing when surfaced): needs an "Abovetgt" target bone like the UEF Tech 2 Sonar.
- Salem: it is amphibious
- Cybran Tech 2 and Tech 3 engineers: would need an "Abovetgt" target bone or a higher hitbox.

**No changes required**
- Neptune Class: T3 Battlecruiser
- Tech 3 Battleships
- Atlantis: Experimental Aircraft Carrier

## Additional context
- For torpedoes to be able to hit naval units properly, a `CollisionOffsetY` of at least `-0.375` is necessary. A `CollisionOffsetY` of `-0.4` is guaranteed to fix the issue and prevents the last very rare misses that occur with `-0.375`.

## Checklist


units/URS0203/URS0203_unit.bp
```diff
@@ -1,6 +1,11 @@
 UnitBlueprint{
     Description = "<LOC urs0203_desc>Attack Submarine",
-    AI = { GuardReturnRadius = 40 },
+    AI = {
+        GuardReturnRadius = 40,
+        TargetBones = {
+            "URS0203",
+        },
```
lL1l1, line 7: The root bone (the unit name one) is the default targeting bone already.

units/XAS0204/XAS0204_unit.bp
```diff
@@ -191,11 +191,10 @@ UnitBlueprint{
     SelectionSizeX = 0.6,
     SelectionSizeZ = 1.8,
     SelectionThickness = 0.37,
-    CollisionOffsetX = 0.05,
```
lL1l1, line 194: The offset is from GPG files and the mesh is symmetrical except the left front panel, the top rib, and the top oval part, so since it's so minor removing this is fine I suppose.

lL1l1 (commented):
Without a testing setup to replicate the issues I can't really test the collision box changes so that lasers/torpedoes hit these units more reliably. Did you have a testing setup?

units/URS0203/URS0203_unit.bp
Basilisk3: Without this bone, lasers will miss the sub when it is surfaced. I used the `NoDamage` command and Seraphim destroyers to test the changes.

units/XAS0204/XAS0204_unit.bp
Basilisk3: Yes, the offset is not necessary and actually makes the hitbox fit worse.

Basilisk3:
I think this PR is ready. It is unfortunate that the Vesper and Cybran Tech 2 and Tech 3 engineers cannot be fixed without the addition of target bones, but that can be tackled at a later date.

Basilisk3:
Can this be part of the patch on the 22nd? I am here to apply feedback if necessary.

clyfordv (approved):
All units spawned and tested against lasers and torpedos.

- No errors or warnings
- Torpedos hit reliably
- Lasers hit reliably
- Collision boxes visually align with unit meshes

image

PR #6406, opened 2024-08-10
Add game version guard in the regular lobby

Garanas:

First implementation of the feature to not only embed the current game version (and commit hash) into what we deploy, but to also act on it to try and prevent desyncs.

One issue that occurs often when a release happens is a large number of desyncs. With these changes we introduce a guard to try and prevent the dissatisfaction that can occur when a release happens. 

You can have multiple instances of the game on a local machine and join the lobby through LAN/IP.

## Checklist


Garanas:
@4z0t would you be available to add a russian translation to the lobby message?

4z0t:
@Garanas added

PR #6413, opened 2024-08-17
Redundancies to avoid lua errors

maudlin27:
This fixes two lua errors I received in a desynced replay (was an AI vs AI game I was testing offline hence no replay ID to refer to).

The first was:
WARNING: Error running lua script: ...\gamedata\units.nx2\units\xab3301\xab3301_script.lua(52): attempt to call method `SetRate' (a nil value)
         stack traceback:
         	...\gamedata\units.nx2\units\xab3301\xab3301_script.lua(52): in function `CreateVisibleEntity'
         	...ata\faforever\gamedata\lua.nx2\lua\remoteviewing.lua(57): in function <...ata\faforever\gamedata\lua.nx2\lua\remoteviewing.lua:49>

I've therefore added a check to make sure that the method SetRate isn't called if it doesn't exist (and for completeness have done the same with the following lines calling a separate method SetTargetSpeed)

The second error was:
WARNING: Error running lua script: ...ithub\fa\lua\sim\weapons\defaultprojectileweapon.lua(414): Game object has been destroyed
         stack traceback:
         	[C]: in function `RemoveEconomyEvent'
         	...ithub\fa\lua\sim\weapons\defaultprojectileweapon.lua(414): in function <...ithub\fa\lua\sim\weapons\defaultprojectileweapon.lua:412>

For this, I added a check if the object in question returned true for IsDestroyed() before going on to call RemoveEconomyEvent.

Re-running the replay with these changes, neither error appeared (and no other 'error running lua script' messages appeared either).

### DIFF
--- a/lua/sim/weapons/DefaultProjectileWeapon.lua
+++ b/lua/sim/weapons/DefaultProjectileWeapon.lua
@@ -411,7 +411,7 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
     ---@param self DefaultProjectileWeapon
     EconomyDrainThread = function(self)
         WaitFor(self.EconDrain)
-        if self.unit.BeenDestroyed and not(self.unit:BeenDestroyed()) then
+        if self.unit and not(IsDestroyed(self.unit)) then
             RemoveEconomyEvent(self.unit, self.EconDrain)
         end
         self.EconDrain = nil
--- a/lua/sim/weapons/DefaultProjectileWeapon.lua
+++ b/lua/sim/weapons/DefaultProjectileWeapon.lua
@@ -411,7 +411,9 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
     ---@param self DefaultProjectileWeapon
     EconomyDrainThread = function(self)
         WaitFor(self.EconDrain)
-        RemoveEconomyEvent(self.unit, self.EconDrain)
+        if not(self:BeenDestroyed()) then
+            RemoveEconomyEvent(self.unit, self.EconDrain)
+        end
         self.EconDrain = nil
     end,
 
--- a/lua/sim/weapons/DefaultProjectileWeapon.lua
+++ b/lua/sim/weapons/DefaultProjectileWeapon.lua
@@ -411,7 +411,7 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
     ---@param self DefaultProjectileWeapon
     EconomyDrainThread = function(self)
         WaitFor(self.EconDrain)
-        if not(self:BeenDestroyed()) then
+        if self.unit.BeenDestroyed and not(self.unit:BeenDestroyed()) then
             RemoveEconomyEvent(self.unit, self.EconDrain)
         end
         self.EconDrain = nil
--- a/units/XAB3301/XAB3301_script.lua
+++ b/units/XAB3301/XAB3301_script.lua
@@ -47,11 +47,12 @@ XAB3301 = ClassUnit( AStructureUnit ) {
             end
 
             if not self.ScryEnabled then
-                self.ScryEnabled = true 
-                
-                self.Animator:SetRate(1)
-                self.RotatorBot:SetTargetSpeed(12)
-                self.RotatorTop:SetTargetSpeed(-8)
+                self.ScryEnabled = true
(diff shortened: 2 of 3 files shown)


lua/sim/weapons/DefaultProjectileWeapon.lua
```diff
@@ -411,7 +411,9 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
     ---@param self DefaultProjectileWeapon
     EconomyDrainThread = function(self)
         WaitFor(self.EconDrain)
-        RemoveEconomyEvent(self.unit, self.EconDrain)
+        if not(self:BeenDestroyed()) then
```
Garanas, line 414: Is this checking for the correct entity to be (not) destroyed?

It checks whether the weapon (`self`) is destroyed, but the economy event is applied to the unit (`self.unit`)

units/XAB3301/XAB3301_script.lua
```diff
@@ -47,11 +47,12 @@ XAB3301 = ClassUnit( AStructureUnit ) {
             end
 
             if not self.ScryEnabled then
-                self.ScryEnabled = true 
-                
-                self.Animator:SetRate(1)
-                self.RotatorBot:SetTargetSpeed(12)
-                self.RotatorTop:SetTargetSpeed(-8)
+                self.ScryEnabled = true
+                if self.Animator.SetRate then self.Animator:SetRate(1) end
```
Garanas, line 51: I think you're checking for the wrong thing here. Whether the field (function) `SetRate` exists is not the problem - the problem is that `self.Animator` is not yet initialized.

These fields are initialized when the unit is done building. See also the body of `OnStopBeingBuilt`. That means that any logic inside this function can not run until the unit is done building. A better solution would be to check and verify that the unit is done building (there is a flag if I am not mistaken, or an engine function) and just early exit the function as a whole.

This also applies to the other changes at lines 52-55.

Garanas (changes requested):
See the comments.

units/XAB3301/XAB3301_script.lua
Garanas: This may apply to all the vision-related functionality in general, but I have not reviewed the remaining code at this moment.

lua/sim/weapons/DefaultProjectileWeapon.lua
maudlin27: Agreed it looks like it should check if self.unit is valid - I've made changes to the code, but my nervousness with this is I couldn't get a log added in the same place to trigger, and I dont have the replay that gave rise to the error anymore

Garanas: You can use the global `IsDestroyed` instead of the table entry `BeenDestroyed` if you're afraid of whether the reference exists or not.

units/XAB3301/XAB3301_script.lua
maudlin27: I've added a check whether the unit itself is constructed (if I've understood your comments correctly) - it runs without error in a sandbox (but then in the sandbox scenario the unit was fully constructed)

lua/sim/weapons/DefaultProjectileWeapon.lua
maudlin27: Presumably that would be IsDestroyed(self.unit)?

units/XAB3301/XAB3301_script.lua
```diff
@@ -36,6 +36,7 @@ XAB3301 = ClassUnit( AStructureUnit ) {
     end,
 
     CreateVisibleEntity = function(self)
+        if not(self.GetFractionComplete) or self:GetFractionComplete() < 1 then return nil end
```
Garanas, line 39: Out of curiosity - why do you check for the existence of the `self.GetFractionComplete` field?

BlackYps:
What's the status here?

units/XAB3301/XAB3301_script.lua
maudlin27: If it doesn't exist (for whatever reason) then the code would cause an error - e.g. if code elsewhere was to pass an invalid unit reference or a nil value, so I added for completeness

maudlin27:
@Garanas See BlackYps query

BlackYps:

PR #6424, opened 2024-08-29
UnitviewDetail.lua: Introduce total DPS calculations

Basilisk3:
Shows a DPS total for all weapons not categorized as 'Anti Air' or 'Anti Navy' in the additional unit details displayed when `Show Armament Detail in Build Menu` is enabled in the settings.
- Useful for units with multiple weapons of this type, such as the Uashavoh (Seraphim Destroyer), Ythotha or the Mongoose.
- Accounts for multiple instances of the same weapon being present. Example: the Fatboy is correctly displayed as having a total DPS of 4000.
- ~~Also introduce a DPS/M calculation; I do not believe this is calulcated anywhere in th UI or the unit databases.~~

## Checklist
- [x] Changes are annotated, including comments where useful.
- [x] Changes are documented in the changelog for the next game version.
- [x] Remove the entry if the unit only has 'Anti Air' or 'Anti Navy' weapons, or no weapons at all.
- [x] Improve the naming of the stats.
- [x] Split the ground DPS stat into direct fire and indirect fire DPS.


BlackYps:
I generally like this addition, but I think displaying a dps/mass stat is going a bit overboard. This seems more appropriate to be displayed in a unit database instead.

lL1l1:
Agree about dps/mass belonging in a unit db.

Basilisk3:
Can you elaborate on why you think they are out of place here? I moved the DPS/M calculations into a mod to test them out and found them to be pretty useful. There is also still place in the UI to accommodate them.

BlackYps:
DPS/mass is a derivative stat. We can think of many more derivative stats that could be used to reason about a unit's efficiency. HP/mass, HP*DPS/mass², and many more.
It's not so much that this single stat will crowd the UI, but what do we say to the calls that want an additional stat implemented? It's not like DPS/mass is vastly superior to any other derivative stat. I don't see a rational argument why we need this one specifically and not one of the others.
If you want to think about which units are generally efficient you don't do that in an actual game anyway, because nobody has time for that, so doing it in a unit db works just as well.
Actually now that I am thinking about it, I'm not sure if implementing DPS/mass in a unit DB is a good idea either. The thought what we should do with other similar stats remains the same. It also has the chance that people will hyperfocus on these stats in their future balance ideas, not realizing that things like speed and range are also important. But I think all this is less severe in a unit DB. We can more easily fit many stats and we already get bad balance ideas anyway. So I am not against implementation in a unit DB, but I think it should be discussed more broadly first, to see what others think of these arguments.

Oh, and before you say that DPS is a derivative stat: That's technically true, but first of all, it's unrealistic to remove this again and it's a very useful stat because it smoothes out arbitrary variations in the firing cycle between units. The firing cycle is only important in edge cases, so arguably DPS is the most important stat of a weapon and not the damage of a single shell.

Basilisk3:
> DPS/mass is a derivative stat. We can think of many more derivative stats that could be used to reason about a unit's efficiency. HP/mass, HP*DPS/mass², and many more. It's not so much that this single stat will crowd the UI, but what do we say to the calls that want an additional stat implemented? It's not like DPS/mass is vastly superior to any other derivative stat. I don't see a rational argument why we need this one specifically and not one of the others. 

This is a slippery slope argument. HP and DPS are by far the most important stats a unit has, so it makes sense to only implement _their_ "per mass" derivatives and not, for example, Speed/Mass or Range/Mass. Adding these two additional stats does not force us to implement more nor do I expect anyone to request something like Speed/Mass. The only other thing I could imagine being useful is BP/Mass for engineering units. I also disagree with not adding more stats simply because they are derivative. If they are important, they should be made visible inside the game.

> It also has the chance that people will hyperfocus on these stats in their future balance ideas, not realizing that things like speed and range are also important.

This is already the case anyway and I think that it is highly unlikely that my changes will have any tangible effect on this. At the end of the day, it is additional information, what other players choose to do with it is not something I can control or will worry about.

lL1l1:
I think a good compromise between having more technical stats available and not having to delve into external websites is keeping DPS/mass in the in-game unit manager, as is already the case although it's showing damage/mass instead of DPS/mass. `UnitsTooltip.lua` is the relevant file.

In my opinion people do not have the time to evaluate the cost/DPS/efficiency/buildtime/buildpower/range/HP and so forth combined stats with their interfering effects when playing a standard FAF match. That is implying that it could be useful as an optional UI mod for players who want to analyze new stats in a slower paced environment like modded games or development environments.

Basilisk3:
The unit manager is only available through the lobby though, right? I have only ever used it for unit restrictions, and aside from you, I don't believe many others use it for anything else. Do not get me wrong, there is no harm in adding these calculations to the unit manager, but I would rather have it all in one place, and I think uvd is our best option.

A combined debugging/analysis tool that does the same as uvd.lua and the blueprint manipulator, but with even more stats, would be a nice addition too.

> In my opinion people do not have the time to evaluate the cost/DPS/efficiency/buildtime/buildpower/range/HP and so forth combined stats with their interfering effects when playing a standard FAF match.

This is the case for a lot of stats that uvd provides, most stats only really make sense if they are compared to other stats from other units. At the end of the day, it is an additional option in the menu and the window is packed with information. Especially newer players are not going to care about this addition. It is going to matter exclusively for seasoned players and people who play with lots of mods.

BlackYps:
> This is a slippery slope argument.

Well yes, but what do you say to people that say "I find BP/mass and shieldHP/mass a useful stat as well, please add it. Also HP*DPS/mass² is a better metric of general efficiency, so please add this as well"?

You say that DPS/mass is important, but you also say that newer players are not going to care about this addition. It is going to matter exclusively for seasoned players and people who play with lots of mods.
I agree with all that and that's why I am not convinced that it should be added. Especially because I have never heard of seasoned players actively wanting this. Mostly, the units are not so similar that you should pick the numerically most efficient one and ditch the rest. For good balancing we want to avoid scenarios like this specifically, because it will always lead to at least one unit losing that comparison and thus never be a good idea to build. Instead the units are differentiated by their role they are useful in.
If we want to help people get a better understanding of what units to use for what, we should rather think about improving the unit description. But I digress because this was never intended for newer players.

My bottom line is, I think this should rather be a UI mod. Otherwise I would need more evidence (read: more people saying they want this feature and an example how this is actually relevant information that drives your choices in the game) that DPS/mass is important enough to add it to the base game.

Basilisk3:
> Well yes, but what do you say to people that say "I find BP/mass and shieldHP/mass a useful stat as well, please add it.

It makes sense to add BP/mass in my opinion. This does not mean we have to add calculations for more than just the important stats.

> Also HP*DPS/mass² is a better metric of general efficiency, so please add this as well"?

As state above, a line has to be drawn somewhere. I think my changes do not go overboard as they only calculate "efficiency" based on broad weapon categories and mass. Introducing three or more variables is where I draw the line. 

> You say that DPS/mass is important, but you also say that newer players are not going to care about this addition. It is going to matter exclusively for seasoned players and people who play with lots of mods. I agree with all that and that's why I am not convinced that it should be added. 

You can apply this argument for almost all changes I did to uvd.lua. All of my prior PRs changing this file are essentially completely irrelevant to newer players.

> Especially because I have never heard of seasoned players actively wanting this. Mostly, the units are not so similar that you should pick the numerically most efficient one and ditch the rest. 

This is because they do not know it is possible or do not know they want the change until it already happened. Again, the same can be said about all of my changes to this file.

> For good balancing we want to avoid scenarios like this specifically, because it will always lead to at least one unit losing that comparison and thus never be a good idea to build. Instead the units are differentiated by their role they are useful in. If we want to help people get a better understanding of what units to use for what, we should rather think about improving the unit description.

Implying that this change will negatively interfere with how the game is balanced is a huge stretch and entirely unrealistic.

The problem you touch upon is called selective perception. If you choose to ignore relevant stats, you could argue, for example, that either the Obsidian or Ilshavoh are strictly superior to each other. This is obviously a fallacy and easily disproven by other stats that are listed in the same window (e.g. range).

> My bottom line is, I think this should rather be a UI mod. Otherwise I would need more evidence (read: more people saying they want this feature and an example how this is actually relevant information that drives your choices in the game) that DPS/mass is important enough to add it to the base game.

UI mods cannot hook strings_db.lua, so converting this PR into a UI mod is not feasible, other people have also run into this issue. Therefore if only the DPS summary is included, all other changes will essentially be lost. Unfortunately, this will be the most likely route for this PR, as I would like that at least the DPS summary gets merged.

BlackYps:
> It makes sense to add BP/mass in my opinion

It doesn't make sense in my opinion.

> You can apply this argument for almost all changes I did to uvd.lua.

Yeah, but I wasn't involved during that time.


> Implying that this change will negatively interfere with how the game is balanced is a huge stretch and entirely unrealistic.

This seems to be a misunderstanding. I'm not saying that this change will affect balance. What I was trying to say is that the units have different roles. There are no units that have the same role and one is more efficient than the other, but I think in mods this is the case? This efficiency difference is what the DPS/mass stat would reveal. But because we don't have this situation in the base game, this stat seems useless to me. Looking at it like this, I fail to see how this would be beneficial for seasoned players either. It might be beneficial for the balance team though.



>> Especially because I have never heard of seasoned players actively wanting this. Mostly, the units are not so similar that you should pick the numerically most efficient one and ditch the rest.
>
> This is because they do not know it is possible or do not know they want the change until it already happened.

> UI mods cannot hook strings_db.lua, so converting this PR into a UI mod is not feasible,

That's unfortunate. But sim mods can? As it is now, as I only see benefit for modded units it wouldn't be an issue that it is a sim mod.
If you still want it in the base game I would need to hear some more voices that this is useful. You could start a discussion in discord that serves as a reality check for us both to how seasoned players really think about this.

Basilisk3:
> Yeah, but I wasn't involved during that time.

`It is going to matter exclusively for seasoned players and people who play with lots of mods. I agree with all that and that's why I am not convinced that it should be added.` 

Your point was that if additions only matter for a smaller portion of players, it does not make sense to add them. My examples aim to show why I disagree with this way of evaluating changes.

> This seems to be a misunderstanding. I'm not saying that this change will affect balance. 

Well, this is what you wrote: 
`For good balancing we want to avoid scenarios like this specifically, because it will always lead to at least one unit losing that comparison and thus never be a good idea to build. Instead the units are differentiated by their role they are useful in.
If we want to help people get a better understanding of what units to use for what, we should rather think about improving the unit description.`

> What I was trying to say is that the units have different roles. There are no units that have the same role and one is more efficient than the other, but I think in mods this is the case? This efficiency difference is what the DPS/mass stat would reveal. But because we don't have this situation in the base game, this stat seems useless to me. Looking at it like this, I fail to see how this would be beneficial for seasoned players either. It might be beneficial for the balance team though.

The stat does matter if you want to deduct the performance of units from their bp-values. This should not be the only course of action when evaluating units (theory crafting), but it does give a first insight as to how a unit might perform.

Examples: 
- DPS/Mass comparison between Interceptors and ASFs
- Obsidian vs. Ilshavoh.
- BP/Mass for engineers

> That's unfortunate. But sim mods can? As it is now, as I only see benefit for modded units it wouldn't be an issue that it is a sim mod. If you still want it in the base game I would need to hear some more voices that this is useful. You could start a discussion in discord that serves as a reality check for us both to how seasoned players really think about this.

Sim mods can hook strings_db.lua. I reverted the changes so at least the DPS summary can get merged. In case someone wants to make a sim-mod, he can take my code from this PR.

changelog/snippets/features.6424.md
```diff
@@ -0,0 +1,5 @@
+- (#6424) Introduce total DPS and DPS/Mass calculations in the additional unit details displayed when `Show Armament Detail in Build Menu` is enabled in the settings. This is useful for summarizing or comparing the stats of units with multiple weapons. Previously, the DPS/Mass stat was not displayed in the UI or the unit databases, so they had to be calculated by hand.
```
BlackYps, line 1: The changelog still mentions DPS/mass. Other than that this PR looks good

BlackYps (changes requested):

BlackYps (approved):

PR #6427, opened 2024-09-03
Fix Seraphim carrier/T3 Sub and Cybran carrier AA missing their targets

lL1l1:
## Issue
The Seraphim carrier and T3 sub have the same weapon as the Seraphim SAM (by name and visuals), but don't have any of the expected tracking properties. Their muzzle velocity and pitch ranges are also slightly different.
Cybran carrier's projectile has tracking which turns off the weapons' target leading. The tracking is not strong enough for the projectiles to correct their aim with the high velocity of the projectiles. The Cybran carrier actually has a worse hit rate than the Seraphim carrier because of the combination of turning off leading and too weak tracking.

Copy the Seraphim SAM projectile physics to the carrier/T3 sub projectile (it is shared). Also copy it to the Cybran carrier's projectile to improve its hit rate.
Copy the Seraphim SAM muzzle velocity and turret pitch to the carrier/T3 sub so they can aim straight upwards.

Turn on `NoDamage` in the console, then patrol a spy plane (fastest aircraft) around the edge of the carrier range and see that it is getting hit.

## Checklist


clyfordv (approved):
Tested, as advertised, aa performance on the given units is greatly improved.

PR #6428, opened 2024-09-03
Implement a Dual Yaw turret controller so that Loyalist can fully aim its secondary

lL1l1:
## Issue
Since Loyalist's secondary is on a turret that is mounted to the Torso bone, it can only aim 30 degrees left or right of where the torso is aiming. In theory, the secondary weapon should be able to fire in any direction since the torso can rotate. However, the torso is only controlled by the primary weapon.

Adds a universal way to add a secondary yaw controller for weapon turrets. This allows the Loyalist's secondary weapon to control the torso when the primary weapon is not targeting anything within range.

Additionally, adds `PrefersPrimaryWeaponTarget = true` in the secondary weapon's BP to fix it getting stuck on T1 units when the primary is aiming at a higher tech unit in a different direction.

The Loyalist now rotates its torso to aim at targets that are out of primary weapon range but within secondary weapon range. The secondary weapon no longer gets stuck on low-tech targets that it cannot hit when the primary weapon is turning the torso elsewhere.

## Screenshot
Notice how both weapons fire at the Othuum. Previously the secondary weapon would get stuck on the T1 tank.
image

## Checklist


clyfordv (approved):
Tested, works as advertised, smooth implementation.

PR #6430, opened 2024-09-04
Fix an OnImpact error for CZAR depth charges

lL1l1:
## Issue
The `HasImpacted = true` line was not added back in during the refactor in #4517.

Add back the line.

The CZAR no longer gives errors after hitting an underwater unit with the depth charges.

## Checklist


BlackYps (approved):
Nice catch!

PR #6434, opened 2024-09-07
Fix and improve the FPS cap option

lL1l1:
## Issues
`gamemain.lua`:
- sets the frame time clamp every time the game starts, overriding the frametime option in the game settings menu (which I don't think changes the fps cap every game startup either).
- has a hard cap of 60 fps for fullscreen adapter options
- always sets the cap to 100 fps for windowed

The vsync game option caps fps at 60 despite the engine doing the capping and syncing correctly by itself.

- For `AdjustFrameRate` in `gamemain.lua`:
  - Use the `frametime` option if it's set (fixes the functionality of the option).
  - Use the framerate of the selected fullscreen adapter instead of capping at 60 fps
  - Find the max framerate for windowed mode by checking the list of adapter options.
- For the vsync option, remove the part that changes frametime back to 16 when enabling vsync.
- Frametime option: decrease minimum frametime further, since it only went to 250 fps before.

Checking with the in-game FPS counter, changing the in-game settings adjusts the framerate correctly, including when restarting the session.

## Checklist


Basilisk3 (approved):
Tested and works. Also does not break FPS mods, so this should be fine to include.

PR #6435, opened 2024-09-07
Add ability for Soul Ripper weapons to fire at gunships

RabidPope:
The Soul Ripper is targetable by gunships, but is useless against them. This change gives it its own niche in that it can contribute to defending against gunship threats. 

Like ground exp's, it loses to lower mass investment in t3 counterparts of around 22k-25k, so that is about 16 wailers worth. 
T2 gunships get completely countered though.


lL1l1 (commented):
Changes to target layers looks correct, I did the same independently before. Do the target priorities need updating with these changes?

RabidPope:
> Changes to target layers looks correct, I did the same independently before. Do the target priorities need updating with these changes?

From what I observed, if it is firing at ground units, it continues doing that. Same apply's to gunships.

clyfordv (approved):
Changes are good, hot SR-on-gunship action confirmed. Synergizes with: https://github.com/FAForever/fa/pull/6383#pullrequestreview-2379671027

RabidPope:
Any chance of this going through or should I close this?

lL1l1:
Keep it open; balance team had a change of leadership and we're busy with many other changes during this surge of activity so we just haven't gotten to this PR.

Tagada14 (approved):

PR #6436, opened 2024-09-07
Prevent the logging of an unecessary warning when certain units make landfall

### LINKED ISSUE #6136: Rare error when the GC, ML and Megalith make landfall
## Describe the bug

This is the same issue as fixed in #5829. It is a rare bug, but it does happen in real games.

The error: 
```
WARNING: *WARNING: No movement effect groups defined for unit ual0401,
Effect groups with bone lists must be defined to play movement effects. 
Add these to the Display.MovementEffects Land. Effects table in unit blueprint. 
```

The definition of the error is found in Unit.lua, line 3732.

## How to reproduce the error:
- Spawn a line of GCs via the Alt+F2 menu (the line must be both on land and in the water)
- Give a movement order to the units you spawned in

## Additional context

Related to: #6135; #5924; #5829

Basilisk3:
Issuing this warning even if a `Footfall` table is defined in the blueprint is incorrect, at least for the units I am familiar with. Units like the GC may not have an `Effects` table in their blueprint, but their movement effects still play correctly without it.

- The warning is not logged anymore if a `Footfall` table is present in the blueprint.
- Remove the doubling of the word `Warning` in the log.
- Remove a now unnecessary change to the Salem.

Resolves #6136.

## The warning and how to reproduce it:

```
WARNING: *WARNING: No movement effect groups defined for unit ual0401,
Effect groups with bone lists must be defined to play movement effects. 
Add these to the Display.MovementEffects Land. Effects table in unit blueprint. 
```

- Spawn a line of GCs via the Alt+F2 menu (the line must be both on land and in the water).
- Give a movement order to the units you spawned in.

The error message also occurrs in non-sandbox games.

All units functions properly, no errors in the log.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6436.md
+++ b/changelog/snippets/fix.6436.md
@@ -1,0 +1,1 @@
+- (#6436) Prevent the logging of an unecessary warning when certain units make landfall.
--- a/lua/shared/components/DebugComponent.lua
+++ b/lua/shared/components/DebugComponent.lua
@@ -35,7 +35,7 @@ DebugComponent = ClassSimple {
     EnabledLogging = true,
     EnabledWarnings = true,
     EnabledErrors = true,
-    EnabledDrawing = true,
+    EnabledDrawing = false,
 
     --#endregion
 }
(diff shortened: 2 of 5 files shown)


lL1l1 (commented):
The root of the issue is fixed with a8c847f8da278d8504a950807da282df4a7c7e2c.

The blueprint annotations for movement effects are rather disconnected from how the tables actually look. That should be fixed next.

lL1l1 (approved):
Tested:
- No warnings when spawning GCs in the water and walking onto land/sea.
- Footfall still does damage.

PR #5061, opened 2023-06-13
Update vector processing

Hdt80bro:
Adds metamethods for vectors (quaternions in particular) and cleans up lots of vector code.

PR #6438, opened 2024-09-09
Finalize "Update vector processing" (#5061)

lL1l1:

1. Rebase #5061 to the latest develop version while fixing conflicts.
2. Fix structure alignment to terrain.
3. Test and code review the rest of the changes.

<details> <summary> Testing structure orientation: </summary>

Put this command into the console then spawn units by pasting from the clipboard like normal.
```lua
UI_Lua 
local squareSize = 9
local outStr = ''
for x=0, squareSize do
   for z=0, squareSize do
      outStr = outStr .. string.format([[CreateUnitAtMouse('ueb0301', 0,    %f,  %f, %f)]]..'\n'
      , x * 8 - squareSize/2*8
      , z * 8 - squareSize/2*8
      , Random() * math.pi
   )
   end
end
CopyToClipboard(outStr)
```
Results (map is "Project Albus"):
image

The raw numbers when logged are very slightly different but I don't see a visual difference.

</details>

<details> <summary> Generate a bunch of wreckages using ScenarioUtilities </summary>

```lua
SimLua 
local SUCreateWreckage = import('/lua/sim/ScenarioUtilities.lua').CreateWreckage
for i = 0, 50 do
   local unit = CreateUnit2('xsa0402', 1, 'Air', 100 * Random(), 100 * Random(), math.pi * Random())
   ForkThread(function()
      WaitTicks(1)
      SUCreateWreckage(unit, true)
   end)
end
```
Seems to still create random orientations. Unfortunately due to the `TryCopyPose` in `Unit:CreateWreckageProp` this doesn't work on units not in the air layer (land, navy, landed air units), so it must be tested with an air unit + 1 tick delay to get into the air layer (spawns on land layer).
</details>

Reviewed the `utilities.lua` and `TerrainUtils.lua` changes to make sure the logic stayed the same.

Unit rotation functions review:
Just spawned a tank unit (walkers always stay vertical) on a slope and saw the commands do what the annotations say:

```lua
SimLua 
ArmyBrains[1]:GetListOfUnits(categories.ALLUNITS)[1]:Rotate(45)
```

```lua
SimLua 
ArmyBrains[1]:GetListOfUnits(categories.ALLUNITS)[1]:SetRotation(45)
```

Ran a small m27AI skirmish with no errors.

## Checklist


lua/shared/quaternions.lua
lL1l1, line 1: Do we want to keep the old quaternions library for backwards compatibility?

lua/sim/units/StructureUnit.lua
```diff
@@ -19,7 +19,6 @@ local EffectTemplate = import("/lua/effecttemplates.lua")
 local TerrainUtils = import("/lua/sim/terrainutils.lua")
 local Buff = import("/lua/sim/buff.lua")
 local AdjacencyBuffs = import("/lua/sim/adjacencybuffs.lua")
-local Quaternion = import("/lua/shared/quaternions.lua").Quaternion
```
lL1l1, line 22: Lost backwards compatibility here.

lua/defaultunits.lua
```diff
@@ -66,7 +66,6 @@ local Buff = import("/lua/sim/buff.lua")
 local AdjacencyBuffs = import("/lua/sim/adjacencybuffs.lua")
 local FireState = import("/lua/game.lua").FireState
 local ScenarioFramework = import("/lua/scenarioframework.lua")
-local Quaternion = import("/lua/shared/quaternions.lua").Quaternion
```
lL1l1, line 69: Lost backwards compatibility here.

lua/sim/units/StructureUnit.lua
```diff
@@ -119,20 +118,18 @@ StructureUnit = ClassUnit(Unit, BlinkingLightsUnitComponent) {
...
-                -- quaternion magic incoming, be prepared! Note that the yaw axis is inverted, but then
-                -- re-inverted again by multiplying it with the original orientation
-                local quatSlope = Quaternion.fromAngle(0, 0 - a2,-1 * a1)
-                local quatOrient = setmetatable(self:GetOrientation(), Quaternion)
-                local quat = quatOrient * quatSlope
-                self:SetOrientation(quat, true)
+            if roll ~= 0 or pitch ~= 0 then
+                -- "q′ = q2 * q1  in which q′ corresponds to the rotation q1 followed by the rotation q2" (wikipedia)
+                -- the unit's orientation comes first, and then is rotated by the terrain angle
+                -- roll is negated because ?
```
lL1l1, line 131: For what reason is the roll negated?
Also I'm not sure why it is important that here we do the `terrain quaternion * orientation` when other parts in this PR do `orientation * quaternion` just fine.

lL1l1:
Also I must've made a mistake in `Projectile.lua` `OnTrackTargetGround` because the game doesn't use `y,z,x,w` format quaternions but I haven't solved that yet, and the code is at least functional.

lL1l1:
The utility tests fails because of https://github.com/FAForever/fa/blob/b93c634653034d4b452c278d978895e7ccbbccad/lua/system/utils.lua#L885
but this line works in-game, since Vector2 is a Core global.

lua/shared/quaternions.lua
Garanas: If you're referring to https://github.com/FAForever/fa/blob/develop/lua/shared/quaternions.lua then perhaps the only mod that may reference it is the FAF Superior Experience mod. I do not recall the GitHub tag of the author.

If we're going to use this new approach it is probably better to just remove the file.

lua/sim/units/StructureUnit.lua
Garanas: That is probably best asked to @Hdt80bro , I do not recall myself. I think I negated it because otherwise it would not orientate properly.

lua/shared/quaternions.lua
Garanas: Ah, I now see the file that is part of the review 😃 .

Garanas:
This is great @lL1l1 , it would be great to finalize this and get rid of all the magic mathematics that is in some of the Lua files 👍

lua/sim/units/StructureUnit.lua
Hdt80bro: You did, and I tried to extract the original mathematics that would interpret roll correctly without negation, but got stuck. Similarly, since I didn't quite figure it out, sometimes it needed to be multiplied on the left to rotate, others multiplied on the right (it is sensitive to the side its multiplied on). It's something I wasn't happy about and why this PR got stuck in limbo.

lua/system/utils.lua
```diff
@@ -1002,12 +1002,6 @@ function vector_metatable.__mul(a, b)
         end
 
         -- Quaternion * Vector (Quaternion with no spin)
-        local len = b1*b1 + b2*b2 + b3*b3
-        if len ~= 1 then
-            -- normalize any vectors that might not represent a direction cosine
-            len = math.sqrt(len)
-            b1, b2, b3 = b1 / len, b2 / len, b3 / len
-        end
```
lL1l1: Is it fine to remove the automatic normalization of vectors when multiplied by a quaternion? I don't think this issue would be that frequent.

Garanas: I don't know - I am not good at maths. Best to ask @Hdt80bro or maybe @BlackYps

BlackYps: I'm good at maths, but haven't really dealt with quaternions, so I can't help you without first spending some hours learning quaternion operations

lL1l1: According to my research quaternions get multiplied by vectors only to rotate a vector by an orientation. This would be operations such as rotating a direction vector (such as the output of `GetBoneDirection`, which is XYZ direction, not RPY fyi), velocity vector (possible use in the projectile splitter component), or position vector (`OnTrackTargetGround`) by an orientation quaternion. In none of these cases do you want to lose the magnitude of the vector. Basically you always want to input and output a non-unit vector.

engine/Core.lua
```diff
@@ -465,51 +483,57 @@ end
 function VDist2Sq(x1, y1, x2, y2)
 end
 
---- Computes the distance between the vectors `a` and `b`
+--- Computes the distance between vectors `a` and `b`
+--- It is always faster to compute it in Lua.
```
4z0t, line 487: Wrong. It was optimized and now it is faster.

lL1l1: Confirmed that the engine function is faster with benchmark.

lua/sim/units/StructureUnit.lua
lL1l1: I spent some time figuring it out, the reason roll is negated is because the terrain calculation's roll is given with respect to the `-z` axis, and units use `+z`.

terrain orientation roll explanation - milton

This is the x component of the basic form of the calculation for small structures (only 1 edge is calculated).
The number we get is `atan2(dy, lx)`, where `dy` is the terrain height in the middle of the `-x` side edge. This number is the trigonometric representation (local perspective) in the bottom right.
But notice how in the diagram of the real world situation (global perspective) in the bottom left the triangle is flipped horizontally, and that when using right-handedness, the angle we get is on the `-z` rotation axis.

Averaging is done correctly by negating the second component, which is measuring the other side's angle which is flipping the axis of rotation 180 degrees.
https://github.com/FAForever/fa/blob/81a95af49e8a6a58c76b3512f56809ef1b9c4505/lua/sim/TerrainUtils.lua#L24-L25

equivalent to `avg = (angle1 + -angle2) / 2`

In comparison to Pitch, Roll is done incorrectly because it starts with `pos - lengthX/2` towards the `-x` side edge which gives an angle around `-z`, but Pitch is correct because it also starts with `pos - lengthZ/2`, and that goes to the `-z` side edge, but that angle rotates around `+x`, which is the same as units.

lL1l1:
@Hdt80bro @Garanas @BlackYps I think everything is figured out and this PR is finished. Please take some time to review it.

lua/system/utils.lua
BlackYps: Yeah, if vector normalization is not in the standard process of multiplying vectors with quaternions then we should not do it. I appreciate the good explanation and agree with the conclusion.

lua/system/utils.lua
Garanas, line 1: As I am not that great at mathematics, I'll leave the review of this particular file to @BlackYps, @Hdt80bro or perhaps even @clyfordv .

Garanas (approved):
I've reviewed the behavior in-game and all appears to be good. The relevant code changes where the new logic is applied look tremendously better. 

An absolute high-quality contribution. Nothing but exceptional. This is the type of in-depth analysis at an almost professional level.

engine/Core.lua
```diff
@@ -434,20 +450,23 @@ function Trace(enable)
...
+---@deprecated It is faster to compute it in Lua with `Vector`
 ---@param a Vector
 ---@param b Vector
 ---@return Vector
 function VDiff(a, b)
 end
 
---- Computes the distance between two points
+--- Computes the distance between two points.
+--- It is always faster to compute it in Lua.
```
BlackYps, line 469: Shouldn't these functions that are faster in lua also be deprecated?

BlackYps (approved):
I didn't trace every single operation, but that everything in the game works correctly gives me confidence that the math is correct too.
The files are much more readable now, very good work!
If I am just out of the loop about which functions are supposed to be deprecated, then you can merge this

engine/Core.lua
lL1l1: Yes, I will mark them as deprecated.

lL1l1:
A problem with the __add metamethod is that it has quite a few safety checks that slow down the operation. It can be faster than engine functions and even lua since it has the vector metatable cached for use with setmetatable (faster than `Vector()`).
Should the vector metatables be split for the different types? Then the only thing that needs to be compared is `getmetatable(b) == vector/quat/vector2_metatable`.

4z0t:
All vector metatables can be expanded via bin patches and done with performance priority

BlackYps:
I'm afraid I neither understand the question, nor your accompanying explanation...

lL1l1:
Currently Vector, Vector2, and Quaternion use the same `vector_metatable` to do their operations. Because of this, a lot of the operations have quite a bit of type checking slowing them down. If each type had its own metatable, then over half the type checks wouldn't be needed as the type would already be known.
Should that be done here or leave it for later?

BlackYps:
It does sound like a good idea, but I suggest to do that in a separate pull request. This one is already quite big and I'd like to have it merged. It's already a viable pull request.
Further improvements can be split out to make reviews easier.

PR #6440, opened 2024-09-09
Unify the MuzzleVelocity stat of the Aeon Cruiser's missiles

Basilisk3:
The Infinity Class Cruiser is equipped with two copies of Zealot AA Missile launchers, both of which target the same unit. These two weapons function identically except for the `MuzzleVelocity` stat, where one weapon has a higher value than the other (though the left weapon also does not have a `TargetCheckInterval`, and prefers the target of the right weapon). This is misleading because both unit databases show the AA batteries as having a `MuzzleVelocity` of `40`, when in reality, this is only true for one of the launchers. The models on the unit also look identical.

I am not aware of any other unit that does this; the Keefer Class has a MuzzleVelocity of 30 for both of its AA batteries. The UI also does not indicate a difference in MuzzleVelocity between both weapons.

**Infinity Class: T2 Cruiser (UAS0202):**
   - Zealot AA Missile (right launcher):
      - MuzzleVelocity: 40 --> 35
   - Zealot AA Missile (left launcher):
      - MuzzleVelocity: 30 --> 35

## Checklist


lua/shared/components/DebugComponent.lua
```diff
@@ -35,7 +35,7 @@ DebugComponent = ClassSimple {
     EnabledLogging = true,
     EnabledWarnings = true,
     EnabledErrors = true,
-    EnabledDrawing = true,
```
Garanas, line 38: Did you start from `deploy/fafdevelop`? These lines should not be changed. Instead, start from `develop` next time.

lua/version.lua
```diff
@@ -28,9 +28,9 @@
 -- - https://github.com/FAForever/fa/blob/develop/.github/workflows/deploy-fafbeta.yaml
 -- - https://github.com/FAForever/fa/blob/develop/.github/workflows/deploy-fafdevelop.yaml
 
-local GameType = 'unknown'  -- The use of `'` instead of `"` is **intentional**
+local GameType = "FAF Develop"  -- The use of `'` instead of `"` is **intentional**
 
-local Commit = 'unknown'    -- The use of `'` instead of `"` is **intentional**
+local Commit = "0a94e1c6818ef5bc937d4b91c86d24cf0bd8262c"    -- The use of `'` instead of `"` is **intentional**
```
Garanas, line 33: Same here, see the previous comment. These are set by the deployment workflows.

Garanas (dismissed):
See the comments.

Basilisk3:
I reverted the changes. I am unsure what caused it, but your explanation makes sense.

BlackYps (approved):

Issue #6238, opened 2024-06-09
Create a Github Workflow to convert markdown changelog files into Lua changelog files


Garanas:
Description
-----------

Needs: https://github.com/FAForever/fa/issues/6237

Create a new workflow file to convert the changelog files from the markdown format to a Lua format. This workflow should use the tool created in https://github.com/FAForever/fa/issues/6237 . The output of the workflow is an artifact with all the changelog files in a Lua format.

Zjonn:
Hi @Garanas, I have some questions about this issue:
- `with all the changelog files in a Lua format.` Do you want this workflow to parse all changleog files each time it's runned?
- Should it be runned  (only) on push to:
  - develop
  - deploy/fafdevelop
  - deploy/fafbeta
 - How about workflow_call and workflow_dispatch, should they be also added?

Garanas:
Yes - the goal is to have one place where we manage the changelog files. These are written in Markdown. We then generate lua-compatible changelog files from the markdown changelog files. For now, this is not a complicated task - we essentially add `"` to the start and end of each line and then generate a Lua table around it with the remaining properties. In the future we may want to extrapolate more information so that we can generate links to pull requests for example, but that's a feature for the future.

> Do you want this workflow to parse all changleog files each time it's runned?

Yes, the input is all Markdown changelog files in an input folder. And the output is one Lua file for each Markdown changelog file in an output folder. We then store that output in an artifact. That's the entire scope of this issue.

> How about workflow_call and workflow_dispatch, should they be also added?

Yes, the workflow should be a separate workflow that can trigger on both `workflow_call` and `workflow_dispatch`. Dispatch allows us to run it manually, useful for testing 😃 . Call allows us to run it from another workflow, which allows us to integrate it into the existing deployment workflows. 

To repeat myself - this issue is only about the new workflow, which should generate an artifact with all the Lua-compatible changelogs. Integrating it with the existing deployment workflows is another pull request/issue. It requires a little more work, as the game also needs to support the new format (one file for each changelog, instead of one massive file with all changelogs in it).

PR #6444, opened 2024-09-12
Create workflow converting markdown-based changlog to lua-based changelog

### LINKED ISSUE #6237: Convert a markdown-based changelog file into a Lua-based changelog file
Description
-----------

Part of https://github.com/FAForever/fa/issues/6236

Create a script, program or some other tool that we can retrieve in a Github Action to convert a markdown file that represents a changelog into a Lua-based template that has the changelog content in it.

As an example, the input could be:

```markdown
# Game version 3769 (24th of September, 2023)

Please refer to the notes of versions 3765, 3766, 3767 and 3768 for the full patch notes.

With gratitude to all those who took the time to report issues,

Jip

## Bug fixes

- (#5454) Fix a bug that can enable the Seraphim MML to launch an endless barrage of missiles.
- (#5454) Fix a bug where the command queue is stuck on a launch order.

This specifically occurred with the tech 3 strategic submarines and the tech 3 Seraphim battleship. As a result, the unit would be 'stuck' on the launch order, and any other order issued after the launch order would never be executed.

- (#5454) Fix a bug where distributed launch orders would not actually launch a missile.

This occurs when you issue launch orders and then distribute them. The new, distributed orders are processed and removed from the queue even though no missiles have been launched.

## Contributors

With thanks to the following individuals who contributed through coding:

- Jip
```

And the output would be:

```lua
{
    version = 3769, -- based on the file name
    description = { -- based on the content of the file
        "# Game version 3769 (24th of September, 2023)",
        "",
        "Please refer to the notes of versions 3765, 3766, 3767 and 3768 for the full patch notes.",
        "",
        "With gratitude to all those who took the time to report issues,",
        "",
        "Jip",
        "",
        "## Bug fixes",
        "",
        "- (#5454) Fix a bug that can enable the Seraphim MML to launch an endless barrage of missiles.",
        "- (#5454) Fix a bug where the command queue is stuck on a launch order.",
        "",
        "This specifically occurred with the tech 3 strategic submarines and the tech 3 Seraphim battleship. As a result, the unit would be 'stuck' on the ",
        "launch order, and any other order issued after the launch order would never be executed.",
        "",
        "- (#5454) Fix a bug where distributed launch orders would not actually launch a missile.",
        "",
        "This occurs when you issue launch orders and then distribute them. The new, distributed orders are processed and removed from the queue even though ",
        "no missiles have been launched.",
        "",
        "- (#5470) Fix a bug that cause gunships to have a tough time to hit certain units.",
        "",
        "When a projectile experiences multiple collisions in a single tick, the engine does not always return the first collision. This issue frequently occurs with gunships, as they often fire from nearly vertical angles. Consequently, gunships can appear to 'hit' the terrain, even though they clearly hit the intended unit. To resolve this, we now ensure that every collision box of every unit has a minimum length of 0.5 ogrids.",
        "",
        "- (#5470) Fix a bug that allows certain units to evade beams in specific conditions",
        "",
        "Most notable the Brick no longer has the ability to evade beams.",
        "",
        "## Contributors",
        "",
        "With thanks to the following individuals who contributed through coding:",
        "",
        "- Jip",
    }
},
```

A couple of considerations:

- (1) the maximum character width on a single line is 150 characters. You'll need to wrap the lines of the markdown file accordingly.

The format of the script, program and tool is up to the implementer. But please, whatever you do - keep it simple and maintainable. It should also be trivial to use it in a Github Action. Personally I think it is possible via a simple bash script that accepts an input file and output file.

Course of action
----------------

One could look into using Pandoc or Prettier to format the markdown file before converting it, see also:

- [Pandoc](https://pandoc.org/MANUAL.html#pandocs-markdown) - [installation instructions](https://pandoc.org/installing.html)
- [Prettier](https://prettier.io/docs/en/options.html) - [can be installed via npmjs](https://www.npmjs.com/package/prettier)

The entire process should work in a local environment.

Zjonn:
Resolves #6237 and #6238
Please take a look @Garanas 

## Description

Following PR is introducing python script converting Markdown files to Lua. 
Markdown files are formatted before conversion to Lua file using [mdformat](https://github.com/executablebooks/mdformat) (it's python alternative to prettier).

Why I chose python? There are several reasons:
- It comes preinstalled in most Linux distibutions
- Installed packages do not pollute the user environment (when using venv)
- Has better linters than for example bash (shellcheck)

Additionaly I wrote `setup_python_env.sh` script for users not familiar with python, it can create python virtual enviorement for them, install required packages and source it.

Tested manually in local environment

## Checklist


Garanas:
Hi @Zjonn, this looks exciting! I'll make some time next weekend to review it. Would you like me to incorporate it further into the workflow, or would you like to try and do that yourself?

Zjonn:
Hi @Garanas 
I will add this to the workflow myself (I hope 😄 ).

BlackYps:
Please set it to ready for review once you are done with your changes

Zjonn:
I added a workflow that can be run manually or through another workflow. 
This has been tested in my fa fork, the results are as follows https://github.com/Zjonn/fa/actions/runs/10984939551

In short:
- it takes all `*.md` files from changelog directory
- runs `markdown2lua.py` on each of them
- saves results in `lua_changlog` directory, which is later uploaded as an artifact

scripts/markdown2lua.py
Garanas, line 1: It is best to put these scripts in the following folder:

- https://github.com/FAForever/fa/blob/develop/.github/workflows/scripts

That way all the source files that are related to workflows are close to the source files of the workflows. 

Perhaps the file can be called `changelog-markdown-to-lua.py`, to match the convention of the other files.

scripts/requirements.txt
Garanas, line 1: I am not _that_ familiar with a Python programming environment. Given the results of Google, I suspect such a file as this is common practice. Just like `scripts/markdown2lua.py`, it is best to move this file to: https://github.com/FAForever/fa/blob/develop/.github/workflows/scripts

But given the generic name, maybe it is better if both this and the python file is in its own subfolder? That way future python scripts with different requirements would not clash. But again - I have little to no experience with Python so do share your thoughts.

scripts/setup_python_env.sh
Garanas, line 1: Same suggestion as with the other two files in this folder - please move it to:
- https://github.com/FAForever/fa/blob/develop/.github/workflows/scripts

Garanas (changes requested):
In general it looks good and the output is as promised! 

A general suggestion:

- (1) Generated files should be marked as generated. Please add a header block of comments stating that the lua changelog files are generated. Preferably with a reference to the source file and the logic responsible for generating it.

And two functional requests: 

- (1) The changelog files will be loaded as regular modules, via either the `import` statement or the `doscript` statement. For this the table inside the file needs to be named. Can you add a `Changelog = {...}` in front of each table?

- (2) It would good if the integrity of the generated lua files are checked automatically. In a separate job you can install Lua, retrieve the artifact and check that all files can be interpret. The step should fail if one (or more) files can not be interpret. Make sure to use the [needs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) field to tell the interpreter about the dependency between the jobs.

The last job is especially relevant since maintainers come and go. Not everyone may be aware of the details of this workflow file. And by testing the output automatically we can catch 'simple' issues long before it would reach any staging or production area.

For future reference, this was the output the review was made with:

- [lua_changelog.zip](https://github.com/user-attachments/files/17257699/lua_changelog.zip)

It took a bit longer to review this than anticipated, life got in between 😃 .

Zjonn:
Thanks for the review @Garanas.
I'll temporarily mark PR as a draft version to satisfy @BlackYps ;)

I like the idea of adding information that the file was generated automatically and how

>  Can you add a Changelog = {...} in front of each table?

Sure, no problem

> ld fail if one (or more) files can not be interpret. Make sure to use the [needs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idneeds) field to tell the interpreter about the dependency between the jobs.

Thanks for the tip

scripts/markdown2lua.py
Zjonn: One of the things I really like about Python is [PEP8](https://peps.python.org/pep-0008/#package-and-module-names), which defines a style guide for the entire language, taking file names into account as well. I wish more languages would do the same 😄. 

Having said that, `-` in file name is no go, but `changelog_markdown_to_lua.py` is, but it seems unnecesarry long to me, how about `changelog_markdown2lua.py`?

scripts/requirements.txt
Zjonn: Yes, requirements.txt is the simplest way to go. If you were developing a package, you would probably choose poetry these days and end up with a .toml file, but there is no need to complicate things, simple, well-known requirements should be enough.

Also, I don't know if anyone will want to use this script on Windows, but if they do, it might be a good idea to put a README.md in the scripts directory with the instruction that they need to install requirements.txt first.

I have mixed feelings about simply adding `python` subfolder in the `scripts`. 

```
scripts
|----python
      |---- markdown2lua.py
|----changelog-combine.sh
```

Doesn't look the best to me. I agree that it's a good idea to keep Python-related stuff in separate directories, so I suggest making separate directories for bash scripts as well, that way we end up with
scripts

```
|---- python
      |---- markdown2lua.py
|---- bash
      |---- changelog-combine.sh
```

What do you think?

Garanas: What if we go with feature-based naming, instead of language-based? For your python script it would be something like `markdown-to-lua` and then the main python script can be called `main.py` (or what the usual convention is). What are your thoughts on that?

Garanas: I see now your previous comment about pep8 - I'll leave this decision to you. Since you're clearly more aware of the best practices than I am, haha. But everything should reside in the `scripts` folder inside the `workflows` folder.

scripts/markdown2lua.py
Garanas: See also my comment here: https://github.com/FAForever/fa/pull/6444#discussion_r1788521168

scripts/markdown2lua.py
```diff
@@ -3,18 +3,27 @@
...
     version = {version},
     description = {{
 {description}
     }}
 }},
 """
 
+HEADER_SEPARATOR = "#" * MAX_LINE_LENGTH
+HEADER = f"""{HEADER_SEPARATOR}
+#{f"This file was autogenerated using {SCRIPT_NAME:}":^{MAX_LINE_LENGTH - 2}}#
```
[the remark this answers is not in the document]
Zjonn: Thanks, I had completely forgotten to check this, muscle memory worked

scripts/requirements.txt
Zjonn: I wanted to revisit your idea of putting Python-related stuff in `scripts/python` (by `scripts` I mean the `scripts` directory in the `workflows` folder.

What do you think of the following directory structure, I will adapt to your decision.
```
scripts
|---- python
      |---- python_related_stuff.py
|---- bash
      |---- currently-existing-scripts.sh
```

scripts/markdown2lua.py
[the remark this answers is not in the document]
Zjonn: Can look like this?
```lua
------------------------------------------------------------------------------------------------------------------------------------------------------
--                                                This file was autogenerated using markdown2lua.py                                                 --
--                                                           Source: fa/changelog/3636.md                                                           --
------------------------------------------------------------------------------------------------------------------------------------------------------
```

scripts/requirements.txt
Garanas: That seems fine, but how does one deal with conflicting `requirements.txt` files if we have multiple in the future?

Zjonn: There should be only one `requirements.txt` file, shared by all scripts. It will be the responsibility of the implementer not to affect other scripts if it decides to change the requirements already defined.

Zjonn:
I think I have made all the changes we have discussed. You can find newest workflow results [here](https://github.com/Zjonn/fa/actions/runs/11225533050).

Validation job reuses already existing `tests/run-syntax-test.sh`, I made it executable (`chmod +x`). 
I also moved all workflow bash scripts into bash subdirectory.

Garanas (approved):
Excellent work 👍

Garanas:
As an extension to your work here: https://github.com/FAForever/fa/issues/6472 , with that we have everything we need to start updating the Lua of the game. We'll save several megabytes of memory all together and we'll have consistent changelogs 😃 !

Garanas:
@Zjonn I forgot - the contributions have no license at the top of the file. Would it be fine to license the contributions of this pull request with the MIT license?

Zjonn:
@Garanas for some reason I don't see notifications when you are pinging me. MIT license sounds perfect! 

About #6472, I can't wait to see the end result. Amount of deleted lines in there looks great.

PR #6445, opened 2024-09-12
Small bug fixes for shields

lL1l1:

- Fix overcharge dealing half damage to shields. Reported on [Discord](https://discord.com/channels/197033481883222026/1283134865604218953).
- Prevent tree effect damage from going through shields
- Move anti-shield area damage before the normal area damage so that the normal damage can overkill things under the shield if the anti-shield damage kills the shield.

Shooting overcharge at a shield in a forest and checking its HP and that the trees on the other side don't get knocked down.

## Checklist


lua/shield.lua
```diff
@@ -516,13 +516,16 @@ Shield = ClassShield(moho.shield_methods, Entity) {
         local tick = GetGameTick()
 
         -- damage correction for overcharge
-        -- These preset damages deal `2 * dmg * absorbMult or armorMult`, currently absorption multiplier is 1x so we need to divide by 2
+        -- If the absorption multi is less than 1, then the shield will get hit by two instances of area damage, one of which is the remainder.
+        -- This means that the following code requires a multiplier to correct the total damage amount since both instances will be overriden.
+        -- In the past it used to be 0.25 absorption for static shields (due to structure armor) and needed a 2x multiplier.
```
BlackYps, line 521:
```suggestion
        -- In the past it used to be 0.25 absorption for static shields (due to structure armor) and needed a 2x division.
```

BlackYps (changes requested):
Other than the slip in the comments it looks good. Also tested it and it works

lua/shield.lua
lL1l1: Seems like the explanation is unclear. With 0.25 absorption we had a 0.25 damage and 0.75 damage instance hitting the shield. Both get set to 800 OC structure damage, but then 0.25x armor is applied again (second `OnGetDamageAbsorption` call), reducing it to 2x200 damage, which requires a 2x multiplication to bring it to the expected 800.

The old version of the line:
https://github.com/FAForever/fa/blob/aa17c0cf95570b71e2702db888e681dcdae23efe/lua/shield.lua#L514-L515

BlackYps: Yeah, it's indeed a bit confusing. I was looking at the removed / 2 division in line 523 and 525 and thought you meant that.

BlackYps: What is the purpose of this comment anyway? I'm not sure what it is supposed to achieve and if I were to read it at some point in the future I'm not sure if I would understand what it is trying to tell me

lL1l1: I'm trying to warn about how setting this damage value is wrong and can lead to issues if not all damage is absorbed, since the unabsorbed remainder also hits the shield and gets set + applied armor against again.

BlackYps: I think we should just include your example in the comment and make the wording more clear. My proposal:

If the absorption multiplier is less than 1, then the shield will get hit by two instances of area damage, the absorbed amount and the remainder.
This means that the following code then requires a multiplier to correct the total damage amount since both instances will be overriden.
For example with 0.25 absorption we would have a 0.25 damage and 0.75 damage instance hitting the shield. Both get set to 800 OC structure damage, but then 0.25x armor is applied again (second OnGetDamageAbsorption call), reducing it to 2x200 damage, which requires a 2x multiplication to bring it to the expected 800.
Currently the absorption multiplier is 1, so we don't need a multiplier on the damage to balance it out.

BlackYps (approved):

PR #6446, opened 2024-09-13
Fix the "assist to upgrade" and "assist to unpause" settings conflicting and not pausing upgraded units during lag

lL1l1:
## Issue
Reported on [Discord](https://discord.com/channels/197033481883222026/408556023180296193/1283069284548870310). 

With sufficient lag, the unpause thread checks the upgrading unit's focus before "assist to upgrade"'s upgrade order gets through. This means the upgrading unit's focus is `nil`, same as the engineer's, so an unpause order is sent at the moment. This unpause order is processed after the upgrade and pause orders of "assist to upgrade", causing the unit to upgrade -> pause -> unpause.

Replicable easily by using `net_Lag 3000` in the console. Make sure to remove the guards in `userInit.lua` not allowing that command to work. Just assist a mass extractor with any engineer and that much lag. It will upgrade, pause, and unpause.

- Check if the target focus is `nil` before comparing it to the engineer's. This fixes the unpausing.
Other style/performance changes:
- Don't create threads for units that don't have anything to unpause.
- immediately kill the unpausing thread when it finishes its work.
- Move the thread out of the function.

Assisting mexes with both options enabled and `net_Lag 3000` no longer unpauses them.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6446.md
+++ b/changelog/snippets/fix.6446.md
@@ -1,0 +1,1 @@
+- (#6446) Fix the "assist to upgrade" and "assist to unpause" game options conflicting and causing upgrading units to unpause during lag spikes.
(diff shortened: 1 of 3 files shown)


Garanas (commented):
I've not run the changes in-game but I personally did not find the cause of this bug. And I'm happy that you did! Absolutely perfect 😃 !

BlackYps (approved):
Tested, works!

Issue #198, opened 2014-12-16
Sinking wrecks block torpedos / weapons


Crotalus:
3637 (and maybe even before that)
When a battleship sinks and something is firing at a unit behind the sinking wreck, the torpedos hit an invisible ship on the surface.

This is because the hitbox is still on the surface while the mesh is animated downwards.

Solution could be to disable collision when the ship is sinking, on the other hand it would be cool if the unit's real body/hitbox followed the animation.

PR #6448, opened 2024-09-14
Fix air units colliding with projectiles after death

lL1l1:
## Issue
Air units do not instantly die (get `Unit:Destroy()` called); they wait until they impact the ground. Since the unit is not destroyed, it still has collision while falling through the air. Units undergoing a death animation should not have collision, this is already the case for land units like the Megalith and Monkeylord.

Adds `self.DisallowCollisions = true` in the `OnKilled` override function used by air units.

Spawn a UEF SAM and an enemy Cybran spy plane in front of the SAM. The SAM shouldn't waste all 6 shots (1200 damage) on the spy plane (700 hp), 2 shots should pass through. 
Make sure that the SAM doesn't kill the spy plane before all 6 shots are fired, or else it will cancel the volley early and you won't see 2 remaining shots.

## Balance implications
This increases the effectiveness of SAMs against groups of aircraft flying towards them, since previously an aircraft getting killed in the front could block a significant number of missiles from hitting the aircraft behind it.

A compromise to solve the bug while maintaining current balance could be to imitate damaging the wreckage of the overkilled aircraft if it gets hit while falling.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6448.md
+++ b/changelog/snippets/balance.6448.md
@@ -1,0 +1,1 @@
+- (#6448) Fix projectiles colliding with dead air units. From a balance perspective, this increases the effectiveness of SAMs against groups of aircraft flying directly at them. 
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1496,7 +1496,7 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
                 self:PlayUnitSound('Killed')
             end
 
-            -- apply death animation on half built units (do not apply for ML and mega)
+            -- apply death animation on half built units (do not apply for half-built units that may animate underground like Monkeylord and Megalith)
             local FractionThreshold = bp.General.FractionThreshold or 0.5
             if self.PlayDeathAnimation and self:GetFractionComplete() > FractionThreshold then
                 self:ForkThread(self.PlayAnimationThread, 'AnimationDeath')
--- a/lua/sim/units/AirUnit.lua
+++ b/lua/sim/units/AirUnit.lua
@@ -192,6 +192,9 @@ AirUnit = ClassUnit(MobileUnit) {
                 self:DoUnitCallbacks('OnKilled')
                 self:DisableShield()
 
+                -- The unit falling is a death animation, so it needs collision turned off
+                self.DisallowCollisions = true
+
                 -- Store our death weapon's damage on the unit so it can be edited remotely by the shield bouncer projectile
                 local bp = self.Blueprint
                 local i = 1
--- a/projectiles/ShieldCollider/ShieldCollider_script.lua
+++ b/projectiles/ShieldCollider/ShieldCollider_script.lua
@@ -51,7 +51,7 @@ ShieldCollider = ClassProjectile(Projectile) {
     ---@param other any
     OnCollisionCheck = function(self, other)
         -- We intercept this just incase the projectile collides with something it shouldn't
-        WARN('Shield collision projectile checking collision! Fix me!')
+        WARN('Shield collision projectile checking collision! Fix me!', debug.traceback())
         if IsUnit(other) then WARN('It was a unit!') return end
 
         Projectile.OnCollisionCheck(self, other)


Garanas (commented):
I understand where the changes are coming from but I do not think this is the right direction. It would be better (and the same would apply to the examples you used, such as the Monkey Lord and/or structures with death animations) if the units in question would keep soaking up the damage. Once they reach the threshold that no wreck would remain, they just instantly combust and are destroyed.

It is a simulated game after all, but I understand that this solution is subjective 😃 .

BlackYps:
> Units undergoing a death animation should not have collision

Says who? I haven't heard that this is a consensus at all.

> This is already the case for land units like the Megalith and Monkeylord

Maybe this is a bug then? I don't see why this should be the case.

I generally agree with Jip that units in their death animation should still soak up damage. I'm not sure if the instant combustion should be a thing, this will look weird. Also note that I haven't thought this through from a balance perspective, so I don't know if there are severe consequences if dying experimentals still soak up bullets. The enemy units still switch targets, so I don't think it is too much of an issue. If the idea is to buff SAMs, we should find a different solution though.

lL1l1:
> > Units undergoing a death animation should not have collision
> 
> Says who? I haven't heard that this is a consensus at all.

It fits the design of the game:

From a gameplay perspective, it makes sense that if the unit has lost all of its HP it should no longer block projectiles and soak up more damage, especially since the player can't see the wreckage HP, see strategic icons, or clear hitbox size.
Other Total Annihilation inspired games (and TA itself) have wreckage get in the way of pathfinding and projectiles which is nice aesthetically but, without even going into the game design of the mechanics, it is not compatible with Supcom's large scale, zoomed out gameplay, or pathfinding issues.

I also could not find any Categories, Prop/Wreckage functions (OnCollisionCheck does nothing), or weapon/projectile/prop properties to enable collision between projectiles and props, so there is no elegant way to fully realize the aesthetic angle anyway because units will magically lose collision upon death, or we make new exceptions with only specific units/props having collidable wrecks by linking separate entities that transfer damage to the wreckage prop.

> > This is already the case for land units like the Megalith and Monkeylord
> 
> Maybe this is a bug then? I don't see why this should be the case.

It's not a bug, collision removal on death was added in https://github.com/FAForever/fa/issues/198 for navy because their hitboxes float on the surface on death. To be fair, that does mean I was wrong that it was added to ML/Mega for balance reasons specifically.

BlackYps:
> From a gameplay perspective, it makes sense that if the unit has lost all of its HP it should no longer block projectiles and soak up more damage

Not necessarily, but let's just agree to disagree here. This seems subjective and I don't think we will reach a consensus on this.

> I also could not find any Categories, Prop/Wreckage functions (OnCollisionCheck does nothing), or weapon/projectile/prop properties to enable collision between projectiles and props, so there is no elegant way to fully realize the aesthetic angle anyway because units will magically lose collision upon death, or we make new exceptions with only specific units/props having collidable wrecks by linking separate entities that transfer damage to the wreckage prop.

Why are collisions with props relevant? Props don't have death animations, at least I can't think of any.

lL1l1:
> Why are collisions with props relevant? Props don't have death animations, at least I can't think of any.

If units with death animations collide and take damage on their wreck it would be expected that props (wreckages) should have collision as well since the game rule would basically be "everything dead can tank damage as well". Making an exception that only units in a death animation can tank damage would be awkward.

BlackYps:
Okay, so you say wreckages blocking shots is impossible? I figure wrecks getting damaged happens only when projectiles hit the ground and the aoe covers the wreck? So the hitbox is still there for damage checks, but projectiles don't trigger on it, is that correct? That would be a bit unfortunate because as you said we can't have an aesthetically consistent solution.

Still that doesn't necessarily mean that units in death animation should not tank projectiles. The life cycle is this:
alive unit -> death animation -> wreckage.
Alive units need to tank projectiles. If wreckages can't ever tank projectiles there will always be an inconsistency and we can decide if units in death animation should tank projectiles or not. I think it is visually more consistent to have this change when the model switched from normal to wreckage. The player can easily see this change. The switch from alive to death animation is much more subtle. When doing this we don't create "an exception that only units in a death animation can tank damage", because the alive unit was tanking damage all the time already.

lL1l1:
> So the hitbox is still there for damage checks, but projectiles don't trigger on it, is that correct?

Yes.

> The switch from alive to death animation is much more subtle.

There are death explosions/sounds, units stop firing, and their strategic icon and health bar disappear, so it's not subtle at all.

> If wreckages can't ever tank projectiles there will always be an inconsistency and we can decide if units in death animation should tank projectiles or not. I think it is visually more consistent to have this change when the model switched from normal to wreckage.

Since the change from alive unit -> death animation is visually distinct, I do not think it is very visually inconsistent that a dying unit does not tank damage, since there is a clear visual cue for when the gameplay side of things happens.

> When doing this we don't create "an exception that only units in a death animation can tank damage", because the alive unit was tanking damage all the time already.

Currently the rule is that shield and unit HP enable collision boxes to collide with projectiles. By enabling death animations to damage the wreck HP, it follows that all collision boxes with wreck HP should collide with projectiles. That is not possible, so there would have to be an exception that death animations are the only collision box with wreck HP that collides with projectiles.

Garanas:
@BlackYps I think this is a balance team question, and not so much a game team question. It may be interesting to bring it up with the balance team. It is a very subjective topic and usually that indicates it is a balance team question.

Either direction, it should be consistent. Albeit I made my own opinion clear that I prefer one direction over the other.

@lL1l1 I did manage to get projectiles colliding with props working a year ago or two. But I do not think that I have the prototype anymore.

It is even possible to have props block the pathing of hover and naval units, but not of submerged units. This is in particular interesting for the ice-themed maps where the ice props can act as a natural boundary for frigates, but not for submarines 😄 . Back then the demo was that the ice would stop projectiles of torpedo bombers, introducing a new aspect to the simulation that is specific to ice-themed maps.

BlackYps:
Yeah, I was about to write exactly that. It seems to be a thing for the balance team to decide.

PR #6449, opened 2024-09-17
Verify death animations in blueprint post-processing 

lL1l1:
## Issue
Some mods (ex: "T4 Energy Generators" by Atom7k) add units that have bad death animation paths which causes their units to never die and be cleaned up.
In my opinion since this causes a permanent problem if the units die, it should be verified by us.

- Use `DiskGetFileInfo` in unit blueprint postprocessing to make sure the death animation exists, and remove it from the blueprint if the animation doesn't exist. Blueprint postprocessing is the cheapest way to do it performance-wise.
- Clarify the annotations that do mention an animation file path that it must be the complete file path and make a distinction between base game paths and mod paths.
- Fix ilshavoh transport drop animation.
- Some other annotation updates I picked up on the way.

Break the `uab1301`'s death animation file path in the bp and then reload the game to see the warning and ctrl-k the unit to make sure it does die instead of getting stuck forever.

## Checklist


clyfordv (changes requested):
General death animation fixes are good, but the ilshie drop animations got a little messed up; they pause above the ground for a little bit, do a little moonwalk, then drop the final distance. Can make a video if you can't reproduce.

lL1l1:
Yeah it looks a bit janky, and both animations have their downsides but I'll revert the fix since the animation does look a bit awkward.

https://github.com/user-attachments/assets/37e59b54-04b6-45ea-a549-df13e1a1dba8


https://github.com/user-attachments/assets/92edc643-1a75-462d-986f-37338c402afe

Basilisk3 (approved):
Tested with the `Advanced ShieldsFixed` mod, which contains a shield generator (SYSW5) with a broken death animation that would never expire. The changes fixed this issue by adding a fallback.

The argument against this change would be the introduction of more complexity. I believe this PR is overall positive though and should be fine to include in the patch.

PR #6450, opened 2024-09-17
Improve beam blueprint annotations

lL1l1:

Greatly improve the detail of beam blueprint annotations so that even a beginner could understand every parameter.

In 13936435abbda7030bb59512c0566cf110b11aef, tested the various properties to see that they work as described. For testing beam length, the soul ripper's exhaust flames are convenient beams: `missile_exhaust_fire_beam_03_emit.bp`.

## Checklist


Garanas (approved):
Better annotations are always great.

Garanas:
@Fichom may also have some interesting insights from his effect adventure from a while back.

lL1l1:
He was working on the wiki page at the time, and the testing setup in this PR would be useful to create some visual aids when the beam emitter page gets made.

PR #6456, opened 2024-09-21
Move `RollOfUnit()` to after the "finished building" animation

### LINKED ISSUE #5625: 3772: Engineers built from Cybran and UEF air factories do not automatically rotate towards the optimal rollout point

lL1l1:

Fixes #5625. UEF and Cybran air factories have an animation that locks the unit's orientation, interfering with the rolloff rotation. The changes move the rolloff rotation to happen after the animation finishes.

Spawn factories, give a rally point on one side, order an engineer, and then change the rally point to the other side before the engineer finishes building and see that it now gets rotated correctly in the UEF and Cybran factories.
```
   CreateUnitAtMouse('xab1401', 0,    0.14,   -8.65, -0.00000)
   CreateUnitAtMouse('xsb0302', 0,   -4.04,    2.16,  0.00000)
   CreateUnitAtMouse('ueb0302', 0,   11.96,    2.16,  0.00000)
   CreateUnitAtMouse('urb0302', 0,    3.96,    2.16,  0.00000)
   CreateUnitAtMouse('uab0302', 0,  -12.04,    2.16,  0.00000)
```

## Checklist

### DIFF
--- a/changelog/snippets/fix.6456.md
+++ b/changelog/snippets/fix.6456.md
@@ -1,0 +1,1 @@
+- (#6456) Fix engineers not rotating towards the optimal rolloff point in factories with "build finished" animations such as the UEF and Cybran air factories.
--- a/lua/sim/units/FactoryUnit.lua
+++ b/lua/sim/units/FactoryUnit.lua
@@ -139,7 +139,21 @@ FactoryUnit = ClassUnit(StructureUnit) {
 
         if not (self.FactoryBuildFailed or IsDestroyed(self)) then
             self:StopBuildFx()
-            self:ForkThread(self.FinishBuildThread, unitBeingBuilt, order)
+
+            -- Moving off factory has to be issued this tick so that rally points are issued after it
+            -- Air units don't need the move order since they fly off the factory by themselves
+            -- Pass the spin up so that engineers can be rotated towards the rolloff point after the "build finished" animation
+
+            local spin = nil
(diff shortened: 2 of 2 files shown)


clyfordv (changes requested):
This introduces a bug where the a land unit built from a relevant air fac wants to nav back to the roll off point after the rally point. See something similar here where an order is slipped in right at the moment of construction with odd behavior: https://discord.com/channels/197033481883222026/1144605729911738439/1238254471662927943 

Stock:
image

With change:
image

My read on why this occurs is that the rally point order chain issuing is extra special somewhere in the engine (apropos nothing it would be nice to expose that to lua a little more, to enable more sophisticated roll off behavior). I think if you want to implement this as-is you'll have to manually re-check the unit's orientation and update if necessary.

lL1l1:
This fixes it but the code style may be sloppy, although I think it's the most efficient solution.

Basilisk3:
Tested and works as expected, in addition to being an important change for gameplay.

clyfordv (approved):
Tested, good to go

BlackYps:
We'll include this one as well because we now have more time for testing with the delayed release

PR #6458, opened 2024-09-23
Annotate the damage data passed from weapons to projectiles

lL1l1:

This is the table passed from weapons to projectiles so that the projectiles know their damage-dealing behavior

Started from `GetDamageTableInternal` and manually traced wherever the data was passed and used.
Also checked that my explanation of the metatable was correct by checking the projectile's vs weapon's `debug.allocatedsize()` of the damage table (40 vs 360 bytes) on the Aeon ACU's main gun.

## Additional context

I noticed that `CollideFriendly`'s functionality is broken. There is also an unused `CollideFriendlyShield` field that could be cleaned up.

## Checklist

### DIFF
--- a/changelog/snippets/sections/other.6458.md
+++ b/changelog/snippets/sections/other.6458.md
@@ -1,0 +1,1 @@
+- (#6458) Annotate the damage data table passed from weapons to projectiles.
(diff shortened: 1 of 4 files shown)


clyfordv (approved):
Annotations are good, no warnings thrown.

PR #6460, opened 2024-09-24
Add a workflow to automatically convert pull requests to draft

BlackYps:
First of possibly many workflows to help us automate project management

### DIFF
--- a/.github/workflows/auto-draft-pr.yaml
+++ b/.github/workflows/auto-draft-pr.yaml
@@ -2,7 +2,7 @@ name: Convert PRs to Draft on Opening
 
 on:
   pull_request:
-    types: [opened]
+    types: [synchronize]
 
 jobs:
   convert_to_draft:
--- a/.github/workflows/auto-draft-pr.yaml
+++ b/.github/workflows/auto-draft-pr.yaml
@@ -1,0 +1,18 @@
+name: Convert PRs to Draft on Opening
+
+on:
+  pull_request:
+    types: [opened]
+
+jobs:
+  convert_to_draft:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install GitHub CLI
+        run: sudo apt-get install gh -y
+
+      - name: Authenticate GitHub CLI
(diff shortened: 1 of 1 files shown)


Garanas:
I am missing the bigger picture here. What does this solve exactly?

BlackYps:
It's solving the problem that I want PRs to be drafts when they are not ready for review because the person that opened it still wants to implement additional things. People will forget to do this, because in the repo settings we can not set PRs to be drafts by default. So I added this action to do it for us.

It's not the most important thing, but it was easier than the manipulations of the project status that I have in mind to automate the milestone overview. I picked this to get familiar with automatic workflows that use github cli and are triggered by events. I should probably describe that in detail in an issue, so someone else can pick the further automations up.

Garanas:
I see, but the downside is that not all pull requests are necessarily unfinished. But there are pros/cons to everything, haha.

.github/workflows/auto-draft-pr.yaml
```diff
@@ -0,0 +1,14 @@
+name: Convert PRs to Draft on Opening
```
Garanas, line 1: Since this repository has no global license I would add the standard MIT license at the top of the file.

BlackYps:
Luckily it's just a click of a button when the pull request is ready to be reviewed

.github/workflows/auto-draft-pr.yaml
BlackYps: Seems a bit unelegant to specify this on the file level, but until we have a better solution I will do this.

Garanas (approved):

Garanas:
I agree, I can see the value. At a minimum it will help to identify what contributions are ready to be reviewed.

Garanas:
@BlackYps I think this approach does not work when a pull request originates from a fork.

image

See also an error message:

```
Run gh pr ready --undo 6518
API call failed: GraphQL: Resource not accessible by integration (convertPullRequestToDraft)
Error: Process completed with exit code 1.
```

Shall I disable the workflow for now?

PR #6464, opened 2024-09-27
Make shield assist start repairing the shield instantly

lL1l1:
## Issue
Compared to other construction/unit repair projects, shields can take a very long time to repair when assisted, varying between 7-11 ticks from the shield's `OnDamage` until `OnStartBuild` is called. Repairing takes 1 tick to start.
The behavior was brought up to attention in a Discord [balance suggestion](https://discord.com/channels/197033481883222026/1286761127870464010/1286864739766505555).

- When a shield is damaged, it issues a repair and guard order to units guarding it. This causes them to start repairing in 1 tick instead of 7-11 ticks.
  - This can happen once per tick.
  - Guards must have an empty command queue aside from the guard order to be affected.

Spawn a shield/paragon/sniper bot/engineer presets and log the time difference using the debug setup commit.
```
   CreateUnitAtMouse('xab1401', 0,   -9.50,    8.32, -0.00000)
   CreateUnitAtMouse('urb4202', 0,    2.50,   -1.68,  0.00000)
   CreateUnitAtMouse('xsl0305', 1,   17.50,  -15.68, -0.00000)
   CreateUnitAtMouse('url0301_engineer', 0,   -2.16,    4.66, -0.00000)
   CreateUnitAtMouse('url0301_engineer', 0,   -3.52,    3.30, -0.00000)
   CreateUnitAtMouse('url0301_engineer', 0,   -4.83,    1.07,  0.00000)
```

## Balance implications
This makes shield assistance significantly more powerful against fast bursts of high damage such as Aeon T3 Arty (6k damage --1 second-> 6k damage), synchronized T3 artillery (especially Aeon artillery), or the Mavor (16k every 4.8 seconds). Outside of defending against strategic artillery, I don't think there are similar situations where shield assist is commonly used.

Buffing shield assistance in this way is fair because it already drains a massive amount of resources to let the player focus on other tasks than microing shields against artillery (which I reckon is not the gameplay that most are interested in). Keeping a consistently achievable shield assistance outcome, unaffected by random assist reaction time vs burst damage, would fulfill this goal of letting the player do other tasks.

It could make shield assist too powerful in which case I would reduce the efficiency of shield repair by adjusting the `RegenAssistMult` (determines much regen is given for an amount of buildpower) or changing the repair cost of shields directly.

## Checklist


clyfordv (approved):
Tested, changes are good, assisters start repairing immediately instead of waiting.

PR #6465, opened 2024-09-28
Improve the visual representation of multiple projectiles by increasing their `StrategicIconSize`

Basilisk3:
**Usha-Ah: Seraphim T3 Sniper Bot (secondary firing mode):**
The current `StrategicIconSize` of this projectile is set to `1`. This is too small, because the projectile is very fast and thereby easy to miss. A `StrategicIconSize` of  `1` also does not represent the actual strength of the projectile well (2000 damage).

I changed the `StrategicIconSize` to `4` in this PR. Alternative and possibly preferable options could be:
- Only increase it to `3` since `4` is only used by TML missiles.
- Decrease the `StrategicIconSize` of the first firing mode to `2` and and only increase the size of the second firing mode to `3`.

**Othuum: T3 Siege Tank (heavy cannon); Yenzyne: T2 Hover Tank:**
StrategicIconSize: 1 --> 2

**Percival: T3 Armored Assault Bot:**
StrategicIconSize: 2 --> 3

## Checklist


lL1l1:
Comparison from left to right of percivals, bricks (both size 2), normal sniper shots (size 3), and sniper sniper shots (size 4).

{165CC18A-52DB-411F-843A-F502A43C6C75}

I think for how high velocity and high damage the projectiles are, it's fine to use the unique size 4 icon. You could argue that normal shots should be size 2 and then size 3 for sniper shots, but I think the argument against that is that the high velocity and long range needs a larger size to be noticed in contrast to the slow/short range percy/brick projectiles.

By the way, Othuums have size 1 projectiles for both weapons. The Thau Cannon could have size 2. Also the Thau cannon is listed as indirect fire in the UI, although indirect fire range rings + direct fire UI text is possible.

lL1l1:
In particular for this type of view the size is even more important:
{37722837-9A88-44C8-B35F-FEB0E6C827FA}

lL1l1 (approved):

Basilisk3:
How about giving the Percival's projectile a `StrategicIconSize` of 3?

> Also the Thau cannon is listed as indirect fire in the UI, although indirect fire range rings + direct fire UI text is possible.

Changed in #6424. I think it makes more sense for it be labed as `Direct Fire`.

PR #6466, opened 2024-09-29
Fix Ythotha being able to move backwards

lL1l1:
## Issue

Ythotha can walk backwards (with 2.5 speed at that) which is obviously overpowered and also looks very weird.

https://github.com/user-attachments/assets/0f12a23d-e2c9-4ee6-8391-00605244a5e6

Set its reverse speed to 0 so it cannot walk backwards.

Spawn it and confirm it cannot walk backwards when given move orders near close behind it.
```
   CreateUnitAtMouse('xsl0401', 0,    0.00,    0.00,  0)
```

## Additional context

`BackUpDistance` gets a default value from the engine, but I'm not sure how it is calculated, so I didn't add a comment in the annotation about it.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6466.md
+++ b/changelog/snippets/balance.6466.md
@@ -1,0 +1,2 @@
+- (#6466) Fix Ythotha being able to walk backwards.
+  - `MaxSpeedReverse`: 2.5 -> 0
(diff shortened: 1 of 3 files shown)


Basilisk3 (approved):
Tested and works. This is an important change, as the bug makes the Ythotha too powerful. Setting the `BackUpDistance` to `0` also fixed the bug for me, but I am not sure if it is the preferred solution.

lL1l1:
`BackUpDistance` being set to 0 like with monkeylord still allows walking backwards because it will try to do a backwards 3 point turn (move backwards while turning to 90 degrees, then move forwards while turning to 180 degrees). Abusing that maneuver behavior, you can give move orders on opposite sides behind the monkeylord to make it walk backwards, it just needs constant micro.

PR #6467, opened 2024-09-30
Fix a crash if a unit has an enhancements table but not an intel table

The-Balthazar:
Fixed a crash if a unit has an enhancements table but not an intel table.

### DIFF
--- a/changelog/snippets/fix.6467.md
+++ b/changelog/snippets/fix.6467.md
@@ -1,0 +1,1 @@
+- (#6467) Fix a crash during blueprint loading if a unit blueprint has an enhancements table but not an intel table.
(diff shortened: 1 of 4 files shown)


Hdt80bro:
Incoming PR's need to be pointed at `develop` now!

Garanas:
And originate from `develop` 😃 !

lL1l1:
Reported on [Discord](https://discord.com/channels/197033481883222026/832710161847287819/1290246295998562386)

Was caused by #4576 I think, since it was bit unexpected for a unit to have no intel at all.

lL1l1 (approved):

PR #6468, opened 2024-10-01
Increased Jester Agility

RabidPope:
Increased acceleration, making the unit more responsive to direction changes.
Also increased circle strafing radius to stay out of t1 mex vision more reliably.

Tested to see that it will still stay in range of a fast moving ground target.


clyfordv (approved):
Tested, changes are good, unsure why the draft conversion check failed

BlackYps:
Still needs to discussed with the balance team

Rhaelya:
The balance team has found the changes not that interesting beyond the idea of a small T1 gunship being more agile than other gunships.

PR #6470, opened 2024-10-03
Implement basic gesture detection to delete commands

Garanas:

Adds a basic gesture detection to allow you to delete commands while in command mode. This is in particular useful when you do not use hotkeys. Often I find myself in the situation that I miss placed a build order. When you do not use hotkeys, this looks like:

- (0) Miss click a build order.
- (1) Right click to cancel command mode.
- (2) Right click to remove the build order.
- (3) Move mouse to the correct build icon and left-click it.

In particular, step 3 can be time consuming. With this feature it can be simplified to:

- (0) Miss click a build order.
- (1) Rage-move the mouse across the build order until it disappears.

And problem solved. See it also in action:

https://github.com/user-attachments/assets/687aceb0-7522-4172-bbbe-106da39a7c5d

All is tested during development. All the code is annotated. Debugging is in particular trivial since it requires no reloading of the game - see also the nice trick with using `OnDirty` and `OnReload` of the module 😃 ! 

## Additional context

Requires https://github.com/FAForever/FA-Binary-Patches/pull/90, which is available on the FAF Develop game type. To test this pull request you'll need to start a lobby using the FAF Develop game type to retrieve the correct executable.

With thanks to @4z0t for his time and effort.

## Checklist

### DIFF
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -7,6 +7,7 @@
 ---@field targetId? EntityId
 ---@field blueprintId? UnitId
 ---@field commandType number
+---@field commandId number
 
 ---@alias SubmergeStatus
 ---| -1  # submerged
@@ -480,7 +481,7 @@ end
 
 ---
 ---@param key string
----@return string[]
+---@return string
 function GetOptions(key)
 end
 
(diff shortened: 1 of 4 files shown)


Garanas:
I've turned this into a UI mod, see also:

- https://github.com/FAForever/fa-gesture-delete-command

vjt4:
Sorry to necro this, but what is the hot key called to undo the most recent green highlight building thing (blueprint)

PR #6476, opened 2024-10-12
Readjust the Cybran ACU's Nanite Torpedo upgrade

Basilisk3:
When the Nano-Repair upgrade was introduced to the Cybran ACU, the `MuzzleSalvoSize` of its torpedo upgrade was nerfed to prevent the combination from being too oppressive. Unfortunately, this change resulted in the torpedo upgrade being weak even against lower quantities of torpedo defenses since a single anti-torpedo projectile can negate 500 damage. To alleviate this issue, the torpedo upgrade now launches `3` instead of `2` torpedoes, but at a slightly lower DPS. Additionally, as with most other Cybran torpedo launchers, the damage is now dealt in multiple pulses.

**Cybran Armored Command Unit (URL0001):**
  - Nanite Torpedo Launcher
    - Damage: 500 --> 60 (DoTPulses: 5)
    - MuzzleSalvoSize: 2 --> 3
    - DPS: 250 --> 225

## Checklist


lL1l1:
After #6457, the DoT time should be 0.4 to match other Cybran torpedoes.

Basilisk3:
Approved by the Balance Team.

PR #6477, opened 2024-10-14
Add `SimUtils` `DrawBone` function

lL1l1:

- 314f7ed4cfcc86b5432308a8e72a491aeaa76c48 Corrects the annotation for `Entity:GetBoneDirection`, as it returns the direction vector and not the roll pitch and yaw of the bone.
- 3a99a93de9da39c653f95983ff098f5c15444ed0 Adds a `DrawBone` function to SimUtils which is practically identical to how the engine displays bones with `ren_showskeletons`.
  - This change depends on https://github.com/FAForever/fa/pull/6438.

Used the DrawBone function in a thread for a unit and made sure it is identical to the engine's version when the unit is moving around.

## Checklist


lL1l1:
Not sure if `VectorCross(v1, v2)` or `v1 * v2` is more efficient.

Garanas (approved):
Function works as advertisement. Out of curiosity, you are aware of this keybinding:

image

That achieves the _exact_ same result, but for all bones:

image

lL1l1:
> Adds a DrawBone function to SimUtils which is practically identical to how the engine displays bones with ren_showskeletons.

Yes I'm aware. I made draw bone for drawing bones that aren't shown by the skeletons, specifically the weird extra bones on collision beams. It was also an exercise with `GetBoneDirection`.

PR #6481, opened 2024-10-19
Balance cost and DPS stats of T3 arty

lL1l1:
## Issue
[Related Discord thread.](https://discord.com/channels/197033481883222026/1272828981367803967) Currently it's well known that Aeon arty is by far the best T3 arty. It has double the alpha damage of other arties (can one shot shields with only two arties, one shots targets if it hits), has better dps per unit and dps per mass than other arties (compared to Cybran: 19%/mass 24%/unit), and has very good accuracy that boosts the damage output compared to others even further.
To a lesser known extent, UEF arty is 10% better DPS per mass than Cybran arty, and 8% compared to Sera, so in general T3 arties could use some balance altogether. Balancing the arties altogether is a good first step before seeing if Aeon arty needs to be addressed.

### Balance of arty DPS by cost:
1. Aeon arty set to 79k mass in proportion to old UEF arty DPS/M.
2. Arty costs spread evenly across 70-79k mass. Arty costs are not reduced below their current costs because that would make them come out earlier.
3. Damage adjusted for more per-artillery variety in shots to kill T3 shields, while also evening out STK vs different faction T3 shields.
4. Reload adjusted so that avg DPS vs T3 shields (this accounts for regen) per mass is equal to Aeon arty.
5. Energy and buildtime adjusted for the new mass costs with some rounding, but practically the same per-mass ratios.

image

## Testing done on the proposed changes

Shooting at a test target, no artillery seems excessively powerful, due to the inaccuracy massively reducing the DPS of artillery with high splash radius.
<details> <summary> Spawn target command: </summary>

UEF structure/Seraphim shield air grid test target spawned at the edge of the arty range so that the top shield edge is roughly at max range of the artillery:
```
CreateUnitAtMouse('xsb4301', 1,    3.63,   11.13, -0.00017)
CreateUnitAtMouse('ueb0302', 1,  -27.38,  -11.88, -0.00009)
CreateUnitAtMouse('ueb1301', 1,   12.63,  -11.88,  0.00010)
CreateUnitAtMouse('ueb1301', 1,  -11.38,   12.13,  0.00000)
CreateUnitAtMouse('ueb1301', 1,  -19.38,    4.13,  0.00000)
CreateUnitAtMouse('zeb9602', 1,  -11.38,   20.13,  0.00018)
CreateUnitAtMouse('zeb9602', 1,  -11.38,    4.13, -0.00024)
CreateUnitAtMouse('zeb9602', 1,  -19.38,   12.13,  0.00012)
[...]

PR #6482, opened 2024-10-19
Balance accuracy stats of T3 Arty

lL1l1:
## Issue
[Related Discord thread.](https://discord.com/channels/197033481883222026/1272828981367803967) On top of the DPS and cost issues mentioned in #6481, there are also significant issues in the effectiveness of artillery due to inaccuracy vs splash damage. T3 Artillery often comes out at the stage of the game where there are only a few highly defended key targets to attack, such as other artillery, game enders, ACUs, nukes, and SMD. This means that high splash, low accuracy, fast reload artillery is generally much less favored than the opposite which is better at breaking through shields in one point and destroying a small target. On the other hand, high splash radius with a high enough accuracy begins multiplies its damage by hitting multiple shields at the same time, which is also unfortunately countered by microing shields so only one shield is active at a time, in which case DPS becomes the most important stat (balanced in the other PR, although it assumes perfect accuracy, it is close enough).

Builds on top of the changes in #6481.

### Balance of arty accuracy:
All faction arties are spawned with full adjacency and in the same exact spot. They shoot at the exact same invincible shields which measure how much direct DPS (no overspill) the arties are dealing to the shields overall. This results in a % DPS multiplier for the arty since it can hit multiple shields. The firing randomness is changed iteratively until the % DPS multipliers match for all arties, which in theory means that they perform equally well against multiple shields despite their accuracy and damage radius differences.

See the `testprocedures.txt` file in the commits.

## Checklist


Tagada14:
The mass cost of Emissary is reversed in the balance change snippet

lL1l1:
I don't really have a solid test plan for testing artillery randomness, but this is better than nothing so I would try it out in games.

I guess a proper way is to create some test unit that generates a target base template and then records data about DPS/time so that you could run the sim at +15 for a bit to get the right numbers, but it seems excessive.

Rhaelya (approved):
Everything looks in order

PR #6485, opened 2024-10-20
Shader rework

BlackYps:
Depends on #6452, so the diff will be misleading until that one is merged.

Now every shader is capable of using the new, better way of calculating the water absorption. The only requirement is that the light multiplier is bigger than 2.1. This is good, because the mesh shader doesn't know what terrain shader is in use and has worked like this before. This could lead to unfitting results, if the light multiplier was big enough, but a terrain shader was used that only supported the legacy water calculation. This is now fixed.

I also made the decals able to use pbr light calculations. Now they behave consistently like the ground they are on. (Also the map editor does this automatically and we can't prevent it, so it helps the editor produce results that are consistent with the game.)
We could also do this for splats, I haven't decided on that yet.

One thing about the decals is that they theoretically have a texture to define the specularity (or roughness), but this texture was black for all decals that I tested. I don't know if there are any decals that have this defined.

## Checklist

### DIFF
--- a/changelog/snippets/other.6485.md
+++ b/changelog/snippets/other.6485.md
@@ -1,0 +1,2 @@
+- (#6485) The new, better way of calculating the water absorption is now available for all terrain shaders. The only requirement is that the light multiplier is set to more than 2.1. Decals now use PBR light calculations if the terrain shader uses it, making them more consistent with the ground they are on.
+- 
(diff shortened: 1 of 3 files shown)


Garanas (approved):
This looks fine to me. The additional documentation is also highly appreciated.

PR #6486, opened 2024-10-21
Adjust aurora handling

lL1l1:

- Boost aurora turn rate significantly so that the force it uses for movement can change direction instantly instead of forcing a turning maneuver.
- Remove reverse speed so that it does not try to do half of the turn in reverse, as that gets it stuck when spamming move orders.

This makes it no longer get stuck for a short time doing a turning circle maneuver every time a move order backwards is given.
It does become harder to micro dodges since you can't dodge with the turning maneuver by occasionally clicking backwards/forward and instead have to click to the sides of the formation, which means that the aurora formation rotates sideways which puts your aurora out of range and in general messes with the overall formation. This more applies at larger counts though. At smaller amounts you don't have to deal with large formations so the desired sideways movement is easier to accomplish without messing up the formation badly, while also responding to commands as expected.

Use `NoDamage` in console, and spawn some auroras and a target to shoot at then move them around comparing the before and after of the changes, with various amounts of auroras and with spam clicking vs single clicking. Use `ren_steering` to show the steering (green means reverse).

Before:

https://github.com/user-attachments/assets/35c86b4b-24c7-48e0-acf0-790dc1502a9c

After:

https://github.com/user-attachments/assets/d3899436-015c-4975-b1e8-cc87128ddbe6

## Additional context

Last time aurora movement was discussed was here: https://forums.faforever.com/viewtopic.php?f=52&t=3205

## Checklist


Garanas:
That is an interesting alternative solution, very creative 😃 . I'll need to toy around with it to see the impact.

lL1l1:
Rather basic changelog snippet.

Rhaelya (approved):
Approved for a test in FAF Beta

PR #6498, opened 2024-10-30
Enhancement Script Change (UEF-ACU)

MrRowey:
### Pull Request Summary

This Pull Request refactors the Enhancement handling logic in the script by replacing the long `if/else` chain with a modular approach where each enhancement has a dedicated function. This change aims to improve code maintainability and readability.

### Background

The current implementation stores enhancements in a lengthy `if/else` chain, making it difficult to manage and understand. After discussing four potential approaches for improvement in Discord, we concluded that the existing method (Option 1) is not viable due to the difficulty in maintaining and extending it. Here’s a summary of each option and the reasoning behind our choice:

1. **Current Implementation (if/else chain)**: Difficult to maintain and understand. Not viable.
2. **Enhancement Upgrades Table within Class with Key Lookup**: Improved maintainability but poses issues with inheritance and potential for modifications to leak through the hierarchy. This approach was ruled out due to concerns raised by hdt80: "Anything that requires several levels of nesting won't work well with inheritance and may confuse as to whether the table is part of the class structure."
3. **Enhancement Upgrades Table as a Module with Key Lookup**: While modular, this approach would require redefining the `CreateEnhancements` function with each new file, creating redundancy. This option was ruled out, as noted by Nomander: "You have to rewrite the enhancements table and `CreateEnhancements` every new file, which is inconvenient."
4. **Inline Enhancement Functions in Unit Class with Key Lookup**: Improves maintainability without complex inheritance issues. This approach aligns with standard methods for modding behaviour and allows each enhancement to be directly modified or extended.

Based on these discussions, **Option 4** was determined to be the most logical and maintainable choice.

### Implementation

With Option 4, each enhancement type (e.g., `Shield`, `ShieldRemove`) will now have its dedicated function. For instance, here’s how the `Shield` enhancement logic changes:

**Before (in `if/else` chain):**
```lua
elseif enh == 'Shield' then
    self:AddToggleCap('RULEUTC_ShieldToggle')
    self:CreateShield(bp)
    self:SetEnergyMaintenanceConsumptionOverride(bp.MaintenanceConsumptionPerSecondEnergy or 0)
elseif enh == 'ShieldRemove' then
    self:DestroyShield()
    RemoveUnitEnhancement(self, 'ShieldRemove')
    self:RemoveToggleCap('RULEUTC_ShieldToggle')
```

**After (Option 4 with dedicated functions):**
```lua
ProcessShieldEnhancement = function(self, bp)
    self:AddToggleCap('RULEUTC_ShieldToggle')
    self:CreateShield(bp)
    self:SetEnergyMaintenanceConsumptionOverride(bp.MaintenanceConsumptionPerSecondEnergy or 0)
end,

ProcessShieldRemoveEnhancement = function(self, bp)
    self:DestroyShield()
    RemoveUnitEnhancement(self, 'ShieldRemove')
    self:RemoveToggleCap('RULEUTC_ShieldToggle')
end,
```

### `CreateEnhancement` Function Redesign

The `CreateEnhancement` function now dynamically calls enhancement functions using a naming convention based on the enhancement name. For example, `Shield` enhancements will map to `ProcessShieldEnhancement`, making the code modular and error-resistant. The updated function is as follows:

```lua
CreateEnhancement = function(self, enh)
    ACUUnit.CreateEnhancement(self, enh)
    local bp = self:GetBlueprint().Enhancements[enh]
    if not bp then return end

    local ref = 'Process' .. enh .. 'Enhancement'
    local handler = self[ref]
    if handler then
        handler(self, bp)
    else
        WARN("Missing enhancement: ", enh, " for unit: ", self:GetUnitId(), " - expected function name: ", ref)
    end
end,
```

This approach offers:

- **Modularity**: Each enhancement is a self-contained function, simplifying future modifications.
- **Error Handling**: The new structure provides clear warnings if an enhancement function is missing, which improves troubleshooting.
- **Readability and Maintenance**: By removing the lengthy `if/else` chain, the code is easier to read, manage, and extend.

## Checklist
- [x] Tested that all Upgrades work in the game both Installing and removing each upgrade.

### DIFF
--- a/units/UEL0001/UEL0001_script.lua
+++ b/units/UEL0001/UEL0001_script.lua
@@ -302,8 +302,8 @@ UEL0001 = ClassUnit(ACUUnit) {
         attachee:SetDoNotTarget(false)
     end,
 
-    --==================================================================================================================================================================================
-    -- Enhancements
+    ---------------------------------------------------------------------------
+    --#region Enhancements
 
     -- Drone Upgrades
 
@@ -686,6 +686,8 @@ UEL0001 = ClassUnit(ACUUnit) {
             WARN("Missing enhancement: ", enh, " for unit: ", self:GetUnitId(), " note that the function name should be called: ", ref)
         end
     end,
+
+    --#endregion
 }
 
 TypeClass = UEL0001
--- a/units/UEL0001/UEL0001_script.lua
+++ b/units/UEL0001/UEL0001_script.lua
@@ -306,6 +306,9 @@ UEL0001 = ClassUnit(ACUUnit) {
     -- Enhancements
 
     -- Drone Upgrades
+
+    ---@param self UEL0001
+    ---@param bp Blueprint unused
     ProcessLeftPodEnhancement = function(self, bp)
         local location = self:GetPosition('AttachSpecial02')
         local pod = CreateUnitHPR('UEA0001', self.Army, location[1], location[2], location[3], 0, 0, 0)
@@ -315,6 +318,9 @@ UEL0001 = ClassUnit(ACUUnit) {
         self.HasLeftPod = true
         self.LeftPod = pod
     end,
+
+    ---@param self UEL0001
+    ---@param bp Blueprint unused
     ProcessLeftPodRemoveEnhancement = function(self, bp)
         if self.HasLeftPod == true then
             self.HasLeftPod = false
@@ -329,6 +335,9 @@ UEL0001 = ClassUnit(ACUUnit) {
         end
         KillThread(self.RebuildThread)
     end,
+
+    ---@param self UEL0001
+    ---@param bp Blueprint unused
     ProcessRightPodEnhancement = function(self, bp)
         local location = self:GetPosition('AttachSpecial01')
         local pod = CreateUnitHPR('UEA0001', self.Army, location[1], location[2], location[3], 0, 0, 0)
@@ -338,6 +347,9 @@ UEL0001 = ClassUnit(ACUUnit) {
         self.HasRightPod = true
         self.RightPod = pod
     end,
+
+    ---@param self UEL0001
+    ---@param bp Blueprint unused
     ProcessRightPodRemoveEnhancement = function(self, bp)
         if self.HasLeftPod == true then
             self.HasLeftPod = false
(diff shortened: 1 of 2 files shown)


units/UEL0001/UEL0001_script.lua
```diff
@@ -302,237 +302,388 @@ UEL0001 = ClassUnit(ACUUnit) {
...
+    ---@param self UEL0001
+    ---@param bp Blueprint unused
+    ProcessAdvancedEngineeringRemoveEnhancement = function(self, bp)
+        local bp = self:GetBlueprint().Economy.BuildRate
+        if not bp then return end
+        self:RestoreBuildRestrictions()
+        self:AddBuildRestriction(categories.UEF *
+            (categories.BUILTBYTIER2COMMANDER + categories.BUILTBYTIER3COMMANDER))
+        self:AddBuildRestriction(categories.UEF *
+            (categories.BUILTBYTIER2COMMANDER + categories.BUILTBYTIER3COMMANDER))
```
sprunk, line 482: duplicate?

MrRowey: All I've done is Coppy Past the old code within the if statement as you can see here

image

It been there all this time xD

So we're already improving maintainability as we can see like you says its a duplicate

sprunk: Maybe it's good to keep it just a Coppy Past for now then (to ensure there are no behaviour changes) and do further improvements separately.

units/UEL0001/UEL0001_script.lua
```diff
@@ -302,237 +302,388 @@ UEL0001 = ClassUnit(ACUUnit) {
...
+
+    ---@param self UEL0001
+    ---@param enh string
+    CreateEnhancement = function(self, enh)
+        ACUUnit.CreateEnhancement(self, enh)
+
+        local bp = self:GetBlueprint().Enhancements[enh]
+        if not bp then return end
+
+        local ref = 'Process' .. enh .. 'Enhancement'
```
Hdt80bro, line 681: Consider renaming methods to `ProcessEnhancement`*name* so it's not hiding in between bookends like this

MrRowey: Updated to use this better naming Schema

units/UEL0001/UEL0001_script.lua
MrRowey: Yup that would be the plan.

Garanas: I agree, it is better to do that separate.

Garanas:
Great job @MrRowey , looking forward to the other (S)ACUs 😃 .

And for small issues that @sprunk mentions, feel free to turn those into GitHub issues and assign the 'good-first-issue' label to it. They're great to introduce new contributors to the development environment.

PR #6499, opened 2024-10-30
Apply new binary patch to Salem

4z0t:
Salem destroyer has special ability to move from water onto land. However in vanilla game this feature sometimes may cause problems with positioning of the unit and it's pathfinding. FAF introduced solution with a binary patch which makes it toggable ability to walk on land. 
But this solution had flaws:
1. works only for salem: you can't make it work for other unit (mod one for example);
2. solution is unsafe: you can't restart game or replay because if any salem walks it just causes a crash;
3. solution uses weird way to do that toggle: it confuses doing this way;
4. this breaks filtering based on blueprint id.

With new assembly patch these flaws are no more. This PR is about applying new assmbly fix.

Applies patch changes and removes old ones.

1. This patch can be applied to any other unit to achive same behavior (no example provided)
2. Start skirmish match. Spawn salem and make it walk. Restart match (not closing game). Spawn salem again and make it walk. Game works with no problems.
3. Function description isn't great since it relies on internal work of the game, but still it isn't done with `GetStat`.
4. Spawn 2 salems, make one of them go to land. Select one of them, press *ctrl-z* (select similar units) or with cltr-left-click, second one is selected too.

## Additional context
This change requires an asm [patch](https://github.com/FAForever/FA-Binary-Patches/pull/94).

## Checklist


changelog/snippets/other.6499.md
```diff
@@ -0,0 +1,23 @@
+- (#6499) Apply new binary patch to Salem
+
+Salem destroyer has special ability to move from water onto land. However in vanilla game this feature sometimes may cause problems with positioning of the unit and it's pathfinding. FAF introduced solution with a binary patch which makes it toggable ability to walk on land.
```
Garanas, line 3: Please do not copy-paste the pull request description into the changelog snippet. The audience of the two are not the same - for a pull request description your audience is other programmers and/or contributors who (soon) have a decent idea of how things work. For the changelog the audience is your average joe, they won't understand all the technical details and they don't have to.

I think you can simplify this to:

```
- (#6499) Fix hard crash when exiting the game when a Salem is on land

The Cybran Tech 2 Destroyer (Salem) has the special ability to move from water onto land. There was a bug in a special, alternative implementation introduced by FAForever. Together with an assembly patch this bug is now fixed.

Interested in how assembly patches work? Reach out to the game team. You can find us easiest via the official Discord server.
```

The above would be of category 'fix'. The below would be of category 'feature':

```
- (#6499) Enable dynamic footprint changes for (modded) units

With thanks to an assembly patch the footprint changes that were previously unique to the Cybran Tech 2 Destroyer can now be applied to any unit. For more details, see the pull request on GitHub. 

Interested in how assembly patches work? Reach out to the game team. You can find us easiest via the official Discord server.
```

Garanas (approved):
Looks good in general. @BlackYps do we want to include this in the release? One less hard crash to desktop is always nice.

changelog/snippets/other.6499.md
4z0t: It is not correct to call it dynamic footprint change. Because internally game still uses both of them based on motion type (if both are provided). Here we are just forcing to use only one in particular for water.

BlackYps:
Yes, this sounds like a good addition

Garanas (approved):

PR #6504, opened 2024-11-01
Tweak the selection boxes of various air units

Basilisk3:
Unlike other air units, the selection box of most transports is almost completely flush with the ground, which results in even slight variations in terrain obstructing the selection box while the transport is grounded. This can be alleviated by moving the box slightly upwards. In terms of gameplay, there is essentially no difference. Additionally, the size of the Vulthoo's selection box is tweaked.

## Checklist


Garanas (approved):
Looks good, fixes some small but important details.

PR #6506, opened 2024-11-03
Update Seraphim Enhancement Script

MrRowey:
Following on the changes of #6498 

## Checklist
- [x] Tested all Enhancements that they apply without any errors with both install and remove


MrRowey:
@Garanas the only one I'm trying to get to work is the regen aura enhancements as I've had to split them due to them being a or in the statement chain when being called. so if you know what needed to be changes with the code that is in this let me know

lL1l1:
Fixed the regen aura. The enhancement name was just being used and it was missing, which shows the downsides of that style of modular code where it just appends strings to make variables.

I think keeping the two regen auras separate is fine. It's duplicate code but the alternative would disrupt the standard we have made a bit. The duplicate code also makes it easier to see what the buffs actually are.

### Additional Note:
- I kept `AdditionalDamage` as an undefined field of `UnitBlueprintEnhancement` (produces a warning with intellisense).
https://github.com/FAForever/fa/blob/3a2267a4986f4673d2b1a10c3aa0e548ff5a08fa/units/XSL0001/XSL0001_script.lua#L518
That should be fixed in a different PR because it involves the UEF ACU (`ZephyrDamageMod`), Cybran/Aeon SACUs (`NewDamageMod`), and Sera ACU (`AdditionalDamage`).

MrRowey:
@Garanas can this be merged in

Garanas (approved):

PR #6507, opened 2024-11-03
Fix depth charges not applying the `ProjectilesToDeflect` blueprint field

Garanas:

Fix an issue where the blueprint field `ProjectilesToDeflect` of a depth charge was ignored. It would always default to the value `3`.

In the following test case I increased the redirect count of the depth charge of the UEF Destroyer to 10.

https://github.com/user-attachments/assets/68fa6e63-9c84-4a6e-a36b-d62aae46c0dd

## Checklist

### DIFF
--- a/changelog/snippets/fix.6507.md
+++ b/changelog/snippets/fix.6507.md
@@ -1,0 +1,3 @@
+- (#6507) Fix depth charges not applying the ProjectilesToDeflect blueprint field
+
+As a result, all depth charges would also deflect up to 3 projectiles. Mostly applies to torpedo defenses.
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -95,6 +95,9 @@ local VectorCached = Vector(0, 0, 0)
 ---@field OriginalTarget? Unit
 ---@field DamageData WeaponDamageTable
 ---@field MyDepthCharge? DepthCharge
+---@field MyFlare? Flare
+---@field MyUpperFlare? Flare
+---@field MyLowerFlare? Flare
 ---@field CreatedByWeapon Weapon
 ---@field IsRedirected? boolean
 ---@field InnerRing? NukeAOE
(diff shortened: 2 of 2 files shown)


lL1l1 (commented):
"Depth charge" is confusing with the depth charge weapon. I think we should use "Anti-Torpedo Flare" because that makes sense when you know about Aeon tactical missile flares, and the name is already used by the Mermaid, Brick, Megalith, and Barracuda. UEF's display name is "Smart Charge AntiTorpedo" and Salem is an outlier with "Anti Torpedo Solution".

lL1l1:
You have to respawn the units for the changes to take effect since this mechanic uses the weapon's blueprint, and blueprints are cached on instantiation (`self.Blueprint`), so you have to respawn the unit and its weapon for it to update.
You may also find useful the hotkey "toggle collision debug information" (equal to `dbg c` console command) to see how the projectiles hits the flare's hitbox, and increasing the projectile lifetime in the antitorp weapon's blueprint so that it can catch more torpedoes.
{6E383769-F0F7-4ED6-925C-A448CD7117EC}

Tagada14 (approved):

Garanas:
> "Depth charge" is confusing with the depth charge weapon. I think we should use "Anti-Torpedo Flare" because that makes sense when you know about Aeon tactical missile flares, and the name is already used by the Mermaid, Brick, Megalith, and Barracuda. UEF's display name is "Smart Charge AntiTorpedo" and Salem is an outlier with "Anti Torpedo Solution".

I agree, but we should process that in a separate pull request. It's also a little tricky because if we change the names (of the Lua classes) then we'll also break mod compatibility.

BlackYps:
When it's just about the class name then I suggest to just add a comment to explain what the class is actually about. This way we don't break any mods

lL1l1:
I meant it in the context of the changelog snippet, but BlackYps makes a good suggestion as well.

PR #6511, opened 2024-11-05
fix mesh table srl0310 (Pulsar)

Uveso:
Removed unnecessary mesh tables that were causing errors.

Caused error:
```
Initializing mesh blueprint for /units/srl0310/srl0310_mesh
    Initializing field LODs of RMeshBlueprint at 0x163e8a80
        WARNING: Invalid type for RMeshBlueprintLOD at 0x12b9b84c: nil
```

This may have been for testing purposes and should be commented out rather than deleted?!?

### DIFF
--- a/units/SRL0310/SRL0310_unit.bp
+++ b/units/SRL0310/SRL0310_unit.bp
@@ -103,16 +103,6 @@ UnitBlueprint{
                     Scrolling = true,
                     ShaderName = "Insect",
                 },
-                {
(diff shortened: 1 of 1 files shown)


Garanas:
I do not see the mentioned error in the game log. What is your game setup (mods, branch)? Because the relevant LODs are created by @The-Balthazar and do exist in the repository. And they work fine when I launch a game from `develop`

PR #6515, opened 2024-11-07
Update UEF SACU Enhancement Script

MrRowey:
Following on the changes of https://github.com/FAForever/fa/pull/6498

## Checklist


units/UEL0301/UEL0301_script.lua
```diff
@@ -171,105 +171,147 @@ UEL0301 = ClassUnit(CommandUnit) {
...
+            WARN("Missing enhancement: ", enh, " for unit: ", self:GetUnitId(), " note that the function name should be called: ", ref)
         end
     end,
 
     ---@param self UEL0301
     ---@param intel IntelType
     OnIntelEnabled = function(self, intel)
         CommandUnit.OnIntelEnabled(self, intel)
-        if self.RadarJammerEnh and self:IsIntelEnabled('Jammer') then
+        if self.ProcessEnhancementRadarJammer and self:IsIntelEnabled('Jammer') then
```
Garanas, line 314: This is an unintentional change - it would always be true. See also the `RadarJammerEnh` field that is set to `true`/`false` depending on whether the enhancement exists.

units/UEL0301/UEL0301_script.lua
```diff
@@ -171,105 +171,147 @@ UEL0301 = ClassUnit(CommandUnit) {
         attachee:SetDoNotTarget(false)
     end,
 
+
+    -- ============================================================================================================================================================
+    -- ENHANCEMENTS
+
+    ProcessEnhancementPod = function(self, bp)
```
Garanas, line 178: Please add the basic annotations, which would boil down to the following in my opinion:

```
---@param self UEL0001
---@param enhancementBlueprint table
```

Garanas (changes requested):

units/UEL0301/UEL0301_script.lua
MrRowey: yes, theses were meant to be marked as draft as I had not got around to doing the annotation.

units/UEL0301/UEL0301_script.lua
MrRowey: After looking at this can't this just be a part of the 2 Functions for the enhancements now?

MrRowey: So something like

```lua
    ---@param self UEL0301
    ---@param bp UnitBlueprintEnhancement
    ProcessEnhancementRadarJammer = function(self, bp)
        self.RadarJammerEnh = true
        self:SetIntelRadius('Jammer', bp.NewJammerRadius or 26)
        self:EnableUnitIntel('Enhancement', 'Jammer')
        self:AddToggleCap('RULEUTC_JammingToggle')

        if self.RadarJammerEnh == true then
            if self.IntelEffects then
                self.IntelEffectsBag = {}
                self:CreateTerrainTypeEffects(self.IntelEffects, 'FXIdle',  self.Layer, nil, self.IntelEffectsBag)
            end
            self:SetEnergyMaintenanceConsumptionOverride(bp.MaintenanceConsumptionPerSecondEnergy or 0)
        end

    end,
    ```

MrRowey: but then do you even need the `self.RadarJammerEnh = true` if the part its setting ti for can now be apart of the same function as it wasn't before ?

lL1l1: You can't inline `OnIntelEnabled` because that function is called when the jamming toggle is pressed (or whenever else the jammer is enabled), while the enhancement function is called only once when the enhancement is built. Same for `OnIntelDisabled`.

If you did inline it, then the jammer's FX and and energy consumption would remain permanently on until the enhancement is removed.

Garanas (approved):

PR #6520, opened 2024-11-07
Nerf Emissary against shield assistance

lL1l1:
## Issue
Emissary is better against shield assistance than even T4 artillery because just two of them can kill any shield in one volley in the span of 1 second between the damage instances.
Currently it is impossible to shield assist against two Emissaries because engineers have a roughly 1 second reaction time when assisting a shield before they start repairing it. With #6464 the reaction time would be 0.1 seconds, resulting in the following absurd numbers of T3 engineers/engineering stations ("Assist" is to out-repair the DPS of 1 Emissary, "NextSurvival" is to survive the overkill from two shots):
image

Other artillery don't require more buildpower to survive the next shot than buildpower to survive the DPS, so Emissary having up to 5x in that place makes little sense.

Building off of #6464's reaction time changes, the DoT time is increased 1s -> 3.8s to produce much more reasonable amounts of buildpower needed to sustain a shield against two Emissaries.
image

Keep in mind the assist cost is for only 1 Emissary's DPS. When looking at 2 Emissaries, the number of engis needed to out-assist DPS would be more than or equal to what is needed to survive the overkill during the DoT for the highest tier shields.

## Reasoning for the changes
While the Emissary does need a good shield-breaking ability to compensate for its low fire rate and AoE being unable to kill shield structures themselves instead of just taking down shields, that ability shouldn't be so powerful that it makes shield assistance completely impossible. It also shouldn't go against the trend of requiring less engis to survive overkill than to survive artillery DPS, as that creates frustrating inconsistencies.

This isn't too strong of a nerf, as compared to other artilleries, you need enough engis to assist against the full DPS of the 2 Emissaries just to survive the overkill, while other artilleries only need half the DPS assistance's engis to survive the next shot and buy time for another shield to recharge. It also still leaves wide open the strong combo of 3 Emissaries killing any assisted shield, since 3 artilleries firing together is generally as powerful as a game ender with any faction's artillery.

## Additional context

A graphics improvement for the Emissary's hit effect would be even more necessary with a longer DoT.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6520.md
+++ b/changelog/snippets/balance.6520.md
@@ -1,0 +1,2 @@
+- (#6520) Increase the DoT time of the Aeon T3 Static Artillery (Emissary) to make it possible to block 2 Emissaries with a reasonable amount of buildpower repairing the shield.
+  - Emissary DoTTime: 1 second -> 3.8 seconds
--- a/units/UAB2302/UAB2302_unit.bp
+++ b/units/UAB2302/UAB2302_unit.bp
@@ -149,7 +149,7 @@ UnitBlueprint{
             DamageType = "Normal",
             DisplayName = "Sonance Artillery",
             DoTPulses = 2,
-            DoTTime = 1,
+            DoTTime = 3.8,
             EnergyDrainPerSecond = 4250,
             EnergyRequired = 17000,
             FireTargetLayerCapsTable = {


lL1l1:
Note:
This would make 1 emissary slightly better against unassisted shields because the DoT would cause regeneration interruption for 6 seconds total (3s interrupt + 3s interrupt after 3.8s) instead of 4 seconds (3s interrupt + 3s interrupt after 1s).
If the DoT is split into more damage instances, the regen interruption would extend to 6.8 seconds.
The difference in actual health regenerated would be rather minor compared to the damage of the artillery (12000): 1344 (4s regen interrupt) ->1008 (6s interrupt) -> 873.6 (6.8s interrupt).

PreciseBump38:
Did some testing using T3 Aeon Engineers (32.5 bp) to assist T3 Aeon shields.

Values below are the number of engineers required to reliably prevent shield drop of a single shield (some very small variability was observed for test with multiple artillery, probably due to slightly random impact time variation depending on exact hit location).

- For testing multiple artillery pieces, the shots were synchronised (as close as possible).
- Artilleries were tested without adjacency bonuses, so in practice required engineers would generally be higher

| Test | Version | Artillery | Engineers Required |
| - | - | - | - |
| 1 | Old/New | 1 x Duke | 6 |
| 2 | Old | 1 x Emissary| 6 |
| 3 | New | 1 x Emissary | 6 |
| 4 | Old/New | 2 x Duke | 13 |
| 5 | Old | 2 x Emissary | 57 |
| 6 | New | 2 x Emissary | 18 |

Conclusions:
- Current Emissary is substantially stronger then Duke (UEF) against shield assisting engies.
- The proposed fix is trivially simple (single value change in the BP).
- The proposed fix does materially reduce the assist required to defend against two Emissary.
- The proposed fix leaves the Emissary with an advantage over the Duke, preserving a degree of faction uniqueness.

All happy here, would just need balance team signoff.

PreciseBump38 (approved):

lL1l1:
This needs a graphical update for the impact effect.

BlackYps:
Closing due to inactivity.

PR #6522, opened 2024-11-09
Fix Solace torpedo damage

Tagada14:
In #5725 the number of child projectiles was accidentally increased from 2 to 3 resulting in increased total damage (6k rather than 4k). The damage is now properly split between child projectiles as well. To compensate the original damage was multiplied by the number of child projectiles.

## Checklist


changelog/snippets/balance.6522.md
```diff
@@ -0,0 +1,4 @@
+- (#6522) In #5725 the number of child projectiles was accidentally increased from 2 to 3 resulting in increased total damage (6k rather than 4k). This change was reversed and the damage is now properly split between child projectiles as well. To compensate the original projectile damage was multiplied by the number of child projectiles.
```
lL1l1, line 1:
```suggestion
- (#6522) In #5725 (version 3777, December 2023) the number of child projectiles for the Solace was accidentally increased from 2 to 3 resulting in increased total damage (6k rather than 4k). This change is reversed, and the damage is now properly split between child projectiles as well. To compensate, the original projectile damage is multiplied by the number of child projectiles.
```
In addition, the technical details about how the projectile damage is split may not be necessary, since the noticeable gameplay change is that the projectiles now deal full damage if they hit an enemy before entering the water.

lL1l1 (dismissed):
The UI uses the projectile's `Fragments` field to multiply damage per projectile, but I guess that's poor design: it causes inconsistent damage amounts if the projectile doesn't split before hitting, and is harder to edit since you have to pay attention to 2 files just to change damage. I would be on board with changing it for the other projectiles that use the Fragments field, but in a separate PR.

Using the `SplitComponent` and changing the projectile damage data (that is then passed to the children) upon entering the water would be the correct way to do this without using `Fragments`. Like in the [Seraphim torpedo script](https://github.com/FAForever/fa/blob/27f5d31f9aefa1ff15da660ea597699b833e9b50/projectiles/SANHeavyCavitationTorpedo01/SANHeavyCavitationTorpedo01_script.lua).

projectiles/AANTorpedoCluster01/AANTorpedoCluster01_script.lua
```diff
@@ -44,10 +44,14 @@ AANTorpedoCluster01 = ClassProjectile(ATorpedoCluster) {
...
-        -- create child projectiles
         local bp = self.Blueprint.Physics
         local numProjectiles = bp.Fragments
-        for i = 0, numProjectiles do
+
+        -- split damage over each child
+        self.DamageData.DamageAmount = self.DamageData.DamageAmount / numProjectiles
+
+        -- create child projectiles
+        for i = 0, (numProjectiles - 1) do
```
Hdt80bro, line 54: This confused me at first, could you do this?
```lua
for i = 1, numProjectiles do
```

lL1l1:
@Garanas Just want a quick review that it's fine to change the split component like this. I do think it needs a refactor in the future to 1. work with the UI through blueprints and 2. have better defined child split and child movement behavior, since it's currently tuned for Cybran projectiles specifically it seems.

lL1l1 (approved):

PR #6523, opened 2024-11-09
UVD: Remove unnecessary UI elements when the selected unit does not have a special `armorType`

Basilisk3:
This is only a small UI tweak for units without a special `armorType` like, for example, the Striker. The current `armorType` portion of uvd.lua looks like this for these units:
```
Armor Type: Normal          (% of damage taken)
Normal - 100.0

```
This is the new output:
```
Armor Type: Normal
```

Both `(% of damage taken)` and `Normal - 100.0` do not provide the player with any relevant information, so they can be hidden.

Only affects units without a special `armorType`.

## Checklist

### DIFF
--- a/changelog/snippets/features.6424.md
+++ b/changelog/snippets/features.6424.md
@@ -3,3 +3,5 @@
     - For example, the Ythotha (Seraphim Experimental Assault Bot), which has multiple weapons, is now displayed as possessing a combined total of `3794` direct fire DPS. This is displayed in an additional field in the UI, the stats of the individual weapons of the unit remain accessible in the same locations as before.
 
 - (#6424) Do not display the Mole's Target Tracker in the additional unit details displayed when `Show Armament Detail in Build Menu` is enabled.
+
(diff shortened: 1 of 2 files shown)


lL1l1:
What if a mod changes the normal armor to take different damage from something else? I think an armor type's properties should not be shown only if they are all 1x damage taken.

Basilisk3:
I applied your feedback. Unfortunately, removing `(% of damage taken)` reliably without any side effects has proven difficult.

lL1l1:
If you don't mind I added a way that removes the "% of damage taken" text. Basically `table.insert` can insert at the first position, making it easy to keep everything the same then just add the header line in the beginning of the lines, after checking if the armor type takes any adjusted damages.

Basilisk3:
Thanks, tested and works.

lL1l1:
I think it's best to display it so that players and modders know that armor types exist, even if normal armor doesn't have any special modifiers.

BlackYps:
Makes sense

BlackYps (approved):

PR #6525, opened 2024-11-10
fix GetAngleInBetween parameter references

relent0r:
## Description of the proposed changes
Fixes a small bug in the refactor of GetAngleInBetween. It was references indexes but the function that calls it is using named keys.

fixes #6524

p.s this needs someone smarter than me to validate I'm actually fixes this correctly as I'm unsure.

## Testing done on the proposed changes
Tested to make sure the scathis does not error and shoots things.

## Checklist
- [ ] Changes are annotated, including comments where useful
- [ ] Changes are documented in the changelog for the next game version

PR #6528, opened 2024-11-10
Use `Vector` types instead of tables for Scathis script 

### LINKED ISSUE #6524: [Bug]: faf develop scathis does not fire
### Describe the Bug

Once a scathis is built or spawned it will not fire.

Error seen in the logs.

warning: Error running lua script: g:\code\relent0r\fa\lua\utilities.lua(412): attempt to perform arithmetic on local `z2' (a nil value)
         stack traceback:
         	g:\code\relent0r\fa\lua\utilities.lua(412): in function `GetAngleInBetween'
         	g:\code\relent0r\fa\units\url0401\url0401_script.lua(78): in function `CreateProjectileAtMuzzle'
         	...ent0r\fa\lua\sim\weapons\defaultprojectileweapon.lua(1056): in function <...ent0r\fa\lua\sim\weapons\defaultprojectileweapon.lua:969>

### Reproduce the bug

Spawn a scathis with a target to fire at on the current faf develop commit as of 5ae907e3dac7516278c3ff8824bc5ac9e1823c04

### Screenshots

_No response_

### Additional context

_No response_

lL1l1:

Fixes #6524.
- Changes the Scathis's script to use cached `Vector`s instead of local tables `{ x: number, y: number, z: number}`
This is more performant and easier to maintain.
  - For performance, the vectors are also accessed with indices instead of hash keys, avoiding a Vector's `__newindex` call.
- Annotate and explain the Scathis script

There are 3 scenarios for the `CreateProjectileAtMuzzle` function changed in this PR: 
- Firing after unpacking (`self.initialaim`)
- Firing while unpacked, but changing targets (`self.losttarget`)
- Firing without changing targets (no special bool)

In all three scenarios the Scathis fires without errors in the log. The fake barrels are misaligned with the real barrel due to a refactoring error fixed in #6525.

## Additional context

- How should unit weapon annotations be handled? Classes seem to be global, so the name of the unit weapon class has to be unique. I opted for "[UnitID]_[WeaponName]" here.
- I dislike how the script is designed because it interrupts the standard weapon cycle for its own animations, and would personally redesign it to use the `OnStartTracking` and `OnStopTracking` weapon functions.

## Checklist

### DIFF
--- a/lua/EffectTemplates.lua
+++ b/lua/EffectTemplates.lua
@@ -1456,7 +1456,7 @@ CArtilleryFlash01 = {
     EmtBpPath .. 'proton_artillery_muzzle_08_emit.bp',
 }
 CArtilleryFlash02 = {
-    EmtBpPath .. 'proton_artillery_muzzle_07_emit.bp',
+    EmtBpPath .. 'proton_artillery_muzzle_07_emit.bp', -- Large, faint rings of air expanding outwards
 }
 
 CArtilleryHit01 = DefaultHitExplosion01
--- a/units/URL0401/URL0401_Script.lua
+++ b/units/URL0401/URL0401_Script.lua
@@ -24,7 +24,9 @@ URL0401 = ClassUnit(CLandUnit) {
         ---@field PitchRotators moho.RotateManipulator[] # Pitch rotators for the fake turret barrels
         ---@field currentbarrel number # Which barrel is currently aligned with the aim's yaw
         ---@field Goal number # Yaw goal of fake barrels
-        ---@field restdirvector { x: number, y: number, z: number }
+        ---@field restdirvector Vector
+        ---@field dirvector Vector
+        ---@field basedirvector Vector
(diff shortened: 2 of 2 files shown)


Garanas:
The use of tables as a replacement for `Vector3(x,y,z)` is not uncommon in the codebase. Do you anticipate more of these type of issues?

> How should unit weapon annotations be handled? Classes seem to be global, so the name of the unit weapon class has to be unique. I opted for "[UnitID]_[WeaponName]" here.

I think that solution is fine.

> I dislike how the script is designed because it interrupts the standard weapon cycle for its own animations, and would personally redesign it to use the OnStartTracking and OnStopTracking weapon functions.

I agree, and I also do not approve of the naming convention that is used 😄 . Feel free to pick this up in a separate pull request.

lL1l1:
> The use of tables as a replacement for Vector3(x,y,z) is not uncommon in the codebase. Do you anticipate more of these type of issues?

As long as the tables are using arrays instead of hash tables the utility functions should work on them; the functions should be written using array accesses.

Garanas (approved):

PR #6530, opened 2024-11-11
Remove ASF armor against Ahwassa bomb

lL1l1:
## Reasoning for the changes
Since a delay was added to the Ahwassa bomb, it is possible to micro ASF out of the explosion, and the armor against it is no longer necessary.

- Removes the armor type: I don't think anyone outside of FAF was using the damage type for this specific balance feature.
- Changes the Ahwassa bomb's damage type to normal.

- Checked that the damage type is not mentioned anywhere.
- Verified that ASF take full damage
- Verified that it is possible to move out of range of a ground fired bomb's explosion even if it lands on top of the ASF
  - Note: since the damage radius is spherical, the high altitude of ASFs means that the bomb actually has as small effective blast radius against ASF, about half the ground radius.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6530.md
+++ b/changelog/snippets/balance.6530.md
@@ -1,0 +1,2 @@
+- (#6530) Remove ASF's armor against Ahwassa bomb, as it is no longer necessary with the new delayed explosion.
+  - Air superiority fighters: Damage taken from Ahwassa bomb 10% -> 100%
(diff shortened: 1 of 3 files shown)


Garanas:
I totally agree 😃 .

Tagada14 (approved):

Rhaelya (approved):

Garanas:
Looks like a great commit message @Rhaelya , good job 👍 . Left out the unnecessary details.

PR #6532, opened 2024-11-12
Fix excess projectiles in the same tick impacting entities that die that tick

lL1l1:
## Issue
If multiple projectiles hit a target in the same tick, they will all impact it regardless of the target's HP.
For example, 6 TML (36000 damage) placed in a line will all impact a 13000 HP shield if they hit on the same tick. In general, any line/concave formation of units could be susceptible to this bug.

This is because collisions are checked for all projectiles before any impact scripts are run, so the usual `OnCollisionCheck` `if self.DisallowCollisions then return false end` doesn't work if `DisallowCollisions` is adjusted due to damage taken from impacts.

- Check `DisallowCollisions` in the `OnImpact` function of the projectile, so that projectiles have an up-to-date status on the entity's collision as HP gets adjusted due to previous impacts.
- Add the `DisallowCollisions` boolean to shields because:
  1. `_IsUp` would be an extra check for all projectiles
  2. `_IsUp` is adjusted by state changes, which are forked threads that happen after all impact scripts are run.

Use a logging statement to see how many projectiles impact a unit or shield and how many impact the terrain. The projectiles are from 15 TML stacked on top of each other, although realistic scenarios like 6 TML in a line can also impact a shield in the same tick.
The logging statement shows that just enough projectiles to kill the unit/shield hit it, and the rest passed on to hit the terrain behind/under it.

Tested with a shield, a Ythotha (unit with death animation), Percival (standard unit), and a Tempest (sinking animation unit). Before the changes, all 15 TML would impact the target, and after the changes only the necessary amount impact the target.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6532.md
+++ b/changelog/snippets/fix.6532.md
@@ -1,0 +1,1 @@
+- (#6532) Fix excess projectiles impacting entities that die in the same tick. For example, 6 tactical missiles impacting a shield at the same time, despite only 3 being needed to bring down the shield.
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -150,6 +150,7 @@ end
 ---@field ImpactMeshBp string
 ---@field SkipAttachmentCheck boolean
 ---@field AbsorptionTypeDamageTypeToMulti table<DamageType, number>
+---@field DisallowCollisions boolean
 Shield = ClassShield(moho.shield_methods, Entity) {
 
     RemainEnabledWhenAttached = false,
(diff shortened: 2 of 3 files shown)


Garanas:
This is tricky. I do not perceive this to be that big of an issue. I am not sure if it is even a bug. It's definitely a balance chance - various units that fire multiple projectiles the same tick may now be more viable to penetrating shields. 

I think this requires more discussion about whether it is a bug, and if it is, whether it is worth introducing the additional logic to fix it.

PR #6534, opened 2024-11-13
Commander Transport Search Hotkey Upgrade

clyfordv:
Modifies the existing select/zoom to commander keyactions to use a new SelectCommander function, which searches possible transports and selects/zooms to them if it doesn't find a commander on the ground.

For cases of multiple commanders:
- Prioritizes the nearest commander not in a transport (like the stock implementation)
- If there's no grounded commanders, it finds one in a transport at random. There's unfortunately no method for the UI to search by both attached units and distance

In case of no commander it clears the selection


Garanas:
@clyfordv this is branched from `deploy/fafdevelop` instead of `develop`. Next time, please branch from `develop`

Garanas (approved):

PR #6535, opened 2024-11-15
Strat agility buff

lL1l1:
## Issue
Strategic bombers are in a difficult spot with balance, having steep restrictions on cost/speed due to strat rushes, restrictions on HP/damage due to undesired snipe gameplay, and restrictions on the damage model due to being a strategic bomber designed to destroy key targets.

Due to energy cost and buildtime nerfs to further disincentivize strat rushes, strats are very underused compared to T3 gunships, which are the dominant air to ground option until air T4s can be built as a mere alternative. Strategic bombers should be buffed to improve variety in T3 air to ground options.

Due to the above restrictions, the only possible buffs are to the maneuverability and reliability of strategic bombers, which is the route this PR takes:
- Maneuverability: 
  - Increase turn speed 0.8 -> 1.5
    This lets it effectively micro against flak damage, and leaves less time/opportunities for ASF/SAMs to deal damage.
  - Increase Lift Factor 7 -> 17
    This improves the speed of the bomber lifting off, complementing the new turn speed.
- Reliability:
  - Bomb drop threshold 3.5 -> 10
    This greatly improves the abilities of the bomber to fire on uneven terrain or during turns, where it may get a less than ideal path. It is necessary because every bomb of a strategic bomber matters, and it can be devastating if one of them doesn't fire because the engine decided the small drop zone was unfortunately in the wrong place. A similar massive increase for the threshold was done for the Ahwassa (4 -> 20) https://github.com/FAForever/fa/pull/2465.
  - Smaller loops on auto attack to maximize DPS: 
  This only increases DPS when bombing a single target with multiple passes, so I wouldn't put it exactly under the HP/DPS buff restriction. Previously bombs would drop only every ~13 seconds, and with the turn speed and breakoff distance reduction they drop every ~8 seconds.
   It can also be argued that this is just simplifying the micro of multiple bombers when you want to maximize DPS. There will of course still be micro against AA or to have good bombing paths that make use of the reload, both of which are far more engaging than figuring out when to turn multiple bombers just to bomb a single target on reload with multiple passes.
    - BreakOffDistance 60 -> 40
    - RandomBreakOffDistanceMult 1.5 -> 1
    
Essentially these changes would encourage a more proactive, powerful usage of strategic bombers taking shots at targets of opportunity before AA can arrive, similar to the Ahwassa.

Tried out an individual bomber in sandbox against idle and moving flak getting a feel for the micro.

Logged the time between bomb drops when groundfiring and targeting a moving flak while adjusting the breakoff distance/turn speed to make use of the 8 second reload.

Used `ai_RenderBombDropZones` console command.

## Checklist

## Summary by CodeRabbit

* **Balance Changes**
  * Strategic bombers gain improved maneuverability (faster turning, greater lift), altered bombing behavior (higher drop threshold, reduced break-off distance, tighter auto-attack range) — resulting in slightly faster bomb cadence (~13s → ~10s) and bombs dropping from farther away.
  * Veterancy progression adjusted: required kills now scale to 2× each bomber's mass.

* **Chores**
  * Changelog and internal configuration notes updated.


StormLanternz:
I like how the changes feel. The strat becomes more agile, which suits its relative small size. That said, I'm not a big fan of making frequent changes to the movement of micro heavy units. This because expertise on microing these units comes with years of practice. When a change like this happens, you lose all that. Besides, wasn't it pretty recent that strats were changed?

So to conclude, I'm a bit on the fence on this one. I really like the removal of limited fuel though. This allows for a reload bar perhaps?

lL1l1:
> I really like the removal of limited fuel though. This allows for a reload bar perhaps?

Unfortunately this breaks automatic undocking from air staging.

Rhaelya (approved):
Everything looks in order.

PR #6543, opened 2024-11-18
Disable collisions with dead units consistently

lL1l1:
## Issue
This bit of code runs for all units and causes them to wait while dead before being destroyed, making them block damage when they shouldn't:
https://github.com/FAForever/fa/blob/9948335f0fe0699dead4f1e0070f08ebcfdd9fcc/lua/sim/Unit.lua#L1982-L1994

This code does the same, but makes up a much more significant portion of the wait time (0-0.5 seconds vs 0.1)
https://github.com/FAForever/fa/blob/9948335f0fe0699dead4f1e0070f08ebcfdd9fcc/lua/sim/Unit.lua#L1921-L1924

For a practical example, spawn a line of hives and an enemy monkeylord. The monkeylord's laser will be blocked while the dead hive is going through its explosion FX.

- add `DisallowCollisions = true` to `OnKilled` as that is the entry point for all the issues that arise with collisions after death (`DeathThreadDestructionWaitTime`, sinking, and death animations)
  - remove the statement in the sinking and death animation cases as it is no longer necessary

- Monkeylord can laser through a line of hives without getting blocked
- Monkeylord can kill a cruiser, then kill another cruiser spawned 1 ogrid behind the dead one (the sinking cruiser's hitbox remains on the surface)
- Monkeylord can kill a monkeylord and the laser immediately goes through to a target behind the monkeylord in the death animation

## Additional context

This does not apply to air units because they overwrite `OnKilled`.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6543.md
+++ b/changelog/snippets/fix.6543.md
@@ -1,0 +1,1 @@
+- (#6543) Fix dead units having collision with projectiles and beams for 0.1-0.6 seconds after death.
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -1474,6 +1474,9 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
         -- this flag is used to skip the need of `IsDestroyed`
         self.Dead = true
 
+        -- don't allow projectiles/beams to collide since we are dead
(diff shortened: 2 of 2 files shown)


clyfordv (approved):
Changes work as advertised. Looks better IMO, definitely a bit of buff to continuous heavy beam weapons.

PR #6544, opened 2024-11-18
Fix unit transfer enabling enhancement-locked weapons

lL1l1:

Fixes a bug where transferring units (Seraphim SACU) enables their enhancement-disabled weapons (Seraphim SACU TML) despite the enhancement not being built.

1. Spawn a Seraphim SACU without any upgrades.
2. Transfer it to another team.
3. It should not start building a missile. You also should not be able to order it to fire a missile by selecting a TML along with the SACU to be able to give the launch and build orders.
  - In contrast, a missile preset is able to load and fire missiles without any workaround.

## Additional context

The overcharge weapon seems to not re-enable itself without these changes anyway, I'm not sure why.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6544.md
+++ b/changelog/snippets/fix.6544.md
@@ -1,0 +1,1 @@
+- (#6544) Fix unit transfer enabling weapons locked by enhancements, such as the TML of the Seraphim SACU.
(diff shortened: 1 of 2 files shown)


clyfordv (approved):
Tested, looks good. The underlying behavior of enhancement activated weapons is a little strange (cybran SACUs can be given attack orders against air units even without the nanite missile enhancement, and will just follow the air unit around as a result, for example) but this fixes other cases like a transferred Sera SACU immediately starting construction of a missile even without the TML enhancement installed.

Garanas (approved):
Great find and looks good!

It may be interesting to refactor this in a different pull request to have a function in the unit class to manage this logic instead. That way the same bug does not creep up in other places.

lL1l1:
Good idea, the bug does creep up in the Megalith's hack peg launcher, so making it accessible from the unit class allows an easy override to fix that.

PR #6546, opened 2024-11-19
Fix billy nuke launches being synced to UI

lL1l1:
## Issue
An exploit that @clyfordv's mod discovered, where you can be notified of enemy billy nuke launches because the launch data is synced despite not having a VO line play.

Fixes the exploit by adding the VO's condition from `Unit.NukeCreatedAtUnit` to the launch data sync condition. Allies still get pings for billy nukes.

Log the data that is synced when a nuke launches:
- Allies/yourself/observers receive data + location data of billy nukes and nukes
- When switching to an enemy army before launching the nuke, you receive data only for nukes and no data at all for billy nukes.

## Additional Context
There is currently no visual indicator for strategic launches by default, but it can be modded in.
Effectively a continuation of the anti-cheat logic in https://github.com/FAForever/fa/pull/5582.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6546.md
+++ b/changelog/snippets/fix.6546.md
@@ -1,0 +1,1 @@
+- (#6546) Fix an exploit that allows being notified of enemy Billy nuke launches.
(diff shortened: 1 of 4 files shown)


clyfordv (approved):
Tested, changes are good, notifications work as expected for billy and non-billy nukes.

PR #6547, opened 2024-11-19
Do not use the script-defined playable area when it is tiny

Garanas:

Reported by @maudlin27 on the AI Discord, see also:

- https://discord.com/channels/619646141306503181/941273132818391100/1289285154145046623

Generated and unexplored maps temporarily set the playable area to 4. This confuses the navigational mesh. When the playable area is too small we instead opt to ignore it. This solution is not ideal.

## Checklist


maudlin27 (approved):
Did a basic test on an unexplored mapgen and on a normal (explored) map and it seemed to work as expected, while the condition makes sense to me

PR #6549, opened 2024-11-22
Fix Czar missing carrier auto deploy functionality

lL1l1:
## Issue
CZAR does not have the `CARRIER` category, so `blueprints-units.lua` external factory processing for carriers does not pick it up:
https://github.com/FAForever/fa/blob/3f034d13505450b406994746590d6668ed322826/lua/system/blueprints-units.lua#L675-L698

- Manually add in the order override and stat toggle added by `blueprints-units.lua`.
- Fix type annotation of the `StatToggles` blueprint field. Not sure why it was incorrectly written in #5581 5dcaac5771cc26d84ec0326de72707e47d4a13d4

The button's description is corrected, and when you right click it gets highlighted and constructed units automatically deploy.
{E107CE8C-0027-4862-99E4-93C3A9E0314A}
{E6F41DE5-5E96-4EB5-A765-5367404447B9}

## Additional context

This is the second bug I've seen because of CZAR being a flying carrier, the first being that intel/counterintel remains enabled for units inside the CZAR because its base class is that of an air transport and not a (naval) aircraft carrier.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6549.md
+++ b/changelog/snippets/fix.6549.md
@@ -1,0 +1,1 @@
+- (#6549) Fix CZAR's deploy button not being like that of other aircraft carriers where you can right click it to enable automatic deployment of built units.
--- a/engine/Core/Blueprints/UnitBlueprint.lua
+++ b/engine/Core/Blueprints/UnitBlueprint.lua
@@ -979,7 +979,7 @@
 --- table of toggle capabilities available for this unit
 ---@field ToggleCaps table<ToggleCap, boolean>
 --- table of boolean toggles set/got with SetStatByCallback/GetStat
----@field StatToggles table<string, table>
+---@field StatToggles? table<string, true>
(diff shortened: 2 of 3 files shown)


Garanas:
What would be the downside of giving it the `CARRIER` category? What are the side effects?

lL1l1:
`CARRIER` is frequently used as if it is implicitly a naval unit category. It's more like the `DESTROYER`, `CRUISER`, and `BATTLESHIP` categories where it signals a rather specific type of unit than the general ability to carry aircraft.

Garanas:
I understand. It may also mess up existing category queries where they expect naval-based carriers, but do not specify the naval category.

Garanas (approved):

lL1l1:
I don't remember why I closed this. I guess because it already has the CARRIER category?

PR #6559, opened 2024-11-28
Fixes and QoL improvements for Megalith's build capabilities

lL1l1:

- Make the build range longer because with 1 build range you can drag the build preview to start building at a great distance (similar to interrupt pathfinding), but without that micro the unit won't build in a sensible fashion, constantly walking around to position itself on top of or next to the build target.
- [x] Update build range preview to work off unit footprint + `MaxBuildDistance` + target skirt size (minimum 1, so we use 1).
  - This will probably involve reducing the build range.
- Add repair capability so that it can finish eggs it started
- Add reclaim capability since it can already reclaim by building eggs on top of wrecks
- Add pause order since it has engineering capabilities
- [x] Fix out of date T3 mobile arty and T3 mobile AA egg cost stats.
- Use `AlwaysAlignToTerrain` for crab eggs since they're temporary and very small, like walls, so they shouldn't flatten the terrain with `FlattenSkirt`.
- Remove the `BuildAttachBone` that doesn't exist for the eggs so it will break if used.

You can place eggs around the Megalith to build them instead of having to drag the build preview out around the Megalith before it starts building them.

Notably Megalith's huge footprint does let it build quite far out of its build range ring if you drag the build preview (to be fixed).

Its footprint also somewhat acts as a minimum build range of sorts, within which its AI starts pathfinding very weirdly. The skirt size might block the build site too.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Extended egg build range around the Megalith; eggs can be placed slightly farther
  * Added Repair, Reclaim and Pause commands for eggs/engineering units

* **Bug Fixes**
  * Build range preview now reflects the real build range

* **Balance Changes**
  * Reduced production costs and build times for select units
  * Updated T3 Mobile Artillery and Anti‑Air stats

* **Improvements**
  * Better terrain alignment and footprint‑aware build overlay visuals


Rhaelya (approved):
Changes work as described

PR #6565, opened 2024-11-29
Fix matchmaker games not launching

BlackYps:
Hosts of matchmaker games got the following error: `warning: GPGNET: Error processing gpg.net command CreateLobby: Attempting to lookup the RType for class Moho::CGpgNetInterface before it is registered.`

According to the debugger, the log statement was at fault.


lL1l1 (approved):
Let's try this.

clyfordv (approved):
Changes are as advertised for removing the LOG argument.

PR #6569, opened 2024-11-30
Expose player ratings and divisions to the session

Garanas:

Extends https://github.com/FAForever/fa/pull/6479 by implementing the details of https://github.com/FAForever/fa/pull/6160. I'm not sure how I forgot about this - the rating and divisions are also not part of the checklist. 

Extended the launch script made by @lL1l1 to include (sub)divisions. Launch the game via the script. Observe the divisions in the scoreboard.

image

And similar for clan tags with [8e2b159](https://github.com/FAForever/fa/pull/6569/commits/8e2b159414680e2c1be02ab68909b82eb0f58765):

image

## Additional context

Discussion on [Discord](https://discord.com/channels/197033481883222026/1312271501776388146) whether we should show this data by default to players.

## Checklist

### DIFF
--- a/lua/ui/lobby/autolobby/AutolobbyController.lua
+++ b/lua/ui/lobby/autolobby/AutolobbyController.lua
@@ -110,7 +110,7 @@ AutolobbyCommunications = Class(MohoLobbyMethods, AutolobbyServerCommunicationsC
 
         self.LocalPeerId = "-2"
         self.LocalPlayerName = "Charlie"
-        self.PlayerCount = tonumber(GetCommandLineArg("/players", 1)[1]) or 2
+        self.PlayerCount = self:GetCommandLineArgumentNumber("/players", 2)
         self.HostID = "-2"
 
         self.GameMods = {}
@@ -150,18 +150,18 @@ AutolobbyCommunications = Class(MohoLobbyMethods, AutolobbyServerCommunicationsC
 
         -- retrieve team and start spot
         info.Team = self:GetCommandLineArgumentNumber("/team", -1)
-        info.StartSpot = tonumber(GetCommandLineArg("/startspot", 1)[1]) or -1 -- TODO
+        info.StartSpot = self:GetCommandLineArgumentNumber("/startspot", -1)
 
         -- determine army color based on start location
         info.PlayerColor = GameColors.MapToWarmCold(info.StartSpot)
         info.ArmyColor = GameColors.MapToWarmCold(info.StartSpot)
 
         -- retrieve rating
-        info.DEV = tonumber(GetCommandLineArg("/deviation", 1)[1]) or 500
-        info.MEAN = tonumber(GetCommandLineArg("/mean", 1)[1]) or 1500
-        info.NG = tonumber(GetCommandLineArg("/numgames", 1)[1]) or 0
-        info.DIV = (GetCommandLineArg("/division", 1)[1]) or ""
-        info.SUBDIV = (GetCommandLineArg("/subdivision", 1)[1]) or ""
+        info.DEV = self:GetCommandLineArgumentNumber("/deviation", 500)
+        info.MEAN = self:GetCommandLineArgumentNumber("/mean", 1500)
+        info.NG = self:GetCommandLineArgumentNumber("/numgames", 0)
+        info.DIV = self:GetCommandLineArgumentString("/division", "")
+        info.SUBDIV = self:GetCommandLineArgumentString("/subdivision", "")
         info.PL = math.floor(info.MEAN - 3 * info.DEV)
 
         return info
@@ -195,7 +195,7 @@ AutolobbyCommunications = Class(MohoLobbyMethods, AutolobbyServerCommunicationsC
         }
 
         -- process game options from the command line
-        for name, value in Utils.GetCommandLineArgTable("/gameoptions") do
+        for name, value in self:GetCommandLineArgumentArray("/gameoptions") do
             if name and value then
                 options[name] = value
             else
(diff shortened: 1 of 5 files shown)


lL1l1 (commented):
- Technically divisions aren't needed in the Sim, `GetFrontEndData` and `SetFrontEndData` would be the correct choice, if we ignore being consistent with ratings and clan tags.
- The `/clan` arg also needs to be passed to the UI of other players for clan tags to work. It goes into the `gameInfo.GameOptions.ClanTags` field https://github.com/FAForever/fa/blob/6208f257759778bf9509455d64dc8d68d52d3053/lua/ui/lobby/lobby.lua#L2238
Unfortunately we can't use front end data due to backwards compatibility unless you add a meta table to `SessionGetScenarioInfo().Options.ClanTags` with 
```lua
__index = function(i) 
  return GetFrontEndData("ClanTags")[i] 
end
```
or simply duplicate the code, if it's not being used often anyway (we will only ever have 2 lobbies).
- The lobby does some safeguards for args in case the arg is an empty string and causes the next arg to be grabbed instead of the value (since an empty string is not a valid token for the arg getter).
https://github.com/FAForever/fa/blob/6208f257759778bf9509455d64dc8d68d52d3053/lua/ui/lobby/lobby.lua#L159-L181
This safeguarding is similar to what I did for `GetCommandLineArgTable` more recently:
https://github.com/FAForever/fa/blob/b872a718f8c026690b6b95e475c06b356137f4e4/lua/system/utils.lua#L744-L769

Garanas:
> * Technically divisions aren't needed in the Sim, `GetFrontEndData` and `SetFrontEndData` would be the correct choice, if we ignore being consistent with ratings and clan tags.

I wonder if these show up in the replay somewhere.

lL1l1:
Mentioning replays, it's best to put it in the sim so that replays have a copy of clan tags/divisions/rating through the scenario info table.

BlackYps:
They should be handled in the same way as ratings and clans. They should all not be lost when watching the replay.

Garanas:
@lL1l1 is https://github.com/FAForever/fa/pull/6569/commits/1c6384b80043bd4ad474886f2dda7f647c6f1880 roughly the right direction for proper command line parsing?

Garanas:
@BlackYps it now also supports clan tags.

lua/ui/lobby/autolobby/components/AutolobbyArguments.lua
```diff
@@ -0,0 +1,127 @@
...
+    ---@param default string
+    ---@return string
+    GetCommandLineArgumentString = function(self, option, default)
+        if not self:ValidCommandLineKey(option) then
+            return default
+        end
+
+        -- try to get the first argument
+        local arguments = GetCommandLineArg(option, 1)
+        if arguments and (not option[ arg[1] ]) then
```
lL1l1, line 84: Why not check `string.sub(arg[1], 0, 1) == "/"`? 
- It's much simpler than keeping a list of all valid command line arguments (which get preceded and effectively overwritten by the client's own cmdline args anyway).

lua/ui/lobby/autolobby/components/AutolobbyArguments.lua
```diff
@@ -0,0 +1,127 @@
...
+
+        return default
+    end,
+
+    --- Attempts to retrieve a number-like command line argument
+    ---@param self UIAutolobbyArgumentsComponent | UIAutolobbyCommunications
+    ---@param option string
+    ---@param default number
+    ---@return number
+    GetCommandLineArgumentNumber = function(self, option, default)
```
lL1l1, line 96: Not sure if verifying as thoroughly as checking types is necessary, it just adds to the load of thinking which function is best to use without improving the solution to the original issue: The guards in `lobby.lua` were implemented to prevent the case of the client giving a command line argument `/arg X /arg2 ...` where `X` is an empty string, and `GetCommandLineArg(option, 1)` would end up returning `{"/arg2"}` because an empty string is not a token.

lua/ui/lobby/autolobby/components/AutolobbyArguments.lua
```diff
@@ -0,0 +1,127 @@
...
+    end,
+
+    --- Attempts to retrieve a table-like command line argument
+    ---@param self UIAutolobbyArgumentsComponent | UIAutolobbyCommunications
+    ---@param option string
+    ---@return table<string, string>
+    GetCommandLineArgumentArray = function(self, option)
+        if not self:ValidCommandLineKey(option) then
+            return {}
+        end
```
lL1l1, line 123: the utils function will return an empty table for a non-existent key, so if we don't want to verify if a requested command line key is valid, then this function is not necessary as the default empty table return value is already implemented.

lua/ui/lobby/autolobby/AutolobbyController.lua
```diff
@@ -488,6 +528,11 @@ AutolobbyCommunications = Class(MohoLobbyMethods, AutolobbyServerCommunicationsC
                         self:SendPlayerOptionToServer(ownerId, 'Faction', playerOptions.Faction)
                     end
 
+                    -- tuck them into the game options. By all means a hack, but
+                    -- this way they are available in both the sim and the UI
```
lL1l1, line 532:
```suggestion
                    -- tuck them into the game options. By all means a hack, but
                    -- this way they are available in both the sim (for replays) and the UI
```

lL1l1 (changes requested):
I think all that is necessary is some "get arg or default" function:
```lua
GetCommandLineArgOrDefault = function(option, default)
  local arg = GetCommandLineArg(option, 1)[1]
  if not arg or string.sub(arg, 0, 1) == "/" then
    return default
  else
    return arg
  end
end
```
This guards against the empty string (fromt client) arg case with the least complexity.

lL1l1 (approved):
Tested on test server with accounts [TEC] clan `KAZAKHNYA` and `Straiken` and the fix works: clans, ratings, and divisions appear in the game options for both the live game and in the replay file. Although the style could be better, this PR is due for a hotfix and it is functional, so I'm approving it.
```
UI_Lua print(repr(SessionGetScenarioInfo().Options))
```

PR #6572, opened 2024-12-03
Fix bombers missing during vertical movement

lL1l1:
## Issue
Bombers miss when moving vertically because bombs inherit launcher XYZ momentum rotated onto the XZ plane, but the Lua calculation assumes only launcher XZ momentum is inherited.

Add a few lines fixing the XZ momentum used by the Lua calcuation.

Order a bomber to attack somewhere where it will be changing altitude, it should be perfectly accurate now. The easiest way is to groundfire in front of a landed bomber. You can also groundfire the peak of a mountain.
Left: no fix. Right: fixed

https://github.com/user-attachments/assets/1cc9f317-dfdd-4961-8bb2-5bdb1e8951d1

## Additional context

The engine still calculates the bomb drop zone independently for firing the bomb, and may not let the bomb fire at all. You can use `ai_RenderBombDropZone` to see the drop zone.

Bombs still have a speed limit in the projectile bp in an effort to reduce the minimum bomb drop time in the case of bombing right under the bomber, but this can make them miss when they fire downwards from a high altitude, where the bomb will accelerate a lot.

## Checklist

### DIFF
--- a/lua/sim/weapons/DefaultProjectileWeapon.lua
+++ b/lua/sim/weapons/DefaultProjectileWeapon.lua
@@ -211,7 +211,7 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
         projVelX = projVelX * multiplier
         projVelZ = projVelZ * multiplier
 
-        local targetPos
+        local targetPos, _
         local targetVelX, targetVelZ = 0, 0
 
         local data = self.CurrentSalvoData
--- a/lua/sim/weapons/DefaultProjectileWeapon.lua
+++ b/lua/sim/weapons/DefaultProjectileWeapon.lua
@@ -13,6 +13,7 @@ local UnitGetVelocity = UnitMethods.GetVelocity
 local UnitGetTargetEntity = UnitMethods.GetTargetEntity
 
 local MathClamp = math.clamp
+local MathSqrt = math.sqrt
 
 ---@class WeaponSalvoData
 ---@field target? Unit | Prop   if absent, will use `targetPos` instead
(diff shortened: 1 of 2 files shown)


lua/sim/weapons/DefaultProjectileWeapon.lua
```diff
@@ -204,6 +204,13 @@ DefaultProjectileWeapon = ClassWeapon(Weapon) {
         local projPosX, projPosY, projPosZ = EntityGetPositionXYZ(projectile)
         local projVelX, _, projVelZ = UnitGetVelocity(launcher)
 
+        -- The projectile will have velocity in the horizontal plane equal to the unit's 3 dimensional speed
+        -- Multiply the XZ components by the ratio of the XYZ to XZ speeds to get the correct XZ components
+        local projVelXZSquareSum = projVelX * projVelX + projVelZ * projVelZ
+        local multiplier = math.sqrt((projVelXZSquareSum + _ * _) / (projVelXZSquareSum))
```
Hdt80bro, line 210: Now that the variable is being used, `projVelY` shouldn't be called `_` anymore

lL1l1: `_` gets reused:
https://github.com/FAForever/fa/blob/2821e42d517ceaadd131a35d1dc2a524c0df3ea4/lua/sim/weapons/DefaultProjectileWeapon.lua#L226
https://github.com/FAForever/fa/blob/2821e42d517ceaadd131a35d1dc2a524c0df3ea4/lua/sim/weapons/DefaultProjectileWeapon.lua#L239

Should it be initialized as a local here?
https://github.com/FAForever/fa/blob/2821e42d517ceaadd131a35d1dc2a524c0df3ea4/lua/sim/weapons/DefaultProjectileWeapon.lua#L214

Hdt80bro: In the end, it's all for readability - here it's just indicating that a variable that needs to be there isn't used. Ideally, it'd be redefined every time so that the function can more easily pop its return values into sequential registers, but I don't recall if shadowed variables in Lua get their register slot freed.

Garanas:
@BlackYps this may also be nice to include.

BlackYps (approved):

PR #6574, opened 2024-12-04
Automatically rotate the Monkeylord towards targets, limited outside a range

lL1l1:

Cherry picked the now-reverted Monkeylord auto-rotation changes from https://github.com/FAForever/fa/pull/6437. In summary, the ML has angle-restricted weapons that it cannot use when idle, and they make up a significant 450 DPS, so it would be nice if they fired automatically when they can, but on the other hand the ML is slow turning and reliant on micro so turning it automatically is not exactly desired.

Add a dummy weapon that limits the bolters' auto rotation behavior to outside 50 range:
- The target check interval is 0.1 so that the ML doesn't rotate in the time it takes for the dummy weapon to acquire the target.
- The rate of fire is 0 so that `OnFire` only ever gets called once.
- The tracking radius is 1 so that the range in the BP is what is actually used for the unit behavior.
- The necessary targeting fields mimic the bolters/laser valid targets.
- The `SlavedToBody` fields stop the autorotation behavior of the bolters.
- It is placed after the laser so that the unit walks into laser range on an attack task, but after the bolters so that its `SlavedToBody` angle limits have priority.

I think this is the absolute minimum for an AI dummy weapon, it should be replicated to the other dummy weapons (various forms of AA-only units, and transports).

Testing from the previous PR (since this is cherry picked):
> The units can hit a spirit moving around the unit at max range.

- ML does not rotate for targets outside 50 range, and it goes into laser range when on attack task (attack order, attack move/patrol order).
- Logging the `OnFire` and `OnGot/LostTarget` functions of the base weapon class.

## Checklist
  - will split blueprint annotations to a different PR since this is a dubious change for balance

### DIFF
--- a/changelog/snippets/balance.6437.md
+++ b/changelog/snippets/balance.6437.md
@@ -1,0 +1,3 @@
+- (#6437) Automatically rotate Cybran Carrier, Othuum, and Monkeylord towards their targets when idle or attacking so that all their weapons can fire.
+    - Monkeylord will not automatically rotate when the laser is firing, to make it get in the way of micro less.
+    - The units rotate in place when aiming like this.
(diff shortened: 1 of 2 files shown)


BlackYps:
Rejected by the balance team lead

PR #6585, opened 2024-12-17
WorldviewCommandModeTableStreamline

clyfordv:
- Modifies `controls/worldview.lua` to call `GetCommandMode` once and store as a local table.
- Modifies `commandmode.lua` to only update the `commandModeTable` (what's returned by `GetCommandMode`) when the command mode changes.
- Minor: removes an unecessary check for the alt attack move cursor that didn't respect preferences.


lua/ui/controls/worldview.lua
```diff
@@ -635,7 +633,7 @@ WorldView = ClassUI(moho.UIWorldView, Control) {
                 local cursor = self.Cursor
                 cursor[1], cursor[2], cursor[3], cursor[4], cursor[5] = UIUtil.GetCursor(identifier)
                 self:ApplyCursor()
-                CommandMode.GetCommandMode()[2].reticle = TeleportReticle(self)
+                commandModeTable[2].reticle = TeleportReticle(self)
```
lL1l1, line 636: Feels wrong to stuff data into the commad mode data that is UI related. Imagine if this is a script command with a reticle and we pass unserializable userdata inside a table: the game would immediately crash (as I discovered when annotating script tasks, and I doubt that LuaParams gets passed with any other orders, as it is logically not necessary).

lL1l1 (commented):
Functionally it makes sense.
It is possible to dirty the command mode table and then that would get used by other UI elements looking for the command mode, but that was already possible before (get table -> change table -> yield -> someone else gets the table and resets your changes) without issues so I don't think it'll become a problem. Perhaps a simple warning that the table shouldn't be modified would be suitable?

lua/ui/controls/worldview.lua
clyfordv: I think that line can just be taken out; originally I had the idea that the command mode table was a good vehicle for different UI elements to get access to the reticle (or anything) but it doesn't apply in this case.

(believe I said this before)

Garanas:
I do not fully understand what the issue is that these changes solve. At the same time, it increases coupling and makes it more difficult to understand how `commandModeTable` operates for future maintainers. Initially it would appear to just update magically 😄 !

Could you help me understand?

clyfordv:
I think it's a good question and I'm not sure what the best answer is.

On one hand we have the issues you describe where it updates magically. But that behavior occurs in other places (see `controls.lua`--also not terribly intuitive).

On the other we have a function needlessly called every frame to update a table which is mostly (on a frame basis) the same. But the performance improvements to be had in that area are marginal.

Garanas:
I think the readability of the code outweighs the possible performance improvements. There's a tiny bit of overhead of a function call, but I do not believe it is a significant factor in the code that is changed.

BlackYps:
I also agree that code readability outweighs theoretical performance improvements that are probably not even noticable.

PR #6586, opened 2024-12-17
Firestate Hotkey Instant Response

clyfordv:
- Fixes the minor aggravation of the delay in the display on the order button when changing firestate via hotkey.
- Prevents unnecessary calls to SetAvailableOrders when new selection is same as the old selection (discovered because `SetFirestate` triggers `OnSelection`, for whatever reason).


lL1l1 (changes requested):
> * Fixes the minor aggravation of the delay in the display on the order button when changing firestate via hotkey.

It's nice to see the fire state not delay, but it doesn't reflect what the Sim does, so I'm not sure it's the best, since you can no longer exactly tell what the game is trying to do. Maybe there could be like a progress bar showing the changing process? I think the worst part is actually that the popup menu closes when the fire state updates on the UI side, so you open the menu to change the state, do it wrong for whatever reason, open it again to fix it, but it auto closes due to the 0.5 delay. The delay itself is not a significant issue, after some experience it's expected due to how the game handles every single command.

> * Prevents unnecessary calls to SetAvailableOrders when new selection is same as the old selection (discovered because `SetFirestate` triggers `OnSelection`, for whatever reason).

- This breaks the updating of orders due to enhancements finishing, like TML or teleporter, which are simply using `Unit.AddCommandCap`.
- This also breaks the updating of the question mark that appears for mixed fire states after setting all the fire states using the popup. The question mark should disappear after you do that since the fire states are no longer mixed.

clyfordv:
- On the UI/sim disconnect, all other similar order interactions (auto mode, clicking the firestate button) all reflect the input immediately even if it takes a half second for the sim to reflect the change. I mean even order waypoints show up immediately despite units taking a second to react.
- ... but gotta go back to the drawing board, because without the modification to `OnSelection` you get flickering through the different states when rapidly pressing the hotkey (the auto closing after 0.5 seconds you mentioned happens for the same reason), which imo is jankier than the delay. Addressing both would require refining the call to `SetAvailableOrders` to not recreate all the order buttons under all circumstances.

lL1l1:
> * On the UI/sim disconnect, all other similar order interactions (auto mode, clicking the firestate button) all reflect the input immediately even if it takes a half second for the sim to reflect the change. I mean even order waypoints show up immediately despite units taking a second to react.

Yeah I agree now, I think I was being a bit too harsh on my expectations of the UI showing the network delay. It would be nice to show the delay but it's not required, so instant UI changes despite the delay is preferred over delayed UI changes.

> * ... but gotta go back to the drawing board, because without the modification to `OnSelection` you get flickering through the different states when rapidly pressing the hotkey (the auto closing after 0.5 seconds you mentioned happens for the same reason), which imo is jankier than the delay. Addressing both would require refining the call to `SetAvailableOrders` to not recreate all the order buttons under all circumstances.

I don't think the call itself needs to be refined, I think the function itself should verify if the orders of the selection changed before refreshing the UI elements. Maybe ideally an element doesn't get refreshed unless that element's order state was changed in the selection.

BlackYps:
Closing due to inactivity.

PR #6588, opened 2024-12-18
Reduce Seraphim Destroyer underwater speed 5 -> 4.5

lL1l1:

The Seraphim destroyer is currently too powerful due to its micro potential, but nerfing its main stats like range or damage would make it weaker than it already is when not micro'd, which would be unfun at lower levels. Considering this, reducing the underwater speed allows counterplay potential without heavily affecting lower level gameplay: Aeon and Cybran destroyers can catch Seraphim destroyers that overcommitted, and all destroyers (particularly UEF) can flee from submerged Seraphim destroyers and get out of torpedo range.

The submerged destroyer is slightly slower than unsubmerged destroyers.

## Checklist


Rhaelya (approved):
Everything looks in order.

PR #6590, opened 2024-12-18
Reduce Aeon and Seraphim T2 PD damage

### LINKED ISSUE #6554: [Balance Idea]: T2 PD adjustments
### Intended Change

Uttaushala (Sera T2 PD): 
Dmg 605->550
Oblivion(Aeon T2 PD):
Dmg 600->560 

### Reasoning

Both Aeon and Sera PDs have higher DPS/mass than UEF PD. Aeon also has higher muzzle velocity while Sera has a beam weapon.
T2 PDs are generally very strong, especially in the later stage of the game so we would rather nerf the best PDs rather than buffing the weaker ones. 

### Expected behavior

The damage stats for both PDs are adjusted accordingly.

lL1l1:

Closes #6554
- Seraphim: 605 -> 550 damage total
  - Beam lifetime preserved for sound effect compatibility (not that it would be very noticeable to be fair).

The point defenses do the correct amount of damage.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6590.md
+++ b/changelog/snippets/balance.6590.md
@@ -1,0 +1,8 @@
+- (#6590) Reduce the damage of Aeon and Seraphim T2 PD, as they have better DPS/mass than the UEF PD, and the Aeon one has higher muzzle velocity while the Seraphim one has a beam weapon.
+
+  - Aeon T2 PD:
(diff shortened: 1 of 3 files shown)


Rhaelya (approved):
Everything looks in order, stat changes are as they should be.

PR #6591, opened 2024-12-18
Buff UEF ACU Drone's rebuild costs

### LINKED ISSUE #6553: [Balance Idea]: Balance C-D1 rebuild cost
### Intended Change

C-D1 drone:
Make the rebuild cost of the drone half of the upgrade cost.

Mass cost: 160->60
Energy cost: 1600->800
Time cost: 100->200

### Reasoning

The upgrade is made in 40 seconds at 3 m/s drain. Rebuilding is done in 10 seconds at 16 m/s drain. That's inconsistent and counter-intuitive for a unit that isn't even present in the client unit database. Rebuilding also costs 1600 energy versus 2400 for an upgrade. A sudden unexpected E-drain spike of 160e/s can ruin a player's eco-balance. No one would reasonably expect the rebuild to drain 2.66x more than the upgrade.

It costs 120 mass to build a drone. It also costs 160 mass to rebuild a drone. This is the only unit in the game that is more expensive to rebuild than build. This was an accidental leftover when the upgrade cost was changed but the cost of rebuilding the drone wasn't adjusted alongside it.

### Expected behavior

The rebuilding costs of both left and right drones are correctly adjusted in the script file.

lL1l1:

- Closes https://github.com/FAForever/fa/issues/6553.
   - "Time cost: 100->200" is interpreted as coming from the default 10 buildpower being multiplied with the 10 seconds of the economy event, so that's how the 100 build time exists despite the actual build time in the drone blueprint being 400.
- The drone build time is now affected by the ACU's buildpower.

Spawn ACUs, upgrade the drone upgrades and various build suite levels. Kill the drones so that they rebuild.
Drones rebuild at a reasonable consumption at T1, and speed up when rebuilt with T2/T3 upgrades. The consumption at higher speeds is still less than any available structures, so the economy being unable to support the consumption rate is not an issue.
```
   CreateUnitAtMouse('xrb0304', 0,   -1.34,   -6.15,  0.00000)
   CreateUnitAtMouse('xrb0304', 0,   -1.34,   -4.15,  0.00000)
   CreateUnitAtMouse('xrb0304', 0,   -0.34,    5.85,  0.00000)
   CreateUnitAtMouse('uel0001', 0,    3.94,    0.48, -0.04558)
   CreateUnitAtMouse('uel0001', 0,    2.90,   -4.92, -2.31662)
   CreateUnitAtMouse('xab1401', 0,   -7.34,    1.85,  0.00000)
   CreateUnitAtMouse('uel0001', 0,    5.89,    3.61, -0.76593)
   CreateUnitAtMouse('xrb0304', 0,   -0.34,    3.85,  0.00000)
   CreateUnitAtMouse('xrb0304', 0,   -0.34,    1.85, -0.00000)
   CreateUnitAtMouse('xrb0304', 0,   -0.34,   -0.15,  0.00000)
   CreateUnitAtMouse('xrb0304', 0,   -1.34,   -2.15,  0.00000)
```

## Additional context

- After testing the rebuilding during a stall, it seems like `CreateEconomyEvent` works in blocks of buildtime in the same way as silos. Reworking the drone rebuild process to be the same as the kennel's (`TPodTowerUnit`) would fix this.
- The Drone auto-rebuild toggle is still using the very old gpg system of removing the drone enhancement when the drone dies. To implement it correctly there would have to be some UI data set on the ACU to create a "rebuild drone" button when the drone is destroyed but not rebuilding, and then that button would issue a sim callback or script task order that initiates the drone rebuild sim-side.

## Checklist


lL1l1:
@clyfordv I think you would be interested in the various possible implementations for improving drone rebuilding, which is related to this PR. I did mention that kennels have their own special way of building drones, but there's also external factories and unit upgrades which might be interesting to make it possible to assist the drone reconstruction and even display the rebuild progress separately from the ACU.

Rhaelya (approved):
Everything seems in order.

PR #6594, opened 2024-12-18
Replace self:GetLauncher()

MrRowey:
self:GetLauncher() > self.Launcher

## Checklist

### DIFF
--- a/lua/sim/CollisionBeam.lua
+++ b/lua/sim/CollisionBeam.lua
@@ -295,7 +295,7 @@ CollisionBeam = Class(moho.CollisionBeamEntity) {
         --  'Shield'
 
         if impactType == 'Unit' or impactType == 'UnitAir' or impactType == 'UnitUnderwater' then
-            if not self:GetLauncher() then
+            if not self.Launcher then
                 return
             end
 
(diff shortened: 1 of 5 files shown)


lua/sim/CollisionBeam.lua
```diff
@@ -295,7 +295,7 @@ CollisionBeam = Class(moho.CollisionBeamEntity) {
         --  'Shield'
 
         if impactType == 'Unit' or impactType == 'UnitAir' or impactType == 'UnitUnderwater' then
-            if not self:GetLauncher() then
```
Garanas, line 298: Have you tested this? I do not think the launcher is stored in the collision beam. And therefore `self.Launcher` may always be nil.

MrRowey: tbh i just did fined and replace with vscode

Garanas: The approach is fine, but you need to verify the changes. Did you do the same for https://github.com/FAForever/fa/pull/6592 and https://github.com/FAForever/fa/pull/6593? 

If so, please test and verify the changes. If you do not have the time to do so it is better to close the pull requests.

PR #6597, opened 2024-12-22
Repr improvement

lL1l1:

More FAF-specific improvements to the inspect library:
- Print the base class of a table, if any.
- Don't print `__index` fields if they're the same as the table. This removes a lot of common recursive clutter (it is still possible to recurse, for example the `Brain.CDR.Brain` or `WorldView.view.WorldView` in the example below).
- Print source for lua functions in a rather short local format.

Use this command on a Percival:
```
SimLua LOG(repr(SelectedUnit(), {meta=true}))
```
[Example Output (Percival).txt](https://github.com/user-attachments/files/18221230/Example.Output.Percival.txt)

Use this command:
```
UI_Lua LOG(repr(import('/lua/ui/game/worldview.lua').GetWorldViews()['WorldCamera'], {meta = true}))
```
[Example Output (main WorldView).txt](https://github.com/user-attachments/files/18221269/Example.Output.main.WorldView.txt)

## Additional context

The [last version](https://github.com/FAForever/fa/blob/5df582851a64ac8cc7545dab243ec6935c495605/lua/system/repr.lua) of the original repr printed class-specific data. I was lazy on implementing that, but it could be done by using a helper function instead of directly the names table. Still, in general these changes lay the foundations for class-specific data printing.

## Checklist


Garanas:
@maudlin27 / @relent0r you both use this function too. It may be interesting for you to look at these changes and share your thoughts.

lua/system/repr.lua
```diff
@@ -47,12 +47,18 @@ local tostring = tostring
 local rep = string.rep
 local flr = math.floor
 local match = string.match
+local substr = string.sub
 local gsub = string.gsub
 local fmt = string.format
 local _rawget = rawget
 local type = type
+local getmetatable = getmetatable
```
4z0t, line 55: Like this diversity in variable naming. It was discussed 3 years ago and we still mix this and that.

lua/system/repr.lua
```diff
@@ -182,11 +189,120 @@ function Inspector:getId(v)
...
+    mesh_methods = "mesh_methods",
+    histogram_methods = "Histogram",
+    PathDebugger_methods = "PathDebugger",
+
+    -- Shared class implemented in Lua
+    Trashbag = "Trashbag",
+}
+
+---@param v table | fa-class | fa-class-state
+local function fafTableToString(v)
```
4z0t, line 288: `ObjectToString` sounds more reasonable.

lua/system/repr.lua
```diff
@@ -223,23 +339,31 @@ function Inspector:putValue(v)
...
+            local checkMeta
             if self.meta then
-                if type(mt) == 'table' and not table.empty(mt) then
-                    if seqLen + keysLen > 0 then puts(buf, ',') end
-                    tabify(self)
-                    puts(buf, '<metatable> = ')
-                    self:putValue(mt)
-                end
+                mt = getmetatable(t)
+                checkMeta = type(mt) == 'table' and not TableEmpty(mt)
```
4z0t, line 354: Metatable can be either table or nil. Shouldn't you just check against nil?

4z0t (changes requested):

lua/system/repr.lua
Garanas: It's a table or an empty table. I've never seen it being nil.

4z0t: Actually it can be other types if `__metatable` is set...

lL1l1: __metatable must've been why the original library had it check the type of the the metatable.

lua/system/repr.lua
```diff
@@ -43,16 +43,23 @@ local Inspector_mt = { __index = Inspector }
...
-local rep = string.rep
-local flr = math.floor
-local match = string.match
-local gsub = string.gsub
-local fmt = string.format
-local _rawget = rawget
+local rawget = rawget
 local type = type
+local getmetatable = getmetatable
+local debugGetInfo = debug.getinfo
```
4z0t, line 56: If you make upvalued functions of certain libraries as Pascal case then follow it.

4z0t (changes requested):

lua/system/repr.lua
lL1l1: Do you have a link to the discussion? I haven't seen it.

Garanas:
These changes are fine in general. It appears however that calling `reprsl` to a unit generates a very large print. What if we by default skip (recursively) printing tables with a metatable, with the exception of course of the table that we're printing initially?

The reason is that at the moment the Brain reference is recursively printed too. And that can hold a lot of information, especially for AIs.

relent0r:
For the uninitiated what does this do for us exactly compared to the repr function now?

lL1l1:
Maybe it is better to skip only brains and blueprints (if they're not the root), and create a flag for that specifically? There could be a wide variety of tables with metatables that we'd want to avoid.

Alternatively the proper solution would be to be like deepcopy and have a "backrefs" table that tells us whether we've seen a table or not before.

lua/system/repr.lua
```diff
@@ -223,23 +341,31 @@ function Inspector:putValue(v)
                         puts(buf, "]")
                     end
                     puts(buf, ' = ')
-                    self:putValue(t[k])
+                    if k == "__index" and tostring(t[k]) == tostring(t) then
+                        puts(buf, StringFormat("{...} -- %s (%g bytes)", 'table (self): ' .. string.sub(tostring(v), 8), DebugAllocatedSize(t)))
```
PreciseBump38, line 345: `StringSub` instead of `string.sub`?

PreciseBump38:
I like these changes.

@Garanas makes a good point about repr size for units, as evidenced by the 1k+ line Percival output.  For me the preference would be to have the default behaviour skip the Blueprint and AIBrain tables by default with an option to reenable - I would worry a bit about skipping things with metatables because that feels like it could have unintended consequences.

The backref tracking suggestion sounds great and a simple addition that would bring some benefit.

Another thing I wondered about was implementing the ability to skip tables where they have already been repr'ed starting from a lower initial depth.  The idea is that tables appearing multiple times (like aibrain) can be repr'ed once and then referenced from elsewhere just by the table address.  This would be more complicated to add though - and would require implementing breadth first searches to identify duplicate entries, so I'd probably lean against even trying it (unless someone can think of a smarter way to do it).

PR looks good enough as is to me though, and I'd be in favour of merging and raising follow on work (BP/AI brain skipping, backref tracking) as issues that can be addressed in a future PR.  Lets not let perfect be the enemy of the good.

PreciseBump38 (approved):

BlackYps:
Are there anymore changes planned or is the conclusion to merge as is?

PR #6600, opened 2024-12-25
Annotate some of the functionality related to worldviews

Garanas:

Annotates the functionality related to world views. Documents to some degree how secondary worldviews (that run on secondary adapters) work. 

Not necessary, these are just changes to annotations.

## Additional context

In preparation of a feature for HintHunter to toggle the split screen.

## Checklist

### DIFF
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -481,7 +481,7 @@ end
 function GetMovieVolume()
 end
 
---- Returns the current number of root frames. There is usually only one root frame for each adapter (monitor). This is often referred to as a 'head' in other comments.
+--- Returns the current number of root frames. There is usually only one root frame for each adapter (monitor). This is often referred to as a 'head' in other comments. The game supports up to two root frames.
 --- 
 --- In the game options you can add a second adapter under the 'Video' tab.
 ---
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -485,7 +485,7 @@ end
 --- 
 --- In the game options you can add a second adapter under the 'Video' tab.
 ---
---- See also `GetFrame(0)` and `GetFrame(1)` to retrieve the root frame for a given adapter. 
+--- See also `GetFrame(0)` to retrieve the root frame of the primary adapter and `GetFrame(1)` to retrieve the root frame of the secondary adapter. 
 --- See also the following modules that manage these frames:
 --- - Primary adapter: lua\ui\game\worldview.lua
 --- - Secondary adapter: lua\ui\game\multihead.lua
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -386,9 +386,17 @@ end
 function GetFireState(units)
 end
 
---- Returns the root UI frame for a given head
+--- Returns the root UI frame for a given adapter. You can use `GetFrame(0)` to retrieve the primary adapter. And you can use `GetFrame(1)` to retrieve the secondary adapter. 
+---
+--- In the game options you can add a second adapter under the 'Video' tab.
+--- 
+--- See also `GetNumRootFrames()` to determine the number of root frames. 
+--- 
+--- See also the following modules that manage these frames:
+--- - Primary adapter: lua\ui\game\worldview.lua
+--- - Secondary adapter: lua\ui\game\multihead.lua
 ---@param head number
----@return Frame
+---@return Frame | nil
 function GetFrame(head)
 end
 
@@ -473,7 +481,14 @@ end
 function GetMovieVolume()
 end
 
---- Returns the current number of root frames (typically one per head)
+--- Returns the current number of root frames. There is usually only one root frame for each adapter (monitor). This is often referred to as a 'head' in other comments.
+--- 
+--- In the game options you can add a second adapter under the 'Video' tab.
+---
+--- See also `GetFrame(0)` and `GetFrame(1)` to retrieve the root frame for a given adapter. 
+--- See also the following modules that manage these frames:
+--- - Primary adapter: lua\ui\game\worldview.lua
+--- - Secondary adapter: lua\ui\game\multihead.lua
 ---@return number
 function GetNumRootFrames()
 end
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -395,7 +395,7 @@ end
 --- See also the following modules that manage these frames:
 --- - Primary adapter: lua\ui\game\worldview.lua
 --- - Secondary adapter: lua\ui\game\multihead.lua
----@param head number
+---@param head 0 | 1
 ---@return Frame | nil
 function GetFrame(head)
 end
--- a/lua/ui/game/multihead.lua
+++ b/lua/ui/game/multihead.lua
@@ -50,6 +50,7 @@ end
 --- This function is referenced directly by the engine.
 ---@return nil
 function CreateSecondView()
+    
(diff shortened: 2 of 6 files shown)


Garanas:
There's some interesting possibilities here with world views. It's sad that the original interface is not friendly towards refactoring. For example, the value `viewLeft` is referenced directly instead of via a function. Because of that we can't refactor it blindly.

lua/ui/game/worldview.lua
```diff
@@ -12,14 +12,28 @@ local Bitmap = import("/lua/maui/bitmap.lua").Bitmap
...
----@type { [string]: WorldView }
+--- A list of all available worldviews.
+---@type table<string, WorldView>
 MapControls = {}
 
+--- A group that covers the entire screen.
+---@type Group | false
 view = false
+
+--- Primary view, if in split screen this is the left view. The left view is always visible.
```
lL1l1, line 28:
```suggestion
--- Primary view, and if in split screen this is the left view. The left view is always visible. 
```

lua/ui/game/worldview.lua
```diff
@@ -12,14 +12,28 @@ local Bitmap = import("/lua/maui/bitmap.lua").Bitmap
...
+---@type table<string, WorldView>
 MapControls = {}
 
+--- A group that covers the entire screen.
+---@type Group | false
 view = false
+
+--- Primary view, if in split screen this is the left view. The left view is always visible. 
+--- 
+--- Most features that rely on a worldview only work for the left worldview. One example is the reclaim overlay.
```
lL1l1, line 30: This is just a lua implementation detail, I don't think it is necessary in the documentation, as it is absolutely possible to have multi-view UI elements.

Hdt80bro (approved):

engine/User.lua
```diff
@@ -473,7 +481,14 @@ end
 function GetMovieVolume()
 end
 
---- Returns the current number of root frames (typically one per head)
+--- Returns the current number of root frames. There is usually only one root frame for each adapter (monitor). This is often referred to as a 'head' in other comments.
```
lL1l1, line 484: > There is usually only one root frame for each adapter (monitor).

This makes it sound like there can be more than two monitors used for the game, although it only supports 2.

engine/User.lua
```diff
@@ -473,7 +481,14 @@ end
...
+--- 
+--- In the game options you can add a second adapter under the 'Video' tab.
+---
+--- See also `GetFrame(0)` and `GetFrame(1)` to retrieve the root frame for a given adapter.
```
lL1l1, line 488: Should specify that 0 is the primary adapter and 1 is the secondary adapter.

lua/ui/game/multihead.lua
```diff
@@ -1,30 +1,69 @@
...
 local LayoutHelpers = import("/lua/maui/layouthelpers.lua")
 local Group = import("/lua/maui/group.lua").Group
-local GameCommon = import("/lua/ui/game/gamecommon.lua")
 local Bitmap = import("/lua/maui/bitmap.lua").Bitmap
 
+---@type WorldView | false
 view = false
 
+--- Shows the logo in the secondary adapter.
+---@return nil
```
lL1l1, line 37:
```suggestion
```

lua/ui/game/multihead.lua
```diff
@@ -1,30 +1,69 @@
...
 
     local rootFrame = GetFrame(1)
     local bg = Bitmap(rootFrame, UIUtil.UIFile('/marketing/splash.dds'))
     LayoutHelpers.FillParentPreserveAspectRatio(bg, rootFrame)
 end
 
+--- Creates a world view on the secondary adapter. The worldview is registered and available in the world view manager.
+---
+--- This function is referenced directly by the engine.
+---@return nil
```
lL1l1, line 56:
```suggestion
```

lua/ui/game/worldview.lua
```diff
@@ -298,9 +321,10 @@ function UnlockInput()
     end
 end
 
--- this function is called by the engine so it can not be removed (its logic could be changed though)
+--- This function is called by the engine so it can not be removed (its logic could be changed though)
```
lL1l1, line 324:
```suggestion
--- This function is called by the engine so it cannot be removed (its logic could be changed though)
```

lua/ui/game/worldview.lua
```diff
@@ -131,6 +150,10 @@ function MarkStartPositions(startPositions)
     end
 end
 
+--- Creates the world view on the primary monitor.
+---@param parent Control
+---@param mapGroup Control
+---@param mapGroupRight Control | nil   # if provided, creates a split view with a separate world view in both map groups
```
lL1l1, line 156:
```suggestion
---@param mapGroupRight? Control # if provided, creates a split view with a separate world view in both map groups
```

lua/ui/game/worldview.lua
Garanas: Yes it is, but it's an intentional choice that we do not. For example, the reclaim overlay would be too expensive.

engine/User.lua
Garanas: Processed in: https://github.com/FAForever/fa/pull/6600/commits/15e6b7c947a4eb001005dee3bf3178c247463bed

engine/User.lua
Garanas: See also: https://github.com/FAForever/fa/pull/6600/commits/0a61a3b39529f660cae471a03a090f4957295a41

lua/ui/game/multihead.lua
```diff
@@ -1,30 +1,67 @@
...
 
+---@type WorldView | false
 view = false
 
+--- Shows the logo in the secondary adapter.
 function ShowLogoInHead1()
-    if GetNumRootFrames() < 2 then return end
+
+    ---------------------------------------------------------------------------
+    -- defensive programming
```
4z0t, line 40: Cluttering comments.

lua/ui/game/multihead.lua
```diff
@@ -1,30 +1,67 @@
...
 end
 
+--- Creates a world view on the secondary adapter. The worldview is registered and available in the world view manager.
+---
+--- This function is referenced directly by the engine.
 function CreateSecondView()
-    if GetNumRootFrames() < 2 then return end
-    
+
+    ---------------------------------------------------------------------------
```
4z0t, line 57: Same.

lua/ui/game/worldview.lua
lL1l1: It's good to document it as a ton of UI Lua being coupled to this specific worldview, but I think what stands out to me with this specific comment is that "features that rely on a worldview only work for the left worldview" sounds as if the engine depends on `viewLeft`, so I'd change it to something that specifies the Lua coupling more specifically.

lua/ui/game/multihead.lua
Garanas: Resolved in https://github.com/FAForever/fa/pull/6600/commits/45831abcc1ec668f727bf4aebab8d1de7aa9ba9b

lua/ui/game/multihead.lua
Garanas: Resolved in https://github.com/FAForever/fa/pull/6600/commits/45831abcc1ec668f727bf4aebab8d1de7aa9ba9b

PR #6601, opened 2024-12-26
Include the projectile in the OnMissileIntercepted callback

maudlin27:
As certain information may be included on the projectile that is of interest to AI when the missile is intercepted

Adds the projectile to the onmissileintercepted callback, for if the projectile contains information of use to the AI.

Did a destructive hook of the code change to check it provided the desired functionality.

## Additional context
The target position provided from the callback is unreliable; this way when a projectile is created I can record against the projectile the intended unit target, and then access this when the intercepted callback is triggered.

### DIFF
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -309,7 +309,7 @@ Projectile = ClassProjectile(ProjectileMethods, DebugProjectileComponent) {
         -- callbacks for launcher to have an idea what is going on for AIs
         local launcher = self.Launcher
         if not IsDestroyed(launcher) then
-            launcher:OnMissileIntercepted(self:GetCurrentTargetPosition(), instigator, self:GetPosition())
+            launcher:OnMissileIntercepted(self:GetCurrentTargetPosition(), instigator, self:GetPosition(), self)
 
             -- keep track of the number of intercepted missiles
             if not IsDestroyed(instigator) and instigator.GetStat then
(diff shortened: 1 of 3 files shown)


maudlin27:
If any issues with this or want to discuss please ping me on discord; I've already covered the scenario with a destructive hook for M28 but figured it'd be worth adding to core functionality to be of benefit more generally

Garanas:
@relent0r I think we spoke of this a few days ago, what are your thoughts about this?

lua/sim/Projectile.lua
Garanas, line 1: Please also update the definition in the unit class to include the additional parameter.

relent0r:
> @relent0r I think we spoke of this a few days ago, what are your thoughts about this?

If its what I think it is its a little different. Maudlin found that when the Aeon TMD intercepted missiles the position being returned by the callback was the TMD rather than the target position. I believe this was a way that Maudlin could track what happened to the missile.

lua/sim/Projectile.lua
maudlin27: I've updated Unit.lua's OnMissileIntercepted, is that what you meant by this?

lua/sim/Unit.lua
```diff
@@ -5083,7 +5083,8 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
     ---@param target Vector
     ---@param defense Unit Requires an `IsDestroyed` check as the defense may have been destroyed when the missile is intercepted
     ---@param position Vector Location where the missile got intercepted
-    OnMissileIntercepted = function(self, target, defense, position)
+    ---@param self Unit
+    OnMissileIntercepted = function(self, target, defense, position, self)
```
4z0t, line 5087: second `self` replaces the first one. Shouldn't it be `missle`?

4z0t (changes requested):

lua/sim/Unit.lua
maudlin27: yes thanks, I've updated

4z0t (approved):

PR #6602, opened 2024-12-28
Fix : Remove unused personality calls in default skirmish AI code

relent0r:
This PR removed unused code from the Skirmish AI factory and platoon former manager.
The personality function calls were present in both the factory manager and platoonformer manager.

In the factory manager they created a ptnsize variable that is unused.

In the platoonformer manager that are used twice in CanFormPlatoon and FormPlatoon as the platoon count parameter. I validated that in the default AI this is always returning 1. So I've removed the function calls and replaced with a static integer of 1.

Tested that this change has no bearing on the AI functionality, nor does it cause a desync when comparing an replay with and without the changes indicating it is doing nothing out of what has already been analyzed.

## Additional context
The functions that are being modified are used by the AI multiple times per second so while removing the function calls may not seem like much it is saving some compute and memory.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6602.md
+++ b/changelog/snippets/fix.6602.md
@@ -1,0 +1,1 @@
+- (#6602) Removed unused personalty function calls in default AI factory manager and platoon former manager.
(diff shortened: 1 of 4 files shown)


relent0r:
@Dhomie These functions are also used in the campaign AI but I'm uncertain if they do anything. My understanding is that the personality system was never completed and required personality templates to be used (there was an example of this in an old piece of demo code in aipersonality.lua but to my knowledge this was never used by scfa.

Dhomie:
It defaults to 1, I haven't encountered where it might be a different value.
The campaign AI is meant to be a blank template that the mission creator then customizes in the map files.
All in all this should be fine.

Garanas (approved):

PR #6603, opened 2024-12-28
Increase novax HP so it survives 1 mavor shell

lL1l1:

While the projectile simulation is cool, it can be game deciding for Novax satellites to be destroyed by essentially random artillery fire. Increasing the HP so it survives at least one shell from any artillery reduces this unfair randomness while keeping a cool aspect of the simulation.
This is a tradeoff between immersion and gameplay balance.

- Novax satellite:
  - HP: 100 -> 16500

It has more HP than any one artillery shell.
Mod compatibility (specifically BlackOpsFAF Unleashed) seems to be fine as the satellite has the `UNTARGETABLE` category.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6603.md
+++ b/changelog/snippets/balance.6603.md
@@ -1,0 +1,4 @@
+- (#6603) Increase Novax HP so that it can survive at least 1 of any artillery shell, reducing the impact of the random chance that artillery hits your novax while it is flying over to the artillery.
+
+  - Novax satellite:
+    - HP: 100 -> 16500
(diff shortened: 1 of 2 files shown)


Hdt80bro:
I think it'd make more sense for it to be unhittable.

Hdt80bro:
While I agree that the simulation is cool, I still think that the Novax shouldn't be able to be hit at all if immersion is the concern - as an orbital weapon, it should have an elevation a 100 times what it is now. It's only been compressed to make it usable for the UI.

BlackYps:
Rejected by the balance team lead

PR #6606, opened 2024-12-28
Sniper bot buff

lL1l1:
## Issue
Snipers bots are too micro-heavy and do not get enough value for their cost so they rarely get built. They used to be a powerful choice for beating Percivals/Bricks, but now they're not even viable and only firebases and T4s remain viable options.

- Code fixes for the Seraphim sniper
  - This includes a nerf: the turret yaw speed goes from 90 to 70 (normal mode) / 50 (sniper mode)

### Both snipers:
- Firing tolerance: set to 0

  This fixes them consistently missing on the first shot after turning/retargeting, especially for the Seraphim sniper which now has lower yaw speed than turn speed. For reference, hitting a size 1 brick at 70 range requires 0.41 degrees firing tolerance.
- Range: Seraphim: 55/65 -> 60/70; Aeon: 60 -> 65

  This partially reverts the range nerf from the most recent changes to snipers. The cost changes solved most of the issues, and the range nerf introduced too much micro and too little range to deal damage to enemy units before they catch up, so the experience was made worse. The range is only partially reverted so that Seraphim sniper mode does not outrange Ravagers (70 range).
  
- Seraphim sniper speed in normal mode swapped with Aeon sniper: Aeon: 2.4 -> 2.2; Seraphim: 2.2 -> 2.4
  
    This creates a more interesting interaction between Aeon and Seraphim snipers as they swap who outranges or outspeeds who, instead of Aeon always outspeeding Seraphim. Aeon has the very fast Absolver to counter Seraphim's alpha strike capability as well.

### Seraphim sniper:
- Sniper mode speed: 1.65 -> 1.8

  Since the sniper mode speed is coded as a multiplier, it was unnecessarily nerfed by the recent speed changes. 1.65 is slower than ACU (1.7), so it is clearly excessive, and made sniper mode even more difficult to use against fast mainline units.

- Turret Yaw speed: 90 -> 70 (normal mode) / 50 (sniper mode)

  This is due to the aforementioned fix in the weapon code. It further emphasizes a steady, long range sniping style for Seraphim in contrast to Aeon's quick turning and more mobile firing.

- The Seraphim sniper has a mini rework to make it more favorable to use the sniper mode regularly, instead of reserving it for alpha-striking experimentals and ACUs:
  - Normal mode and sniper mode DPS swapped: 153/132 -> 132/152 respectively
  - Sniper mode damage reduced, reload sped up: 2000 dmg/15.2s (132 DPS) -> 1570 dmg/10.3s (152 DPS)
  
    This reduces for how long the sniper slows down after firing and makes it overkill Harbingers and Othuums much less (30%/28% overkill -> 2%/0%).

## Checklist


lL1l1:
We decided to buff speed instead of range as it would be helpful in low apm/1v1 scenarios and because it would encourage more proactive back and forth gameplay.

We decided to tackle the sniper mode DPS and damage model changes in the next patch so we can have more time to discuss it.

Rhaelya (approved):
Everything looks in order.

PR #6607, opened 2024-12-31
Improvements to the functionality of a number of anti-torpedo weapons

Basilisk3:
Improve the functionality of a few anti-torpedo weapons. 

The addition of a `MinRadius` prevents the weapon from locking onto projectiles it cannot reasonably intercept anymore. `UseFiringSolutionInsteadOfAimBone` is now enabled for the Aeon Tech 3 Sonar and Seraphim Tech 1 Submarine since these weapons are turreted. This also prevents a bug that causes the anti-torpedo weapon of the Aeon Tech 3 Sonar to shoot in the opposite direction in some rare cases.

Additionally, audio queues are added to the Ajellu Anti-Torpedo Defenses of the Seraphim Tech 1 and Tech 3 subs.

**Aeon T3 Sonar Platform (UAS0305):**
- Quasar Anti Torpedo
  - MinRadius: 0 --> 5 
  - `UseFiringSolutionInsteadOfAimBone`: `false` --> `true`

**Sou-istle: T1 Attack Submarine (XSS0203):**
- Ajellu Anti-Torpedo Defense
  - MinRadius: 0 --> 5
  - `UseFiringSolutionInsteadOfAimBone`: `false` --> `true`
  - Audio queue added

**Yathsou: T3 Submarine Hunter (XSS0304):**
- Ajellu Anti-Torpedo Defense (x2)
  - MinRadius: 0 --> 10
  - Audio queue added

**Barracuda (XRS0204) and Megalith (XRL0403):**
- Anti-Torpedo Flare
  - MinRadius: 0 --> 5

The performance of the torpedo defenses is improved, greatly in some cases. The Quasar Anti Torpedo of the Aeon Tech3 Sonar Platform is fixed.

## Checklist


lL1l1:
> This also prevents a bug that causes the anti-torpedo weapon of the Aeon Tech 3 Sonar to shoot in the opposite direction in some rare cases.

Do you have specifics for those rare cases? The only reason I can think that it would fire backwards is that it doesn't have `LeadTarget = false`, but that should be false by default because its firing a projectile with tracking in its blueprint.

As for using the firing solution instead of the aim bone, I don't think being turreted is a valid enough reason, but the T1 sub and T3 sonar don't have a visible turret, so I guess it's fine to use the cheaty firing solution. I would increase the firing tolerance to 180 to truly make it fire instantly in any direction, since currently it is still limited by the turret's (fast) turn rate before it can fire.

Also, it would be worth investigating torpedo flares. The Barracuda, Salem, Cooper, and Valiant have minimum ranges of 0, 5, 10, and 15 respectively, and maximum range of 25, 30, 25, and 30 respectively. Neptune is 10-30 as well.

Basilisk3:
> Do you have specifics for those rare cases? The only reason I can think that it would fire backwards is that it doesn't have `LeadTarget = false`, but that should be false by default because its firing a projectile with tracking in its blueprint.

I tested this by spawning in a couple of HARMS and enabling `NoDamage`. The anti-torp weapon eventually bugs out and either refuses to shoot or shoots in the opposite direction. With "rare" I meant that in regular games, this usually should not happen all that often. In test games it is easily reproducible though. HARMS torpedoes are pretty fast, maybe that is causing it.

> As for using the firing solution instead of the aim bone, I don't think being turreted is a valid enough reason, but the T1 sub and T3 sonar don't have a visible turret, so I guess it's fine to use the cheaty firing solution. I would increase the firing tolerance to 180 to truly make it fire instantly in any direction, since currently it is still limited by the turret's (fast) turn rate before it can fire.

I meant that these weapons don't look like they are turreted, so visually, it's fine to change. Using the firing solution instead of the aim bone greatly increased the functionality, which is the actual reason I included it. Note that I didn't add it for the Yathsou because it would look off.

Increasing the firing tolerance to 180 is a good suggestion, I'll add that.

Garanas (approved):
Changes look good to me.

PR #6608, opened 2025-01-01
Fix text overflow in fixed width tooltips when using UI scaling

speed2CZ:
The mod manager is using some custom logic for the tooltips and its was not scaling the width of the tooltip causing text overflow on couple of the elements.

### DIFF
--- a/changelog/snippets/fix.6608.md
+++ b/changelog/snippets/fix.6608.md
@@ -1,0 +1,1 @@
+- (#6608) Fix text overflow in mods manager tooltips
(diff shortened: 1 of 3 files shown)


lL1l1 (commented):
What if the width is scaled in the base class? It's where this width value is going.
https://github.com/FAForever/fa/blob/7f5ce745e8308aff1975be44d6963fd1edebbd06/lua/ui/game/tooltip.lua#L298-L303

lL1l1:
Happens when UI scaling is used with fixed-width tooltips (such as activate favorites button, sort dropdown, mod website button, and github url button).
{5DE49450-C539-42F3-A891-4A9006B16120}

lL1l1 (approved):

PR #6609, opened 2025-01-02
Fix drones used in unit transfer to rebuild units not being destroyed

lL1l1:

Fixes a bug reported in [Discord](https://discord.com/channels/197033481883222026/1323455309661863936) and shown in replay 23962051 where rebuilder drones were not getting cleaned up after use.
This is caused by the `rebuilders` table being indexed by the index of `trackers`, but this indexing can be skipped if `trackers.success` is true, which causes `ipairs` to not completely iterate over `rebuilders`, causing some rebuilder drones to not be destroyed.

Launch replay 23962051 from the branch `deploy/faf` at commit 805bdd344 with the logging commit cherrypicked on top. It will take around 11 minutes in-game for the bug to show, so you can use `wld_RunWithTheWind`. 
The log will show that the tracker with the 2nd index fails, which causes the next iteration of rebuilders (for the next army) to create a rebuilder for tracker 2, but it doesn't iterate over `rebuilders = {[1] = nil, [2] = rebuilder}`, so rebuilder 2 is not destroyed.

By adding in the commits that remove the `ipairs` iterators, the same replay shows that the rebuilders are all destroyed every army iteration despite tracker 2 failing.

## Checklist


lua/SimUtils.lua
```diff
@@ -533,7 +533,7 @@ end
 ---@param army Army
 function TryRebuildUnits(trackers, army)
     local rebuilders = {}
-    for k, tracker in ipairs(trackers) do
```
Hdt80bro, line 536: How does this affect the outcome? The trackers are still created using `ipairs`.

lL1l1: It doesn't, I just removed it for consistent code style.

speed2CZ (approved):

PR #6620, opened 2025-01-14
Add feature to only dock damaged air units

Garanas:

Introduces the ability to dock the damaged air units in your current selection. This can either be achieved via a hotkey or by right-clicking the dock button in the UI.

Originates from a suggestion on our [Discord server](https://discord.com/channels/197033481883222026/13283730906771619940).

https://github.com/user-attachments/assets/a8ee642b-b513-4e80-945b-dde5999b2748

Have damaged air units in your selection. Use the hotkey and/or the UI.

## Checklist


lua/ui/game/hotkeys/dock-damaged.lua
```diff
@@ -0,0 +1,68 @@
...
+        local isDamagedSufficiently = health / maxHealth < ratio
+
+        if isAirUnit and canUseAirStaging and isDamagedSufficiently then
+            TableInsert(damaged, unit)
+        else
+            TableInsert(remaining, unit)
+        end
+    end
+
+    -- prevents losing command mode
```
lL1l1, line 54:
```suggestion
    -- Since `IssueUnitCommand` does not work with docking orders, use the selection-only `IssueDockCommand` function
    -- prevents losing command mode
```

lua/ui/game/orders.lua
```diff
@@ -280,11 +280,22 @@ end
...
+            IssueDockCommand(true)
+        end
     end
+
+    if modifiers.Right then
+        if modifiers.Ctrl then
+            import("/lua/ui/game/hotkeys/dock-damaged.lua").DockDamaged(0.75)
+        else
+            import("/lua/ui/game/hotkeys/dock-damaged.lua").DockDamaged(0.50)
+        end
```
lL1l1, line 296: The default sim behavior docks aircraft below 75% hp (`NeedRepairThresholdRatio` in console), so 50% and 75% are quite low values. I would set a 90% threshold because below 90% is when ASF-ASF and inty-inty shots to kill change.

I don't think a Ctrl modifier is needed since people won't be clicking mid-combat to dock damaged aircraft (like attack move would with 75%), and they just want a dock-damaged modifier to not dock their full HP units and waste time repairing the fleet to full HP.

The default value for the hotkey itself should be updated too.

lua/ui/game/hotkeys/dock-damaged.lua
```diff
@@ -0,0 +1,68 @@
...
+    SelectUnits(damaged)
+    IssueDockCommand(true)
+    SelectUnits(remaining)
+
+    -- prevents losing command mode
+    gameMain.SetIgnoreSelection(false)
+    commandMode.RestoreCommandMode(true)
+
+    -- inform user
+    print(string.format("Docking %i units", TableGetn(damaged)))
```
lL1l1, line 67: for style consistency
```suggestion
    print(string.format("Docking %d units", TableGetn(damaged)))
```

lL1l1 (changes requested):
The dock button's tooltip should be updated with the new behavior.

I think a fuel limit should be added to the button's behavior. The sim uses 20% fuel (Console `NeedRefuelThresholdRatio`) in patrol/attack move so I'd use the same. It would be practical for users.

lua/ui/game/orders.lua
Garanas: I was not aware of that threshold. I initially had it at 0.25 (25% hp remaining) but noticed that a lot of units that I would personally want to go and repair too did not go. In some cases no air craft would try to dock.

lL1l1: Do you want me to just set the default to 90% and remove the ctrl modifier for the button and then we can merge this in?

Garanas: Yeah, that works for me.

lL1l1 (approved):
@Garanas I made a few changes to better match how orders and order buttons normally work. Can you take a quick look and merge it?
Only thing is in the future someone will have to pick up localization because I can't decide on a tag for it right now. Otherwise it's all polished in my opinion.

Garanas:
Merged it in. I do feel that the threshold is too high for it to be useful this way. But we'll see.

lL1l1:
With the snowbally nature and extremely high density of air fights/ground attacks I wouldn't be surprised that a lot of units (ASF in particular) come out entirely undamaged.

PR #6621, opened 2025-01-14
Rework shader names and add biplanar shader

BlackYps:
There are several changes here:

- I noticed that we actually disabled being able to scale the slot that is supposed to store the macrotexture. This makes sense when only thinking about the scenario where you provide a custom map-wide texture. Then the user does not have to worry about providing the right scaling to make the texture fit neatly over the whole map. However, there are some gpg macrotextures that are supposed to tile. If we don't provide the ability to control the scale of this texture we are artificially limiting mapmakers, so I added this ability again.

- I reworked the naming scheme because I noticed that the previous naming scheme was too complicated and made it hard to remember what is what.

- I found a usage for previously unused texture channels in the Terrain003 (old name, new one is Terrain001) shader. This gives more artistic control and could produce interesting results if someone wants to make a stylized map.

- I added a new shader that uses biplanar mapping in two texture slots. Triplanar mapping is a well-known technique to minimize the problem of stretched textures. You can see an example in the beginning of this article here: https://bgolus.medium.com/normal-mapping-for-a-triplanar-shader-10bf39dca05a 
Actually I have to read this article properly at some point in the future to see if there are some easy improvements we could make. Thanks to @Garanas for linking it [here](https://github.com/FAForever/fa/pull/3585). This collection of links has been a great resource!
The shader only uses biplanar mapping because we omit the projection from the top. This is faster and as we want to use it for cliffs we don't need the top projection. The result looks like this:
grafik

- during the development of this shader I realized some peculiarities about texture coordinates so I did some small refactors to improve the code.

### DIFF
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -2080,7 +2080,7 @@ float4 TerrainPBRAlbedoPS ( VS_OUTPUT inV) : COLOR
 // Append a "B" to the name, if the shader uses Biplanar mapping.
 
 // ----------------------------------------------------------------------------
-//#region TerrainXP+, Terrain000 and Terrain050
+//#region TTerrainXPExt, Terrain000 and Terrain050
 
 // Layer| Albedo stratum                                               | Normal stratum
 //      | R           | G             | B            | A               | R             | G             | B             | A            |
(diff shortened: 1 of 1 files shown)


Garanas:
This is amazing @BlackYps , I wanted to built this years ago. But at the time I could not figure it out. I'm looking forward to this!

PR #6622, opened 2025-01-15
Fix Seraphim sniper's barrel not aiming with LOD1 model

lL1l1:

- Renames the vertex group/bone responsible for the barrel vertices from `lod1_Barrel` to `Barrel`

The barrel can be seen rotating when zoomed out:

{B222C165-5952-496E-BEB0-8CC8789D7C42}

Freecam can help hide the strategic icon if it obscures your view.

## Checklist


lL1l1:
I don't know how to verify all the other models or if there are any other known similar issues, since this was quite a human mistake that is likely to be repeated somewhere.

Also not sure where to document this behavior of turrets and LOD models.

Garanas (approved):
Delivers as promised. I also noticed this, but I had no idea how to fix it.

With regards to documentation, maybe this should be on the wiki. For example:

- https://wiki.faforever.com/en/Development/Modding/Creating-models

lL1l1:
I added a note about LoDs on that wiki page.

PR #6629, opened 2025-01-18
Check if instigator is destroyed in Fatboy `DeathThread`

lL1l1:
## Issue 
Since the instigator is used in a thread, and threads start 1 tick later, it can be passed from OnKilled and then get destroyed.
Replay: https://replay.faforever.com/24087043 Right side fatboy for south team at 29:00 errors out when dying
Error:
```
warning: Error running lua script: ...\gamedata\units.nx2\units\uel0401\uel0401_script.lua(263): Game object has been destroyed
         stack traceback:
         	[C]: in function `GetPositionXYZ'
         	...\gamedata\units.nx2\units\uel0401\uel0401_script.lua(263): in function <...\gamedata\units.nx2\units\uel0401\uel0401_script.lua:253>
```

- Check if the instigator was destroyed before calling `GetPositionXYZ` on it. I didn't opt for using `Unit.Dead` because this code won't run often and `.Dead` is true after dying, which is before C-object destruction.

- Running the replay from deploy/faf (805bdd34446cf8e8664ae74743c597cdb2ad56a4) with the fix commit cherry picked on top causes the fatboy to successfully die
- Checked that no other `DeathThread`s use the instigator value.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6629.md
+++ b/changelog/snippets/fix.6629.md
@@ -1,0 +1,1 @@
+- (#6629) Fix a rare error that would cause the Fatboy to not finish dying.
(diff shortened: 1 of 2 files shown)


Garanas (approved):
Nice catch!

PR #6633, opened 2025-01-20
Improve documentation of terrain types

BlackYps:
I had a closer look at the terrain types while I was working with the map editor and noticed that the documentation what they actually do is a bit lacking, so we should fix that.

I tried to understand how types that differ from the default are handled, but I have not fully figured it out. Apparently the game sometimes falls back onto the default terrain type if it can't find certain effects, but it also seems possible to remove default effects. @Hdt80bro you had a closer look at this all a while ago, can you provide some insights?

Somewhat related:
A discussion on discord: https://discord.com/channels/197033481883222026/1044974379232149535/1048015638502846535
A mod by balthazar to edit unit blueprints to use the feature: https://github.com/The-Balthazar/BrewLAN/tree/master/mods/BrewLAN_Plenae/ExpertCamo

## Summary by CodeRabbit

* **Documentation**
  * Renamed "Map Style List" heading to "Style Key List" in terrain docs.
  * Added a multi-line explanatory note clarifying an unused texture/mesh swap feature, its high memory cost, and low likelihood of being implemented.

### DIFF
--- a/changelog/snippets/other.6633.md
+++ b/changelog/snippets/other.6633.md
@@ -1,0 +1,1 @@
+- (#6633) Improve documentation of the terrain type features
--- a/lua/TerrainTypes.lua
+++ b/lua/TerrainTypes.lua
@@ -23,7 +23,7 @@
     200-219 reserved for snowy
     220-255 reserved for water
  
-- Map Style List (7)-
+- Style Key List (7)-
     Desert
     Evergreen
     Geothermal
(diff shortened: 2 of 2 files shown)


BlackYps:
And what is going on here? 
https://github.com/FAForever/fa/blob/fc5368a21a84eca4e0540350d35ac83c53455972/lua/TerrainTypes.lua#L119-L127

Hdt80bro:
Which effect table is used is handled [here](https://github.com/FAForever/fa/blob/develop/lua/sim/Unit.lua#L3612). Those block types are the four ways [this](https://github.com/FAForever/fa/blob/fb0457f9d0d1b0334d844ad70b6b152a640ab797/lua/sim/Unit.lua#L3647) function gets called. The FX tables have different key names of the same block type I annotated next to it (So the `LayerTerrainEffectType` tables all have key names of `Layer`, the `FXImpact` table has keys from `ImpactType`, etc).

BlackYps:
I found this again as we are currently talking about terrain types again. I'd like to merge this, as it is an improvement that can already stand on its own and I obviously didn't dive deeper into it.

Hdt80bro:
If we're only documenting the style keys, this is good to go. If you were hoping to document the rest of the items you had questions about, you can still do that. I'm unclear if you still plan to, however.

PR #6635, opened 2025-01-25
Enable counterintel by default for AI units

lL1l1:
Implements a [discord suggestion](https://discord.com/channels/197033481883222026/1133712738380677150).
- Replaces toggles like:
    ```lua
    --Turns Jamming off when unit is built
    self:SetScriptBit('RULEUTC_JammingToggle', true)
    ```
    with
    ```lua
    -- Don't turn off jamming for AI so that it uses it by default
    if self.Brain.BrainType == 'Human' then
        self:SetScriptBit('RULEUTC_JammingToggle', true)
    else
    end
    ```
- Formatting
- Annotations

All units work as expected when spawned for the player and the default civilian AI.
<details> <summary> Spawn command: </summary>

```lua
CreateUnitAtMouse('ura0401', 0,    1.93,  115.47,  0.00000)
CreateUnitAtMouse('ura0304', 0,   14.43,  117.47,  0.00000)
CreateUnitAtMouse('uea0304', 1,  -11.18,  -89.86,  0.00000)
CreateUnitAtMouse('uea0304', 1,  -22.74,  -49.33,  0.00000)
CreateUnitAtMouse('uea0304', 1,  -14.93,  -80.66,  0.00000)
CreateUnitAtMouse('xab1401', 0,    1.29,   99.40,  0.00000)
CreateUnitAtMouse('dra0202', 1,   -6.43,  -78.16,  0.00000)
CreateUnitAtMouse('ura0304', 1,   15.07,  -80.66,  0.00000)
CreateUnitAtMouse('xra0305', 1,   28.57,  -81.16,  0.00000)
CreateUnitAtMouse('ura0401', 1,    2.65,  -83.16,  0.16435)
CreateUnitAtMouse('xra0305', 0,   27.93,  116.97,  0.00000)
CreateUnitAtMouse('url0101', 0,  -25.07,  115.97,  0.00000)
CreateUnitAtMouse('ura0303', 1,   23.07,  -81.66,  0.00000)
CreateUnitAtMouse('xsb3201', 1,    3.12,  -67.09, -0.00000)
CreateUnitAtMouse('uea0304', 1,   -7.37,  -78.44,  0.00000)
CreateUnitAtMouse('xab1401', 1,    2.29,  -98.60, -0.00000)
CreateUnitAtMouse('url0101', 1,  -24.43,  -82.16,  0.00000)
CreateUnitAtMouse('uea0304', 1,   -9.34,  -60.59,  0.00000)
CreateUnitAtMouse('ura0303', 0,   22.43,  116.47,  0.00000)
CreateUnitAtMouse('dra0202', 0,   -7.07,  119.97,  0.00000)
CreateUnitAtMouse('uea0304', 0,  -15.57,  117.47,  0.00000)
CreateUnitAtMouse('xsb3201', 0,    1.29,   92.40,  0.00000)
```

</details>

## Checklist

### DIFF
--- a/changelog/snippets/ai.6635.md
+++ b/changelog/snippets/ai.6635.md
@@ -1,1 +1,1 @@
-- (#6635) Enable counterintel by default for units created by AIs. This includes Cybran aircraft stealth, Cybran land scout's cloak, and UEF strategic bomber's jamming.
+- (#6635) Enable counterintel by default for units created by campaign or skirmish (easy/medium/tech/rush/turtle/adaptive) AIs. This includes Cybran aircraft stealth, Cybran land scout's cloak, and UEF strategic bomber's jamming.
--- a/changelog/snippets/ai.6635.md
+++ b/changelog/snippets/ai.6635.md
@@ -1,0 +1,1 @@
+- (#6635) Enable counterintel by default for units created by AIs. This includes Cybran aircraft stealth, Cybran land scout's cloak, and UEF strategic bomber's jamming.
--- a/changelog/snippets/ai.6635.md
+++ b/changelog/snippets/ai.6635.md
@@ -1,1 +1,1 @@
-- (#6635) Enable counterintel by default for units created by campaign or skirmish (easy/medium/tech/rush/turtle/adaptive) AIs. This includes Cybran aircraft stealth, Cybran land scout's cloak, and UEF strategic bomber's jamming.
+- (#6635) Enable counterintel by default for units created by AIs. This includes Cybran aircraft stealth, Cybran land scout's cloak, and UEF strategic bomber's jamming.
--- a/engine/Enums.lua
+++ b/engine/Enums.lua
@@ -33,15 +33,15 @@
 ---| "RULEUCC_Invalid"
 
 ---@alias EngineToggleCap
----| "RULEUTC_ShieldToggle"
----| "RULEUTC_WeaponToggle"
----| "RULEUTC_JammingToggle"
----| "RULEUTC_IntelToggle"
----| "RULEUTC_ProductionToggle"
----| "RULEUTC_StealthToggle"
----| "RULEUTC_GenericToggle"
----| "RULEUTC_SpecialToggle"
----| "RULEUTC_CloakToggle"
+---| "RULEUTC_ShieldToggle"     # Bit 0
+---| "RULEUTC_WeaponToggle"     # Bit 1
+---| "RULEUTC_JammingToggle"    # Bit 2
+---| "RULEUTC_IntelToggle"      # Bit 3
+---| "RULEUTC_ProductionToggle" # Bit 4
+---| "RULEUTC_StealthToggle"    # Bit 5
+---| "RULEUTC_GenericToggle"    # Bit 6
+---| "RULEUTC_SpecialToggle"    # Bit 7
+---| "RULEUTC_CloakToggle"      # Bit 8
 
 ---@alias UnitMotionType
 ---| "RULEUMT_None"
--- a/engine/Sim/Entity.lua
+++ b/engine/Sim/Entity.lua
@@ -314,6 +314,7 @@ end
 function Entity:RemoveShooter(shooter)
 end
 
+--- Calls OnSelectionChanged if the unit is currently selected.
 function Entity:RequestRefreshUI()
 end
 
--- a/engine/Sim/Unit.lua
+++ b/engine/Sim/Unit.lua
@@ -617,9 +617,9 @@ end
 function Unit:SetRegenRate(rate)
 end
 
---- sets the script bit
----@param bit number | ToggleCap
----@param state boolean
+--- sets the script bit, which calls `OnScriptBitSet` or `OnScriptBitClear` for the Sim Unit and `OnSelectionChanged` on the User side if the bit is toggled
+---@param bit ToggleCap
+---@param state boolean # `true` -> `OnScriptBitSet`; `false` -> `OnScriptBitClear`
 function Unit:SetScriptBit(bit, state)
 end
 
--- a/engine/Sim/Unit.lua
+++ b/engine/Sim/Unit.lua
@@ -618,7 +618,7 @@ function Unit:SetRegenRate(rate)
 end
 
 --- sets the script bit
----@param bit number|string
+---@param bit number | ToggleCap
 ---@param state boolean
 function Unit:SetScriptBit(bit, state)
 end
@@ -700,6 +700,7 @@ end
 --- Tests if the unit has this toggle cap specified in the blueprint spec.
 --- May not always work.
 ---@param capName ToggleCap
+---@return boolean
 function Unit:TestToggleCaps(capName)
 end
 
--- a/lua/aibrains/adaptive-ai.lua
+++ b/lua/aibrains/adaptive-ai.lua
@@ -169,28 +169,6 @@ AIBrain = Class(StandardBrain, EconomyComponent) {
         end
     end,
 
-    --#region Unit Events
-
-    --- Called by a unit as it is finished being built
-    ---@param self AIBrainAdaptive
-    ---@param unit Unit
-    ---@param builder Unit
-    ---@param layer Layer
-    OnUnitStopBeingBuilt = function(self, unit, builder, layer)
-        StandardBrain.OnUnitStopBeingBuilt(self, unit, builder, layer)
-
-        -- enable counterintel
-        -- cloaking is ignored because it is expensive and uncommon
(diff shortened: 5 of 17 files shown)


relent0r:
@lL1l1 
Sorry if this is a very late comment since I'm asking for a possible change.

Rather than doing this at the unit level. Since its for AI only could we perhaps tackle at the aiBrain layer? Jip introduced callback triggers against the aiBrains for OnUnitStopBeingBuilt which I believe could tackle this.

A reason why this is a good approach is that we can make it only impact the AI's that we are interested in such as the default and campaign. 3rd party AI's handle their own counter intel units(which isn't really relevant since this just changes the on built state). There is a set of brain files that live under /lua/aibrains with one for each AI type.

Just a thought.

lL1l1:
So you want me to put the callback into `campaign-ai.lua` and `base-ai.lua`? I'm guessing this would work for enabling counterintel in all the coop missions, would it work for spawned units in skirmishes like survival maps?

relent0r:
> So you want me to put the callback into `campaign-ai.lua` and `base-ai.lua`? I'm guessing this would work for enabling counterintel in all the coop missions, would it work for spawned units in skirmishes like survival maps?



> So you want me to put the callback into `campaign-ai.lua` and `base-ai.lua`? I'm guessing this would work for enabling counterintel in all the coop missions, would it work for spawned units in skirmishes like survival maps?

I was just having a look as I've forgotten. I couldn't find a reference to it using the base-ai.lua file

In the OnCreateArmyBrain function I could see this.

```
    if (not info.Human) then
        local reference
        local keys = import("/lua/aibrains/index.lua").keyToBrain
        if (not info.Civilian) and (info.AIPersonality != '') then
            -- likely a skirmish scenario
            reference = keys[info.AIPersonality]
        else
            -- likely a campaign scenario
            if ScenarioInfo.type != 'skirmish' then
                reference = keys['campaign']
            end
        end

        if reference then
            local instance
            if not table.empty(getmetatable(reference)) then
                instance = reference
            else
                instance = import(reference[1])[reference[2]]
            end

            if instance then
                setmetatable(brain, instance)
            end
        end
    end
```
    
    I know the table here
    
    `keyToBrain = {
    -- default
    default = { "/lua/aibrains/base-ai.lua", "AIBrain" },
    campaign = { "/lua/aibrains/campaign-ai.lua", "AIBrain" },

    -- base AI
    tech = { "/lua/aibrains/tech-ai.lua", "AIBrain" },
    turtle = { "/lua/aibrains/turtle-ai.lua", "AIBrain" },
    rush = { "/lua/aibrains/rush-ai.lua", "AIBrain" },
    easy = { "/lua/aibrains/medium-ai.lua", "AIBrain" },
    medium = { "/lua/aibrains/medium-ai.lua", "AIBrain" },
    adaptive = { "/lua/aibrains/adaptive-ai.lua", "AIBrain" },
    random = { "/lua/aibrains/adaptive-ai.lua", "AIBrain" },

    -- base AIX
    techcheat = { "/lua/aibrains/tech-ai.lua", "AIBrain" },
    turtlecheat = { "/lua/aibrains/turtle-ai.lua", "AIBrain" },
    rushcheat = { "/lua/aibrains/rush-ai.lua", "AIBrain" },
    adaptivecheat = { "/lua/aibrains/adaptive-ai.lua", "AIBrain" },
    randomcheat = { "/lua/aibrains/adaptive-ai.lua", "AIBrain" },
}`

Provides all the key references. So I think the base-ai.lua is a base template rather than something that is used.

So in theory if we want all the default AI including campaign to use it we update all the brain files for the various AI.

In answer to your other question. Sadly no it won't apply to spawned units since they don't perform that callback. The callback is just for when a unit is built (I couldn't remember if it applies to just factories or if both factories and engineers).

If you want me to validate first I can run the changes locally here before you put in the effort.

lL1l1:
@relent0r I moved the counterintel enabling to the campaign AI and the default AIs. Can you review it please?

relent0r (dismissed):

relent0r:
> @relent0r I moved the counterintel enabling to the campaign AI and the default AIs. Can you review it please?

Looks good. 

@Dhomie fyi

Dhomie:
Yep, looks good indeed, although I do have a question.

When it comes to spawning units, doesn't `Unit:OnCreate()` get called?
I assume it's also called when the unit entity is created when it has just started building, but would it be possible to handle the intel/counter intel for spawned units in `Unit:OnCreate()` ?

relent0r:
> Yep, looks good indeed, although I do have a question.
> 
> When it comes to spawning units, doesn't `Unit:OnCreate()` get called? I assume it's also called when the unit entity is created when it has just started building, but would it be possible to handle the intel/counter intel for spawned units in `Unit:OnCreate()` ?

Yea OnCreate does get called. From an AI developers pov OnCreate would be better (its what m28/m27 use as the hook for assigning units to logic). I always assumed there was a valid reason why it was never used as a callback for ai. e.g what have a heavy platoon manager when you could just use oncreate callbacks.

Dhomie:
I second the idea of having the whole thing handled by `Unit:OnCreate()`, as it would simply deal with both spawning, and building cases. If there's no `AIBrain:OnUnitCreate(unit)` callback yet, it can be introduced for this use case, and the idea of keeping this QoL update to specific brains is still applicable.

As for why GPG did things this way, I think they preferred the idea of manually enabling these whenever the mission, and/or AI developers have seen it fit, or simply didn't have the time nor resources to seperate AI and player cases.
These intel features aren't enabled for players either by default, and I'd leave the player matter up to the balance, and senior staff.
However for AIs really, it's just a chore having to always manually enable the intel features on units.

relent0r:
> I second the idea of having the whole thing handled by `Unit:OnCreate()`, as it would simply deal with both spawning, and building cases. If there's no `AIBrain:OnUnitCreate(unit)` callback yet, it can be introduced for this use case, and the idea of keeping this QoL update to specific brains is still applicable.
> 
> As for why GPG did things this way, I think they preferred the idea of manually enabling these whenever the mission, and/or AI developers have seen it fit, or simply didn't have the time nor resources to seperate AI and player cases. These intel features aren't enabled for players either by default, and I'd leave the player matter up to the balance, and senior staff. However for AIs really, it's just a chore having to always manually enable the intel features on units.

I think for the sake of getting this implemented we should run with what we have as hooking OnCreate is something that requires a bigger discussion given the potential for impact where as this change is a known quantity.

I know Jip spent a bunch of time attaching callbacks to the aiBrain class so that we could leverage added functionality but I never thought to ask why he didn't add OnCreate as part of that work. Perhaps we could create an issue and get a wider discussion on it?

Dhomie:
> I think for the sake of getting this implemented we should run with what we have as hooking OnCreate is something that requires a bigger discussion given the potential for impact where as this change is a known quantity.
> 
> I know Jip spent a bunch of time attaching callbacks to the aiBrain class so that we could leverage added functionality but I never thought to ask why he didn't add OnCreate as part of that work. Perhaps we could create an issue and get a wider discussion on it?

Agreed, this might be an issue that could have further implications.

lL1l1:
@relent0r 
> I logged the callback and its firing, its detecting that the unit supports stealth and attempts to set the script bit but the unit itself never goes into stealth. I double checked the unit id and its passing the correct unit.
> 
> Thoughts?

In the unit scripts it runs the base function first and then the rest, so it's enabling the intel but re-enabling it in the unit script. Seems like the unit scripts need to be edited after all.

relent0r:
> @relent0r
> 
> > I logged the callback and its firing, its detecting that the unit supports stealth and attempts to set the script bit but the unit itself never goes into stealth. I double checked the unit id and its passing the correct unit.
> > Thoughts?
> 
> In the unit scripts it runs the base function first and then the rest, so it's enabling the intel but re-enabling it in the unit script. Seems like the unit scripts need to be edited after all.

ooooh ok I see what you mean. I didn't spot that the first time around. In that case I apologize from making you do a bunch of extra work. I guess if you revert to your previous commit I'll retest on that.

lL1l1:
> ooooh ok I see what you mean. I didn't spot that the first time around. In that case I apologize from making you do a bunch of extra work. I guess if you revert to your previous commit I'll retest on that.

I think a combination of the two is the best: the unit script only disables it for human units and it does nothing for the AI, leaving the implementation up to the AI brain callback. Keeping it in the script for humans makes it easy for balance team to really fine tune the gameplay experience per-unit, and having it in the callback for AI keeps all behavior in one place for AI devs.

relent0r:
Oh yea valid point. Sounds like a good middleground.

relent0r (approved):

relent0r:
ok have retested it and its working as expected.

PR #6637, opened 2025-01-26
Improve merge bp compatibility with removing fields

lL1l1:
## Issue
Merge blueprints cannot set fields in the base bp to `nil` since lua does not allow storing table keys with a value of nil. The alternative to doing that is to set those keys to `false`, but not all code works well with that.

Fix errors/bugs I encountered when modding SACUs, so code using blueprints and related to enhancements, enhancement presets, and weapons.

- Units with enhancement presets that have a `false` value do not error and the enhancements work properly.
- Setting a preset to `false` does not error.
- Setting weapon buffs to `false` doesn't break it.

## Additional context

An alternative could be to create a special value like `"__nil"` that `table.merged` treats as setting a field to `nil`.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6637.md
+++ b/changelog/snippets/fix.6637.md
@@ -1,0 +1,1 @@
+- (#6637) Improve merge blueprint compatibility with removal of enhancements.
--- a/changelog/snippets/fix.6637.md
+++ b/changelog/snippets/fix.6637.md
@@ -1,1 +1,1 @@
-- (#6637) Improve merge blueprint compatibility with removal of enhancements.
+- (#6637) Improve merge blueprint compatibility with removal of enhancements, enhancement presets, and weapon Buffs.
--- a/engine/Core.lua
+++ b/engine/Core.lua
@@ -4,6 +4,9 @@
 ---@class FileName: string, stringlib
 ---@operator concat(FileName | string): FileName
 
+---@class string: stringlib
+---@operator concat(FileName): FileName
+
 ---@class VectorBase
 ---@field [1] number    # x
 ---@field [2] number    # y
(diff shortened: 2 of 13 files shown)


lL1l1:
Another pain I encountered while modding is the `Weapons` table, which is an array and index 1 controls the AI. This solution seems okay-ish? At least I'm not using debug functions like I thought of initially, and it still works from inside a `.bp` file.
```lua
local oldModBp = ModBlueprints
---@param all_bps BlueprintsTable
ModBlueprints = function(all_bps)
    local origBp = all_bps.Unit['url0301']--[[@as UnitBlueprint]]
    local origWeapon = origBp.Weapon --[[@as WeaponBlueprint[] ]]
    table.insert(origWeapon, 1, { ... })

    oldModBp(all_bps)
end
```

lL1l1:
Also there is a bug with reloading merge bps in that the mesh gets deleted due to how mesh bps are extracted and defaulted, ignoring `__blueprints` when `merge = true`.

lL1l1:
> ```lua
> local oldModBp = ModBlueprints
> ---@param all_bps BlueprintsTable
> ModBlueprints = function(all_bps)
>     local origBp = all_bps.Unit['url0301']--[[@as UnitBlueprint]]
>     local origWeapon = origBp.Weapon --[[@as WeaponBlueprint[] ]]
>     table.insert(origWeapon, 1, { ... })
> 
>     oldModBp(all_bps)
> end
> ```

Has to be table assignment and not table.insert or else you get infinite copies.

lL1l1:
Actually you can't even use ModBlueprints from `*.bp` files since they're run with doscript and will hook ModBlueprints an additional time every time you reload. So your reloaded file has no guarantee for the execution order of `ModBlueprints` hooks within your own mod.

lL1l1:
Closing in favor of https://github.com/FAForever/fa/pull/6669

PR #6639, opened 2025-01-31
Strat idea

Spikey84:
Essentially this change causes strat bombers to consume a percentage of their fuel when dropping a bomb. With this change in mind, their attack damage and radii have increased. The intention is to give the strategic bombers a unique role compared to that of other air units: to strike important enemy targets and to help suppress high-tier land armies. This change also alters the ability for bombers to snipe high HP enemies with little recourse quickly. With these changes, a player with 1 uef strat can bomb an acu dealing quite a noticeable amount of damage, but they have to choose how many bombs to drop. If they take 1 pass the damage is quite small. If they take two, they must accept that their strat is now extremely slow and every vulnerable to a few inties. Of course it is still possible to spam a large quantity of bombers late game to deal with a high HP target, but now even if they are allowed to kill their target uncontested they cannot simply kill every valuable unit on the map. 

I understand this change may not seem intuitive. I am hesitant to suggest a change that alters a hitherto untouched mechanic. I honestly just expect it to make for some fun gameplay. I miss microing strats. Yet even back when I could rush a strat and kill every t2 mex on the map uncontested, it felt unfulfilling. I hope with the buff to reload time and air physics that strats will be far more fun to use, be useful to make, and be less overwhelming super late game. 

Everything is subject to change and I cant be bothered proof reading this lol
Working on this...

## Additional context
 This is a very WIP change. I do not want this to be merged.

## Checklist


lL1l1:
Your base branch should be `develop` and not `deploy/fafdevelop` so that the bot doesn't add commits.

lL1l1:
Closed in favor of #6535

PR #6641, opened 2025-02-01
Nerf Brick range and DPS

lL1l1:
## Issue
Relevant [Balance Team discord thread](https://discord.com/channels/197033481883222026/1334704579887235144).
The brick has become too powerful after its +2 range buff, in a way that it can quite well stand up to Percies in direct combat, on top of being far more versatile with their speed and ability to handle T1/T2 spam. Essentially Bricks became a better version of Percies. Cybran doesn't need this as they have two direct fire T4s, which UEF doesn't.

The purpose of the range buff was to help Bricks fight ACUs, since previously they had a 2 range advantage at 30 vs 32 range, which made it too difficult to stay out of range of ACUs, while Bricks are very weak against ACU overcharge because they have no shields.

Two side effects of the range buff were that Bricks became far more effective kiting against Othuums/Harbingers, and that Percies no longer had any range advantage to rely on when being chased by Bricks.
For Harbingers, Bricks at 32 range used to deal 2841 damage before the Harbinger got in range, and 34 range increased that to 3977 (+1136 or +40%). Othuums are not very relevant as they have extremely short range + lower speed so the matchup was not ever close, although they are worse off too.
For Percies, Percies with a range advantage of 2 were able to fire 3.3 shots (The initial shot out of range is extremely important: `3.3 shots while running = (2 distance / 0.2 speed difference) / (1 shot/4.3 seconds) + 1 initial shot`) before the Brick got in range if the Percy was running away while being chased. A Percy takes 6 shots to kill a Brick so this range being removed is a massive difference in this scenario.

- Reduce the range of the Brick by 1 (34 -> 33):
  - For Harbingers, this reduces the extra damage they take compared to the old balance by half (1136-> 568).
  - For Percies, it buys them time to shoot 2.16 shots before the Brick gets into range in a kiting situation. The range should be helpful in general for efficiency of formations too.
- Reduce the HP of the Brick 7500 -> 7250:
  - This slightly helps Harbingers/Othuums fight Bricks, although understandably the brick's HP is not as relevant when it is taking no damage with its longer range.
  - This stops Percy from massively overkilling the Brick (5 shots is 7250 damage, 6 is 8700), so the intended effect is to bring the aforementioned 2.16 shot advantage to 3.16, in an effort to match the old Percy vs Brick balance.
    - There is an argument that supporting units like T1 artillery, random bombers, or leftover Titans/T2 units can deal the 250 damage necessary to prevent overkill, but that is quite inconsistent and unsatisfying design-wise.
- Reduce the DPS of the brick 312.5 -> 300:
  - This is intended to reduce the extra damage dealt against Harbingers, from 568 after the range nerf to 432. The final result is 2841 -> 3272 damage dealt while kiting (+431 or +15%).
  - It will have effects in all areas, so it is a rather small DPS decrease.
- Due to the range decrease, increase the turn rate of the Brick from 68.75 (derived from `TurnRadius`: `180 / (pi * 2 radius / 2.4 speed) = 68.75`) to 75 so that the reaction to turn around when an ACU starts coming into range has the same reaction time given as before.
  - This will also slightly help the Brick pick better fights and retreat, which is fair enough considering the now nerfed stats (Just the DPS and HP changes alone are combining into a 7.2% HP*DPS power decrease).

In an ideal kiting scenario (both units start facing the correct direction, and are moving at full speed before anyone fires):
- New changes: a Percy now kills a brick with 3240 HP remaining
- Current balance: 200 HP remaining
- Old balance (32 range): 3325 HP remaining

In a scenario where the Percy approaches the brick:
- New changes: it kills the brick with 1920 HP remaining
- Current balance: 200 HP remaining
- Old balance: 450 HP remaining

The drastic difference is due to the removal of overkill. If there is 250 damage from supporting units, the difference would not be so large.

Unlike with a 2 range difference, the Percy cannot automatically stop out of range of the brick when on attack move/attack order.
- This is because the AI only starts stopping the unit once the target is already in range, and the Percy has a stopping distance of 1.1. Increasing the `MaxBrake` would help, but it would adversely buff its micro potential overall, and in a massive way, since decreasing the stopping distance from 1.1 to something more fitting the 1 range difference requires a large change of 2.2 to 3.025 (0.8 stopping distance) or more depending on how unfavorable the AI is behaving (target check interval, which is 1.1 seconds for the Percy by default due to its fire rate, also plays a huge role how far away the Percy stops).

In sandbox testing, there is a noticeable range where the Percy can walk around in and fire without being fired back at, which is important but I don't think it will play as strong of a role in real games.

7v7 scenario Bricks walking into Percies:
- New changes: 4 Percies survive easily.
- Current balance: 2 Percies survive, or they can lose without micro due to overkill
- Old balance: 3-4 Percies survive, but they can lose without micro due to overkill

<details> <summary> Test spawn units </summary>

```
   CreateUnitAtMouse('xrl0305', 1,   22.47,   -4.70, -1.60165)
   CreateUnitAtMouse('xrl0305', 1,   22.31,   -1.65, -1.61377)
   CreateUnitAtMouse('xel0305', 0,  -21.44,  -10.00,  1.54845)
   CreateUnitAtMouse('xrl0305', 1,   22.32,    9.69, -1.44405)
   CreateUnitAtMouse('xel0305', 0,  -22.43,    4.98,  1.51267)
   CreateUnitAtMouse('xel0305', 0,  -23.43,    8.03,  1.59532)
   CreateUnitAtMouse('xrl0305', 1,   22.51,    3.70, -1.60165)
   CreateUnitAtMouse('xel0305', 0,  -22.43,   -3.95,  1.63072)
   CreateUnitAtMouse('xel0305', 0,  -22.45,   -1.01,  1.52899)
   CreateUnitAtMouse('xel0305', 0,  -22.43,    2.09,  1.70456)
   CreateUnitAtMouse('xel0305', 0,  -22.42,   -6.99,  1.56046)
   CreateUnitAtMouse('xrl0305', 1,   22.50,    6.78, -1.60166)
   CreateUnitAtMouse('xrl0305', 1,   22.42,   -7.74, -1.60869)
   CreateUnitAtMouse('xrl0305', 1,   22.50,    0.76, -1.60166)
```

</details>

With micro the bricks can win in any scenario because they can dodge huge chunks of damage from the Percies.

## Checklist


lL1l1:
### New changes:
HP: 7500 (no longer changed)
The reduction to 7250 HP made Percies way too good against Bricks in a fight without micro, which is undesirable for what are basically identical type mainline units.

Damage (DPS): 125 (312.5) -> 120 (300) -> 118 (295)
A little bit more of a DPS nerf since the HP is no longer reduced. 5.6% lower DPS compared to 7.2% lower HP * DPS from before.

### Testing:
Brick right clicked onto Percy: Percy survives with 474 HP

Brick dodges 1 shot fighting Percy: Brick lives with 250 HP

Percy kiting brick: Percy survives with 2008 HP (was 1920 with 120 dmg, 1700 with 125 dmg)

Rhaelya (approved):
Everything is as it should be.

PR #6646, opened 2025-02-07
Remove the chat sound when watching a replay

Garanas:

This refers to the sound when you receive a message. The sound plays in replays. I personally find it to be an annoying sound, but this is of course subjective.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6646.md
+++ b/changelog/snippets/fix.6646.md
@@ -1,1 +1,1 @@
-- (#6646) Remove the chat bleeps when watching a replay
+- (#6646) Remove the chat beeps when watching a replay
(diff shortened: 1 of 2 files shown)


changelog/snippets/fix.6646.md
```diff
@@ -0,0 +1 @@
+- (#6646) Remove the chat bleeps when watching a replay
```
lL1l1, line 1: A bleep is the sound used to censor swear words.
```suggestion
- (#6646) Remove the chat beeps when watching a replay
```

lL1l1 (approved):
The current implementation of the chat beep doesn't make sense so removing it is okay. There is a chat beep UI mod (ChatBeepLite) that has a more reasonable design (beeps only for ally chat/private chat messages). Maybe that could be a feature to implement with configuration options.

PR #6647, opened 2025-02-07
Add a guard to prevent a buffer overflow when saving a game

Garanas:

According to @4z0t there's a possible buffer overflow in this function. This function is exposed to UI mods. It would technically allow a UI mod to run arbitrary code when the UI mod is enabled.

## Checklist

### DIFF
--- a/lua/ui/globals/InternalSaveGame.lua
+++ b/lua/ui/globals/InternalSaveGame.lua
@@ -31,11 +31,11 @@ do
     ---@param filename string
     _G.InternalSaveGame = function(filename, friendlyFilename, onCompletionCallback)
         if DebugAllocatedSize(filename) > 50 then
-            filename = filename:sub(0, 50)
+            filename = filename:sub(1, 50)
         end
 
         if DebugAllocatedSize(friendlyFilename) > 50 then
-            friendlyFilename = friendlyFilename:sub(0, 50)
+            friendlyFilename = friendlyFilename:sub(1, 50)
         end
 
         return oldInternalSaveGame(filename, friendlyFilename, onCompletionCallback)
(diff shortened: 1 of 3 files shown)


lua/ui/globals/InternalSaveGame.lua
```diff
@@ -0,0 +1,43 @@
...
+do
+    local DebugAllocatedSize = debug.allocatedsize
+
+    local oldInternalSaveGame = _G.InternalSaveGame
+
+    --- Hook to fix a buffer overflow security issue in the engine
+    ---@param filename string
+    _G.InternalSaveGame = function(filename, friendlyFilename, onCompletionCallback)
+        if DebugAllocatedSize(filename) > 50 then
+            filename = filename:sub(0, 50)
```
4z0t, line 34: Lua is 1-indexed.

Garanas: Fixed in https://github.com/FAForever/fa/pull/6647/commits/f8ef9726fc7eb58728580c9a95bd6b4a82ffe8c4

4z0t (approved):

Hdt80bro (changes requested):
50 characters is too short (the buffer size windows uses is 260) and checking the C object size is incorrect; Lua's `string.len()` gets the raw character length.

Garanas:
But characters can be 8 or 16 bytes. How do I know which ones they are?

Hdt80bro:
I don't think Lua cares; it says a string's length is always the raw number of bytes composing it, regardless of multicharacter glyphs. I don't know how LuaPlus changes that with wide characters, but it should be easy to test at least. I'm realizing that it isn't an exact science anyway since this filename is going to appended to the user's save directory (which also uses some of that buffer), so probably setting it to 100 characters and shipping it is fine.

Hdt80bro (approved):

Hdt80bro:
This completely broke file saving for some people.

4z0t:
Then we have to fix the engine.

PR #6648, opened 2025-02-08
Make Terrain000 use the water channel in the map info texture

BlackYps:
The map editor will be able to automatically generate the map info texture with all the channels, so there is no reason not to use the higher resolution water texture.

## Checklist

### DIFF
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -2097,7 +2097,7 @@ float4 TerrainPBRAlbedoPS ( VS_OUTPUT inV) : COLOR
 // | S6 | R             G               B              specularity     | X               Y               Z               unused       |
 // | S7 | R             G               B              specularity     | X               Y               Z               unused       |
 //  ----            ---             ---             ---              ---             ---             ---             ---            ---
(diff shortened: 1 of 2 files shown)


Garanas (approved):

effects/terrain.fx
```diff
@@ -2221,17 +2222,13 @@ float4 Terrain000AlbedoPS ( VS_OUTPUT inV, uniform bool halfRange ) : COLOR
     float3 r = reflect(normalize(inV.mViewDirection), normal);
     float3 specular = pow(saturate(dot(r, SunDirection)),80) * albedo.aaa * SpecularColor.a * SpecularColor.rgb;
 
-    // compute lighting
-    float dotSunNormal = max(0, dot(SunDirection, normal));
+    float dotSunNormal = dot(SunDirection, normal);
```
Garanas, line 2225: You want to allow negative numbers here?

BlackYps: two lines below the value is clamped anyway. I just removed redundancy

PR #6649, opened 2025-02-08
Nerf the Seraphim Destroyer's main turret in exchange for a buff to its torpedo defense

Basilisk3:
**Uashavoh: T2 Destroyer (XSS0201):**
- Front Ultrachromatic Beam Generator
  - Damage: 50 --> 45
  - Total directfire DPS: 209 --> 195
- Ajellu Anti-Torpedo Defense
  - FiringTolerance: 90 --> 180
  - MinRadius: 15 --> 5
  - Reload: 3.3s --> 3.0s
  - `UseFiringSolutionInsteadOfAimBone`: `false` --> `true`

- Performance against Tech 2 Torpedo Launchers stays more or less the same.
- The Seraphim Destroyer is slightly more effective against subs.

## Checklist


Basilisk3:
I also applied the same changes to torpedo defenses from #6607 here. Visually there is basically no difference, as the turret could turn 360 degrees already. So now it just works better, and there is also no more need for the pretty large `MinRadius` of `15`.

Now the question is whether we still want the 0.3s buff to its reload time.

Tagada14 (approved):

PR #6653, opened 2025-02-14
Close automated attack move loophole

lL1l1:
## Issue
The attack move sim callback relies on position data written in lua on the user side, which allows for automation of attack move orders.

Thanks to an engine patch the sim can inspect a unit's command queue.
- Change the callback to get the target position from the command queue.
  - There is no way to add a position to that command queue data automatically, so the automation is fixed.
- related annotations

- Use the attack move button to issue shift/not shift attack moves
- Use this script (use in clipboard) to see that it doesn't work for lua script tasks:
```lua
    IssueUnitCommandToUnit(GetSelectedUnits()[1], "UNITCOMMAND_Script", {TaskName = 'AttackMove'})
```
- And this one to see the callback doesn't work automatically.
```lua
    local cb = { Func = "AttackMove", Args = { Clear = true, Target = {50, 50, 50} } }
    SimCallback(cb, true)
```

An unavoidable part of the implementation since the very beginning is that the unit queue is filled with useless script commands, so the unit is slightly unresponsive as it goes through the script command before moving on to the attack move task.

## Checklist

### DIFF
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -17,6 +17,7 @@ local SimPing = import("/lua/simping.lua")
 local SimTriggers = import("/lua/scenariotriggers.lua")
 local SUtils = import("/lua/ai/sorianutilities.lua")
 local ScenarioFramework = import("/lua/scenarioframework.lua")
+local UnitQueueDataToCommand = import("/lua/sim/commands/shared.lua").UnitQueueDataToCommand
 
 -- upvalue table operations for performance
 local TableInsert = table.insert
@@ -26,6 +27,7 @@ local TableMerged = table.merged
 
 -- upvalue scope for performance
 local type = type
+local TableGetn = table.getn
 local Vector = Vector
 local IsEntity = IsEntity
 local GetEntityById = GetEntityById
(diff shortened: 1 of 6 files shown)


lua/SimCallbacks.lua
```diff
@@ -264,14 +266,28 @@ Callbacks.ValidateAssist = function(data, units)
...
 Callbacks.AttackMove = function(data, units)
-    -- exclude structures as it makes no sense to apply a move command to them
+    -- exclude structures as they can't move and there is no sim command to issue attack move rally points for factories
     local allNonStructures = EntityCategoryFilterDown(categories.ALLUNITS - categories.STRUCTURE, units)
 
+    -- Verify that the user manually clicked to issue the command, and use that position.
+    -- assume all units in the selection were given the same order, so we only need to check one unit
+    local commandQueue = allNonStructures[1]:GetCommandQueue()
+    local lastcommand = commandQueue[TableGetn(commandQueue)]
+    LOG(repr(lastcommand), debug.traceback())
```
Hdt80bro, line 280: Looks like some debug code got left

Hdt80bro (changes requested):

Hdt80bro (approved):

PR #6173, opened 2024-05-12
Use tooltip class for order buttons' tooltips instead of custom code

lL1l1:
## Description of the proposed changes

Fixes abnormally slow alpha change of order button tooltips. Removes code duplication.

## Testing done on the proposed changes

Mousing over orders creates a smooth transparency transition just like the tooltips for the construction menu buttons (tech level, templates, unit selector, construction selector, enhancements selector).

## Additional context

The old code is from GPG files.

## Checklist

- [x] Changes are documented in the changelog for the next game version

PR #6654, opened 2025-02-14
Fix order button tooltips not disappearing on mouse exit

lL1l1:
## Issue
https://github.com/FAForever/fa/pull/6173 broke the mouse exit behavior because it didn't update all usages of the tooltip, under the incorrect assumption that the create function would return a Control that can be destroyed by the rest of the code like usual.

Update the remaining code for the order button tooltips to use the Tooltip class correctly.

Mouse over order buttons then move the mouse away and the tooltips should disappear. Previously the order button tooltips would only disappear when a new tooltip was created.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6654.md
+++ b/changelog/snippets/fix.6654.md
@@ -1,0 +1,1 @@
+- (#6654) Fix order button tooltips not disappearing when the mouse is moved off the order button.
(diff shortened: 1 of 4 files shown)


lL1l1:
This has some mod conflicts when they try to add new order buttons :/
Mods: M28AI, Advanced Target Priorities Tweaked

speed2CZ (approved):

PR #6667, opened 2025-02-19
Implement decapitation victory condition

### LINKED ISSUE #5347: Decapitation as a rated victory condition
Decapitation is currently an unrated, modded victory condition implemented by Sheikah. It lets people continue playing even after their commander dies, but the game ends in a loss for the team that loses all commanders.
It's a nice feature for full-share games with a mix of new and veteran players because it avoids the problems associated with the highest rated getting double eco after a beginner dies, and it punishes beginners less for using their commanders, which is otherwise one of the hardest things to learn because of the extreme cost associated with losing it. From the tests I ran with Decapitation, the game balance was less volatile than standard full-share.

Integrating it as a core feature would increase it's visibility (since by default, modded games are not shown in the custom game list) and making it rated has already been ok'd by the balance team in this [forum thread](https://forum.faforever.com/topic/5739/decapitation-should-be-a-rated-victory-condition?_=1692789042117)

lL1l1:
Implement the decapitation victory condition, where all commanders in a team must die for the team to be defeated. Resolves #5347.

Still needs translation for other languages.

How it looks in the lobby UI:
{16FA0A28-0A3F-4180-B25B-C1BAAA7EEF42}

What the log looks like when all brains are defeated:
```
info: { -- table: 20847D98 (48 bytes)} -- 1 ai is missing the ACU but is not defeated because of the alliance defeat requirement
info: { -- table: 272F1190 (48 bytes)}
info: { -- table: 1CEEE550 (72 bytes) -- both ais are now defeated, so the table fills
info:   {...} -- table: 205A6208 (2600 bytes),
info:   [3] = {...} -- table: 2065D730 (2600 bytes)
info: }
info: { -- table: 2722F938 (72 bytes) -- second call to the function to "take into account cascading effects"
info:   {...} -- table: 205A6208 (2600 bytes),
info:   [3] = {...} -- table: 2065D730 (2600 bytes)
info: }
debug: Matchstate - defeated: Thel-Awloshi (AI: Easy)
debug: Matchstate - defeated: Ronald (AI: Easy)
debug: Matchstate - won: Oum-Ullehesh (AI: Easy)
debug: Matchstate - won: Player
```

Defeat works when alliances are changed. Ran `SimLua SetAlliance(1, 3, 'enemy')` in the console to test. The game continues as expected afterwards (since I set myself to 1 team, my ally army 3 to another, and armies 2/4 to a third team).

## Additional Context
I did look at the mod by Sheikah but went with a different implementation that doesn't iterate unit checking over all brains multiple times.

## Checklist


Hdt80bro:
My first thought was "uh, all commanders on a team needing to be destroyed for the team to be out is normal assassination" - but what this actually does is only make an army go out once all other teammates' commanders also get destroyed? If so, we should clarify that.

lL1l1:
I was matching it with the simple language used in other victory conditions, where "Game Ends" means "Your army is put out of the game" and 2nd person isn't used, but you're right it's confusing that way.

loc/US/strings_db.lua
```diff
@@ -4533,6 +4533,8 @@ lobui_0126="Annihilation"
 lobui_0127="Game ends when all units are destroyed"
 lobui_0128="Sandbox"
 lobui_0129="Game never ends"
+lobui_0152="Decapitation"
+lobui_0131="Game ends when all allied commanders are destroyed"
```
Hdt80bro: Wrong tag

lL1l1: Fixed

Garanas (approved):

PR #6674, opened 2025-02-24
Improve the creation of chunk templates

Garanas:

Improves upon https://github.com/FAForever/fa/issues/5633 by copying the template to the clipboard. 

With thanks to @4z0t for the [engine patch](https://github.com/FAForever/FA-Binary-Patches/pull/50) that makes this possible.

Verified that the same output is generated as before. As an example, it turns the following selection:

image

Into the following table:

```Lua
{
  Faction = UEF, 
  Size = 16, 
  BuildAreas = { 
    {  }, 
    { { -3.50, 1.50 }, { -1.50, 1.50 }, { 6.50, -0.50 }, { -5.50, 1.50 }, { 4.50, -0.50 }, { 2.50, -0.50 },  }, 
    {  }, 
    {  }, 
    {  }, 
    { { 4.50, -4.50 },  }, 
    {  }, 
    { { 4.50, 4.50 }, { -3.50, -3.50 },  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
  }, 
} 
```

That is available in your clipboard. The developer can then proceed to copy it into a Lua module.

## Additional context

There is still some missing functionality. As an example, the faction is hardcoded to UEF. This pull request does not address that. It just makes it easier to use to create templates.

## Checklist


relent0r:
Hey @Garanas 
So to confirm what I should be looking to test here. Is it effectively a simulation of what the AI would be doing when the OnUnitStopBuild callback fires and a structure is completed, its then updating a chunk with the SkirtSizeX/SkritSizeZ of the unit?

Garanas:
> Hey @Garanas So to confirm what I should be looking to test here. Is it effectively a simulation of what the AI would be doing when the OnUnitStopBuild callback fires and a structure is completed (assuming I took that route), its then updating a chunk with the SkirtSizeX/SkritSizeZ of the unit?

No, chunks are a way for us to define 'chunks' of bases. This pull request just improves the tooling to create the chunks. The actual interaction with the chunks does not exist yet.

What you can test is whether you can indeed create a chunk and that the result of the chunk is copied to your clipboard. It's a relatively small test.

My idea of the nearby future is to:

- 1) Have a folder of 'chunks', one for each faction. We can populate and extend these overtime. And because they are on a per-faction basis, we can define different base patterns for each faction.

The fun part here is that technically anyone can make chunks. Meaning that if this starts to take off we can surely find some excited contributors that just want to build chunks for their favorite faction.

- 2) Introduce an AIBrain component that manages chunks. This component provides an API (functions) to load and retrieve chunks. You can then populate the manager with the chunks that you're interested in. These could originate from the repository, or from an AI mod.

This will also include some debugging interfaces to allow you to determine what chunks are loaded. And to preview chunks, to make it easier to understand what the options of the AI are when it is expanding its base.

- Introduce an AIBrain component that manages build areas. A build area is extremely similar to a square of the navigational mesh. Unlike the navigational mesh, this component only manages build area's for land and (surface) naval. Since you can't build anywhere else in the default game.

This is where things get interesting! Essentially we're building a replacement for `CAiBrain:FindPlaceToBuild` where instead of passing a gigantic table of possibilities (from `basetemplates.lua`) we instead look at what we're trying to build, go through the applied chunks so far and see if there's a place for it to build. Especially late game it would reduce the number of 'attempts'  from hundreds or thousands to just a few.

This focus of this pull request is to make step 1 easier, the creation of chunks.

relent0r:
I tested this with a few scenario's. Though I couldn't remember which each key size meant.
Was it related to SkirtSize or something else?

I also tried naval as I thats one of the painpoints where the findplacetobuild returns bad positons.

image

{
  Faction = UEF, 
  Size = 32, 
  BuildAreas = { 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    { { 6.50, -6.50 }, { -7.50, -6.50 }, { -7.50, 7.50 }, { 6.50, 7.50 },  }, 
    {  }, 
    { { -0.50, -7.50 }, { -0.50, 8.50 }, { 7.50, 0.50 }, { -8.50, 0.50 }, { -0.50, 0.50 },  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
  }, 
} 

image

This one is interesting as the borders go outside the 16x16 square but the units themselves stay in.

{
  Faction = UEF, 
  Size = 16, 
  BuildAreas = { 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    { { -3.50, -3.50 }, { 4.50, -3.50 }, { -3.50, 4.50 },  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
  }, 
} 

image

{
  Faction = UEF, 
  Size = 16, 
  BuildAreas = { 
    {  }, 
    { { -5.50, -4.50 }, { 2.50, -4.50 }, { -3.50, -2.50 }, { -3.50, -4.50 }, { -1.50, 3.50 }, { 0.50, 3.50 }, { 2.50, 3.50 }, { 4.50, -2.50 }, { 4.50, -0.50 }, { -3.50, 1.50 }, { 0.50, -4.50 }, { -3.50, -6.50 }, { -1.50, -4.50 }, { -3.50, -0.50 }, { 4.50, 1.50 },  }, 
    {  }, 
    {  }, 
    {  }, 
    { { 0.50, -0.50 },  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
    {  }, 
  }, 
}

relent0r:
Overall I'm really interested to see how this might work compared to our existing stuff now. 

The two pain points I know we have that this could potentially resolve.
T2+ factories causing issues with build positions being returned that are within the boundary.
Naval factory build positions (and torpedo/sonar) being returned that cause pathing issues. Though I'm unsure if we might run into rotation issues with the factories?

relent0r (approved):

PR #6676, opened 2025-02-26
Fix dual yaw controller affecting unit pitch

lL1l1:
## Issue
The dual yaw controller used by the loyalist pitches the unit at the torso (or whatever bone is chosen). It isn't noticeable with the loyalist because it has a high muzzle velocity for its range, but it can be clearly seen if the weapon has low muzzle velocity:
{115B36D0-B9D3-43FA-A84C-F535C11670C9}

- Remove the pitch component from the dual yaw controller's initialization and firing arcs.

Spawn loyalists with the test scenario commit and see that they don't pitch with their torso:
```lua
   CreateUnitAtMouse('url0303', 0,   12.00,    0.50, -0.00000)
   CreateUnitAtMouse('url0303', 1,  -12.00,   -0.50, -0.00000)
```
{5439C585-96F4-4851-9119-846EA0381C76}

## Additional context
I noticed that an additional bonus of the fixed dual yaw controller is that it can compensate for the movement animation of walkers, removing the necessity of `UseFiringSolutionInsteadOfAimBone = true`.

## Checklist


Hdt80bro:
@lL1l1 this has conflicts now, can you resolve those?

lL1l1:
@Hdt80bro I resolved the conflicts

speed2CZ:
This looks like a significant change the these units. Could it affect modded units in some negative way?

lL1l1:
> This looks like a significant change the these units. Could it affect modded units in some negative way?

I added the dual yaw controller only 5 months ago, and it was never intended to be used as a secondary pitch controller (the naming and comments make it clear), so I don't think any mods will be affected negatively.

speed2CZ (approved):

speed2CZ:
Then ship it! :D

PR #6679, opened 2025-02-28
Resolve circular dependency of WorldView and CommandMode

4z0t:
Remove import of `WorldView.lua` in `CommandMode.lua` which is imported inside `WorldView.lua` causing circular dependency of these 2 modules.

## Checklist

### DIFF
--- a/changelog/snippets/other.6679.md
+++ b/changelog/snippets/other.6679.md
@@ -1,0 +1,1 @@
+- (#6679) Resolve circular dependency of WorldView and CommandMode.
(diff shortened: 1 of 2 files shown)


lL1l1 (commented):
I looked into the history and the local was added in #3484 and usage of it was removed in #4124 which is not that long of a time and is quite specific in purpose so I think it's fine to remove.
Can you add a snippet?

4z0t:
bump

lL1l1 (approved):

PR #6680, opened 2025-03-01
Annotate SetWeaponPriorites callback and usage

lL1l1:
Annotate `SetWeaponPriorites` callback and usage

Reading the code and references.

## Checklist


lua/WeaponPriorities.lua
```diff
@@ -131,6 +140,11 @@ function SetWeaponPriorities(data)
 
         --checks if unit is already in requested mode
         if weaponCount > 0 and unit.Sync.WepPriority ~= name then
+            ---@alias UnitSyncWepPriority 
+            ---| string # Custom priorities with a name
+            ---| "?????" # Custom priorities without a name
+            ---| "Default"
+
```
Hdt80bro, line 147: I prefer all annotation data to be at the top of the file. Thoughts?

lL1l1: - For viewing, I use F12 to jump to definitions and having the definition next to the implementation makes it easy to understand if i need to.
- For editing, putting the definition next to the implementation makes it obvious that it needs to be changed when that code is edited. It makes the diff a bit less clean but I value that less than having it next to relevant code.

Hdt80bro: That makes sense

Hdt80bro (approved):

PR #6681, opened 2025-03-01
Don't initialize/alter objects during module loading in `Construction.lua`

4z0t:
Initializing/altering objects during loading of a module is a bad habbit, especially if it is from another module. In this case it can cause error and game freeze if UI mod tries to access `construction.lua` before UI is created.

Queue dragging works as normal.

## Checklist


lL1l1 (commented):
I tested this and found no issues.

Can you add a snippet?

4z0t:
bump

PR #6683, opened 2025-03-01
Yet another shader rework

BlackYps:
It is time for another stab at the terrain shader file. I gathered some new insights on what is important to produce the best results possible, partly from looking at BAR maps, partly from looking at maps from @sting-2 and these insights are the basis for the following changes.

The way to achieve the best quality seems have a huge (4-8k) map-wide albedo texture and to interpolate it with the other stratums that provide details. This way the map albedo texture can break up the texture repetition and provide much more color variation than you could by just tiling 8 regular textures.
The mapwide normal should ideally also be huge. To justify thse file sizes we need compression. Until now we stored the shadows, water depth and map normals all in one texture. Compressing these leads to severe artifacts and basically all our inputs are sensitive to artifacts. The dds compression was meant for colors, but if we store non-color data we notice this much more.
But we can utilize a trick. The alpha channel does not suffer from artifacts and the color channels also have basically no compression if we store the same value in all color channels. So we can store two channels per texture, with compression, and it still looks good. We put the normals in the texture like this:
R: normal y
G: normal y
B: normal y
A: normal X
A nice upside of this is that normal decals are compatible to this encoding. So if someone has made a normal decal for their map normals in the past, they can use this texture directly.
By separating the normals from the shadows and water, we can crank up the resolution on the normal texture and choose a lower resolution on our shadows and water, saving disk space.

In fact, we only need the water channel when we have a map with biplanar mapping (BPM), because then we need to sample one more texture and we are already at the limit of texture samplers. We store the channels like this:
R: AO or water depth
G: AO
B: AO
A: shadow
When we have a non-BPM map, then we can fill the water channel with AO and save the texture with compression without artifacts. On BPM maps we have to save uncompressed because we need the water channel.

The downside of all this is that we have one more texture now. I changed the slots of our map textures to accomodate this as well as possible.

The macro texture now does not have a dedicated slot anymore. Instead the lower albedo can get used as a macrotexture. Different shader variants allow control over how the lower albedo gets treated.

I changed the order of the height an roughness channels in the PBR atlas. This way you can omit the alpha channel with BC1 compression for shaders that don't need the height information. This helps saving a little bit of disk space.

**Summary of the changes:**

_Previous layout:_
upper albedo: map info texture: map normal + water depth + shadow
stratum7 albedo: macrotexture with alpha
stratum7 normal: heightroughness texture

_New Layout:_
upper albedo: RoughnessAndHeight texture
stratum7 albedo: terrain info texture: water depth + ambient occlusion + shadow
stratum7 normal: terrain normal: normal y + normal x

- Usage of these textures is controlled by their respective scale.
- Added a biplanar mapping version for the Terrain100 shaders.
- Added shaders for different ways to blend the lower albedo on top.
- Increased the contrast of the height blending
- Switched height and roughness channels

### DIFF
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -1918,6 +1918,35 @@ float blendHeight(float3 position, float2 blendWeights, uniform float2 nearscale
     return (heightNearYZ + heightFarYZ) / 2 * blendWeights.x + (heightNearXZ + heightFarXZ) / 2 * blendWeights.y;
 }
 
+float2 calculateBlendWeights(float2 position) {
+    float2 terrainNormal = tex2D(Stratum7NormalSampler, position).ag * 2 - 1;
+    float2 blendWeights = pow(abs(terrainNormal), 3);
+    blendWeights = blendWeights / (blendWeights.x + blendWeights.y);
+    return blendWeights;
+}
+
+float3 applyPbrShading(VS_OUTPUT inV, float2 position, float4 albedo, float3 normal, float roughnessVariation) {
+    // We need to add 0.01 as the reflection disappears at 0
+    float roughness = saturate(albedo.a * roughnessVariation * 2 + 0.01);
+
+    float waterDepth = tex2D(Stratum7AlbedoSampler, position).a;
+    float3 color = PBR(inV, albedo, normal, roughness, waterDepth);
+    color = ApplyWaterColor(-1 * inV.mViewDirection, inV.mTexWT.z, waterDepth, color);
+    return color;
+}
+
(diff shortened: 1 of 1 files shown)


BlackYps:
I am not sure if anyone actually feels comfortable reviewing this, as it is quite a special area

Garanas (approved):
I've experimented with it both alone and together with @BlackYps  - combined with the changes to the editor it is one bulk of amazing work!

See also:

- https://github.com/FAForever/FAForeverMapEditor/pull/65
- https://github.com/FAForever/FAForeverMapEditor/pull/66

PR #6684, opened 2025-03-02
Console improvements

speed2CZ:
- search results are now sorted alphabetically
- clicking on an item doesnt steal the focus from the Edit, so UP/DOWN keys can be used to keep moving through the list
- the item list correctly follows the focus item and wraps around if needed


lL1l1 (approved):
Works.

PR #6688, opened 2025-03-11
Proper generic intellisense for LazyVar

4z0t:
Annotations for generic LazyVar. It addresses issues that were appearing when working with generic types of this class.
* `Lazy<T>` - common generic for it. Represents object or function.
* `LazyVar<T>` - represents only object that has methods `Set`, `SetValue`, `SetFunction`, `Destroy` and `OnDirty`.
* `LazyOrValue<T>` - `Lazy<T>` or `T`. 

## *Why like this?*
*This is done like this because our intellisense doesn't support generic classes properly and also generic operators*.

## Checklist

### DIFF
--- a/engine/User/CMauiBitmap.lua
+++ b/engine/User/CMauiBitmap.lua
@@ -1,6 +1,8 @@
 ---@meta
 
 ---@class moho.bitmap_methods : moho.control_methods
+---@field BitmapWidth Lazy<number>
+---@field BitmapHeight Lazy<number>
 local CMauiBitmap = {}
 
 ---
--- a/engine/User/CMauiControl.lua
+++ b/engine/User/CMauiControl.lua
@@ -1,13 +1,13 @@
 ---@meta
 
 ---@class moho.control_methods : Destroyable
----@field Left LazyVar<number>
----@field Width LazyVar<number>
----@field Right LazyVar<number>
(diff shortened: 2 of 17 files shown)


4z0t:
@lL1l1, @Garanas What do you think about naming? Doesn't it look confusing for you?

lL1l1 (commented):
- I don't like using just `Lazy` because it doesn't really make sense to use an adjective as the name of an object.
- on the other hand `LazyVarOrValue<Type>` would be a bit verbose. `LazyVar<Type> | Type` would be repetitive.
- Changing the type to `Lazy` is also going to break any UI mod annotations that use lazyvars.
- Lazyvars would be the word when we talk about lazyvars but the `Lazy<Type>` syntax wouldn't match.

I think weighing those factors, `LazyVar<T>` and `LazyVarOrValue<T>` should be what is used in UI, and to make annotations work `LazyVar<T>` can be `LazyVarClass<T>` (for example).

Hdt80bro (changes requested):
One consideration to make is that, for controls, the engine is the one that gives their fields their lazy vars from the LazyVar file and it expects them to be such when it retrieves or sets the values. So, they can't be `Lazy`, they are explicitly `LazyVar`s.

PR #6690, opened 2025-03-13
Buff UEF Destroyer DPS/speed/turn rate at cost of HP

lL1l1:
## Issue
With the current UEF destroyer rebalance, it still has issues in navy because while it does have superior HP and gun DPS, it does not have the speed to force favorable fights, its short range loses it a lot of HP over time, and the gun DPS advantage is equalized once torpedo DPS comes into play.

UEF Destroyer (UES0201):
- Help the destroyer take better fights:
  - Max speed/acceleration/braking: 5.0 -> 5.5
  - Turn rate: 50 -> 55
- Buff DPS and revert the HP buff so that more proactive usage of the destroyer is better rewarded:
  - Gauss cannon damage (DPS): 275 (275) -> 305 (305)
    - It can now 6 shot exodus (6600 -> 7320 damage per 6 volleys); it is unlikely due to inaccuracy, but it is worth noting.
  - HP: 8000 -> 7200

Spawn other faction destroyers and give an attack order (make sure to surface Seraphim destroyers, as they are cheat spawned submerged):
<details> <summary>  Here is an automatic script to test this, just run it from your clipboard:
</summary>

```lua
CreateUnitAtMouse('uas0201', 1,  -57.92,   79.42, -0.00000)
CreateUnitAtMouse('urs0201', 1,  -55.42,    6.92, -0.00000)
CreateUnitAtMouse('xss0201', 1,  -51.92,  -82.58, -0.00000)
CreateUnitAtMouse('ues0201', 0,   49.08,  -88.58, -0.00000)
CreateUnitAtMouse('ues0201', 0,   64.08,   -0.58, -0.00000)
CreateUnitAtMouse('ues0201', 0,   52.08,   85.42, -0.00000)
ConExecute([[SimLua 
---@type AIBrain
local hum = ArmyBrains[1]
---@type AIBrain
local civ = ArmyBrains[2]
local humDestros = hum:GetListOfUnits(categories.DESTROYER, false, true)
table.sort(humDestros, function (a, b)
    return a:GetPosition().z < b:GetPosition().z
end)
local civDestros = civ:GetListOfUnits(categories.DESTROYER, false, true)
IssueDive(civ:GetListOfUnits(categories.SERAPHIM, false, true))
IssueAttack({humDestros[1]}, civ:GetListOfUnits(categories.SERAPHIM, false, true)[1])
IssueAttack({humDestros[2]}, civ:GetListOfUnits(categories.CYBRAN, false, true)[1])
IssueAttack({humDestros[3]}, civ:GetListOfUnits(categories.AEON, false, true)[1])
SimConExecute("wld_runwiththewind")
ForkThread(function ()
    local liveDestro = true
    while liveDestro do
        WaitTicks(1)
        liveDestro = false
        for _, u in civDestros do
            if not u.Dead then liveDestro = true break end
        end
    end
    WaitSeconds(3)
    LOG(string.format('Uaashavoh: %d leftover hp', not humDestros[1].Dead and humDestros[1]:GetHealth() or 0))
    LOG(string.format('Salem: %d leftover hp', not humDestros[2].Dead and humDestros[2]:GetHealth() or 0))
    LOG(string.format('Exodus: %d leftover hp', not humDestros[3].Dead and humDestros[3]:GetHealth() or 0))
    for _, u in humDestros do
        u:Destroy()
    end
    for _, u in civDestros do
        u:Destroy()
    end
    SimConExecute("wld_runwiththewind")
end)
]])
```

</details>

Before changes:
Uaashavoh: ~1000/~1750 leftover HP
Salem: ~200/~1300/~1900 leftover HP
Exodus: 1640/580/0 (draw) leftover hp

It seems like inaccuracy plays a huge role in these simple direct fire tests, so the results are inconsistent. UEF Destroyer has the highest firing randomness at 0.35, compared to 0.3 for Salem and 0.25 for Exodus (Exodus also has 1.4 instead of 1 splash).

After changes:
Uaashavoh: ~1000/~1750 leftover HP
Salem: ~200/~800/~1200 leftover HP
Exodus: 1900/840/0 (draw or loss) leftover HP

It seems like the DPS buff is compensating the HP loss appropriately, so leftover HP stays the same. This indicates a buff because we have more DPS but the same HP left over after a fight. These UEF Destroyers should snowball better than the HP buff variant, which is acceptable because they're at a range disadvantage, which is a primary factor in real games.
The speed and turning buff help a little bit in this scenario too, countering the range advantage slightly.

## Checklist


Rhaelya (approved):
Everything looks in order

PR #6691, opened 2025-03-13
Reduce the delay before pausing to 1 tick

Garanas:

Reduces the delay before a unit is paused to just 1 tick. I'm not entirely sure why it was 5 ticks to begin with. Because of the additional ticks _some_ resources would be spent on the unit in question.

Based on a discussion on [Discord](https://discord.com/channels/197033481883222026/1349762080428789791).

Run a game and observe that the feature still works as expected. No resources are 'leaked' to the upgrade.

## Checklist


BlackYps (approved):
Looks good. I didn't test it in game again, but the change seems small enough

PR #6694, opened 2025-03-18
Add benchmarking

Hdt80bro:
Completes #4087, adding in-game benchmarking capabilities to the profiler.

## Checklist


lua/SimCallbacks.lua
```diff
@@ -1258,6 +1258,22 @@ Callbacks.NavGenerate = function(data, units)
     import("/lua/sim/navgenerator.lua").Generate()
 end
 
+-- Allows searching for benchmarks
+Callbacks.FindBenchmarks = function (data)
+    import("/lua/sim/profiler.lua").FindBenchmarks(data.Army)
```
speed2CZ, line 1263: I think all these callbacks should be available only with cheats enabled. So it cant be abused to lag the game.

speed2CZ (changes requested):

lua/SimCallbacks.lua
Hdt80bro: That is done in the callbacks themselves - do you think the check should be done sooner?

Garanas: The old profiler code also have the cheat guards in the respective functions.

I think it would make more sense to do the cheat guards to happen in `SimCallbacks.lua`, outside of the actual logic of the callback. That allows you to, purely theoretically in this case, re-use the actual functions in sim code.

Garanas (commented):
Started the game and did a few tests. This is essentially what I could only dream of when I was trying to improve the performance of the game.

Great job!

lua/SimCallbacks.lua
speed2CZ: It makes sense to do the checks there, as thats the entry point from the UI. Its also consistent with the rest of the code that does the checks there

Hdt80bro: Okeedokee

lua/SimCallbacks.lua
```diff
@@ -811,28 +811,34 @@ end
...
+end
 
-    -- allow when cheats are enabled
-    return PassesAntiCheatCheck()
+--- An anti-cheat check for profiling that only passes when the correct army
+--- triggered the callback
+---@return boolean
+local function PassesProfilingAntiCheatCheck()
+    if import("/lua/simutils.lua").OriginalFocusArmy == GetCurrentCommandSource() then
+        return SessionIsReplay() or PassesAntiCheatCheck()
```
Garanas: I'm not sure if this condition makes sense. You can't issue callbacks when watching a replay. And this guard means these callbacks would not work during the game, but then in the replay the callback would happen and it suddenly does work.

lua/SimUtils.lua
```diff
@@ -772,3 +772,5 @@ function DrawBone(entity, bone, length)
     -- Z axis
     DrawLine(pos, pos + forward * length, '0000ff')
 end
+
+OriginalFocusArmy = GetFocusArmy()
```
Garanas: Probably best not to store this here. It's a value that's different for all players. If you use it then you're guaranteed to have a desync.

lua/shared/DebugFunction.lua
Garanas, line 1: Does this file originate from a library? If not, please add an MIT license if you're fine with that.

lua/ui/lobby/lobby.lua
```diff
@@ -2812,7 +2812,7 @@ function CreateSlotsUI(makeLabel)
         newSlot.HandleEvent = defaultHandler
 
         -- Slot number
-        local slotNumber = UIUtil.CreateText(newSlot, i, 14, 'Arial')
+        local slotNumber = UIUtil.CreateText(newSlot, tostring(i), 14, 'Arial')
```
Garanas, line 2815: A random change I think - it is fine though.

lua/ui/maputil.lua
Garanas, line 1: Changes appear unintentional, they're fine though.

Garanas (commented):
Reviewed all the files - looks fine in general. Just a few small questions and then I think we can approve it.

lua/SimCallbacks.lua
Hdt80bro: I was going back and forth on it-I couldn't remember why it was there, so I left it. I'll take it out though, with that explanation.

lua/ui/lobby/lobby.lua
Hdt80bro: I somehow ran into an issue where `i` wasn't a number one time and the function errored. It could be a separate PR but I'm somewhat unmotivated to track down the issue.

lua/ui/maputil.lua
Hdt80bro: I don't know what it was about either, but it appeared in the diff and seemed fine to leave.

Hdt80bro:
I wanted a way to make sure the sim functions only ran for the person who requested them. It didn't seem like there was a good way though. If you think doing that will desync, I'll give in and just let it run on all computers even though it doesn't make sense. I'll make the changes later today when I get home.

Garanas:
> I wanted a way to make sure the sim functions only ran for the person who requested them. It didn't seem like there was a good way though. If you think doing that will desync, I'll give in and just let it run on all computers even though it doesn't make sense. I'll make the changes later today when I get home.

The value existing there won't cause a desync. But a future maintainer using that value while not understanding it may be a good source of desyncs.

It's probably best to just have it run for all machines. There's a cheat guard anyway - once that is enabled then there's a thousand ways to break a game.

Garanas (approved):

speed2CZ (approved):

PR #6698, opened 2025-03-19
Update hover cursor texture only when changed

speed2CZ:
## Checklist


Garanas (approved):
That makes sense, all appears to be fine!

PR #6704, opened 2025-03-27
Shader improvement

BlackYps:
Two small additions to the recent PR #6683.

_1)_ The specular color channels now control more behavior in the Terrain200 shaders. The green channel controls the contrast increase of the heightmaps that are used for texture blending. The blue and alpha channels store the values of the rotation matrix. This makes the rotation angle adjustable which could be beneficial in some cases.

The editor will not expose these values for now. Instead it will provide reasonable defaults. I plan to do it this way to give people time to first learn the important features of the new shaders. Then when they need more control I can expose these values. By doing this change now in the fa shaders and having the editor provide good default values, we ensure that all maps using Terrain200 shaders have reasonable values in the specular color and we won't break existing maps when we finally introduce the feature officially.

_2)_ The blurriness slider was not behaving properly after my recent refacter. Part of the reason is that the code was very hard to understand. The code should be more readable now and the behaviour is fixed. Now a high blurriness creates a smooth gradient over the whole mask range.

Low blurriness:
grafik

High blurriness:
grafik

old behaviour with high blurriness: The gradient is only in the lower half.
grafik

## Checklist

### DIFF
--- a/changelog/snippets/graphics.6683.md
+++ b/changelog/snippets/graphics.6683.md
@@ -1,0 +1,3 @@
+- (#6683 and #6704) Rework and improve the newly added terrain shaders.
+
+The actual changes are pretty technical, but if you are a mapper you can find an in-depth explanation on how to use the new shaders here: https://wiki.faforever.com/en/Development/Shaders/terrain-shaders
--- a/effects/terrain.fx
+++ b/effects/terrain.fx
@@ -1795,33 +1795,26 @@ technique LowFidelityLighting
 /* # Blending techniques # */
 
 float splatLerp(float t1, float t2, float t2height, float opacity, uniform float blurriness) {
-    // We need to increase the contrast of the height
-    float height2 = (1.6 * (t2height * (1 - 2 * blurriness) + blurriness) - 0.3) + opacity;
-    float threshold = max(1, height2) - blurriness;
-    float factor = 0;
-    if (opacity > 0) {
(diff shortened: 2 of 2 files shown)


Garanas:
I'm not sure this is a great way to use the specular texture. For me personally, it would make more sense to have the roughness and/or height texture to be read from the specular texture. I know this was initially not done because all existing game textures don't have this. But if we are going to use the specular texture it would make more sense to me to do it like a lot of modern engines and/or tooling do. 

It would also be more convenient to use then the current approach where they are separate. Assuming that a user wants to go into this direction then I suspect he or she is likely using custom textures anyway.

Specifically, I see no need to be able to control the rotation matrix. No external tooling supports that. And to me, it would be very difficult to obtain some effect out of it beyond the benefit that it's rotated in general.

BlackYps:
SpecularColor is not a texture. It's just a single RGBA value. So there is not too much we can use it for.
It's true that controlling the rotation angle does not have too many use cases. I can think of something though: For example if you have a texture of rectangular concrete tiles you might want to change the rotation angle from the default 30° to 90°.
The editor would provide ui functionality so that you can input the angle directly and you don't have to provide the individual rotation matrix entries.

BlackYps:
Still, if you have better ideas how we could use some float values to control interesting properties of the shader I would like to hear them

Garanas:
Oh - I miss understood the meaning of 'specular color channels'. I thought you were talking about a texture there.

In that case - forget what I said!

Garanas (approved):

PR #6705, opened 2025-03-29
Adjust Aeon range/speed gun upgrade costs

lL1l1:
## Issue
For some reason the Aeon range + speed upgrade combo does not match the cost of other factions' gun upgrades.

https://github.com/FAForever/fa/pull/5020 says the cost of the range upgrade should've been reduced 30%.
Actual changes:
- Mass: 500 -> 300 (-40%)
- Buildtime: 500 -> 350 (-30%)
- Energy: 15000 -> 6000 (-60%)

I believe what happened is that the energy got reduced 30% (15000 ->10500), then scaled with the mass (10500 -> 6300), and finally rounded down to 6000.

Judging by the mass cost there seems to be an intention of making it 800 mass in total like all other gun upgrades, but then the energy cost should be 9000 and the buildtime 300.

Neither the PR nor the parent issue give any reason for the energy difference. I also asked @Spikey84 about it on discord and he didn't say it was intentional.

## Balance reasoning
There is an argument to be made that the reduced energy cost allows rushing the upgrade faster, but that is not necessary because there is already a 40% cost reduction, giving a lot of time to use the 30 range while other factions finish building their guns.

  - Aeon ACU (UAL0001):
    - First range upgrade:
      - Mass cost: 300 -> 400
      - Energy cost: 6000 -> 12000 (20 -> 30 energy per mass)
      - Buildtime: 350 -> 400
    - Fire rate upgrade:
      - Mass cost: 500 -> 400
      - Energy cost: 15000 -> 12000
      - Buildtime: 500 -> 400
 
With these stats range + speed for Aeon match the cost of gun upgrades for other factions.

The upgrades are made equal cost so that neither can be rushed quicker than the other since they are both powerful.

The total mass/energy/buildtime are the same for Aeon and other gun upgrades.

## Checklist


lL1l1:
After a new discussion on [Discord](https://discord.com/channels/197033481883222026/1385738193080553545) I made the guns equal cost since either one can be very powerful on smaller maps, so making them equal cost makes them as difficult to rush any individual upgrade while keeping the total cost of range + speed equal to other factions.

Tagada14 (approved):

Tagada14:
If you feel inclined, you could update the description of the PR to reflect the changes after the aforementioned discord discussion.

PR #6708, opened 2025-03-31
Fix the tombstone's normals

lL1l1:
Sets the normals to `#808080` gray.

Before:
image
After:
image


BlackYps:
is the alpha channel 50% as well?

BlackYps:
should be 50% as well, because that is the second direction and 50% is the neutral value

lL1l1:
Ok I figured out how to do it, here's the result:
{94A38BFF-FC02-476E-BA72-B601525CCD32}

PR #6714, opened 2025-04-04
Fix the hitbox of the Volcano

Basilisk3:
Lasers could still miss the Volcano sometimes. The Volcano's `SizeY` now corresponds to the unit's height when it's deploying its flare.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6714.md
+++ b/changelog/snippets/fix.6714.md
@@ -1,1 +1,1 @@
-- (#6714) Prevent beam weapons from missing the Volcano (Aeon TMD).
+- (#6714) Prevent beam weapons from missing the Volcano (Aeon TMD) while it is firing. Also increase its hitbox size to match the firing animation.
--- a/changelog/snippets/fix.6714.md
+++ b/changelog/snippets/fix.6714.md
@@ -1,0 +1,1 @@
+- (#6714) Prevent beam weapons from missing the Volcano (Aeon TMD).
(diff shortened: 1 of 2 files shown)


lL1l1:
Last time you said that you tested with Seraphim destroyers with increased range. This is the only way I can consistently get the lasers to miss, but the necessary height of the hitbox depends on this range, so what range are you optimizing the hitbox for?

This may be an engine issue with how accurate it can get the turret to aim the beams if this is just a range issue. It is weird that the further away the destroyer is, the higher it aims its beam, even when using groundfire (notice the green lines for the beams).
{09C70408-7CE2-426D-B684-F9AC72244445}

Basilisk3:
> Last time you said that you tested with Seraphim destroyers with increased range. This is the only way I can consistently get the lasers to miss, but the necessary height of the hitbox depends on this range, so what range are you optimizing the hitbox for?

The miss is reproducible without increasing the range of the Seraphim destroyer, and I also noticed it in-game. It only occurs when the Volcano is opening up to deploy its flare, because the target bones move around. Judging from my tests, this is fixed with these changes.

lL1l1:
Ah I see, I can reproduce it too. Setting the targeting bone fixes this where the lasers miss while it is firing:

{9E6254E1-0FA2-4F6C-A153-B81DC736BBC3}

And as you said, increasing the hitbox size makes it match the animation:

{BE194AC3-9D92-4A4D-9F15-398B06ED83E6}

lL1l1 (approved):

PR #2721, opened 2019-02-11
Supreme Props Mod FAF integration

badbwoi:
Here is the workflow on how I integrated an improved version of the "Supreme Props Mod" into FAF.

Infos about the original mod can be found here: http://forums.faforever.com/viewtopic.php?f=41&t=15934#p161087
BUG & LOD FIXES are integrated like in the original mod (video presentation: https://tinyurl.com/y6zteb8r)
The reclaim values remain mostly untouched, except I considered them as BUG FIXES. The props I touched reclaim wise are highlighted in the picture, see here: http://forums.faforever.com/viewtopic.php?f=41&t=15934&p=171871#p171312 The applied values match "mass_value A", except "D", special "M" and "map elements".

CHANGES:
1) FILE INPUT & FILE FIXES (manual)
- added 234 original FA env "_prop.bp" files to FAF
- replaced '# ' with '-- ' comments
- removed white spaces, tabs to spaces, corrected formatting
- fixed structure in \env\Geothermal\Props\Rocks\GeoRock01_prop.bp
- added new 'Oak01_Group2_LOD2_albedo.dds' to fix shade error!

2) READ ALL CURRENT BP DATA & WRITE CSV (python)
- fix values like SizeZ = .0 --> 0.0
- excluded Seraphim II, marker, emitter and other unsuited props

3) FIXES AND CHANGES IN CSV (manual)
- apply Supreme Props LOD where it was needed
- fix wrong or none LOD 3 (including shadername3 and adapted shadername2 if needed)
- fix SIZE where it is off (too huge block e.g. Crystalline; to small block e.g. Aeon, Cybran & Seraphim ). This also improves the collision boxes (check in game on fa_props map with ctrl+shift+alt+c). Some conflict could not be resolved due to the prop origin not being in the (volume) center
- fix RECLAIM VALUES where it was needed e.g. Structures (had no value), most of broken 300 mass props like small crystals (this allows map maker to use the Crystalline props more often), Aeon, Cybran, Seraphim & UEF and some props that were clearly off. See info and change pic: http://forums.faforever.com/viewtopic.php?f=41&t=15934&start=10#p171312
- NOTE: props needs mass value >0 for footprint and blockpath to work

- TROPICAL single ficus' and palm's remain untouched

4) READ CHANGED BP CSV & OVERRIDE CURRENT PROP BP (python)
- ensure right structure-tree for different prop types (e.g. rock, tree, structures & wreckage have different value requirements)
[...]

PR #6715, opened 2025-04-04
Enable tree groups to be seen at higher zoom levels

Basilisk3:
As per the title. Tree groups are generally a pretty high priority, so it makes sense to keep them rendered for longer. The performance impact of this is minimal.

## Checklist


Garanas (commented):
I've not run these changes, but there's a reason that trees and tree groups share the same LOD values. See also this forum post:

- https://forum.faforever.com/topic/3852/scouts-and-labs-should-not-break-tree-groups/1

Tree groups are replaced _through_ the fog of war. I suspect that if the LODs are different, then you can see the tree groups 'pop' (disappear) through the fog of war when you're zoomed out. This is especially relevant for early raids.

Basilisk3:
That is a good point, I guess tree groups simply disappearing looks odd. I noticed very little performance impact from increasing the render distance of single trees but perhaps someone with a weak PC can confirm.

Garanas:
> That is a good point, I guess tree groups simply disappearing looks odd.

It is not so much about looking odd - you're giving away intel through the fog of war. It will affect the game.

> but perhaps someone with a weak PC can confirm.

You don't need a computer with light hardware to run into issues. My desktop specs that are relevant:

- CPU: 7950x3d AMD
- GPU: RX 7800 XT

If you have extended graphics enabled and maximum LOD settings then I can make my desktop go below 60FPS when zoomed out. If you do not have extended graphics enabled before starting the map, make sure to restart. 

Open the console and disable vsync (`SC_VerticalSync`) and then remove the frame time clamp (`sc_FrameTimeClamp 0`). Then enable statistics (`ShowStats`). Open up the `EntityCount` and the `Frame` drawers.

Make sure that you're at the maximum zoom level right before tree groups disappear. That's your base line.

Then you spawn in a 100 scouts and break all the tree groups. You can see the frame time go up as the prop count does. 

image
 
This example may appear academic. But it's not. Seton's Clutch is relatively light on tree groups in comparison to some maps. And players with less game knowledge will be less likely to reclaim tree group(s) and more likely to break them. Let alone that the map generator can generate some pretty tree-heavy maps. This happens - and it happens often.
 
The approach to rendering of this game was advanced for its time. But it's no where near the capabilities of modern game engines. This is just a fundamental issue that we run into. 

The current values are based on tinkering and tweaking across various months and deployments. See also:

- https://github.com/FAForever/fa/pull/4675
- https://github.com/FAForever/fa/pull/5021
- https://github.com/FAForever/fa/pull/3662

tldr; I don't think these should be adjusted. And if you still want it, then in my opinion it requires much more data sampling on how the adjusted settings impact the performance of commonly played maps where players do break the tree groups, but they don't clean them up.

Basilisk3:
> > That is a good point, I guess tree groups simply disappearing looks odd.
> 
> It is not so much about looking odd - you're giving away intel through the fog of war. It will affect the game.

To be fair, this has already been the case before these changes.

> This example may appear academic. But it's not. Seton's Clutch is relatively light on tree groups in comparison to some maps. And players with less game knowledge will be less likely to reclaim tree group(s) and more likely to break them. Let alone that the map generator can generate some pretty tree-heavy maps. This happens - and it happens often.

I see your points, but please also keep in mind that these changes basically only revert the LODs back to how they are are Steam. In fact, the LOD levels of some tree groups are higher on Steam. So, I am not sure what causes these changes to be an issue now, and I fear it is the experimental extended graphics option you use. 

The basic idea for this PR stems from galaticfear, and old player who liked the old tree group LODs more. Please note that the last time he played (with the old LOD values) tree breaking with scouts was already a thing.

https://discord.com/channels/197033481883222026/1089784297789083730/1350680864702333008
https://discord.com/channels/197033481883222026/1089784297789083730/1349441462479360063

I think he is right, the older LOD levels look nicer and are easier to play with. If this is an issue caused by the extended graphics option (which I do not run as it is labeled as `experimental` and `potentially unstable`) then, in my opinion, that is an issue with the setting and not the LOD levels.

Garanas:
> To be fair, this has already been the case before these changes.

Sure, but the difference is now minimized. That was the goal at the time. 

> I see your points, but please also keep in mind that these changes basically only revert the LODs back to how they are are Steam. In fact, the LOD levels of some tree groups are higher on Steam. So, I am not sure what causes these changes to be an issue now, and I fear it is the experimental extended graphics option you use.

The order of history is different. We initially sanitized the LODs in this pull request:

- https://github.com/FAForever/fa/pull/3662

Please note the two videos in that pull request. Based on feedback we provided players with an option to extent the rendering distance at their own risk:

- https://github.com/FAForever/fa/pull/3775

With this setting it would be roughly similar to what it was before in FAForever. This is why the option is called potentially unstable because that is exactly what it is. It does not guarantee a stable experience on all maps, or with all player ratings. Which brings me to my next point.

You mentioned GalacticFear. Based on his recent replay history of the past month he only appears to play Seton's Clutch. There are more maps then Seton's Clutch. The default game configuration should provide a stable game experience on all maps, for all player ratings and for all reasonable hardware configurations. The player ratings is a unique problem to Supreme Commander. If you do not reclaim, then the game still needs to work within reason.

If the extended graphics settings does not suffice for GalacticFear then I'd recommend him to write a custom UI mod to set the value to what he wants each game. It can be done through this UI function:

- `ConExecute("cam_SetLOD WorldCamera 0.70")`

The lower the number, the longer things keep rendering.

Basilisk3:
> Sure, but the difference is now minimized. That was the goal at the time.

It is still very easily doable and players do it all the time.

> With this setting it would be roughly similar to what it was before in FAForever. This is why the option is called potentially unstable because that is exactly what it is. It does not guarantee a stable experience on all maps, or with all player ratings. Which brings me to my next point.

Extended graphics also increases shadow fidelity and needs the rest of the settings to be set to high too. LOD is only one component of this and if the setting was not forcing you to play at high settings, the LOD would not be a problem to begin with. Realistically, I do not want to increase my graphics settings just to prevent a regression in the LODs of a couple of props. 

My testing indicates the performance issues only occur at high graphics settings, with shadows enabled. When a player has performance problems, lowering shadows is probably the first thing he does. For me and probably most people, the LOD changes caused a regression, but at the same time I have also never heard of FPS problems when too many trees were run over, not even when I was still new and players had worse PCs than they do now..

Also, this makes your point `Sure, but the difference is now minimized. That was the goal at the time.` not applicable, since then my PR effectively does the same, with the twist of enabling everyone to take advantage of the higher LODs, instead of only those that know about the advantages of this setting. Also, my PR should have better performance since I use the NormalMappedAlpha instead of the models for longer.

> You mentioned GalacticFear. Based on his recent replay history of the past month he only appears to play Seton's Clutch. There are more maps then Seton's Clutch. 

The only playable map I know that has more trees than Seton's is Selkie and that map essentially has broken amounts of trees.

> There are more maps then Seton's Clutch. 

I am fully aware that there is more than one map and when making these changes I consider and test more than just one map. Maps that have similar or more trees than Selkie I and most of the community will consider broken though.

> If the extended graphics settings does not suffice for GalacticFear then I'd recommend him to write a custom UI mod to set the value to what he wants each game. It can be done through this UI function:
> 
> * `ConExecute("cam_SetLOD WorldCamera 0.70")`
> 
> The lower the number, the longer things keep rendering.

I order for this to work properly, changes from this PR would be required, notably:

https://github.com/FAForever/fa/blob/521f13baa2b3f7a3dcf528745d40e41030fcdc47/lua/system/blueprints-props.lua#L130

Simply making a UI mod does not seem possible here, since the original code overrides the LOD of NormalMappedAlpha shader which is done in the sim state of the game. However when increasing LOD levels, you realistically want this shader to be displayed for as long as possible for performance reasons. In essence, all this PR does is display the NormalMappedAlpha for longer. 

My initial changes were similar to what you propose here. Gameplay wise they are regression since the NormalMappedAlpha is more useful than the model and has better performance.

lua/system/blueprints-props.lua
```diff
@@ -126,12 +126,16 @@ local function ProcessLOD(prop)
         for k = 1, n do
             local data = prop.Display.Mesh.LODs[k]
 
-            -- https://www.desmos.com/calculator (x * x)
-            local factor = (k / n) * (k / n)
-            local LODCutoff = factor * lod
+            -- Do not adjust the LOD of the 'NormalMappedAlpha' shader
+            if data.ShaderName ~= "NormalMappedAlpha" then
```
Garanas, line 130: Unrelated but related, the purpose of this function is to create cutoffs that go from low to high. This requires to update the value in the blueprint. Or there's no guarantee that the premise holds.

Besides, if you don't override the cutoff because the shader is `NormalMappedAlpha`. Then what is the value in practice? Whatever is defined in the blueprint? The default engine value?

Both are undesirable. We override it to guarantee a good performance.

Garanas (changes requested):
> The only playable map I know that has more trees than Seton's is Selkie and that map essentially has broken amounts of trees

> I am fully aware that there is more than one map and when making these changes I consider and test more than just one map. Maps that have similar or more trees than Selkie I and most of the community will consider broken though.

These are pretty vague statements. And in my point of view you also appear to use a logical fallacy by appealing to an anonymous authority. And I'd argue that it's not even true. As an example, take these maps:

Seton's Clutch: 			-> ~5500 props
Round 1. Phenom Spartiate v2.		-> ~6900 props
Round 2. Adaptive Moon. 		-> ~6400 props (no trees on map though)	
Round 3. Adaptive aloren. 		-> ~5100 props
Round 4. Adaptive Ndera. 		-> ~31000 props (!)
Round 5. Adaptive tabula rasa. 		-> ~24000 props (!)
Round 6. Adaptive Wonder Open. 		-> ~8000 props

These maps were played during the [Double Allied Tournament 4](https://forum.faforever.com/topic/8910/double-allied-tournament-4/1). I can't imagine these maps to be considered broken. Let alone that I already mentioned that generated maps can also have a relative high amount of tree groups. And people play these just fine, multiple times on a daily basis.

I'll also casually ignore that there are original maps that have more trees that Seton's Clutch. As an example, Syrtis Major (~6100) and Crossfire Canal (~21000) (that both have a FAF version) have more prop instances. And of those, the majority are trees just like in Seton's Clutch.

> Simply making a UI mod does not seem possible here, since the original code overrides the LOD of NormalMappedAlpha shader which is done in the sim state of the game. However when increasing LOD levels, you realistically want this shader to be displayed for as long as possible for performance reasons. In essence, all this PR does is display the NormalMappedAlpha for longer.

You're making a claim here, but I don't see any data to back it up. Based on my experience, why would the use of `NormalMappedAlpha` even be more efficient anyway when the issue is not the shader, it is the draw calls from the CPU to the GPU that is the bottleneck.

I understand you feel the need for the changes. And I feel the need to do a proper analysis of the impact of these changes to understand the impact it will have on the average player. With the proposed changes you're toying with the stability of the game. And I've not seen any data from you to back it up. This was done in the past. Some results of that are in the pull requests. Sadly, most of it were private messages between Strogoo and myself.

In my opinion it is more relevant to guarantee a proper performance across a wide range of game, hardware and player configurations over the convenience of seeing props sooner. Especially when there's another approach to that.

I'm going to reject these changes. Another maintainer can override that as they see fit. But I would be strongly against doing so without a proper analysis with plain and simple data about the impact of these changes.

Basilisk3:
> These are pretty vague statements. And in my point of view you also appear to use a logical fallacy by appealing to an anonymous authority. And I'd argue that it's not even true. As an example, take these maps:
> 
> Seton's Clutch: -> ~5500 props Round 1. Phenom Spartiate v2. -> ~6900 props Round 2. Adaptive Moon. -> ~6400 props (no trees on map though) Round 3. Adaptive aloren. -> ~5100 props Round 4. Adaptive Ndera. -> ~31000 props (!) Round 5. Adaptive tabula rasa. -> ~24000 props (!) Round 6. Adaptive Wonder Open. -> ~8000 props

There is no logic fallacy. Selkie is widely considered to have a broken number of trees, this is a fact.

I tested all maps you listed and they all run fine, even with experimental graphics enabled. This is because, even though spamming thousands of single trees everywhere is terrible map design, the trees are spread out relatively well, especially on Ndera. It's on Seton's where I experienced the same frame dips you did, probably because they are all groups. Also, if the game runs badly, turning down the settings is the first thing I do in a videogame.

> You're making a claim here, but I don't see any data to back it up. 

The `NormalMappedAlpha` corresponds to the highest LOD level before the trees disappear. It is definitely less detailed than the previous levels, so I made the reasonable assumption that this LOD level has better performance than the (far more detailed), previous one.

> With the proposed changes you're toying with the stability of the game.

In short, no, and I take offense with this statement. Calling this "toying" is an extreme mischaracterization of what I am doing.

> In my opinion it is more relevant to guarantee a proper performance across a wide range of game, hardware and player configurations over the convenience of seeing props sooner.

This is also a mischaracterization. Seeing props sooner is not for merely for "convenience" it's gameplay relevant. Identifying tree groups and larger forests is very important, especially on the maps you listed above. Also note, that you provide no proof that these performance problems have ever been reported anywhere, even though they have been in the game for 15 years, if not longer.

Basilisk3:
Closing this as it has been rejected. It is disappointing to see a gameplay-relevant regression compared to the Steam version get introduced for all players and then have player feedback and a proposed remedy get dismissed after just three days.

Garanas:
> In short, no, and I take offense with this statement.

There's no need to take offense. I toy around all the time. It's just a way of phrasing. Don't look too deep into it.

> Also note, that you provide no proof that these performance problems have ever been reported anywhere, even though they have been in the game for 15 years, if not longer.

I did. I pointed you to the video's in question. See also this pull request:

https://github.com/FAForever/fa/pull/3662
The whole reason that pull request exists is because people told me about the problem. And I investigated it together with Strogoo at the time. I don't see how that is a 'lack of proof' that the problem exists, but I do agree that the pull request is a little undocumented.

That was quite common until recently.

Note that I'm just another maintainer. I'm not the team lead. Closing the pull request is premature. If another maintainer disagrees with my statements then that's fine and these changes can still get merged in.

I just don't think anyone should make this type of changes without careful consideration of various maps and the impact on those maps. Together with Strogoo we investigated it extensively. And we noticed some significant issues. Some were fixed through an assembly patch. As an example:

https://github.com/FAForever/FA-Binary-Patches/pull/7
Others are fixed through pull requests such as #3662. These are just addressing the symptoms, not the root cause. The root cause is the lack of modern instancing to reduce draw calls. That is sadly out of reach. Not just because it would require an extremely difficult binary patch. But also because we don't use the latest and greatest of DirectX.

With all of that said - I asked you to provide data. I asked two times. You still did not do that. These changes are measurable. I even told you how to measure it. The whole discussion would be quite different from my point of view if we were talking about data.

But now all I could talk about were your words and assumptions. That immediately makes it needlessly more personal.

Basilisk3:
> The whole reason that pull request exists is because people told me about the problem. 

The same can be said for this PR. I personally did not notice this change when it was introduced somehow, but when I decided to start the Steam version of the game, I immediately realized what galacticfear meant; our current LOD levels for trees are much inferior in terms of playability.

I also think you undervalue the importance of the changes in general. Tree groups are a major aspect of the game and making them less visible for one of the most common zoom levels can have massive implications, especially for newer players who will have a harder time learning them.

> And I investigated it together with Strogoo at the time. I don't see how that is a 'lack of proof' that the problem exists, but I do agree that the pull request is a little undocumented.

This is not what I am talking about and you are leaving out important context. The complaints you are talking about were mostly caused by other **major** changes in prop LOD behavior (see https://github.com/FAForever/fa/pull/2721 among others; this one changed over `10000` lines of code). It is disingenuous to leave out this information, because the changes you mention are not a direct response to the base game. Your data is based on a previous, major rework of props, which I know caused problems because it increased the LODs of various props. In a way it is a bug fix or a refinement of previous changes, it is **not** proof that the original LOD levels (the one's my PR is based on and aims to return to) caused issues to warrant this kind of nerf.

> With all of that said - I asked you to provide data. I asked two times. You still did not do that. These changes are measurable. I even told you how to measure it. The whole discussion would be quite different from my point of view if we were talking about data. But now all I could talk about were your words and assumptions. That immediately makes it needlessly more personal. 

I do not have the testing setup to reliably track FPS numbers and deviations from it. My preliminary testing indicated a negligible decrease in FAF unless very high settings are used and 50 Spirits break every tree group in sight.

I edited your post by accident, it should be restored now.

BlackYps:
I don't think performance data can solve this discussion. The fps impact is minimal until the render thread saturates one core of the cpu and then the fps start to tank. Depending on the fps overhead you have with your gpu you might feel this immediately or only later. The point is, it depends massively on your hardware setup and the type of games you play. I suspect the variance here is so high that we can find a high lod cutoff value where we can reasonably assume that it works for the vast majority of players. Instead I propose a different approach:
In my view we should make this a graphics option. We should set this to a medium value by default and then people can set it to a higher value depending on their preference.
The experimentel graphics option was a nice addition, but it bundles different settings that are not necessarily related which is bad. When we split these settings we allow players to set the individual graphics settings to values that best fit their needs. This way we don't have to spend a lot of energy trying to find the ideal cutoff value that we believe most machines can still handle.

Garanas:
> I don't think performance data can solve this discussion. The fps impact is minimal until the render thread saturates one core of the cpu and then the fps start to tank.

This is not true. If you disable vsync and remove the frame time clamp then you can measure it as the trees break. As an example:

https://github.com/user-attachments/assets/1e08f363-93fd-423c-86c1-40c0d1c17afd

And because we're talking about draw calls, it is irrelevant whether you have the latest and greatest or something that is decent. It only becomes relevant when you're in the below-decent area where draw calls get more expensive then usual. See also this topic:

- https://stackoverflow.com/questions/4853856/why-are-draw-calls-expensive

I don't think I agree with the existence of game options that just don't work on the majority of maps if the map is played 'poorly' (by not reclaiming). I agree that the experimental game option is not optimal, but at least it makes it clear that you're on your own.

Basilisk3:
> I agree that the experimental game option is not optimal, but at least it makes it clear that you're on your own.

I see the following options how we could tackle this:
- Make it a boolean option. As in: `Experimental: Increase LOD of treegroups`. The default should be false.
- Additionally we could add a performance notice.

The problem is, that this is not easily done in the way I wanted, which is why I will keep this on hold indefinitely.

@BlackYps Some more context on my process that is relevant here:

While I was testing, I tried out the experimental graphics setting. It yielded suboptimal results as it basically increases every LOD level, which means the more detailed tree models are also displayed earlier. This plays worse because the last LOD level is what you want for identifying tree groups (at least that is how I do it), and I also assume this is worse performance wise. This is why I went with excluding the `NormalMappedAlpha` shader from the LOD computation here: 

https://github.com/FAForever/fa/pull/6715/files#diff-afa305d131383c60aebe3bbe028d3d6e3edb36871123e926095af45e118a5dc7R130. 

In essence, I only wanted to display the last LOD level for longer, with the other LOD levels remaining unchanged. I managed to achieve this with this PR, the least detailed tree model remains rendered for longer. The changes more or less bring FAF back in line with Steam or with how FAF used to render tree groups.

Please also note that I **did** performance tests on multiple maps. At the highest graphics setting and if scouts systematically break most tree groups, there is a noticeable performance impact on Seton's.

BlackYps:
> This is not true. If you disable vsync and remove the frame time clamp then you can measure it as the trees break. 

Yes, you can measure it _on your machine_. What I meant is that we don't have data about the hardware that people use. We can accurately determine the lod value that starts creating issues on a particular machine, but we can't confidently generalize it to what value would work for _most_ machines. That's the issue I was getting at. And I don't think we can solve this. But we don't have to. We can just give people options and they can decide for themselves. Many games allow you to crankup the settings to a level where most hardware (that was common at the release of the game) will struggle. If we expose an option to tweak the tree lod (or alternatively if we can't have it affect only trees, the general lod) and set the default to a sane value, for example the value we have now, then people can dial in a value that works for them. If we couple this with a notice that this setting is very performance intensive, then I don't think we do anything unusual.

The experimental game option is suboptimal because it bundles the shadow and lod settings that don't have to be bundled and it doesn't describe well what it actually does. But if we use this as a framework and split the experimental option into two settings we are almost at the solution I described above.

BlackYps:
> The problem is, that this is not easily done in the way I wanted, which is why I will keep this on hold indefinitely.

Can you elaborate what problems you encountered?

I understand the intent to only increase the last LOD level. I just think that the way you coded it is suboptimal. First of all, it's not obvious that the NormalMappedAlpha shader is the last LOD and why that should be treated differently, so from reading the code it's hard to understand why this is happening. You added a comment in the code, but there you basically explained what the next line does, instead of explaining _why_ you do this.
Secondly, this opens the theoretical possibility that the last LOD has a lower value than the previous LOD.

Thinking about this whole thing more, why do we recalculate the LOD values here at all? Isn't it more straightforward to have these values in the blueprints? Then you know the actual values that will be used in the game instead of whatever values are in the blueprint file. That's a question for @Garanas I suppose, as you made the feature.

Basilisk3:
> Can you elaborate what problems you encountered?

Jip told me this, you are better off asking him I think.

> I just think that the way you coded it is suboptimal.

This is true, this was the first time I have ever worked with LOD values and I am still learning.

> Thinking about this whole thing more, why do we recalculate the LOD values here at all?

The LOD values in the blueprints are overridden in the file I changed.

Garanas:
> Yes, you can measure it on your machine. What I meant is that we don't have data about the hardware that people use. We can accurately determine the lod value that starts creating issues on a particular machine, but we can't confidently generalize it to what value would work for most machines.

This is not how draw calls work. They're slow on all machines. A very fast GPU or CPU barely have an effect on that. But I'm not going to argue this fact again.

> Thinking about this whole thing more, why do we recalculate the LOD values here at all? Isn't it more straightforward to have these values in the blueprints? Then you know the actual values that will be used in the game instead of whatever values are in the blueprint file. That's a question for @Garanas I suppose, as you made the feature.

At the time there was no easy way to bake in these values. Changing the LOD value for hundreds, if not thousands of fields was not an option. Yet, the game very clearly broke on some maps. See also https://github.com/FAForever/fa/issues/3662. 

With thanks to @The-Balthazar we now have a method of mass processing blueprint files. I explored this  last year. It's done by running his [Wiki generator](https://github.com/The-Balthazar/BrewWikiGen) which has a side effect of being able to load, edit and then re-write all the blueprints. 

It appears I already did a run of baking in the values: https://github.com/FAForever/fa/pull/6274 .

It also appears that preparing for batch processing of props is why we have the LOD-logic happening twice during blueprint loading.

BlackYps:
I would like to schedule a voice call with you to talk about the technical limitations we have to consider. I feel this will work better than trying to work out the details via text. I think when we did this, then we can plan a way forward. The approach in this PR has issues, as we have discussed already, but I think there is a possible path that can improve the situation while addressing the raised concerns. Maybe @Basilisk3 wants to join as well, if he wants to get some additional info to tackle the future implementation.

PR #6309, opened 2024-06-26
Implement variable transport speeds

Basilisk3:
Implement transports moving at different speeds depending on how many and which types of units they have loaded. Closes #6186.

## Description of the proposed changes
- Each Tech 1 unit decreases the transport's MaxAirspeed by `0.15`
- Each Tech 2 unit decreases the transport's MaxAirspeed by `0.3`
- Each Tech 3 unit decreases the transport's MaxAirspeed by `0.6`
- ACUs decrease the transport's MaxAirspeed by `1`

## Stat changes:
**Transport**
MaxAirspeed: 10 --> 10.75 (Tech 1 Transports)
MaxAirspeed: 13.5 --> 14.5 (Tech 2 Transports)
MaxAirspeed: 15 --> 17.5 (The Continental)

**Land units**
TransportSpeedReduction: 0.15 (Tech 1 land units)
TransportSpeedReduction: 0.3 (Tech 2 land units)
TransportSpeedReduction: 0.6 (Tech 3 land units)
TransportSpeedReduction: 1 (ACUs and SACUs)
TransportSpeedReduction: 1 (Tech 4 land units, to ensure compatibility with survival maps)
TransportSpeedReduction: 1 (naval units, to ensure mod compatibility)

The `TransportSpeedReduction` stat can be tweaked for each unit individually, though I did not do it here for consistency reasons. For example, you can make LABs lighter than tanks. The `TransportSpeedReduction` stat has also been made visible in the UI in the additional unit details displayed when `Show Armament Detail in Build Menu` is enabled.

## Additional context
- As Jip already pointed out, the change introduces a new balance knob.
- Enables more teamplay, as getting an empty transport to your teammates is faster.
- The changes would introduce a new choice for the player, allowing him to decide between a swift and less damaging drop or a slower but more damaging one.
- Nerfs arty drops, specifically Seraphim transports fully loaded with Zthuees, in an intuitive way. As Seraphim has the highest capacity transports (and the strongest Tech 1 Artillery), I regard this as a positive.

## Checklist
- [x] ~~Populate the `TransportSpeedReduction` blueprint field of all land units~~ (alternative approach via blueprints-units.lua)
- [x] Changes are annotated, including comments where useful
- [x] Changes are documented in the changelog for the next game version

PR #6720, opened 2025-04-06
Remove redundant sonar and radar stats from various units

Basilisk3:
Currently, several units have a `SonarRadius` that is smaller than their `WaterVisionRadius`. This makes their sonar pointless and adds another range ring that is not useful. There are two options: 
- Remove the sonar
- Make it large enough to where it matters

I think we should remove the sonar in cases where a larger radius would increase the amount of intel the unit provides too much (ideally, sonar should be a bit larger than water vision).

**Examples:** 
- The Torrent has a `WaterVisionRadius` of `48` but only a `SonarRadius` of `20`. I would propose removing the sonar in this case.
- The Salem Class has a `WaterVisionRadius` of `44` but only a `SonarRadius` of `32`. I think both removing the sonar or buffing it could make sense here.

Also increase the Salem's radar to match the other Destroyers.

## Checklist


lL1l1:
Buffing sonar to match water vision would allow those units to see surface ships at a greater distance, since for some reason water vision only applies underwater, so I would go with removing the sonar to maintain the current balance. Destroyers do have a radar that allows them to detect surface ships, but it's only 36 or 40 range (this difference is a mistake made when fixing the ranges for the intel grid, it used to be 35), which is much less than the 44 water vision range, considering how much it would change the effectiveness of frigates at closing in on lone destroyers to buff the sonar that much.

The water vision for destroyers roughly matches their torpedo range, so it's not like it is completely invisible as it often is.

Basilisk3:
I agree, I'll remove the sonars.

Garanas:
Removing the sensors may not negatively impact the base game, but it can impact mods that work with cloak (fields).

Basilisk3:
The counter for cloaking fields would be making sonars (448 radius) or units with proper sonar capabilities. I do not believe anyone uses Battleship (20) or Destroyer (36) sonars for countering cloak fields.

Garanas:
No, but the sonar is a prime target to snipe of course. I would not built a Destroyer (or Battleship) for its sonar capabilities. But it can be useful to have. And they also have it on Steam. 

Where does this change originate from? Was it a request of the balance team?

Basilisk3:
> No, but the sonar is a prime target to snipe of course.

Sure, but even with this PR, you still have plenty of units that offer sonar capabilities. Cruiser sonar has a radius of 120, Tech 2 subs have 60, even Tech 1 subs have 60. So I highly doubt this would impact modded games with cloaked naval units basically at all.

I also think an argument can be made that, from a gameplay perspective, this change actually improves gameplay for these modded lobbies. Now not every random naval unit has sonar anymore, which is something you can tactically play around and consider when microing your units. The balance team made similar changes (e.g. GC omni removal) to promote the use of actual counters, instead of random units having one.

> I would not built a Destroyer (or Battleship) for its sonar capabilities. But it can be useful to have. 

See above. A Destroyer's or a Battleship's sonar coming in handy is extremely unlikely, given that a single Cruiser covers 10x the area of a Battleship's sonar at less than a quarter of its cost. Also, sniped sonars are always quickly rebuilt, they are necessary to have proper engagements from the mid Tech 2 stage and beyond.

> Where does this change originate from? Was it a request of the balance team?

It was my idea.

BlackYps:
I understood the main reason of the PR as removing visual clutter in the range rings and I agree with that.
From my impression with unit mods they generally don't seem like they are carefully crafted to be balanced with the current unit capabilites. And if a mod really relies on this a lot then it can just readd the sonar capabilites on the base units again. So I think it's fine to make this change.
I'll add the balance label, so the balance team knows to have a look at this, although I don't think they will have a problem with the idea. But I want to make sure that we don't accidentally overlook an important gameplay interaction.

lua/system/blueprints-units.lua
```diff
@@ -328,6 +328,22 @@ local function PostProcessUnit(unit)
         end
     end
 
+    -- Check if a unit's WaterVisionRadius is larger than its SonarRadius
+    if unit.Intel.WaterVisionRadius and unit.Intel.SonarRadius then
+        if unit.Intel.WaterVisionRadius > unit.Intel.SonarRadius then
```
Hdt80bro: As pointed out by @lL1l1, the `SonarRadius` needs to be smaller than both the `WaterVisionRadius` and the `VisionRadius` or `RadarRadius`. Thus, most of these changes are incorrect.

Hdt80bro (changes requested):

Hdt80bro:
I don't like the thought of removing functionality from some units just because it looks cluttered when you click on it. It some cases it's crucial - for example, with these changes, the DN1 now gets demolished by frigates because it can't see them on sonar. But it can still shoot submarines just fine.

Basilisk3:
Thanks for your review @Hdt80bro, I made a mistake and changed a few units I should not have. I will fix that in the coming days. The goal is to have no change in functionality.

BlackYps:
> the DN1 now gets demolished by frigates

What's the DN1?

Hdt80bro:
That is the name of the UEF T1 torpedo defense.

Hdt80bro:
I just remembered that sonar & radar can detect cloaked units while vision cannot - so there can't ever be a way to truly ever do this without removing some functionality unless there's omni involved.

Basilisk3:
As for the sonar changes, there is no change in functionality in the base game. Concerning the radar changes, only the tiny radars of Battleships are removed, which will basically never matter the base game (the Mole and Fire Beetle have cloak but not stealth; Battleship radar is too small to matter for these units though).

Regarding if this is good/bad for modded games, see here: https://github.com/FAForever/fa/pull/6720#issuecomment-2785915396.

Hdt80bro:
Well, the Cybran Commander can be cloaked, so I wouldn't say removing sonar has no functionality loss in the base game, but other than that, my reservations are based on immersion and the large-scale nature of naval combat. If it feels like only random units were given sensors (which I'm unsure about), I'd rather give them to all naval units first than take them away from only some of them. I'd need to take a closer look, but I think I prefer extending the range in second place.

Your comments about mods don't hold water for me, so I'd like input from someone with more experience with them. Maybe @The-Balthazar can give us feedback?

Basilisk3:
> Well, the Cybran Commander can be cloaked

The stealth upgrade is a prerequisite for the cloak upgrade, so the changes will not affect it.

> I'd rather give them to all naval units first [...]

The vast majority of naval units have sonar. The only exceptions I can think of are the Shard and the Bulwark.

In my opinion, a stat should serve a purpose - the sonar of these units does not serve one, so it should either be made useful or removed. Removing has the advantage of not incurring any balance changes and removing range rings.

Hdt80bro:
Ah, you're right about the commander

Hdt80bro:
> The vast majority of naval units have sonar. The only exceptions I can think of are the Shard and the Bulwark.

Then I understand even less why this won't affect mods. I've talked to Blathazar and he confirmed that this would break some design considerations that mods have made, such as for cloaked mines or cloaked amphibious units. These stats serve a purpose, and it's to create a consistent intel environment that mods can play with. It would be unfair to them to suddenly change that and make having naval sonar an inconsistent thing just because it happens to not matter to the base game and can be optimized out.

Basilisk3:
> I've talked to Blathazar and he confirmed that this would break some design considerations that mods have made, such as for cloaked mines or cloaked amphibious units. 

I think this is a stretch. If almost every unit has sonar, then that means basically every unit also hard-counters mines. If these mines were implemented in the base game, no one would use them. Similarly, if Pillars or Mongoose had radar, that would reduce the number of Fire Beetles built to approximately zero.

Also, a single Tech 3 sonar or even a single cruiser hard counters these units anyway, unless you also stealth them, at which point these changes would not alter the interaction anyway, since you would then need an omni. 

Why does Battleships having a sonar radius of 20 matter for mines? It is so low, it basically cannot matter, even in the scenarios you lined out. And increasing the radius like you suggested earlier would break the balance of these mods even more, because now every naval unit is even more of a hard counter. Please also note that this would make the modded unit **perform worse**, while my changes theoretically make them sightly **more effective**.

> It would be unfair to them to suddenly change that and make having naval sonar an inconsistent thing just because it happens to not matter to the base game and can be optimized out.

Overall, I do not find these arguments convincing, because by the time any of these units arrive, you, in all likelihood, have either a sonar, a single Cruiser, Frigates or a even a sub. I can however see how mines might gain a slight indirect buff against pure Destroyer mixes that lack sonars, cruisers, and subs. I have extensive experience with navy and no one I know plays like this.

Additionally, basically every balance change of the previous patch altered the game's interaction with mods in some way. In the past there were even changes like the GC tractor beam fix, that broke the balance of survival maps and modded Tech 3 units. My changes do not even approach this magnitude of disruptiveness, at most they are a small buff to the modded units, which by itself is way less disruptive than making their usage actively more difficult.

If you want Destroyers to gain a larger sonar radius instead, then I think you can make that case. But please also consider that now you are **nerfing** the modded units and that **base game balance implications** arise from this.

The-Balthazar:
Stubling over chestertons fence so hard it hurts.

Basilisk3:
> Stubling over chestertons fence so hard it hurts.

Chesterton's fence implies that one removes an institution without considering adverse consequences. Here, have a look at the first result on Google: https://theknowledge.io/chestertons-fence-explained/. So no, this does not apply here, since I **did** consider adverse consequences, gave every scenario the benefit of the doubt and explained in detail why the above considerations do not apply for the base game, and are also barely relevant for modded games.

The-Balthazar:
That article doesn't support your narrow read of the principle, and doesn't change the fact that you're messing with numbers that are only important to an environment you have no experience with, for next to no benefit outside of that environment.

The-Balthazar:
The concept that because a given thing has no use in the base game is, frankly, hostile to modding in general.

Basilisk3:
> That article doesn't support your narrow read of the principle, 

My read of the principle captures its essence, while your interpretation conveniently ignores the fact that adverse consequences of actions need to remain unconsidered for it to apply.

> and doesn't change the fact that you're messing with numbers that are only important to an environment you have no experience with

You do not know me, you do not know my in-game name, you do not know for how long I have been playing, you do not know which games I play. Maybe consider making fewer assumptions and spawn in a Summit or Torrent to see for yourself if it is reasonable to assume that removing their sonars will break mods.

> The concept that because a given thing has no use in the base game is, frankly, hostile to modding in general.

Can you please explain how a Summit (10000 mass each) having a `SonarRadius` of 20 is important for mods? Do you realize that this radius is smaller than a vanilla ACU's gun range? The fact that I even need to ask this already tells me everything.

Also, the narrative you are trying to spin, that I somehow do not care about mod compatibility is easily disproven by this PR, which is fully compatible with mods and even takes survival maps into account: https://github.com/FAForever/fa/pull/6309.

The-Balthazar:
I am aware of how small it is, and yes, I do believe that that sonar radius of 20 has more value to mods than the value added from simplifying the base game by removing it. There are mods that can buff intel values, and mods that add cloaked ships, subs, and other amph things that would suddenly be able to take out a lone Summit that it could previously detect and in some of those cases retaliate against.
Removing it is close to pointless at best, and hostile at worst.

Basilisk3:
Let me rephrase my question, can you please explain how 20 sonar radius on a Summit is relevant for mods when a Tech 3 sonar has a sonar radius of 448 for 1/10 of the price? Feel free to exchange "Tech 3 sonar" with "any sub", "cruisers" or even "frigates"´, which you usually have in front of your navy.

> suddenly be able to take out a lone Summit that it could previously detect and in some of those cases retaliate against.

Lone Summits usually do not exist, and when they do, that is a misplay of gigantic proportions. I do not assume everyone plays perfectly, but these kinds of misplays are not (and cannot) be taken into account for game balance, because they display a fundamental misunderstanding of naval combat. And even if they were taken into account, this also assumes that the player does not have a single sonar, a cruiser or even a sub remotely nearby, which is a another skill-related issue.

> There are mods that can buff intel values

In this case, the Summit's sonar would **matter even less**, considering units with larger radii (so almost every other unit in the game), **benefit more** than units with low sonar values like the Summit.

As a compromise I propose removing the sonar from units where it makes absolutely no sense (Battleships) but in turn increase it for Destroyers. I don't like this, and, as pointed out by Nomander, this changes the balance of the base game.

Hdt80bro:
> Lone Summits usually do not exist, and when they do, that is a misplay of gigantic proportions. I do not assume everyone plays perfectly, but these kinds of misplays are not (and cannot) be taken into account for game balance, because they display a fundamental misunderstanding of naval combat. And even if they were taken into account, this also assumes that the player does not have a single sonar, a cruiser or even a sub remotely nearby, which is a another skill-related issue.

I thought this was a discussion about loss of functionality for mods, not game balance. I agree that these changes shouldn't affect gameplay, but that's not where my qualms come from.

Basilisk3:
> > Lone Summits usually do not exist, and when they do, that is a misplay of gigantic proportions. I do not assume everyone plays perfectly, but these kinds of misplays are not (and cannot) be taken into account for game balance, because they display a fundamental misunderstanding of naval combat. And even if they were taken into account, this also assumes that the player does not have a single sonar, a cruiser or even a sub remotely nearby, which is a another skill-related issue.

> I thought this was a discussion about loss of functionality for mods, not game balance. 

My argument is that if you play with these mods, the loss of sonar radius will only affect you if you do not build sonars, or have almost any other naval unit near the Battleship. In other words, basically if someone does not really know how to play. Having isolated Summits and no sonar or other nearby ships is akin to only building Tech 1 arty and never building scouts or radars.

This is not something that can be taken into account for making balancing decisions (which this is), because at that point, the player is doing all kinds of mistakes, and a sonar radius of 20, **on a battleship of all units**, is the least of his concerns. This sounds harsh but is not meant to be taken as an offense to new or bad players, obviously. The balance team operates in a similar manner, e.g. it cannot take players that do not build Spirits into account when balancing the Aurora.

Hdt80bro:
Ah, my opinion is that we should not balance the base game for mod authors. That is their own job. If we try to things for them based on speculation, then we run the risk of interfering with a mod's decisions.

The-Balthazar:
This is a change that makes basically no difference to the base game, and some difference to mods. To me that says you should not make the change. Saying "oh but that's not likely to happen even in modded games" assumes a lot about the environment those mods create that I don't think you should be assuming. Why are you so against letting them have their token abilities?

Basilisk3:
> This is a change that makes basically no difference to the base game

Just because it is a detail does not mean it makes basically no difference. It is important that the UI only conveys important information and not random stats that do not make any sense and were placed there with little consideration. Sonar, radar, sonar stealth, radar stealth, vision and watervision are already difficult concepts to understand and we should not keep them artificially convoluted by including things that do not make any sense.

> Saying "oh but that's not likely to happen even in modded games" assumes a lot about the environment those mods create that I don't think you should be assuming.
 
I already explained my case above. My assumptions are based on extensive experience with both navy and heavily modded games. You, however, assume that players basically only build a single type of naval unit and no sonars (which is the prerequisite for these super weak sonars to matter). I try to keep my changes as least disruptive to mods as possible, but this is not something I can consider.

I also do not see how this _breaks_ any mods, the absolute worst that could happen is that a few units are slightly more effective, if someone really does not know what he is doing. Virtually every single balance change also does this.

speed2CZ (approved):
Looks fine.

I highly doubt this would be a mod breaking change.

BlackYps:
Looking at this more, I noticed that the vision and water vision radius differs considerably sometimes. As far as I can see there is no visual indication of water vision in the game. There is just the text in the unit tooltip. And water vision only applies to submarines and not to normal ships, is that correct?
Before we come to a conclusion here I would like to have a better understanding of the situation and to see some consideration of alternative approaches.
Is there a specific reason that the relation between vision, water vision and sonar are so different on different units, or has the balance team just never touched them because it wasn't important enough?  There is also the question if it is even desirable to have different values for vision and water vision.
Instead of removing the sonar completely on some ships, wouldn't it make more sense to set the radius to the water vision radius if the sonar radius was smaller before?

Basilisk3:
> Is there a specific reason that the relation between vision, water vision and sonar are so different on different units, or has the balance team just never touched them because it wasn't important enough? There is also the question if it is even desirable to have different values for vision and water vision.

Aligning water vision with vision is not a bad idea per se, but I think it would be difficult to get right, at least for certain subgroups of units.

Take the Seraphim sub hunter as an example, it has a high water vision radius. Increasing its vision to match its watervision radius would be a significant buff to the unit overall.

lL1l1:
> Take carriers as an example...

Carriers already have 100 vision + 100 underwater vision.

Here are all the units with less vision than water vision. Also notice that a lot of them have less sonar than water vision too.
```
AEON     TECH1        Torpedo Launcher               Tide                           Water vision 44 Vision 12 (-32) Sonar 36 (-8)
AEON     TECH2        Destroyer                      Exodus Class                   Water vision 44 Vision 32 (-12) Sonar 36 (-8)
AEON     TECH2        Submarine Hunter               Vesper                         Water vision 44 Vision 32 (-12) Sonar 60 (+16)
AEON     TECH2        Torpedo Launcher               Wave Break                     Water vision 44 Vision 20 (-24) Sonar 36 (-8)
AEON     TECH3        Strategic Missile Submarine    Silencer                       Water vision 36 Vision 16 (-20) Sonar 60 (+24)
CYBRAN   TECH1        Torpedo Launcher               Scuttle                        Water vision 44 Vision 12 (-32) Sonar 36 (-8)
CYBRAN   TECH2        Destroyer                      Salem Class                    Water vision 44 Vision 32 (-12) Sonar 32 (-12)
CYBRAN   TECH2        Torpedo Launcher               Nanite Torpedo Array           Water vision 44 Vision 20 (-24) Sonar 36 (-8)
CYBRAN   TECH3        Strategic Missile Submarine    Plan B                         Water vision 36 Vision 16 (-20) Sonar 60 (+24)
CYBRAN   TECH3        Torpedo Ambushing System       HARMS                          Water vision 44 Vision 20 (-24) Sonar 40 (-4)
CYBRAN   TECH3        Torpedo Ambushing System       HARMS                          Water vision 44 Vision 20 (-24) Sonar 40 (-4)
SERAPHIM TECH1        Torpedo Launcher               Sou-atha                       Water vision 44 Vision 12 (-32) Sonar 36 (-8)
SERAPHIM TECH2        Destroyer                      Uashavoh                       Water vision 44 Vision 32 (-12) Sonar 36 (-8)
SERAPHIM TECH2        Torpedo Launcher               Uosthu                         Water vision 44 Vision 20 (-24) Sonar 36 (-8)
SERAPHIM TECH3        Siege Tank                     Othuum                         Water vision 24 Vision 22 (-2) Sonar 00 (-24)
SERAPHIM TECH3        Submarine Hunter               Yathsou                        Water vision 56 Vision 32 (-24) Sonar 88 (+32)
UEF      TECH1        Torpedo Launcher               DN1                            Water vision 44 Vision 12 (-32) Sonar 36 (-8)
UEF      TECH2        Destroyer                      Valiant Class                  Water vision 44 Vision 32 (-12) Sonar 36 (-8)
UEF      TECH2        Torpedo Boat                   Cooper                         Water vision 56 Vision 32 (-24) Sonar 36 (-20)
UEF      TECH2        Torpedo Launcher               Tsunami                        Water vision 44 Vision 20 (-24) Sonar 36 (-8)
UEF      TECH3        Battlecruiser                  Neptune Class                  Water vision 48 Vision 40 (-8) Sonar 20 (-28)
UEF      TECH3        Strategic Missile Submarine    Ace                            Water vision 36 Vision 16 (-20) Sonar 60 (+24)
```

Here are the units with more vision than water vision, excluding many +2 water vision intel grid rounding errors. I also included units with less sonar range than water vision (carriers, fatboy, battleships, wagner).
```
info: AEON     TECH1        Frigate                        Beacon Class                   Water vision 16 Vision 32 (+16) Sonar 16 (+0)
info: AEON     TECH2        Cruiser                        Infinity Class                 Water vision 60 Vision 64 (+4) Sonar 120 (+60)
info: AEON     TECH3        Aircraft Carrier               Keefer Class                   Water vision 100 Vision 100 (+0) Sonar 40 (-60)
info: AEON     TECH3        Battleship                     Omen Class                     Water vision 48 Vision 50 (+2) Sonar 20 (-28)
info: AEON     TECH3        Missile Ship                   Torrent Class                  Water vision 48 Vision 70 (+22) Sonar 20 (-28)
info: AEON     TECH3        Sonar Platform                 nil                            Water vision 24 Vision 32 (+8) Sonar 448 (+424)
info: CYBRAN   COMMAND      Armored Command Unit           nil                            Water vision 24 Vision 26 (+2) Sonar 24 (+0)
info: CYBRAN   TECH1        Frigate                        Trident Class                  Water vision 16 Vision 32 (+16) Sonar 16 (+0)
info: CYBRAN   TECH2        Amphibious Tank                Wagner                         Water vision 20 Vision 20 (+0) Sonar 00 (-20)
info: CYBRAN   TECH2        Cruiser                        Siren Class                    Water vision 60 Vision 64 (+4) Sonar 120 (+60)
info: CYBRAN   TECH3        Aircraft Carrier               Command Class                  Water vision 100 Vision 100 (+0) Sonar 40 (-60)
info: CYBRAN   TECH3        Battleship                     Galaxy Class                   Water vision 48 Vision 50 (+2) Sonar 20 (-28)
info: CYBRAN   TECH3        Sonar Platform                 Flood XR                       Water vision 24 Vision 32 (+8) Sonar 448 (+424)
info: SERAPHIM TECH1        Frigate                        Hau-esel                       Water vision 16 Vision 32 (+16) Sonar 16 (+0)
info: SERAPHIM TECH2        Cruiser                        Ithalua                        Water vision 60 Vision 64 (+4) Sonar 120 (+60)
info: SERAPHIM TECH3        Aircraft Carrier               Iavish                         Water vision 100 Vision 100 (+0) Sonar 40 (-60)
info: SERAPHIM TECH3        Battleship                     Hauthuum                       Water vision 48 Vision 50 (+2) Sonar 20 (-28)
info: UEF      EXPERIMENTAL Experimental Mobile Factory    Fatboy                         Water vision 32 Vision 32 (+0) Sonar 00 (-32)
info: UEF      TECH1        Frigate                        Thunderhead Class              Water vision 16 Vision 32 (+16) Sonar 16 (+0)
info: UEF      TECH2        Cruiser                        Governor Class                 Water vision 60 Vision 64 (+4) Sonar 120 (+60)
info: UEF      TECH3        Battleship                     Summit Class                   Water vision 48 Vision 50 (+2) Sonar 20 (-28)
info: UEF      TECH3        Sonar Platform                 SP3 - 3000                     Water vision 24 Vision 32 (+8) Sonar 448 (+424)
```

Basilisk3:
> > Take carriers as an example...
> 
> Carriers already have 100 vision + 100 underwater vision.

I updated my comment, I wanted to write about subs initially but then went with carriers for some reason.

> ```
> AEON     TECH1        Torpedo Launcher               Tide                           Water vision 44 Vision 12 (-32) Sonar 36 (-8)
> AEON     TECH2        Destroyer                      Exodus Class                   Water vision 44 Vision 32 (-12) Sonar 36 (-8)
> ```

Out of curiosity, you gathered this data by including a log statement roughly similar to mine, correct?

lL1l1:
> Out of curiosity, you gathered this data by including a log statement roughly similar to mine, correct?

Yes, I added the following to `PostProcessUnits` in `blueprints-units.lua`.
```lua
    for _, unit in units do
        if not unit.CategoriesHash['INSIGNIFICANTUNIT'] and not unit.CategoriesHash['UNSPAWNABLE']
            and unit.Intel.VisionRadius and unit.Intel.WaterVisionRadius
            and (
                unit.Intel.VisionRadius ~= unit.Intel.WaterVisionRadius
                or (unit.Intel.SonarRadius or 0) < unit.Intel.WaterVisionRadius
            )
        then
            LOG(string.format('%-8s %-12s %-30s %-30s Water vision %02d Vision %02d (%+02d) Sonar %02d (%+02d)'
                , tostring(unit.FactionCategory)
                , tostring(unit.TechCategory)
                , tostring(unit.Description):gsub("<LOC .*>", '')
                , tostring(unit.General.UnitName):gsub("<LOC .*>", '')
                , unit.Intel.WaterVisionRadius
                , unit.Intel.VisionRadius
                , unit.Intel.VisionRadius - unit.Intel.WaterVisionRadius
                , (unit.Intel.SonarRadius or 0)
                , (unit.Intel.SonarRadius or 0) - unit.Intel.WaterVisionRadius
            ))
        end
    end
```

JaggedAppliance:
How about trying to reduce the amount of visible rings for navy in general? Select a frigate and you can have 6 or 7 rings, many of them very similar range. Sonar or vision could be set to the water vision radius to at least give some indication of its range. 

I don't think it's important to remove sonar just because it's redundant. I also don't know anything about how this could affect mods, an example would help in this case.

Basilisk3:
I wrote a short summary why this is still a good idea in my opinion:

- The UI should not convey useless information to the player.
- Units like the Salem have 6 range rings, which is quite a lot. 5 is still not great but more manageable.
- Removing these range rings is slightly beneficial to performance.
- Units having random stats that were placed there with little consideration is bad game design.
- Mods are usually not very balanced and it's unlikely they were crafted to consider these obscure values.
- Pretty much every regular balance change will impact mods more than this PR.
- This is not necessarily a bad balance change for mods. Shifting sonars onto fewer but more effective units actually creates more interesting counterplay, since not every random unit ends up with one.

**In summary:**
In a modded game that introduces these kinds of units, cloaked but unstealthed underwater units could now slip past, e.g., a Summit within a range of 20 (shorter than the range of a Tech 1 tank) if there are no sonars, intel ships (cruisers), or scouts with sonars. I think this is a very rare edge case and assumes the player has a unit mix of exclusively 1 to 2 units.

**Final thoughts:**
We do not have to remove these stats from all units changed here. We could also only remove it for _some_ units and make it useful for others (perhaps those with sonar ranges above 20?) by increasing it. I think this might be a fair compromise.

Hdt80bro:
Well, it sounds like people have strong opinions about this - I like the idea of a compromise. That might actually be even more interesting. What units did you have in mind?

Basilisk3:
I would definitely remove it for battleships and destroyers. Cruisers should be the primary intel unit at the Tech 2 stage, and battleships simply don't need sonars. Carriers could have their sonar radius increased because it fits their strong intel capabilities well. However, since almost everyone will have a Tech 3 sonar by the time carriers appear, the actual gameplay impact of this change would be negligible.

BlackYps:
Closing due to inactivity.

PR #6724, opened 2025-04-09
Refactor unit cap watch thread for default AI

### LINKED ISSUE #6722: [Feature]: Refactor UnitCapWatchThread for default AI
### Describe the Feature

The UniCapWatchThread is a piece of logic that runs every 30 seconds. It checks the unit cap and kills units if it is close to its cap.

The AI also has some builder conditions that will return false if the AI is within a certain ration of its cap.

The issue : The current logic uses this to determine when to action its logic.

if GetArmyUnitCostTotal(aiBrain:GetArmyIndex()) > (GetArmyUnitCap(aiBrain:GetArmyIndex()) - 10) then

This is not inline with its builder condition ratios so the AI will stop producing units long before this triggers.

The more important part is how it handles culling units. 
It will get a list of T1 mobile land units and kill up to 20 of them.
It will get a list of T1 direct fire defense units and kill up to 20 of them.

I'll be refactoring this to make it take into account other unit types and perform more validation of the units that it culls so that on 20km+ maps the AI does not constantly run into unit count issues.

### Expected Behavior

AI can cull lower tier units when it hits unit counts that will be inline with what it currently has and covers more unit types such as Air, Naval etc.

### Screenshots

_No response_

### Additional context

I'm going through the process on rewriting how this functions for my own AI so will port back improvements that I believe will help the default AI.

relent0r:
This pull request refactors the unitcap watch thread that runs against the AI every 30 seconds.
Original functionality.
Loop every 30 seconds and check if we are within 10 of the unit cap. If we are will kill up to 20 T1 units, on the next loop we would kill all the DIRECTFIRE point defense. Then on the next loop we would reset. This logic did not do a very good job of keeping the AI under unit cap.

fixes #6722

Refactored functionality.
Loop every 30 seconds through a data table of possible unit categories that can be culled. 
The table supports options such as 
- Perform compares before initiating the cull, for example if you want to make sure you have higher tier units before culling lower tier ones you can.
- Check if the unit is attached to a transport or  air staging facility to avoid errors
- Check if the unit is idle in case you do not want to cull units that are not idle (such as engineers that are building something)

The table can be modified and added to so that we can support better lists of units should it be desired. The table is designed to support certain mods that have units such as naval or air engineers.

The loop logic supports various configuration variables to tune its behavior. The goal is that the loop will not cull ALL of the units of the category but will dynamically decide based on how close to the unit cap we are. This will allow the loop to cull various unit categories per pass to achieve a better balance.

With the default unit cap of 1000 the AI usually does not hit the unit cap until it is on a 20km+ map size. Though there are scenarios such as water maps where the AI will build land units and not be able to transport them to the enemy.

Have run various scenarios with various unit caps to validate that the culling logic is working correctly.

## How to test
With the default unit cap of 1000 testing can bit a little painful so I recommend using lower unit caps for testing. The idea is that if the AI can better manage its unit cap when it is lower then it will have no issues at higher unit caps. The main issues I saw when testing lower unit caps was around the issues of hitting unit cap before the AI had shifted to Tier 2. 

## Additional context
The logic is designed to work in tandem with the build condition UnitCapCheckLess. Ideally they will be tuned to work in harmony but this pull request does not retune the build condition in order to keep the code review simple. The UnitCapCheckLess will return false once the unit cap is beyond a certain amount. With the current unitCapDesiredRatio variable being set to 0.9 (90% of unit cap) the logic will not activate until the build condition is already returning false in 'most' cases.

## Checklist


Hdt80bro:
Are you sure it checks every 3 seconds? It only looks like 30 to me.

relent0r:
> Are you sure it checks every 3 seconds? It only looks like 30 to me.

Your totally right, my brain was translating it to WaitTicks despite it being quite clear :D

lua/AI/aiarchetype-managerloader.lua
```diff
@@ -84,45 +84,121 @@ function SetupMainBase(aiBrain)
...
+    -- The cull table describes the types of units we want to cull if we get close to our unit cap
+    -- Options :
+    -- categories - The category of the units we want to cull
+    -- compare - If we want to validate if another type of unit already exist before we cull them
+    -- compareTo - The categories we want to compare against
+    -- checkAttached - If we want to make sure they are not attached to a transport or air staging before culling them
+    -- Remember that this table will run in order, so we want the most deisred to cull first
+    -- There is also two configurable settings
+    -- cullPressure - Indicates how many units we want to cull per pass. The closer we are to the unit cap the more units we will cull.
+    -- dynamicRatioThreshold - What sort of ratio we want when performing compares. So that we dont instantly cull lots of units just because
```
Hdt80bro: These could be an annotation type

lua/AI/aiarchetype-managerloader.lua
```diff
@@ -84,45 +84,121 @@ function SetupMainBase(aiBrain)
...
+    -- compare - If we want to validate if another type of unit already exist before we cull them
+    -- compareTo - The categories we want to compare against
+    -- checkAttached - If we want to make sure they are not attached to a transport or air staging before culling them
+    -- Remember that this table will run in order, so we want the most deisred to cull first
+    -- There is also two configurable settings
+    -- cullPressure - Indicates how many units we want to cull per pass. The closer we are to the unit cap the more units we will cull.
+    -- dynamicRatioThreshold - What sort of ratio we want when performing compares. So that we dont instantly cull lots of units just because 
+    -- one of the next tier is available
+
+    local cullTable = {
```
Hdt80bro: This table should be outside of the function so it can be hooked

lua/AI/aiarchetype-managerloader.lua
```diff
@@ -84,45 +84,121 @@ function SetupMainBase(aiBrain)
...
+        },
+        T1LandUnits = {
+            categories = categories.MOBILE * categories.TECH1 * categories.LAND - categories.ENGINEER,
+            compare = false,
+            cullRatio = 0.2,
+            checkAttached = true
+        },
+        T1LandEngineer = {
+            categories = categories.MOBILE * categories.TECH1 * categories.LAND * categories.ENGINEER - categories.COMMAND,
+            compare = true,
```
Hdt80bro: The `compare` field seems unnecessary - if there's a `compareTo` field, it should be checked

lua/AI/aiarchetype-managerloader.lua
```diff
@@ -84,45 +84,121 @@ function SetupMainBase(aiBrain)
...
+            compareTo = categories.MOBILE * categories.LAND * categories.ENGINEER * (categories.TECH2 + categories.TECH3)- categories.COMMAND - categories.SUBCOMMANDER - categories.POD - categories.FIELDENGINEER,
+            cullRatio = 0.2,
+            checkAttached = true,
```
Hdt80bro: If `checkAttached` is always true, should it be default behavior? When would we not want to check for this?

lua/AI/aiarchetype-managerloader.lua
```diff
@@ -84,45 +84,121 @@ function SetupMainBase(aiBrain)
...
+        },
+    }
+    local unitCapDesiredRatio = 0.9
+    local maxCullNumber = 30
```
Hdt80bro: These would be good to leave outside the function as well

lua/AI/aiarchetype-managerloader.lua
```diff
@@ -84,45 +84,121 @@ function SetupMainBase(aiBrain)
...
-                    v:Kill()
-                    count = count + 1
-                    if count >= 20 then break end
+        local brainIndex = aiBrain:GetArmyIndex()
+        local currentCount = GetArmyUnitCostTotal(brainIndex)
+        local cap = GetArmyUnitCap(brainIndex)
+        local capRatio = currentCount / cap
+        if capRatio > unitCapDesiredRatio then
+            local cullPressure = math.min((capRatio - 0.80) / 0.2, 1)
+            local dynamicRatioThreshold = 2.0 - (capRatio - 0.80) * 9
```
Hdt80bro: Where do the constants in this math come from? Should they be named somewhere?

lua/AI/aiarchetype-managerloader.lua
```diff
@@ -84,45 +84,121 @@ function SetupMainBase(aiBrain)
...
+            local culledUnitCount = 0
+            for k, cullType in cullTable do
+                if cullType.compare then
+                    local compareFrom = aiBrain:GetCurrentUnits(cullType.categories)
+                    local compareTo = aiBrain:GetCurrentUnits(cullType.compareTo)
+                    if compareTo > 0 and compareFrom > 0 then
+                        local ratio = compareFrom / compareTo
+                        if ratio > dynamicRatioThreshold then
+                            local toCull = math.min(compareTo, math.ceil(compareTo * ratio * cullType.cullRatio * cullPressure))
```
Hdt80bro: Isn't `compareTo * ratio` just `compareFrom`?

Hdt80bro (changes requested):

relent0r:
@Hdt80bro thanks for the feedback, I'll go over them tonight. Some good valid points.

One question I did have was about externalizing the tables / variables.  What benefits do we gain from that? I've always been cautious about that sort of thing as its easy to 'lose' them at some point.

Hdt80bro:
Exposing parameters outside of functions means they aren't hidden from modders and can be easily controlled. If a mod wants to change something inside a function, they have to replace the entire thing with a copy and make their one change. This has caused problems in the past because now the mod will no longer work correctly when we update anything in that function.

There are usually also performance/memory concerns if you have to reprocess things every function call (particularly with recreating the same table structure), but that's not an issue here since the thread is only created once.

I'm not sure what you mean by things getting lost, but if you meant that it's hard to see that the function uses the variable as an upvalue, I would just put the function controls immediately above the function.

relent0r:
> Exposing parameters outside of functions means they aren't hidden from modders and can be easily controlled. If a mod wants to change something inside a function, they have to replace the entire thing with a copy and make their one change. This has caused problems in the past because now the mod will no longer work correctly when we update anything in that function.
> 
> There are usually also performance/memory concerns if you have to reprocess things every function call (particularly with recreating the same table structure), but that's not an issue here since the thread is only created once.
> 
> I'm not sure what you mean by things getting lost, but if you meant that it's hard to see that the function uses the variable as an upvalue, I would just put the function controls immediately above the function.

Ah ok I never thought about the hooking part, makes sense thanks. I'll put some time into it this week and make some changes.

lua/AI/aiarchetype-managerloader.lua
relent0r: Have updated to be annotations on the table.

lua/AI/aiarchetype-managerloader.lua
relent0r: Have externalised the table.

lua/AI/aiarchetype-managerloader.lua
relent0r: Your right, I've removed it and added a compareTo check instead.

lua/AI/aiarchetype-managerloader.lua
relent0r: This one, I wouldn't checkAttached if I'm working with structures or units that can't be attached. The original didn't perform this check and it wasn't until I started seeing errors when units got culled when they were on transports (or air staging) that I thought to add the check.

lua/AI/aiarchetype-managerloader.lua
relent0r: I've changed these to be parameters of the function itself since. Not sure if that offers better configuration though.

lua/AI/aiarchetype-managerloader.lua
relent0r: I did document these in the notes of the function. I could put more details in the description if its feels too abstract to understand. These are designed to increase or limit aggressiveness of culling depending on how close you are to the limit and also to make sure the AI isn't overly aggressive at culling categories just because they are the first in the loop.

relent0r:
PR is too old and the commit history is too messy now. Will create new at a later date.

PR #6725, opened 2025-04-09
Tactical Paint

### LINKED ISSUE #5976: Project - Painting feature
# Painting 

## Description

At the moment we have a painting feature for observers. By painting we only mean the ability to draw lines. The feature allows an observer to paint on the map and other observers can view the painting. 

image

This feature can be quite powerful and we wish to extend it. In particular we want regular players to be able to paint, but we also want to improve the painting itself. 

## Tasks

- [ ] Interpolation I
- [ ] Interpolation II
- [ ] Update frequency
- [ ] Message in chat

4z0t:
This feature adds ability to draw lines in the worldview which teammates can see. (I'm bad and describing things xd)
It is possible thanks to [this binary patch](https://github.com/FAForever/FA-Binary-Patches/pull/110).
tp

- Adds canvas mode where you can draw lines with color of player's army. This mode is enabled via keybind "Toggle Tactical Paint canvas" under "'Tactical Paint" section.
- You can also access canvas via hold version of keybind. It is 'Hold to toggle Tactical Paint canvas'.
- When in paint mode you can draw lines holding left mouse button and erase them holding right mouse button.
- These drawings are shared between teammates.
- You can mute specific player if one annoys you.
- Can be turned off completely (it is on by default).
- These drawings are left in replays, so mind what you draw.
- You can adjust the duration of **your** drawings up to a maximum of 1 minute.
- Observers see all drawings and can draw their own.

* With setup of 4 players 2 for team.
* With replay of previous game.

## Checklist
- [x] Add replay sync
- [x] Add mute ability
- [x] Allow hiding all drawings
- [x] Optimize network usage
- [x] Make lines decay


Garanas (commented):
Overall - looks exciting!

Overall a few code style suggestions:

- Classes should reside in their own files. If you do not do this they can become impossible to hook properly (hello collection of unit, weapon and projectile classes files). 
- There's some hooks directly into the worldview class in the `Main` function. Probably best to just add the logic to the worldview class itself. 

And it appears (based on your video) that besides painting, you can also remove lines. I understand the idea, but does it not make it needlessly complex from all directions? More complex to implement. More complex to sync. More complex to maintain. More complex to moderate in case people make weird drawings...

What if instead we just only provide the ability to draw lines and then they fade out relatively quickly?

4z0t:
> And it appears (based on your video) that besides painting, you can also remove lines. I understand the idea, but does it not make it needlessly complex from all directions? More complex to implement. More complex to sync. More complex to maintain. More complex to moderate in case people make weird drawings...

> What if instead we just only provide the ability to draw lines and then they fade out relatively quickly?

I think it may cause some unwanted cases when you draw something important and It disappears. I understand that logic may become overcompilcated, but it is worth to do for such feature.

Garanas:
> I think it may cause some unwanted cases when you draw something important and It disappears. I understand that logic may become overcompilcated, but it is worth to do for such feature.

I don't agree. A lot of features in FAF are too complicated for what they need to be. And it reduces the ability to maintain it and the ability make improvements based on feedback.

As an example, this is all that you can do in Dota 2:

- https://www.youtube.com/watch?v=dCA8xlS2shw

Of course, in Dota it only lingers for seconds. We could keep it for 10-20 seconds.

lua/ui/game/TacticalPaint/Main.lua
```diff
@@ -1,124 +1,20 @@
...
-    for id, player in GetArmiesTable().armiesTable do
-        if player.nickname == nickname then
-            return IsAlly(focus, id)
-        end
-    end
-end
-
-local myColor = getArmyColor()
-
-function FindClients(id)
```
Garanas: You still need this logic in some place to make it work between observers.

4z0t: Doing it right now.

lua/ui/game/TacticalPaint/Main.lua
```diff
@@ -0,0 +1,203 @@
...
+    for _, view in views do
+        ---@type Canvas?
+        local canvas = view._canvas
+        if canvas then
+            canvas:SetActive(true)
+        end
+    end
+
+    ForkThread(UnlockThread)
+end
```
lL1l1: There is a bug here where the thread happens before the key action event at high FPS (248-249 for me), and causes tactical paint to disable itself and quickly re-enable itself, which causes the mouse to lose focus. 

Solution: `WaitFrames` should be `WaitSeconds` instead. 

5 frames at 60 fps is 0.083 seconds, and at 250 fps is 0.02 seconds. I tested how often my keyboard input repeats the `OnHold()` function, and got 0.03-0.05 seconds between each. I also had fps lag when drawing due to some other bug, but it didn't seem to affect my input interval, so I would recommend `WaitSeconds(0.1)`.

Ideally we'd actually have a key pressed/key released event for this to get perfect timings. Checking `IsKeyDown` in the render update might also be a possibility. Either way requires accounting for the full keybind with modifiers, would be nice to have a utility function for that.

lua/ui/game/TacticalPaint/Main.lua
```diff
@@ -0,0 +1,203 @@
...
+            canvas:SetActive(not canvas:GetActive())
+        end
+    end
+
+    -- if canvas:GetActive() then
+    --     local EscapeHandler = import('/lua/ui/dialogs/eschandler.lua')
+    --     EscapeHandler.PushEscapeHandler(function()
+    --         canvas:SetActive(false)
+    --     end)
+    -- end
```
lL1l1: Why is this commented out?

lua/keymap/keydescriptions.lua
```diff
@@ -218,6 +218,10 @@ keyDescriptions = {
     ['toggle_map_control'] = "<LOC key_desc_0109>Toggles the map control window",
     ['cam_free'] = "<LOC key_desc_0104>Toggles camera free mode",
 
+    ["tpaint_open"] = 'Open Tactical Paint canvas',
+    ["tpaint_hold"] = 'Hold Draw',
```
lL1l1:
```suggestion
    ["tpaint_open"] = 'Toggle Tactical Paint canvas',
    ["tpaint_hold"] = 'Hold to toggle Tactical Paint canvas',
```

lua/options/options.lua
```diff
@@ -1351,28 +1351,35 @@ options = {
...
             },
-
             {
-                title = "<LOC OPTIONS_0309>Painting",
-                key = 'casting_painting',
+                title = "Painting enabled",
+                key = 'tpaint_enabled',
                 type = 'toggle',
-                default = 18,
+                default = 'off',
```
lL1l1: It should default to `on`. When I first used the hotkeys to activate tactical paint I was confused why it wasn't drawing because the setting was off but the canvas opened anyway. There should be some warning if painting is disabled.

lua/options/options.lua
```diff
@@ -1351,28 +1351,35 @@ options = {
...
-
                 -- these are expected everywhere
                 default = '',
                 key = '',
             },
-
             {
-                title = "<LOC OPTIONS_0309>Painting",
-                key = 'casting_painting',
+                title = "Painting enabled",
```
lL1l1, line 1361: This name is inconsistent with "Tactical Paint". Perhaps it should just be called "Painting" since it is not a mod anymore.

lL1l1 (commented):
After drawing a few lines I lose a lot of FPS but only while drawing (left click held down + canvas active). Not sure why.

lua/ui/game/TacticalPaint/Main.lua
4z0t: This thing is rather experimental, because it is very inconsistent. I want to find the other way around to detect when a key is unpressed and then disable the canvas. And there is a solution actually, it is how @Garanas ~~does right now~~ did initially, but we have to make it detect whether all modifiers are also active.

lua/ui/game/TacticalPaint/Main.lua
4z0t: I wanted to detect when I press ESC to close canvas, but it doesn't work for me.

4z0t:
> After drawing a few lines I lose a lot of FPS but only while drawing (left click held down + canvas active). Not sure why.

~~I found an issue... Turns out high DPI mouse makes more calls in `HandleEvent` before frame is even rendered. Which causes it to call checks more than your framerate per second (like 10-50 per frame)~~
Maybe not...

4z0t:
@lL1l1 Thanks for pointing out at the problem. Turned out I messed up with removing algorithm :D

4z0t:
[This](https://github.com/FAForever/FA-Binary-Patches/pull/112) is also related.

lua/options/options.lua
4z0t: Some people may complain that they don't want to have things by default, and I understand them. I'd let them choose whether to use it or not.

Garanas: I agree with @lL1l1 - this should be turned on by default. Those that are vocal enough to complain also either have, or have the ability to get the knowledge to turn it off. 

If you don't enable it by default then in my own experience a large chunk of the player base will simply not use it.

lL1l1: This is a communication tool like pings or chat, and most people like to use/see those features. If they want to disable it entirely or mute a specific player it's up to them just like it is with chat (idk if pings can be muted).

lua/ui/game/TacticalPaint/Main.lua
4z0t: This solution now should be better.

4z0t: @lL1l1 check this one now. It must be much better.

changelog/snippets/features.6725.md
```diff
@@ -0,0 +1,13 @@
+- (#6725) Tactical Paint feature:
+
+    This feature adds ability to draw lines during game which teammates and observers can see.
+    It is possible thanks to [this binary patch](https://github.com/FAForever/FA-Binary-Patches/pull/110).
+
+    - Adds canvas mode where you can draw lines with color of player's army. This mode is enabled via keybind "Toggle Tactical Paint canvas" under "'Tactical Paint" section.
+    - When in paint mode you can draw lines holding left mouse button and erase them holding right mouse button.
+    - These drawings are shared between teammates.
+    - You can mute specific player if one annoys you.
+    - Can be turned off completely **(it is by default, you have to turn this on to see!)**.
```
lL1l1:
```suggestion
    - Can be turned off completely (it is on by default).
```

changelog/snippets/features.6725.md
```diff
@@ -0,0 +1,13 @@
...
+    - These drawings are left in replays, so, mind what you draw.
```
lL1l1:
```suggestion
    - These drawings are left in replays, so mind what you draw.
```

changelog/snippets/features.6725.md
```diff
@@ -0,0 +1,13 @@
...
+    - You can adjust duration of **your** drawings with maximum of 1 minute.
```
lL1l1: Grammar fix suggestion, but really there should be details in this point or in an additional point on how long your teammates' or your own drawings stay on the map.
```suggestion
    - You can adjust the duration of **your** drawings up to a maximum of 1 minute.
```

changelog/snippets/features.6725.md
```diff
@@ -0,0 +1,13 @@
...
+    - Observers see these between each over and can draw their own.
```
lL1l1:
```suggestion
    - Observers see all drawings and can draw their own.
```

lua/ui/game/TacticalPaint/Main.lua
```diff
@@ -149,43 +141,120 @@ function Clear()
...
         end
     end
 end
 
 function OnHold()
-    lock = true
-    if isEnabled then
+    local keymap = Prefs.GetFromCurrentProfile('UserKeyMap')
+    ---@type string?
+    local key = table.find(keymap, "tpaint_hold")
```
lL1l1: Maybe this can be used instead?
https://github.com/FAForever/fa/blob/5ab1d3ef227e983e5cd6fbd2c21ee08b1daf00cc/lua/keymap/keymapper.lua#L95-L98

lua/ui/game/TacticalPaint/Main.lua
```diff
@@ -149,43 +141,120 @@ function Clear()
...
+    while checkFn() do
+        WaitFrames(1)
+    end
+
+    SetCanvasesActive(false)
+end
+
+---@param keyData NormalizedKeyBinding
+---@return fun():boolean
+local function GetKeyDataPressedChecker(keyData, keyCodeString)
```
lL1l1: I think such a function should be in `uiutils.lua` or `keymapper`

lL1l1 (commented):
The minimum distance is too low so it's hard to draw with precision. For example I find it difficult to draw an arrow head, or to circle a mex spot when zoomed in
I had to take some time to find what the actual minimum distance is calculated from in `Canvas.lua`
https://github.com/FAForever/fa/blob/e38d19fb70b9b37cc85368a9b10da888dd06316b/lua/ui/game/TacticalPaint/Canvas.lua#L279
I would make that `100` value a local variable so it can be adjusted easily just like `minDist`. Same for the `50` value used for the eraser radius.
It would also be nice to have a line connecting to your cursor while you are drawing. The line would be "placed" once it is longer than minDist or when drawing ends.

lua/ui/game/TacticalPaint/Main.lua
lL1l1: Yes it turns off instantly now when the keybind is released and no longer lags. There is an engine issue where if you press a key before the modifier then the modified action doesn't register but that makes sense since a modified key wasn't input in the first place, and usually keybinds are key presses, not key holds.

lua/ui/game/TacticalPaint/Main.lua
lL1l1: It's because `uimain` has the actual escape handler, which is set to nil when a game session starts.
https://github.com/FAForever/fa/blob/6780001ed8a649cd4c9af492c7e892c6e3e17877/lua/ui/uimain.lua#L142
https://github.com/FAForever/fa/blob/b1e44ebd48b5e7dd82d9007ec10f261a5ed5732e/lua/ui/menus/menucommon.lua#L82-L90

4z0t:
I think this is enough for initial stage of testing on faf develop. Further improments/changes can be made after people try it and share their opinion.

4z0t:
Discovered bugs:
* ~~Can't either move or delete markers~~
* Unknown error which makes it impossible to draw after pausing right after game starts
* ...

Garanas:
As discussed with the game team meeting on the 3th of May we're going with https://github.com/FAForever/fa/pull/6726. You do receive full credits in the changelog, my name is not even mentioned there 😄!

PR #6726, opened 2025-04-10
Alternative implementation of tactical paint

### LINKED ISSUE #5976: Project - Painting feature

Garanas:

For comparison, see #6725.

Related issues: https://github.com/FAForever/fa/issues/5976.

Introduces an alternative and more simple approach to painting. Inspired partially by #6725, #4393 and of course the original UI mod that introduced basic painting. Requires the assembly patch https://github.com/FAForever/FA-Binary-Patches/pull/110 made by @4z0t which is available on FAF Develop at the moment of writing.

https://github.com/user-attachments/assets/8d1a2f2d-82d5-416e-9817-f4ce80614f3e

At the moment this implementation supports the following features:

- Support for multiple world views: once shared, the painting is available on all world views.
- Support for replays: once shared, the painting shows in the replay.
- - This enables moderators to moderate on the paintings when necessary.
- Support for multiplayer: once shared, paintings show for all allied players.
- Support for use of team colors (https://github.com/FAForever/fa/pull/6726/commits/822df23f7761e1d95e7bd59facf560eb5008bddd)
- Support of gradual fade out (https://github.com/FAForever/fa/pull/6726/commits/93eab5a69fdbe1c22a2d4a379e84d1190211326e)
- Support to enable or disable the feature as a whole (https://github.com/FAForever/fa/pull/6726/commits/6eed3e45b583d14ccbdadcf4151baa2f51e6a0ad)
- Support to abort a painting (https://github.com/FAForever/fa/pull/6726/commits/788a7fd1f975d60284fc7fa0772598e5bb7569b5 and https://github.com/FAForever/fa/pull/6726/commits/461231675cced7e8acb03f5b586c99fe9c3cf535)
- Support to delete a painting (ALT + Right Mouse Button)
- Support to mute a painter for the current session (CTRL + ALT + Right Mouse Button)
- Drawings have decreased resolution as you zoom out
- Drawings have increased resolution as you make curves/bends

And the following sub-features:

- Active painting is aborted when switching focus army (https://github.com/FAForever/fa/pull/6726/commits/04bebb2f7e8b9297af7cd4ed550e2936f4a11b55)
- Active painting is aborted when turning on or off NIS mode for cinematics (https://github.com/FAForever/fa/pull/6726/commits/e2e7ff7de0ceee2f19e20ab189ec37e54a3e5647)
- Paintings are automatically de-allocated by using weak tables (https://github.com/FAForever/fa/pull/6726/commits/433636d4b4935bb3b7fcf32d061c6ed0f7ef4f12)

Painting works by pressing your right mouse button. You can not paint when you have a unit selection. You can abort the painting by pressing escape. Painting is also aborted when you switch focus armies. And when you enter cinematic mode. You can delete paintings using ALT + right mouse button. You can mute painters using CTRL + ALT + right mouse button.

Remaining open problems are discussed in https://github.com/FAForever/fa/pull/6726#issuecomment-2797132078

## Code style

The feature as a whole is complicated. There's multiple type of interactions. And there's multiple worldviews. And there's a network sync involved. This implementation attempts to apply the single responsibility pattern to make it more maintainable. It also has the factory pattern to abstract away the complicated network sync logic.

See also this list of responsibilities:

- Active brush stroke: responsible for a brush stroke that is 'under construction'.
- Brush stroke: responsible for a single brush stroke by describing its functionality (rendering, decay, collision computation), but not calling it.
- Painting: responsible for managing all brush strokes.
- Painting Brush: responsible for interacting with paintings and creating new paintings.
- Painting Canvas: wrapper for each world view, responsible for gluing together the adapter, brush and  the painting with its brush strokes.
- Painting Canvas Adapter: responsible for all cross-worldview and cross-network related logic.
- - Adapter Factory: responsible for choosing the correct adapter.
- - Observer Adapter: subscribes via sim callbacks, subscribes and publishes via chat messages.
- - Replay Adapter: subscribes via sim callbacks.
- - Player Adapter: subscribes and publishes via sim callbacks.

Almost all logic and state is part of a class. Almost everything is documented. And the painting canvas supports hot reload. This allows you to make changes and when you hit save, the new class is loaded in live. No need to restart the map. 

- Tested in skirmish, with one world view.
- Tested in skirmish, with split world views.
- Tested in multiplayer, with one world view.
- Tested in multiplayer, with split world views.
- Tested in multiplayer, with observers.
- Tested in a replay.

## Additional context

Note that this pull request is not intended to be a direct replacement. That would be a little rude as I did not discuss it with 4z0t. The purpose of this pull request is to show the value of simplicity over complexity. See also my comment here: https://github.com/FAForever/fa/pull/6725#issuecomment-2791609195

I strongly believe this feature is a significant improvement. But I also strongly believe that a simple solution is often the best. Especially with open source projects such as these where maintainers come and go.

## Checklist


Garanas:
## Open user problems

- Where to manage muting/filtering of the paintings of peers?

I was thinking of using the chat configuration for this. That is also where you filter chat messages.

Implemented in https://github.com/FAForever/fa/pull/6726/commits/8833816dfad97ed26c4e9fa900b1c1e8a12ae56a. Peers are muted for the remainder of the session.

- How to delete unwanted paintings?

I was thinking of holding ALT + Right Mouse Button to act as a 'reverse' brush. Paintings can not be destroyed 'partially' though. If you touch a painting to delete it then the entire thing is deleted at once.

Implemented in https://github.com/FAForever/fa/pull/6726/commits/a1033cdec242aa0c4f83c69506864cf163a7450a. The interaction is not shared across the network, paintings are only deleted locally.

## Open technical problems

- Chat messages have a maximum size. We'll need to overcome that somehow.

We can reduce the number of bytes used by changing the format from a 'array of structures' to a 'structure of arrays'. That won't fix the problem for really long drawings, but it will make it less common.

Implemented with: https://github.com/FAForever/fa/pull/6726/commits/9a6e38b2f377cb9a733ea932521df6804b281a21

- What type of format checks should we do to prevent malicious shared paintings from having consequences?

The draw line function has the following interface:

```lua
--- Draws a line in world. Must be called within `WorldView:OnRenderWorld`
---@param position1 Vector
---@param position2 Vector
---@param color Color
---@param thickness? number
function UI_DrawLine(position1, position2, color, thickness)
end
```

Only the positional data is shared across the network. We verify this using the function `tonumber`. All samples that don't meet the requirement are not included. The color is determined by data from the engine. And the thickness is always 0.

- There should probably be a 3rd adapter for replays that never tries to send any data.

Implemented in https://github.com/FAForever/fa/pull/6726/commits/c9f6bbf32ac520c58a2590807159f97c31065b45.

4z0t:
> ... But I also strongly believe that a simple solution is often the best...

I start to wonder how *complex* solution may look like...

Garanas:
> > ... But I also strongly believe that a simple solution is often the best...
> 
> I start to wonder how _complex_ solution may look like...

Thank you for the feedback. At the end of the day it's an opinion in programmer land 😄 !

The feature as a whole is of course complicated. There's multiple type of interactions. And there's multiple worldviews. And there's a network sync involved. This implementation attempts to apply the single responsibility pattern to make it more maintainable. It also has the factory pattern to abstract away the complicated network sync logic.

See also this list of responsibilities:

- Brush stroke class: responsible for a single brush stroke by describing its functionality (rendering, decay, collision computation), but not calling it.
- Painting class: responsible for managing all brush strokes.
- Active painting class: responsible for a painting under construction.
- Painting Canvas: responsible for glueing together the adapter, brush and painting.
- Painting Brush: responsible for interacting with paintings and creating new paintings.
- Painting Canvas Adapter: responsible for all cross-worldview and cross-network related logic.
- - Adapter Factory: responsible for choosing the correct adapter.
- - Observer Adapter: subscribes via sim callbacks, subscribes and publishes via chat messages.
- - Replay Adapter: subscribes via sim callbacks.
- - Player Adapter: subscribes and publishes via sim callbacks.

Almost all logic and state is part of a class. Almost everything is documented. And the painting canvas supports hot reload. This allows you to make changes and when you hit save, the new class is loaded in live. No need to restart the map.

lua/ui/game/painting/PaintingCanvas.lua
```diff
@@ -0,0 +1,327 @@
...
+        Bitmap.Destroy(self)
+        self.Trash:Destroy()
+
+        -- deregister us to render
+        self.WorldView:UnregisterRenderable(tostring(self))
+    end,
+
+    --- Checks if the painting feature is disabled as a whole.
+    IsEnabledByGameOptions = function(self)
+        local gameOption = GetOptions("painting")
```
sam-site, line 103: I'm new to this code base so please disregard my comment if it's not relavent.
How expensive is the `GetOptions()` call?
If `HandleWorldViewEvent` is called every frame then even a relatively low cost look up can add up. Consider caching the option instead and "push" any changes to the "painting" option (which I guess would be infrequent).

lua/ui/game/painting/PaintingBrush.lua
```diff
@@ -0,0 +1,267 @@
...
+    GetColorOfActiveBrushStroke = function(self)
+
+        -- in this case we just want to use the color of the
+        -- current focus army. This may not always match the
+        -- command source, but the color is only visible to
+        -- the person drawing it and not to the rest of the
+        -- peers.
+
+        local armiesTable = GetArmiesTable().armiesTable
+        return armiesTable[GetFocusArmy()].color or 'ffffffff'
```
sam-site, line 114: consider making the default colour a game option? Or configurable via some config file or something along those lines. It would give casters more options to customize their presentation.

sam-site: In replay mode when I select an army I am not able to draw. I don't know if that's intentional but this would seem to imply that you should be able to draw and it will match the colour of the focused army.

lua/ui/game/painting/PaintingBrush.lua
```diff
@@ -0,0 +1,267 @@
...
+        if self.ActiveBrushStroke then
+            self.ActiveBrushStroke:OnRender(delta)
+        end
+
+        -- feature: visualize interactions with the canvas
+        local brushRadius = self:GetBrushRadius()
+        local brushCoordinates = self.WorldCoordinates
+        local brushType = self:GetBrushType()
+        if brushType == 'Create' then
+            UI_DrawCircle(brushCoordinates, 0, self:GetColorOfActiveBrushStroke(), 0)
```
sam-site, line 234: size = 0, thickness = 0, so does this do anything?
In game and in replay I saw no circles.
(replay https://replay.faforever.com/24714660)

lua/ui/game/painting/BrushStroke.lua
```diff
@@ -0,0 +1,457 @@
...
+    --- Renders the brush stroke.
+    ---@param self UIBrushStroke
+    ---@param delta number
+    OnRender = function(self, delta)
+        local decayProgress = 0
+        if self.Decay then
+            decayProgress = math.clamp((GetGameTimeSeconds() - self.Decay.StartTime) / self.Decay.Duration, 0, 1)
+        end
+
+        local decayedColor = self:ComputeDecayedColor(self.Color, decayProgress)
```
sam-site, line 82: Do you need to do this if it's not in Decay? They decayProcess will be 0, and so the line should be 100% opaque. 
You could save a couple of math function calls, parsing the color and setting the color.

lua/ui/game/painting/PaintingCanvas.lua
Garanas: As far as I am aware it is not much more expensive then the average engine call is. In this particular example it is also allocation free. A lot happens each frame, and a lot of that is not allocation free. We should be good here.

lua/ui/game/painting/BrushStroke.lua
sam-site:
```suggestion
local renderColor = self.Color
if self.Decay then
   local decayProgress = 0
   decayProgress = math.clamp((GetGameTimeSeconds() - self.Decay.StartTime) / self.Decay.Duration, 0, 1)
   renderColor = self:ComputeDecayedColor(self.Color, decayProgress)
end

...

            UI_DrawLine(position1, position2, renderColor, 0)
```

lua/ui/game/painting/PaintingBrush.lua
Garanas: I'll check and see what's going wrong! You're right that it would imply that your drawings suddenly have a different color, that would be confusing.

lua/ui/game/painting/PaintingBrush.lua
Garanas: There's a tiny circle at the position of your mouse. It's of the same color that you're drawing with. Ideally we also change the cursor to a pencil 😃 , but at the moment I can understand it appears meaningless.

lua/ui/game/painting/BrushStroke.lua
```diff
@@ -0,0 +1,457 @@
...
+    ---@param y0 number
+    ---@param z0 number
+    ---@param x1 number
+    ---@param y1 number
+    ---@param z1 number
+    ---@param x2 number
+    ---@param y2 number
+    ---@param z2 number
+    ---@return number?
+    CurvatureOfSegments = function(self, x0, y0, z0, x1, y1, z1, x2, y2, z2)
```
sam-site: minor: `self` parameter does not appear to be used.

lua/ui/game/painting/ActiveBrushStroke.lua
```diff
@@ -0,0 +1,152 @@
...
+    ---@return boolean
+    DebounceSample = function(self, sample)
+        -- feature: debounce samples that are in too quick succession
+        local now = GetGameTimeSeconds()
+        if now - self.LastEdited < self.DebounceTimeThreshold then
+            return true
+        end
+
+        -- feature: debounce samples that are too close to each other
+        local debounceDistance = self:GetDebounceDistance(sample)
```
sam-site: minor: for clarity rename `debounceDistnace` to `minimumDistance`

lua/ui/game/painting/ActiveBrushStroke.lua
```diff
@@ -0,0 +1,152 @@
...
+            -- quickly get us two samples to help determine accurate debounce distance
+            return 1
+        end
+    end,
+
+    --- Responsible for debouncing samples. This is to reduce the
+    --- bandwidth it requires to share the brush stroke once it is finished.
+    ---@param self UIActiveBrushStroke
+    ---@return boolean
+    DebounceSample = function(self, sample)
```
sam-site, line 140: Suggestion: Allow higher resolution line drawing if the user is drawing slowly.
i.e. reduce the minimum distance if the mouse position delta is small.

At the moment it feels a bit clunky that I can't draw a complete circle even when moving my mouse slowly and carefully and I'm assuming it's because the position of my mouse cursor is within the mininum "debounce" distance from the previous sample point.

I can think of two ways to do this:

1. have a maximum `DebounceTimeThreshold`, something like 100ms. `if self.LastEdited > self.MaxDebounceTimeThreshold then return false end` (you could still set an absolute min distance). This would probably still feel pretty clunky.
2. measure the accelation of the mouse movement. If the accelation is low then allow the sample. Again you can still allow for some (smaller) absolute minimum distance.

lua/ui/game/painting/BrushStroke.lua
Garanas: You're correct. I wanted to give the mathematics a name. At the moment we don't have a place for these type of calculations. So I kept it inside the class for now, just to keep the code together.

We can refactor that in the future again, unless maybe @Hdt80bro has a good idea where to put this type of mathematics in the repository?

lua/ui/game/painting/ActiveBrushStroke.lua
Garanas: I've implemented something similar in https://github.com/FAForever/fa/pull/6726/commits/c8ee655d17faf5105c748a94c32bea474a0e3065. It decreases the debounce distance as you make a curve.

lua/ui/game/painting/BrushStroke.lua
Hdt80bro: `/lua/system/utils.lua` makes the most sense right now if you're looking for a spot to move it, but I'm not convinced it needs to be.

lua/ui/game/painting/PaintingCanvas.lua
```diff
@@ -0,0 +1,332 @@
...
+        Bitmap.__init(self, worldview)
+
+        self.Trash = TrashBag()
+        self.Painting = CreatePainting()
+        self.Adapter = self.Trash:Add(GetPaintingCanvasAdapter(self))
+
+        self.WorldView = worldview
+
+        -- register us to render
+        self.WorldView:RegisterRenderable(self, tostring(self))
```
lL1l1: Update to match refactor
```suggestion
        self.WorldView:AddShape(self, tostring(self))
```

lua/ui/game/painting/PaintingCanvas.lua
```diff
@@ -0,0 +1,332 @@
...
+            :End()
+    end,
+
+    ---@param self UIPaintingCanvas
+    OnDestroy = function(self)
+        Bitmap.OnDestroy(self)
+        self.Trash:Destroy()
+
+        -- deregister us to render
+        self.WorldView:UnregisterRenderable(tostring(self))
```
lL1l1:
```suggestion
        self.WorldView:RemoveShape(tostring(self))
```

lua/ui/game/painting/PaintingCanvas.lua
```diff
@@ -0,0 +1,332 @@
...
+        -- feature: inhibit painting when holding shift
+        if IsKeyDown('Shift') then
+            self:AddInhibition('shift')
+        else
+            self:LiftInhibition('shift')
+        end
+
+        -- limitation: only enable painting when no unit is selected. This limitation enables us 
+        -- to use a wide range of (mouse) buttons that would otherwise be used by engine functions
+        if table.getn(GetSelectedUnits()) == 0 then
```
lL1l1: - GetSelectedUnits() returns `nil` with an empty selection (this needs to be documented)
- `table.getn` errors on nil, unlike `table.getsize` which returns 0
  - the `table.getn` comparison might as well be removed since the empty selection returns `nil`, unless you know of a way to get an empty table from `GetSelectedUnits()`
```suggestion
        local selection = GetSelectedUnits()
        if not selection or table.getn(GetSelectedUnits()) == 0 then
```

lua/ui/game/renderable/line.lua
lL1l1, line 1: This file conflicts with the new refactor that has `lua\ui\game\shapes\LineShape.lua`

lua/options/options.lua
```diff
@@ -785,6 +785,54 @@ options = {
...
+                custom = {
+                    min = 10,
+                    max = 40,
+                    inc = 1,
+                },
+            },
+
+
+            {
+                title = "<LOC options_painting_duration_title>Duration of when observing",
```
lL1l1:
```suggestion
                title = "<LOC options_painting_duration_title>Duration of paintings when observing",
```

lua/ui/controls/worldview.lua
```diff
@@ -1023,6 +1037,9 @@ WorldView = ClassUI(moho.UIWorldView, Control, WorldViewShapeComponent) {
     ---@param self WorldView
     ---@param pingData table
     DisplayPing = function(self, pingData)
+
+        reprsl(pingData)
+
```
lL1l1: Remove before merging.
```suggestion
```

lua/ui/controls/worldview.lua
```diff
@@ -1103,6 +1120,7 @@ WorldView = ClassUI(moho.UIWorldView, Control, WorldViewShapeComponent) {
                 LayoutHelpers.AtCenterIn(PingGroup.Marker.TeamColor, PingGroup.Marker)
 
                 PingGroup.Marker.HandleEvent = function(marker, event)
+                    LOG("HANDLE EVENT OF MARKER")
```
lL1l1: Remove before merging
```suggestion
```

lL1l1 (changes requested):
- This needs to be changed to calling `OnRender`. `Render` is not used in the other refactored shapes.
https://github.com/FAForever/fa/blob/15346a0c92ff1e07dfe0a38f5d07ac5ad808c527/lua/ui/controls/components/WorldViewShapeComponent.lua#L90

- There are mod incompatibility bugs with mods that make automatic selections.
  1. The automatic selection stops the current stroke and clears it.
  2. The automatic selection disables drawing until a normal selection is made and cleared.
  Example command to reproduce these modded bugs:
```
  local f = import('/lua/ui/game/selection.lua').Hidden
f(function ()
    UISelectionByCategory('ALLUNITS', false, false, false, false)
    LOG('selected this many units: ', table.getsize(GetSelectedUnits()))
end)
```
The other way to make hidden selections is gamemain's `SetIgnoreMode`:
https://github.com/FAForever/fa/blob/310684d764638b4566858110980c9145299bb239/lua/ui/game/gamemain.lua#L40-L43
- I turned off all my UI mods, and if I draw enough with a single line it stops drawing. If there is a max line length or something like that i needs an indicator (like a pencil icon with a filled bar).
- How do I know the controls for drawing and erasing? They should be in the option tooltip at minimum.
- 4z0t's implementation does have a very nice mute menu. Though imo it should be integrated into the chat mute menu (which needs to be made more visible too).
- When erasing, if you alt + right click it shows the eraser circle but does not perform the erasing action until you move 1 pixel.
- It would be good to have a preview of the current line segment being drawn from the end point of the last "saved" line to the cursor.
- The brush stroke color does not update with custom team color changes. There is a score board color function: https://github.com/FAForever/fa/blob/5667965b205f1b6f30994de3163d37cdb2580e01/lua/ui/game/multifunction.lua#L1212

lua/ui/game/renderable/line.lua
Garanas: See also: https://github.com/FAForever/fa/pull/6726/commits/ba0ef811d4d3f120a69911fc14f5ce900f740913

Garanas:
> * This needs to be changed to calling `OnRender`. `Render` is not used in the other refactored shapes.

It is used in the base shape: 

```lua
--- Renders the shape if it is not hidden.
---@param self UIShape
---@param delta number
Render = function(self, delta)
    if self.Hidden then
        return
    end

    self:OnRender(delta)
end,

--- The logic to render the shape.
---@param self UIShape
---@param delta number
OnRender = function(self, delta)
    -- to be defined by the subclass
    error("OnRender must be overridden by subclass")
end,
```

Fixed it in https://github.com/FAForever/fa/pull/6726/commits/b37bfe44e1b40a6af9417b6099b6503631ddb8bf.

Garanas:
> There are mod incompatibility bugs with mods that make automatic selections.
> 
> 1. The automatic selection stops the current stroke and clears it.
> 2. The automatic selection disables drawing until a normal selection is made and cleared.
>    Example command to reproduce these modded bugs:

I've removed the logic in `gamemain.lua` and now just rely on the call to `GetSelectedUnits`. See also https://github.com/FAForever/fa/pull/6726/commits/b48e9a16d7e5463fcd42326b955b49551d039fb3. I think this should fix any incompatibility. Can you give it another try @lL1l1 ?

In general, it is a little sad that working with selections is so sensitive because of legacy. It would be great to have the engine function that was developed by @4z0t to get lists of units without selecting them.

Garanas:
> * I turned off all my UI mods, and if I draw enough with a single line it stops drawing. If there is a max line length or something like that i needs an indicator (like a pencil icon with a filled bar).

This is true - in practice very long brush strokes (a single line) are never quite... useful. By limiting it we encourage users to make smaller brushstrokes and it's also technically nice as it limits the size of the brush strokes that we sent across the interwebs.

In general - when painting (or erasing) it would be great to have a different cursor icon. I've started working on that with ChatGPT (to make the icon), but decided that the pull request is large enough as it is right now.

Garanas:
> * It would be good to have a preview of the current line segment being drawn from the end point of the last "saved" line to the cursor.

Implemented in https://github.com/FAForever/fa/pull/6726/commits/8aa41fa690a9c8299f08daccd630d56313401568

https://github.com/user-attachments/assets/c83f99d2-e780-4513-ada5-c686f0455da1

One thing that I don't like about it is that it introduces a little bit of 'disturbance' when you stop drawing. Perhaps we should always add the last position of the mouse to the painting when we stop?

Garanas:
> * The brush stroke color does not update with custom team color changes. There is a score board color function:

I lack the inspiration at the moment on how to approach this the best way. Do you have a suggestion?

It seems to me that it would be nice to have some module that is responsible for determining the color of an army.

lua/ui/game/painting/PaintingBrush.lua
Garanas: Introduced a small preview: https://github.com/FAForever/fa/pull/6726#issuecomment-2849011032

lua/ui/game/painting/BrushStroke.lua
Garanas: All brush strokes decay, the only brush stroke that does not decay is the active brush stroke. With https://github.com/FAForever/fa/pull/6726/commits/a7602e540647271c06ef583a5b6e92a85b4aef78 I took a different approach. This makes sure that the functions are still being called, but in the case of an active brush stroke it simply returns the same color.

I'm not sure what approach is better for the project as a whole (future maintainability, etc), what do you think?

lua/ui/game/painting/ActiveBrushStroke.lua
Garanas: Applied in https://github.com/FAForever/fa/pull/6726/commits/c5150760fe7f59faeb6d77d390931aac31b7cf8f

lL1l1:
> Perhaps we should always add the last position of the mouse to the painting when we stop?

Yes, it would more accurately reflect the user's intentions.

lL1l1:
> It seems to me that it would be nice to have some module that is responsible for determining the color of an army.

I think the simplest solution is hooking the global `TeamColorMode` and sending a color table to callbacks. The current color string, color mode (boolean), default army color table, and custom army color table (btw this should be filled with '00000000' by default) would be cached.
When the new input is a string different from the old string, the custom army color table would be updated from `StringSplit(mode, '%X')`.
When the new input is a boolean different from the old boolean, callbacks would be triggered with the appropriate army color table.

Simplest way to initialize the color state for others would be to call the callback when it is added. Having this be the only way to inspect the existing colors would be annoying but maintain consistency across UI implementations?

The hook should be in user init and a module would have `AddOnTeamColorsChangedCallback(fn)` and `OnTeamColorsChanged(colorTable)` (only way to not have this second function is if the first is a global).

Garanas:
@lL1l1 I think it would be best if we first merge this pull request before I proceed to make more changes. Otherwise the pull request keeps creeping in size.

Garanas:
I've created some issues of the open changes that may still be interesting to incorporate.

lL1l1:
I think muting players is a must-have feature for this. I created a basic working implementation here: https://github.com/FAForever/fa/pull/6783

Garanas:
There is already a basic implementation available for muting: use of CTRL + ALT + Right Mouse Button.

lL1l1:
I think the issue with my mods were two-fold: 
- Crotalus's common UI tools has its own precursor copy of FAF's `lua\ui\game\selection.lua` with `mods\common\select.lua`, which is used by old mods to make hidden selections, but FAF doesn't detect when that file is making a hidden selection. This is an outdated mod issue. It can be fixed by just replacing that file's functions with FAF's equivalents (that's what I did in my copy and it works).
- The command mode can't be changed in a hidden way as there is no ignore mode in `StartCommandMode`, which is used when a hidden selection restores the old command mode, so you can't rely on it only ever changing for user actions. This is fixable in FAF's `selection.lua` `Hidden` by replacing `CM.StartCommandMode` with `CM.SetIgnoreSelection` (I did this and it works).
https://github.com/FAForever/fa/blob/f22707113c9305abf2a547a7ddb4b8426aea94e0/lua/ui/game/selection.lua#L47-L57

I opened a PR into this branch with the changes here: https://github.com/FAForever/fa/pull/6786

lL1l1:
> There is already a basic implementation available for muting: use of CTRL + ALT + Right Mouse Button.

I only just found out about this while trying out random keys for what key modifiers work with painting lol. The controls for painting should be written in the painting feature tooltip in the options at minimum. Also this keybind is a bit weird since Ctrl + Shift + RMB deletes orders and markers.

PR #6728, opened 2025-04-10
Adjust Othuum hitbox and DPS distribution

lL1l1:
- Make the hitbox shorter so that Othuums in formation bump into each other less often (bumping causes pathfinding calculation and slows them down a lot).
- ~~Make the hitbox slightly wider to fit the model better.~~
- Allocate more DPS to the long range gun
  - 256 short range + 156 long range DPS -> ~~208 short range + 208 long range DPS~~ 228 short range + 184 long range DPS

### Othuum vs Harbinger
Othuums are too underwhelming until a large amount of them is built, particularly against Harbingers which should be a more even matchup since they are of the same unit class. This makes them excessively weak at stopping raids where low unit counts are the norm.

Moving more DPS to the long range gun lets the Othuum compete with Harbs more fairly as Harbs do not get such a massive advantage when kiting Othuums. It is chosen over range because it has less effects in other Othuum vs Harb situations and most directly addresses the situation where the Harb kites the Othuum, which is by far the weakest point.

Keep in mind the Harbinger has a 1k HP shield that recharges in 30s/regens in 33s. This lets it snowball much better than Othuums if it ever wins a fight. It can also regenerate or recharge fully during extended fights due to its low HP.

### Othuum Pathfinding
Othuum's hitbox is adjusted so that medium/large numbers perform better with the average move order, where Othuums are notoriously poor because they break up formation and get into lines due to bumping into each other's front/rear all the time. Good pathfinding is critical to the Othuum since it has short range and low speed for its range at its tech tier, and needs to catch up to the range of Percy/Brick and the speed/range of Harbingers.

1 Othuum vs 1 Harbinger micro tests and 10v10 micro tests (by myself). For the 1v1 tests I also tested range increases for the long range gun without DPS reallocation.

In 1v1 the changes greatly decreases how huge of an advantage a Harbinger has against the Othuum with correct micro.

In 10v10, the Harbinger formation is harder to manage so they don't have as much of an advantage, so they can lose, but they can still win if the formation is managed skillfully by utilizing flanking (Harbs are more maneuverable).

<details> <summary> 1v1 tests </summary>

- (1) Harb kiting an othuum by running away and only being in the long range gun's range, while the Othuum just moves forward:
   - 28 range othuum: Harb 2160/4600 HP remaining
   - 30 range othuum: Harb 1535 HP
   - 208 long DPS: Harb 880 HP remaining
- (2) Harb and othuum walking straight into eachother: 
   - 28 range Othuum: Othuum 860/4700 HP remaining
   - 30 range Othuum: Othuum 1020 HP
   - 208 long DPS: Othuum 1020 HP
- (3) Harb walking into stationary Othuum:
   - 28 range Othuum: Othuum 700 HP
   - 30 range Othuum: Othuum 860 HP
   - 208 long DPS: Othuum 860 HP
- (4) Othuum kiting a Harb by running away (perfect micro like that described in the Percy/Brick 1 range matchup):
   - 28 range Othuum: Harb 285 HP
   - 29 range Othuum: Othuum 2460 HP
   - 30 range Othuum: Othuum 4700 HP
  - 208 long DPS: Othuum 860 HP
- (5) Othuum kiting a harb and then reversing into it once its in range to get in short range
   - 28 range Othuum: Othuum 1980 HP
   - 29 range Othuum: 2940 HP
   - 30 range Othuum: unnecessary
   - 208 long DPS: Othuum 2300 HP

</details>

For pathfinding, a subjective analysis seems to see less bumping.

## Checklist


lL1l1:
With @Rhaelya's suggestion I added 1 range to the Thau cannon and moved some DPS back to the short range gun in return.

With @K-ETFreeman's concern about it, I removed the width change because it does have a small negative effect on gameplay, while the hitbox matching the model is really only something a dev would notice.

Rhaelya (approved):
Everything seems in order

PR #6729, opened 2025-04-11
Fix strategic bomber's 2nd formation

Strogoo:
A more wider 2nd formation for strat bombers was added a long time ago (see: https://github.com/FAForever/fa/pull/2740)
Then it got broken (well, it's basically just my code being bad but w/e)

An old vid displaying changes: https://youtu.be/KDxRhz5J-h4

## Checklist

### DIFF
--- a/lua/formations.lua
+++ b/lua/formations.lua
@@ -813,6 +813,8 @@ function GrowthFormation(formationUnits)
     BlockBuilderLand(subUnitsList, subBlock, SubCategories, 1)
 
     if not table.empty(unitsList.Air.Bomb3) then
+        -- unitsList.Air.Bomb3 contains no more than one table with selected strat bombers in it
+        -- Cycle puts it at index 1 (as it was in the past) otherwise some code below won't work.
         for k,v in unitsList.Air.Bomb3 do
             if k != 1 then 
                 unitsList.Air.Bomb3[1] = unitsList.Air.Bomb3[k]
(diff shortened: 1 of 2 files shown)


lua/formations.lua
```diff
@@ -812,15 +812,21 @@ function GrowthFormation(formationUnits)
     BlockBuilderLand(seaUnitsList, seaBlock, NavalCategories, 1)
     BlockBuilderLand(subUnitsList, subBlock, SubCategories, 1)
 
-    if unitsList.Air.Bomb3[1] then
+    if not table.empty(unitsList.Air.Bomb3) then
+        for k,v in unitsList.Air.Bomb3 do
```
speed2CZ, line 818: Whats the point of this loop?

Strogoo: `unitsList.Air.Bomb3` always contains one single table with selected strat bombers. In past this table index was `1` (and further code relies on this fact), now it's always `4` (probably due to some changes in formation's code). So this cycle just put it again to `1`

Also want to mention that there are 2 formations:
default one (is processed by `AttackFormation()`) - called very often, basically after every move command
2nd (`GrowthFormation()`) <- We are here.

Second one almost never called (no one really use formations and they are mostly the same tbh). So there is no point to dig too deep into it unless we make full refactor (I think I saw a PR for this?)

speed2CZ: Probably worth to left a comment in the code for confused individuals like me?

speed2CZ (approved):

PR #6730, opened 2025-04-12
Balance Idea: move the Seraphim ACU's advanced gun upgrade to the right arm

Basilisk3:
This simple PR moves the Seraphim ACU's advanced gun upgrade to the right arm. The regular (speed) gun upgrade is now the prerequisite for upgrading the advanced (splash) gun.

## Advantages of this approach
- The upgrade is underused due to competing with Tech 3 and Regen Aura, which are more versatile options. Both Tech 3 and Regen Aura are also cheaper, which means that by the time you can upgrade the advanced gun, you face the additional barrier of entry of having to remove two, very useful, upgrades.

## Miscellaneous advantages
- Making it alongside the regular (speed) gun upgrade is basically a necessity anyway.
- The advanced gun upgrade alters the effectiveness of the gun, which is on the right arm, so it makes more sense for it to also be on the right arm.
- Having a unified and clear upgrade path is less confusing for new players who do not instinctively know which upgrade to get.
- Fewer misclicks.
- (Even number of upgrades (4) per slot).

## Checklist

## Summary by CodeRabbit

* **Balance Changes**
  * Moved Seraphim ACU's Refracting Chronotron Amplifier (splash gun) from left arm to right arm.
  * Made Chronotron Accelerator a prerequisite for that upgrade.
  * Adjusted related weapon behavior to match the new placement.

* **Bug Fixes**
  * Removing the upgrade now correctly restores firing rate and effective radius for the Chronotron and its OverCharge variants.

* **Documentation**
  * Added changelog entry describing these changes.


JaggedAppliance:
I think this is ok.

Isocinetic:
I think there should be noticeable visual difference between just having gun and having gun+splash in order to allow intel counterplay

vjt4:
Fully agree. Always such an awkward upgrade to squeeze in.

FAUSheppy:
I have recently seen some discussion about the state of the Seraphim Splash Gun, and I want to provide a perspective from actually using it. I am not a top player, but I am perhaps the only consistent user. I have used it in probably half of my team games over the last two months.

TL;DR:
- It's fun and definitely viable sub-2k in late T2/early T3, however partially because of lack of visual clarity and lack of knowledge in the playerbase.
- It's good to punish over-reliance on OC as a defense in many-ACU team games.
- Sniperbots are too much of a counter for it to be viable at higher ratings long term; there must be a better follow-up/defense against Sniperbots.
- ..for example by buffing Adv Nano to reduce damage from Snipers/possibly splitting Adv Nano into its current form and a cheaper anti-Sniperbot upgrade.

## My Playbook

Normally I get:
- 2 Pgens + 4 E-Storages @60-70 mass income, go gun instead of T3 upgrade.
- ..or 3 Pgens + 4 E-Storages on higher-income maps, go gun instead of T3 upgrade.
- ..rarely, if I have too much mass, go T3 land first and gun after T3 power.

Usually there is also overflow or direct share from my team (which is part of the reason I think this upgrade is better in team games).

This usually means:
- Get gun as usual or on the first T2 PGen, and get Nano after the first T2 PGen.
- Eco to the required mass threshold, start gun with 3-4 engie support, switch factories to AA. One of the main ways to get impact is to tank a strat or gunships from the enemy air player.
- If you are already T3, go mostly shields, Sniperbots, and AA.
- Push enemy com/firebase with some T1/T2 units, go deeper depending on how the map looks, or switch position with a T2 transport to threaten other coms.

## What is it good against?

I think it is currently one of the most effective ways to punish over-ecoing and relying on OC and PD to hold early T3. It gives a significant, immediate power spike (meaning lower buildpower required, ACU usually already close to front), and defending against this timing seems to be non-trivial for a lot of players. At 5400 mass, it is equivalent to a T3 upgrade + 7 light T3 bots, however this would be 200k build time total + walking to front vs only 4.6k build time for the upgrade.

## Why is it bad above a certain skill level

Pushing into a lot of 2k enemies recently, I can begin to see the cracks in this strategy.
- For Aeon & Sera, the most problematic thing to play against are *Sniperbots*. If your enemy spots your upgrade and is willing to possibly give up some space to get 4-5 Snipers out, you are completely done, because the health pool and regen of the commander is too low for that. You need your own Snipers and T3 mobile shields to deal with that, which you will be delayed on because you went gun.
- The obvious counter for Cybran should in some way be to go laser, but in 50+ splash gun games, nobody has done that. Make of this information what you will.
- UEF, imho, is the most cooked. You just have to hang on with Percies and shields, I guess.
- There is no follow-up to double gun. I tried working with Adv Nano a few times, but it doesn't work. It just costs too much for what it does, and it doesn't help against Snipers.

## What would need to change?

- Sera gun needs its own sound effect and visual indicator. Laser is red and loud and scary, and splash gun barely has its own impact animation. People are constantly confused by it up until the moment they die. Make it more peow-peow and blue or something. I feel like this would help a lot.
- Snipers are too strong against this upgrade and the investment you put into it. If everybody would just give up 3 mexes, use their extra mass to go 5 Snipers & shield, it would fail every time. I don't know why Snipers are allowed to have ACU focus, other than it being "logical" because they are Snipers. Removing this feature would certainly help.
- There needs to be something that can punish Sniperbot spam if it is done to counter splash. Things in my mind are: give the double Nano reduction to long-range fire. Currently second Nano gives you 150 regen, that's literally the DPS of one single Sniper. That's just giga useless.
- There are pretty good upgrades blocked by the splash gun. Upon reflecting about it after Rhealy first suggested it, moving the upgrade to middle, or even as an upgrade to first gun, would open T2 (-1000m compared to Nano) as an alternative to Nano Resto Field as the meta play, or TML as a pathway to get the required advantage to pull this off. However, moving it to left arm would also buff Sera Tele quite a bit.

I had a lot of fun using this upgrade. I think it currently suffers from a bad reputation in low rating and bad follow-up and too-direct counters in high rating. I would look forward to having more visual/audio clarity to stop people from getting tilted by it, and buffs which help with its direct counters to make it more viable at high rating.

PR #6735, opened 2025-04-17
AI - add filter for non engineering units in Engineer Manager callback

relent0r:
This PR adds a filter to the UnitConstructionFinished function of the AI engineer manager. It was found that any units in the ConsumptionUnits table were being put into the engineer assignment function loop. This results in many units checking for engineer builders that will not get any due to not being engineers, but will consume compute resources due to this.

It also separates engineers from engineering stations in the consumption table of the engineer manager, this is done so that functions that return engineer counts against engineer managers are not incorrectly counting engineering stations as engineers.

Added logs and validated that invalid units were being assigned engineering tasks.
Added logs and validated that changes to code stopped this from happening.
Detected HIVES were being counted as normal engineers for builder conditions and reran replays to confirm this was no longer happening post changes.

## Additional context
This has been a silent failure in the default AI and was present in the original SCFA code base. Since the units did not pick up engineering jobs and silently failed no one would know it was happening.

## Checklist

### DIFF
--- a/lua/sim/EngineerManager.lua
+++ b/lua/sim/EngineerManager.lua
@@ -666,15 +666,12 @@ EngineerManager = Class(BuilderManager) {
     ---@param unit Unit
     ---@param finishedUnit Unit
     UnitConstructionFinished = function(self, unit, finishedUnit)
-        local dontAssignEngineerTask = true
         if EntityCategoryContains(categories.FACTORY * categories.STRUCTURE, finishedUnit) and finishedUnit:GetAIBrain():GetArmyIndex() == self.Brain:GetArmyIndex() then
             self.Brain.BuilderManagers[self.LocationType].FactoryManager:AddFactory(finishedUnit)
         end
         if finishedUnit:GetAIBrain():GetArmyIndex() == self.Brain:GetArmyIndex() then
-            if EntityCategoryContains(categories.ENGINEER - categories.ENGINEERSTATION, finishedUnit) then
-                dontAssignEngineerTask = false
-            end
-            self:AddUnit(finishedUnit, dontAssignEngineerTask)
+            local dontAssign = not EntityContainsCategory(finishedUnit, categories.ENGINEER - categories.ENGINEERSTATION)
+            self:AddUnit(finishedUnit, dontAssign)
         end
     end,
 
--- a/lua/sim/EngineerManager.lua
+++ b/lua/sim/EngineerManager.lua
@@ -670,11 +670,15 @@ EngineerManager = Class(BuilderManager) {
     ---@param unit Unit
     ---@param finishedUnit Unit
     UnitConstructionFinished = function(self, unit, finishedUnit)
+        local dontAssignEngineerTask = true
         if EntityCategoryContains(categories.FACTORY * categories.STRUCTURE, finishedUnit) and finishedUnit:GetAIBrain():GetArmyIndex() == self.Brain:GetArmyIndex() then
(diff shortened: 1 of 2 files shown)


lua/sim/EngineerManager.lua
```diff
@@ -670,11 +666,15 @@ EngineerManager = Class(BuilderManager) {
...
+        local dontAssignEngineerTask = true
         if EntityCategoryContains(categories.FACTORY * categories.STRUCTURE, finishedUnit) and finishedUnit:GetAIBrain():GetArmyIndex() == self.Brain:GetArmyIndex() then
             self.Brain.BuilderManagers[self.LocationType].FactoryManager:AddFactory(finishedUnit)
         end
         if finishedUnit:GetAIBrain():GetArmyIndex() == self.Brain:GetArmyIndex() then
-            self:AddUnit(finishedUnit)
+            if EntityCategoryContains(categories.ENGINEER - categories.ENGINEERSTATION, finishedUnit) then
+                dontAssignEngineerTask = false
+            end
+            self:AddUnit(finishedUnit, dontAssignEngineerTask)
```
Hdt80bro: Does this simplify to `self:AddUnit(finishedUnit, EntityContainsCategory(finishedUnit, categories.ALL - categories.ENGINEERING + categories.ENGINEERINGSTATION))`? It looks a little hard to read right now, especially with the variable definition away from where it's being used.

Hdt80bro (changes requested):

lua/sim/EngineerManager.lua
relent0r: Yea I actually thought the same thing at first, but I thought that would be harder to read for someone who wasn't familiar than what I'd done. But if everyone reckons that easier then I'll change it.

Hdt80bro: I typically prefer to read one line of code over five - it's just faster for me. Also, in this case when using boolean expressions to assign other boolean expressions instead simply using the original expression, it came across as verbose. It actually took me quite a bit of time to figure out what you wanted to do as you were using both conditions and categories that seemed to put the focus on engineering stations - I almost thought you intended `self:AddUnit(finishedUnit, EntityContainsCategory(finishedUnit, categories.ENGINEERINGSTATION))`. If you don't like how much there is going on on the one line, you could do
```Lua
local dontAssign = not EntityContainsCategory(finishedUnit, categories.ENGINEERING - categories.ENGINEERINGSTATION)
self:AddUnit(finishedUnit, dontAssign)
```

Hdt80bro: Though, frankly, just moving the `local dontAssignEngineerTask = true` to just above the if-statement to make the idiom clear would make me happy to merge this - the rest is just food for thought about code style.

relent0r: > I typically prefer to read one line of code over five - it's just faster for me. Also, in this case when using boolean expressions to assign other boolean expressions instead simply using the original expression, it came across as verbose. It actually took me quite a bit of time to figure out what you wanted to do as you were using both conditions and categories that seemed to put the focus on engineering stations - I almost thought you intended `self:AddUnit(finishedUnit, EntityContainsCategory(finishedUnit, categories.ENGINEERINGSTATION))`. If you don't like how much there is going on on the one line, you could do
> 
> ```lua
> local dontAssign = not EntityContainsCategory(finishedUnit, categories.ENGINEERING - categories.ENGINEERINGSTATION)
> self:AddUnit(finishedUnit, dontAssign)
> ```

I've updated it to be like this since that looks quite clean and easy to read.

Hdt80bro (approved):

PR #6738, opened 2025-04-19
Preliminary tweaks for TML and TMD

Basilisk3:
Various adjustments to TMLs and TMDs to improve their functionality and address balancing problems. There is still work to be done regarding these units, but the changes should increase consistency and make for more balanced gameplay.

### TMLs
- The hitboxes of all tactical missiles are increased slightly, to prevent TMDs from missing them by overshooting.
- Seraphim and UEF TMLs fly lower.
- The speed of the Seraphim TML is nerfed because the difference between its speed and that of other TMLs became too large after previous changes. At longer ranges, it was able to reach its target up to (and sometimes over) 10 seconds faster than other TMLs. With these changes, it is still excellent (and the fastest), but its lower terminal speed should now serve as a legitimate balancing factor.
- Nerf the Seraphim ACU TML's really oppressive insta-dink potential by reducing its speed.

### TMDs
- Cybran and UEF TMDs no longer run out of beamlength/lifetime, which could previously cause their projectiles to expire before reaching their target. This change should also ensure compatibility with mods that introduce missiles flying at very high altitudes.
- Unify the `MuzzleVelocity` stats of all UEF TMDs.
- Remove unnecessary firing tolerance stats, which theoretically could have caused TMDs to miss.

It would be appreciated if someone could double-check if Zappers still miss with these changes.

## Checklist


lL1l1:
It may be worth taking a look at the muzzle velocity for the TML weapons, it determines the initial speed of the TML, and when I was working on it it was a significant part of how I made Sera TML get up to speed quickly, especially for navy.

I'm sad to see Sera TML go (I over tuned it/finetuned it after all). I don't think the speed matters a huge amount in terms of TMD timing unless you're sniping early TMD or making a hidden TML battery, which are rather niche cases and thus acceptable.

Btw if UEF TML flies lower it might hit TMD at 14 speed. I set terminal speed at 13 for 30 distance as a maximum speed it can have without hitting TMD.

Basilisk3:
currentTML
This is the current balance between UEF and Seraphim TMLs. I clicked launch at the same time. The Seraphim TML arrives 10 seconds earlier, partly because it flies much faster (which makes avoiding intermediary TMDs easier) and because it launches more quickly (easier to react to mobile units). In my opinion some kind of nerf is necessary here, especially when considering that the UEF missiles are the second fastest flying ones.

- **Seraphim TML:** time until impact: 18s --> 22s
- **UEF TML:** time until impact: 28s --> 25s
- **Aeon and Cybran TMLs:** time until impact: 28s (unchanged)

The changes are still WIP, so if you think the Seraphim TMLs lose too much of their identity, I can adjust them accordingly.

lL1l1:
I feel like the most significant issue is Sera ACU TML being used in combat vs other ACUs. Without any downside in HP or significant upsides on other faction TML, I will agree that it's badly imbalanced, even though the situations where Sera TML excels in are not very common aside from the ACU TML dueling other ACUs. 

The only reason to ignore the imbalance is just some notion of unique design, but then other factions would need actual upsides to their TML, for example: Cybran functional split missile accuracy for bypassing TMD, Aeon ??, UEF bypassing all terrain with their altitude. Sera could actually have more speed if it had 1 HP, which would be a real downside compared to the minor example upsides. These things are outside the scope of the PR though.

Basilisk3:
How about:

- Aeon: 3 HP but slower
- UEF: slightly larger explosion radius
- Seraphim: fast missiles
- Cybran: split missiles

lL1l1:
Not sure how benefits of 3 HP compares with downsides of being slower if in the end it still has to not directly hit TMD and not pass directly over TMD.
UEF having a larger damage radius doesn't seem like a fun difference to me. Just makes it easier to hit ACUs and makes storages die on mex.

Basilisk3:
The Aeon change is more of a faction diversity change, similar to how FAF changed the Evensong. As for the UEF TML, I would propose a radius of at most 2.5, so the Seraphim TML is not straight up better in all situations.

Basilisk3:
I did some more tests, and I think this is ready to be reviewed. I tried to keep it simple—the PR aims to address the low-hanging fruits (e.g., functionality) so another PR can work out the more complicated aspects, like introducing more faction diversity, later.

changelog/snippets/balance.6738.md
```diff
@@ -0,0 +1,14 @@
+- (#6738) Various adjustments to TMLs and TMDs to improve their functionality and address balancing problems. There is still work to be done regarding these units, but the changes should increase consistency and make for more balanced gameplay.
+
+  **Tactical Missile Launchers (TMLs)**
+
+    - The hitboxes of all tactical missiles are increased slightly, to prevent TMDs from missing them by overshooting.
+    - Seraphim and UEF TMLs fly lower.
+    - The speed of the Seraphim TML is nerfed because the difference between its speed and that of other TMLs became too large after previous changes. At longer ranges, it was able to reach its target up to (and sometimes over) 10 seconds faster than other TMLs. With these changes, it is still excellent (and the fastest), but its lower terminal speed should now serve as a legitimate balancing factor.
+    - Nerf the Seraphim ACU TML's really oppressive insta-dink potential by reducing its speed.
```
lL1l1:
```suggestion
    - Nerf the Seraphim ACU TML's really oppressive close range combat potential by reducing its speed.

```

changelog/snippets/balance.6738.md
```diff
@@ -0,0 +1,14 @@
+- (#6738) Various adjustments to TMLs and TMDs to improve their functionality and address balancing problems. There is still work to be done regarding these units, but the changes should increase consistency and make for more balanced gameplay.
+
+  **Tactical Missile Launchers (TMLs)**
+
+    - The hitboxes of all tactical missiles are increased slightly, to prevent TMDs from missing them by overshooting.
+    - Seraphim and UEF TMLs fly lower.
+    - The speed of the Seraphim TML is nerfed because the difference between its speed and that of other TMLs became too large after previous changes. At longer ranges, it was able to reach its target up to (and sometimes over) 10 seconds faster than other TMLs. With these changes, it is still excellent (and the fastest), but its lower terminal speed should now serve as a legitimate balancing factor.
```
lL1l1: Terminal speed is a term in the code but not known by players.

changelog/snippets/balance.6738.md
Basilisk3: Looks like I forgot to change this when copying over the text from the PR.

changelog/snippets/balance.6738.md
Basilisk3: If you have a better alternative, I am open to changing the wording. I think terminal speed is relatively self-explanatory though. Terminal velocity would also be fitting but sounds even more technical.

lL1l1: I think you should split the nerf of the max speed and the terminal speed into two separate sentences, that will make it a bit easier to understand while reading, thus allowing us to fit in a few more words.
```suggestion
    - The max speed of the Seraphim TML is reduced because it was too fast compared to other TMLs; for example, it was able to reach its target over 10s faster at longer ranges. Additionally, the terminal speed of the missile as it nears its target is also reduced, so that it serves as a more legitimate balancing factor.

```

lL1l1 (approved):

PR #6739, opened 2025-04-20
Refactor custom world rendering of shapes

Garanas:

With the recent surge in demand for rendering shapes (#6725 and #6726) I noticed that the existing interface can be a little lacking. With this pull request I try to improve the situation before it's too late to go back!

Verified that the shape used for an area attack command still works.

## Additional context

Replaces the word 'Renderable' with just 'Shape'. To keep the pull request small I only factor out the logic of the world view. If we all agree with this naming convention then I can refactor the circle, square and line accordingly.

## Checklist

### DIFF
--- a/lua/ui/controls/components/RenderShapesComponent.lua
+++ b/lua/ui/controls/components/RenderShapesComponent.lua
@@ -22,7 +22,7 @@
 
 --- A component to encapsulate the logic used to render shapes in the world.
 ---@class UIRenderShapesComponent
----@field Renderables table<string, Renderable>
+---@field Shapes table<string, Renderable>
 RenderShapesComponent = ClassSimple {
 
     -- Related binary patches that make this possible:
@@ -32,29 +32,32 @@ RenderShapesComponent = ClassSimple {
 
     ---@param self UIRenderShapesComponent | WorldView
     __post_init = function(self)
-        self.Renderables = {}
+        self.Shapes = TrashBag()
     end,
 
-    --- Register a renderable to render each frame
+    --- Adds a shape that we should render to the world view.
+    --- 
+    --- The shape is stored in a weak table. If there's no other active  
+    --- reference to the shape then the garbage collector will clean it up.
     ---@param self UIRenderShapesComponent | WorldView
-    ---@param renderable Renderable
+    ---@param shape Renderable
     ---@param id string
-    RegisterRenderable = function(self, renderable, id)
-        self.Trash:Add(renderable)
-        self.Renderables[id] = renderable
+    AddShape = function(self, shape, id)
+        self.Trash:Add(shape)
+        self.Shapes[id] = shape
 
-        if not table.empty(self.Renderables) then
+        if not table.empty(self.Shapes) then
             self:SetCustomRender(true)
         end
     end,
 
-    --- Unregister a renderable
+    --- Removes a shape. It will no longer be rendered to the world view.
     ---@param self UIRenderShapesComponent | WorldView
     ---@param id string
-    UnregisterRenderable = function(self, id)
-        self.Renderables[id] = nil
+    RemoveShape = function(self, id)
+        self.Shapes[id] = nil
 
-        if table.empty(self.Renderables) then
+        if table.empty(self.Shapes) then
             self:SetCustomRender(false)
         end
     end,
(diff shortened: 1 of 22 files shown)


lua/ui/controls/components/RenderShapesComponent.lua
```diff
@@ -0,0 +1,98 @@
...
+
+    --- Is called each frame for each shape to render it. 
+    ---@param self UIRenderShapesComponent | WorldView
+    ---@param shape Renderable
+    ---@param delta number
+    ---@return boolean  # if false then some error happened during rendering. 
+    OnRenderShape = function(self, shape, delta)
+        -- sanity check on the data
+        if not shape or not shape.OnRender then
+            return false
```
lL1l1: if the shape doesn't have an OnRender shouldn't there be a warning?

lua/ui/controls/components/RenderShapesComponent.lua
```diff
@@ -0,0 +1,98 @@
...
+RenderShapesComponent = ClassSimple {
+
+    -- Related binary patches that make this possible:
+    -- - https://github.com/FAForever/FA-Binary-Patches/pull/47
+    -- - https://github.com/FAForever/FA-Binary-Patches/pull/111
+    -- - https://github.com/FAForever/FA-Binary-Patches/pull/112
+
+    ---@param self UIRenderShapesComponent | WorldView
+    __post_init = function(self)
+        self.Shapes = TrashBag()
```
lL1l1: Should add a comment somewhere that specifies that you can call `WorldView.Shapes:Destroy()` to destroy all shapes.

Otherwise the trashbag's `Destroy` functionality isn't used, maybe it should go into `WorldView.OnDestroy`? That way instead of waiting for the shapes to be garbage collected they can be cleaned up promptly.

lua/ui/controls/components/RenderShapesComponent.lua
```diff
@@ -0,0 +1,98 @@
...
+--** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+--** AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+--** LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+--** OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+--** SOFTWARE.
+--******************************************************************************************************
+
+--- A component to encapsulate the logic used to render shapes in the world.
+---@class UIRenderShapesComponent
+---@field Shapes table<string, Renderable>
```
lL1l1:
```suggestion
---@field Shapes table<string, Renderable> | TrashBag
```

lua/ui/controls/components/RenderShapesComponent.lua
Garanas: The reason I use a trash bag here is because it is a weak table. Meaning if all references to the shape are lost, then it won't linger around just because someone forgot to deregister it.

It should go into WorldView.OnDestroy perhaps, yes. Not sure about the possibility to just destory all the shapes - it sounds like something you don't want to do.

Garanas: Thinking about it more - pings renew. Should shapes renew too? I'm not sure what the right approach here is. We can't just go and update all existing references.

lua/ui/controls/components/RenderShapesComponent.lua
Garanas: Applied in https://github.com/FAForever/fa/pull/6739/commits/2ed6648745cc75d7af9b11f04bf7b3479059e7ac

lua/ui/controls/components/RenderShapesComponent.lua
Garanas: Applied in: https://github.com/FAForever/fa/pull/6739/commits/1fd36184679ffdc949b0a35a0ca5749652ee0e13

lL1l1: Pings renew for army changes because the ping feature depends on perspective. Shapes are too abstract to depend on perspective, so it should be up to whatever feature is creating the shapes to renew them once perspective changes.

lL1l1 (commented):
I can approve this after a snippet is added.

Garanas:
I'd like to first merge https://github.com/FAForever/fa/pull/6740/files into this pull request, before merging this pull request into develop.

lL1l1 (approved):

PR #6740, opened 2025-04-20
Refactor current shape definitions

Garanas:

Refactors how the shapes are defined. 

Verified that the shape used for an area attack command still works.

## Additional context

With the introduction of a line shape it also becomes possible to create more complex shapes, such as a series of line segments or a box that we can rotate. We can then abstract that with a `UIComplexShape` class that introduces functions to apply (matrix) operations to rotate, translate or even scale shapes that are based on line segments!

## Checklist

### DIFF
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -1287,28 +1287,46 @@ end
 function UIZoomTo(units, seconds)
 end
 
---- Draws a circle in world. Must be called within `WorldView:OnRenderWorld`
----@param pos Vector
+--- Draws a circle in world. Must be called from `WorldView:OnRenderWorld` or it won't draw anything.
+---@param pos Vector  # in world coordinates
 ---@param size number
 ---@param color Color
 ---@param thickness? number
 function UI_DrawCircle(pos, size, color, thickness)
+
+  -- Introduced by an assembly function, see also:
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/47
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/111
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/112
+
 end
 
---- Draws a rectangle in world. Must be called within `WorldView:OnRenderWorld`
----@param pos Vector
+--- Draws a rectangle in world. Must be called from `WorldView:OnRenderWorld` or it won't draw anything.
+---@param pos Vector  # in world coordinates
 ---@param size number
 ---@param color Color
 ---@param thickness? number
 function UI_DrawRect(pos, size, color, thickness)
+
+  -- Introduced by an assembly function, see also:
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/47
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/111
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/112
+
 end
 
---- Draws a line in world. Must be called within `WorldView:OnRenderWorld`
----@param position1 Vector
----@param position2 Vector
+--- Draws a line in world. Must be called from `WorldView:OnRenderWorld` or it won't draw anything.
+---@param position1 Vector  # in world coordinates
+---@param position2 Vector  # in world coordinates
 ---@param color Color
 ---@param thickness? number
 function UI_DrawLine(position1, position2, color, thickness)
+
+  -- Introduced by an assembly function, see also:
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/47
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/111
+  -- - https://github.com/FAForever/FA-Binary-Patches/pull/112
+
 end
 
 ---
--- a/engine/User.lua
+++ b/engine/User.lua
@@ -1287,7 +1287,7 @@ end
 function UIZoomTo(units, seconds)
 end
 
----Draws circle in world. Must be called within `WorldView:OnRenderWorld`
+--- Draws a circle in world. Must be called within `WorldView:OnRenderWorld`
 ---@param pos Vector
 ---@param size number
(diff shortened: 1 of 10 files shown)


lua/ui/game/shapes/SquareShape.lua
```diff
@@ -75,11 +66,38 @@ UIRenderableSquare = ClassSimple {
...
+---@param ox number # in world coordinates
+---@param oy number # in world coordinates
+---@param oz number # in world coordinates
+---@param size number
+---@param color Color
+---@param thickness number
+---@return UISquareShape
+CreateSquareShapeXYZ = function(worldview, ox, oy, oz, size, color, thickness)
+    local shape = SquareShape(worldview, ox, oy, oz, size, color, thickness) --[[@as UISquareShape]]
+    return shape
```
GodFuper: ```return SquareShape(worldview, ox, oy, oz, size, color, thickness)```

lua/ui/game/shapes/LineShape.lua
```diff
@@ -0,0 +1,67 @@
...
+--- Creates a line segment with the given properties.
+---@param worldview WorldView
+---@param p1 Vector   # in world coordinates
+---@param p2 Vector   # in world coordinates
+---@param color Color
+---@param thickness number
+---@return UILineShape
+CreateLineShape = function(worldview, p1, p2, color, thickness)
+    local shape = LineShape(worldview, p1, p2, color, thickness) --[[@as UILineShape]]
+    return shape
```
GodFuper: ```retutn LineShape(worldview, p1, p2, color, thickness)```

lua/ui/game/shapes/SquareShape.lua
Garanas: Sadly, the plugin loses the type information if we write it that way. Especially if anyone in the long term adds something that needs to happen for each instance, after it is done being created. 

Practically you're right, but I favor the type information here.

engine/User.lua
```diff
@@ -1287,22 +1287,30 @@ end
...
----Draws rectangle in world. Must be called within `WorldView:OnRenderWorld`
+--- Draws a rectangle in world. Must be called within `WorldView:OnRenderWorld`
 ---@param pos Vector
 ---@param size number
 ---@param color Color
 ---@param thickness? number
 function UI_DrawRect(pos, size, color, thickness)
 end
 
+--- Draws a line in world. Must be called within `WorldView:OnRenderWorld`
```
lL1l1: I think there should be a message about the assembly patch in the UI shape drawing functions so that people don't get it confused with vanilla.
Maybe also mention `SetCustomRender` needs to be enabled, although that's already indirectly mentioned in the comment for `OnRenderWorld`, so maybe just that comment needs to be more specific. The language in that comment should also be exactly the same as in the `SetCustomRender` description so that it is easy to search.

lua/ui/game/shapes/Shape.lua
```diff
@@ -0,0 +1,85 @@
...
+        end
+
+        self:OnRender(delta)
+    end,
+
+    --- The logic to render the shape.
+    ---@param self UIShape
+    ---@param delta number
+    OnRender = function(self, delta)
+        -- to be defined by the subclass
```
lL1l1, line 61: Maybe make it error so it gets deleted in worldview and gives a message + traceback for the developer.

lua/ui/game/shapes/SquareShape.lua
lL1l1: I don't understand how the type information is lost. Shouldn't the `@as` comment give type information to the object return by the Class call (classes don't return the correct type upon call because generics are not working in our version of the server iirc)?

Garanas: I think you're right. This is with the local:

image

And this is when I apply the `@as` on the return value:

image

It still gives me the error message that I argued about here. I did not think it would work. It took a few seconds for the plugin to figure out there's a miss match. Must've been too impatient at the time.

I'll adjust it, great thinking!

Garanas: Applied in https://github.com/FAForever/fa/pull/6740/commits/075e345bc91e435b96e0f7b4bd75ff46d2c5ec54

lua/ui/game/shapes/LineShape.lua
Garanas: Applied in https://github.com/FAForever/fa/pull/6740/commits/075e345bc91e435b96e0f7b4bd75ff46d2c5ec54

lua/ui/game/shapes/Shape.lua
Garanas: Great thinking, applied in https://github.com/FAForever/fa/pull/6740/commits/1c219dba61f0fb2dc1b2fd9abb309dcf8e4c0cd1

engine/User.lua
Garanas: Applied it in: https://github.com/FAForever/fa/pull/6740/commits/c9cb856792b318c8b627244b5a353f9b9d5c49ec, I've intentionally put it in the function body as I don't feel it adds much to the description of the functionality itself. What do you think?

Garanas:
A few more questions about this design:

At the moment a shape can only be rendered by one worldview. This means that if you want it to be rendered across multiple worldviews, then you'll have to define the same shape with the same data multiple times. This limitation exists now because we pass the worldview as an argument to the constructor. But what if we do not pass it to the construct, and instead provide a function to register the shape to a world view? That way the same instance can be used to render on multiple worldviews. 

There's logic to hide the shape. Does it make sense to exist?

What about complicated shapes, such as a series of line segments? Should those originate from 'ComplexShape' class that inherits from the 'Shape' class? The reason I am asking is because I don't want to fall into the 'inheritance' trap. I'd rather make it a composition. But I'm not sure how to make this 'composable'.

PR #6741, opened 2025-04-21
Refactor screen indicator

Garanas:

Factors out the camera indicator into a separate and extendable class. The camera indicator was previously baked into the world view. By factoring out we reduce the clutter in the world view class.

Tested using split and non-split world views. 

## Additional context

And we make it possible to re-use the camera indicator, for paintings for example 😃 !

## Checklist

### DIFF
--- a/lua/ui/game/common/AnimatedGlow.lua
+++ b/lua/ui/game/common/AnimatedGlow.lua
@@ -23,10 +23,9 @@
 local LayoutHelpers = import("/lua/maui/layouthelpers.lua")
 local Bitmap = import("/lua/maui/bitmap.lua").Bitmap
 
+--- A small utility class that will 'animate' a bitmap by pulsing it on and off by manipulating its alpha channel.
 ---@class UIAnimatedGlow : Bitmap
 AnimatedGlow = Class(Bitmap) {
-
-    DebugName = "AnimatedGlow",
 
     ---@param self UIAnimatedGlow
     ---@param parent Control
@@ -39,6 +38,7 @@ AnimatedGlow = Class(Bitmap) {
             :End()
     end,
 
+    --- An 'animation' that will pulse over the current time to influence the opacity of the texture.
     ---@param self UIAnimatedGlow
     ---@param delta number
     OnFrame = function(self, delta)
--- a/lua/ui/game/common/AnimatedGlow.lua
+++ b/lua/ui/game/common/AnimatedGlow.lua
@@ -1,0 +1,57 @@
+--******************************************************************************************************
+--** Copyright (c) 2022  Willem 'Jip' Wijnia
+--**
+--** Permission is hereby granted, free of charge, to any person obtaining a copy
+--** of this software and associated documentation files (the "Software"), to deal
+--** in the Software without restriction, including without limitation the rights
+--** to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+--** copies of the Software, and to permit persons to whom the Software is
+--** furnished to do so, subject to the following conditions:
+--**
+--** The above copyright notice and this permission notice shall be included in all
+--** copies or substantial portions of the Software.
+--**
+--** THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+--** IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+--** FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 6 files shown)


lua/ui/game/common/AnimatedGlow.lua
```diff
@@ -0,0 +1,57 @@
...
+--** OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+--** SOFTWARE.
+--******************************************************************************************************
+
+local LayoutHelpers = import("/lua/maui/layouthelpers.lua")
+local Bitmap = import("/lua/maui/bitmap.lua").Bitmap
+
+--- A small utility class that will 'animate' a bitmap by pulsing it on and off by manipulating its alpha channel.
+---@class UIAnimatedGlow : Bitmap
+AnimatedGlow = Class(Bitmap) {
```
lL1l1: Since this is a general purpose utility class, the glow rate should be adjustable.

lL1l1 (changes requested):
Works without errors but still needs some details fixed.
- Needs a snippet
- It wasn't mentioned, but this refactor fixes the glow's pulse duration being influenced by framerate. 👍
- The refactor has a bug where the indicator can clip offscreen when it is pointing to an indicator near the corner but not outside the corner. This wasn't present before the refactor.
{DEEE0DD4-6A9C-49CE-A5E3-34B4A59CB304}

lua/ui/game/common/AnimatedGlow.lua
Garanas: Applied in https://github.com/FAForever/fa/pull/6741/commits/aecb2e7147e68299dd7f19226284eb169e2b38e2

speed2CZ:
What about having the all the textures in a single table indexes by colours. And having just one function to create the marker with the colour passed as the argument. That way you dont need a function per each colour and it would be easier to add a new one if anyone decides to do so, as it would be just a new entry in the table.

Garanas:
@speed2CZ Do you mean like this:

```lua
local TexturePresets = {
    Blue = {...},
    Red = {...},
    Yellow = {...}
}

---@param parent Control
---@param worldView WorldView
---@param preset 'Yellow' | 'Red' | 'Blue'
---@param target Vector # in world coordinates
---@return UIOffscreenMarkerIndicator
CreateOffScreenMarkerIndicatorFromPreset = function(parent, worldView, target, preset)
    local texturePreset = TexturePresets[preset]
    if not texturePreset then
        error(string.format("Invalid texture preset for off screen marker indicator: %s", preset))
    end

    local indicator = OffscreenMarkerIndicator(parent, worldView, texturePreset, target) --[[@as UIOffscreenMarkerIndicator]]
    return indicator
end
```

I think it would make sense if we can have dozen(s) of presets. Are you having any plans 😃 ?

lua/ui/game/common/AnimatedGlow.lua
Garanas: Snippet in https://github.com/FAForever/fa/pull/6741/commits/307c108dc611702020ac4e6517fa1c259c598383

Garanas: Attempt to fix the bug about the corners: https://github.com/FAForever/fa/pull/6741/commits/8a2342508ef4ebe8a13873663c828a6d3c08f6dd.

It now snaps a little bit to the corner, I'm not sure how to make it pixel-perfect smooth without interpreting the layout files.

speed2CZ:
It makes sense even now, at least to me. Its less code = more simple :D
And instead of erroring out, you could just use the default colour

Garanas:
It's a different way of structuring it. I'm fine either way. The advantage to your approach is that you can pass around, serialize and deserialize the color value.

edit: implemented it in https://github.com/FAForever/fa/pull/6741/commits/569bf2d17ebc7ccd95bf330b19306a4dbae49dcf

Hdt80bro:
Currently, there's a lot of paths I have to read for a single texture set - 25 filename which all begin with the same prefix. I think I'd like a generator function for readability.

Garanas:
@Hdt80bro How about https://github.com/FAForever/fa/pull/6741/commits/84df7d3c65c2f164ae7146c9abe449400ee6ee78 ?

Hdt80bro:
Looks good! I would have gone one step further and created another function that returns each individual preset that the generator then calls nine times, but looking at what we have again, I'm realizing that would be overkill.

lua/ui/game/common/AnimatedGlow.lua
lL1l1: I was mistaken, it didn't snap to corners before, still it's a nice visibility improvement. What do you mean by layout files? I opened the texture and the left arrow texture has 10px of empty space from the bottom edge, so the snapping is caused by that 10px of empty space.

lL1l1 (approved):

PR #6742, opened 2025-04-22
Fix shield assist issuing orders to an empty table

lL1l1:
## Issue
Reported on [Discord](https://discord.com/channels/197033481883222026/1363827638895902912/1363827638895902912) for replay [24723646](https://replay.faforever.com/24723646) (faf develop so it will desync in the future), and with the following error message:
```
WARNING: Error running OnDamage script in Entity  at 2531f880: ...faforever\replaydata\gamedata\lua.nx5\lua\shield.lua(605): Expected a game object. (Did you call with '.' instead of ':'?)
         stack traceback:
             [C]: ?
             ...faforever\replaydata\gamedata\lua.nx5\lua\shield.lua(605): in function `ApplyDamage'
             ...faforever\replaydata\gamedata\lua.nx5\lua\shield.lua(653): in function `ApplyDamage'
             ...faforever\replaydata\gamedata\lua.nx5\lua\shield.lua(520): in function <...faforever\replaydata\gamedata\lua.nx5\lua\shield.lua:511>
             [C]: in function `DamageArea'
             ...eplaydata\gamedata\lua.nx5\lua\sim\collisionbeam.lua(133): in function `DoDamage'
             ...eplaydata\gamedata\lua.nx5\lua\sim\collisionbeam.lua(313): in function `OnImpact'
             ...im\collisionbeams\orbitaldeathlasercollisionbeam.lua(32): in function <...im\collisionbeams\orbitaldeathlasercollisionbeam.lua:23>
```

The error is caused by fully clearing the guards table and still using it to issue orders afterwards.

- Keep track if we left any guards in the table and only issue the orders if we did.

To reproduce the error:
1. Damage a shield
2. Assist the shield with low enough buildpower that it doesn't repair fully before the next damage instance
3. Queue another order before the next damage instance arrives, do this for all guards.
4. The next damage instance no longer causes an error when it arrives.

## Checklist

### DIFF
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -595,16 +595,22 @@ Shield = ClassShield(moho.shield_methods, Entity) {
                 local guards = UnitGetGuards(owner)
                 if not TableEmpty(guards) then
                     -- filter out guards with something queued after the shield assist order, as to not delete clear their queue
+                    -- keep track if we have guards left over to issue orders to
+                    local hasValidGuards = false
                     for i, guard in guards do
                         if TableGetn(UnitGetCommandQueue(guard)) >= 2 then
                             guards[i] = nil
+                        elseif not hasValidGuards then
+                            hasValidGuards = true
                         end
                     end
 
-                    -- For the filtered guards, clear their assist order, order repair, then re-add the assist order after
-                    IssueClearCommands(guards)
-                    IssueRepair(guards, owner)
-                    IssueGuard(guards, owner)
+                    if hasValidGuards then
+                        -- For the filtered guards, clear their assist order, order repair, then re-add the assist order after
+                        IssueClearCommands(guards)
+                        IssueRepair(guards, owner)
+                        IssueGuard(guards, owner)
+                    end
                 end
             end
 
--- a/lua/shield.lua
+++ b/lua/shield.lua
@@ -595,17 +595,13 @@ Shield = ClassShield(moho.shield_methods, Entity) {
                 local guards = UnitGetGuards(owner)
                 if not TableEmpty(guards) then
                     -- filter out guards with something queued after the shield assist order, as to not delete clear their queue
-                    -- keep track if we have guards left over to issue orders to
-                    local hasValidGuards = false
                     for i, guard in guards do
                         if TableGetn(UnitGetCommandQueue(guard)) >= 2 then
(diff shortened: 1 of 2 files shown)


lua/shield.lua
```diff
@@ -595,16 +595,22 @@ Shield = ClassShield(moho.shield_methods, Entity) {
...
+                        elseif not hasValidGuards then
+                            hasValidGuards = true
                         end
                     end
 
-                    -- For the filtered guards, clear their assist order, order repair, then re-add the assist order after
-                    IssueClearCommands(guards)
-                    IssueRepair(guards, owner)
-                    IssueGuard(guards, owner)
+                    if hasValidGuards then
```
speed2CZ: Why don't you just check again if the guard table isnt empty?

lL1l1: Since it's in the damage part of the shield script it has to be optimized and it's better to just use this local instead of an entire `table.getn`, `next`, or `table.empty` function.

Hdt80bro (approved):

lua/shield.lua
Hdt80bro: Have you tested it? Even my first thought was that it was less optimal to test if `hasValidGuards` was false before setting it to true instead of just always setting it to true in an else statement, but it could also just be a coin toss. I'd probably grab for `next` first as that's a C function that's going to be fast and gets rid of any of the overhead in the loop you're currently dealing with. It's the kind of thing I haven't tested yet, but if we're being this speculative about performance, it likely doesn't matter much either way.

lL1l1: I didn't test it. Yeah I didn't realize it's checking `next` once compared to checking the boolean during the entire loop.

Garanas: I think there's other, more significant, overhead in this function then whether we make a call to `table.empty` or not. 

As an example, the use of [Unit:GetCommandQueue](https://github.com/FAForever/fa/blob/develop/lua/sim/Unit.lua#L5047-L5063) allocates a table. And there's a lot of logic involved to populate that table. And this happens for every unit.

I'd argue for readability that we should just call `table.empty`.

edit: in another PR it may be interesting to add a `GetCommandQueueCount` function that only retrieves the result of the engine and counts that. It would still allocate a table, but at least we would not be iterating over it in Lua (for every unit).

PR #6743, opened 2025-04-22
Fix error when air staging gets killed when dead

lL1l1:
## Issue
The air staging's transport status goes away when it's declared dead through the first C function call, which can cause an error because lua allows units to keep being killed after being killed once.
```
warning: Error running OnDamage script in Entity xsb5202 at 28315608: Unit:TransportDetachAllUnits can only be called for transports
         stack traceback:
         	[C]: in function `TransportDetachAllUnits'
         	...\lua\sim\units\airstagingplatformunit.lua(34): in function `Kill'
         	...\lua\sim\unit.lua(1402): in function <...\lua\sim\unit.lua:1378>
         	...\lua\sim\units\structureunit.lua(705): in function `DoTakeDamage'
         	...\lua\sim\unit.lua(1368): in function <...\lua\sim\unit.lua:1352>
         	[C]: in function `DamageArea'
         	...\lua\sim\collisionbeam.lua(133): in function `DoDamage'
         	...\lua\sim\collisionbeam.lua(313): in function `OnImpact'
         	...\lua\sim\collisionbeams\phasonlasercollisionbeam.lua(30): in function <...\lua\sim\collisionbeams\phasonlasercollisionbeam.lua:21>
```

- Check if the air staging is dead before detaching all units

Spawn a Galactic Colossus and a few enemy air stagings, there should be no errors.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6665.md
+++ b/changelog/snippets/balance.6665.md
@@ -1,1 +1,1 @@
-- (#6665) Air staging now detaches aircraft when killed or ctrl-k'd (to work around an engine bug where aircraft get stuck inside air staging).
+- (#6665, #6743) Air staging now detaches aircraft when killed or ctrl-k'd (to work around an engine bug where aircraft get stuck inside air staging).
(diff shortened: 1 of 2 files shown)


Hdt80bro (approved):

speed2CZ:
Why is Kill called twice?

Hdt80bro:
Where do you see that?

lL1l1:
Because there's nothing stopping a dead unit from getting damaged when it's dead.
https://github.com/FAForever/fa/blob/9120b9c0bc5e72c1c690171f502e3e32c04d4bfd/lua/sim/Unit.lua#L1373-L1405

Units stop colliding with lasers/projectiles after they die, but AoE damage bypasses that.

PR #6744, opened 2025-04-22
Buff Cybran navy's missile deflectors

lL1l1:
## Issue
[Discord balance team thread](https://discord.com/channels/197033481883222026/1314781654841299014). Cybran navy is unable to properly defend itself against missile ships after they were changed to deflector type TMD with heavily nerfed fire rate and range.

Increase the fire rate and range to match standard naval TMD.

Cybran cruiser/carrier TMD:
- Range: 20/26 -> 60
- Max target height: 8/10 -> infinite
- Reload time: 4/2.5 seconds -> 1 second

Cybran cruisers/carriers can now defend fully against enemy missile cruisers. They can even kill UEF cruisers over some time, so it's not like the missiles being sent back are completely useless.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6744.md
+++ b/changelog/snippets/balance.6744.md
@@ -1,6 +1,6 @@
-- (#6744) Buff Cybran navy's missile deflector TMD so that they can defend against missile cruisers of other factions. Since deflectors ignore missile HP, they are given less fire rate than the gun TMD of UEF and Seraphim.
+- (#6744) Buff Cybran navy's missile deflector TMD so that they can defend against missile cruisers of other factions. Since deflectors ignore missile HP, they are given less fire rate and range than the gun TMD of UEF and Seraphim.
 
   Cybran cruiser (URS0202) and carrier (URS0303) Missile Deflectors:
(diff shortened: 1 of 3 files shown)


Garanas:
This is a **massive** buff for a unit that ignores the health of the missiles it intercepts. I'm fine either way, but maybe first increase the range to 40 and half the fire rate?

lL1l1:
- Reload 1.0 -> 1.9 seconds. This way it can push back against Sera's 1 HP, 2s reload missiles. It still defends against UEF's volley of 4 2 HP missiles spaced 1.5 seconds apart.
- Range 60 -> 44 which lets it defend a spread out navy but still gives some extra opportunity to hit stuff compared to gun TMD since it ignores missile HP.

I tested a small formation of frigates around the cruiser and groundfired too to see how far away from the cruiser the TMD of Cybran compared to Sera can defend.
I also tested Sera cruiser and saw that it lets UEF cruiser missiles get quite close, but I can't really compromise on range or fire rate to make the equivalent in Cybran.

Rhaelya (approved):
Everything looks in order

PR #6752, opened 2025-04-27
Add a workflow to bake blueprint properties of props

Garanas:

Introduces a workflow to make it easier to bake prop blueprints. This workflow uses the [BrewWikiGen](https://github.com/The-Balthazar/BrewWikiGen) tool made by @The-Balthazar. We entirely ignore the wiki part of the tool, and instead just use it to bake in properties into blueprints.

With this workflow it now becomes possible to mass-process blueprints. I'll try and think through a proposal for the exact procedure the following days.

Some advantages are:

- Reduce loading time.
- Reduce post processing during run time.
- Blueprints reflect the 'truth' better.

See also [this commit](https://github.com/Garanas/fa/commit/2b4a7d96b8d370e24bc3d906b9cbf7e04271a479) of a fork of the fa repository, that is a result of [this workflow run](https://github.com/Garanas/fa/actions/runs/14690995272).

## Additional context

This allows us to bake in properties of blueprints. 

The capability to do the baking so conveniently is entirely new. It was previously either out of reach (before the creation of the tool by Balthazar), or difficult to setup (using the tool of Balthazar).

With this new capability there's also a new type of question: what properties should we bake in, and what properties are better to (also) be applied during runtime? The advantage of post processing at run time is that it's also applied to all mods.

This is a difficult question! Some values are critical for the functionality of FAForever, and others are too sensitive that I strongly believe they should always be overwritten at runtime. As a few (quick) examples:

- `AI.GuardScanRadius` of unit blueprint: extremely sensitive to the performance of the simulation.
- `Weapon.TargetCheckInterval` of unit blueprint: extremely sensitive to the performance of the simulation.
- `Defense.ThreatXYZ` of unit blueprint: sensitive to the performance of AIs, as balance changes these should change accordingly.
- `Display.Mesh.LODs` of unit and prop blueprint: sensitive to the rendering performance of the game. 
- `ExtractWreckageBlueprint` of general post processing: Introduction of wreck blueprints.
- `ExtractBuildMeshBlueprint` of general post processing: Introduction of build blueprints.
- `ExtractCloakMeshBlueprint` of general post processing: introduction of cloak blueprints.

And a few examples that may be interesting to bake in:

- `HandleUnitWithBuildPresets` of general post processing: Creation of SACU presets.

## Features for future pull requests

- Add support for unit blueprints.
- Add support for projectile blueprints.
- Add support for beam blueprints.
- Add support for emitter blueprints.
- Bake all blueprints when deploying to `deploy/faf`, `deploy/fafbeta` or `deploy/fafdevelop`.

## Checklist


.github/workflows/bake-blueprints.yaml
```diff
@@ -0,0 +1,109 @@
...
+          sparse-checkout-cone-mode: |
+            .github/*
+            loc
+            lua/ui/help/unitscription.lua
+            lua/ui/help/tooltips.lua
+            lua/sim/AdjacencyBuffs.lua
+            lua/system/Blueprints.lua
+            units/*_unit.bp
+            units/*_LOD0.scm
+            projectiles
```
lL1l1: The fa checkout takes the longest time alongside the lua installation, so it's worth looking at. This list of patterns doesn't actually have prop blueprints, so I think sparse checkout is just not being used because the [documentation](https://github.com/actions/checkout#:~:text=%23%20Default%3A%20null-,sparse%2Dcheckout%3A%20%27%27,-%23%20Specifies%20whether) says that `sparse-checkout-cone-mode` is a boolean to use "cone mode" while `sparse-checkout` is the correct place to place patterns to use for git.
This same mistake (if I understood correctly) is in the total mayhem repo btw.

After that, the checkout should be slimmed down to just blueprint files if the wiki gen baking works with that.

.github/workflows/scripts/baking/bake-props.lua
```diff
@@ -0,0 +1,209 @@
...
+--------------------------------------------------------------------------------
+
+--[{ ---------------------------------------------------------------------- ]]--
+--[[ Inputs -- NOTE: Mod input files must be valid lua                      ]]--
+--[[ ---------------------------------------------------------------------- ]]--
+local OutputDirectory = "D:/faf-development/fa.wiki/"
+local WikiGeneratorDirectory = "D:/faf-development/BrewWikiGen/"
+local FADirectory = "D:/faf-development/fa/"
+
+-- This section deals with overriding the OutputDirectory and WikiGeneratorDirectory if required
```
lL1l1: At a glance the comment looks like it describes the function (that's how it is for all other lua in this repo). I think we should use a region and add a newline.
```suggestion
--#region This section deals with overriding the OutputDirectory and WikiGeneratorDirectory if required by command-line arguments

```

.github/workflows/scripts/baking/bake-props.lua
```diff
@@ -0,0 +1,209 @@
...
+end
+if args["FADirectory"] then
+    FADirectory = args["FADirectory"]
+end
+
+print("Directories set")
+print("Output Directory: " ..OutputDirectory)
+print("Wiki Generator Directory: " ..WikiGeneratorDirectory)
+print("FA Directory: " ..FADirectory)
+
```
lL1l1, line 45:
```suggestion
--#endregion

```

changelog/snippets/other.6752.md
```diff
@@ -0,0 +1,3 @@
+- (#6752) Add a GitHub Action workflow to bake properties into blueprints of props
+
+This enables us to reduce the blueprint processing during run time. That will reduce the loading time of users.
```
lL1l1: - Indentation is needed again after double newline to match the bullet point's indentation.

```suggestion
  This enables us to reduce the blueprint processing during run time, which will reduce the loading time for users.
```

.github/workflows/bake-blueprints.yaml
```diff
@@ -0,0 +1,109 @@
...
+      reference:
+        required: true
+        type: choice
+        description: The reference (branch or tag) to use to compile the changelog snippets from
+        options:
+          - develop
+          - staging/faf
+          - staging/fafbeta
+          - staging/fafdevelop
+
```
lL1l1: This is unused.
```suggestion
```

lL1l1 (changes requested):
I familiarized myself with workflows and brew wiki gen just now, so maybe it's not the best but I took a look.

.github/workflows/bake-blueprints.yaml
Garanas: Great observation, I'll look into it!

.github/workflows/bake-blueprints.yaml
Garanas: It should be used though! I'm just not sure if this is the correct direction to take.

In my opinion, I would always prefer to do the baking as part of a separate pull request and/or commit (directly to `develop`). That way you can reason about what to bake and have a small number of file changes. Where as the actual baking will always involve hundreds of files, slowing or even breaking down the preview on GitHub.

The idea of this input would be to use the latest version of the workflow file, while baking what is on the given branch.

.github/workflows/bake-blueprints.yaml
Garanas: Fixed! I've also adjusted it for Total Mayhem: https://github.com/FAForever/fa-total-mayhem/pull/52

Garanas:
@lL1l1 I've spent an hour or two. It now creates a pull request if you trigger the workflow manually. this allows you to review it. It creates a commit if you trigger it from another workflow, which would be done when deploying.

As an example of a pull request: 

- https://github.com/Garanas/fa/pull/4

edit: there are small changes in here to test the workflow. I'll remove those when we're ready for a merge.

.github/workflows/bake-blueprints.yaml
Garanas: Applied in: https://github.com/FAForever/fa/pull/6752/commits/18e8b4e1475f9b52a9dfe205fa2abc7780665764

lL1l1: I think I understand now: If the workflow ever needs an update it would be unrelated to what happens in the baking for blueprints, which is going to be based on whatever `BakeXBlueprints` does on the chosen branch, so it's nice to be able to separate those two and not have to wait for updates on the staging branches if something is wrong with the workflow.

The input's description needs to be updated.

Garanas: Done in https://github.com/FAForever/fa/pull/6752/commits/3eb65650bd6da512e5c5606c08137cb4532dcb2c

Garanas:
~~Adjusted the test workflow to only test the syntax of files that we ship. It now tests 6863 files instead of 7030 files. The largest difference (78 Lua files) is the engine folder. The other remaining 89 Lua files are scattered among the other folders that we do not ship.~~

All of the above is nonsense - I forgot there are layout files in the textures folder. The syntax check script also ignores the files in the engine folder by default. The difference is now 7, which makes more sense 😄 !

lL1l1 (approved):
Just have to remove the testing changes in `blueprints-props.lua` and this can be merged.

Also the commit description should note that it fixes the sparse checkout for the lua syntax test as well.

PR #6753, opened 2025-04-27
Fix various anti-air weapons being unable to engage units that fly right above them

Basilisk3:
As per the title. As far as I know, this is another issue that has been present since the game's release.

The bug is easiest to reproduce with the Burst Master (Cybran Tech 2 stationary flak) by enabling `NoDamage` in the console and hovering air scouts directly over it. There are some exceptions, but generally, the weapon's `TurretPitch` and `TurretPitchRange` should sum up to 90 (degrees).

Previously affected units:

```
Archer: T1 Mobile Anti-Air Gun (UEL0104)
Sky Slammer: T1 Mobile Anti-Air Gun (URL0104)
Cougar: T3 Mobile Rapid-fire AA Cannon (DELK002)
Thunderhead Class: T1 Frigate (UES0103)
Trident Class: T1 Frigate (URS0103)
Air Cleaner: T2 Anti-Air Flak Artillery (UEB2204)
Burst Master: T2 Anti-Air Flak Artillery (URB2204)
Marr: T2 Anti-Air Flak Artillery (UAB2204)
Sinnatha: T2 Anti-Air Flak Artillery (XSB2204)
```

I decided to add the balance labels, as while it is mostly a bug fix, the changes do have balance implications.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6753.md
+++ b/changelog/snippets/balance.6753.md
@@ -4,7 +4,6 @@
   
   <details><summary>Previously affected units:</summary>
 
-      ```
       Archer: T1 Mobile Anti-Air Gun (UEL0104)
       Sky Slammer: T1 Mobile Anti-Air Gun (URL0104)
       Cougar: T3 Mobile Rapid-fire AA Cannon (DELK002)
@@ -14,5 +13,5 @@
       Burst Master: T2 Anti-Air Flak Artillery (URB2204)
       Marr: T2 Anti-Air Flak Artillery (UAB2204)
       Sinnatha: T2 Anti-Air Flak Artillery (XSB2204)
-      ```
-    </details>
+
+  </details>
--- a/changelog/snippets/balance.6753.md
+++ b/changelog/snippets/balance.6753.md
@@ -1,0 +1,18 @@
+- (#6753) Fix various anti-air weapons being unable to engage units that fly right above them.
+
+  This bug was particularly significant for the Burst Master (Cybran Tech 2 Flak).
+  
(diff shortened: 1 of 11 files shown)


Garanas:
I don't agree with calling these changes a bug fix. There's no unintentional error or defect here - the unit just stops firing. The original values feel like a design decision to me. Therefore, in my opinion these, changes are just a balance change. 

The biggest issue I've found with these units is that the weapon of the Burstmaster has no recoil when firing. It looks silly now that I am aware of that 😄 !

Basilisk3:
It is entirely possible that this was a design decision, though I personally believe it was not. It is problematic because:

- It is not indicated by the UI, nor the rest of the blueprint in the form of a minimum range. There are units with similar angle limits, such as the Wailer, but the Wailer also has a minimum range.
- There is no clear pattern to it. Why does the Archer have it but the Sera MAA does not? I cannot think of any gameplay, faction diversity or balance reasons. Same goes for the frigates.

The lack of a minimum range also causes the AAs to stay locked onto units they cannot reach, even though alternative targets are available. I would classify this behavior as buggy, and it's not really desirable. 
targeting

We can either introduce a minimum range for all of these units or add a couple of degrees to the pitch range. I prefer option two.

Garanas:
Was checked out on the wrong branch, apologies!

lL1l1 (approved):
There were complaints in the discord discussion that it looked silly when the weapons aim at that angle, but in my opinion their angle limits before were still high enough to look silly, and the issue is with inherent to using heading/pitch while aiming at something above.

Issue #6075, opened 2024-04-12
Sera transport cannot land on narrow platoes


K-ETFreeman:
**Describe the bug**
Sera transport is being unusable on narrow platoes. This is really bad cuz it affects balance, on maps with such platoes sera will have disadvantage. Transports of other factions seems to have no such issues, or atleast far less. Atleast, on this particular plato in the video, any other t1 transport landed each time i tried, only sera is cursed.
Better to check out more narrow platoes and after fix make sure that any transport can land there regardless of faction

https://github.com/FAForever/fa/assets/55841348/743e7d4a-9e07-4c37-b803-6781bd6ed0ed

**Additional context**
replay id is https://replay.faforever.com/22266592 , but you can try it yourself

Garanas:
This is not a game/balance issue. It is a map issue. Not every unit fits everywhere. I think it's better to ask the author of the map to increase the size of the plateau.

K-ETFreeman:
> This is not a game/balance issue. It is a map issue. Not every unit fits everywhere. I think it's better to ask the author of the map to increase the size of the plateau.

Not sure what you meant
I mean, the platoe is large enough for transport to land. It lands sometimes, sometimes dont. If it just was too big to fit there, transport would never land, but it lands. So, there is no size issue here?
Also we have more narrow platoes on the same map, and yes, its still possible to land there, just sometimes it doesnt work for some reason

K-ETFreeman:
I just again faced this issue in ladder, really sucks
So, why it is labeled unfixable?
Ok lets assume that this is some engine behavior, or smth
Cant we just make some new unit which has uef/aeon transport as base, but replace the model with sera transport, or something like that?
So, engine-wise it will work more like uef/aeon transport (no big issues with langing), but visually will be somewhat the same?
Add missing transport slots, then replace original bugged sera transport with new fake-sera which does the same but without this issue
Should work?

K-ETFreeman:
i've got some confirmation from @Basilisk3 that this should be technically possible to fix with new-unit-to-replace-old approach, so i ask @MrRowey to remove type: unfixable from this issue

BlackYps:
I would like someone to explain or investigate why exactly the sera transport has more issues than the other transports. My first two ideas are the number of loaded units and the physical arrangement of the transport clamps. I don't think the uef transport and the sera transport use different classes in lua, but I am not sure about that. The root of the issue affects what solutions are viable, so we need to figure this out first.

[...]

Issue #2037, opened 2017-05-27
most transports doesn't move enough in order for unit to build


shalkya:
i didn't test if that is the case of other units.
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=6271632

sometimes it moves well enough, sometimes it doesn't move enough. happen 2 times in this replay.

shalkya:
been able to replicate that on sera and cybran T1 transport
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=6271690 from 4:50 to 8 min

shalkya:
on cybran and sera T2 transport too (didn't test aeon T2 transport)
http://content.faforever.com/faf/vault/replay_vault/replay.php?id=6271737

PR #6755, opened 2025-05-02
Revert of increased sera transport footprints

K-ETFreeman:
So, this is quick fix of #6075 . Whole issue was due to big sera transports footprint sizes, which were increased a year ago in #5885 . This PR basically reverts changes of #5885 for sera transports, so #2037 should be reopened again and fixed somehow else.
Blocking construction of the building is much more minor problem than transport not being able to drop units where other races can

I also revert t1 transport sizes to 1/3 instead of 1/4, because 1/3 behaves noticeably  better when dropping on small platoes, which is very common usage of t1 transport.

t2 footprint size shrinked to 3/3 also for better landing on small platoes.

Possible improvements for later: model changes, make the model smaller / shift it relative to the footprint for better transport behaviour.  So, this is not full solution for #6075, because it causes some minor issues, but this is still a major improvement. Full proper fix (probably) requires some model changes, adjusting the model so unit drop zone is about the same as footprint in both size and position.


Basilisk3:
It is unfortunate that this causes the bug detailed in #2037 to resurface, but I will agree that being able to drop plateaus is generally more important.

changelog/snippets/fix.6755.md
```diff
@@ -0,0 +1 @@
+- (#6755) Fix seraphim t1 and t2 transports being unable to drop units on small platoes
```
lL1l1:
```suggestion
- (#6755) Fix seraphim t1 and t2 transports being unable to drop units on small plateaus.
```

lL1l1 (approved):

PR #6763, opened 2025-05-04
Add a game option for shadow LOD

Garanas:

Introduces a game option to adjust the LOD for shadows. This is introduced because of an observation of @Strogoo at https://github.com/FAForever/FA-Binary-Patches/pull/115.

Observed that the shadow rendering distance is adjusted (live) as you change the settings.

## Additional context

The default value is 250 in Steam. The options default to 260. The default value in FAForever was 20.000 the past two years, see also https://github.com/FAForever/fa/issues/5093. 

## Checklist


Garanas:
@BlackYps It may be nice to include this

BlackYps:
If this isn't a bugfix (and it doesn't look like it is) then I would rather not include last minute changes in the patch

PR #6766, opened 2025-05-04
Refactor match state by applying the factory pattern to reduce complexity

Garanas:

With the introduction of decapitation in https://github.com/FAForever/fa/issues/6667 the current setup for determining the victory condition gets a bit [complected](https://youtu.be/SxdOUGdseq4?si=7sMI1igw3CGiuDWo&t=2544). By applying a factory pattern we can reduce the state to a minimum and create more meaning and order.

## Checklist


lL1l1 (commented):
I think there should be a class that handles the logic of everything except decapitation, since only self.UnitCategories is changed in those.

Garanas:
@lL1l1  Applied it in https://github.com/FAForever/fa/pull/6766/commits/7a3a88d6f88d95f6e2a4332e47a9413c72e99c8e . I understand where you come from, it's a balance:

- By introducing a single abstraction for multiple conditions we run the risk of a [leaky abstraction](https://en.wikipedia.org/wiki/Leaky_abstraction).
- By introducing a single abstraction we run the risk that it's not trivial to figure out what conditions it matches.
- By introducing a single abstraction, we move the responsibility of _what units_ trigger the condition to the abstract factory.

In this example - it is very, very unlikely that the behavior of Assasination, Supremacy and Annihilation would ever change. And therefore I think the argument of a leaky abstraction does not apply. In practice however - these things change!

lua/sim/Matchstate/AbstractVictoryCondition.lua
```diff
@@ -0,0 +1,230 @@
...
+        return false
+    end,
+
+    --- A utility function that retrieves all brains that are still participating in the match.
+    ---@param self AbstractVictoryCondition
+    ---@return AIBrain[]
+    GetArmyBrains = function(self)
+        local participatingArmyBrains = {}
+        for k, aiBrain in ArmyBrains do
+            if not (ArmyIsOutOfGame(aiBrain:GetArmyIndex() or aiBrain:IsDefeated())) then
```
lL1l1: - `ArmyIsOutOfGame` had its parentheses around the entire `or` statement
- `IsDefeated` already checks `ArmyIsOutOfGame`:
https://github.com/FAForever/fa/blob/e16b829ee2603450988f283e7ee1776026c53061/lua/aibrain.lua#L411-L417
```suggestion
            if not aiBrain:IsDefeated() then
```

lua/sim/Matchstate/AbstractVictoryCondition.lua
```diff
@@ -0,0 +1,230 @@
...
+            end
+        end
+
+        return false
+    end,
+
+    --- A utility function that retrieves all brains that are still participating in the match.
+    ---@param self AbstractVictoryCondition
+    ---@return AIBrain[]
+    GetArmyBrains = function(self)
```
lL1l1: The function name implies this gets all army brains. Maybe it should be something like `GetPlayingArmyBrains` or "Eligible" or "InPlay".

lua/sim/Matchstate/AbstractVictoryCondition.lua
```diff
@@ -0,0 +1,230 @@
...
+    --- Ends the game by starting a thread that ends the game 3 seconds later.
+    ---@param self AbstractVictoryCondition
+    EndGame = function(self)
+        self.Trash:Add(ForkThread(self.EndGameThread, self))
+    end,
+
+    --- Ends the game.
+    ---@param self AbstractVictoryCondition
+    EndGameThread = function(self)
+        WaitTicks(30)
```
lL1l1: Why not `WaitSeconds` to match the comment above?

lua/sim/Matchstate/AbstractVictoryCondition.lua
```diff
@@ -0,0 +1,230 @@
...
+
+        for _, v in GameOverListeners do
+            pcall(v)
+        end
+
+        Sync.GameEnded = true
+        WaitTicks(1)
```
lL1l1: Do we know the purpose of this wait?

Also previously it was `WaitSeconds(0.1)` which is `WaitTicks(2)`?

lua/sim/Matchstate/AbstractVictoryCondition.lua
```diff
@@ -0,0 +1,230 @@
...
+        EndGame()
+    end,
+
+    --- Utility function for the logic to remove an army from the game by turning it into an observer.
+    ---@param self AbstractVictoryCondition
+    ---@param aiBrain AIBrain
+    ToObserver = function(self, aiBrain)
+        local brainIndex = aiBrain:GetArmyIndex()
```
lL1l1:
```suggestion
        local brainIndex = aiBrain.Army
```
(and for all other `aiBrain:GetArmyIndex()` calls too)

lua/sim/Matchstate/AbstractVictoryCondition.lua
```diff
@@ -0,0 +1,230 @@
...
+
+        Sync.GameEnded = true
+        WaitTicks(1)
+        EndGame()
+    end,
+
+    --- Utility function for the logic to remove an army from the game by turning it into an observer.
+    ---@param self AbstractVictoryCondition
+    ---@param aiBrain AIBrain
+    ToObserver = function(self, aiBrain)
```
lL1l1: I no longer see the `SetCommandSource` part, where humans lose permission to do anything, does that matter?

lua/sim/Matchstate/UnitCondition.lua
```diff
@@ -0,0 +1,91 @@
...
+
+local AbstractVictoryCondition = import("/lua/sim/Matchstate/AbstractVictoryCondition.lua").AbstractVictoryCondition
+
+-- upvalue for performance
+local TableGetn = table.getn
+local TableInsert = table.insert
+
+---@class DecapitationCondition : AbstractVictoryCondition
+---@field UnitCategories EntityCategory
+DecapitationCondition = Class(AbstractVictoryCondition) {
```
lL1l1:
```suggestion
UnitCondition = Class(AbstractVictoryCondition) {
```

lua/sim/Matchstate/UnitCondition.lua
```diff
@@ -0,0 +1,91 @@
...
+            self:EndGame()
+            return
+        end
+    end,
+}
+
+---@param entityCategories EntityCategory
+---@return DecapitationCondition
+CreateUnitCondition = function(entityCategories)
+    return DecapitationCondition(entityCategories)
```
lL1l1:
```suggestion
    return UnitCondition(entityCategories)
```

lua/sim/Matchstate/UnitCondition.lua
```diff
@@ -0,0 +1,91 @@
...
+
+            self:EndGame()
+            return
+        end
+    end,
+}
+
+---@param entityCategories EntityCategory
+---@return DecapitationCondition
+CreateUnitCondition = function(entityCategories)
```
lL1l1: Instead of creating these functions you could add an `---@operator call(EntityCategory): DecapitationCondition` to the class annotation to make intellisense work, albeit manually (generics don't work in our outdated version).

lua/sim/Matchstate/DecapitationCondition.lua
```diff
@@ -0,0 +1,100 @@
...
+
+            if not self:BrainHasEligibleUnits(aiBrain, self.UnitCategories) then
+                TableInsert(defeatedBrains, aiBrain)
+            else
+                TableInsert(aliveBrains, aiBrain)
+            end
+        end
+
+        -- check if defeated brains still have allies with ACUs
+        local decapitatedBrains = {}
```
lL1l1: This table shouldn't be necessary, you can just remove brains from defeated brains when they have an ally. Using `table.getn` for iteration does make it annoying though, you have to do this trick:
```lua
local n = TableGetn(t)
for i = n, 1, -1 do
    local v = t[i]
    if cond() then
        t[i], t[n] = t[n], nil
        n = n - 1
    end
end
```
which is hard to read lol so maybe its easier to just iterate with a foreach and set keys to `nil`.

lua/sim/Matchstate/DecapitationCondition.lua
```diff
@@ -0,0 +1,100 @@
...
+        for k = 1, TableGetn(defeatedBrains) do
+            local defeatedBrain = defeatedBrains[k]
+
+            -- check if the defeated brain is allied with an army that is not defeated (yet!), if
+            -- so then the brain is not 'defeated' yet and should not be removed from the game
+            for l = 1, TableGetn(aliveBrains) do
+                local aliveBrain = aliveBrains[l]
+                if IsAlly(defeatedBrain:GetArmyIndex(), aliveBrain:GetArmyIndex()) then
+                    TableInsert(aliveBrains, defeatedBrain)
```
lL1l1: The decapitated brains table is not being updated in any way.

lua/sim/Matchstate/DecapitationCondition.lua
```diff
@@ -0,0 +1,100 @@
...
+--** SOFTWARE.
+--******************************************************************************************************
+
+local AbstractVictoryCondition = import("/lua/sim/Matchstate/AbstractVictoryCondition.lua").AbstractVictoryCondition
+
+-- upvalue for performance
+local TableGetn = table.getn
+local TableInsert = table.insert
+
+---@class DecapitationCondition : AbstractVictoryCondition
```
lL1l1: Class descriptions seem like a good idea.
```suggestion
--- Victory condition where brains are considered defeated if they have no allied ACUs alive.
---@class DecapitationCondition : AbstractVictoryCondition
```

lua/sim/Matchstate/DecapitationCondition.lua
lL1l1: Also if you want to aim for peak optimization you can track the `n` of tables as you insert into them from the very start, since they're not going to be edited outside this scope.

lua/sim/Matchstate/VictoryConditionFactory.lua
```diff
@@ -0,0 +1,43 @@
...
+    elseif victoryConditionType == 'domination' then
+        return import('/lua/sim/Matchstate/UnitCondition.lua').CreateUnitCondition(categories.STRUCTURE + categories.ENGINEER - categories.WALL)
+    elseif victoryConditionType == 'eradication' then
+        return import('/lua/sim/Matchstate/UnitCondition.lua').CreateUnitCondition(categories.ALLUNITS - categories.WALL)
+    elseif victoryConditionType == 'sandbox' then
+        return import('/lua/sim/Matchstate/SandboxCondition.lua').CreateSandBoxCondition()
+    end
+
+    -- default to sandbox
+    WARN("Unknown victory condition type: " .. victoryConditionType)
```
lL1l1: concatenation operator errors with non-strings.
```suggestion
    WARN("Unknown victory condition type: " .. tostring(victoryConditionType))
```

lua/sim/Matchstate/AbstractVictoryCondition.lua
```diff
@@ -0,0 +1,230 @@
...
+
+        -- only units that are finished are taken into account.
+
+        if unit:GetFractionComplete() <= 1.0 then
+            return false
+        end
+
+        -- only units that are not recalling are taken into account.
+
+        if unit.RecallingAfterDefeat then
```
lL1l1: I think this should be refactored to something like `IsNotEligibleForVictory` in the future.

lL1l1 (changes requested):
> By introducing a single abstraction we run the risk that it's not trivial to figure out what conditions it matches.

I think the 3 files with only units being different makes it more difficult than looking at what the Factory is doing.

> By introducing a single abstraction, we move the responsibility of what units trigger the condition to the abstract factory.

I'm not sure how this matters, whatever creates a victory instance can modify the units through class fields anyway since we don't implement private fields in lua (not that its possible with the debug library being accessible).

Speaking of "whatever creates a victory instance", I don't think the factory is necessary because it would only be called once due to its game-wide, engine-behavior effects. All of its code could just be in MatchState's `Setup` function.

Garanas:
> Speaking of "whatever creates a victory instance", I don't think the factory is necessary because it would only be called once due to its game-wide, engine-behavior effects. All of its code could just be in MatchState's `Setup` function.

I disagree because it dissolves the pattern. And by dissolving the pattern it makes it harder to reason about.

I've reworked it a bit. I think it's more clear now, and ready to be merged in. I also checked whether coop interact with these functions and they don't.

lua/sim/Matchstate/AbstractVictoryCondition.lua
Garanas: I have no idea. I'll leave it in for now.

lua/sim/Matchstate/AbstractVictoryCondition.lua
Garanas: Applied in https://github.com/FAForever/fa/pull/6766/commits/26a33782a23213715aab80300cf5c21be6642fdd

lua/sim/Matchstate/AbstractVictoryCondition.lua
Garanas: Applied in https://github.com/FAForever/fa/pull/6766/commits/1387401439bdd6bc0afc6e90088d3291e0bd1561

lua/sim/Matchstate/AbstractVictoryCondition.lua
Garanas: Re-introduced it with https://github.com/FAForever/fa/pull/6766/commits/506ac5ec22aed2eadc784605a0f61cbc19121d99

lua/sim/Matchstate/UnitCondition.lua
Garanas: Great find, applied it in https://github.com/FAForever/fa/pull/6766/commits/eb05d058399686fd388b5dc2ec8029754a5e6a86

lua/sim/Matchstate/DecapitationCondition.lua
Garanas: There's no performance concerns here. There's always only one instance. Readability is more favorable in my opinion.

PR #6771, opened 2025-05-07
Fix Aeon sniper aim while walking

lL1l1:
## Issue
Aeon sniper's walk animation pitches it up and down a lot, and the turret's pitch cannot keep up with this, so when walking about half the time the weapon is not pitched on target, and should not fire.
Keep in mind this is a pitch issue, so the issue appears when walking towards/away from the target.

The engine understands it should not fire, so in the test setup where it fires every tick the sniper does not always fire. Unfortunately the Lua places a muzzle charge delay on the weapon, which means the engine can fire when aimed but the delay creates the projectile when the weapon is no longer aimed, causing it to miss.

Increase the pitch speed 30 -> 50 so that the sniper is always aimed; now the delay can never create the projectile when the muzzle isn't aimed.

Use the test setup where the sniper fires every tick. Before the changes, the sniper would fire intermittently when moving towards/away from the target. After the changes, it fires every tick.

## Additional context
An alternative solution is to use the `RackSalvoChargeTime` value so that the weapon starts charging when it acquires the target, and only fires when it is exactly on target. This would also mean a sniper can charge up out of range and hold that shot forever though, which is a big difference visually so I didn't go with this solution.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6771.md
+++ b/changelog/snippets/balance.6771.md
@@ -1,0 +1,4 @@
+- (#6771) Fix the Aeon sniper missing when walking towards/away from the target due to insufficient turret pitch speed to compensate the walk animation.
+
+  **Sprite Striker: T3 Sniper Bot (XAL0305)**
+  - Turret pitch speed: 30 -> 50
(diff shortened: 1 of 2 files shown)


Rhaelya (approved):
Works as intended

PR #6784, opened 2025-05-08
Implement proper transfer of unbuilt units in factories/external factories

lL1l1:
Fixes an issue reported on [Discord](https://discord.com/channels/197033481883222026/1364371300474421370) where units in factories rebuilt by the unfinished unit transfer function were blocking the new factory.

- Disable "unfinished unit transfer" (the one that builds units on the ground) for units in factories.
- Instead, implement a factory rebuild function, which basically gives the factory cheats to rebuild the unit to the given % upon transfer (and then pauses). The function has compatibility with external factories.

Copy and execute the following command. It transfers selected units to a different army and then back to yourself so that you can see that everything works (factory rebuilds the unit to the correct %, the unit has correct HP if it was damaged in the fac, and the fac is paused (make sure to turn off ecomanager if you use it, as it doesn't handle sim-paused units so it unpauses them)).
```
ConExecute([[SimLua 
ForkThread(function ()
    local focus = GetFocusArmy()
    local f = import('/lua/SimUtils.lua').TransferUnitsOwnership
    local units = DebugGetSelection()
    units = f(units, focus + 1, false)
    WaitTicks(10)
    f(units, focus, false)
end)
]])
```

This command does the old behavior where units are rebuilt in-place. It should do nothing with the fix. Without the fix, it can cause various errors since the unit is destroyed while the factory isn't ready for it, but it shouldn't be an issue in real games as the original factory is also destroyed upon transfer (so it cannot error when the unit being built is destroyed).
```
ConExecute([[SimLua 
local f = import("/lua/SimUtils.lua").TransferUnfinishedUnitsAfterDeath
local focus = GetFocusArmy()
ForkThread(f, ArmyBrains[focus]:GetListOfUnits(categories.ALLUNITS, false, false), {focus})
]])
```

## Checklist


Garanas (commented):
Discussed it over voice together with @lL1l1 - changes look good and the behavior is functional. Three concerns:

- When a unit is 50% or higher then the wreck remains.
- When the receiving army has no income whatsoever, the rebuilt unit is not created.
- When the receiving brain is at unit cap, the rebuilt unit can not be created.

The first concern should be fixed, the second only happens in a sandbox environment. The third concern can be part of another pull request.

lL1l1:
For testing exfac vet dispersal: spawn an Aeon carrier (4k xp) and build Solaces (0-1650 xp). Groundfire the GC on the outside of the carrier to avoid killing the Solace under construction. You should see the GC gain more than 4k xp.

The same can be applied with land factories to test normal factories.

To test fatboy wreckage creation: spawn a fatboy and build a unit to >50% then ctrl-k the fatboy.

Garanas:
The issue that the receiving army needs to have _some_ type of income is still there. This won't happen often in practice, but it does 'break' the feature.

Garanas (approved):
I'm approving this because it fixes a bug where improperly rebuilt units occupy the build pad. And the game team would like to include that fix with the upcoming hotfix. The behavior is functional and the changes look good.

However - I feel the complexity of the solution we have here does not outweigh the problems that it is fixing. The problem (where units that are being rebuilt improperly block the build pads) is likely an unintended bug. And that bug is turned into a feature here. Introducing a wide variety of complexity to an already complicated transfer procedure!

It feels like to me that (properly) destroying the unit would have been a sufficient solution here. It mimics the behavior we've had for years. And the bug would have been fixed. And we would not have all of this complicated logic that has edge cases where it does not work. Even if those edge cases only happen in a sandbox scenario.

Similar for the veterancy dispersion - dispersing the veterancy of the unit being built (or of units in a transport, in general) feels random to me. The behavior for the cargo of transports already existed. This pull request introduces similar behavior for units that are being built. In practice it means a random unit - of which the player has no real control - will receive a random (and sometimes massive) veterancy buff. To what purpose, I ask? It introduces a lot of conditions into the code. It makes the code less maintainable. And in practice it has very little to no impact to gameplay. The unit is likely to just get destroyed a few seconds later anyway as veterancy only has a meaningful impact when it is a unit with a very high health pool (like the ACU, experimental units). And specifically for those type of units we introduced a hard limit to just vet a single level to make sure it does not get out of control!

BlackYps:
I tend to agree. Trying to preseve the build status in factories seems like a very complicated process for something that is already an edge case. It happens a times per game at maximum. And I don't think it's too severe that you lose some unit progress when a player dies. Yes, it can be punishing to lose an almost built battleship, but there is an obvious answer: just don't get killed if you don't want to bring your team to a disadvantage.

I don't have a strong opinion about the veterancy dispersion, but I do wonder if units that are only partly built should give veterancy upon destruction at all.

lL1l1:
>  just don't get killed

Sometimes you disconnect, which gives you no choice in not getting killed.

I'm 100% for anything that helps mitigate disconnects or fixes general gameplay issues when someone dies, because these things are a bit on the meta level (they don't make sense in-simulation) and often outside your control due to network issues, team balance, or just natural variation in strategies people play.

> And I don't think it's too severe that you lose some unit progress when a player dies.

This depends on the level of competition. The margins get thin at higher levels, so losing an early 90% complete unit, or a bunch of ~50% complete ones can become quite annoying because it has an impact you can feel.

Basilisk3:
> I'm 100% for anything that helps mitigate disconnects or fixes general gameplay issues when someone dies, because these things are a bit on the meta level (they don't make sense in-simulation) and often outside your control due to network issues, team balance, or just natural variation in strategies people play.

I agree with this. Additionally, unbuilt units disappearing also makes comebacks less likely to happen since the team that just lost a player gets an additional disadvantage (especially relevant for 2 vs. 2).

PR #6790, opened 2025-05-09
Fix cancelling tempest firing cycle

### LINKED ISSUE #6443: [Bug]: Moving the attack order while the Tempest charges its cannon cancels the shot but activates its reload time
### Describe the Bug

As per the title. Especially relevant when groundfiring but also when targeting units regularly.

### Reproduce the bug

Groundfire a Tempest and move the attack order while it is in the process of charging its cannon.

### Screenshots

_No response_

### Additional context

_No response_

lL1l1:
Resolves #6443
MuzzleChargeDelay is bad because it desyncs the projectile creation and the projectile firing from the engine when the weapon is aimed. We ignore the engine's behavior by instead basing the reload on rack salvo reload time, and setting the fire rate low. This allows us to cancel firing during the charging sequence without firing and without triggering the reload delay.

- The tempest can be retargeted without triggering a long reload now.
- The tempest doesn't charge its weapon against units outside of range and it cannot hold its charge (this mimics the target acquisition->firing behavior of MuzzleChargeDelay, it is what "fires after charge" is for.)

The tempest will charge and fire when it *gets* a target instead of when it is *on* target, which can easily cause it to miss due to the turn speed. This is fixed with https://github.com/FAForever/fa/pull/6791.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6790.md
+++ b/changelog/snippets/balance.6790.md
@@ -1,0 +1,1 @@
+- (#6790) The tempest can now be retargeted while it is charging without triggering a full 12.5 second reload.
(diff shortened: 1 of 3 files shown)


Basilisk3 (approved):
Works in conjunction with #6791.

PR #6794, opened 2025-05-12
Change Aeon ACU's personal shield from blue to green

lL1l1:
- Changes the shader so that it matches other Aeon units.

Old|New
-|-
{82C05C8E-66C0-4C6E-ABE5-4BD55E0EA8FB} | {CA556765-33DA-40BB-89BD-CF87E6883CDC}

## Additional Context

This change was originally made on gapforever https://github.com/gapforever2/fa/commit/5709d4aee66f261f856bbb95ea7e9d4cfc39eb1b.

## Checklist
~~- [ ] Changes are annotated, including comments where useful~~ (we don't have an aesthetics guideline afaik)

### DIFF
--- a/changelog/snippets/graphics.6794.md
+++ b/changelog/snippets/graphics.6794.md
@@ -1,0 +1,1 @@
+- (#6794) Change Aeon ACU's personal shield from UEF's blue shield to Aeon's green shield effect.
--- a/units/UAL0001/UAL0001_PhaseShield_mesh.bp
+++ b/units/UAL0001/UAL0001_PhaseShield_mesh.bp
@@ -4,7 +4,7 @@ MeshBlueprint{
         {
             LODCutoff = 195,
             LookupName = "/effects/entities/PhaseShield/PhaseShieldLookup.dds",
-            ShaderName = "PhaseShield",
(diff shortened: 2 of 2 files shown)


Garanas (approved):
Looks like a great change to me!

lL1l1:
I got a comment that the Aeon shields are hard to see so I made a comparison. Bloom off:
 
https://github.com/user-attachments/assets/f7b8a992-c98d-4039-b180-f5bec6ef604a

Bloom with default 15 intensity:

https://github.com/user-attachments/assets/32be466f-b4ce-4282-8504-571b3dff4359

Minimum 10 intensity looks much more like bloom off.

RegalEagIe:
What about the seraphim SCU personal shield?

lL1l1:
It would need an original new design. The shader already exists and is used by the SACU: https://github.com/FAForever/fa/blob/12c3305bf0f6023581fdaaf319fba50e9e0a135d/effects/mesh.fx#L4163

but it's not much different from the UEF shield. Also the SACU rebalance was considering removing the personal shield anyway so maybe it's not worth the effort (although removing it might mess with map script backwards compat).

PR #6802, opened 2025-05-17
Fix a bug where recall does not work

Garanas:

Fix a typo from https://github.com/FAForever/fa/pull/5971


lL1l1 (changes requested):
Specifically typo is here:
https://github.com/FAForever/fa/pull/5971/files#diff-514781ee9a4283ed2420eb74be449318d9f307f381cb34a11f563b69d5f2fe01L759-R579

That diff doesn't have recall.lua OnArmyDefeat (since that handles updating the state of the current recall vote in case an army is defeated mid-vote), so that should be removed.

lL1l1 (approved):

PR #6803, opened 2025-05-18
Map Textures Wasteland Biome - Reduced Size

sting-2:
Adds a few layer strata and normal decals as a wasteland biome. 

@BlackYps  This commit brings the size of env.nx2 from about 413 mb to about 489 mb. 

image

image


BlackYps (approved):
Looks good. This time we should not run into cloudflare delivery problems

PR #6813, opened 2025-05-22
Re-implement announcements

### LINKED ISSUE #6812: [Bug]: Campaign
### Describe the Bug

After latest patch a lot missions do not work properly. Examples: objective in last aeon mission show in weird place and mouse interact with objective icon in place where there is no objective icon. FA 1st mission. Cutscene does not end

### Reproduce the bug

Run the missions

### Screenshots

_No response_

### Additional context

_No response_

Garanas:

Implements how announcements work from the ground up! There was an error introduced with https://github.com/FAForever/fa/pull/5971 but I could not figure out where the circular dependency was. The existing code was quite like spaghetti and after half an hour I figured - let's just rework it.

The new setup produces an animation that works exactly like the old one. It's now also easier to introduce different type of announcements. 

Introduced a debug hotkey that you can use to trigger the announcement. Played campaign maps to make sure objectives work properly again.

## Additional context

The way announcements work is very interesting! To help the player understand to what the announcement is related it 'morphs' from a control to the center of the screen when expanding, just to morph back again when it's contracting. The content does not morph - it just becomes opaque to become transparent again by adjusting the alpha.

As an example of how the old animation looks like:

https://github.com/user-attachments/assets/088bea20-3d5a-4cd6-b8b1-f8b90a552586

And it starts like this:

image

image

image

Up to the point where it encapsulates the content of the announcement. Once it arrived at its destination the content shows up by reducing the transparency.

To make this easier to work with I introduced a `ContentArea` field which is where the announcement will morph to. Where ever that is, and of whatever size it is - when morphing the announcement will wrap itself around it. This makes it much more convenient to create separate types of announcement that involve more then just text! 

## Checklist

### DIFF
--- a/lua/ui/game/announcement.lua
+++ b/lua/ui/game/announcement.lua
@@ -132,7 +132,8 @@ end
 ---@param bodyText? UnlocalizedString
 ---@param goalControl? Control          # if defined, the announcement visually expands and contracts to this control.
 function CreateAnnouncement(titleText, bodyText, goalControl)
-    if type(bodyText) == "string" then
+    local typeOfBodyText = type(bodyText)
+    if typeOfBodyText == "string" or typeOfBodyText == "number" then
         return import("/lua/ui/game/announcement.lua").CreateTitleTextAnnouncement (titleText, bodyText, goalControl)
     else
         return import("/lua/ui/game/announcement.lua").CreateTitleAnnouncement(titleText, goalControl)
--- a/lua/ui/game/announcement.lua
+++ b/lua/ui/game/announcement.lua
@@ -83,11 +83,11 @@ CreateTitleAnnouncement = function(titleText, goalControl)
     goalControl = goalControl or CreateDefaultGoalControl(frame)
 
     -- developers note: lazy load the module so that it remains unloaded unless used
-    local SmallAnnouncement = import("/lua/ui/game/announcement/SmallAnnouncement.lua").SmallAnnouncement
+    local TitleAnnouncement = import("/lua/ui/game/announcement/TitleAnnouncement.lua").TitleAnnouncement
 
     -- create the announcement
     ---@type UIAbstractAnnouncement
-    local announcement = SmallAnnouncement(frame, titleText)
+    local announcement = TitleAnnouncement(frame, titleText)
     announcement:Animate(goalControl, 1.4)
     Announcements:Add(announcement)
 
--- a/lua/ui/game/announcement.lua
+++ b/lua/ui/game/announcement.lua
@@ -5,19 +5,24 @@
 --*
 --* Copyright � 2007 Gas Powered Games, Inc.  All rights reserved.
 --*****************************************************************************
-local LayoutHelpers = import("/lua/maui/layouthelpers.lua")
-local Layouter = LayoutHelpers.ReusedLayoutFor
-
-local UIUtil = import("/lua/ui/uiutil.lua")
 local Group = import("/lua/maui/group.lua").Group
-local Bitmap = import("/lua/maui/bitmap.lua").Bitmap
-local TextArea = import("/lua/ui/controls/textarea.lua").TextArea
 
 local SmallAnnouncement = import("/lua/ui/game/announcement/SmallAnnouncement.lua").SmallAnnouncement
 
-local MATH_Lerp = MATH_Lerp
+---@type TrashBag | UIAbstractAnnouncement[]
+local Announcements = TrashBag()
 
-local bg = false
+--- Creates the default goal control that originates from the top of the screen, in the center.
+---@param frame Frame
+local function CreateDefaultGoalControl(frame)
+    goalControl = Group(frame)
+    goalControl.Left:Set(function() return frame.Left() + 0.49 * frame.Right() end)
+    goalControl.Right:Set(function() return frame.Left() + 0.51 * frame.Right() end)
+    goalControl.Top:Set(function() return frame.Top() end);
+    goalControl.Height:Set(0)
+
+    return goalControl
+end
 
 --- Create an announcement UI for sending general messages to the user
 ---@param text UnlocalizedString # title text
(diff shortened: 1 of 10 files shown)


Garanas:
An example of the 'new' animation:

https://github.com/user-attachments/assets/49dd3823-ae18-4595-8633-708e0d22df07

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+---@field BackgroundBorderTopRight Bitmap       # Border element of the background
+---@field BackgroundBorderLeft Bitmap           # Border element of the background
+---@field BackgroundBorderRight Bitmap          # Border element of the background
+---@field BackgroundBorderBottomLeft Bitmap     # Border element of the background
+---@field BackgroundBorderBottom Bitmap         # Border element of the background
+---@field BackgroundBorderBottomRight Bitmap    # Border element of the background
+---@field ContentArea Group                     # Content area that we animate to/from
+AbstractAnnouncement = ClassUI(Group) {
+
+    -- Announcements works by morphing the background from a control, to the content are and back. By doing so,
```
lL1l1:
```suggestion
    -- Announcements works by morphing the background from a control, to the content area, and then back. By doing so,
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+    -- we give the player an idea what the announcement is connected to. The content of the announcement is never 
+    -- moved, only the background is. The alpha of the content is adjusted when the announcement arrives and 
+    -- leaves again.
+
+    -- To make it more concrete:
+    -- - The 'ContentArea' is the area that we animate to/from
```
lL1l1: to/from is easy to confuse as the origin control, which would be very similarly worded with from/to.
```suggestion
    -- - The 'ContentArea' is the area that we animate towards when displaying the announcement, and away from when hiding it
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+    --- The default animation sequence for an announcement. The announcement is destroyed at the end of the animation.
+    ---@param self UIAbstractAnnouncement
+    ---@param control Control
+    AnimateThread = function(self, control)
+        -- internal configuration of the animation
+        local expandDuration = 0.4
+        local contractDuration = 0.4
+        local contentDuration = 2.0
+        local contentShowDuration = 0.6
+        local contentHideDuration = 0.4
```
lL1l1: If someone wishes to change this then it would duplicate the animation code.
You can say that the duration of movement should stay consistent, but how long the content is shown is important if more information needs to be displayed. On the other hand that can simply be considered misusing announcements and instead the transmission log should be used, what do you think?

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+        local from = self.ContentArea:GetAlpha()
+        self.AnimateContentThreadInstance = self.Trash:Add(ForkThread(self.AnimateContentThread, self, duration, from, targetTransparency))
+    end,
+
+    --- Animates the alpha value of the content area.
+    ---@param self UIAbstractAnnouncement
+    ---@param duration number
+    ---@param target number
+    AnimateContentThread = function(self, duration, from, target)
+        -- animate it
```
lL1l1:
```suggestion
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+            if currentTime > endTime then
+                break
+            end
+
+            local progress = (currentTime - startTime) / duration
+            local alpha = math.clamp(MATH_Lerp(progress, from, target), 0, 1)
+            self.ContentArea:SetAlpha(alpha, true)
+
+            WaitFrames(1)
+        end
```
lL1l1: Not a suggestion, but this makes me think of OnFrame and using States to change what OnFrame does. I don't think I've ever seen States used in UI. It isn't necessary in this case because this is a simple UI progressing linearly. I don't think the complexity of states should ever be necessary for the UI though, it would mean your UI element is unapproachable to the user and also there are no engine calls for the UI you really need to have states for.

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+    --- Contracts the background of the announcement, the provided control is the destination to contract to.
+    ---@param self UIAbstractAnnouncement
+    ---@param control Control
+    ---@param duration number
+    ContractBackground = function(self, control, duration)
+        self:AnimateBackground(control, duration, 'Contract')
+        PlaySound(Sound({Bank = 'Interface', Cue = 'UI_Announcement_Close'}))
+    end,
+
+    --- Animates the background. Use the utility functions `ExpandBackground` and `ContractBackground` to
```
lL1l1, line 282: Incomplete comment

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+
+            local progress = (currentTime - startTime) / duration
+            local alpha = math.clamp(MATH_Lerp(progress, from, target), 0, 1)
+            self.ContentArea:SetAlpha(alpha, true)
+
+            WaitFrames(1)
+        end
+    end,
+
+    --- Expands the background of the announcement, starting at the provided control towards the center of the screen.
```
lL1l1: Should be changed in other comments too.
```suggestion
    --- Expands the background of the announcement, starting at the provided control towards the content area.
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+
+    --- Cancels the animation thread.
+    ---@param self UIAbstractAnnouncement
+    CancelAnimation = function(self)
+        if self.AnimateThreadInstance then
+            self.AnimateThreadInstance:Destroy()
+        end
+    end,
+
+    --- Cancels the animation of the background, stopping it in its tracks.
```
lL1l1, line 441:
```suggestion
    --- Cancels the animation of the background.
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+    ---@param self UIAbstractAnnouncement
+    CancelBackgroundAnimation = function(self)
+        if self.AnimateBackgroundThreadInstance then
+            self.AnimateBackgroundThreadInstance:Destroy()
+        end
+    end,
+
+    --- Cancels the animation of the content, stopping it in its tracks.
```
lL1l1:
```suggestion
    --- Cancels the animation of the content.
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+                break
+            end
+
+            WaitFrames(1)
+        end
+
+        self:Destroy()
+    end,
+
+    --- Cancels the animation thread.
```
lL1l1: I would call it main because it executes both the background and content animations.
```suggestion
    --- Cancels the main animation thread.
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+
+    --- Aborts the announcement by turning it completely transparent. The announcement is destroyed at the end of it.
+    ---@param self UIAbstractAnnouncement
+    AbortAnimationThread = function(self)
+        local alphaContent = self.ContentArea:GetAlpha()
+        local alphaBackground = self.Background:GetAlpha()
+
+        while not IsDestroyed(self) do
+            alphaContent = math.clamp(alphaContent - 0.05, 0, 1)
+            alphaBackground = math.clamp(alphaBackground - 0.05, 0, 1)
```
lL1l1: This should be proportional to system time, not framerate.

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+        self:CancelAnimation()
+        self:CancelContentAnimation()
+
+        self.AbortAnimationThreadInstance = self.Trash:Add(ForkThread(self.AbortAnimationThread, self))
+    end,
+
+    --- Aborts the announcement by turning it completely transparent. The announcement is destroyed at the end of it.
+    ---@param self UIAbstractAnnouncement
+    AbortAnimationThread = function(self)
+        local alphaContent = self.ContentArea:GetAlpha()
```
lL1l1, line 408: When I use the dummy announcement hotkey I get this error:
```
WARNING: Error running lua script: ...fa\lua\ui\game\announcement\abstractannouncement.lua(376): Game object has been destroyed
         stack traceback:
         	[C]: in function `GetAlpha'
         	...fa\lua\ui\game\announcement\abstractannouncement.lua(376): in function <...fa\lua\ui\game\announcement\abstractannouncement.lua:375>
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,421 @@
...
+        local startTime = GetSystemTimeSeconds()
+        local endTime = startTime + duration
+        while not IsDestroyed(self) do
+            local currentTime = GetSystemTimeSeconds()
+            if currentTime > endTime then
+                break
+            end
+
+            -- compute where we are in the animation
+            local percentage = math.clamp((currentTime - startTime) / (endTime - startTime), 0, 1)
```
lL1l1, line 359: I think lerp and clamped lerp should be in lua utils. As usual, Lua is significantly faster than the engine function so replacing what I was using is good.

lL1l1 (changes requested):
Testing:

https://github.com/user-attachments/assets/583d85ea-acc9-4b01-b2ab-1e3ea4fc9578

- In my testing the new announcement is slower, which makes sense because the values are different. I think it being displayed for longer is nice but it should expand and hide faster.

- There is an issue with what happens when a new announcement is ordered. The new method fixes the bug where the text gets full alpha and goes to 0 even though its outside the background, but now the new announcement instantly disappears instead of fading out like the old one. This can cause you to miss an announcement, for example if multiple players get shared to you at a time.

Code style:
- The new announcement is quite heavy on line count and is annoying to read for me. I'm biased because I understand the old announcement code but it's been changed 214 lines -> 581 lines across multiple files. A lot of UI is simple and doesn't need this much space imo.

- There would need to be a larger announcement class to fit secondary text, I assume that's the next step after this initial review though.

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: It feels a little tricky. I think the expand/contract duration should always be the same out of consistencies sake. I'm open to parameterize the duration of the content. To me that is the only part that is relevant when you create a new type of announcement with more content in it.

The idea that I had in mind when writing this is that if people want to customize this then it's probably better to just overwrite it in your class that inherits from the abstract announcement class. There's no 'logic' in this function, just behavior. And it's fine to overwrite behavior.

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: I actually tried doing that 😃 ! To me it felt very restricting. Similarly, only using `OnFrame` also felt off.

And therefore I decided to just take this approach instead. You're right in that it's very similar, if not equal to the concept behind `OnFrame`. This setup is just more flexible - it's easier to cancel things.

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: You're right!

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: That's odd, it worked fine yesterday. I'll give it a view.

Garanas:
> * The new announcement is quite heavy on line count and is annoying to read for me. I'm biased because I understand the old announcement code but it's been changed 214 lines -> 581 lines across multiple files. A lot of UI is simple and doesn't need this much space imo.

You're right in that it's more lines of code. It's written in a way that you could technically also test it. All the behavior is in a separate function. And the actual animation is just a function that calls the behavior (other functions) in the 'right' order.

It feels to me that half of the lines are comments, annotation or just white space. I can be a bit verbose with those.

> * There would need to be a larger announcement class to fit secondary text, I assume that's the next step after this initial review though.

That would be the next step!

> * There is an issue with what happens when a new announcement is ordered. The new method fixes the bug where the text gets full alpha and goes to 0 even though its outside the background, but now the new announcement instantly disappears instead of fading out like the old one. This can cause you to miss an announcement, for example if multiple players get shared to you at a time.

I was thinking of this too while implementing it. I think the easiest solution is to start the next announcement when the current announcement is ready to contract. That way you can 'queue' them. What do you think?

Other options I thought of was to show multiple announcements at once, but to show them in a vertical order. That way you can get a column of announcements if multiple show up. I'm not sure about that though.

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,420 @@
...
+        self.Trash:Destroy()
+    end,
+
+    ---------------------------------------------------------------------------
+    --#region Interface to animate
+
+    --- The default animation sequence for an announcement. The announcement is destroyed at the end of the animation.
+    ---@param self UIAbstractAnnouncement
+    ---@param control Control
+    Animate = function(self, control)
```
4z0t: I'd consider using approach as I ended up in my UI mods. https://github.com/4z0t/FAF-UI-Mods/blob/main/mods%2FReUI.UI.Animation%2FMain.lua

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -0,0 +1,420 @@
...
+        end
+
+        self:AnimateContent(contentHideDuration, 0.0)
+        WaitSeconds(contentHideDuration)
+        if IsDestroyed(self) then
+            return
+        end
+
+        -- contract animation
+        self:ContractBackground(control, contractDuration)
```
4z0t, line 212: Same applies to any other animation.

lua/ui/game/announcement/AbstractAnnouncement.lua
lL1l1: It doesn't always happen, you might have to mess with it a bit.

lL1l1:
> I was thinking of this too while implementing it. I think the easiest solution is to start the next announcement when the current announcement is ready to contract. That way you can 'queue' them. What do you think?
> 
> Other options I thought of was to show multiple announcements at once, but to show them in a vertical order. That way you can get a column of announcements if multiple show up. I'm not sure about that though.

I think just queueing them is enough. You could combine the queuing with the column idea by having queued announcements stack up behind each other with a slight offset like a deck of cards. This would let you know when the announcement happened exactly, but still give you time to read it and it wouldn't take up much space or be too distracting while queued.

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: Processed in: https://github.com/FAForever/fa/pull/6813/commits/ad8415f09c9fba9b41723167036333adaa8287ab

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: I think it is this: https://github.com/FAForever/fa/pull/6813/commits/91fac83582e4ef49f7cd569af3370e4f1a3b2ca6

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: I don't understand, I still use `MATH_Lerp` in `AnimateContentThread`

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: Maybe in the next iteration, this is part of a hotfix!

4z0t:
The way animations are done must be done as standalone module. Right now I see hardcoded and rigid way of doing it with `ForkThread`. Although such approach may be viable in sim, in UI it bloats code making it hard to read and understand as well as maintain. I suggest including/implementing Animator and Animation classes to handle all effects and animations that may happen in UI.

Garanas:
@4z0t Introducing a module full of abstractions of 500+ lines of code that also has a dependency on other facets of your library is not exactly hotfix material.

The current implementation is fine. It's not perfect, but it's much more maintainable then it was before. It fixes the issue with coop (that I could not find in the original code). Together it's a step forward. And I encourage you to help us understand the module you're referring to and how it can make the code base better in a separate voice chat/issue/pull request.

lua/ui/game/announcement/AbstractAnnouncement.lua
Garanas: How about https://github.com/FAForever/fa/pull/6813/commits/652d4745e47785a93de637b5bdfdbed7135ac4cd

Garanas:
@lL1l1 The title-with-text announcement exists now too. The `TextArea` class is an odd one 😄 !

There's a separate hotkey to test it. I did not yet implement the ability to queue announcements, I think that was also not really supported by the original implementation and therefore it's fine to not include that in this implementation. This is still a 'hotfix'.

https://github.com/user-attachments/assets/6a7b158b-2182-42c0-81e8-4f2deed27907

I also did not see where the title-with-text announcement was used, do you know where?

lL1l1:
@Garanas 
> I did not yet implement the ability to queue announcements, I think that was also not really supported by the original implementation and therefore it's fine to not include that in this implementation. This is still a 'hotfix'.

That is fine.

> I also did not see where the title-with-text announcement was used, do you know where?

Originally announcements were composed of two Text controls, one for the large primary text and the other for the smaller secondary text. Examples include the pause announcement:
https://github.com/FAForever/fa/blob/a34ffb7af22e51c24ac613e65a9fb177e24562f3/lua/ui/game/tabs.lua#L854
And the original fullshare announcement:
https://github.com/FAForever/fa/blob/633f5b8cc2528c8bafc281ad7077014dc86a748a/lua/UserSync.lua#L163-L166

I wanted to add longer secondary text that could span multiple lines (see below), so I changed the entire secondary text to a TextArea just to fit that newline.
https://github.com/FAForever/fa/blob/e16b829ee2603450988f283e7ee1776026c53061/lua/UserSync.lua#L180-L188

lua/ui/game/announcement.lua
```diff
@@ -1,213 +1,146 @@
...
+end
 
 --- Create an announcement UI for sending general messages to the user
 ---@param text UnlocalizedString # title text
 ---@param goalControl? Control The control where the announcement appears out of.
----@param secondaryText? UnlocalizedString # body text
----@param onFinished? function
-function CreateAnnouncement(text, goalControl, secondaryText, onFinished)
-    local frame = GetFrame(0)
+function CreateAnnouncement(text, goalControl)
```
lL1l1: Removing the secondary text entirely breaks backwards compatibility:
https://github.com/FAForever/fa/blob/093737430c275bb3ebab1cd2376bccff1f8537c3/lua/ui/game/build_templates.lua#L43

https://github.com/FAForever/fa/blob/a34ffb7af22e51c24ac613e65a9fb177e24562f3/lua/ui/game/tabs.lua#L854

lua/ui/game/announcement/TitleTextAnnouncement.lua
```diff
@@ -0,0 +1,87 @@
...
+
+    ---@param self UITitleTextAnnouncement
+    ---@param parent Control
+    ---@param text string
+    __init = function(self, parent, text, bodyText)
+        AbstractAnnouncement.__init(self, parent)
+
+        self.Title = UIUtil.CreateText(self.ContentArea, text, 22, UIUtil.titleFont)
+        self.Text = TextArea(self.ContentArea, 600, 60)
+        self.Text:SetText(bodyText)
```
lL1l1:
```suggestion
        self.Text:SetText(LOC(bodyText))
```

lua/ui/game/announcement/TitleTextAnnouncement.lua
```diff
@@ -0,0 +1,87 @@
...
+
+--- An announcement with a title and some text below it.
+---@class UITitleTextAnnouncement : UIAbstractAnnouncement
+---@field Title Text
+---@field Text TextArea
+TitleTextAnnouncement = ClassUI(AbstractAnnouncement) {
+
+    ---@param self UITitleTextAnnouncement
+    ---@param parent Control
+    ---@param text string
```
lL1l1:
```suggestion
    ---@param text UnlocalizedString
    ---@param bodyText UnlocalizedString
```

lua/ui/game/announcement/TitleTextAnnouncement.lua
```diff
@@ -0,0 +1,87 @@
...
+        AbstractAnnouncement.__init(self, parent)
+
+        self.Title = UIUtil.CreateText(self.ContentArea, text, 22, UIUtil.titleFont)
+        self.Text = TextArea(self.ContentArea, 600, 60)
+        self.Text:SetText(bodyText)
+    end,
+
+    ---@param self UITitleTextAnnouncement
+    ---@param parent Control
+    ---@param text string
```
lL1l1:
```suggestion
    ---@param text UnlocalizedString # unused
    ---@param bodyText UnlocalizedString # unused
```
Rowey used to add `# unused` comments but I'm not sure if they're needed, what do you think? You can see if a variable is unused in the function code but not in the call signature preview from the extension, that's where the comment provides new information.

lua/ui/game/announcement/TitleTextAnnouncement.lua
```diff
@@ -0,0 +1,87 @@
...
+    __post_init = function(self, parent, text, bodyText)
+        AbstractAnnouncement.__post_init(self, parent)
+
+        Layouter(self.Title)
+            :AtCenterIn(parent, -250)
+            :End()
+
+        local text = self.Text
+        text:SetFont(UIUtil.bodyFont, 18)
```
lL1l1, line 60: Setting the font after setting the text clips the announcement created by this command:
```
primary ="<LOC usersync_0005>Disconnect Temporary Share"
secondary ="<LOC usersync_0007>%s\'s units and ACU transferred to you until ACU dies/recalls"
import("/lua/ui/uiutil.lua").CreateAnnouncementStd(primary, secondary, nil)
```
This is due to a bug in how TextArea reflows text after setting a font. Implemented in ItemList, SetFont doesn't actually change the font through dirtying the font lazyvars, but just calls the internal font change directly after both vars are dirtied. So the font vars are dirtied, the font doesn't change but the text is reflowed, and then the font changes, thus the bug.

https://github.com/FAForever/fa/blob/e16b829ee2603450988f283e7ee1776026c53061/lua/maui/itemlist.lua#L86-L94

lua/ui/game/announcement/AbstractAnnouncement.lua
lL1l1: I figured it out. It happens when the old announcement finishes and destroys itself, but the Announcements trashbag in announcements.lua still has a weak reference to destroyed announcement, so it tries to abort that old announcement when you create a new one.

This can be fixed in that specific place and then all the early exits can be removed.

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -373,14 +400,25 @@ AbstractAnnouncement = ClassUI(Group) {
...
+        end
+
         local alphaContent = self.ContentArea:GetAlpha()
         local alphaBackground = self.Background:GetAlpha()
 
+        local lastTime = GetSystemTimeSeconds()
         while not IsDestroyed(self) do
-            alphaContent = math.clamp(alphaContent - 0.05, 0, 1)
-            alphaBackground = math.clamp(alphaBackground - 0.05, 0, 1)
+            -- frame rate independent
```
lL1l1:
```suggestion
```

lua/ui/game/announcement/AbstractAnnouncement.lua
```diff
@@ -373,14 +400,25 @@ AbstractAnnouncement = ClassUI(Group) {
...
+            local currentTime = GetSystemTimeSeconds()
+            local diff = 2 * (currentTime - lastTime)
```
lL1l1:
```suggestion
            local diff = 2 * (currentTime - lastTime) -- 100% alpha change in 0.5 seconds
```

lua/ui/game/announcement/TitleTextAnnouncement.lua
lL1l1: Fixed in 8d7c129

lua/ui/game/announcement/TitleTextAnnouncement.lua
Garanas: I don't think it's necessary, especially in the __post_init or __init functions as you don't see those. They are called internally.

lua/ui/game/announcement.lua
Garanas: You're right, fixed in https://github.com/FAForever/fa/pull/6813/commits/6447ba3efa9957fdad6ab5e17e68de53269524d8

Garanas:
I think we're good here now. What do you think @lL1l1 ?

lua/ui/game/announcement.lua
```diff
@@ -106,12 +120,25 @@ CreateTitleTextAnnouncement = function(titleText, bodyText, goalControl)
...
 end
 
+--- A general function to create an announcement UI for sending general messages to the user.
+---
+--- Exists for backwards compatibility.
+---@param titleText UnlocalizedString
+---@param bodyText? UnlocalizedString
+---@param goalControl? Control          # if defined, the announcement visually expands and contracts to this control.
+function CreateAnnouncement(titleText, bodyText, goalControl)
+    if type(bodyText) == "string" then
```
lL1l1: This conflicts with `UnlocalizedString` which is `string | number`

lua/ui/game/announcement/SmallAnnouncement.lua
```diff
@@ -33,17 +33,17 @@ SmallAnnouncement = ClassUI(AbstractAnnouncement) {
...
     ---@param self UISmallAnnouncement
     ---@param parent Control
-    ---@param text string
-    __init = function(self, parent, text)
+    ---@param titleText UnlocalizedString
+    __init = function(self, parent, titleText)
         AbstractAnnouncement.__init(self, parent)
 
-        self.Title = UIUtil.CreateText(self.ContentArea, text, 22, UIUtil.titleFont)
+        self.Title = UIUtil.CreateText(self.ContentArea, LOC(titleText), 22, UIUtil.titleFont)
```
lL1l1: CreateText already localizes the text input, that's why I didn't add it here.

lL1l1 (approved):
Some minor changes left but it can be merged

lua/ui/game/announcement.lua
Garanas: Fixed with https://github.com/FAForever/fa/pull/6813/commits/9e0356601433340d0da2fcec436efd448ab4a735

PR #6814, opened 2025-05-23
Fix issues with transport speed reduction

lL1l1:
## Issue
- I heard Nomads transport starts moving backwards when loaded with huge amounts of units.
- The map "Survival Versus" places structures on transports (flying Paragon bombs), and I was told the map was broken due to this.
- Another map is Direct Strike, which places air units inside transports. Maybe that was the actual map I'm remembering.
- I also have this error from game 24916350, it seems to be due to a czar (uaa0310) so it's probably related to attaching aircraft.
```
warning: Error running OnTransportDetach script in Entity uaa0310 at 2e23b308: ...\gamedata\lua.nx2\lua\sim\units\airtransportunit.lua(83): attempt to perform arithmetic on field `TransportSpeedReduction' (a nil value)
         stack traceback:
         	...\gamedata\lua.nx2\lua\sim\units\airtransportunit.lua(83): in function `ReduceTransportSpeed'
         	...\gamedata\lua.nx2\lua\sim\units\airtransportunit.lua(103): in function <...\gamedata\lua.nx2\lua\sim\units\airtransportunit.lua:100>
```

- Give transports 1 speed reduction for mod compatibility, just like we did with naval and experimental units. In theory all units should have the stat now.
- Add a nil check just in case some unit doesn't have the stat (air unit or no bp.Physics table)
- Add a minimum transport speed of 30% the base speed.

The following UI code mimics the code Direct Strike uses, which is the scenario framework function. It needs multiple cargo because GetCargo doesn't see the very first unit attached. It uses Janus for the cargo since it is an air unit.
You can switch `dea0202` for `xab1401` to test with 2 Paragon structures attached.
You can adjust the max value in the loop to overload the transport with weight.
```lua
CreateUnitAtMouse('uea0107', GetFocusArmy() - 1, 0, 0, 0)
ForkThread(function ()
    WaitTicks(6)
    UISelectionByCategory('uea0107', false ,false, true, false)
    ConExecute([[SimLua 
    local t = {}
    for i = 1, 2 do
        t[i] = CreateUnitHPR('dea0202', GetFocusArmy(), 0, 0, 0, 0, 0, 0)
    end
    import('/lua/ScenarioFramework.lua').AttachUnitsToTransports(t, {SelectedUnit()})
    ]])
end)
```

## Checklist

### DIFF
--- a/lua/sim/units/AirTransportUnit.lua
+++ b/lua/sim/units/AirTransportUnit.lua
@@ -80,7 +80,9 @@ AirTransport = ClassUnit(AirUnit, BaseTransport) {
         local transportspeed = self.Blueprint.Air.MaxAirspeed
         local totalweight = 0
         for _, unit in self:GetCargo() do
-            totalweight = totalweight + unit.Blueprint.Physics.TransportSpeedReduction
+            local reduction = unit.Blueprint.Physics.TransportSpeedReduction
+            if not reduction then continue end
+            totalweight = totalweight + reduction
 	    end
         self:SetSpeedMult(1 - (totalweight / transportspeed))
     end,
--- a/lua/sim/units/AirTransportUnit.lua
+++ b/lua/sim/units/AirTransportUnit.lua
@@ -90,7 +90,7 @@ AirTransport = ClassUnit(AirUnit, BaseTransport) {
                 totalweight = maxWeight
                 break
             end
-	    end
+        end
         self:SetSpeedMult(1 - (totalweight / transportspeed))
     end,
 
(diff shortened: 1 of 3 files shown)


Garanas (approved):

lua/sim/units/AirTransportUnit.lua
```diff
@@ -78,9 +78,18 @@ AirTransport = ClassUnit(AirUnit, BaseTransport) {
...
-            totalweight = totalweight + unit.Blueprint.Physics.TransportSpeedReduction
+            local reduction = unit.Blueprint.Physics.TransportSpeedReduction
+            if not reduction then continue end
+            totalweight = totalweight + reduction
+
+            if totalweight > maxWeight then
+                totalweight = maxWeight
+                break
+            end
 	    end
```
Basilisk3:
```suggestion
	end
```

Basilisk3 (commented):
This was something I also wanted to do but did not get to it. Looks good to me.

lua/sim/units/AirTransportUnit.lua
lL1l1: Fixed in https://github.com/FAForever/fa/pull/6814/commits/bfc9e05c66df18a5a3a8273f48a793e693e94c8c

PR #6823, opened 2025-05-26
Proper type annotation for `GetCommandMode`

4z0t:
## Checklist

### DIFF
--- a/changelog/snippets/other.6823.md
+++ b/changelog/snippets/other.6823.md
@@ -1,0 +1,1 @@
+- (#6823) Proper type annotation for `GetCommandMode`.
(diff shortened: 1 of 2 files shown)


lL1l1 (approved):
Issue was that `[1]` was `CommandModeDataOrder` where `CommandModeDataOrder` is actually a sub type of `CommandModeData` which is already in `[2]`. Instead `[1]` should've been `CommandMode`.
Specifically, this function is the source that sets these types:
https://github.com/FAForever/fa/blob/f7892ccff7cfe822172aaaae1a24da141827d1b3/lua/ui/game/commandmode.lua#L176-L187

These changes fix the issue correctly.

PR #6824, opened 2025-05-27
Fix factories destroying finished built units

lL1l1:
## Issue
Caused by #6784
There was no check if the unit being built was finished, so finished units, particularly factory upgrades, were being destroyed because no new unit started building before the factory was destroyed and the variable is cached even after the unit finishes.

Check if the unit is finished.

Factory upgrades no longer Destroy themselves on upgrade completion.

## Checklist


lL1l1:
I also ran an AIx Adaptive 3v3 with navy/land and there were no unusual errors or behaviors.

PR #6366, opened 2024-07-24
Extend logging of various game objects

Garanas:
## Description of the proposed changes

Introduces various logging functions and a drawing function for units, weapons and projectiles. This should make it easier to debug things during testing.

## Testing done on the proposed changes

Add debug logic and spawn in and interact with units.

## Checklist

- [x] Add logic to the deployment workflows to automatically turn off debugging for `deploy/faf` and `deploy/fafbeta`
- [x] Changes are annotated, including comments where useful
- [ ] ~~Changes are documented in the changelog for the next game version~~

PR #6830, opened 2025-05-30
Fix desync when using debug draw

lL1l1:
## Issue 
Drawing seems harmless but it actually affects the RNG output of the sim, which can desync the game in serious ways. Debug draw draws only for the focus army/observers, which causes a desync.

Draw for all armies, but keep the focus army feature by changing the color to transparency.

1. Use the debug setup.
2. Start a local multiplayer test with 2 players.
3. Spawn some units on screen for both players.
4. Kill those units. You should see a white circle for the player focused on the unit army and nothing for the player not focused.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6830.md
+++ b/changelog/snippets/fix.6830.md
@@ -1,0 +1,1 @@
+- (#6830) Fix unit/projectile/weapon debug component logging and drawing causing desyncs.
--- a/lua/SimHooks.lua
+++ b/lua/SimHooks.lua
@@ -40,45 +40,6 @@ do
 end
 
 do
-
-    -- upvalue for performance
-    local Random = Random
-
-    local oldDrawCircle = _G.DrawCircle
-    _G.DrawCircle = function(position, diameter, color)
-
-        -- cause a desync when players during non-ai games try and call this function separate from other players
-        if not ScenarioInfo.GameHasAIs then
-            Random()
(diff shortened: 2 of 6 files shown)


4z0t (approved):
I approve this, but consider making it not copy-pasted 3 times.

Garanas:
This behavior is intentional. See also https://github.com/FAForever/fa/pull/3810. At the time, there were suspicions that these (engine) functions were used to cheat. 

If we feel this is unnecessary then it's better to remove the hooks in my opinion. That would also solve the problem. And it would keep the drawing more clean, limited to just the players that match the army of the unit.

lL1l1:
We don't have any sandboxing, so circumventing hooks is not too difficult if you already know how to load your own sim lua. I'd just remove them.

lL1l1:
The hooks also cause any logging to desync in replays, which makes the debugging module unusable, even though it can be useful for replays.

Garanas:
> We don't have any sandboxing, so circumventing hooks is not too difficult if you already know how to load your own sim lua. I'd just remove them.

This is more involved then you make it appear to be. 

It doesn't matter - I'm fine with removing them. They've served their purpose. And I agree that it's incompatible with the introduction of https://github.com/FAForever/fa/issues/6366.

Garanas (approved):

PR #6831, opened 2025-05-30
Fix draws, recalls, and disconnects resulting in unknown results or incorrect victories

lL1l1:
## Issue
The server relies on game results reported by the game according to what is described in https://github.com/FAForever/fa/pull/4016, but the refactor in #6766 broke those results.
- 1 defeat is never reported if the last players die at the same time and leave no living brains
- 2 there is only 1 tick given for draws, so incorrect victory results are often given due to a random delay in the default unit death thread.
- 3 disconnects and recalls are never reported because they evade monitoring by being set to a defeated status (ineligible for monitoring) before units are destroyed and checked.

Replay analysis on [Discord](https://discord.com/channels/197033481883222026/1376572524208459936) for the replays: 
- [24950680](https://replay.faforever.com/24950680): 1v1 draw, no results sent
- [24932027](https://replay.faforever.com/24932027): 1v1 draw, no results sent
- 24971738: 3v3 TMM game ending with unknown result due to recall
- 24940485: 4v4 Custom game ending with unknown result due to disconnect (and maybe most players leaving too early to give a victory result?)

Fixes the listed issues.
1: [Process defeated brains before ending the game](https://github.com/FAForever/fa/commit/7423b8b432b1df654ebac391d4e160ed555982f2)
2: [Account for the random death thread delay](https://github.com/FAForever/fa/commit/8ac5eb174d973b36c92cf9095806a8e211ae1310): Increases the draw window from 1 tick to 6 ticks.
3: [Add recall to the victory condition](https://github.com/FAForever/fa/commit/dcd8a1b12a229fb7eff02d5d81baa0e7c0d9a873) and [Sync defeat of disconnected players](https://github.com/FAForever/fa/commit/4b22f1db4c3590c16796c380fac3eb4a67a183a3)

1. Use the debug setup.
2. Start a local test with 2 players in a 1v1 with Assassination.
---
3. Teleport one ACU to the other and speed up the game until both ACUs explode.
4. The game should end with defeat results for both players.
---
3. Recall one player.
4. The game should with a defeat result for one player, and victory for the other player.
---
3. Disconnect one player.
4. The game should end with a defeat result for one player, and victory for the other player.

## Additional context
The draw window used to be [5 seconds](https://github.com/FAForever/fa/blob/5667965b205f1b6f30994de3163d37cdb2580e01/lua/victory.lua#L111) before #4016. After that it was 1 tick * number of living brains due to [this line](https://github.com/FAForever/fa/blob/e16b829ee2603450988f283e7ee1776026c53061/lua/sim/MatchState.lua#L65).

## Checklist

### DIFF
--- a/lua/sim/VictoryCondition/DecapitationCondition.lua
+++ b/lua/sim/VictoryCondition/DecapitationCondition.lua
@@ -74,6 +74,13 @@ DecapitationCondition = Class(AbstractVictoryCondition) {
             end
         end
 
+        -- process all defeated brains. At this stage, it is an entire team that is defeated at once
+        -- defeated brains must be processed before the game (potentially) ends to sync game results to be sent to server
+        for k = 1, TableGetn(defeatedBrains) do
+            local defeatedBrain = defeatedBrains[k]
+            self:DefeatForArmy(defeatedBrain)
+        end
+
         -- no remaining players, just end the game
         if table.empty(aliveBrains) then
             if self.EnabledSpewing then
@@ -82,12 +89,6 @@ DecapitationCondition = Class(AbstractVictoryCondition) {
 
             self:EndGame()
             return
-        end
-
-        -- process all defeated brains. At this stage, it is an entire team that is defeated at once
-        for k = 1, TableGetn(decapitatedBrains) do
-            local defeatedBrain = decapitatedBrains[k]
-            self:DefeatForArmy(defeatedBrain)
         end
 
         -- check if all remaining players want to forfeit
--- a/lua/sim/VictoryCondition/DecapitationCondition.lua
+++ b/lua/sim/VictoryCondition/DecapitationCondition.lua
@@ -74,11 +74,17 @@ DecapitationCondition = Class(AbstractVictoryCondition) {
             end
         end
 
-        -- process all defeated brains. At this stage, it is an entire team that is defeated at once
-        -- defeated brains must be processed before the game (potentially) ends to sync game results to be sent to server
-        for k = 1, TableGetn(defeatedBrains) do
-            local defeatedBrain = defeatedBrains[k]
-            self:DefeatForArmy(defeatedBrain)
+        -- defeated brains must be processed first to sync game results to be sent to server
+        -- At this stage, it is an entire team that is defeated at once
+        local numDefeated = TableGetn(defeatedBrains)
+        if numDefeated > 0 then
+            for k = 1, numDefeated do
+                local defeatedBrain = defeatedBrains[k]
+                self:DefeatForArmy(defeatedBrain)
+            end
+            -- Give other brains some time to be defeated for a draw
+            WaitTicks(6) -- 6 ticks due to random 0-6 tick delay in Unit death thread
+            return self:EvaluateVictoryCondition()
         end
 
         -- no remaining players, just end the game
--- a/changelog/snippets/fix.6831.md
+++ b/changelog/snippets/fix.6831.md
@@ -1,0 +1,1 @@
+- (#6831) Fix draws, recalls, and disconnects resulting in unknown results or incorrect victories.
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -1418,7 +1418,7 @@ end
 Callbacks.ModeratorEvent = function(data)
     -- show up in the game logs
     local brain = GetArmyBrain(GetCurrentCommandSource())
-    SPEW(string.format("Moderator event for %s (Army %d): %s", tostring(brain.Nickname), brain.Army, tostring(data.Message)))
+    SPEW(string.format("Moderator event for %s: %s", tostring(brain.Nickname), tostring(data.Message)))
 end
 
 --#endregion
--- a/lua/SimCallbacks.lua
+++ b/lua/SimCallbacks.lua
@@ -1418,7 +1418,7 @@ end
 Callbacks.ModeratorEvent = function(data)
     -- show up in the game logs
     local brain = GetArmyBrain(GetCurrentCommandSource())
-    SPEW(string.format("Moderator event for %s: %s", tostring(brain.Nickname), tostring(data.Message)))
+    SPEW(string.format("Moderator event for %s (Army %d): %s", tostring(brain.Nickname), brain.Army, tostring(data.Message)))
 end
 
 --#endregion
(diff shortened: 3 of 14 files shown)


lL1l1:
Can't sync abandonedbyplayer, apparently disconnecting causes all others to abandon too, so in a 1v1 you'll get two contradicting results of victory/defeat vs defeat/victory.

lua/sim/VictoryCondition/DecapitationCondition.lua
```diff
@@ -74,6 +74,19 @@ DecapitationCondition = Class(AbstractVictoryCondition) {
...
+        -- At this stage, it is an entire team that is defeated at once
+        local numDefeated = TableGetn(defeatedBrains)
+        if numDefeated > 0 then
+            for k = 1, numDefeated do
+                local defeatedBrain = defeatedBrains[k]
+                self:DefeatForArmy(defeatedBrain)
+            end
+            -- Give other brains some time to be defeated for a draw
+            WaitTicks(6) -- 6 ticks due to random 0-6 tick delay in Unit death thread
+            return self:EvaluateVictoryCondition()
```
Garanas, line 87: I do not see the need for a recursive call.

I understand that not all units are destroyed equally. I think it's better to catch this behavior in the monitoring thread. If we increase the delay to check the game state from 4 ticks to (say) 11 ticks then we catch this situation. We can add the comment there to describe why the delay is 11 ticks. And then there's also no need for the recursive call - we just let the monitor thread process the rest of the game state in the next evaluation.

Garanas (commented):
Great analysis. I've not yet looked into the logic for recall, I'll try and look into that later.

@BlackYps I think we should do another hotfix in a week or so. This isn't game breaking, but it surely is annoying if it is the status quo for month(s).

Garanas:
> Can't sync abandonedbyplayer, apparently disconnecting causes all others to abandon too, so in a 1v1 you'll get two contradicting results of victory/defeat vs defeat/victory.

The old logic did not take this event into account. Looking at the [logic](https://github.com/FAForever/fa/blob/2c230d5a240b92fa0de500c75c29fbb07f1ffc77/lua/aibrain.lua#L475-L556) I think the army simply loses his ACU and is then considered defeated because of that. 

I think the same applies to the recall behavior. The distinction does not matter - the army is defeated because it loses the units required for the victory condition. And therefore the army is defeated.

lua/sim/VictoryCondition/DecapitationCondition.lua
lL1l1: Increasing the monitoring delay decreases the chance of catching two units killed at the same tick but destroyed at different ticks, but it does not eliminate it.

The primary problem is that the monitoring thread immediately kills itself and stops monitoring upon the game ending, even if the game would actually be a draw a short time later.
In both versions of the old code there was a similar repetition of the evaluation where evaluation was repeated after a delay until the defeated brain count stopped changing.

lL1l1:
I believe the way that the old logic handled disconnects is that it relied on the ACU being Destroyed on a delay long enough for the disconnecting player to stop their sim so that they don't send any game results.

lL1l1:
I think it is clean to not rely on side effects of recall or abandonment to leave us in the correct match state, so I'd like to have direct sync calls for recall/abandon.
Logically an army that is abandoning should not sync game results as it no longer has an accurate simulation. This would be accomplished by flagging when the client has called `SessionEndGame()` and then discarding game result sync when the session is in the process of ending.

Also it's worth noting that this incorrect victory report only happens when you use the ["End Game" button in our LAN setup](https://github.com/FAForever/fa/blob/a34ffb7af22e51c24ac613e65a9fb177e24562f3/lua/ui/game/tabs.lua#L186), which is replaced by ["Exit Game" in "gpgnet" FAF games](https://github.com/FAForever/fa/blob/a34ffb7af22e51c24ac613e65a9fb177e24562f3/lua/ui/game/tabs.lua#L234).
This is because Exit Game exits the application 0.1 seconds after ending the game session, which gives no time for game results to sync.

BlackYps:
> @BlackYps I think we should do another hotfix in a week or so. This isn't game breaking, but it surely is annoying if it is the status quo for month(s).

I think so too. But maybe a week is a bit soon. This issue here is not so severe that we need to do a patch ASAP, so I think we can take the time to maybe find out why the replays are the wrong length and include that too.

Garanas:
> > @BlackYps I think we should do another hotfix in a week or so. This isn't game breaking, but it surely is annoying if it is the status quo for month(s).
> 
> I think so too. But maybe a week is a bit soon. This issue here is not so severe that we need to do a patch ASAP, so I think we can take the time to maybe find out why the replays are the wrong length and include that too.

I personally have no idea where to look for that though. The behavior of replays is all engine work. Maybe there's some (side) effect in the parsing of replays that triggers sooner now.

edit: whoops

Garanas:
> I believe the way that the old logic handled disconnects is that it relied on the ACU being Destroyed on a delay long enough for the disconnecting player to stop their sim so that they don't send any game results.

Maybe, but as the person who wrote it that was certainly not part of the idea at the time. The original module that was responsible is this one:

- https://github.com/FAForever/fa/blob/develop/lua/victory.lua

It still exists for backwards compatibility for mods I think - I'm not sure. It also immediately exits, but only checks the game state every 3 seconds.

lL1l1:
It doesn't immediately exit. When it detects a "win" due to all living brains being allied or requesting an allied victory, it adds those winners to the `potentialWinners` table and sets `victoryTime` 5 seconds into the future. Only after those 5 seconds pass does it assign `"OnVictory"` as the callback to end the game, but only if the `stillAlive` table matches the `potentialWinners` (i.e. nobody was defeated in 5 seconds, which is what a draw looks like from the player perspective).

Garanas:
Oh, you're right. It's hidden here:

```lua
        local callback = nil
        if win then
            local equal = table.equal(stillAlive, potentialWinners)
            if not equal then
                victoryTime = GetGameTimeSeconds() + 5
                potentialWinners = stillAlive
            end

            if equal and GetGameTimeSeconds() >= victoryTime then
                callback = 'OnVictory'
            end
        elseif draw then
            callback = 'OnDraw'
        else
            victoryTime = nil
            potentialWinners = {}
        end
        ```
I did not see that. It indeed waits to call out the winners. We can add that logic back in. Let me think about it.

Garanas:
I took a more simple approach. It appears to catch all described cases while testing. See also:

- https://github.com/FAForever/fa/pull/6840

It introduces a new field `ProcessedBrains` that keeps track which brains we've processed. A brain remains eligible for being processed until we actually processed it, regardless of the internal defeat status of the brain. This way we guarantee that all relevant brains are processed. That appears to fix the issues too. While testing it shows all events being processed.

PR #6833, opened 2025-05-30
Fix profiler's `RemoveOutliers` when there is 0 IQR

lL1l1:
## Issue
When all benchmark runs give the same exact results, all the results are removed from the final displayed statistics because they are within 1.5x of the IQR (which is 0), and are considered outliers. This shouldn't happen, as the comment for the function says that outliers are points "more than" 1.5x IQR, so keeping values equal to 1.5x IQR is correct.

Don't consider points as outliers when they are equal to 1.5x the IQR.

Fast benchmarks that return the exact same results now have data outputs instead of just all being zeroes.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6833.md
+++ b/changelog/snippets/fix.6833.md
@@ -1,0 +1,1 @@
+- (#6833) Fix profiler discarding all benchmark data as outliers when all data points are the same exact value.
--- a/lua/shared/statistics.lua
+++ b/lua/shared/statistics.lua
@@ -156,7 +156,7 @@ function RemoveOutliers(t, n)
     local size = 0
     for i = 1, n do
         local result = sorted[i]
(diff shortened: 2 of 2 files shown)


Hdt80bro (approved):

lL1l1:
It failed the test because the IQR calculation is incorrect? The input is {1,2,3,4,15} which should result in q1 = 2 q2 = 3 q3 = 4 and an iqr of 2, and 15 is 2 * 1.5 from the median (3). When I log it it gives me an IQR15 of 12, so in the past it was very specifically excluding 15 for being equal to 3 + 12, but that doesn't make sense.

Hdt80bro:
Ah, you're correct - this is why we have tests :)
I'll put in a fix tonight.

Hdt80bro:
Close in favor of #6848

PR #6837, opened 2025-05-31
Improve annotations related to Buffs

lL1l1:
## Description of the proposed changes
- Solves annotation warnings without changing functionality.
  - except for fixing an incorrect return value in buff calculate, but it's just for a warning message and returns defaults just like when the function fails due to lacking a buff affect.
- adds missing annotations for buff effect functions
  - I wish I could do `table<BuffAffectName, fun(buffdef: BuffBlueprint ... afterRemove: boolean )>` but function annotations only work for function params (calling the param inside the function and inputting params when calling the function)

## Testing done on the proposed changes
Verified that all the changes are refactoring how code is placed or what comments are written. Also spawned a preset SACU to test that buffs work on a basic level.

## Checklist
- [x] Changes are annotated, including comments where useful
- [x] Changes are documented in a changelog snippet according to the [guidelines](https://faforever.github.io/fa/development/changelog#format-of-a-snippet).
- [ ] Request 2-3 reviewers from the [list of reviewers and their areas of knowledge](https://github.com/FAForever/fa/blob/deploy/fafdevelop/CONTRIBUTING.md#reviewers).

PR #6839, opened 2025-05-31
Fix outdated `Health` and `DamageRadius` buffs

lL1l1:
Based on https://github.com/FAForever/fa/pull/6837

Fix outdated buffs that would just error if you used them.

Use debug setup and spawn units. The SACU will spawn with 10k HP damaged and it can groundfire to damage the paragon due to the damage radius buff.
```
   CreateUnitAtMouse('url0301_cloak', 0,    3.00,    0.61,  0.00000)
   CreateUnitAtMouse('xab1401', 0,   -3.00,   -0.61,  0.00000)
```

## Checklist

### DIFF
--- a/changelog/snippets/fix.6839.md
+++ b/changelog/snippets/fix.6839.md
@@ -1,0 +1,1 @@
+- (#6839) Fix `Health` and `DamageRadius` buff effect types. Previously they would just error if you used them.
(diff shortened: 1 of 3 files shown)


lL1l1 (commented):
An idea for better compatibility between DamageRadius buff affects and `weapon.AddDamageRadiusMod` would be to have a new field `wep.DamageRadiusBuffMod` and treat `bp.DamageRadius + wep.DamageRadiusMod` as the initial value.

The same applies to the Damage buff affect.

Also imo there should be an option to specify `BuffedWeapons = Label[]` in the buff definition so then we can iterate that table and check unit.WeaponInstances for what label weapons to buff. Integer indices would also technically work but really all non-dummy weapons should be labeled properly for mod compatibility.

speed2CZ (approved):

PR #6840, opened 2025-05-31
Fix issues with the victory results not being sent to the server

Garanas:

Alternative solution to https://github.com/FAForever/fa/pull/6831

### DIFF
--- a/lua/sim/VictoryCondition/AbstractVictoryCondition.lua
+++ b/lua/sim/VictoryCondition/AbstractVictoryCondition.lua
@@ -288,6 +288,10 @@ AbstractVictoryCondition = Class(DebugComponent) {
     ---@param self AbstractVictoryCondition
     ---@param aiBrain AIBrain
     DrawForArmy = function(self, aiBrain)
+        if self.EnabledSpewing then
+            SPEW("Army forfeit the game: ", aiBrain.Nickname)
+        end
+
         self:FlagBrainAsProcessed(aiBrain)
         self:ToObserver(aiBrain)
         aiBrain:OnDraw()
@@ -300,6 +304,10 @@ AbstractVictoryCondition = Class(DebugComponent) {
     ---@param self AbstractVictoryCondition
     ---@param aiBrain AIBrain
     VictoryForArmy = function(self, aiBrain)
+        if self.EnabledSpewing then
+            SPEW("Army is victorious: ", aiBrain.Nickname)
+        end
+
         self:FlagBrainAsProcessed(aiBrain)
         self:ToObserver(aiBrain)
         aiBrain:OnVictory()
@@ -312,6 +320,10 @@ AbstractVictoryCondition = Class(DebugComponent) {
     ---@param self AbstractVictoryCondition
     ---@param aiBrain AIBrain
     DefeatForArmy = function(self, aiBrain)
+        if self.EnabledSpewing then
+            SPEW("Army is defeated: ", aiBrain.Nickname)
+        end
+
         self:FlagBrainAsProcessed(aiBrain)
         self:ToObserver(aiBrain)
         aiBrain:OnDefeat()
--- a/lua/sim/VictoryCondition/AbstractVictoryCondition.lua
+++ b/lua/sim/VictoryCondition/AbstractVictoryCondition.lua
@@ -114,19 +114,12 @@ AbstractVictoryCondition = Class(DebugComponent) {
         return false
     end,
 
-    --- Retrieves a unique name for the given brain.
-    ---@param self AbstractVictoryCondition
-    ---@param aiBrain AIBrain
-    GetBrainName = function(self, aiBrain)
-        return aiBrain.Name
-    end,
-
     --- A utility function that returns whether a brain is eligible to be considered for victory conditions.
     ---@param self AbstractVictoryCondition
     ---@param aiBrain AIBrain
     ---@return boolean
     BrainIsEligible = function(self, aiBrain)
-        if self.ProcessedBrains[self:GetBrainName(aiBrain)] then
+        if self.ProcessedBrains[aiBrain.Name] then
             return false
         end
 
@@ -348,7 +341,7 @@ AbstractVictoryCondition = Class(DebugComponent) {
     ---@param self AbstractVictoryCondition
     ---@param aiBrain AIBrain
     FlagBrainAsProcessed = function(self, aiBrain)
-        local aiBrainName = self:GetBrainName(aiBrain)
+        local aiBrainName = aiBrain.Name
 
         if self.EnabledSpewing then
             SPEW("Flagging brain as processed: " .. aiBrainName)
@@ -361,7 +354,7 @@ AbstractVictoryCondition = Class(DebugComponent) {
     ---@param self AbstractVictoryCondition
     ---@param aiBrain AIBrain
     DrawForArmy = function(self, aiBrain)
-        local aiBrainName = self:GetBrainName(aiBrain)
+        local aiBrainName = aiBrain.Name
 
         if self.EnabledSpewing then
             SPEW("Army forfeit the game: ", aiBrainName)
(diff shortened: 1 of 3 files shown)


lua/sim/VictoryCondition/AbstractVictoryCondition.lua
```diff
@@ -101,7 +109,7 @@ AbstractVictoryCondition = Class(DebugComponent) {
     ---@param aiBrain AIBrain
     ---@return boolean
     BrainIsEligible = function(self, aiBrain)
-        if aiBrain:IsDefeated() then
+        if self.ProcessedBrains[aiBrain.Nickname] then
```
lL1l1: Can brains have identical nicknames? Can it be nil? `OnCreateArmyBrain` in siminit doesn't have this annotated.

As an alternative we have `aiBrain.Name` which corresponds to a start slot, so it should definitely be a unique string.

lua/sim/VictoryCondition/AbstractVictoryCondition.lua
```diff
@@ -189,7 +197,7 @@ AbstractVictoryCondition = Class(DebugComponent) {
     MonitoringThread = function(self)
         while not IsGameOver() do
             self:EvaluateVictoryCondition()
-            WaitTicks(4)
+            WaitTicks(11)
```
lL1l1: With increased delay, there is still an opportunity for the random delay on unit destruction to cause incorrect results if there is no second check.

lL1l1:
I like the simplicity.
Tests:
- Wwith some `net_lag` and setting the processing wait time to 1 tick and couldn't get `SessionEndGame()` to produce conflicting game results, so there shouldn't be a problem when a player just leaves a game.
- Disconnects (I just ejected the peer manually) can still produce conflicting results which will matter in 1v1 where there are no additional players to rely on. I don't think this is solvable as nobody knows who disconnected for what reason.
- Setting the processing wait to 1 tick and having the two players draw causes one player to win. So in theory the "draw bug" is still possible.

lua/sim/VictoryCondition/AbstractVictoryCondition.lua
Garanas: That's probably the better candidate to use then, yes!

Garanas:
> * Setting the processing wait to 1 tick and having the two players draw causes one player to win. So in theory the "draw bug" is still possible.

This is a different type of draw bug! 😄 

What did you change exactly, the wait duration of the monitoring thread?

> * Disconnects (I just ejected the peer manually) can still produce conflicting results which will matter in 1v1 where there are no additional players to rely on. I don't think this is solvable as nobody knows who disconnected for what reason.

I'm not sure how to deal with this situation. As you state, there's a wide range of reasons why one peer may not send any game results. And no Lua check can force a player to send the data.

lua/sim/VictoryCondition/AbstractVictoryCondition.lua
Garanas: Okay, so let's think about this.

For my idea - this does not apply to assassination, right? The explosion of ACUs is not random I think. The unit does 'last' a few ticks, I am not sure if it is considered 'destroyed' at that point. I think this is more about Supremacy, where my last engineer died on land and yours is still sinking?

And what would be the correct definition of verifying whether the two checks mean the same? Should it be the same brains winning?

We could for example make a function called 'TryEndGame' which attempts to end the game. And if it's called a second time with the 'correct conditions' then it proceeds to kill the monitoring thread and actually end the game.

lL1l1:
> What did you change exactly, the wait duration of the monitoring thread?

Yes, I just changed it to 1 tick so that it can always catch differences in the time until an ACU is Destroyed.

https://github.com/FAForever/fa/blob/a1b013d94d5402ffaa831feb63c35050aa739611/lua/sim/VictoryCondition/AbstractVictoryCondition.lua#L198-L201

lua/sim/VictoryCondition/AbstractVictoryCondition.lua
lL1l1: This does apply to Assassination because all units have a built-in random delay in their death sequence before they have Destroy called on them:
https://github.com/FAForever/fa/blob/0a6c5cec2b2a8470dae109b7cae60ee3599a24bf/lua/sim/Unit.lua#L1984

The delay is there to make explosions not look exactly the same when multiple units die. You could get rid of it for only ACUs but then you'd get issues with modded Command units, alternative victory Unit Conditions, or visual animation stuff like sinking or transports.

The correct definition would be the same brains winning (or more accurately, the same brains being alive) after the 2nd check, but I don't think ending the game should rely on the same function being called twice with the same conditions since it leaves no room to set a specific time within which you can draw (I mentioned in the other PR that it was 5 seconds in victory.lua, and it was like `numLivingBrains` ticks in matchstate.lua, which I consider extremely low actually).

lL1l1:
> I'm not sure how to deal with this situation. As you state, there's a wide range of reasons why one peer may not send any game results. And no Lua check can force a player to send the data.

I think you misunderstood; I disconnected the players from each other but not the server, so they both think they won because the other player abandoned and lost. It's not possible to address and not worth addressing since typically disconnects would be caused by internet outages (no connection to server) or closing the game (session ends before thread can call SyncGameResults).

Garanas:
> > I'm not sure how to deal with this situation. As you state, there's a wide range of reasons why one peer may not send any game results. And no Lua check can force a player to send the data.
> 
> I think you misunderstood; I disconnected the players from each other but not the server, so they both think they won because the other player abandoned and lost. It's not possible to address and not worth addressing since typically disconnects would be caused by internet outages (no connection to server) or closing the game (session ends before thread can call SyncGameResults).

I miss understood. How did you create this test scenario exactly?

edit: understood and deleted, since the described approach is trivial to use by players.

lL1l1:
I played a 2 player test and just had 1 player eject the other using the console `EjectSessionClient(2)`. It's what I meant by "(I just ejected the peer manually)"

lua/sim/VictoryCondition/AbstractVictoryCondition.lua
Garanas: Processed in https://github.com/FAForever/fa/pull/6840/commits/c87caff89f836cfe20e3ebb9d66b2dffce2d0397

lL1l1: Why the function just to retrieve the table value? If it's a documentation issue, the class annotations should describe the value sufficiently.

Garanas: Fine with me, see https://github.com/FAForever/fa/pull/6840/commits/219c6127b69b4eb41043d582b05c84091c3391c7

lua/sim/VictoryCondition/AbstractVictoryCondition.lua
Garanas: How about https://github.com/FAForever/fa/pull/6840/commits/93c47cdbe539ca9ba8efa4df1f2dfd17d08c9b29 ?

lL1l1 (changes requested):
- I tested a draw in a 1v1, and didn't see any report of defeat, which will result in unknown results. These two bits need to be swapped:
https://github.com/FAForever/fa/blob/93c47cdbe539ca9ba8efa4df1f2dfd17d08c9b29/lua/sim/VictoryCondition/UnitCondition.lua#L59-L74

- The monitoring thread delay can be reduced again.

- Recall reports defeat correctly. For a future fix, I noticed that in 1v1 there is no recall announcement and that in team games the recall announcement gets hidden somewhat quickly by defeat announcements.

Everything else looks good and I can approve it after the draws are fixed.

Garanas:
@lL1l1 There you go!

I may create a new pull request in the near future to fix some naming conventions. They're a little over the place at the moment. Not crucial for this fix though.

lL1l1 (approved):

PR #6842, opened 2025-06-04
Give land scouts radar structure priority

lL1l1:
Land scouts have very low dps that is often not useful but can still kill radars, damaged units, and other land scouts. The target priorities are changed to better suit that.

Spawning the units gives no errors and they target radars over other units.

## Checklist

### DIFF
--- a/units/UAL0101/UAL0101_unit.bp
+++ b/units/UAL0101/UAL0101_unit.bp
@@ -203,7 +203,7 @@ UnitBlueprint{
             RangeCategory = "UWRC_DirectFire",
             RateOfFire = 10/20, --10/integer interval in ticks
             TargetPriorities = {
-                "(MOBILE * SCOUT + INTELLIGENCE * STRUCTURE * RADAR)",
+                "RADAR INTELLIGENCE STRUCTURE",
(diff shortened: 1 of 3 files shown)


units/UAL0101/UAL0101_unit.bp
```diff
@@ -203,8 +203,8 @@ UnitBlueprint{
             RangeCategory = "UWRC_DirectFire",
             RateOfFire = 10/20, --10/integer interval in ticks
             TargetPriorities = {
+                "RADAR INTELLIGENCE STRUCTURE",
```
4z0t, line 206: Shouldn't these be writted the way removed ones were done?

lL1l1: We have two category expression string parsers:
https://github.com/FAForever/fa/blob/309b408c827e7dc60a24b7c78bfa64ac25d114c7/lua/sim/weapon.lua#L58-L62

The engine one only handles intersections, while the lua one also parses unions and exclusions. A parentheses is used to pick the lua parser.

Rhaelya (approved):
Everything looks in order.

PR #6843, opened 2025-06-04
Reduce Cybran T1 mobile AA's damage

lL1l1:
Since the projectiles can retarget the AA doesn't need so much DPS to stay equal with other factions.

- **Sky Slammer: T1 Mobile Anti-Air Gun (URL0104)**
  - Nanodart Launcher (Anti Air)
    - Damage: 16 -> 14

A bomber flying over the AA now takes 182 damage instead of 208. Aeon and Sera deal 168 and 169 damage, while UEF deals 182 damage too.

A group of 3 AA seems to deal with 3 bombers just slightly faster than other factions, owing to its inability to miss and still slightly higher DPS.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6843.md
+++ b/changelog/snippets/balance.6843.md
@@ -1,0 +1,5 @@
+- (#6843) Reduce the damage of the Cybran T1 mobile anti-air since its projectiles can now retarget, and it no longer needs so much damage to match the AA of other factions.
+
+  **Sky Slammer: T1 Mobile Anti-Air Gun (URL0104)**
+    - Nanodart Launcher (Anti Air)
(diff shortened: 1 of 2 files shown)


RabidPope:
I dont really understand why you would lower the lowest dps anti air. The main problem stems from them shooting things after they leave the range of the unit.

lL1l1:
It has a reload time of 0.5, so the DPS goes from 32 to 28. UEF/Sera AA have 26 DPS and Aeon has 24 DPS with 3 more range. I'll add the DPS value to the patch note since it was confusing.

Rhaelya (approved):
Everything looks in order

PR #6845, opened 2025-06-05
Fix various issues with announcements

### LINKED ISSUE #6841: [Bug]: When an objective fails, so does the announcement
### Describe the Bug

Objectives create an announcement when they are created. They also produce one when they are [completed, or failed](https://github.com/FAForever/fa/blob/develop/lua/ui/game/objectives2.lua#L398). The control that is responsible for showing the announcement is destroyed in the same frame, right after the announcement is created.

This is problematic. In the new implementation the [animation is part of a thread](https://github.com/FAForever/fa/blob/develop/lua/ui/game/announcement/AbstractAnnouncement.lua#L302) and therefore when we interact with the control we may already have lost the reference that we need. That produces a load of errors in the log. 

### Reproduce the bug

Play the map 'Rainmakers Survival'. Ignore the warnings about deep copy. Wait for the time-based objective 'The calm before the storm' to finish. See the logs full of errors about functions that no longer exist.

### Screenshots

_No response_

### Additional context

_No response_

Garanas:

Fixes various issues with the new setup for announcements. Closes #6841 .

It fixes the following behavior:

- Re-introduces the callback, which is now called when the announcement is finished. There was 1 place where this was used: objectives to cleanup after themselves when the announcement ends!
- Refactors the animation to be a separate function called `SetupBackgroundAnimation`. The applied animation can then be controlled by setting the progress variable that is returned.
- Fixes an edge case where the animation breaks if the origin/target control is destroyed during the animation.

Also fixes a subtle issue where the position of the origin/target control must be known before we fork a thread. By forking, we push the remaining logic to the end of the internal call stack. In the meantime the control can be destroyed, before the animation even had the time to process it.

Launched various campaign maps and campaign-like maps to see and verify the interactions with the objectives. Specifically:

- Rainmakers Survival (ignore the initial warnings about deep copying).
- The first campaign map of the FA campaign.

## Checklist

### DIFF
--- a/lua/ui/game/announcement.lua
+++ b/lua/ui/game/announcement.lua
@@ -81,9 +81,11 @@ CreateTitleAnnouncement = function(titleText, goalControl, onCompleteCallback)
 
     -- create a dummy goal control if we don't have one
     local frame = GetFrame(0) --[[@as Frame]]
-    if not goalControl or IsDestroyed(goalControl) then
+    if IsDestroyed(goalControl) then
         goalControl = CreateDefaultGoalControl(frame)
     end
+
+    ---@cast goalControl -nil, -unknown
 
     -- developers note: lazy load the module so that it remains unloaded unless used
     local TitleAnnouncement = import("/lua/ui/game/announcement/TitleAnnouncement.lua").TitleAnnouncement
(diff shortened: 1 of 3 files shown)


lua/ui/game/announcement.lua
```diff
@@ -80,15 +81,17 @@ CreateTitleAnnouncement = function(titleText, goalControl)
 
     -- create a dummy goal control if we don't have one
     local frame = GetFrame(0) --[[@as Frame]]
-    goalControl = goalControl or CreateDefaultGoalControl(frame)
+    if not goalControl or IsDestroyed(goalControl) then
```
4z0t: `IsDestroyed` already checks for nil.

Garanas: Yes, but the plugin does not understand that 😸

4z0t: I wonder if 
```lua
---@overload fun(object:nil): true
...
```
will work.

Garanas: It appears to not work, unfortunately:

image

lL1l1: Searching the issue for the language server they recommend using `---@cast goalControl -nil`.
docs: https://luals.github.io/wiki/annotations/#cast

lL1l1: Which results in this:
{74822151-ED75-45AF-A504-EA3B8C2799EB}

Garanas: Applied in https://github.com/FAForever/fa/pull/6845/commits/b7068b1f03fcc0160fe6357519c966b39f53b96d

PR #6848, opened 2025-06-07
Fix `RemoveOutliers` statistics

Hdt80bro:
When all values in a list are the same, the IQR is 0 and the `RemoveOutliers` function was incorrectly handling that case. However, #6833 revealed more bugs that required more testing and exposing the quartiles as a separate function in `GetQuartiles` made that easier. `RemoveOutliers` no longer sorts the values as a consequence.


lL1l1 (approved):
Maybe a snippet could be added.

PR #6850, opened 2025-06-07
Refactor regen buff

lL1l1:
## Issue
- The default BuffCalculate function hijacks the Regen mult calculation as said in https://github.com/FAForever/fa/pull/5170#issuecomment-1619073588 by @Hdt80bro.

- The `UniqueBuffs` table does not give a way to stack different unique buffs; the unique buff has to do correct calculations for all buffs, while adding its own unique calculation in the middle.
  - It's worth noting the `UniqueBuffs` table was never used because BuffCalculate never passed in a buffname for Regen (the second param is `nil`).
  https://github.com/FAForever/fa/blob/5adcca7113beda198176d9511ffc827cd30bc855/lua/sim/Buff.lua#L325

Branch is based on #6837.

Refactor the regen aura buff calculations into a new `UniqueAffectCalculation` function table. `UniqueAffectCalculation` can be used to give specific buffs a unique way of calculating adds or mults for a buff. `Floor` and `Bool` are not made available as they don't seem to make sense to adjust uniquely because they override values of the entire affect stack. This allows a single buff to have a unique calculation for its contribution to the overall affect, instead of having it calculate the entire stack with all buffs + its unique effect.
  - Since the UniqueBuffs table was not used, I refactored the base regen calculation and not the unique one into the new function.

This would break other mods that rely on the existing regen mult behavior and aren't using the correct buff names.

Regen Aura ACU gives the correct regen amount to SACUs with regen upgrades.
```
   CreateUnitAtMouse('xab1401', 0,   -5.67,   -2.00, -0.00000)
   CreateUnitAtMouse('xsl0001', 0,    1.33,    2.00, -0.00000)
   CreateUnitAtMouse('ual0301_nanocombat', 0,    4.33,    0.00, -0.00000)
```

## Additional context
BlackOps FAF Unleashed shouldn't break due to these changes because it uses the same buff names as FAF:
[`SeraphimACURegenAura`](https://github.com/Uveso/BlackOpsFAF-ACUs/blob/e6cccf541a120546508f9ac4ddf793f2437e284d/units/ESL0001/ESL0001_script.lua#L479)
[`SeraphimACUAdvancedRegenAura`](https://github.com/Uveso/BlackOpsFAF-ACUs/blob/e6cccf541a120546508f9ac4ddf793f2437e284d/units/ESL0001/ESL0001_script.lua#L561C29-L561C57)

## Checklist

### DIFF
--- a/changelog/snippets/features.6850.md
+++ b/changelog/snippets/features.6850.md
@@ -1,0 +1,5 @@
+- (#6850) Create a new `UniqueAffectCalculation` table in Buffs.lua, which allows a buff to have a unique calculation for how much it adds/multiplies an effect, while letting all other buffs that use the effect to calculate independently (unlike the old `UniqeBuffs` table).
+
+  An example is the Seraphim regen aura which adds regeneration to a unit by a percent of its max HP with maximum amounts based on tech level. Previously it was hijacking the entire multiplier calculation for the regen effect and also had to calculate additive regeneration effects from other buffs. Now it is refactoed to use the new system so it works without affecting the multipliers from other buffs and doesn't have to calculate the additive effects of other buffs.
+
+  Note for mod developers: If your mod relies on the regen multiplier effect to be based on unit HP, and isn't using the `SeraphimACURegenAura` or `SeraphimACUAdvancedRegenAura` buff names, it will break since the multiplier will be multiplying the additive effect of regen buffs.
--- a/lua/sim/Buff.lua
+++ b/lua/sim/Buff.lua
@@ -102,7 +102,7 @@ UniqueAffectCalculation = {
     SeraphimACUAdvancedRegenAura = { Regen = regenAuraCalculate },
 }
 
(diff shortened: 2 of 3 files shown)


lL1l1:
I added changes I made while I was working on my sacu balance mod so I had the opportunity to test how the new refactor works.
Originally I wanted to pass in the `buffsForAffect` so that I could count how much regen I added from different regen field buffs (SACU, ACU lvl1, ACU lvl2) to a unit, but this ended up having issues because of the varying max regen amount. I ended up using `UniqueBuffs` instead for more power, so that I could get all the adds from the unique calculations per regen aura affect buff, sort the adds by the max allowed by their buff, and then apply those adds in a sorted order according to their max, so that no add goes over their own buff's max. But if someone has a different idea that doesn't run into such per-buff-definition issues, then I think `buffsForAffect` is the most reasonable place to store the data, compared to the next accessible table which is the unit table.

BlackYps:
Closing due to inactivity.

PR #6761, opened 2025-05-03
Experiment setup to determine sane LOD values for tree (groups)

Garanas:
## Description of the proposed changes

Now that we have https://github.com/FAForever/fa/pull/6752 to bake properties we have the ability to bake in the LOD values that we want. And since https://github.com/FAForever/fa/pull/6751 is merged in we can safely assume that all players will start with the 'High' settings for LODs. Users that take the 'Extreme' setting are on their own.

With that we have everything we need to restart the discussion in https://github.com/FAForever/fa/pull/6715.

This pull request provides a standardized test that we can use to inspect the behavior and compare it between hardware configurations. We have to make some assumptions.

- We assume that individual trees and tree groups should have the same LOD to minimize the intel that is leaked through the fog of war when units move through tree groups.
- We assume that the map that is being played is Seton's Clutch. We also assume that this map is representative.
- We assume that players do not play optimal, and therefore do not reclaim tree groups. As a result there will be a lot of broken tree groups because of units traversing them. 
- We assume that breaking half the tree groups is representative of this.

With these assumptions we can get some data. The tree groups are broken automatically. There's two hotkeys with a standardized camera configuration:

image

Then you can open the console using `~`. You can open the statistics by typing `ShowStats`. If you open the entity count tab then you can see there's quite a few props. If you open the frame tab then you can see the current time (in milliseconds) it takes to render a frame. This is a combination of he Lua code being run (in the UI thread, each frame) and the rendering of what is on screen. You can also see the current frames per second (fps).

Then the experiment can start:

- Set game option to for level of detail in the graphics tab to 'High'.
- Change the `weight` factor of tree groups in [blueprint-props.lua](). Ideally we check all values ranging from 2.6 up to 14.0 in steps of 2.0.
- (Re)start the map (default hotkey is CTRL + F10)
- Open the statistics.
- For each hotkey with fixed camera positions
- - Write down the frame time. 

## Additional context

Note that https://github.com/FAForever/FA-Binary-Patches/pull/115 also exists. It does not solve the problem of this pull request - if you increase the LOD by too much it will still just lag. but it may (instead) give players the ability to scale the LOD values of the entities (units, projectiles or props) that the player is most interested in. 

I'm not sure if being able to scale LOD values of props, units and/or projectiles is a healthy game option to have. It's super specific and technical. Not a lot of players will understand that.

PR #6851, opened 2025-06-08
Extend the rendering distance of all tree (groups)

Garanas:

Based on https://github.com/FAForever/fa/pull/6761 and https://github.com/FAForever/fa/pull/6763 I think it's now safe to push the rendering distance for trees by quite a bit. 

## Additional context

The initial comment of https://github.com/FAForever/fa/pull/6761.

## Checklist


Garanas:
Once merged in, I'll bake in the new values into the blueprints.

lua/system/blueprints-props.lua
```diff
@@ -152,6 +152,9 @@ end
 --- - .github\workflows\bake-blueprints.yaml
 ---@param blueprints BlueprintsTable
 function BakePropBlueprints(blueprints)
+
+
+
```
lL1l1:
```suggestion
```

changelog/snippets/graphics.6851.md
```diff
@@ -0,0 +1,4 @@
+- (#6851) Push the rendering distance of tree (groups) to better match the values in Steam
+
+After thorough testing we think it is safe to increase the rendering distance without running the risk to destroying the framerate of the average user. With thanks to an observation of Strogo, most of the frame time was spent on the shadow pass of the image. Two years ago we made it so that shadows would always render. Regardless of the distance. This means whenever there's a mesh, the mesh would have a shadow. This was not the case in Steam. And apparently this is what makes tree (groups) very expensive. Now, we turned that into a slider. That way players with better hardware can be more on the edge of what is possible, while players with weaker hardware can still see the trees at a greater distance, but without a shadow.
```
lL1l1:
```suggestion
After thorough testing we think it is safe to increase the rendering distance without running the risk to destroying the framerate of the average user. With thanks to an observation of Strogo, most of the frame time was spent on the shadow pass of the image. Two years ago we made it so that shadows would render regardless of the distance. This means that whenever there's a mesh, the mesh would have a shadow, which was not the case in Steam, and apparently this is what makes tree (groups) very expensive. With #6763, we turned the shadow rendering distance into a slider in the game options, so players with better hardware can be more on the edge of what is possible, while players with weaker hardware can still see the trees at a greater distance, but without a shadow.
```

Aside from these wording changes, this snippet is repeating #6763's snippet too much. I would look at the two together to make sure they don't repeat information.

lL1l1 (commented):
Shadow distance: 440 (max)
It feels a bit low imo. I feel like I could handle it if it was higher but we will have to see (would personally test in live games with console commands). I wouldn't be against increasing the max value for the slider.

High LOD min zoom to see all trees:
{C1D9C0C5-DF52-4855-A902-E2A041DCB7C7}

High LOD has a zoom step where trees are partially visible. Shadows also pop-in due to the hard LoD limit. I assume both are dependent on user zoom settings though so they can't be fixed.
{DCC62CFA-49CF-409E-B134-C5755B2EDAA7}


Extreme LOD min zoom to see all trees:
{DD0352D7-DB1B-40AE-A4C9-2138D8CD601F}

PR #6857, opened 2025-06-13
Add UIUtil.CreateSCXMenuPanelBorder

4z0t:
Take the implementation of the scx border creation function out of many other identical ones to UIUtil.

## Checklist

### DIFF
--- a/lua/ui/dialogs/disconnect.lua
+++ b/lua/ui/dialogs/disconnect.lua
@@ -16,7 +16,6 @@ local scenario = SessionGetScenarioInfo()
 
 local parent = nil
 local myIndex = ''
-local CreateBorder
 
 function DestroyDialog()
     SessionClients.ResetInterval()
@@ -42,7 +41,7 @@ local function CreateDialog(clients)
     bg = Bitmap(parent, UIUtil.UIFile('/scx_menu/panel-brd/panel_brd_m.dds'))
     LayoutHelpers.FillParent(bg, parent)
 
-    bg.border = CreateBorder(bg)
+    bg.border = UIUtil.CreateSCXMenuPanelBorder(bg)
 
     local dlgTitle = UIUtil.CreateText(bg, "<LOC UI_Disco0000>Connectivity", 18)
     LayoutHelpers.AtTopIn(dlgTitle, bg, -52)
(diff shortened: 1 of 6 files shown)


lua/ui/uiutil.lua
```diff
@@ -1492,3 +1492,68 @@ function CreateHorzFillGroup(parent, filename)
...
+---@field tr Bitmap
+---@field l Bitmap
+---@field r Bitmap
+---@field bl Bitmap
+---@field bm Bitmap
+---@field br Bitmap
+
+---@param parent Control
+---@return SCXMenuPanelBorderTable
+function CreateSCXMenuPanelBorder(parent)
```
lL1l1, line 1500: Can you add a description of what this border looks like/where it is used?
Image for reference:
{906C54CE-5A1E-4823-94FF-957EE5203190}

Also for context it is used in the restricted units dialog, but that's only the old one, which we don't use anymore. So basically its only seen in the disconnect/connectivity dialogs.

lua/ui/uiutil.lua
```diff
@@ -1492,3 +1492,68 @@ function CreateHorzFillGroup(parent, filename)
...
+
+---@class SCXMenuPanelBorderTable
+---@field tl Bitmap
+---@field tm Bitmap
+---@field tr Bitmap
+---@field l Bitmap
+---@field r Bitmap
+---@field bl Bitmap
+---@field bm Bitmap
+---@field br Bitmap
```
lL1l1:
```suggestion
```

lua/ui/uiutil.lua
```diff
@@ -1492,3 +1492,68 @@ function CreateHorzFillGroup(parent, filename)
...
+
+---@param parent Control
+---@return SCXMenuPanelBorderTable
+function CreateSCXMenuPanelBorder(parent)
+    ---@type SCXMenuPanelBorderTable
```
lL1l1: Assigning the class alongside the actual table would make the annotations more concise.
```suggestion
    ---@class SCXMenuPanelBorderTable
```

lL1l1: You also have to add 
```
---@overload fun(parent: Control, filename: Lazy<FileName>, debugname: string?): Bitmap
````
to the Bitmap class for intellisense to get the correct type of the table values.

File: lua/ui/dialogs/disconnect.lua [hunk 4]
```diff
@@ -235,57 +234,6 @@ function Update()
 
 end
 
-function CreateBorder(parent)
```
lL1l1, line 238: backwards compat is broken but these UIs are probably never modded

speed2CZ (approved):

PR #6858, opened 2025-06-14
water absorption improvements

BlackYps:
- Mesh emission was erroneously applied after the water color
- Code should be easier to understand now
- Main change: The water changes less when tilting the camera when using advanced water absorption. The effect was too big before, now it looks more reasonable. Which is not surprising, because I did some research to model real-world behaviour more closely.

Tested it on Runestone that has a custom water ramp.
More testing with different maps and water ramps is needed.

## Checklist


BlackYps:
I did some more testing. This change makes the water appear a bit brighter than before. The affected maps can compensate for this by increasing the abyss height. The change is somewhat subtle so this is not strictly necessary in most cases, if the map author does not have a very particular opinion about how dark the water should look.
The only affected maps are maps that use the advanced water absorption. So far this is only a handful of maps. 99% of the maps are not affected at all by this change.

PR #6860, opened 2025-06-15
Reduce weapon damage table computations when applying buffs by storing constant data

ostrovaya:

When a buff is applied to a weapon, parts of its damage table needs to change. This PR makes it so that this computation only occurs to the parts of the damage table that need to be recomputed and avoids computing the rest by caching them.

Renames the `...Internal` function to have a more meaningful name.

### Replay simulation

Rebase changes onto `1c652a79bfde09257c78d8fe612333cfb606a6f6` and simulate replay 25053165 on game version 3824.

Observe many desyncs starting early into the match that continue to desync.

Observe no differences in end game state before and after changes.

Screenshot from 2025-06-15 22-48-45

Screenshot from 2025-06-15 22-27-05

## Checklist

### DIFF
--- a/lua/sim/weapon.lua
+++ b/lua/sim/weapon.lua
@@ -81,7 +81,7 @@ local WeaponMethods = moho.weapon_methods
 ---@field DamageMod number
 ---@field DamageModifiers number[] # Set of damage multipliers used by collision beams for the weapon
 ---@field DamageRadiusMod number
----@field damageTableCache WeaponDamageTable?
+---@field damageTableCache WeaponDamageTable? # nil when the damage table needs to be refreshed from the weapon blueprint
 ---@field damageTableCacheValid boolean? # false when the damage table needs to be updated with changed damage/radius modifiers or Buffs
 ---@field DisabledBuffs table
 ---@field DisabledFiringBones Bone[] # Bones that `Unit.Animator` cannot move when this weapon has a target
--- a/lua/sim/weapon.lua
+++ b/lua/sim/weapon.lua
@@ -475,31 +475,40 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
 
     ---@param self Weapon
     ---@return WeaponDamageTable
-    RefreshDamageTable = function(self)
+    GetNewDamageTable = function(self)
         local weaponBlueprint = self.Blueprint
+
+        local damageFriendly = weaponBlueprint.DamageFriendly
+        local damageTable = {
+            DamageToShields = weaponBlueprint.DamageToShields,
+            InitialDamageAmount = weaponBlueprint.InitialDamage or 0,
+            DamageType = weaponBlueprint.DamageType,
+            DamageFriendly = damageFriendly == nil and true or damageFriendly,
+            CollideFriendly = weaponBlueprint.CollideFriendly or false,
+            DoTTime = weaponBlueprint.DoTTime,
+            DoTPulses = weaponBlueprint.DoTPulses,
+            MetaImpactAmount = weaponBlueprint.MetaImpactAmount,
+            MetaImpactRadius = weaponBlueprint.MetaImpactRadius,
+            ArtilleryShieldBlocks = weaponBlueprint.ArtilleryShieldBlocks,
+        }
+        damageTable.__index = damageTable
+
+        return damageTable
+    end,
+
+    ---@param self Weapon
+    ---@return WeaponDamageTable
+    RefreshDamageTable = function(self)
         local damageTable = self.damageTableCache or nil
 
         -- Setup the table for values that won't change later
         if not damageTable then
-            local damageFriendly = weaponBlueprint.DamageFriendly
-            ---@diagnostic disable-next-line: missing-fields
-            damageTable = {
-                DamageToShields = weaponBlueprint.DamageToShields,
-                InitialDamageAmount = weaponBlueprint.InitialDamage or 0,
-                DamageType = weaponBlueprint.DamageType,
-                DamageFriendly = damageFriendly == nil and true or damageFriendly,
-                CollideFriendly = weaponBlueprint.CollideFriendly or false,
-                DoTTime = weaponBlueprint.DoTTime,
-                DoTPulses = weaponBlueprint.DoTPulses,
-                MetaImpactAmount = weaponBlueprint.MetaImpactAmount,
-                MetaImpactRadius = weaponBlueprint.MetaImpactRadius,
-                ArtilleryShieldBlocks = weaponBlueprint.ArtilleryShieldBlocks,
-            }
-            damageTable.__index = damageTable
+            damageTable = self:GetNewDamageTable()
         else
             damageTable = self.damageTableCache --[[@as WeaponDamageTable]]
         end
 
+        local weaponBlueprint = self.Blueprint
         -- Add buff
         damageTable.DamageRadius = weaponBlueprint.DamageRadius + self.DamageRadiusMod
         damageTable.DamageAmount = weaponBlueprint.Damage + self.DamageMod
--- a/lua/sim/weapon.lua
+++ b/lua/sim/weapon.lua
@@ -489,20 +489,19 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
         -- Setup the table for values that won't change later
         if self.damageTableCacheValid == nil then
             ---@diagnostic disable-next-line: missing-fields
-            damageTable = {}
-            damageTable.DamageToShields = weaponBlueprint.DamageToShields
-            damageTable.InitialDamageAmount = weaponBlueprint.InitialDamage or 0
-            damageTable.DamageType = weaponBlueprint.DamageType
-            damageTable.DamageFriendly = weaponBlueprint.DamageFriendly
-            if damageTable.DamageFriendly == nil then
-                damageTable.DamageFriendly = true
-            end
-            damageTable.CollideFriendly = weaponBlueprint.CollideFriendly or false
-            damageTable.DoTTime = weaponBlueprint.DoTTime
-            damageTable.DoTPulses = weaponBlueprint.DoTPulses
-            damageTable.MetaImpactAmount = weaponBlueprint.MetaImpactAmount
-            damageTable.MetaImpactRadius = weaponBlueprint.MetaImpactRadius
-            damageTable.ArtilleryShieldBlocks = weaponBlueprint.ArtilleryShieldBlocks
+            local damageFriendly = weaponBlueprint.DamageFriendly
+            damageTable = {
+                DamageToShields = weaponBlueprint.DamageToShields,
+                InitialDamageAmount = weaponBlueprint.InitialDamage or 0,
+                DamageType = weaponBlueprint.DamageType,
+                DamageFriendly = damageFriendly == nil and true or damageFriendly,
+                CollideFriendly = weaponBlueprint.CollideFriendly or false,
+                DoTTime = weaponBlueprint.DoTTime,
+                DoTPulses = weaponBlueprint.DoTPulses,
+                MetaImpactAmount = weaponBlueprint.MetaImpactAmount,
+                MetaImpactRadius = weaponBlueprint.MetaImpactRadius,
+                ArtilleryShieldBlocks = weaponBlueprint.ArtilleryShieldBlocks,
+            }
         else
             damageTable = self.damageTableCache --[[@as WeaponDamageTable]]
         end
--- a/lua/sim/weapon.lua
+++ b/lua/sim/weapon.lua
@@ -479,11 +479,14 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
         local weaponBlueprint = self.Blueprint
 
         local damageFriendly = weaponBlueprint.DamageFriendly
+        if damageFriendly == nil then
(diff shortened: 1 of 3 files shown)


ostrovaya:
@4z0t @speed2CZ @Hdt80bro I kindly request your review. It is appreciated as always.

lua/sim/weapon.lua
```diff
@@ -472,29 +473,43 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
...
+    RefreshDamageTable = function(self)
         local weaponBlueprint = self.Blueprint
         ---@type WeaponDamageTable
-        local damageTable = {}
+        local damageTable
+
+        -- Setup the table for values that won't change later
+        if self.damageTableCacheValid == nil then
+            ---@diagnostic disable-next-line: missing-fields
+            damageTable = {}
```
4z0t: You can clear previous table instead of allocating new one.

ostrovaya: If `damageTableCacheValid == nil`  then no table has yet existed so it must be created.

edit: The function is also going to be called this way in `CreateProjectileForWeapon`.

lua/sim/weapon.lua
```diff
@@ -472,29 +473,43 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
...
+            damageTable.DamageFriendly = weaponBlueprint.DamageFriendly
+            if damageTable.DamageFriendly == nil then
+                damageTable.DamageFriendly = true
+            end
+            damageTable.CollideFriendly = weaponBlueprint.CollideFriendly or false
+            damageTable.DoTTime = weaponBlueprint.DoTTime
+            damageTable.DoTPulses = weaponBlueprint.DoTPulses
+            damageTable.MetaImpactAmount = weaponBlueprint.MetaImpactAmount
+            damageTable.MetaImpactRadius = weaponBlueprint.MetaImpactRadius
+            damageTable.ArtilleryShieldBlocks = weaponBlueprint.ArtilleryShieldBlocks
```
lL1l1:
```suggestion
            local damageFriendly = weaponBlueprint.DamageFriendly
            damageTable = {
                DamageToShields = weaponBlueprint.DamageToShields,
                InitialDamageAmount = weaponBlueprint.InitialDamage or 0,
                DamageType = weaponBlueprint.DamageType,
                DamageFriendly = damageFriendly == nil and true or damageFriendly,
                CollideFriendly = weaponBlueprint.CollideFriendly or false,
                DoTTime = weaponBlueprint.DoTTime,
                DoTPulses = weaponBlueprint.DoTPulses,
                MetaImpactAmount = weaponBlueprint.MetaImpactAmount,
                MetaImpactRadius = weaponBlueprint.MetaImpactRadius,
                ArtilleryShieldBlocks = weaponBlueprint.ArtilleryShieldBlocks,
            }
```

lua/sim/weapon.lua
```diff
@@ -472,29 +473,43 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
...
-    GetDamageTableInternal = function(self)
+    ---@return WeaponDamageTable
+    RefreshDamageTable = function(self)
         local weaponBlueprint = self.Blueprint
         ---@type WeaponDamageTable
-        local damageTable = {}
+        local damageTable
+
+        -- Setup the table for values that won't change later
+        if self.damageTableCacheValid == nil then
```
lL1l1: I don't think keeping state through nil and false is great.
```suggestion
        if not damageTable then
```

lua/sim/weapon.lua
```diff
@@ -472,29 +473,43 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
...
+    -- true -> fresh, ready to use
+    damageTableCacheValid = nil,
     ---@param self Weapon
-    GetDamageTableInternal = function(self)
+    ---@return WeaponDamageTable
+    RefreshDamageTable = function(self)
         local weaponBlueprint = self.Blueprint
         ---@type WeaponDamageTable
-        local damageTable = {}
+        local damageTable
```
lL1l1:
```suggestion
        local damageTable = self.damageTableCache or nil
```

lua/sim/weapon.lua
```diff
@@ -506,15 +521,16 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
...
         return damageTable
     end,
 
-    damageTableCache = false,
     ---@param self Weapon
     ---@return WeaponDamageTable
     GetDamageTable = function(self)
-        if not self.damageTableCache then
-            self.damageTableCache = self:GetDamageTableInternal()
+        if not self.damageTableCacheValid then
```
lL1l1, line 531: Second check is for backwards compatibility.
First check would be for a new way that doesn't reset the entire damage table from the blueprint.
```suggestion
        if not self.damageTableCacheValid or not self.damageTableCache then
```

lL1l1 (changes requested):
This breaks xsl0305 as it changes the `Blueprint` to a new table, and it also relies on `damageTableCache = false` resetting the damage table.
https://github.com/FAForever/fa/blob/e725005f374984ca49120ea8fe930ffac47bd70a/units/XSL0305/XSL0305_script.lua#L135-L137

We can either add backwards compatibility like I suggested or just update xsl0305.

ostrovaya:
Thanks for the tip here. I'll add some extensive global searching for these sorts of changes for next time to catch that.

I'll amend the PR as suggested and test it again.

lua/sim/weapon.lua
```diff
@@ -472,29 +473,41 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
     OnMotionHorzEventChange = function(self, new, old)
     end,
 
+    damageTableCache = nil,
+    -- Determine the validity of the damageTableCache:
+    -- nil -> never initialized;
+    -- false -> stale, needs to be updated;
+    -- true -> fresh, ready to use
+    damageTableCacheValid = nil,
```
lL1l1: `key = nil` doesn't do anything in table initialization technically, and it's already documented in the class annotation so I don't see them as necessary.
Also the `damageTableCacheValid ` comments are now outdated.
```suggestion
```

lua/sim/weapon.lua
```diff
@@ -81,7 +81,8 @@ local WeaponMethods = moho.weapon_methods
 ---@field DamageMod number
 ---@field DamageModifiers number[] # Set of damage multipliers used by collision beams for the weapon
 ---@field DamageRadiusMod number
----@field damageTableCache WeaponDamageTable | false # Set to false when the weapon's damage is modified
+---@field damageTableCache WeaponDamageTable?
+---@field damageTableCacheValid boolean? # Set to false when the weapon's damage is modified and nil when it hasn't been tracked yet
```
lL1l1:
```suggestion
---@field damageTableCacheValid boolean? # false when the damage table needs to be updated with changed damage/radius modifiers or Buffs
```

lua/sim/weapon.lua
```diff
@@ -81,7 +81,8 @@ local WeaponMethods = moho.weapon_methods
 ---@field DamageMod number
 ---@field DamageModifiers number[] # Set of damage multipliers used by collision beams for the weapon
 ---@field DamageRadiusMod number
----@field damageTableCache WeaponDamageTable | false # Set to false when the weapon's damage is modified
+---@field damageTableCache WeaponDamageTable?
```
lL1l1:
```suggestion
---@field damageTableCache WeaponDamageTable? # nil when the damage table needs to be refreshed from the weapon blueprint
```

lua/sim/weapon.lua
```diff
@@ -678,7 +692,7 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
         if dmgRadMod then
             self.DamageRadiusMod = self.DamageRadiusMod + dmgRadMod
         end
-        self.damageTableCache = false
+        self.damageTableCacheValid = false
```
lL1l1: The validation can be set to false inside the conditional.

lua/sim/weapon.lua
```diff
@@ -506,15 +519,16 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
 
         damageTable.__index = damageTable
```
lL1l1: The `__index` assignment can also be moved to table initialization.

lua/sim/weapon.lua
```diff
@@ -472,29 +473,41 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
...
+    damageTableCache = nil,
+    -- Determine the validity of the damageTableCache:
+    -- nil -> never initialized;
+    -- false -> stale, needs to be updated;
+    -- true -> fresh, ready to use
+    damageTableCacheValid = nil,
     ---@param self Weapon
-    GetDamageTableInternal = function(self)
+    ---@return WeaponDamageTable
+    RefreshDamageTable = function(self)
```
lL1l1: The new name loses the meaning that it returns a damage table, so something like `GetFreshDamageTable` or `GetUpdatedDamageTable` would be more appropriate.

ostrovaya: I took this advice and reduced the code's complexity so there's now one function to setup the table and another to update it. Named appropriately.

ostrovaya:
Test plan successfully rerun.

lua/sim/weapon.lua
```diff
@@ -475,29 +475,32 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
...
+            DamageType = weaponBlueprint.DamageType,
+            DamageFriendly = damageFriendly == nil and true or damageFriendly,
+            CollideFriendly = weaponBlueprint.CollideFriendly or false,
+            DoTTime = weaponBlueprint.DoTTime,
+            DoTPulses = weaponBlueprint.DoTPulses,
+            MetaImpactAmount = weaponBlueprint.MetaImpactAmount,
+            MetaImpactRadius = weaponBlueprint.MetaImpactRadius,
+            ArtilleryShieldBlocks = weaponBlueprint.ArtilleryShieldBlocks,
+        }
+        damageTable.__index = damageTable
```
speed2CZ, line 500: Whats the purpose of this line?

ostrovaya: I have to review Lua tables again to answer that. I have no idea and kept it around :)

ostrovaya: Ah. It's to better support inheritance? Not sure why it's applied here though.

https://www.lua.org/pil/13.4.1.html

lua/sim/weapon.lua
```diff
@@ -475,29 +475,32 @@ Weapon = ClassWeapon(WeaponMethods, DebugWeaponComponent) {
 
     ---@param self Weapon
     ---@return WeaponDamageTable
-    RefreshDamageTable = function(self)
+    GetNewDamageTable = function(self)
+        local weaponBlueprint = self.Blueprint
+
+        local damageFriendly = weaponBlueprint.DamageFriendly
+        local damageTable = {
```
speed2CZ, line 485: I would very much like to see following if condition here:
```LUA
if damageFriendly == nil then
    damageFriendly = true
end
```
That is in the case that I understood that hell of the logic condition few lines below (that the default value is `true`)

Whoever wrote that one was rude to the reader :D (and should watch more of uncle bob)

ostrovaya: Sure! I was thinking that it was obscured in some way and it also took me some time to understand.

lua/sim/weapon.lua
speed2CZ: `__index` field is used inside metatables for a lookup of table keys when they're not present in the table. But the `__index` key **has to be defined inside a metatable**. `damageTable` doesn't look like a metatable to me in this code.

ostrovaya: Reading through the Lua docs though I would think that this line does... nothing? Happy to remove it if that's the case.

speed2CZ: Purge it

ostrovaya: I'll run it through the replay again and push it if it's the same. Which I would expect!

lL1l1: It's needed because the damage table is used by the projectile as a meta table.

https://github.com/FAForever/fa/blob/151673247fb0cdeaed29503b205ecbe0647c9581/lua/sim/Projectile.lua#L609-L612

This saves memory by keeping a small table in the projectile that just passes any nonexistent key references to the weapon damage table, while also allowing any damage modifications that happen within the projectile to happen when needed.

speed2CZ: Ok, its needed then, but Id say in a different form. The function name GetNewDamageTable gives no clue about such thing.
Should this logic for metatable be placed somewhere close if not at the place where the metatable is assigned, or make it in other ways clear about this intend?

lL1l1: Placing it where the metatable is assigned would create an unnecessary table operation every time a projectile is created, so that's definitely not the place to put it.

I don't think there is a better place to put it than in the one-time creation of the table. A comment should be sufficient to explain how it works: there's already a comment in the description of the WeaponDamageTable class and it can just be duplicated to the assignment in the function. I don't think you can miss how it works after that, whether you approach from the projectile side or the weapon firing side.

lL1l1 (approved):

PR #6862, opened 2025-06-16
Reduce table accesses of projectiles especially on impact

ostrovaya:

In `Projectile` there are a few places that benefit from using a `local` of constant values. 

In `OnImpact`, there are many table lookups that use values available outside of that table. The table itself is kept for functions that require it and is properly updated for that purpose.

A `local` is used for booleans that are recomputed for each emitter on impact when their evaluation doesn't change.

Uses some upvalues for functions that appear in a few places throughout the file.

### Replay simulation

Rebase changes on game version 3824 and simulate replay 25053165.

Observe no desyncs and unrelated warnings.

```
WARNING: Duplicate definition of console command ""
WARNING:  NUM PROPS = 5182
WARNING: GetResource: Invalid name ""
WARNING: GetResource: Invalid name ""
WARNING: GetResource: Invalid name ""
WARNING: ACU kill detected. Rating for ranked games is now enforced.
WARNING: FactoryRebuildUnits failed to rebuild correctly for factory urb0303 (entity ID 7340047).
WARNING: Rebuild data:
WARNING: Progress: 0.916706
WARNING: BuildTime: 23100.998047
WARNING: Health: 40667.328125
WARNING: 150
WARNING: FactoryRebuildUnits failed to rebuild correctly for factory zab9501 (entity ID 7341109).
WARNING: Rebuild data:
WARNING: Progress: 0.073846
WARNING: BuildTime: 47.999996
WARNING: Health: 26.107695
WARNING: 40
WARNING: FactoryRebuildUnits failed to rebuild correctly for factory zab9503 (entity ID 7341108).
WARNING: Rebuild data:
WARNING: Progress: 0.401329
WARNING: BuildTime: 3210.627930
WARNING: Health: 1304.484009
WARNING: 90
WARNING: FactoryRebuildUnits failed to rebuild correctly for factory zrb9603 (entity ID 1048668).
WARNING: Rebuild data:
WARNING: Progress: 0.186747
WARNING: BuildTime: 4706.027344
WARNING: Health: 5810.049805
WARNING: 150
WARNING: FactoryRebuildUnits failed to rebuild correctly for factory zrb9503 (entity ID 1048699).
WARNING: Rebuild data:
WARNING: Progress: 0.417000
WARNING: BuildTime: 3336.003662
WARNING: Health: 1252.000000
WARNING: 90
```

## Checklist

### DIFF
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -807,24 +807,25 @@ Projectile = ClassProjectile(ProjectileMethods, DebugProjectileComponent) {
         if not tbl then return end
         if not tbl.Radius then return end
         local radius = tbl.Radius
+        local category = tbl.Category
 
         self.MyFlare = Flare {
             Owner = self,
             Radius = radius,
-            Category = tbl.Category, -- We pass the category bp value along so that it actually has a function.
+            Category = category, -- We pass the category bp value along so that it actually has a function.
         }
         if tbl.Stack == true then -- Secondary flare hitboxes, one above, one below (Aeon TMD)
             self.MyUpperFlare = Flare {
                 Owner = self,
                 Radius = radius,
                 OffsetMult = tbl.OffsetMult,
-                Category = tbl.Category,
+                Category = category,
             }
             self.MyLowerFlare = Flare {
                 Owner = self,
                 Radius = radius,
                 OffsetMult = -tbl.OffsetMult,
-                Category = tbl.Category,
+                Category = category,
             }
             self.Trash:Add(self.MyUpperFlare)
             self.Trash:Add(self.MyLowerFlare)
--- a/lua/sim/Projectile.lua
+++ b/lua/sim/Projectile.lua
@@ -42,6 +42,7 @@ local EntitySetHealth = EntityMethods.SetHealth
 local EntityGetPositionXYZ = EntityMethods.GetPositionXYZ
 local EntityDestroy = EntityMethods.Destroy
 local EntityGetOrientation = EntityMethods.GetOrientation
+local EntityPlaySound = EntityMethods.PlaySound
 
 local TrashBag = TrashBag
 local TrashBagAdd = TrashBag.Add
@@ -461,9 +462,9 @@ Projectile = ClassProjectile(ProjectileMethods, DebugProjectileComponent) {
         -- Sounds for all other impacts, ie: Impact<TargetTypeName>
         local snd = blueprintAudio['Impact' .. targetType]
         if snd then
-            self:PlaySound(snd)
+            EntityPlaySound(self, snd)
         elseif blueprintAudio.Impact then
-            self:PlaySound(blueprintAudio.Impact)
+            EntityPlaySound(self, blueprintAudio.Impact)
         end
 
         -- Possible 'target' values are:
@@ -546,7 +547,7 @@ Projectile = ClassProjectile(ProjectileMethods, DebugProjectileComponent) {
             end
 
             if blueprintAudio then
-                self:PlaySound(blueprintAudio)
+                EntityPlaySound(self, blueprintAudio)
             end
         end
     end,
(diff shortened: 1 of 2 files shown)


lua/sim/Projectile.lua
```diff
@@ -367,43 +368,46 @@ Projectile = ClassProjectile(ProjectileMethods, DebugProjectileComponent) {
         local instigator = launcher or self
 
         -- localize information for performance
-        local vc = VectorCached
-        vc[1], vc[2], vc[3] = EntityGetPositionXYZ(self)
+        local vcx, vcy, vcz = EntityGetPositionXYZ(self)
```
ostrovaya, line 371: Given that the prior code also used this but didn't make use of the local `number` values, this seems fine.

lua/sim/Projectile.lua
```diff
@@ -367,43 +368,46 @@ Projectile = ClassProjectile(ProjectileMethods, DebugProjectileComponent) {
...
+                local velx, vely, velz = targetEntity.Owner:GetVelocity()
+                vcx = vcx + velx
+                vcy = vcy + vely
+                vcz = vcz + velz
             end
         end
 
+        -- localize information for performance
+        local vc = VectorCached
+        vc[1], vc[2], vc[3] = vcx, vcy, vcz
```
ostrovaya, line 392: `vc` doesn't change after this point

ostrovaya:
@Hdt80bro @speed2CZ @Basilisk3 I kindly request your review. It is appreciated as always.

lua/sim/Projectile.lua
```diff
@@ -803,29 +809,35 @@ Projectile = ClassProjectile(ProjectileMethods, DebugProjectileComponent) {
     AddFlare = function(self, tbl)
         if not tbl then return end
         if not tbl.Radius then return end
+        local radius = tbl.Radius
+        local category = tbl.Category
```
speed2CZ: Since all the values except the offset are the same for all 3 flares, why not to reuse the specs table for initialization? Like this:

```LUA
AddFlare = function(self, tbl)
    if not (tbl and tbl.Radius) then return end
    local flareSpecs = {
        Owner = self,
        Radius = tbl.Radius,
        Category = tbl.Category, -- We pass the category bp value along so that it actually has a function.
    }
    self.MyFlare = Flare(flareSpecs)
    TrashBagAdd(self.Trash, self.MyFlare)

    if tbl.Stack == true then -- Secondary flare hitboxes, one above, one below (Aeon TMD)
        local offsetMutl = tbl.OffsetMult
        flareSpecs.OffsetMult = offsetMutl
        self.MyUpperFlare = Flare(flareSpecs)
        TrashBagAdd(self.Trash, self.MyUpperFlare)

        flareSpecs.OffsetMult = -offsetMutl
        self.MyLowerFlare = Flare(flareSpecs)
        TrashBagAdd(self.Trash, self.MyLowerFlare)
    end
end,
```

You save both on accessing the values, and on allocating additional tables that are thrown away right away.

speed2CZ (changes requested):

lua/sim/Projectile.lua
ostrovaya: Even better - thanks. I'll amend the PR with this change.

ostrovaya:
@speed2CZ thank you for the feedback. I have amended the PR with your suggestion and an annotation.

speed2CZ (approved):

Garanas (commented):
A more general note may be what the purpose of the `VectorCached` upvalued global is, and whether it's not too risky to keep.

Originally I introduced this value* because I thought that using `entity:GetPosition()` would generate a new table on each call. I spoke about this with I think @The-Balthazar a year a go or so. And during that conversation it was mentioned that the table is not re-allocated each time it is called. 

Instead, the engine keeps a table reference. It updates it internally each time `entity:GetPosition()` is called. It then returns the reference.

If that is true then using the `VectorCached` upvalued global just introduces more risk with little to no benefit. It also makes the code much more error prone and more difficult to maintain.

I like the changes here where you compute more with locals where that is applicable. That is great! But maybe it's interesting to then ditch the globally shared vector table too (in a separate pull request?), and just make a call to `entity:GetPosition`. We can then update that accordingly, and re-use that.

Garanas (approved):

PR #6863, opened 2025-06-16
Fix adjacency error caused by dummy unit from drones/external factories

### LINKED ISSUE #6425: [Bug]: Frequent `OnAdjacentTo` error
### Describe the Bug

Originates from #6093. The new `ZXA0003` unit seems to cause a bug with adjacency since adjacency expects a `Unit.EntityId` but `DummyUnit` does not assign `self.EntityId`.
The line numbers below are modified to include my warning code of which adjacent unit doesn't have the entity ID, the correct line is 787 instead of 790.
```
warning: adjacent unit does not have self.EntityId: zxa0003
warning: Error running OnAdjacentTo script in Entity ueb0302 at 1b9ffe08: ...\documents\github\fa\lua\sim\units\structureunit.lua(790): table index is nil
         stack traceback:
         	...\documents\github\fa\lua\sim\units\structureunit.lua(790): in function <...\documents\github\fa\lua\sim\units\structureunit.lua:776>
```

### Reproduce the bug

Replay 23216422

### Screenshots

_No response_

### Additional context

_No response_

lL1l1:

Fixes #6425. 

- Adds the dummy unit OnCreate to zxa0003's script. This fixes EntityId being missing from the unit.
- Disable doing adjacency for dummy units since it can allocate many tables considering that it occurs for all UEF construction drones which may be flying around building many structures.

The t2 mass fab is placed here so that it triggers the adjacency for the dummy unit of the fatboy. Using debug logging it shows that the dummy unit is ignored correctly.
You can also trigger the adjacency by having the UEF engineer SACU build any structures.
```lua
   CreateUnitAtMouse('uel0301_engineer', 0,    1.93,   -0.48, -0.00000)
   CreateUnitAtMouse('uel0301_engineer', 0,   -2.00,   -0.48, -0.00000)
   CreateUnitAtMouse('uel0301_engineer', 0,   -3.32,   -0.48, -0.00000)
   CreateUnitAtMouse('uel0301_engineer', 0,   -4.84,   -0.48, -0.00000)
   CreateUnitAtMouse('uel0301_engineer', 0,   -0.72,   -0.48, -0.00000)
   CreateUnitAtMouse('uel0301_engineer', 0,    5.19,   -0.48, -0.00000)
   CreateUnitAtMouse('uel0301_engineer', 0,    3.34,   -0.48, -0.00000)
   CreateUnitAtMouse('uel0301_engineer', 0,    0.40,   -0.48, -0.00000)
   CreateUnitAtMouse('ueb1104', 0,   -2.32,    3.52,  0.00000)
   CreateUnitAtMouse('uel0401ef', 0,   -1.49,    2.89,  1.35140)
   CreateUnitAtMouse('uel0401', 0,    3.15,    3.92,  1.35140)
   CreateUnitAtMouse('xab1401', 0,    0.68,   -6.48, -0.00000)
```

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed errors involving units attached to external factories and engineering drones.
  - Improved construction pod attachment handling to prevent incorrect guard behavior.
  - Corrected unit targeting and spawning classifications for improved in-game behavior.
  - Updated attached-unit behavior so hidden units remain properly unselectable.

- **Diagnostics**
  - Added optional logging to help identify active unit states during gameplay troubleshooting.

- **Documentation**
  - Added a changelog entry detailing the attachment-related error fix.

### DIFF
--- a/lua/sim/units/StructureUnit.lua
+++ b/lua/sim/units/StructureUnit.lua
@@ -772,7 +772,9 @@ StructureUnit = ClassUnit(Unit, BlinkingLightsUnitComponent) {
     OnAdjacentTo = function(self, adjacentUnit, triggerUnit)
 
         -- make sure we're both finished building
-        if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt() then
+        if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt()
+            or adjacentUnit.Blueprint.CategoriesHash["DUMMYUNIT"]
+        then
             return
         end
 
@@ -832,7 +834,9 @@ StructureUnit = ClassUnit(Unit, BlinkingLightsUnitComponent) {
     OnNotAdjacentTo = function(self, adjacentUnit)
 
         -- make sure we're both finished building
-        if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt() then
+        if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt()
+            or adjacentUnit.Blueprint.CategoriesHash["DUMMYUNIT"]
+        then
             return
         end
 
--- a/lua/sim/units/StructureUnit.lua
+++ b/lua/sim/units/StructureUnit.lua
@@ -775,7 +775,6 @@ StructureUnit = ClassUnit(Unit, BlinkingLightsUnitComponent) {
         if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt()
             or adjacentUnit.Blueprint.CategoriesHash["DUMMYUNIT"]
         then
-            LOG('OnAdjacentTo return')
             return
         end
 
@@ -838,7 +837,6 @@ StructureUnit = ClassUnit(Unit, BlinkingLightsUnitComponent) {
         if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt()
             or adjacentUnit.Blueprint.CategoriesHash["DUMMYUNIT"]
         then
-            LOG('OnNotAdjacentTo return')
             return
         end
 
--- a/lua/sim/units/StructureUnit.lua
+++ b/lua/sim/units/StructureUnit.lua
@@ -775,6 +775,7 @@ StructureUnit = ClassUnit(Unit, BlinkingLightsUnitComponent) {
         if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt()
             or adjacentUnit.Blueprint.CategoriesHash["DUMMYUNIT"]
         then
+            LOG('OnAdjacentTo return')
             return
         end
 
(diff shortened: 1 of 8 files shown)


Garanas:
Great catch on the missing call to `OnCreate`.

I do not fully understand the context about UEF construction drones (from kennels), these generate calls to `OnAdjacentTo`? That would be unexpected, I thought that was a structure-only engine event.

lL1l1:
It is, but we have a "guard dummy" unit (zxa0003) which uses the `RULEUMT_None` motion type and is attached to the drone itself. When the drone finishes construction, the attached dummy unit can be in an ogrid where it is considered adjacent.

The dummy is used to save assist orders onto the drone after a drone docks.

lua/sim/units/StructureUnit.lua
```diff
@@ -772,7 +772,9 @@ StructureUnit = ClassUnit(Unit, BlinkingLightsUnitComponent) {
     OnAdjacentTo = function(self, adjacentUnit, triggerUnit)
 
         -- make sure we're both finished building
-        if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt() then
+        if self:IsBeingBuilt() or adjacentUnit:IsBeingBuilt()
+            or adjacentUnit.Blueprint.CategoriesHash["DUMMYUNIT"]
```
speed2CZ: By the suggested logic the dummy unit should not have adjacency.

Whats the reason to put this check into `Structure`?

Why not to add `OnAdjacentTo OnNotAdjacentTo` to the `Dummy` class that would return?

clyfordv: ^Second this suggestion. Keeping things modular means that any future changes/removals don't leave lingering logic in the base class. Additionally, as the original implementer of zxa0003 I'll say that I knew enough to get it working, but not enough to know all the details exhaustively. Somebody with more experience in the unit blueprints could remove the structure-ness of it and address the problem at the source (add a movement type, maybe?).

Garanas:
Now I understand the context more, the guard unit is introduced with: https://github.com/FAForever/fa/pull/6176

And it only applies to the construction pods of the (S)ACUs. The drones from Kennels do not use this class. They are 'regular' construction units.

lua/sim/units/StructureUnit.lua
lL1l1: > Why not to add OnAdjacentTo OnNotAdjacentTo to the Dummy class that would return?

The error pops up in the script of the structure that is adjacent. Returning in the dummy class does nothing to prevent the adjacent structure unit script from running.

> don't leave lingering logic in the base class

I agree, I'll try to find a different solution.

>  remove the structure-ness of it and address the problem at the source (add a movement type, maybe?).

I tried other motion types but the engine removes the assist order once the drone is attached to the SACU (and the dummy unit is finally attached indirectly to the SACU).
Fortunately, upon some investigation, it seems related to the "Attached" state of the unit, so we can do `self:SetUnitState("Attached", false)` in `OnAttachedToTransport` and then remove all the guard re-assignment behavior, since assist orders no longer get removed.

I haven't yet figured out how to deal with fatboy detaching its external factory when you give the dummy unit a motion type.

PR #6871, opened 2025-06-18
Increase the `StrategicIconSize` of the Seraphim ACU's OC projectile to match that of the other ACUs

Basilisk3:
As per the title.

## Checklist


Garanas (approved):
That makes sense, looks like a fixed oversight!

PR #6877, opened 2025-06-24
Integrate Saver's Aeon T2 -> T3 Shield upgrade mod

lL1l1:
Integrates the unit `uab42022` from the mod "AEON T2 to T3 Shild" created by [Saver](https://forum.faforever.com/user/saver) and @CDRMV. Approval was given by Saver in this [forum post](https://forum.faforever.com/post/66347), and in general that thread has the the history of the mod (variant 3 is being integrated).

The upgrade ability was already approved by the balance team in the past, it just needed an animation.

- Allow the Aeon T2 shield to upgrade to the T3 shield.
  - There is a new model and animation to fit the upgrade ability.
  - There is also AI support using the existing T2 shield upgrade platoon.

https://github.com/user-attachments/assets/191a74a1-9c6e-45c0-b7e5-32ea841a4ee2

I didn't find any errors or abnormalities when:
- The unit is upgraded.
- The unit is upgraded with assistance.
- The unit is killed while upgrading.
- The unit is destroyed while upgrading.
- The shield is enabled while upgrading.
- The shield is disabled while upgrading.

Specific visuals:
- The emitter effects line up with the model.
- The upgrade animation lines up with the T3 shield model at the end.
- It looks sufficiently realistic:
  - The mercy pools obscure any mesh clipping (this is the most important part). An improvement could be separating out the animated parts into a separate entity so that they can use the Aeon build shader.
  - A minor criticism is the unit lifting itself magically near the end of the animation. Maybe it could use the new legs it creates as levers to lift itself instead.
  - There is an issue in that the animation and the rotator of the orb conflict, so the orb has to stop spinning during the upgrade even though the shield is enabled. It might be an issue of the bone hierarchy.

@MadMaxFAF If during your review it seems possible, could you fix the animation issues?

@relent0r @Garanas Can you tell me how to test the AI behavior?

## Checklist


relent0r:
Will require a small change to the AI, I can take a look this week(on phone at the no).
If your in a rush the upgrade builder is here
https://github.com/FAForever/fa/blob/b84226246e8880adc606ac57e7584167c09ca7b9/lua/AI/AIBuilders/AIDefenseBuilders.lua#L991

There is a faction condition, you need to add faction 2 (I think that was Aeon).
I haven't looked at the platoon temple to make sure it doesn't restrict the aeon shield yet.

edit. Got in front of a computer.

So this condition gets a 2 added to it.

`{ MIBC, 'FactionIndex', {1, 4}},`

This platoon template

```
PlatoonTemplate {
    Name = 'T2Shield',
    Plan = 'UnitUpgradeAI',
    FactionSquads = {
        UEF = {
            { 'ueb4202', 0, 1, 'support', 'None' }
        },
        Seraphim = {
            { 'xsb4202', 0, 1, 'attack', 'None' }
        },
    }
}
```

Gets the following added

```
Aeon = {
    {'uab4202', 0, 1, 'support', 'None' }
},
```

I can see you've got the structureupgrades.lua file updated which the UpgradeAI uses for finding what to upgrade to.

I should get a bit of time tonight to test this and confirm.

relent0r:
Tested and the upgrade started and completed with the updates you made so all good.

image

relent0r (approved):
From an AI perspective this PR works fine. I didn't see any issues, it leverages existing known good AI builder logic.
Did not see any errors during game play.

Quick note, as I hadn't considered it. I'm unsure if this has any impact on campaign. I wouldn't think so .

PR #6879, opened 2025-06-27
Fix Cybran engis not turning wheels back to land mode

lL1l1:
## Issue
Reported by @4z0t on [Discord](https://discord.com/channels/197033481883222026/984529346498813992/1387915477078114529)
image

When cybran engis move from water onto land they do not turn their wheels back to normal.

Pass the status of being in the water to the transformation function only if the new layer is water.

Cybran engis transform to land mode after exiting from water onto land or onto a transport.

## Checklist

### DIFF
--- a/lua/sim/units/cybran/CConstructionUnit.lua
+++ b/lua/sim/units/cybran/CConstructionUnit.lua
@@ -113,7 +113,7 @@ CConstructionUnit = ClassUnit(ConstructionUnit, CConstructionTemplate) {
         end
 
         if old ~= 'None' then
-            self.TerrainLayerTransitionThread = TrashBagAdd(trash, ForkThread(self.TransformThread, self, true))
+            self.TerrainLayerTransitionThread = TrashBagAdd(trash, ForkThread(self.TransformThread, self, (new == 'Water')))
         end
     end,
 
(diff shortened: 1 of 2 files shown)


lua/sim/units/cybran/CConstructionUnit.lua
```diff
@@ -113,7 +113,7 @@ CConstructionUnit = ClassUnit(ConstructionUnit, CConstructionTemplate) {
         end
 
         if old ~= 'None' then
-            self.TerrainLayerTransitionThread = TrashBagAdd(trash, ForkThread(self.TransformThread, self, true))
+            self.TerrainLayerTransitionThread = TrashBagAdd(trash, ForkThread(self.TransformThread, self, (new == 'Water')))
```
Garanas, line 116: I'm not sure what the original behavior was. They do not apply the animation when they enter a transport:

image

We could have that if we turn it into `new ~= 'Land'` I think. But I'm fine either way.

lL1l1: This is the old code before the file separation.
https://github.com/FAForever/fa/blob/e56d9f7144d8c97120e306ab3ab3944555260570/lua/cybranunits.lua#L479

Garanas (approved):

PR #6881, opened 2025-06-28
Fix treads spawning for units that are on a transport

Garanas:

A bug was reported on [Discord](https://discord.com/channels/197033481883222026/1388492188887023677) that treads would spawn when units are attached to transports.

image

Spawn in units that create treads. See if their behavior on transports is correct.

## Additional context

When the transport is idle the layer of attached units is considered to be 'Land', even though they're quite a bit up in the air. This causes movement effects to spawn in.

https://github.com/user-attachments/assets/cd7b9780-52c0-4a8a-bd5f-9632837d4e57

## Checklist

### DIFF
--- a/changelog/fix.6881.md
+++ b/changelog/fix.6881.md
@@ -1,0 +1,1 @@
+- (#6881) Fix tread splats spawning for units that are attached to a transport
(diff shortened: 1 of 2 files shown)


Basilisk3 (approved):
Tested and works.

PR #6882, opened 2025-06-29
Fix bug where dummy weapon bps are assigned to non-dummy weapons

lL1l1:
## Issue
Discovered when I was helping a new modder on [Discord](https://discord.com/channels/197033481883222026/832710161847287819/1388716991124213762).

If a dummy weapon is before a non-dummy weapon in the unit blueprint's `Weapon` table, then the engine assigns the dummy weapon's bp to the non-dummy weapon. This is because the engine counts n real weapons, and simply assigns indices 1-n to weapons 1-n. If a dummy weapon is in 1-n then it gets assigned to a real weapon which breaks that weapon.

This is an easy issue to run into if using the `ModWeapon` field in blueprints, which by default places weapons at the end of the weapons table. 
It can also happen if someone doesn't know about this issue and places a dummy weapon in the middle by chance.

In post-mod weapon processing, detect if non-dummy weapons are placed after dummy weapons, and re-order the dummy weapons to the end of the array.

Also check for dummy weapons in the weapon merge function.

The error from the test commit disappears after applying the changes.

## Checklist


Garanas (commented):
Would it not be better to throw a warning and let the author fix it?

I'm not sure if re-ordering the weapons is the best thing to do here. It would make a unit 'work' for FAF, but not for Steam or LOUD for example.

lL1l1:
It already gives a warning. I agree that the author should learn and fix their mistakes, improving compatibility with other environments, if they do a custom implementation of the weapons table (as simple as writing it in a blueprint). But the `MergeWeaponByLabel` feature does need this behavior for it to be easy to use, so I'll keep it in that function.

lL1l1:
It's unlikely, but auto-fixing it could also break mods that hacked around the bug or fix old mods that had the bug. Would just have to see if there are any new warnings when I play.

Garanas:
Maybe @The-Balthazar has an idea how often that can happen in practice, based on his own sim mods and unit packs.

speed2CZ (approved):

The-Balthazar:
It probably doesn't help that `MergeWeaponByLabel` was written by someone who assumes all weapons have unique labels.
```lua
    for i, w in weaponTable do
        if w.Label == label then
            weaponTable[i] = BlueprintMerged(w, newBp)
            return
        end
```
Should probably not return there if you want a label merge to apply to each of a given label assuming merges are unique labels only, or it needs a full rewrite to keep a count of relative instances of given labels in the base vs the merge.

Did the dummies in this case have labels that matched other later indexed weapons?

lL1l1:
I addressed your concern in #7191.

The dummies didn't have labels that matched other weapons. The modder was adding a brand new weapon to the end of the UEF ACU's weapons table, but since the UEF ACU has a death nuke dummy weapon, that was causing the bug described in the PR.

PR #6883, opened 2025-06-29
Update blueprint modding documentation

lL1l1:
Responds to feedback from a [discord message](https://discord.com/channels/197033481883222026/832710161847287819/1388658835765465220) about the `ModWeapons` table.
> I guess for starters, it doesn't specify where you need to put the ModWeapon table at, whether it's an actual, full-blown table like the normal Weapons one, or if it's a nested table placed inside of it. I assumed the former, but that did through me off for a moment.  It also makes you think that MergeLabel is for both merging and adding weapons with the way it is worded. AddIndex is not very clear on what index number you should input and kind of assumes you know how the weapon table is structured already. Do you count weapons from top to bottom or from bottom to top? I know it's from top to bottom, but I'm just pointing that out. Some people need more help than others. Also, do Dummy and Teleport weapons count? And lastly, it suggests that it's useful to place weapons at index "1", which obviously didn't work out too well for me. Hope all these scatterbrained thoughts of mine helped. 

Also updates the blueprint scripting ("Balance Changes") section to include the new function `SetModBlueprintFunction`.

## Checklist

### DIFF
--- a/changelog/snippets/other.6883.md
+++ b/changelog/snippets/other.6883.md
@@ -1,0 +1,1 @@
+- (#6883) Update blueprint modding documentation in the game repository (MODS.LUA).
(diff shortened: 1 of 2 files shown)


Basilisk3 (approved):
Lgtm.

PR #6887, opened 2025-07-16
Make structure hp decay quickly below 10% completion

### LINKED ISSUE #6625: [Balance Idea]: Remove ACU blocking with unbuild PDs/Walls
### Intended Change

Currently it is possible to permanently block ACUs with unbuilt buildings, such as Tech 1 PDs. This should not be possible for several reasons:
- It can cause ACUs to be stuck inside the PD, making it almost impossible to escape
- It shifts the balance of the game towards making ACU pushes or small micro mistakes very punishing. This stands in contrast with a wide range of balance changes that have been made in the past to allow for more liberal usage of the ACU. See the changes to unit density etc.
- It abuses pathfinding problems to create an advantage
- It is very unintuitive and disruptive

**Possible workarounds:**
- Fixing it via code, for example by speeding up the rate at which PDs decay. Unfortunately this is likely handled by the engine. Introducing a limit on how often Tech 1 PDs can be selected might be reasonable to do in LUA.
- Making it an exploit

### Reasoning

Here is an example of this happening (thank you to @lL1l1 for posting it on Discord): 

https://cdn.discordapp.com/attachments/476200727912644618/1329258619711131740/wall_block_bm.mp4?ex=6789b00f&is=67885e8f&hm=aff848fbcfa4caf18ca68a4707eb18290cfd0b3172a96f1893c8af32563b8188&

Please note that with more training (and some other tricks I won't mention here) it can be even more effective than shown in the video.

### Expected behavior

Blocking ACUs with buildings is fine, blocking them permanently is a no-go.

lL1l1:
## Issue
Blocking with PD is considered unfun and has a bad aesthetic. This is because it can allow for cheap ACU snipes by getting enemy ACUs stuck in rapidly formed walls of unbuilt PD, which decay slowly unlike low cost walls (unit build progress decays at 0.1 resource per tick from the highest out of the mass/energy/buildtime requirements).

Implement a component that lets a unit start decaying quickly while it is below a given fraction complete.
 - The default decay rate is equivalent to how long walls would take to decay if built with the same buildpower.
 - The default completion threshold is equivalent to how long walls take to build if built with the same buildpower.

You can no longer easily block an ACU from running away with your own ACU by spamming unbuilt PDs (using two friendly ACUs to test works).

https://github.com/user-attachments/assets/1c966904-a9c4-4a41-9900-9ecc41ef7ce1

The ACU eventually gets out of the trap. It also requires some more skill than just spamming walls all over the place, since the buildpower investment matters significantly for how long the blocking pd stays alive.

## Checklist

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -364,8 +364,8 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
             if decayPerTick < 0 then
                 local decayedHp = 0
                 while self:IsBeingBuilt() do
-                    -- unit is not being actively built and is low completion
-                    if not self:IsUnitState("NoReclaim") and self:GetFractionComplete() <= 0.1 then
+                    -- unit is not being actively built and is low completion (10% for a t1 pd, less for very expensive buildings)
+                    if not self:IsUnitState("NoReclaim") and self:GetFractionComplete() * highestEcoStat <= 200 then
                         decayedHp = decayedHp + decayPerTick
                         self:AdjustHealth(self, decayPerTick)
                     else
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -353,6 +353,31 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
 
         -- Temporarily disable the unit's weapons when it is transferred to prevent bypassing the fire rate
         self:AddOnGivenCallback(self.OnGivenDisableWeapons)
+
+        TrashAdd(self.Trash, ForkThread(function()
+            local maxHealth = self:GetMaxHealth()
+            -- compensate for natural decay rate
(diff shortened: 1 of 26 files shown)


Garanas (commented):
It feels to me it is better if this is triggered from a blueprint property. That way it is more convenient to add or remove the feature. 

Why did you use `OnCreate` instead of `OnStartBeingBuilt` as a trigger to create the decay thread? By using `OnStartBeingBuilt` the logic is skipped for units that are created via script.

Great idea to use a separate component 👍

lL1l1:
> It feels to me it is better if this is triggered from a blueprint property. That way it is more convenient to add or remove the feature.

Good idea. It will also give the UI a way to accurately know what structures decay quickly, if someone wants to note it in the abilities section. I think I started editing the script directly because of unnecessary optimization 😅, just adding one category check in the base unit class won't hurt.

> Why did you use OnCreate instead of OnStartBeingBuilt as a trigger to create the decay thread? By using OnStartBeingBuilt the logic is skipped for units that are created via script.

I didn't come to mind at first; I'll try doing it with OnStartBeingBuilt.

Rhaelya (approved):
Changes work as intended

PR #6890, opened 2025-07-21
Nerf Cybran ACU nano regen cost

Isocinetic:
Increase all cost stats of Cybran nano by 33%

**Cybran ACU**
  - Nano regen: 
    - BuildCostEnergy 45000 -> 60000
    - BuildCostMass = 1500 -> 2000
    - BuildTime = 1500 -> 2000
    
## Additional context
https://discord.com/channels/197033481883222026/1385993698277331146


changelog/snippets/balance.6890.md
```diff
@@ -0,0 +1,7 @@
+(#6890) Increase cost of Cybran ACU nano upgrade by 33%
```
lL1l1: 1. There should be a short explanation for why the changes were made in terms of gameplay. It would also be nice to extend the explanation in the PR description as well, or at least copy over the short explanation (see my other PRs).
2. It needs to be a bullet point (add `- ` at the beginning).
```suggestion
- (#6890) Increase the cost of the Cybran ACU's nano upgrade by 33%. It is meant to help defend against rambo ACUs at early T3 stage, but the low cost made it come out in mid T2 stage where Cybran was powerful enough. It was especially powerful in comparison to the Seraphim ACU's nano upgrade, which had the same HP, worse regen, a higher 1800 mass cost, and no personal stealth.
```

changelog/snippets/balance.6890.md
```diff
@@ -0,0 +1,7 @@
...
+
+**Cybran ACU (URL0001)**
+  - FAF_SelfRepairSystem: 
+    - BuildCostEnergy 45000 -> 60000
+    - BuildCostMass = 1500 -> 2000
+    - BuildTime = 1500 -> 2000
```
lL1l1: 1. This part needs to be indented because of the linebreak on line 2. Here's an example of what I mean:
```
- (#)

desc
```
- (#)

desc
```
- (#)

  desc
```
- (#)

  desc
  
2. Snippets should use a player-friendly format (mass cost first) and player-friendly names.

```suggestion
  **Cybran ACU (URL0001)**
    - Nano-Repair System: 
      - Mass Cost: 1500 -> 2000
      - Energy Cost: 45000 -> 60000
      - Buildtime: 1500 -> 2000
```

Basilisk3:
This upgrade needs a nerf, but I think it would make more sense if it were made weaker instead of more expensive. With the statline you propose, Cybran Nano would be both weaker and more expensive than Sera Nano. 

This can work, but would also result in an overall Nano-balance that looks like this:
- UEF: weaker but cheap
- Seraphim: strong but expensive
- Cybran: strong but very expensive

**My proposal:**
|                   | UEF   | Cybran         | Seraphim |
| ----------------- | ----- | -------------- | -------- |
| +HP               | 1500  | 1500           | 2000     |
| Current +Regen            | 40    | 60             | 60       |
| **Updated +Regen** | 40    | **50 (or 45)** | 60       |
| Mass Cost         | 800   | 1500           | 1800     |
| Energy Cost       | 24000 | 45000          | 56000    |

Like this we would keep the theme of Seraphim being the "strong Nano faction" alive, while Cybran gets a cheaper, but now also appreciably weaker one.

lL1l1:
Since the issue is with early t3 stage instead of t2 stage, it would be better to increase the cost over lowering the stats. Increasing the cost has the additional benefit of making stealth more used instead of people rushing all the way to nano. The stats also need to be high enough to not die to an enemy upgraded ACU.

Also I'd like to comment on the "+HP" part that Cybran stealth-nano has equal HP with Sera nano (10000 + 2000 + 1500 vs 11500 + 2000).

Basilisk3:
> Since the issue is with early t3 stage instead of t2 stage

I disagree with this. The upgrade is almost always problematic (including when the ACU is dropped in the water) because:
1. The statline is simply over tuned.
2. You have the additional advantage of Stealth, which is more of a problem in the Tech 2 stage. Stealth also deserves a nerf.

> Increasing the cost has the additional benefit of making stealth more used instead of people rushing all the way to nano.

This is an interesting perspective, given how cheap and effective Stealth is. From my experience, and from the games I've watched recently, rushing all the way to Nano doesn't displace non-Nano builds meaningfully, as Nano arrives considerably later. I don't think this is problematic. Additionally, nerfing its statline will also give Stealth more space, simply because it will no longer be as worthwhile.

> Also I'd like to comment on the "+HP" part that Cybran stealth-nano has equal HP with Sera nano (10000 + 2000 + 1500 vs 11500 + 2000).

I didn't include Stealth to keep things comparable. While Stealth adds additional costs, you also get a lot of functionality in return, so I think it's fine to exclude from the table. It doesn't sway the analysis against Cybran Nano in any case.

Isocinetic:
> The upgrade is almost always problematic (including when the ACU is dropped in the water)

In my opinion Cybran torp com is pretty balanced. I would note that after Cybran nano nerf (whatever it will be) it might be a good idea to increase Cybran torp ACU salvo size while keeping same DPS

> Like this we would keep the theme of Seraphim being the "strong Nano faction" alive

There also can be an argument for keeping the theme of Cybran being "strong regen faction"

> * UEF: weaker but cheap
> * Seraphim: strong but expensive
> * Cybran: strong but very expensive

I do recognize that nerfing it regen-wise is an alternative, but I would rather call Cybran **very** strong and very expensive. Cybran nano has in-built stealth and more regen (+78 compared to +70 and +50) than other nanos. Having very strong and very expensive upgrade path makes sense. It's already kinda this way if you take into account the total cost of getting nano com, 650+1500 is more than 1800. The problem is that Cybran don't really pay enough premium for having that extra.

I don't think stealth needs nerf. Especially right now, as long as it's overshadowed by how op nano is, there is no need to touch it.

Basilisk3:
> In my opinion Cybran torp com is pretty balanced. 

It's still on the weaker side, the only reason it can sometimes be effective is that Nano is overtuned and the ACU can be used to absorb a lot of damage. Increasing the torpedo upgrade's salvo size is a sensible idea though, and something I also suggested some time ago.

> There also can be an argument for keeping the theme of Cybran being "strong regen faction"

You can make that argument, though I think Cybran being the "weaker ACU faction with stronger units" fit the game better than what we have now.

> Cybran nano has in-built stealth and more regen (+78 compared to +70 and +50) than other nanos. 

In my opinion, Cybran Nano should not simply be a worse version of Sera Nano. Sure Stealth is also a factor, though you can also make the argument that e.g., Seraphim can also get Regen Aura or Tech 2. Comparing the Nano upgrades directly is the most sensible option in my opinion.

The higher base regen is another can of worms, and if you ask me, increasing it to 18 was not necessary either.

> I don't think stealth needs nerf. Especially right now, as long as it's overshadowed by how op nano is, there is no need to touch it.

Stealth giving lots of HP is a leftover from the time period before Cybran Nano and just after its introduction. 

With its current statline, Stealth is easily one of the best upgrades in the game because you get an extremely useful ability no other ACU has, on top of 2000 HP, all while not even blocking any other important upgrade paths in the early game. If it's not built enough, that is simply a misplay. Although, speaking from my experience, it is not overshadowed by Nano.

I will make a counterproposal in another PR and I am also available in Discord, if you rather talk there.

Isocinetic:
> though you can also make the argument that e.g., Seraphim can also get Regen Aura or Tech 2

They get it in different upgrade slot, not in the same. This comparison doesn't apply unless the upgrade path is regen aura -> nano regen. As long as we're comparing same slot & same type upgrade that's a pretty direct comparison.

Basilisk3:
I disagree, but let's assume this gets implemented like proposed in this PR:
- Cybran now has the option to get a really expensive Nano upgrade, that is also by far the worst of the bunch.
- Cybran retains its extremely strong and unique Stealth upgrade, which is a no-brainer to build shortly or directly after gun.

What are the immediate effects of this? Cybran Nano gets pushed to the early Tech 3 stage, because it is now the most expensive Nano upgrade while also being the worst, and requiring Stealth beforehand.

Factoring in the context of Stealth and higher base regen is nice, but the upgrade's statline itself also has to make sense, otherwise there is no point in having the option altogether. Adding new upgrades to ACUs is delicate, and even this relatively straightforward addition is/was far from an uncontroversial decision. The upgrade is too strong currently, but a 33% nerf is excessive, and it also doesn't need to be pushed to the early Tech 3 stage.

lL1l1:
Stale and alternative nerf to regen was implemented #6894

PR #6893, opened 2025-07-31
Make Yolona outer damage ignore shields

lL1l1:
## Issue
Yolona's outer ring damage (7500) is lower than T3 shield HP, so it gets fully absorbed and deals no damage.

Making the ring damage deal damage to shields is not viable because ring damage is repeated as the nuke shockwave travels, so it would deal many instances of 7500 damage to shields.

Add a new damage type for the Yolona that ignores shields.

A new damage type keeps backwards compatibility with nuclear repulsor shields because the `SlowNuke` used by Yolona overrides the damage type to `Nuke` in those mods (they hook `NukeDamage.lua` and replace the lua `DamageArea` with the engine `DamageArea`).

Switching normal nukes to the new damage type would break those mods.

Nuke the paragon using the outer damage and confirm that it ignores the shield.

{4FCA0683-918E-494C-8A15-4FB1D4DA84A6}

```
   CreateUnitAtMouse('xab1401', 0,   -7.75,  -31.75, -0.00000)
   CreateUnitAtMouse('xab1401', 0,   12.25,   23.25, -0.00000)
   CreateUnitAtMouse('xsb4301', 0,   -7.75,  -13.75, -0.00000)
   CreateUnitAtMouse('xsb2401', 0,    3.25,   22.25, -0.00000)
```

## Checklist

### DIFF
--- a/lua/ShieldAbsorptionValues.lua
+++ b/lua/ShieldAbsorptionValues.lua
@@ -35,5 +35,6 @@ shieldAbsorptionValues = {
 	["StaticShield"] = { -- For mod support, auto-assigned if the owner has STRUCTURE category.
 		["Deathnuke"] = 1.0,
 		["Overcharge"] = 1.0,
+		["NukeIgnoreShields"] = 0.0,
 	},
 }
--- a/lua/armordefinition.lua
+++ b/lua/armordefinition.lua
@@ -20,6 +20,7 @@
 ---| "FireBeetleExplosion"
 ---| "Normal"
 ---| "Nuke"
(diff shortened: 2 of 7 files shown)


BlackYps:
Doesn't this look like intended behaviour that the shield is capable of absorbing damage that is less than the shield HP?

lL1l1:
Standard nukes bypass shields for balance reasons, including the miniscule outer damage, hence why it's a bit unintuitive that Yolona's outer damage is blocked by shields.

PR #6900, opened 2025-08-08
Remove the Seraphim Lightning Tank's `FiringTolerance`

Basilisk3:
Beam weapons should not have firing tolerances. I assume the firing tolerance could have caused the unit to miss aircraft occasionally, but this is next to impossible to test offline. There should be no adverse consequences to removing it, so it's a best practice to do so.

## Checklist


Isocinetic:
While it's impossible to confirm that **all** misses are due to firing tolerance, we can be quite sure that [**some** misses](https://discord.com/channels/197033481883222026/1392925692169031822) are caused by it, so that's a good thing to fix.

lL1l1 (approved):

PR #6905, opened 2025-08-16
Fix an edge case where announcements would break cinematics

Garanas:

Fix an edge case where an announcement would break.

There's no guarantee that the control is layout before the animation starts. These values act as a 'last known place'. Initializing with 0 means that if the control is never valid (destroyed immediately, for example) the animation will originate from the top-left corner.

Start the campaign map Black Day and watch it successfully finish the cinematic.

## Additional context

Related pull requests:

- https://github.com/FAForever/fa/pull/6845 (where the bug occured)
- https://github.com/FAForever/fa/pull/6813 (re-implementation of announcements to make it more flexible)

With thanks to @lL1l1 for his time and effort to help find this.

## Checklist

### DIFF
--- a/changelog/snippets/fix.6905.md
+++ b/changelog/snippets/fix.6905.md
@@ -1,0 +1,1 @@
+- (#6905) Fix an edge case where announcements would break cinematic mode
--- a/lua/ui/game/announcement/AbstractAnnouncement.lua
+++ b/lua/ui/game/announcement/AbstractAnnouncement.lua
@@ -300,11 +300,12 @@ AbstractAnnouncement = ClassUI(Group) {
         ---@type LazyVar
         local animationProgress = CreateLazyVar(0)
 
-        -- use last known position of control if it is destroyed, this happens when you fail an objective.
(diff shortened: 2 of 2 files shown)


lL1l1 (approved):

lL1l1:
The cause of the bug was that the goal control for the animation is initialized before the announcement, but laid out only after the announcement was created.

https://github.com/FAForever/fa/blob/07024ed605a971ed4a5d976538a35cc6cec4873c/lua/ui/game/objectives2.lua#L426-L437

BlackYps:
> Start the campaign map Black Day and watch it successfully finish the cinematic.

How do I actually do that?

Garanas:
@BlackYps The `init_coop.lua` file works very similar to the standard init file for development. At the top of the file you can replace the two `locationOfRepositoryX` with the paths to the repositories on your machine. Then make a copy of the bat file that you use to launch the development environment. Update the reference to the init file to use in the copy.

See also:

- https://github.com/FAForever/fa-coop/blob/master/init_coop.lua

PR #6915, opened 2025-08-27
Fix weapons with fire rate modifiers getting stuck waiting for energy drain

lL1l1:
## Issue
Weapons with too high fire rate bonus can get stuck waiting for their energy drain because the fire rate isn't considered in the drain/s of the economy event.
This issue arose in the past with the Salvation. It was fixed previously by increasing the drain in the blueprint, but I found it again in a mod because mods use old blueprints to override everything.

Increase energy drain/s proportional to fire rate changes.

Spawn a UEF T3 artillery and 9 adjacent T3 pgens (use spawn menu's unit count selection to stack them). It should fire when the work progress bar is full instead of pausing for a few seconds (I assume that's a target check interval).

## Checklist

### DIFF
--- a/changelog/snippets/fix.6915.md
+++ b/changelog/snippets/fix.6915.md
@@ -1,0 +1,1 @@
+- (#6915) Fix weapons with fire rate modifiers getting stuck waiting for energy drain.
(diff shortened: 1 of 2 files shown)


Basilisk3 (approved):
Tested and works.

PR #6931, opened 2025-09-21
Fix missing unit param for SpecificUnitsInArea objective type

Dhomie:
## Issue
Normally when an objective has reached some kind of result (success, failure), depending on its type, it provides the result, and additional data for callback functions to use, like units.
The `SpecificUnitsInArea` type however is missing that extra data, specifically the destroyed unit if it triggers objective failure.
Right now this oversight breaks the Seraphim mission **Tha-Atha-Aez**, as the callbacks tries to reference the `unit`, which is just `nil`, and basically softlocking the death camera mission failure cinematics. The game has to be closed.

Simply add the `unit` as a parameter for `OnResult` when it's called via the unit's death.
Not only it fixes the map (mostly), it also prevents the same errors for future missions where mission developers want to use the unit via objective callbacks.

The mission Tha-Atha-Aez is relevant for this.
The first 2 allied ACU escort mission failures should trigger proper.
The 3rd however has an error in the map script, there's no `unit` param defined in the `AddResultCallback` section, so it's map specific.

### DIFF
--- a/lua/SimObjectives.lua
+++ b/lua/SimObjectives.lua
@@ -859,7 +859,7 @@ function SpecificUnitsInArea(Type, Complete, Title, Description, Target)
         total = total - 1
         if objective.Active and total < numRequired then
             objective.Active = false
-            objective:OnResult(false)
+            objective:OnResult(false, unit)
             UpdateObjective(Title, 'complete', 'failed', objective.Tag)
             KillThread(watchThread)
         end
--- a/lua/shared/components/DebugComponent.lua
+++ b/lua/shared/components/DebugComponent.lua
@@ -35,7 +35,7 @@ DebugComponent = ClassSimple {
     EnabledLogging = true,
     EnabledWarnings = true,
     EnabledErrors = true,
-    EnabledDrawing = true,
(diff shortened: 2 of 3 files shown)


4z0t:
Undo file changes unrelated to resolved issue.

4z0t (dismissed):
Undo changes in unrelated files.

Dhomie:
I haven't the slightest idea what those changes are.
When I synced my FAF fork, it was simply there.
Edit: Issue lied in my fork using the `deploy/fafdevelop` branch as the default.
Anyhow, I undid those changes.

Garanas:
@Dhomie The base branch is `develop`. It used to be `deploy/fafdevelop` for a long time. The commit originates from the [deployment workflow](https://github.com/FAForever/fa/actions/workflows/deploy-fafdevelop.yaml). It helps us identify what game version and what specific git commit the user is running when they share a game log.

For future pull requests, start from `develop` 😄 !

Garanas (approved):
Change seems fine to me. I think it is backwards compatible because of how Lua works with function arguments.

PR #6932, opened 2025-09-24
Adjust default frametime to not interfere with vsync

Garanas:

Adjusts the default frame time that was introduced with https://github.com/FAForever/fa/commit/05610664912efb980c5566752488f6568cf4de11. It now defaults to 8 instead of 16. This prevents it from interfering with vsync behavior. This was reported on the forums:

- https://forum.faforever.com/topic/9520/supreme-commander-forged-alliance-game-not-utilizing-gpu./10

Observe that the changes take place when editing the settings. Unfortunately, I do not observe the behavior myself.

## Checklist
- ~~Changes are annotated, including comments where useful~~

### DIFF
--- a/changelog/snippets/fix.6932.md
+++ b/changelog/snippets/fix.6932.md
@@ -1,0 +1,1 @@
+- (#6932) Adjust default frametime to not interfere with vsync
--- a/changelog/snippets/fix.6932.md
+++ b/changelog/snippets/fix.6932.md
@@ -1,1 +1,1 @@
-- (#6932) Adjust default frametime to not interfere with vsync
+- (#6932) Adjust default maximum frame time to not interfere with vsync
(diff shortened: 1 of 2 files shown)


lL1l1 (commented):
I don't see any mention of vsync in the forum post. Did you mean that reducing the frametime somehow removes stutters? Since the poster mentioned that they had 60+/-40 fps and increasing fps fixed it.

Garanas:
That is my suspicion. Vsync forces your framerate to be either 60 or 30 to prevent tearing.

If we set our frame time to 16 then there's very little room for overhead. I'm not sure what type of overhead we are talking about here. It depends on what the engine does inside the desired frame time and what it does outside of it. Maybe for example garbage collection is done outside of this frame time. That could force the framerate from 58 to a flat 30 for a short time when vsync is enabled.

Either way, what happens outside of it can have a significant impact since a few milliseconds is the difference between a good frame rate and one that is uncomfortable to play.

lL1l1 (approved):

PR #6942, opened 2025-10-13
Annotate `OnMouseButtonPress` and related code

lL1l1:
## Description of the changes
Annotate `OnMouseButtonPress` and related code

read output from clicking with left/right/middle mouse buttons after using this command
```
local ui = import("/lua/ui/uimain.lua")
local f = function(event)
	reprsl(event)
end
ui.AddOnMouseClickedFunc(event)
ForkThread(function()
	WaitSeconds(5)
	ui.RemoveOnMouseClickedFunc(f)
end)
```

## Checklist

### DIFF
--- a/changelog/snippets/other.6942.md
+++ b/changelog/snippets/other.6942.md
@@ -1,0 +1,1 @@
+- (#6942) Annotate the UI function `OnMouseButtonPress` and related code.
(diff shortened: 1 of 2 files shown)


4z0t:
Perhaps creating an alias for this function type would be nice.

4z0t (approved):

lL1l1:
I don't think this type of function should appear as a param anywhere else, so it should be ok without an alias.

PR #6943, opened 2025-10-14
Add annotations about the sensitivity of the guard scan radius

Garanas:

Through the discussion in https://github.com/FAForever/fa/pull/6939 I noticed that a lot of relevant information surrounding this blueprint value is lost. With these annotations I hope to prevent this loss of information for the foreseeable future.

## Checklist

### DIFF
--- a/engine/Core/Blueprints/UnitBlueprint.lua
+++ b/engine/Core/Blueprints/UnitBlueprint.lua
@@ -145,7 +145,7 @@
 ---@field GuardRadius number
 --- maximum range from the guarded unit before initiating return
 ---@field GuardReturnRadius number
---- Guard scan range for the unit. The guard scan radius is used applied for attack move and patrol orders. Through this scanning mechanic the patrolling (or attack moving) unit can find a target to chase. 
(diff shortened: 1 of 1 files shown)


engine/Core/Blueprints/UnitBlueprint.lua
```diff
@@ -145,7 +145,15 @@
 ---@field GuardRadius number
 --- maximum range from the guarded unit before initiating return
 ---@field GuardReturnRadius number
---- guard range for the unit, automatically added if absent
+--- Guard scan range for the unit. The guard scan radius is used by attack move and patrol orders. Through this scanning mechanic the patrolling (or attack moving) unit can find a target to chase. The scanning behavior includes neutral units.
```
lL1l1, line 148: There could be another line mentioning how it is automatically set in blueprint loading to reasonable values based on the unit's weapon ranges with a limit based on tech tier.

Garanas: Processed in https://github.com/FAForever/fa/pull/6943/commits/316acaf91a715ed42d748b79ba801bbcfa8b412e

lL1l1 (approved):

PR #6957, opened 2025-11-05
Fix ravager size

maudlin27:
Buliding blueprints include the SIZEx category, with SIZE4 being for units like T2 PD and T1 pgens, SIZE12 being T2 pgen sized units, etc.; in all cases I'm aware of the size reflects the building's UI size, except the ravager which shows as a SIZE12 building instead of a SIZE4 building.

This fixes the blueprint category

As a basic test I created a T3 UEF engineer and checked it could build a ravager.

## Summary by CodeRabbit

* **Bug Fix**
  * Corrected the size category for a ravager unit to ensure its classification is accurate (no gameplay behavior changes).
* **Chores**
  * Added a changelog entry documenting the size category fix.

### DIFF
--- a/changelog/snippets/fix.6957.md
+++ b/changelog/snippets/fix.6957.md
@@ -1,0 +1,1 @@
+- (#6957) Fix the ravager blueprint size category
--- a/units/XEB2306/XEB2306_unit.bp
+++ b/units/XEB2306/XEB2306_unit.bp
@@ -18,7 +18,7 @@ UnitBlueprint{
         "PRODUCTFA",
         "RECLAIMABLE",
         "SELECTABLE",
-        "SIZE12",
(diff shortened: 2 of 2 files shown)


4z0t (approved):

The-Balthazar:
This is does affect gameplay related to adjacency. It was set larger than it should be so that it gets less bonuses than it otherwise would. Notable for rate of fire from T1 generators.

Basilisk3:
The Ravager doesn't gain an increased rate of fire from pgens, so I think this is fine to include.

The-Balthazar:
Oh right, RoF adjacency is for artillery only at size4, and I guess it
never hit any important breakpoints for RoF at size12. So just the minor
energy recharge cost then.

PR #6959, opened 2025-11-10
Add 'Salvation' to notify's default messages

Gatsik:
I noticed that there's no notify messages for 'Salvation' in the replay, but I'm not sure if this is all that's needed to add it

## Summary by CodeRabbit

* **New Features**
  * Added support for the experimental unit "Salvation" with display-name recognition and integration into Notify's default messages.

* **Documentation**
  * Added a changelog entry noting inclusion of "Salvation" in Notify's default messages.


lL1l1 (approved):
Tested and works: salvation message appears in chat and the notify options menu.

PR #6960, opened 2025-11-10
Report recall outcome to server

Hdt80bro:
Reports recalling as an army outcome to the server. Dependent on [this PR](https://github.com/FAForever/server/pull/1073) to be deployed first.


Hdt80bro:
We'll pivot our approach to use `EnforceRating` instead.

PR #6964, opened 2025-11-13
Remove time requirement for rating games when recalling

Hdt80bro:
The time requirement for ranking custom games is in place so that connection issues can be discovered early in the match and rehosted without penalty to rating. However, if a team recalls during that time frame, it is safe to presume that it was not due to any connection issues and they actually do want to forfeit the match.

This PR using the `EnforceRating` system used to do something similar for ACU kills but upon recalling. It also changes the log warning to reflect that better.

Replaces #6960

## Summary by CodeRabbit

* **Updates**
  * Updated messaging regarding time requirements for rating games
  * Adjusted notifications for game recall events to reflect current rating enforcement policies

### DIFF
--- a/lua/aibrain.lua
+++ b/lua/aibrain.lua
@@ -574,6 +574,9 @@ AIBrain = Class(FactoryManagerBrainComponent, StatManagerBrainComponent, JammerM
         -- TODO: create a common function for `OnDefeat` and `OnRecall`
         self.Status = "Recalled"
 
+        Sync.EnforceRating = true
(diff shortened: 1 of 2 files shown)


Isocinetic:
It's supported by various players' complaints/requests ([1](https://discord.com/channels/197033481883222026/1409256374885552188), [2](
https://discord.com/channels/197033481883222026/1428141018297012265) discord threads) too, so this is a good update.

BlackYps (approved):
I guess we can only really test this once it is deployed

PR #6965, opened 2025-11-13
Simplify defeat backend

Hdt80bro:
This PR refactors the AIBrain's defeat code. The main change is moving the unit destruction logic to be all in one spot, in `SimUtils.lua`, no matter how the brain was defeated.

In the process, in was discovered that the armies are handled differently in a recall (it transferred only built units before the 10 seconds wait - it now also transfers unbuilt units, and after the 10 grace period like in a regular defeat), so that has now been standardized. `TransferUnitsToKiller` was also using the incorrect callback in one particular case.

## Summary by CodeRabbit

* **Bug Fixes**
  * Defeat/recall/disconnect flows now transfer units consistently and behave more reliably at end of game.

* **New Features**
  * Utilities for safer unit transfers, timed waits, and targeted handling of unsafe/abandoned commanders.
  * Kennel pause/upgrade and unit-upgrade behaviors for transferred units.
  * Lobby option types refined for clearer share/disconnect choices.
  * End-of-game delay reduced to 1 second (configurable constant).

* **Refactor**
  * Centralized defeat/recall state and consolidated army/unit management paths.

* **Documentation**
  * Added changelog entry documenting the fix.


lua/SimUtils.lua
```diff
@@ -1052,25 +1075,46 @@ function KillArmy(self, shareOption)
...
+-- When using the delayed recall disconnect option, commanders will be shared at
+-- least until this time in game has passed
+MinimumShareTime = 5 * 60 * 10 -- 5 minutes
 
 --- Shares all units including ACUs. When the shared ACUs die or recall after `shareTime`, kills my army according to the given share condition.
 ---@param self AIBrain
----@param shareOption 'FullShare' | 'ShareUntilDeath' | 'PartialShare' | 'TransferToKiller' | 'Defectors' | 'CivilianDeserter'
----@param shareTime number Game time in ticks
+---@param shareOption ShareOption
+---@param shareTime integer Game time in ticks
```
lL1l1, line 1203:
```suggestion
---@param shareTime integer # Game time in ticks
```

lua/SimUtils.lua
```diff
@@ -1163,19 +1208,85 @@ function KillArmyOnACUDeath(self, shareOption)
...
+
+        -- Only handle Assassination victory, as in other settings the player is unlikely to be defeated soon
+        if victoryOption == 'demoralization' and table.empty(safeCommanders) then
+            shareOption = ScenarioInfo.Options.Share
+        end
+
+        -- non-assassination modes can have armies abandon without commanders
+        if shareAcuOption == 'Recall' and not table.empty(safeCommanders) then
+            -- KillArmy waits 10 seconds before acting, while FakeTeleport waits 3 seconds, so the ACU shouldn't explode.
+            FakeTeleportUnits(safeCommanders, true)
```
lL1l1: No longer forking the teleport as a thread means it always recalls the ACU before possibly killing it, so the comment isn't needed.

It also means that the grace period for an abandoned army is 13 seconds instead of 10. In my opinion that is reasonable.

lL1l1 (commented):
Looks mostly good to me. Needs a snippet and after you unset the draft status I can approve it.

lua/SimUtils.lua
Hdt80bro: As I understand, the majority of the repo doesn't use this optional syntax

lua/SimUtils.lua
Hdt80bro: Ah, thanks, I didn't notice, but that is reasonable indeed - it mirrors the recall behavior where the defeat doesn't happen until the recall finishes (which uses the same function), except that the defeat state is set first.

lua/SimUtils.lua
```diff
@@ -740,63 +740,67 @@ end
 local CalculateBrainScore = import("/lua/sim/score.lua").CalculateBrainScore
 local FakeTeleportUnits = import("/lua/scenarioframework.lua").FakeTeleportUnits
 
----@param owner number
--- categoriesToKill is an optional input (it defaults to all categories)
+local defaultTransferCategory = categories.ALLUNITS - categories.WALL - categories.COMMAND
+-- only units in this category will be shared during a partial share
```
lL1l1: "will be shared under partial share rules" is better

lua/SimUtils.lua
```diff
@@ -740,63 +740,67 @@ end
...
     end
-    if not categoriesToKill then
-        sharedUnits[owner] = {}
-    end
 end
 
 --- Given that `deadArmy` just died, redistributes their unit cap based on the scenario options
 ---@param deadArmy integer
 function UpdateUnitCap(deadArmy)
     -- If we are asked to share out unit cap for the defeated army, do the following...
```
lL1l1: unnecessary comment

lua/SimUtils.lua
```diff
@@ -740,63 +740,67 @@ end
...
     for index, brain in ArmyBrains do
-        if (mode == 'all' or (mode == 'allies' and IsAlly(deadArmy, index))) and not ArmyIsCivilian(index) then
-            if not brain:IsDefeated() then
-                aliveCount = aliveCount + 1
-                alive[aliveCount] = brain
-                local cap = GetArmyUnitCap(index)
-                caps[aliveCount] = cap
-            end
+        if not ArmyIsCivilian(index) and not brain:IsDefeated() and
+            (modeAll or IsAlly(deadArmy, index))
```
lL1l1: "shareToAll" would be a better variable name

lua/SimUtils.lua
```diff
@@ -1023,7 +1026,7 @@ local function TransferUnitsToKiller(self)
         end
 
         if killerIndex then
-            TransferUnitsToBrain(self, { ArmyBrains[killerIndex] }, true, nil, "TransferToKiller")
+            TransferUnitsToBrain(self, { GetArmyBrain(killerIndex) }, true, nil, "TransferToKiller")
```
lL1l1: Both of these work; is global table access or an engine function call better?

lua/SimUtils.lua
```diff
@@ -1086,21 +1088,61 @@ function KillRecalledArmy(self, shareOption)
...
 
     KillRemaining(self)
 end
 
+--- Blocks the current thread until all units in a list are dead, or until an
+--- optional timeout. Returns total ticks elapsed, or the original timeout
+--- (which could have been negative or fractional) if it was reached after
+--- checking the units. If all units are dead upon calling, returns `0`.
+---@param units Entity[]
+---@param timeout? integer in ticks
```
lL1l1: `timeoutTicks` would be better

lua/SimUtils.lua
```diff
@@ -1086,21 +1088,61 @@ function KillRecalledArmy(self, shareOption)
...
+---@return integer elapsed
+function ThreadSuspendUntilUnitsDead(units, timeout)
```
lL1l1: `WaitUntilUnitsDeadOrTimeout` would be better because "thread suspension" is closer to what's going on in shield.lua.

lua/SimUtils.lua
lL1l1: It would be nice to clean up the comment style for annotations (using a # or not) in a separate pr then, though that would be low priority.

lua/aibrain.lua
lL1l1: The variable is no longer used and can be removed (it is quite new).

lua/aibrain.lua
```diff
@@ -8,14 +8,10 @@
...
-local TransferUnfinishedUnitsAfterDeath = import("/lua/simutils.lua").TransferUnfinishedUnitsAfterDeath
+local KillAbandonedArmy = import("/lua/simutils.lua").KillAbandonedArmy
 local KillArmy = import("/lua/simutils.lua").KillArmy
-local KillArmyOnDelayedRecall = import("/lua/simutils.lua").KillArmyOnDelayedRecall
-local KillArmyOnACUDeath = import("/lua/simutils.lua").KillArmyOnACUDeath
+local KillRecalledArmy = import("/lua/simutils.lua").KillRecalledArmy
 local DisableAI = import("/lua/simutils.lua").DisableAI
-local TransferUnitsToBrain = import("/lua/simutils.lua").TransferUnitsToBrain
-local TransferUnitsToHighestBrain = import("/lua/simutils.lua").TransferUnitsToHighestBrain
 local UpdateUnitCap = import("/lua/simutils.lua").UpdateUnitCap
```
lL1l1: Simutils gets imported 5 times so maybe it should be in a variable.

lL1l1 (commented):
Some testing using the launch script and 4 players (fullshare, assassination):
- Defeat through recalling seems fine, although I know that in the live game that it kills units before the sim stops.
- Defeat through ctrl k seems fine
- Defeat through abandonment kills units before the sim stops. 

There is a ~109 tick difference between `AbandonedByPlayer` being called and the game ending, which varies due to the ACU kill and victory condition running on different threads. The difference in recall between here and live game is probably due to the additional 3 seconds to the grace period.

The ideal solution would be to make recall and abandon work sequentially with the victory condition thread, but for now the grace period can be increased.

lua/SimUtils.lua
lL1l1: Ran a simple benchmark, even a global is 52x faster.

lL1l1 (approved):

Hdt80bro:
Your changes look good to me, here's my recap so we're on the same page:
* Your local variables name are definitely better than the ones I was able to come up with.
* The function name `ThreadSuspendUntilUnitsDead` was intended to mirror the other thread functions like `ThreadSuspend` and `ThreadResume` but reading it back, it does make more sense to mirror the higher level functions like `WaitSeconds` (which it itself does call) instead. I'm torn on whether the `OrTimeout` part is necessary in `WaitUntilUnitsDeadOrTimeout` as it's an optional parameter that most functions would have; I'll defer to your judgment here as you have better names :)
* Comment updates are good; the one you removed was one I copied. There were several like that I deleted, but missed that one.
* I changed `ArmyBrains[]` to `GetArmyBrains()` purely for consistency knowing it was slower, since it looked like the previous author preferred it and it wasn't being called in performance critical paths; changing all of them back is fine.
* Compacting the imports doesn't affect performance since it's called literally once, but it is more readable, so I like that.
* We really should establish commenting best practices. I'll bring it up in Discord, though it's unlikely anyone will have strong opinions...

At a glance, increasing the grace period to 12 seconds seems curious to me because even 10 seconds already seems too long to me for players to have to wait for. If these changes really started blowing the units too soon, I think we should decrease the victory monitor's wait interval instead.

lL1l1:
> I'm torn on whether the `OrTimeout` part is necessary in `WaitUntilUnitsDeadOrTimeout` as it's an optional parameter that most functions would have; I'll defer to your judgment here as you have better names :)

I think it's nice to have in the name because as you said, most functions would have it. It also makes more clear what the options are for using the function.

> At a glance, increasing the grace period to 12 seconds seems curious to me because even 10 seconds already seems too long to me for players to have to wait for. If these changes really started blowing the units too soon, I think we should decrease the victory monitor's wait interval instead.

There's a 5 second delay that checks for draws, and the wait interval is already 4 ticks (5 in code but WaitTicks is weird). There's `DelayBeforeGameEnds` at 3 seconds, but it doesn't seem to do anything gameplay-wise since it comes after victory is already declared, so I'll reduce that by 2 seconds instead. I agree that the grace period should be minimized, although there is a discussion to be had considering the reward of killing a player and being able to abuse their idle units for 10 seconds (or however long). Which, if official, means there should be a patch for giving units shortly after dying before the victory condition detects defeat.

lL1l1:
Another way to get all units to blow up is to have a player die before the game ends, and then have their last teammate die in the middle of the transfer. That could be another common scenario in games since one player leaving or dying can result in the rest of their team leaving shortly after. I think it's better to close up the PR though, this is "simplify defeat backend", not adding some mechanism that integrates army transfer on defeat and the overall match state.

lL1l1 (approved):

Hdt80bro:
Yeah, that will have to wait for another time.
Everything looks good to me then!

PR #6966, opened 2025-11-13
Revert T1 interceptor and Swiftwind StartTurnDistance 10 -> 5

lL1l1:
## Issue
After #5042, interceptor micro feels bad due to the unintuitive nature of the huge `StartTurnDistance`, which is the distance within which aircraft will hover instead of turn to move to a waypoint.
1. it's unintuitive for a winged aircraft to hover
2. most air have a distance of 5, so micro skills do not transfer properly.
3. there is no visual indicator for the distance so move orders meant to micro turns end up hovering inties, especially when large numbers of inties are in formation

[Related discord post.](https://discord.com/channels/197033481883222026/1438677366925688994)

Revert the `StartTurnDistance` from 10 to 5 for inties and swiftwinds.

- Inties now usually try to turn when I want them to by giving short nearby move orders.
- Inties still dock at air staging in a reasonable time (with ASF as a baseline), taking 1-2 loops around the air staging to dock.

## Checklist

### DIFF
--- a/changelog/snippets/balance.6966.md
+++ b/changelog/snippets/balance.6966.md
@@ -1,0 +1,5 @@
+- (#6966) Reduce the distance at which T1 Interceptors and Swiftwinds hover instead of turning when given a move order.
+
+  **T1 Interceptors and Swiftwind (UAA0102, UEA0102, URA0102, XSA0102, XAA0202)**
+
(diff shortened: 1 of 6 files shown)


Rhaelya (approved):
Works as intended

PR #6978, opened 2025-12-01
Set game speed adjustable when cheats enabled

4z0t:
Forces game speed to be set to "adjustable" when cheats are enabled and is single player.

## Additional context
Useful for testing scenarios.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * GameSpeed can now be adjusted while cheats are enabled in single-player games.

### DIFF
--- a/lua/ui/lobby/lobby.lua
+++ b/lua/ui/lobby/lobby.lua
@@ -2267,6 +2267,9 @@ local function TryLaunch(skipNoObserversCheck)
         if scenarioInfo.AdaptiveMap then
             gameInfo.GameOptions["SpawnMex"] = gameInfo.SpawnMex
         end
+        if gameInfo.GameOptions["CheatsEnabled"] == "true" then
+            gameInfo.GameOptions["GameSpeed"] = "adjustable"
+        end
 
         HostUtils.SendArmySettingsToServer()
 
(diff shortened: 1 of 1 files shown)


BlackYps (approved):
I don't think we can enable cheats from the command line, can we? This would be also nice to add, but can be done in a separate pull request.

4z0t:
@BlackYps Well, actually that would be much better to make it as a command line argument.

Garanas:
> @BlackYps Well, actually that would be much better to make it as a command line argument.

It already is, see also:

- https://github.com/FAForever/fa/blob/develop/scripts/LaunchFAInstances.ps1#L28

BlackYps:
I didn't know about that, because I was looking here and it isn't mentioned: https://wiki.faforever.com/en/Development/Game-Development/Command-Line-Switches
I'll add it to the article once I understood how it really works.

Looking through the repo, the /gameoptions option is only checked in the autolobbyController. When is a game launch using the autolobby path?

BlackYps:
More generally speaking, we should have documentation somewhere how launching the game works. This is also important information for developers working on the client or the server. Is there something somewhere already? If not we should really start creating that.

Garanas:
The way we start the game through script is very different to how the client starts the game. If I am not mistaken, the client sends a crafted GPG message to inform the game to join a particular lobby. That's not something that we can do.

PR #6981, opened 2025-12-09
Fix ship wake effects disappearing after a while

### LINKED ISSUE #6980: Naval units' movement effects suddenly disappear when moving
### Describe the issue

This is a problem present in both the original FA and FAF versions, and I suspect the cause lies in the engine. However, I believe it's possible to fix this issue using other methods.

How to reproduce this bug: Simply give five UEF T3 warships (or any other naval units) and command them to move to a distance. You'll notice that these units initially have movement effects (trails), but after a few seconds, one or two ships suddenly lose their effects. This disappearance occurs abruptly without requiring camera movement. Furthermore, these disappearing trails do not reappear when the camera is repeatedly moved.

The only way to restore the effect to normal is to issue a stop command to the unit, followed by a move command. The movement effect will then reappear.

It's worth noting that reproducing this issue requires very few ships, generally only five units, and this problem occurs very frequently, significantly impacting the visual effects of naval units in the game.

Furthermore, reproducing this issue doesn't require camera movement, thus ruling out LOD or excessive unit effects as the cause. Possible Fix: I used GPT to generate some code that, by hooking /lua/sim/units/SeaUnit.lua, can partially correct this issue.

It should be noted that since I don't know Lua programming, and this code was generated by GPT, I don't own any rights to this work. This code is merely material for illustrating the problem or for discussion, and is not a mod I created.

Furthermore, this code causes the naval unit's wake effect to continuously generate, regardless of whether the unit is moving. However, visually, this doesn't seem to be a problem; naval units still have water splashes near their hulls when stationary, which seems acceptable.

Of course, hopefully someone can truly solve this problem. I'm merely providing a possible direction for a fix: to continuously generate the wake effect of the naval unit's movement.

Since I can't upload the Lua file, I've put the code generated by GPT here.

local OldSeaUnit = SeaUnit

-- Do not change original MobileUnit / Unit behavior; just add an extra "fallback wake" layer on SeaUnit
SeaUnit = ClassUnit(OldSeaUnit) {

    ------------------------------------------------------------------
    -- Config: map bp Type to a list of emitter blueprints
    ------------------------------------------------------------------
    NavyWakeFxMap = {

        -- Rear wake trails; you can add or remove emitters here
        BackWake = {
            '/effects/emitters/water_move_trail_back_01_emit.bp',
            '/effects/emitters/water_move_trail_back_02_emit.bp',
            '/effects/emitters/water_move_trail_back_03_emit.bp',
            '/effects/emitters/water_move_trail_back_04_emit.bp',

        },
    },

    ------------------------------------------------------------------
    -- After the unit is finished: run original logic, then start the wake watchdog thread
    ------------------------------------------------------------------
        OldSeaUnit.OnStopBeingBuilt(self, builder, layer)
        self:StartNavyWakeThread()
    end,

    OnKilled = function(self, instigator, type, overkillRatio)
        self:StopNavyWakeThread()
        OldSeaUnit.OnKilled(self, instigator, type, overkillRatio)
    end,

    OnDestroy = function(self)
        self:StopNavyWakeThread()
        OldSeaUnit.OnDestroy(self)
    end,

    ------------------------------------------------------------------
    -- Start / stop the watchdog thread
    ------------------------------------------------------------------
    StartNavyWakeThread = function(self)
        if self.NavyWakeThread then
            KillThread(self.NavyWakeThread)
            self.NavyWakeThread = nil
        end

        local wakeDefs = nil

        if bp.Display
            and bp.Display.MovementEffects
            and bp.Display.MovementEffects.Water
            and type(bp.Display.MovementEffects.Water.Effects) == 'table'
        then
            wakeDefs = bp.Display.MovementEffects.Water.Effects
        end

        -- If the unit has no water MovementEffects defined, there is nothing to do
        if not wakeDefs then
            return
        end

        self.NavyWakeDefs = wakeDefs
        self.NavyWakeFxBag = {}

        self.NavyWakeThread = self:ForkThread(self.NavyWakeThreadFn)
    end,

    StopNavyWakeThread = function(self)
        if self.NavyWakeThread then
            KillThread(self.NavyWakeThread)
            self.NavyWakeThread = nil
        end

        local bag = self.NavyWakeFxBag
        if bag then
            for k, fx in pairs(bag) do
                if fx and not fx:BeenDestroyed() then
                    fx:Destroy()
                end
                bag[k] = nil
            end
        end
    end,

    ------------------------------------------------------------------
    -- Worker thread: use position delta + bp structure to create / remove wake effects
    ------------------------------------------------------------------
    NavyWakeThreadFn = function(self)
        local wakeDefs = self.NavyWakeDefs
        if not wakeDefs then
            return
        end

        local army = self:GetArmy()
        local map  = self.NavyWakeFxMap

        -- Infer movement by checking position changes between samples
        local lastX, lastY, lastZ = unpack(self:GetPosition())
        -- Threshold: if distance^2 between samples exceeds 0.01 (~0.1 units), treat as moving
        local moveThresholdSq = 0.01

        while not self.Dead do
            local layer = self:GetCurrentLayer()

            if layer == 'Water' then
                local x, y, z = pos[1], pos[2], pos[3]

                local dx = x - lastX
                local dz = z - lastZ
                local distSq = dx * dx + dz * dz

                local moving = distSq > moveThresholdSq

                -- Update last position
                lastX, lastY, lastZ = x, y, z

                -- Check if our own wake emitters are present
                local bag   = self.NavyWakeFxBag
                local hasFx = false
                for _, fx in pairs(bag) do
                    if fx and not fx:BeenDestroyed() then
                        hasFx = true
                        break
                    end
                end

                -- 1) Create: if the unit is clearly moving AND we have no wake yet
                if moving and not hasFx then
                    for _, def in ipairs(wakeDefs) do
                        if type(def) == 'table' and def.Bones then
                            local scale  = def.Scale or 1
                            local fxList = map[def.Type]

                            if fxList then
                                for _, boneName in ipairs(def.Bones) do
                                    for _, fxPath in ipairs(fxList) do
                                        local fx = CreateAttachedEmitter(self, boneName, army, fxPath)
                                        fx:ScaleEmitter(scale)
                                        table.insert(bag, fx)
                                    end
                                end
                            end
                        end
                    end

                -- 2) Cleanup: if the unit is effectively not moving AND our wake is present
                elseif (not moving) and hasFx then
                    for k, fx in pairs(self.NavyWakeFxBag) do
                        if fx and not fx:BeenDestroyed() then
                            fx:Destroy()
                        end
                        self.NavyWakeFxBag[k] = nil
                    end
                end
            else
                -- Not on the Water layer (airlifted / flying / sinking etc.) → remove our wake
                if self.NavyWakeFxBag then
                    for k, fx in pairs(self.NavyWakeFxBag) do
                        if fx and not fx:BeenDestroyed() then
                            fx:Destroy()
                        end
                        self.NavyWakeFxBag[k] = nil
                    end
                end
            end

            -- Check roughly every 0.2 seconds
            WaitTicks(6)
        end
    end,
}

### Expected Behavior

When naval units move, they should continuously generate contrails, just like normal warships in the video.

### Screenshots

https://github.com/user-attachments/assets/1072aed4-5c01-46c2-b8d7-34a20d5baad5

speed2CZ:
Fixes #6980
When ships start moving the 'Default' terrain type is used for picking the effects, which has Wake effects defined.

But as the ship is moving it encounters different WaterXX terrain types and they dont specify any extra affects, so the 'Default' one should be used. That's what this change fixes.

As a side effect it might fix some other missing effects, since the 'Default' terrain type is suppose to be used when the non default one doesnt specify the effects.
- moving using around the map and watching the movement effects

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored disappearing terrain visual effects (notably ship wake), so environmental effects remain visible and consistent during gameplay.
* **Documentation**
  * Added changelog entry documenting the fix for issue #6981.

### DIFF
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -3561,8 +3561,8 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
     OnMotionTurnEventChange = function() end,
 
     ---@param self Unit
-    ---@param new string
-    ---@param old string
+    ---@param new TerrainType
+    ---@param old TerrainType
     OnTerrainTypeChange = function(self, new, old)
         self.TerrainType = new
         if self.MovementEffectsExist then
@@ -3729,13 +3729,13 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
     ---@param effectsBag? TrashBag
     ---@param terrainType? TerrainType
     CreateTerrainTypeEffects = function(self, effectTypeGroups, fxBlockType, layer, typeSuffix, effectsBag, terrainType)
-        local effects, terrainFX, GetTerrainTypeEffects
+        local GetTerrainTypeEffects = self.GetTerrainTypeEffects
         local pos = self:GetPosition()
         local army = self.Army
+
+        local terrainFX
         if terrainType then
             terrainFX = terrainType[fxBlockType][layer]
-        else
-            GetTerrainTypeEffects = self.GetTerrainTypeEffects
         end
 
         for _, typeGroup in effectTypeGroups do
@@ -3745,11 +3745,8 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
                 continue
             end
 
-            if terrainType then
-                effects = terrainFX[typeGroup.Type]
-            else
-                effects = GetTerrainTypeEffects(fxBlockType, layer, pos, typeGroup.Type, typeSuffix)
-            end
+            -- Use TerrainType specific effects or fallback to 'Default' TerrainType effects
+            local effects = terrainType and terrainFX[typeGroup.Type] or GetTerrainTypeEffects(fxBlockType, layer, pos, typeGroup.Type, typeSuffix)
             if table.empty(effects) then
                 continue
             end
--- a/changelog/snippets/fix.6981.md
+++ b/changelog/snippets/fix.6981.md
@@ -1,0 +1,1 @@
+- (#6981) Fix disappearing terrain effects (Mainly ship wake effects, but other effects might be affected as well).


lua/sim/Unit.lua
```diff
@@ -3745,11 +3743,8 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
                 continue
             end
 
-            if terrainType then
-                effects = terrainFX[typeGroup.Type]
-            else
-                effects = GetTerrainTypeEffects(fxBlockType, layer, pos, typeGroup.Type, typeSuffix)
-            end
+            -- Use TerrainType specific effects or fallback to 'Default' TerrainType effects
+            effects = terrainType and terrainFX[typeGroup.Type] or GetTerrainTypeEffects(fxBlockType, layer, pos, typeGroup.Type, typeSuffix)
```
Hdt80bro: I'd move the `local effect` declaration from line 3732 to here if there's now going to be a single assignment - and then at that point you might as well give `terrainFX` and `GetTerrainTypeEffects` their own declarations as well so you don't have a multiassignment with a padding nil.

Hdt80bro:
My one other comment us that the change from the `if` expression to the ternary expression (though unnecesary with LuaPlus's nil indexing if you simply use `or`, please leave it for readability 🙂) is subtle but entirely crucial here - most people would see them as equivalent, so your exact intention might get overlooked. However, the alternative is verbose, so I'm not convinced that simply adding a comment like you did isn't going to end up being good enough. So just keep in mind I guess.

Garanas:
Great find! I thought it was just the engine being buggy. I'm glad to be wrong! I can't ping him, but Zhang may also be excited about this.

zhanghm18GIT:
Thank you all for your work. I think this fix will greatly improve the visual effects of naval battles in FAF!

lua/sim/Unit.lua
```diff
@@ -3745,11 +3745,8 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
                 continue
             end
 
-            if terrainType then
-                effects = terrainFX[typeGroup.Type]
-            else
-                effects = GetTerrainTypeEffects(fxBlockType, layer, pos, typeGroup.Type, typeSuffix)
-            end
+            -- Use TerrainType specific effects or fallback to 'Default' TerrainType effects
+            local effects = terrainType and terrainFX[typeGroup.Type] or GetTerrainTypeEffects(fxBlockType, layer, pos, typeGroup.Type, typeSuffix)
```
BlackYps, line 3749: Expanding on what Hdt80bro said, I don't understand how this is different to the old code. Can you explain it for a scrub like me?

BlackYps:
If I understand the whole situation correctly, then the comment in the TerrainTypes file `If there is no effects defined for a specific terrain, the 'Default' terrain type will be used.` was wrong in the sense that this fallback did in fact not work, right?
And this PR fixes it, which is very nice. I have one more question: When a terraintype aims to specifically disable an effect that is defined in the default terrain type. Would that work by specifically defining that effect as empty? For example, to disable the back wakes, would this work?
```
Water = {
    BackWake = {},
}

lua/sim/Unit.lua
speed2CZ: Previously if terrain type was specified it got the effects from it and when not found, nothing happened.
Now the default terrain type is used instead

speed2CZ:
> If I understand the whole situation correctly, then the comment in the TerrainTypes file `If there is no effects defined for a specific terrain, the 'Default' terrain type will be used.` was wrong in the sense that this fallback did in fact not work, right? And this PR fixes it, which is very nice. I have one more question: When a terraintype aims to specifically disable an effect that is defined in the default terrain type. Would that work by specifically defining that effect as empty? For example, to disable the back wakes, would this work?
> 
> ```
> Water = {
>     BackWake = {},
> }
> ```

Correct and yes

BlackYps (approved):
Looks good, just the snippet should have a different text, because it doesn't just affect ships.
"Fix disappearing terrain effects" should be sufficient if you don't want to go into more details.

zhanghm18GIT:
@speed2CZ Thank you for your work. However, I found that even if this PR was merged, it only solved the mobility issue for naval units. Still unable to solve the problem of treads not displaying on certain undefined terrains.

speed2CZ:
That does sound like a different issue

PR #6985, opened 2025-12-20
Fix profiler benchmarking sending unserializable data

lL1l1:
## Issue
The profiler can send unserializable data through the upvalues provided in function debug information. The code handles basic cases with functions and cfunctions, but not everything. Tables are the main issue since they're easy to find reasons to upvalue and  they can have cyclic references (the engine recursively serializes without cycle protection, so it causes a stack overflow) and they can reference unserializable data of a wide variety of types.

1. Look in the engine to find serializable types, and document them in `utilities.lua` in the `SERIALIZABLE_TYPES` table
2. Add the utility function `SerializableDeepCopy` to convert unserializable values/tables to serializable versions.
3. Add a benchmark comparing accessing values via cfunc vs lua. This is what led me to discover the issue in the first place, since it upvalued ArmyBrains, which has cyclic reference through Brain.CDR.Brain and references userdata (`_c_object`s)/other objects (threads).

The benchmark loads when you navigate to its file in the profiler and select one of the local/upvalue lua functions (this means the ArmyBrains table was serialized successfully).

An example command:
```
local a = {ForkThread(function()end)}
local b = {}
local c = {}
local d = {}
a.b = b
a.c = c
b.d = d
c.d = d
d.a = a
local utils = import('/lua/utilities.lua')
reprsl(utils.SerializableDeepCopy(a), {depth = 4})
```

## Checklist

## Summary by CodeRabbit

* **Chores**
  * Added a new benchmarking suite to measure and analyze performance metrics.
  * Enhanced serialization utilities to improve data handling and compatibility across system boundaries.
  * Improved profiler infrastructure for better data processing and cycle detection.

### DIFF
--- a/lua/utilities.lua
+++ b/lua/utilities.lua
@@ -569,3 +569,55 @@ function RotateVectorXYZByQuat(vX, vY, vZ, q)
             vY * qW + vZ * qX + vW * qY - vX * qZ,
             vZ * qW - vY * qX + vX * qY + vW * qZ
 end
+
+--- These are the types that the engine can send across the ui-sim boundary.
+--- Tables have to be handled for loops and unserializable keys/values.
+SERIALIZABLE_TYPES = {
+    number = true,
+    string = true,
+    boolean = true,
+    ["nil"] = true,
+}
+local serializableTypes = SERIALIZABLE_TYPES
+
+--- Returns a deep copy with unserializable values converted using `tostring` 
+--- and cyclic references converted to strings referencing the `_seenTableId` key.
+--- The result can be sent across the ui-sim boundary.
+---@generic T
+---@param t T
+---@return T
+function SerializableDeepCopy(t)
+    local st = serializableTypes
+    local type_t = type(t)
+    if type_t ~= 'table' then
+        return st[type_t] and t or tostring(t)
+    end
+
+    local backrefs = {}
+    local function CreateSerializableAny(_t)
+        local type_t = type(_t)
+        if type_t ~= 'table' then
+            return st[type_t] and _t or tostring(_t)
+        end
+
+        local b = backrefs[_t]
+        if b then
+            -- store table id since table `tostring` uses memory location which won't be the same between Sim and UI
+            if not b._seenTableId then
+                b._seenTableId = tostring(_t)
+            end
+            -- format makes it easy to find in repr output
+            return '_seenTableId = "' .. b._seenTableId .. '"'
+        end
+
+        local r = {}
+        backrefs[_t] = r
+        for k, v in _t do
+            r[CreateSerializableAny(k)] = CreateSerializableAny(v)
+        end
+
+        return r
+    end
+
+    return CreateSerializableAny(t)
+end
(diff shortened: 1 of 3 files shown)


lua/utilities.lua
```diff
@@ -569,3 +569,55 @@ function RotateVectorXYZByQuat(vX, vY, vZ, q)
...
+--- and cyclic references converted to strings referencing the `_seenTableId` key.
+--- The result can be sent across the ui-sim boundary.
+---@generic T
+---@param t T
+---@return T
+function SerializableDeepCopy(t)
+    local st = serializableTypes
+    local type_t = type(t)
+    if type_t ~= 'table' then
+        return st[type_t] and t or tostring(t)
```
Hdt80bro: This pattern will incorrectly stringify `false` values or a `nil` passed in at the top level.

lua/utilities.lua
```diff
@@ -569,3 +569,55 @@ function RotateVectorXYZByQuat(vX, vY, vZ, q)
...
+        local type_t = type(_t)
+        if type_t ~= 'table' then
+            return st[type_t] and _t or tostring(_t)
+        end
+
+        local b = backrefs[_t]
+        if b then
+            -- store table id since table `tostring` uses memory location which won't be the same between Sim and UI
+            if not b._seenTableId then
+                b._seenTableId = tostring(_t)
```
Hdt80bro, line 615: Using `tostring` on the table here seems to go against the comment above

lua/utilities.lua
```diff
@@ -569,3 +569,55 @@ function RotateVectorXYZByQuat(vX, vY, vZ, q)
...
+            end
+            -- format makes it easy to find in repr output
+            return '_seenTableId = "' .. b._seenTableId .. '"'
```
Hdt80bro, line 618: If the table ID never changes, why dirty the table with an unoriginal field to store the seen table ID? We could use `tostring` every time.

Hdt80bro (changes requested):

lua/utilities.lua
lL1l1: At first I just incremented an ID but then decided to use the table string as a unique ID because then if you have access to a debugger the information is there.

lua/utilities.lua
lL1l1: The table ID should change after being copied across the sim-ui boundary, so the table `tostring` won't give useful information.
An alternative to not dirty the table would be to return an extra table documenting all the references that were converted to strings.

Hdt80bro: But you already are using `tostring`; as you've written it right now, if a table's `_seenTableId` field exists, it is always filled with `tostring` applied to that table. You gave a fine reason to use `tostring` in your previous reply, so I'm not sure why you're saying it doesn't give any useful information here.

lL1l1: >  if a table's _seenTableId field exists, it is always filled with tostring applied to that table

This is not true because the table is presumedly copied across sim-ui, where the field will be filled with tostring applied to the old table before the copying. When you read the copied table you need the old `tostring` ID to rebuild cyclic references.

Hdt80bro: Oh I see - I thought `_seenTableId` was only being used for serialization (the return value on line 618). But you also want to leave it as a marker as part of the data in case you want to deserialize it (which we haven't set up) rather than only using the data we pass for textual purposes.

Hdt80bro (approved):

PR #6991, opened 2025-12-25
Hide chat in NIS mode

4z0t:
Hides chat in NIS mode for better immersion in coop missions.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed chat display behavior during non-interactive sequences to prevent unnecessary UI updates and unintended scrolling, ensuring a cleaner viewing experience during cinematics and other special game moments.

### DIFF
--- a/lua/ui/game/chat.lua
+++ b/lua/ui/game/chat.lua
@@ -289,10 +289,6 @@ function CreateChatLines()
     end
 end
 
-function OnNISBegin()
-    CloseChat()
-end
-
 function SetupChatScroll()
     GUI.chatContainer.top = 1
     GUI.chatContainer.scroll = UIUtil.CreateVertScrollbarFor(GUI.chatContainer)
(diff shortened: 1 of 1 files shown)


speed2CZ:
I saw the message "Player unpaused the game" each time the cinematics start, but afaik thats not pausing the game. Do you know why that message is triggered in the first place?

4z0t:
It comes from `LockInput` in worldview.lua

speed2CZ:
Can that be changed? Since I would say the main issue here is that there's a message triggered when it shouldnt.
Hiding the chat during NIS mode is a good idea as well, but currently its just hiding the main issue.

4z0t:
I don't know what side effects could be there if I remove it

lL1l1:
I think it would be fine to add a way to force resuming the session without sending a chat message, like a param to the hooked function or a separate global. Then you get to remove the chat message without removing the resumption behavior.

4z0t:
> I think it would be fine to add a way to force resuming the session without sending a chat message, like a param to the hooked function or a separate global. Then you get to remove the chat message without removing the resumption behavior.

Tbf I'd remove it completely from coop

PR #7003, opened 2026-01-27
adding "Reverse move" hotkey

Strogoo:
1. Works only with units that have MaxSpeedReverse > 0 in blueprint.
2. You can't queue it with shift or combine with other orders. Single order only (see video)
3. **Balance tweaks are needed before merging this**. Many units have their MaxSpeedReverse = MaxSpeed (*forward move). Worst case scenario - we can disable this command for certain units.
4. If you have any suggestions about improving Lua part of it - feel free to post it here

  .exe for testing will be in Zulip -> engine-development -> reverse move

## Additional context

requires https://github.com/FAForever/FA-Binary-Patches/pull/116

https://github.com/user-attachments/assets/eaf894da-0bd6-4692-8fea-b37f815d113d

## Summary by CodeRabbit

* **New Features**
  * Reverse move command mode allowing selected units with reverse-capable engines to move backward
  * Two new keyboard shortcuts to enter reverse-move command mode (including shift variation)
  * New reverse-move cursor visuals when command mode is active
  * Command system now tracks and toggles reverse-move state for affected units during orders


lua/SimCallbacks.lua
```diff
@@ -826,6 +826,38 @@ do
...
+do
+
+    ---@param data { Enable: boolean, ShowMsg: boolean }
+    ---@param selection Unit[]
+    Callbacks.ForceReverseMove = function(data, selection)
+        -- verify selection
+        if not data.Units then
+            selection = SecureUnits(selection)
+        else
+            selection = SecureUnits(data.Units)
```
Garanas, line 838: Would it not be better to always use the current selection? That way you don't even need to use the SecureUnits function as far as I am aware.

lua/keymap/keydescriptions.lua
```diff
@@ -559,6 +559,8 @@ keyDescriptions = {
     ['shift_load_transports_clear'] = '<LOC key_desc_shift_load_transports_clear>Load into transports. Applies immediately',
     ['copy_orders'] = '<LOC key_desc_copy_orders>Copy orders of the unit the mouse is on top of',
     ['shift_copy_orders'] = '<LOC key_desc_shift_copy_orders>Copy orders of the unit the mouse is on top of',
+    ['reverse_move'] = '<LOC key_desc_reverse_move>Reverse move.',
+    ['shift_reverse_move'] = '<LOC key_desc_shift_reverse_move>Reverse move.',
```
Garanas: These should not end with a full stop, none of the other hotkeys (in the GitHub preview) do this.

Strogoo: yea, idk where it came from :) I think I just copy-pasted something above and changed the text. Will fix it

lua/SimCallbacks.lua
Strogoo: There are cases when we don't need all selected units but only part of them (exlcude units without MaxSpeedReverse or with MaxSpeedReverse = 0). I do this check on UI and sim side just in case. Theoretically we can do this check only in sim and just use current selection, idk what is better

BlackYps:
I'm not sure this is a useful addition. Please start a discussion in the suggestions channel on discord to involve more people.

BlackYps:
Closing for now due to inactivity

PR #7009, opened 2026-02-10
Rework HQ and Support Factories

Saver27:
**### Redesign of all HQ and Support Factories 2.0**

As described in the forum post (https://forum.faforever.com/topic/5790/redesign-of-all-hq-and-support-factories), the first step towards redesigning the factory models was started some time ago. Since this work was unfortunately never completed, I have picked up the project again and continued it.

**Overview of Changes**

All models were recreated and adjusted based on the FAF templates, enabling upgrades to HQ and Support factories

All models now have their own build animations

LOD1 models were largely adjusted to match the new LOD0 models

**### Specific Changes**
**Build Area of Naval Factories**

The required build area for all naval factories has been adjusted.
This change is based on the observation that when an existing factory was upgraded to a higher tech level, the build footprint shifted once the mesh expanded.

This adjustment prevents this issue/effect from occurring (see forum post 
https://forum.faforever.com/topic/5790/redesign-of-all-hq-and-support-factories/26).

**Visual Adjustments to Aeon T3 Naval Ship Construction**

As described in forum post (19), the Mercury Pool visuals were displayed disproportionately large compared to the ships.

These visuals have now been adjusted to better match the ship sizes (https://forum.faforever.com/topic/5790/redesign-of-all-hq-and-support-factories/63).

**Added Death Animation for Seraphim Naval Factories**

While creating the death animations for the Seraphim faction, it was discovered that the animations were not being played.
This issue was fixed by adding the PlayAnimation(...) function to the relevant scripts.

See: (https://forum.faforever.com/topic/5790/redesign-of-all-hq-and-support-factories/90)

**Adjustment of TargetBones for Land and Air Factories**

During testing, I noticed that some units were unable to damage certain factories as intended. Since no changes have been made to the hitboxes so far, this issue is likely caused by the reorientation of various bones in the updated models.

To verify this assumption, several tests were conducted with modified configurations. As a result, the UEF, Aeon, and Cybran factory models now use the following configuration:

AI = {
    TargetBones = {
        "Attachpoint",
    },
},

I would suggest revising the hitboxes of the land and air factories. Since this issue was only briefly mentioned in the forum post, here are a few images to illustrate the point more clearly.


Saver27:
While reviewing the files on GitHub, I noticed several issues (e.g., incorrect spacing and unwanted symbols). Additionally, some values were modified since my last update that were not intended to be changed (e.g., HP).
I will review the data again and correct these issues accordingly.

BlackYps:
Very nice writeup on the changes, good job!

lL1l1 (commented):
Aeon factories reviewed:
Aeon T1->T2 land factory upgrade detaches the left arm which looks weird because its physically impossible and the factory becomes asymmetrical during the upgrade. Not sure about the solution, it is a difficult problem.

Aeon T1 navy fac water effects and hitbox do not match the mesh. 

Image

Do you know if `Scrolling = true` was used on the aeon factor lod 0 meshes?

Saver27:
@IL1I1 thank you very much for your testing.

**Aeon T1 Land Factory Upgrade**

Unfortunately, while creating the animation, I couldn’t come up with a better solution than the current one. Previously, both roof elements kept their positions during the upgrade process.

Since the tower positions change in the T2 HQ and Support models, I considered this approach to be the most reasonable way to visually represent the upgrade.

I may revise the entire Aeon Land Factory model again. I’ve received feedback from multiple sources expressing the wish for a model that more closely resembles the original design from before the changes.

**Aeon Naval Factory – Hitbox / Effects**

I have not implemented effects of this kind in the project so far. However, I am still aware of the hitbox issue.

If I decide to adjust the hitbox, I would also review other hitboxes as part of that process, since some of them do not fully match the updated model. 

### Description

The following hitboxes need adjustment, as they currently do not properly match their respective models:

_Affected Units_

Cybran Land Factory
The visual model is significantly wider than the current hitbox.

Seraphim Air Factory
The T3 hitbox should be applied to all tech levels, as the base structure of the model remains identical.

Aeon Naval Factory
The hitbox should be reduced to match the actual model size.
Note: Additional testing is required to ensure that the previously described bug is not triggered again.

_Expected Behavior_

Hitboxes accurately match the visible model dimensions.

Consistent hitbox structure for units sharing the same base model.

No regression of previously identified bugs.

_To Do_

 Increase width of the Cybran Land Factory hitbox

 Apply T3 hitbox to all Seraphim Air Factory models

 Reduce Aeon Naval Factory hitbox size

 Perform regression testing for the known issue

Below is a short overview along with images of the currently used hitboxes in the game.

**_Lines marked in yellow indicate a deviation from the other tech levels of the same
model._**




















[Factory Hitbox.pdf](https://github.com/user-attachments/files/25457123/Factory.Hitbox.pdf)

Saver27:
I'm currently reviewing and correcting all the blueprints and scripts.

After that, I'll work on the hitboxes. 

Since I haven't quite understood Git yet, I would be grateful if someone could review this work afterwards.

Saver27:
Proposal: Adjustment of the Aeon Naval Hitbox

The proposed adjustment should be carefully reviewed and evaluated.

Purpose of the change

The hitbox will be aligned more accurately with the existing model, ensuring better visual and technical consistency with the actual factory size.

image


Potential impact

The hitbox will more closely match the model.

However, when multiple factories are placed in a row, their build position may shift after upgrading to the next tech level.

The factories remain fully functional.

There is a possibility of model overlap if units were placed before upgrading to the next tech level.

image


Detailed Changes

Aeon Naval T1, T2, T3

T1

SkirtOffsetZ: -1 → -2

SizeZ: 13 → 10.5

T2

SkirtOffsetZ: -1 → -2

SizeZ: 13 → 12

T3

SkirtOffsetZ: -1 → 0

SizeZ: 13 → 14


The current state of the project only presents the possibility that the hitbox of the T1 version is slightly too large. However, all other errors have been corrected.

units/UAB0101/UAB0101_unit.bp
```diff
@@ -71,13 +81,16 @@ UnitBlueprint{
                 {
                     LODCutoff = 200,
                     ShaderName = "Aeon",
+                    AlbedoName = "/units/uab0101/uab0101_albedo.dds",
+                    NormalsName = "/units/uab0101/uab0101_normalsts.dds",
+                    SpecularName = "/units/uab0101/uab0101_specteam.dds",
```
BlackYps: Aren't these just default values here? I don't think they are needed when the texture is actually from the same unit. The higher tier factories do need explicit texture paths, because they reference textures from a different unit.

Same thing in other T1 units.

units/UAB0102/UAB0102_script.lua
```diff
@@ -13,4 +13,4 @@ local AAirFactoryUnit = import("/lua/aeonunits.lua").AAirFactoryUnit
 ---@class UAB0102 : AAirFactoryUnit
 UAB0102 = ClassUnit(AAirFactoryUnit) {}
 
-TypeClass = UAB0102
+TypeClass = UAB0102
```
BlackYps: You have this change in all the script files. Please get rid of it. It's probably because of a setting in your editor. See: https://gist.github.com/olivertappin/19a95236d8b69f1c9eadfc799a9ab7fc

units/UAB0201/UAB0201_unit.bp
```diff
@@ -15,7 +17,7 @@ UnitBlueprint{
         DoneBeingBuilt = Sound { Bank = 'UAB',        Cue = 'UAB0201_Activate',           LodCutoff = 'UnitMove_LodCutoff' },
         UISelection    = Sound { Bank = 'Interface',  Cue = 'Aeon_Select_Factory',        LodCutoff = 'UnitMove_LodCutoff' },
     },
-    AverageDensity = 6400,
+    AverageDensity = 6600,
```
BlackYps: Isn't this value irrelevant on buildings?

units/UEB0101/UEB0101_Albedo.dds
BlackYps, line 1: I'm not a fan of the new textures of the UEF land factory. They look too clean and the normals are inverted in the sense that ridges seem to stick out. I'd just remove these textures for now.
We can always think about a texture rework at a later date in a separate pull request.

BlackYps (commented):
I left some comments, other than that it looks good.

units/UAB0201/UAB0201_unit.bp
Saver27: Unfortunately, I don't know. However, since this value is included in the blueprints of the standard game, I've added it to all factories accordingly.

units/UEB0101/UEB0101_Albedo.dds
Saver27: That's perfectly fine with me. I will remove the file first; it is still available in my mod on FAF if necessary.

little explanation about this:
I took this on because Jip suggested it and Baltazar made his work available to me.

units/UAB0101/UAB0101_unit.bp
Saver27: You're probably right. Since I'd only worked with mods before, I overlooked the issue with the T1 models.

The current FAF only refers to

AlbedoName = "ueb0101_lod1_albedo.dds",

SpecularName = "ueb0101_lod1_specteam.dds",

units/UAB0102/UAB0102_script.lua
Saver27: Thank you for the explanation. I use Note++ and didn't know how the symbol came about or what it meant.

I'll make the change tomorrow.

Saver27:
I hope I was able to address the obvious issues in this branch. Please let me know if you notice any further problems.

units/UAB0201/UAB0201_unit.bp
BlackYps: According to the documentation it's for determining which unit gets bumped out of the way when two units collide. So for buildings this is an unncessary value and probably the reason why most factories do not have it set explicitly.

units/UEB0102/UEB0102_script.lua
BlackYps, line 1: Most changes to unit script files only have whitespace changes. Could you revert these? Especially in big pull requests like these it's important to not build in meaningless changes that make the diff look even bigger.

BlackYps (approved):
Very nice to see this project finished!

Issue #6989, opened 2025-12-21
[Bug]: T1 Bombers do not auto attack


JaggedAppliance:
### Describe the Bug

T1 bombers, if given a move order, will just land and will not acquire any targets. The will be totally idle.

### Reproduce the bug

Give a move order to a bomber.

4z0t:
Same with attack-move order.

lL1l1:
Would also be worth taking a look at making t2 bombers auto-attack. I was told that they do so on steam but not on faf.

PR #7022, opened 2026-02-11
Fix bombers do not auto attack nearby enemies

Strogoo:
Fix this: https://github.com/FAForever/fa/issues/6989

Also some info from Nomander:
изображение

## Summary by CodeRabbit

* **Bug Fixes**
  * Bombers now automatically attack nearby enemies during combat encounters instead of remaining locked on their initial target, improving responsiveness and reducing missed opportunities to engage threats. This fix ensures bomber units will re-evaluate and select closer targets when appropriate, leading to more consistent combat behavior.

### DIFF
--- a/changelog/snippets/fix.7022.md
+++ b/changelog/snippets/fix.7022.md
@@ -1,0 +1,1 @@
+- (#7022) Fix bombers do not auto attack nearby enemies.
--- a/changelog/snippets/fix.7022.md
+++ b/changelog/snippets/fix.7022.md
@@ -1,1 +1,1 @@
-- (#7022) Fix bombers do not auto attack nearby enemies.
+- (#7022) Fix bombers not automatically attacking nearby enemies.
--- a/lua/sim/Unit.lua
+++ b/lua/sim/Unit.lua
@@ -3509,21 +3509,6 @@ Unit = ClassUnit(moho.unit_methods, IntelComponent, VeterancyComponent, DebugUni
         elseif new == 'Bottom' or new == 'Hover' then
             -- Play the "landed" sound
             self:PlayUnitSound('Landed')
-            
-            -- fix bombers do not auto attack nearby enemies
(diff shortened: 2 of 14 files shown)


speed2CZ:
Im not so sure about the solution that adds a new category. Wont that leave modded units broken?
Isnt this a new regression bug introduced lately?

Garanas (dismissed):
The correct place to fix this is here:

- https://github.com/FAForever/fa/blob/develop/lua/system/blueprints-weapons.lua#L108-L130

Note that this flag (when set to `true`) can have a real impact on performance.

Strogoo:
@Garanas  Yea, I know that the root cause is `weapon.AlwaysRecheckTarget = false` for all `weapon.NeedToComputeBombDrop` in blueprints-weapons.lua. But if we revert this

изображение

then `bombers to suddenly retarget and don't drop teir bomb` bug will be a thing again?

@speed2CZ it can be done via checking weapon.NeedToComputeBombDrop instead, I just thought that adding a new category to the unit and using CategoriesHash will be a bit faster performance wise. But since it's not a hot code at all, I think it's ok to check the weapon bp instead

Strogoo:
@Garanas @speed2CZ Made a new version. This one is much better (thanks to speed2 and his mention about modded units). Also I added a comment to the code why we fix it in `unit.lua` and not in `blueprints-weapons.lua`.

lL1l1:
When I was testing after that comment in the OP, I found no evidence that `AlwaysRecheckTarget` can cause sudden retargeting, which I think is because `AutoInitiateAttackCommand` forces the unit attack and kill the first target it targets. The reason bombers might've retargeted previously is because of the `ResetTarget()` in `RackSalvoReadyState`, and that might've mislead the original comment in blueprints-units.

I'm not sure on the exact engine targeting details, but despite that bombers should always have `AlwaysRecheckTarget` because they will always have `AutoInitiateAttackCommand` since that attack command allows them to do attack run behavior.

lL1l1 (approved):

PR #7039, opened 2026-02-17
Salem Class: fix Nanite Torpedoes reducing main cannon ROF

NoryGit:
When the Salem uses its torpedoes, its main cannon begins to fire at a slower rate (in bursts), causing it to lose DPS.

Tested with the Proton Cannon ROF set to 1. The bug also occurs at the default ROF, although it is less noticeable.

## Checklist

## Summary by CodeRabbit

* **Gameplay Balance**
  * Salem Class (URS0201) Nanite Torpedoes no longer prevent other weapons from firing, allowing simultaneous weapon fire for new tactical options.
* **Documentation**
  * Added a changelog entry describing the fix for Nanite Torpedoes blocking other weapons.


lL1l1 (approved):
Tested and works. Fixes it not shooting every 4 shots at max; the actual effect depends on how the firing cycles lined up due to engage distances/layers and turret aim speed.

Other destroyers seem to have the same problem but I reckon their fire rate makes it less noticeable. I suspect it's an issue on a lot of units in general.

PR #7050, opened 2026-02-22
Bump up delay to end game back to 3 seconds

Garanas:

As posted on [Discord](https://discord.com/channels/197033481883222026/14723323538579169700), this fixes the bug where a campaign mission no longer shows the 'Operation Complete/Failed' dialog. It takes three seconds for the dialog to show, this is likely for dramatic effect. In https://github.com/FAForever/fa/pull/6965 the delay before the game ends was reduced from 3 seconds to 1 second, as a result the operation data was not synced and therefore the dialog would never show.

Run using the [init of coop](https://github.com/FAForever/fa-coop/blob/master/init_coop.lua#L5) in dev-mode by pointing to the repositories and confirm that the dialog shows with the changes made.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased post-victory wait time (1s → 3s) so campaign and co-op missions conclude cleanly and the “continue” dialog no longer reappears after the score/dialogue screens.
  * Added a shutdown safeguard to prevent unit removal/transfer during end-of-game processing, avoiding post-victory side effects.

* **Documentation**
  * Added a changelog entry describing the fix for missions not ending properly (fix #7050).

### DIFF
--- a/changelog/snippets/fix.7050.md
+++ b/changelog/snippets/fix.7050.md
@@ -1,3 +1,3 @@
 - (#7050) Fix campaign/coop missions not ending properly after completion
 
-The dialog to continue to the mission after dialogue and/or score screen pops up again.
+  The dialog to continue to the mission after dialogue and/or score screen pops up again.
--- a/changelog/snippets/fix.7050.md
+++ b/changelog/snippets/fix.7050.md
@@ -1,0 +1,3 @@
+- (#7050) Fix campaign/coop missions not ending properly after completion
+
+The dialog to continue to the mission after dialogue and/or score screen pops up again.
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -938,6 +938,10 @@ end
 -- would have blown up.
 EndGameGracePeriod = 10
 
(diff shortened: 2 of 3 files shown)


speed2CZ:
I think the delay was to ensure the final score is fully synced, since syncing doesnt happen every tick

speed2CZ:
iirc this wasnt the case just for the coop, but generally for all games. There was a bug when the final stats didnt match the game, since they were not fully synced. That was a long time ago, maybe things has changed a lot since then in how the score or game ending works, but thats what I remember.

Garanas:
Maybe we should hardcode the delay then, to prevent regression in the future?

lL1l1:
All the defeat/end-of-game stuff is overly complicated timing-wise because the victory thread and ai brain callback threads are too separated. The reason the delay was decreased was because the unit transfer thread could finishing waiting while the game is over and then it would kill all units because the game was over and no valid allies were left over. https://github.com/FAForever/fa/pull/6965#pullrequestreview-3580693572

lL1l1 (approved):

PR #7055, opened 2026-02-28
Balance navy and torp bomber interactions

lL1l1:
The aim of these changes is to make torpedo bombers more viable by making them more certain to snipe targets. If they can better snipe T2 navy then that should put an end to the Destroyer spam meta as it is possible to snipe low-HP destroyers which means that frigates will be farmed less, so frigates can play a better role overall.

The following changes are with this reasoning in mind:
- Remove splash from cruiser AA missiles and reduce flak range on Seraphim cruiser.
- Increase torpedo count for torpedo bombers so they better get through torpedo defenses.
  - Seraphim torpedo bomber already has 6 projectiles due to 2 projectiles splitting into 3 children each, so it was not changed.
  - I didn't touch UEF/Cybran flares because I don't know how to balance it across factions. Flares have poor targeting behavior compared to Aeon/Sera projectile antitorps. UEF/Cyb have extra antitorp on T2 support units while Aeon/Sera have it on T1 frigs and subs. The smallest change for flares is a large 3 -> 2 max deflections/flare which results in 6 -> 4 max deflections/volley firing.

The limit for these changes is to not bring torpedo bombers back to the state where they could snipe any cruisers mass-efficiently and then proceed to destroy an entire navy (except a Cybran frigate fleet). This old state caused a meta, especially in 1v1, where frigates + torpedo bombers/air control overwhelmingly dominated game-time because T2 navy required too large of an economy backing it to not get sniped by torpedo bombers while also being able to defend against high-statted frigates.

To prevent this, we have to consider several buffs that remain since then:
- Cruiser AA missiles still retarget, which is a huge buff vs low-HP torpedo bombers, especially for Cybran.
- Torpedo defenses are still effective against torpedo bombers. Through this mechanic, a large navy boosts the survivability of a cruiser, which means that you need more torpedo bombers per cruiser vs a large navy, unlike before where you only needed enough torpedo bombers to kill the cruisers, which was generally an easy mass investment because most of the cruisers are multi-purpose.

Additionally there is a buff to prevent that old state:
- Carriers are given splash damage to provide a further late-game AA solution, particularly against torpedo bombers which stack up easily.
  - Cybran carriers are given more damage radius since they don't have access to mobile shields, and shielded cruisers are stronger than their own faction's carrier.

## Additional context
[Discord post for the naval discussion](https://discord.com/channels/197033481883222026/1464889716481458277).

## Checklist

## Summary by CodeRabbit

* **Balance Updates**
  * Adjusted weapon damage and blast radii across multiple units.
  * Modified salvo sizes and salvo timing for torpedoes and other weapons.
  * Tweaked AA range/effectiveness for several cruisers and carriers.
  * Updated some weapon targeting/collision flags for select launchers.

* **Documentation**
  * Added changelog entry summarizing the above balance adjustments.


hotcheese4444:
I'm not on the balance team or anything. But this has so many moving pieces its absurd.

Increasing the projectile count up to five and reducing Torp cost back to 240M 4800E from 270M 8000E is a good start.

If torps are still useless, perhaps do the AOE swap of cruisers and carriers. Next balance patch.

Rhaelya (approved):
Everything looks in order

PR #7077, opened 2026-04-01
URA0103: resolve lingering bomb issue; fix failure to destroy on water impact

NoryGit:
Fixes two separate issues:
1. Bomb projectiles not getting destroyed on water impact (they travel through the water).
2. Bomb projectiles lingering and eventually falling through the map, thereby perpetually cluttering the screen.

Example of the bomb lingering issue: #26768413 5:50 at the northern rock base.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a failsafe to remove lingering cluster bomb projectiles after a short delay.
  * Projectiles now mark impact reliably on first hit; fragment spawning only occurs for non-air impacts.
  * A specific neutron cluster bomb variant is now destroyed on water contact.

* **Documentation**
  * Added a changelog snippet describing these fixes.

### DIFF
--- a/lua/cybranprojectiles.lua
+++ b/lua/cybranprojectiles.lua
@@ -506,21 +506,15 @@ CNeutronClusterBombProjectile = ClassProjectile(SinglePolyTrailProjectile) {
 
             if targetType ~= 'Air' then
                 local Random = Random
+                local child = self.ChildProjectile
 
-                local SpawnFragment = function(vx, vy, vz)
-                    local fragment = self:CreateChildProjectile(self.ChildProjectile)
-                    if fragment then
-                        fragment:SetVelocity(vx, vy, vz)
-                    end
-                end
-
-                SpawnFragment(0, Random(1,3), Random(1.5,3))
-                SpawnFragment(Random(1,2), Random(1,3), Random(1,2))
-                SpawnFragment(0, Random(1,3), -Random(1.5,3))
-                SpawnFragment(Random(1.5,3), Random(1,3), 0)
-                SpawnFragment(-Random(1,2), Random(1,3), -Random(1,2))
-                SpawnFragment(-Random(1.5,2.5), Random(1,3), 0)
-                SpawnFragment(-Random(1,2), Random(1,3), Random(2,4))
+                self:CreateChildProjectile(child):SetVelocity(0, Random(1,3), Random(1.5,3))
+                self:CreateChildProjectile(child):SetVelocity(Random(1,2), Random(1,3), Random(1,2))
+                self:CreateChildProjectile(child):SetVelocity(0, Random(1,3), -Random(1.5,3))
+                self:CreateChildProjectile(child):SetVelocity(Random(1.5,3), Random(1,3), 0)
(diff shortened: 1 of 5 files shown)


lua/cybranprojectiles.lua
```diff
@@ -491,16 +501,28 @@ CNeutronClusterBombProjectile = ClassProjectile(SinglePolyTrailProjectile) {
...
-            self:CreateChildProjectile(self.ChildProjectile):SetVelocity(0,Random(1,3),-Random(1.5,3))
-            self:CreateChildProjectile(self.ChildProjectile):SetVelocity(Random(1.5,3),Random(1,3),0)
-            self:CreateChildProjectile(self.ChildProjectile):SetVelocity(-Random(1,2),Random(1,3),-Random(1,2))
-            self:CreateChildProjectile(self.ChildProjectile):SetVelocity(-Random(1.5,2.5),Random(1,3),0)
-            self:CreateChildProjectile(self.ChildProjectile):SetVelocity(-Random(1,2),Random(1,3),Random(2,4))
+
+            if targetType ~= 'Air' then
+                local Random = Random
+
+                local SpawnFragment = function(vx, vy, vz)
```
speed2CZ: Whats the reason for this local function instead of the original chained one?

Are there issues of the projectile not being created, that you've added an extra check for that?

NoryGit: I preferred how the code looked, but you're right, I was being overcautious.

lua/cybranprojectiles.lua
```diff
@@ -483,6 +483,16 @@ CNeutronClusterBombProjectile = ClassProjectile(SinglePolyTrailProjectile) {
     OnCreate = function(self)
         SinglePolyTrailProjectile.OnCreate(self)
         self.Impacted = false
+        self.Trash:Add(ForkThread(self.FailsafeDestroyThread, self))
```
lL1l1: This should never be necessary, the engine gives projectiles a lifetime which will destroy it the same way.

lua/cybranprojectiles.lua
```diff
@@ -491,16 +501,22 @@ CNeutronClusterBombProjectile = ClassProjectile(SinglePolyTrailProjectile) {
     ---@param targetType string
     ---@param targetEntity Unit
     OnImpact = function(self, targetType, targetEntity)
-        if self.Impacted == false and targetType ~= 'Air' then
+        if self.Impacted == false then
```
lL1l1: an `Air` impact type occurs when the projectile is in the air and runs out of lifetime. The original OnImpact should destroy the projectile. So previously there was somehow an air collision type which meant the projectile never destroyed itself, and permanently stopped running the collision script with Impacted set to true, so it just stayed around forever.

lL1l1: I made a mistake in this explanation assuming that OnImpact always destroys the projectile. If the target entity does not allow collisions then OnImpact returns early. Units stop allowing collisions after they die. This leads to a much more reasonable explanation of the bomb impacting an engi directly, but that engi died to earlier bombs in the cluster, so `Impacted` is set to true, but Destroy is not called.

lL1l1 (approved):

PR #7089, opened 2026-04-11
Only build website when game files have changed

BlackYps:
The website build workflow fails when there are no balance snippets yet on deploy/fafdevelop or deploy/fafbeta. This is normal shortly after a release, but it looks concerning in the github UI.
It makes sense to only run this workflow when an actual deployment of game files has happened.
Or was there a specific reasoning to run on each develop update? @Garanas 

## Summary by CodeRabbit

* **Chores**
  * Updated documentation deployment workflow configuration to trigger from different source branches.

### DIFF
--- a/.github/workflows/docs-build.yml
+++ b/.github/workflows/docs-build.yml
@@ -22,7 +22,7 @@ name: Documentation - Build and deploy
 
 on:
   push:
-    branches: ["develop", "deploy/fafdevelop", "deploy/fafbeta"]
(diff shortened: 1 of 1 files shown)


BlackYps:
~~Hrm, I just noticed that the workflow (the old workflow file, currently on develop) did not trigger when I deployed to fafdevelop manually. This is strange.~~

Nevermind, I just forgot to update the staging branch, so no commits were actually deployed.

Garanas:
I don't recall why I set it to `develop`, but in any case `deploy/faf` should be in that list 😃

PR #7096, opened 2026-04-16
randomise default air unit elevation

cx-koraka:
Randomises aircraft default elevation upwards or downwards 5%. Total change in elevation value cannot be mroe than 5 in either direction.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Air units now spawn with small, bounded random variations to their flight elevation for more natural height differences.
  * Variation bounds are configurable and respect a per-unit opt-out flag so specific air types can keep a fixed elevation.
* **Documentation**
  * Added a changelog entry describing the randomized elevation behavior.

### DIFF
--- a/lua/sim/units/AirUnit.lua
+++ b/lua/sim/units/AirUnit.lua
@@ -47,12 +47,12 @@ AirUnit = ClassUnit(MobileUnit) {
         local maxElevationPercentageIncrease = 5
         local maxElevationPercentageDecrease = 5
 
-+        local blueprint = self.Blueprint
-+        local blueprintPhysics = blueprint and blueprint.Physics
-+        local originalElevation = blueprintPhysics and blueprintPhysics.Elevation
-+        if type(originalElevation) ~= 'number' then
-+            return
-+        end
+        local blueprint = self.Blueprint
+        local blueprintPhysics = blueprint and blueprint.Physics
+        local originalElevation = blueprintPhysics and blueprintPhysics.Elevation
+        if type(originalElevation) ~= 'number' then
+            return
+        end
 
         local elevationMultiplier = Random(100-maxElevationPercentageDecrease, 100+maxElevationPercentageIncrease) * 0.01
 
--- a/lua/sim/units/AirUnit.lua
+++ b/lua/sim/units/AirUnit.lua
@@ -20,11 +20,6 @@ AirUnit = ClassUnit(MobileUnit) {
     DestructionExplosionWaitDelayMax = 0,
     DestroyNoFallRandomChance = 0.5,
 
-    -- ELEVATION PARAMS
-    MaxElevationDelta = 5,
-    MaxElevationPercentageIncrease = 0.05,
-    MaxElevationPercentageDecrease = 0.05,
-
     ---@param self AirUnit
     OnCreate = function(self)
         MobileUnitOnCreate(self)
@@ -50,28 +45,23 @@ AirUnit = ClassUnit(MobileUnit) {
     RandomizeElevation = function(self)
         local blueprint = self.Blueprint
 
-        local blueprintCategoriesHash = blueprint.CategoriesHash
-
-        if (blueprintCategoriesHash["NORANDOMELEVATION"]) then
-            return
-        end
-
         local blueprintPhysics = blueprint.Physics
         local originalElevation = blueprintPhysics and blueprintPhysics.Elevation
         if originalElevation == nil then
             return
         end
 
-        local minVal = 1.00 - self.MaxElevationPercentageDecrease
-        local maxVal = 1.00 + self.MaxElevationPercentageIncrease
+        local blueprintAir = blueprint.Air
+        local minVal = 1.00 - blueprintAir.RelativeElevationDelta
+        local maxVal = 1.00 + blueprintAir.RelativeElevationDelta
         local elevationMultiplier = Random() * (maxVal - minVal) + minVal
 
         local newElevation
 
         if elevationMultiplier > 1 then
-            newElevation = math.min(originalElevation * elevationMultiplier, originalElevation + self.MaxElevationDelta)
+            newElevation = math.min(originalElevation * elevationMultiplier, originalElevation + blueprintAir.MaxElevationDelta)
         else
-            newElevation = math.max(originalElevation * elevationMultiplier, originalElevation - self.MaxElevationDelta)
+            newElevation = math.max(originalElevation * elevationMultiplier, originalElevation - blueprintAir.MaxElevationDelta)
         end
 
         self:SetElevation(newElevation)
--- a/lua/sim/units/AirUnit.lua
+++ b/lua/sim/units/AirUnit.lua
@@ -41,7 +41,7 @@ AirUnit = ClassUnit(MobileUnit) {
         end
     end,
 
-    --@paranm self AirUnit
+    --@param self AirUnit
     RandomiseElevation = function(self)
         local maxElevationDelta = 5
         local maxElevationPercentageIncrease = 5
--- a/lua/sim/units/AirUnit.lua
+++ b/lua/sim/units/AirUnit.lua
@@ -20,12 +20,17 @@ AirUnit = ClassUnit(MobileUnit) {
     DestructionExplosionWaitDelayMax = 0,
     DestroyNoFallRandomChance = 0.5,
 
+    -- ELEVATION PARAMS
+    MaxElevationDelta = 5,
+    MaxElevationPercentageIncrease = 0.05,
+    MaxElevationPercentageDecrease = 0.05,
+
     ---@param self AirUnit
     OnCreate = function(self)
         MobileUnitOnCreate(self)
         self.HasFuel = true
         self:AddPingPong()
-        self:RandomiseElevation()
+        self:RandomizeElevation()
     end,
 
     ---@param self AirUnit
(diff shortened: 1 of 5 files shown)


The-Balthazar:
image

you weren't supposed to include the `+`'s at the start of lines

BlackYps:
It has been brought to my attention that it would be good to be able to disable this behavior for certain units with a blueprint value. Something like adding the category `NORANDOMELEVATION`. This way mod units can disable the randomization if it's important for them.

lua/sim/units/AirUnit.lua
```diff
@@ -40,6 +41,32 @@ AirUnit = ClassUnit(MobileUnit) {
         end
     end,
 
+    --@param self AirUnit
```
speed2CZ: This needs to start with 3 --- to be a valid annotation

lua/sim/units/AirUnit.lua
```diff
@@ -40,6 +41,32 @@ AirUnit = ClassUnit(MobileUnit) {
...
+    RandomiseElevation = function(self)
+        local maxElevationDelta = 5
```
speed2CZ: You define this max delta, but you dont use in in the function

lua/sim/units/AirUnit.lua
```diff
@@ -40,6 +41,32 @@ AirUnit = ClassUnit(MobileUnit) {
...
+        local maxElevationPercentageIncrease = 5
+        local maxElevationPercentageDecrease = 5
```
speed2CZ: Why 2 different values for increase and decrease when they are the same?
Its not like a mod could easily adjust this, since its hardcoded in this function

lua/sim/units/AirUnit.lua
```diff
@@ -40,6 +41,32 @@ AirUnit = ClassUnit(MobileUnit) {
...
+
+        local blueprint = self.Blueprint
+        local blueprintPhysics = blueprint and blueprint.Physics
```
speed2CZ: Check for `blueprint` is not needed as it's always defined

lua/sim/units/AirUnit.lua
```diff
@@ -40,6 +41,32 @@ AirUnit = ClassUnit(MobileUnit) {
...
+        local originalElevation = blueprintPhysics and blueprintPhysics.Elevation
+        if type(originalElevation) ~= 'number' then
```
speed2CZ: You can just test if an elevation is defined, the type is set by the annotations and if someone decides to put anything else in it, it will most likely complain in the engine before the code gets to Lua

lua/sim/units/AirUnit.lua
```diff
@@ -40,6 +41,32 @@ AirUnit = ClassUnit(MobileUnit) {
         end
     end,
 
+    --@param self AirUnit
+    RandomiseElevation = function(self)
```
speed2CZ: It should be `Randomize` with `z`. The code is originally and mostly written in american english

lua/sim/units/AirUnit.lua
```diff
@@ -40,6 +41,32 @@ AirUnit = ClassUnit(MobileUnit) {
...
+        local maxElevationPercentageDecrease = 5
+
+        local blueprint = self.Blueprint
+        local blueprintPhysics = blueprint and blueprint.Physics
+        local originalElevation = blueprintPhysics and blueprintPhysics.Elevation
+        if type(originalElevation) ~= 'number' then
+            return
+        end
+
+        local elevationMultiplier = Random(100-maxElevationPercentageDecrease, 100+maxElevationPercentageIncrease) * 0.01
```
speed2CZ: This will produce 11 different elevation, the original, 5 lower, 5 higher ones. Is that intentional? Or was it suppose to be anything in the range?

speed2CZ (changes requested):

lua/sim/units/AirUnit.lua
cx-koraka: Fixed

lua/sim/units/AirUnit.lua
cx-koraka: If I moved these outside of the function, similar to the existing destruction params, would they be able to be changed by mods?
The original version of this had 30 and 0 as the max percentage increase and decrease respectively, this is a holdover

lua/sim/units/AirUnit.lua
cx-koraka: A smooth range is preferable.

Would Random(1-(maxElevationPercentageDecrease/100), 1+(maxElevationPercentageIncrease/100)) create a smooth range?

lua/sim/units/AirUnit.lua
speed2CZ: Yes, if they are saved on the class, mods can easily adjust them without touching the function

lua/sim/units/AirUnit.lua
speed2CZ: Idk, try that and you'll see. But if you give it two integers, it will pick a random integer from that range, not float.

lua/sim/units/AirUnit.lua
cx-koraka: Done. Hopefully this will make it easier to experiment whilst testing as well

lua/sim/units/AirUnit.lua
cx-koraka: I've got it working with the entire range now, couldn't tell by eye, so it was checked with debug logs

speed2CZ (approved):

speed2CZ:
I think this requires a green light from the balance team as well before it can be merged, as this will reduce the effect of AoE of AA weapons.

Garanas:
There's also this console command: `RandomElevationOffset`, I've not tested it. Maybe it is worth exploring that?

BlackYps:
I did some testing and the 5% change is enough that planes clip a lot less while still looking like being at the same general flight level. I like it!
I also noticed that - unrelated to this change - gunships fly up and down a lot when they shoot. So I am a lot less worried about balance implications now.

BlackYps:
From the discussions with the balance team so far it looks positive, but we should add the "NORANDOMELEVATION" category to transports, to make sure we don't run into weird behaviour with dropping or picking up units.

speed2CZ:
Would there really be weird behaviour with transports? The elevation at which they fly should not affect this in anyway, since they go into a different state for picking up units, using a different height. And if they fly over hill and take longer to land would be basically the same thing.

But Im not against the category being introduced.

lua/sim/units/AirUnit.lua
```diff
@@ -20,11 +20,17 @@ AirUnit = ClassUnit(MobileUnit) {
     DestructionExplosionWaitDelayMax = 0,
     DestroyNoFallRandomChance = 0.5,
 
+    -- ELEVATION PARAMS
+    MaxElevationDelta = 5,
+    MaxElevationPercentageIncrease = 0.05,
+    MaxElevationPercentageDecrease = 0.05,
+
```
lL1l1, line 27: If these parameters are never changed in runtime it would be better to assign them in blueprint loading (blueprints-units.lua).
Storing it as an absolute change in the blueprint may also be more convenient for authors who edit their blueprint specifically. You can use the percentage multiplication to calculate that absolute value when it doesn't exist.
Storing it in the blueprint has an added benefit of not requiring a new category to exclude units from the mechanic, since the author can just set the random offset min/max values to 0.

lL1l1 (commented):
Balance-wise this is fine since it has quite a low impact:
- Bombers fly relatively low so they won't move much with 5%, and that small movement transforms into an even less significant bomb drop time change (~0.1s out of 2.26s for ahwassa, the highest flying bomber, and <0.1s changes are undetectable because the game runs in 0.1s ticks).
- ASF at 22 elevation will move up/down 0.9 with 5%, so the lowest flying asf may block the 1.5 radius splash SAM from hitting the highest flying ASF, but that's unlikely and will easily average out over the large numbers built.

Perhaps instead of using a percentage, a proportion of the hitbox could be used as an approximate for the mesh's size. This may result in even smaller variations with even less balance impact while keeping the visual impact.
It would also reasonably restrict the up/down movement for high-flying units.

lua/sim/units/AirUnit.lua
BlackYps: I agree that per-unit control can be beneficial. I had a look at blueprints-units.lua and I am a bit confused. This seems to mostly alter and process some blueprint values, but very little actual default values. Where are the default values to the various blueprint fields specified that are used when a specific unit doesn't specify that field? I feel like the new variable definitions should go there.
I'm not yet convinced that storing it as an absolute change is really more convenient.

BlackYps: Similarly, I don't like switching to a proportion of the hitbox. I understand the idea, but I don't see the need to make the variations even smaller, the point of this PR is to create visible variations in the first place, and it could lead to weird effects with bulky units like gunships. Or the Soulripper for that matter. I think just using a flat percentage makes it easier to predict the results.

lL1l1: >  Where are the default values to the various blueprint fields specified that are used when a specific unit doesn't specify that field?

Most of the default values are in the engine. Anything FAF adds should be in blueprint loading in Blueprints.lua (old organization) or the files that edit specific parts like blueprints-units.lua (better organization).

> Similarly, I don't like switching to a proportion of the hitbox. I understand the idea, but I don't see the need to make the variations even smaller, the point of this PR is to create visible variations in the first place, and it could lead to weird effects with bulky units like gunships. Or the Soulripper for that matter. I think just using a flat percentage makes it easier to predict the results.

I was only suggesting that it could be a smaller variation since you would have more direct control over the visual outcome of the assigned values. The soul ripper is a good outlier to point out: there should be some limits to the values that are automatically assigned based on hitbox. But I don't think including some boundaries is much different from the max elevation delta that the current implementation has.

BlackYps:
@lL1l1 please have a look now if the blueprint loading is now done as you would expect it

PR #7105, opened 2026-05-08
Add custom text truncation support, and improve scoreboard legibility

Lightningbulb2:
I wanted to improve scoreboard legibility.

I started by separating game speed, and game quality into their own controls to make modding and future changes easier. I added dropshadows, tooltips, and right aligned the ratings while making sure to crop player names if necessary. This led to the-

Implementation of SetTruncateText() which allows custom trailing strings like "...", "-", or even "NOFIT". This is executed through SetClipToWidth() and only activates if a truncation string is set.

To do this cleanly, and in a reusable way, I split *SetText()* into ***SetText()*** and its implicitly called ***SetDisplayText()***.

**SetText()** - controls the source text and behaves practically the same as before

**SetDisplayText()** - sends text changes to the engine without changing the Text object's internal string. This allows for better mod and source code capabilities for fancier text. The first example being custom text truncation.

I loaded up private AI matches and replays to check for console and visual errors. I got screenshots:

**Original for reference** (plus rank cropping issue)
Original

---

Game start with lots of players
GameStartExample

---

Full scoreboard
NEWSCOREBOARD

---

Replay scoreboard
REPLAY

---

Show full player names after hovering for a moment (in-case of cropped names)
CROPPEDHOVER

---

Tooltips added for new players
gameQuality

---

gameSpeed

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added game speed display (requested vs. actual) and game quality to the score overlay.
* **Improvements**
  * Refreshed score and mini-score layouts for improved legibility, alignment, and tooltip support (including hover previews for full player names).
  * Upgraded text truncation so the displayed value auto-updates based on available width, with a customizable truncation suffix.


lua/maui/text.lua
```diff
@@ -33,6 +33,38 @@ Text = ClassUI(moho.text_methods, Control) {
...
+        self._fullText = nil
+
+
+        --- Direct Engine SetText() that changes what text is displayed
+        ---@type function
+        ---@param str string | number
+        self.SetDisplayText = self.SetText
+
+
+        --- FAF extensible SetText() that uses SetDisplayText() but can retain it's original text for fancy text display setups like truncation
```
lL1l1:
```suggestion
        --- FAF extensible SetText() that uses SetDisplayText() but can retain its original text for fancy text display setups like truncation
```

lua/maui/text.lua
```diff
@@ -33,6 +33,38 @@ Text = ClassUI(moho.text_methods, Control) {
...
+        --- Direct Engine GetText() for getting the current displayed value
+        ---@type function
+        ---@return string
+        self.GetDisplayText = self.GetText
+
+        --- FAF extensible GetText() that retrieves raw original text that isn't modified for display 
+        ---@return string
+        self.GetText = function (self)
+            return self._fullText
+        end
```
lL1l1: Why not make these part of the class? If you don't know how to, you can do it like this:
```lua
Text = ClassUI(moho.text_methods, Control) {
    SetDisplayText = moho.text_methods.SetText,
    SetText = function() end,
    GetDisplayText = moho.text_methods.GetText,
    GetText = function() end,
}
```

lua/maui/text.lua
```diff
@@ -41,15 +73,58 @@ Text = ClassUI(moho.text_methods, Control) {
...
 
+    --- Internal function to fit the truncation string inside the max width
+    _applyTruncation = function(self)
+        local maxWidth = self.Width()
+        if maxWidth <= 0 then
+             return
+        end
+
+        -- ellipsis is the trailing '...' on truncated text
+        local ellipsis = self._truncationText
```
lL1l1:
```suggestion
        local truncText = self._truncationText
```

lua/maui/text.lua
4z0t: Make a separate class for Truncatable text. You are polluting Text class with functionality used by scoreboard only.

4z0t (changes requested):

lua/maui/text.lua
Lightningbulb2: There are really 2 changes here. A text rendering feature (better truncation), and a text architecture change (display text) which the truncation relies on.

### 1. Better Text Truncation
It's only used by the scoreboard because it didn't exist to be used for other things yet.

Text can already be truncated, but this simply allows the option for setting _**how**_ it truncates it. You may want a dash, dots, or warning text (_for texts that shouldn't be truncated but would be hard to notice if cut off at the right spot_).

It's more of a small extension to SetClipToWidth() than a new special text type.

SetTruncationText() is just setting the trailing characters, if there was any confusion there. Maybe SetTrailingCharacters() would be better?


### 2. SetText() and SetDisplayText()

The separation of the full text and the "displayed text" sent to the engine, is much more useful than just supporting truncation text. It allows much more elaborate text animation/visual effects for modding and development because it's no longer on you to keep track of the old state. Plus anything that tries to GetText() will get it's real value instead of the partial text effect value that may have been setup. GetDisplayText() would be used for getting what the engine sees.

---

Hopefully that makes sense. Thanks for the review!

4z0t: Can I get a response by the **Human** being? Still i insist on creating a separate class.

lua/maui/text.lua
```diff
@@ -33,6 +33,38 @@ Text = ClassUI(moho.text_methods, Control) {
...
             self:SetNewColor(var())
         end
+        self._truncationText = ""
+        self._fullText = nil
+
+
+        --- Direct Engine SetText() that changes what text is displayed
+        ---@type function
+        ---@param str string | number
+        self.SetDisplayText = self.SetText
```
4z0t: Each Text instance has this field, not its meta.

lua/maui/text.lua
```diff
@@ -33,6 +33,38 @@ Text = ClassUI(moho.text_methods, Control) {
...
+
+
+        --- FAF extensible SetText() that uses SetDisplayText() but can retain it's original text for fancy text display setups like truncation
+        ---@param text string | number
+        self.SetText = function (self, text)
```
4z0t: Each Text instance has this field, not its meta.

lua/maui/text.lua
```diff
@@ -33,6 +33,38 @@ Text = ClassUI(moho.text_methods, Control) {
...
+                if self._initialized then
+                    self:_applyTruncation()
+                end
+            end
+        end
+
+        --- Direct Engine GetText() for getting the current displayed value
+        ---@type function
+        ---@return string
+        self.GetDisplayText = self.GetText
```
4z0t: Each Text instance has this field, not its meta.

lua/maui/text.lua
```diff
@@ -33,6 +33,38 @@ Text = ClassUI(moho.text_methods, Control) {
...
+
+        --- FAF extensible GetText() that retrieves raw original text that isn't modified for display 
+        ---@return string
+        self.GetText = function (self)
```
4z0t: Each Text instance has this field, not its meta.

4z0t (changes requested):

lua/maui/text.lua
Lightningbulb2: I typed all that myself, dome to keyboard :(

The last line "Hopefully that makes sense." was about **my writing**, not some AI garbage. I try to include a thanks when I can because this is volunteer work, and programming feedback tends to be heavy on critique so the "thanks" is as much for you as it is for me (to not take it personally).

---

Aaaanyway, yeah, you're right about the function instancing. That modding override trick is great for single function rewrites, but I understand avoiding it here.

How about this solution in the body rather than inside __init? That would prevent all the instancing right?

``` lua
---@class Text : moho.text_methods, Control, InternalObject
Text = ClassUI(...)

    __init()..


    --- Direct Engine SetText() that changes what text is displayed
    ---@type function
    ---@type fun(self: Text, str: string | number)
    SetDisplayText = moho.text_methods.SetText,

    --- Direct Engine GetText() for getting the current displayed value
    ---@type function
    ---@return string
    GetDisplayText = moho.text_methods.GetText,

    --- FAF extensible SetText() that uses SetDisplayText() but can retain it's original text for fancy text display setups like truncation
    ---@param text string | number
    SetText = function(self, text)
        self:SetDisplayText(text)
        self._fullText = text
        if self._truncationText ~= nil and self._truncationText ~= "" then
            if self._initialized then
                self:_applyTruncation()
            end
        end
    end,

    --- FAF extensible GetText() that retrieves raw original text that isn't modified for display
    ---@return string
    GetText = function(self)
        return self._fullText
    end,

    --- Sets custom truncation trailing characters like "..." or "-". Set to "" to disable.
    ---@param text string | number
   SetTruncationText = function(...)...

```


Also, should I keep it as "truncation text" or call it "trailing characters" instead, for clarity?

` self._truncationText` would then be `self._trailingCharacters`

`SetTruncationText()` would then be `SetTrailingCharacters()`

---

Claude.ai showed me I could use "moho.text_methods.SetText" instead of "self.SetText()"
 
Although it defined local variables outside the class to reference it first and then assign the Display functions in the __init(), but I found that wasn't necessary.

(**Hand written reply**)

4z0t: Sorry if I offended you. I just don't like when people drop a lot of text especially made with ai not even understanding what's it about.

As I said, you are modifying class that is used everywhere in the code and any side effects are very unwanted. Please just create a separate class that inherits from Text. Like TruncatableText and instantiate it in scoreboard. It will make it easier to control scope of the change and will be easier to maintain.

Lightningbulb2: The latest commits remove a lot of unnecessary code. It updated fine even when I was messing with the text values in-game.

The only 2 checks that happen now are of the boolean `self._truncationEnabled` inside both `SetText()`, and the `OnDirty` from `SetClipToWidth()`

If it's not enabled, nothing else runs, so it shouldn't bother the rest of the backend text.

The additional fields are

Bool: `self._truncationEnabled` (default: false)
String: `self._truncationText` (default: "...")
String: `self._fullText` (default: nil)

Aside from the truncation stuff, I would still want to keep the main Text changes. So adding another class would just move the truncation stuff for not much gain, plus more inconvenience because any new text features would have to be added to both.

Ignoring all the truncation stuff, this is ALL the core changes to how Text operates

``` lua
   __init = function(...)

     ...

     self._fullText = nil
  end,


    SetDisplayText = moho.text_methods.SetText,

    GetDisplayText = moho.text_methods.GetText,

    SetText = function(self, text)
        self:SetDisplayText(text)
        self._fullText = text
    end,

    GetText = function(self)
        return self._fullText
    end,

```

So while I understand being hesitant to make sweeping changes that might break things. I've been extremely careful to keep things compatible and only those few lines actually matter for the rest of the repository. The other changes in Text.lua are from the optional text truncation, which I've kept as unobtrusive as possible so it can stay as a standard text feature.

If you can think of any side effects of adding this, let me know, otherwise we'll just have to find out if it breaks.

Thanks for reading

PR #7106, opened 2026-05-08
Generate coordinates instead of loading them

PreciseBump38:
## Summary
Instead of storing a clearly code generated set of coordinates and taking up over 1MB (mostly whitespace) in source code, generate the coordinates when we import the file.

Also use references instead of duplicating each set of coordinates for each faction.

## Pros/Cons
Cons:
- We use extra CPU cycles when we first import the file

Pros:
- We save 1MB disk space for developers (less for users because it compresses really well)
- We save load times as less has to be read from disk
- We use fewer CPU cycles parsing lua code
- VSCode doesn't have to tell me "Too large file: fa\lua\basetemplates.lua skipped. The currently set size limit is: 500 KB, and the file size is: 1166.749 KB."

## Testing
Across 5 tests I got the following load times:
- Original version: 0.031453132629395, 0.031542778015137, 0.031722068786621, 0.032179832458496, 0.077125549316406
- New version: 0.0019760131835938, 0.0020065307617188, 0.0020112991333008, 0.0020360946655273, 0.0020914077758789
Taking the median import time for old vs new gives a speedup of 93.66% for an absolute saving of 0.0297 seconds.
I'm quite sure savings will vary by machine depending on CPU and SSD speeds.

In terms of parity to the original version of the file, there are some differences:
1. There are 4 value differences (as noted in the code)
    - MovedTemplates7[2][2][1][22] and MovedTemplates7[2][2][1][23] are swapped
    - MovedTemplates8[1][2][1][23] and MovedTemplates8[1][2][1][24] are swapped
    - These could be made the same as the original if we really wanted, but I assume it was a mistake that they weren't consistent with all the other unit lists already.
2. The per-faction arrays are now references to the same array in memory, not duplicates.  If people are doing side effects through the basetemplate coordinate systems they may get unexpected behaviour.  But also, why would you do that??

## Don't forget
Some thought may also be necessary for how this affects mods, e.g. nomads.  I haven't looked into or thought about it.

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Updated changelog: BaseTemplate tables are now generated at import time rather than parsed from source files.

* **Bug Fix / Behavior**
  * Improved template import handling so computed coordinate data is preserved more accurately during template creation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FAForever/fa/pull/7106?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Garanas:
I think in practice (the bottom line) is not different in-game, is that correct? It will still load the same amount of memory*. Just computed now as a side effect of importing it, instead of pre-computed on disk.

PreciseBump38:
Yeah I can't think why it would break other mods but wanted just to sanity check that.  I had a quick look at the nomads mod (was it last updated 2024??) and they have a separate function for loading in this same data for a 5th faction, but again that keep working fine.

Some other things this code makes me think about (for a separate/future PR):
- All the coordinates are 3-tuples with 0 as the final number - if its always 0 why include it?
- There are (almost) no differences between factions - so lets not index by faction.
- Why are some search patterns spirals and others not?  Can they not be the same?
- Keeping duplicates of the same coordinates but offset by a fixed amount feels like an unnecessary memory hog.
- Could we just replace the file with a generator function instead of a datastructure?

Garanas:
> * All the coordinates are 3-tuples with 0 as the final number - if its always 0 why include it?

It is included because engine functions of the brain expect the third parameter to exist. The function that determines a valid build location, for example. Otherwise you'd have to convert the table each time you use it before you pass it through.

The first two arguments are the X/Z plane (Y is up in Supreme Commander), the third argument is about rotation in case anyone reading this is wondering that 😃.

edit: I think a lot of your questions make sense, I've asked myself the same questions in the past. Unfortunately, I don't know how far you can go without introducing subtle bugs and/or flatout breaking existing functionality. 

In an ideal scenario we would also apply a structure of arrays, instead of an array of structures. This change alone would shave off tens of kilobytes of memory. But it would be incompatible with the exposed engine functions and other sim-related logic that interact with this.

lL1l1 (commented):
I was looking at usages of base templates and I primarily found it to be used in CAiBrain:FindPlaceToBuild. The annotation for that function uses parameter of type `BuildingTemplate`. 
1. This means that the new alias `BaseTemplateType` is not necessary. The `BuildingTemplate` object is what we are generating.
2. `BuildingTemplate` uses `Vector2` instead of `Vector`. I suspected that this means the engine accepts Vector2, so I removed the 0 at the end of all the vectors and it seems to work with my debug logging and running a game with AI Adaptive.

It needs to be noted that type `UnitList` is names defined in `BuildingTemplates.lua`.

lL1l1:
Also from the mods I've checked on my disk:
- RNGAI runs its own separate BaseTemplates.lua file
- BrewLAN uses table.find and table.insert to modify the templates
- SCTAFix adds tables under faction indices 6 and 7 (it is a total annihilation faction mod).

Nomads you tested that it works. It's a featured mod so I think it just overrides the file.

lL1l1 (approved):

PR #7111, opened 2026-05-09
Battleships: Disable fireclock rendering on aft turrets

NoryGit:
The aft turret usually fires after the forward turret(s), which causes the fire clock to reset when it fires, making it useless.

## Checklist
- [ ] ~Changes are annotated, including comments where useful~
- [ ] ~Request 2-3 reviewers from the [list of reviewers and their areas of knowledge](https://github.com/FAForever/fa/blob/deploy/fafdevelop/CONTRIBUTING.md#reviewers).~

### DIFF
--- a/changelog/snippets/fix.7111.md
+++ b/changelog/snippets/fix.7111.md
@@ -1,0 +1,1 @@
+- (#7111) Fixed an issue where battleships' aft turrets reset the fire clock.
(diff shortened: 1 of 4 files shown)


NoryGit:
@lL1l1 @Rhaelya This change is harmless and can be merged for the patch next week. It is cherry-picked from #7101.

lL1l1:
What if we made it so that only the oldest reload clock is shown, and once it finishes the next oldest is shown?

lL1l1:
I'd rather have a more generalized solution like only setting the progress from the oldest fire clock.

PR #7119, opened 2026-05-15
Track Marks Fix

Saver27:
**Description**

This fix resolves an issue where incorrect track marks were occasionally generated due to a problematic albedo texture. Replacing the first fix, which doesn't work 100% (Tread marks glitch fix #7021) 

**Forum Reference:**
https://forum.faforever.com/topic/9739/unit-treadmarks-sometimes-defective/8

**Affected Texture**

TreadMarks/tank_treads06_albedo

**Affected Units**
UEL0401 — Fatboy (UEF)
URL0111 — Viper (Cybran)
URL0304 — Trebuchet (Cybran)

**Cause**

The issue occurred exclusively with the following tread mark albedo texture: _tank_treads06_albedo_

Here is an image of the error from May 11, 2026.
image

**Solution**

The affected units now use: _tank_treads03_albedo_

This alternative does not produce the issue and still matches the affected units visually.
Before-and-After Images for Comparison
image
image

**Additional Note About the Fatboy**

Since the Fatboy uses four tread drives, it would theoretically be possible to add two additional track marks.

However, this approach was intentionally not used because the additional tracks visually overlap and result in an unclean appearance.
image

The current implementation therefore only uses the rear treads as the reference for the generated track marks.
image

Here is another picture showing how the tracks were set:
image

## Summary by CodeRabbit

* **Style**
  * Updated tread-mark visuals for three units to improve detail during land movement.
  * Adjusted tread spacing, size and placement for more consistent markings.
  * Introduced supplemental tread entries and distinct left/right offsets for select units to enhance realism and visual consistency of movement effects.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FAForever/fa/pull/7119?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


Saver27:
**Small additional note regarding TreadMarks/tank_treads06_albedo:**

While I personally think this version looks visually better, I would still avoid using it as long as the described issue can occur.

The bug noticeably breaks immersion during gameplay and creates distracting visual artifacts during matches.

BlackYps:
looking at the textures I noticed that tank_treads06_albedo has a bigger size (512) than the others (256). That is probably the culprit. For the smaller units we could probably just scale these textures down.
For the fatboy this would probably make the texture too low resolution, but the way the texture space is used is extremely inefficient, so we could improve that.
First of all we could easily double the amount of marks (indicated in green) and then crop the texture as indicitated in yellow. This would mean that we have a separate texture for left and right, but as we have more marks on one texture, we can reduce the interval accordingly. This overall avoids an increase of the amount of tread mark splats, which is important because there is an engine limit.
The smaller units would also benefit from doubling the amount of marks in the texture.
Would you be up to implement this plan?
grafik

Saver27:
Your suggestion sounds significantly better than the first attempt above. Therefore, I would like to try to implement it.

BlackYps:
Can you add screenshots or a short video of your solution?

Saver27:
> Können Sie Screenshots oder ein kurzes Video Ihrer Lösung hinzufügen?

Currently, the fix is on hold because the texture resolution is not good enough. The user Fichorn offered to improve the texture for the track tread, so I would like to wait for his work before continuing with the fix.

PR #7128, opened 2026-05-23
Fix typo: RedRock → Redrock in weatherdefinitions.lua (#7118)

### LINKED ISSUE #7118: [Bug]: Typo in MapWeatherList
### Describe the Bug

The list uses "Red**R**ock".
https://github.com/FAForever/fa/blob/3868ca880acbfdaf44dda798d594092158add45d/lua/weatherdefinitions.lua#L62

But in fact, 'MapWeatherList' uses "Red**r**ock".
https://github.com/FAForever/fa/blob/3868ca880acbfdaf44dda798d594092158add45d/lua/weatherdefinitions.lua#L243

### Reproduce the bug

1

### Screenshots

 2

### Additional context

 3

niuniu319:
## Fix FAForever/fa#7118

MapWeatherList used `'RedRock'` but the actual definition key is `Redrock`. Fixed to match.

Closes #7118

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed the casing of the "Red Rock" weather style so it is now consistently recognized and listed in weather selectors and maps, ensuring the style displays and applies correctly across the app.

### DIFF
--- a/lua/weatherdefinitions.lua
+++ b/lua/weatherdefinitions.lua
@@ -240,7 +240,7 @@ MapWeatherList = {
         CirrusMedium4096 = CirrusMedium4096Definition,
         CirrusHeavy4096 = CirrusHeavy4096Definition,
     },
-    Redrock = {
+    RedRock = {
         LightStratus = {
             {
                 EmitterBasePath .. 'weather_stratus_09_emit.bp',	-- 40x40		Med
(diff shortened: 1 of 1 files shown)


lL1l1 (changes requested):
The MapStyleList value should be `RedRock` to be in line with the terrain style's name. Instead, the key for the MapWeatherList should be changed.

lL1l1 (approved):

lL1l1:
@BlackYps You're more familiar with mapping, this is the correct fix considering the mapping workflow, yes?

BlackYps:
Interesting, I am not familiar with this file at all, but RedRock is the spelling as it is used for the map textures for example

PR #7132, opened 2026-06-02
Add moderation exception to self destruct

Garanas:

Adds the exception that when you self destruct a selection with one or more ACUs it will only self destruct the ACUs. 

Launched the game and verified that the ACU is isolated when I try to self destruct all my units.

## Additional context

For this fortunate enough to view it:

- https://discord.com/channels/197033481883222026/1511365928745439272

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * In Full Share mode, self-destructing a selection now targets only selected Command Units when any are present, leaving other selected units unaffected.
  * Self-destruct commands now validate that a unit selection is provided and confirm the action is permitted for the requesting owner.
  * When no Command Units are included in a Full Share selection, self-destruct behavior continues to apply to the original selection.


lua/SimCallbacks.lua
```diff
@@ -179,7 +179,34 @@ Callbacks.UpdateMarker = SimPing.UpdateMarker
...
+---@param data ToggleSelfDestructData
+---@param units Unit[]
+Callbacks.ToggleSelfDestruct = function(data, units)
+    -- prevent malformed input
+    if (not units) or (table.getn(units) == 0) then
+        return
+    end
+
+    -- prevent abuse
+    if (not data) or (not data.owner) or (not OkayToMessWithArmy(data.owner)) then
```
lL1l1, line 191: Using user-defined Args in the `data` table cannot be relied upon for security.

When I look into it further, the verification for this function is ending up a bit all over the place:
1. `data.owner` is verified with `OkayToMessWithArmy` in this callback and the import function.
2. `data.owner` is confirmed to not be -1 (observer) in the import function. Which I guess prevents observers from self-destructing units in cheat games? But `data.owner` is `GetFocusArmy` from UI side, and observers can't select units so they can't trigger the callback with any selection to act upon while also sending an owner of -1.
3. `units` is checked with `SecureUnits`, which loops over units and checks `OkayToMessWithArmy(unit.Army)` (it also has a part converting entity ids into entities but that isn't relevant here). This is all sim-side info and the units list is from the engine UI code, so it is secure. The issue is the duplication where this `OkayToMessWithArmy(unit.Army)` is used in your category filter, and the imported function (twice at that).

Overall, the import function has the final results and always checks `OkayToMessWithArmy(unit.Army)`, so it works, but it seems like the rest of the checks can be replaced. `data.owner` can instead be an early exit using `GetCommandSourceArmy` or `GetFocusArmy` sim-side. Checks for `OkayToMessWithArmy(unit.Army)` can be replaced with SecureUnits in the main callback function replacing the units variable, and then sending it to the imported function.

lua/SimCallbacks.lua
```diff
@@ -179,7 +179,34 @@ Callbacks.UpdateMarker = SimPing.UpdateMarker
...
+    -- moderation rule: if you self destruct with one or more ACUs in the selection when playing full 
+    -- share, then you only self destruct the ACUs. This does not make it impossible to abuse, but it 
+    -- does introduce a simple guardrail. 
+    if ScenarioInfo.Options.Share == "FullShare" then
+        local commandUnits = EntityCategoryFilterDown(categories.COMMAND, SecureUnits(units))
+        if table.getn(commandUnits) > 0 then
+            import("/lua/selfdestruct.lua").ToggleSelfDestruct(data, commandUnits)
+            return
+        end
+    end
```
lL1l1, line 204: This is trivial to bypass with UI mods splitting the user's order into two callback calls, so why not implement it in the UI layer? Then it's easy to have a handler with proper user feedback.

lua/SimCallbacks.lua
Garanas: This is just me pulling the logic from the SelfDestruct function to the boundary, see also lines 35: `if data.owner ~= -1 and OkayToMessWithArmy(data.owner) then`.

I personally do not think that the `units` list needs to be checked, since it's the engine that appends it. Unless we also want to guard against hacked clients, where you can append arbitrary units.

Note that I intentionally did not mess with the existing functionality, since that's not what was requested.

lua/SimCallbacks.lua
Garanas: It's also trivial to identify that this happened when parsing the replay. Malicious intent is then confirmed, easier to ban in my point of view. 

You're free to also implement something in the UI layer, by all means.

lL1l1 (approved):

PR #7135, opened 2026-06-05
Allow docking/building of mercies on carriers and air staging.

Nuggets75:
**Intended Change**
Remove the CANNOTUSEAIRSTAGING category from the Mercy (DAA0206) so it can use air staging facilities and aircraft carriers, like most other air units.

**Reasoning**
Since the mercy rework, the mercy is now an anti t1/t2 army / structure / buildpower unit and no longer a snipe-unit so I see no reason to lock it out of carriers. I believe the main reason it was locked out was to limit their attack range with low amount of fuel before the rework. Now the mercy, while good, has an incredibly small window of opportunity that can be expanded with this change.

**Expected behavior**
The Mercy can now land on air staging and dock on / be built by aircraft carriers. No change to its health, damage, cost, build time, or any other stat.

## Summary by CodeRabbit

* **Gameplay Changes**
  * Updated air-staging interaction rules: this unit can now use air-staging right-click interactions.
  * Expanded command capabilities to include transport/docking/guard/patrol.
  * Increased fuel recharge rate.
  * Improved primary weapon target evaluation with more frequent re-checks and a larger tracking radius.
* **Documentation**
  * Added an explanatory note to engine category documentation regarding air-staging behavior.

### DIFF
--- a/engine/Core/Categories.lua
+++ b/engine/Core/Categories.lua
@@ -55,6 +55,7 @@ categories = {
     BUILTBYTIER3COMMANDER = categoryValue,
     BUILTBYTIER3ENGINEER = categoryValue,
     BUILTBYTIER3FACTORY = categoryValue,
+    --- Used by engine to disallow air units to right click into transports/air staging.
     CANNOTUSEAIRSTAGING = categoryValue,
     CANTRANSPORTCOMMANDER = categoryValue,
     CAPTURE = categoryValue,
--- a/units/DAA0206/DAA0206_unit.bp
+++ b/units/DAA0206/DAA0206_unit.bp
@@ -116,7 +116,10 @@ UnitBlueprint{
     General = {
         CommandCaps = {
             RULEUCC_Attack = true,
+            RULEUCC_CallTransport = true,
+            RULEUCC_Dock = true,
             RULEUCC_Move = true,
+            RULEUCC_Patrol = true,
             RULEUCC_RetaliateToggle = true,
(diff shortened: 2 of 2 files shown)


lL1l1:
I tested it and found that nothing actually worked. I implemented the proposed changes of docking in air staging and building in carriers, but this necessarily comes with the side effect of mercy recharging fuel while on the ground. It's the standard 1s fuel per 2s landed time since that rate is multiplied by air staging when refueling, and the unit needs to refuel fully to leave air staging, so a low base refuel rate is unacceptable. I also needed to add the guard order for mercies to be able to leave carriers using the deploy command, which issues a guard (assist) command on the ground where you click the deploy order.

Since it can now refuel, I thought it made sense to add the patrol ability. 

When I was testing patrol, I discovered that it wasn't working because of a long target check interval, low tracking radius, and missing `AlwaysRecheckTarget = true`, so I fixed all of that. This also fixes the attack behavior on guard and attack move commands, so it is quite necessary for those commands, not just patrol.

Garanas:
> When I was testing patrol, I discovered that it wasn't working because of a long target check interval, low tracking radius, and missing `AlwaysRecheckTarget = true`, so I fixed all of that. This also fixes the attack behavior on guard and attack move commands, so it is quite necessary for those commands, not just patrol.

This was not a bug as far as I am aware. These parameters were intentional so that the Mercy does not trigger on a tech 1 scout that happens to walk by. This was, as far as I am aware, by design 😃 .

PR #7141, opened 2026-06-19
Partial SACU Rebalance

lL1l1:
> I think the current HP-focused combat + engineering SACUs can be an interesting addition into the current balance as a long-term value type of unit vs T3, they could be a bit stronger vs T4, and the engi upgrades could be more efficient. None of this requires whole identity changes or new upgrades, just changing some numbers.

From this [discord post](https://discord.com/channels/197033481883222026/1508956770809282640)

Summary of changes:
1. Quantum Gateways
  a. Quantum gateway cost/buildpower buffed to lower barrier of entry into SACU tech.
  b. QG HP reduced since it is a "high tech" structure
  c. QG adjacency massively buffed to allow for long-term SACU production.
2. General SACU Changes
    1. Base SACU cost reduced by 400 mass
        1. This required shifting around Cybran SACU's mass costs so that SACU with just the dmg/range upgrade is not overpowered.
        2. RAS cost increased by 100 mass so that 5 Aeon SACU sacrifice into GC still.
    2. Buildtime generally reduced to buff QG mass drain from 20-27 mass/s to 30-34
    3. Energy ratios reduced to ~18 for non-combat presets, ~22-26 for combatants, ~32-35 for rambos, and 40 for cloak/shield
    4. Energy upkeeps reduced 30%
    5. Engineer presets heavily buffed in cost to make them a more viable buildpower source. They're still significantly harder to build than even lvl 3 hives, but it's much easier than before. Mainly it is difficult to make the mass/buildtime costs of the base sacu + engi upgrade compete with hives/engis with only 98 buildpower.
       1. Sera TML cost increased by 300 mass to counteract one of the double buffs from engi and base sacu cost buffs. Sera TML preset is quite a powerful tool in the Sera roster already.
3. Specific SACU Changes:
   1. UEF Bubble Shield prerequisite of the personal shield is removed, buffing the shield preset by 2k mass. Due to the cost decrease, the bubble shield is also rebalanced to be more similar to mobile shields.
   2. Seraphim overcharge upgrade cost reduced to 1k mass, since it needs a massive cost in infrastructure to be supported.
   3. Added RAS to Seraphim on the right arm (OC + tele arm) with a RAS preset, and reduced the base cost and resource production of the SACU to standard values.
   3. Seraphim shield upgrade moved to right arm (OC + tele arm).
      1. This nerfs Seraphim Teleport SACU, which is extremely strong with 59500 HP and 400 DPS being harder to kill than a telemazer before it kills your SMD. Players will have to get used to sending multiple SACU to kill more important targets now. A nano tele SACU is only 19150 mass, and the energy infrastructure needed to build/teleport it is reusable, so for killing a 70-79k mass arty or 220-250k mass game ender, that is a fair price (arguably it is still too powerful, as teleport has no good counters besides spamming shields trying to get them in between the target and the SACU gun).
      2. The Rambo preset was previously useless because it was shield + nano + overcharge, so it had no range from the sensors upgrade to be able to use overcharge. The new Rambo preset is shield + sensors + nano. It is similar to the Nano Combatant in that it only has 400 single-target DPS, but its power/mass relies more on HP than regen, so it does better late-game against concentrated DPS from T4s.
  
I tested basic functionality by spawning in units and checking the UI but this will need thorough gameplay testing.

## Checklist

## Summary by CodeRabbit

* **Balance Updates**
  * Rebalanced SACU/Support Commander and related economy/defense values across multiple factions, including lower durability, adjusted build costs/cadence, and retuned engineering/shield/resource upgrades
  * Retuned Quantum Gateway performance and supporting adjacency effects
  * Updated Seraphim Rambo preset loadout and right-arm shield generator behavior, including changes that affect teleport-related SACUs
* **Bug Fixes**
  * Shield Generator Field now activates immediately (no delayed shield rebuild)
  * Resource Allocation upgrades now correctly apply/revert Death Nuke damage and radius
* **Documentation**
  * Updated upgrade tooltips in multiple languages with detailed “resource generation & volatility” stats (including Death Nuke)


hotcheese4444:
I'm partial to moving Sensor system before moving Shield. Sensor system makes Overcharge so much more useful/usable it should be a pre-requisite to OC. It just seems like shield+OC is not ever a viable option currently or with the proposed changes.

lL1l1:
@hotcheese4444 

> It just seems like shield+OC is not ever a viable option ... with the proposed changes.

Shield + OC is removed with the proposed changes. The new preset is shield sensors nano which is an extremely tanky single-target DPS option for Sera. It is much more efficient than OC's energy infrastructure costs and doesn't require micromanaging the drain of auto OCs going off at bad times and causing lower damage than desired because of empty storages.

> Move sensors to right arm as OC prerequisite

Then we would have range OC nano shield SACUs which seems too powerful compared to advanced nano + gun ACU:
- 30 vs 35 range
- 1.7 vs 2.2 speed
- 33500 +220/s hp vs 29500 +295/s hp 20000 shield
- 8.4k mass 530k e vs 7.1k mass 300k e + QG 2.55k m 25.5k e
- 10 T1 BP vs 56 T3 BP + QG 160 BP
- 200 gun DPS vs 400 gun DPS
- 5s auto / 3.3s manual oc vs 7.5s oc

You can give the ACU 5.5k +30/s hp with t3 for 3.2k mass 71k e. Definitely still not enough to be anywhere near as effective as proposed preset.

I also want OC to be the AoE SACU upgrade equivalent for Sera, and since it is very powerful against poor pathfinding with its one shot kills I think the OC SACU shouldn't have so much HP as to be unstoppable with that early 11k-14k mass cost rush (this is an estimate including the energy infra mass cost for it to be constantly overcharging bricks and hitting 2 or 1, but this can be even lower because you will have 1-1.5k mass in storages for ACU usage anyway).

I also think you have to consider that the adv combatant preset by itself has huge stat changes:
9850 mass -> 5950 (-3.9k or -40%)
420k e -> 208k e (-50%)
36k bt -> 28k bt (-22%)
This could easily make it viable already. It could snowball with its T3 efficiency and usage of overbuilt energy infrastructure and just win games.

Rhaelya (approved):
Changes work as intended.

PR #7143, opened 2026-06-20
Remove beetle cloak, increase hp, and reduce speed

lL1l1:
My reasoning from the [Discord thread](https://discord.com/channels/197033481883222026/1492225419527852073):

> Problem 1: Beetle is uninteractable due to its cloak until omni comes out or ACU is available on the front (and then ACU is immediately in danger to beetle snipe...)
> 
> Remove cloak. Considering my experience with beetle performance under omni at T3 stage the speed is still enough to keep beetles good.
> 
> Problem 2: Beetle is barely interactable due to its high speed because there isn't enough time to turn around and kite.
> 
> Reduce the speed so it can be kited viably but greatly increase the HP so that retreating beetles is also possible. This allows a bit of a back-and-forth trying to kill off beetles without running into them while trying to get to the higher range T2/T3 cybran units.
> I suggest 5 -> 3.5 speed (faster than mainline land and mongoose/hoplite/mobile shields, but slower than hover and titan/loya) and 400 -> ~1.2k HP (arbitrary choice based on existing unit hp/mass) as a starting point.
> 
> Problem 3: Evaluate the mass efficiency of the unit.
> 
> This is a more nebulous topic but after the above changes reworking the unit the cost and dmg the unit deals needs to be evaluated.
> I think the HP buff will make it much better against structures when dropped, so its damage should be reduced around 33% so it 3 shots t2 mex (1100 dmg -> 900-500 dmg).
> The energy drain of the unit would have to be removed as there is no cloak and any other abilities would not make sense (you would turn off personal stealth in favor of deceivers, and for weapon firing costs - it only fires once). I would like to keep it because it scales the cost of the unit down in later stage where T3 units are better at countering it due to range and DPS density.
> A different way that beetle can scale better into T3 stage is through its stun, since the stun deals "damage" in proportion to enemy dps density, and just 1 beetle getting in deals very similar stunning as multiple beetles.
> 

Adjustments after testing:
- 3.8 speed since kiting is very powerful for even slower units. The amount of HP to balance 3.5 speed would be excessive.
- Give 10 regen to allow beetles to have secondary value in tanking DPS and getting out of range before dying. It creates an interesting decision for both the beetle user and their opponent. This synergizes in a good way with Rhino/Hoplite/Brick range.
- HP slightly higher at 1340 instead of 1200 since kiting seemed quite powerful even for slower units.

See how 10 beetles trade against 8 obsidians/ilshies in sandbox. It's a large mass deficit for beetles (1900 vs  2880) but the beetles can damage around half of the enemy units' total HP anyway, which means it is quite powerful, considering that it has additional scaling with stuns and huge AoE damage.
Harb (840 mass) can kite 4 beetles (760 mass) before taking damage (it starts taking damage at 5 beetles 950 mass) because of bad beetle pathfinding when right clicked. But that Harb has perfect pathfinding due to no other units or shields and beetle will get better value from AoE, stealth, and bricks in practical scenarios.

I noticed that formations for beetles are terrible. Beetles are in the directfire bot category which goes behind t4-t3 and t2 tanks, so a lot of the time they'll just be in the middle of the formation and not ready to engage the enemy.

## Checklist

## Summary by CodeRabbit

* **Balance Changes**
  * Fire Beetle (T2 Mobile Bomb) rebalanced: removed cloak ability and related effects, increased health from 500 to 1340, added health regeneration of 10 per second, reduced movement speed from 5 to 3.8, and removed the unit's maintenance energy consumption.


Rhaelya (approved):
Works as intended.

PR #7146, opened 2026-06-20
Reduce claw range and dps but allow targets to be shot

lL1l1:
[GC vs Percy/T3 balance Discord thread](https://discord.com/channels/197033481883222026/1505365987422244995)
> GC is often considered the best T4, and mainly its a bit too good vs T3 units, which especially hurts UEF/Cybran who rely on their T3 to deal meaningful damage to T4s to protect their own less efficient, long range T4s.

- Claw range 41 -> 30: 
  GC is too good at skirmishing with T3 units because it has higher speed than percy/brick, higher range, and the claws basically guarantee 2 units to die. The range is reduced so that GC actually has to walk into the enemy formation to use the additional DPS of the claws.
- Claw DPS 729x2 -> 240x2: 
  GC's DPS/HP stats are too good against T3 units, making it extremely difficult to beat mass-for-mass considering the infrastructure investment needed for enough T3 units to kill just 1 GC (nevermind more, or with a supporting army). The claw DPS is reduced to balance it out compared to Ythotha.
- Can target units grabbed in claw: false -> true:
  Units being tractored or being in the claw cannot be targeted, which is not immersive and can create unwanted situations if they shoot your units other than the GC and kill them. This is especially bad with the DPS nerf, so targeting of units in the claw is enabled. Targeting of units being tractored is not enabled because the engine doesn't properly aim at units attached to other units, so shots targeting tractored units just miss, however once the unit is grabbed, the GC is slow enough that most shots hit.

Spawn harbs/snipers behind a GC and see how they fire at bricks grabbed by the claws while the GC is stationary, moving, and/or turning.

Testing T4s vs Percies:

"Formation move" is just starting in the default formation and normal move ordering into range in front of the T4.
Script for perfect surround:
```lua
CreateUnitAtMouse('ual0401', 1, 0, 0, 0)
local n = 16
local d = 70
local rads = 2*math.pi/n
for i = 1, n do
    CreateUnitAtMouse('uel0305', 0, d * math.cos(rads*i), d * math.sin(rads*i), -rads*i-math.pi/2)
end
```
Before changes:
- ML: 12, 9 if u attack from behind, 1 if u attack out of range and from behind, dunno if u kite
- GC: 18
- Ythotha: 20 if without dodging or splitting, 11 if you have a perfect surround in a circle (no splash damage taken), 8 with manual dodging of all projectiles + perfect surround
- Megalith: 19 formation move on top of it, 16 if you perfectly surround in a circle, 14 if you manually dodge projectiles + perfect surround -- overall very low percy count for Megalith's cost

After changes:
It takes 16 Percies to beat a solo GC (average of the 11-20 percies of Ythotha).

## Checklist

## Summary by CodeRabbit

* **Balance Changes**
  * Galactic Colossus Tractor Claw weapons rebalanced for improved gameplay: damage reduced from 729 to 240, attack range reduced from 41 to 30
  * Enemy units attached to tractor claws can now be targeted by allied forces, enabling coordinated attacks and deepened strategic options during combat


Rhaelya (approved):
Works as intended.

PR #7167, opened 2026-07-05
Reduce sniper speed and increase range

lL1l1:
- Reduce sniper bot speed so that they can't kite GC/Ythotha/Monkeylord, as otherwise they have no lategame counter for Aeon/Sera. This also makes them weaker against T3 since its harder to escape a micro mistake.
- Add 5 range to compensate the speed nerf. 

Some considerations of the effects of these changes:
 - Reaction time for starting kiting is higher with more range, but kiting in general is worse with less speed.
 - More units/HP is needed to walk into range of an idle sniper army due to higher range.
 - Ythotha and GC had a somewhat small range gap previously, so +5 range increase to that small gap is a big difference, especially for Seraphim sniper in normal mode. Ythotha 47 range vs 55/60/65 = 8/13/18 gap -> 13 (+62.5%)/18/23 gap. GC 40 range so 15/20/25 gap -> 20/25/30 gap
 - Megalith gets outranged by 1 -> 6 by Seraphim snipers in sniper mode, but Seraphim snipers are 0.3 slower so mega catches up in only 20s, which means that the snipers can get only 2 shots off before they start getting shot at with high range high damage splash.

To be tested on develop/beta.

## Checklist

## Summary by CodeRabbit

* **Balance**
  * Reduced mobility for T3 sniper bots by lowering movement/tuning speeds (including standard and sniper-mode movement).
  * Increased engagement range to compensate, expanding weapon effectiveness for both standard and sniper firing modes across the affected variants.
* **Chores**
  * Removed leftover placeholder content from an empty changelog snippet entry.


Rhaelya (approved):
Looks good to me

PR #7168, opened 2026-07-05
fix acu teleport damage not aligning with blueprint

### LINKED ISSUE #7006: [Bug]: Teleport does 1dmg more than it should
### Describe the Bug

ACU Teleport weapon does 101 dmg, despite it should do exactly 100 (according to blueprint)

### Reproduce the bug

sandbox, put any unit, teleport on him with ACU (i didnt test with SACU, probably the result will be the same)

### Screenshots

_No response_

### Additional context

Its probably connected to 'Force" damage

Lightningbulb2:
Acu teleport now does exactly as much damage as it should (according to the blueprint)

fixes [#7006](https://github.com/FAForever/fa/issues/7006)

## Checklist
~~- [ ] Changes are documented in a changelog snippet according to the [guidelines](https://faforever.github.io/fa/development/changelog#format-of-a-snippet).~~
~~- [ ] Request 2-3 reviewers from the [list of reviewers and their areas of knowledge](https://github.com/FAForever/fa/blob/deploy/fafdevelop/CONTRIBUTING.md#reviewers).~~

not necessary, tiny change

## Summary by CodeRabbit

* **Bug Fixes**
  * Adjusted teleport-in effects during the brief hidden phase so the area damage uses a tree-force type, avoiding unintended damage behavior while remaining consistent in size and magnitude.
* **Tuning**
  * Slightly increased damage values for several related unit weapon configurations (including teleport-in dummy and death-triggered nuke/weapon placeholders) from 100 to 101 for improved consistency.

### DIFF
--- a/lua/EffectUtilities.lua
+++ b/lua/EffectUtilities.lua
@@ -1380,7 +1380,7 @@ function PlayTeleportInEffects(unit, effectsBag)
 
             local fn = function(unit)
                 CreateLightParticle(unit, -1, unitArmy, 4, 10, 'glow_03', 'ramp_yellow_01')
-                DamageArea(unit, unit:GetPosition(), 9, 0.000000001, 'Force', true)
+                DamageArea(unit, unit:GetPosition(), 9, 0.001, 'Force', true)
 
                 unit.TeleportFx_IsInvisible = true
                 unit:HideBone(0, true)
@@ -1407,7 +1407,7 @@ function PlayTeleportInEffects(unit, effectsBag)
             end
 
             CreateLightParticle(unit.TeleportCybranSphere, -1, unitArmy, 4, 10, 'glow_02', 'ramp_white_01')
-            DamageArea(unit, unit:GetPosition(), 9, 0.000000001, 'Force', true)
+            DamageArea(unit, unit:GetPosition(), 9, 0.001, 'Force', true)
 
             CreateDecal(unit:GetPosition(), decalOrient, 'Scorch_generic_002_albedo', '', 'Albedo', 7, 7, 200, 300, unitArmy)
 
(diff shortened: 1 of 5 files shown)


lL1l1 (changes requested):
I think a more broad solution can be done:
1. Force type damage should be ignored by units instead. I'm not sure why they aren't, maybe @Garanas knows because he reworked the tree knocking behavior.
2. Alternatively all scripted Force damage types should be converted to TreeForce. This would have the least effect on mods. I prefer this solution with what I know.

Context for 1:
Units already handle some tree damage types:

https://github.com/FAForever/fa/blob/1d1530b5a656a7213acc14b7330fe8847a16bdbb/lua/sim/Unit.lua#L1377

This function handles the damage types used by trees for context:

https://github.com/FAForever/fa/blob/a7ccd80040e9625d203f6013ba56e06f0f661264/lua/proptree.lua#L95-L137

Lightningbulb2:
> I think a more broad solution can be done:
> 
>     1. Force type damage should be ignored by units instead. I'm not sure why they aren't, maybe @Garanas knows because he reworked the tree knocking behavior.
> 
>     2. Alternatively all scripted Force damage types should be converted to TreeForce. This would have the least effect on mods. I prefer this solution with what I know.
> 
> 
> Context for 1: Units already handle some tree damage types:
> 
> https://github.com/FAForever/fa/blob/1d1530b5a656a7213acc14b7330fe8847a16bdbb/lua/sim/Unit.lua#L1377
> 
> This function handles the damage types used by trees for context:
> 
> https://github.com/FAForever/fa/blob/a7ccd80040e9625d203f6013ba56e06f0f661264/lua/proptree.lua#L95-L137

I added option 2: It works perfectly fine using TreeForce so I don't see why not, plus it better indicates what it's actually doing in this case. (Damaging trees, not troops) I kinda thought it was there for some other sort of compatibility reason until I checked the 12 year old PR that added it.

lL1l1 (approved):

PR #7169, opened 2026-07-05
Add custom text truncation support, and improve scoreboard legibility

Lightningbulb2:
***Sorry, had to remake pull request to move it to a feature branch instead of my develop branch.***

Original PR:
https://github.com/FAForever/fa/pull/7105

I wanted to improve scoreboard legibility.

I started by separating game speed, and game quality into their own controls to make modding and future changes easier. I added dropshadows, tooltips, and right aligned the ratings while making sure to crop player names if necessary. This led to the-

Implementation of SetTruncateText() which defaults to "..." but can have custom trailing strings like "**/**", "**-**", or even "**ITDOESNTFIT**" ***(for larger amounts of text where cropping bugs could be missed)***.

This is executed through SetClipToWidth() and only activates if SetTruncationEnabled(true) is set.

To do this cleanly, and in a reusable way, I split *SetText()* into ***SetText()*** and its implicitly called ***SetDisplayText()***.

**SetText()** - controls the source text and behaves practically the same as before

**SetDisplayText()** - sends text changes to the engine without changing the Text object's internal string. This allows for better mod and source code capabilities for fancier text. The first example being custom text truncation.

I loaded up private AI matches and replays to check for console and visual errors. I got screenshots:

**Original for reference** (plus rank cropping issue)
Original

---

Game start with lots of players
GameStartExample

---

Full scoreboard
NEWSCOREBOARD

---

Replay scoreboard
REPLAY

---

Show full player names after hovering for a moment (in-case of cropped names)
CROPPEDHOVER

---

Tooltips added for new players
gameQuality

---

gameSpeed

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added game speed and estimated game quality indicators to the scoreboard, with explanatory tooltips.
  * Improved scoreboard readability with clearer player divisions, aligned controls, hover tooltips, and better highlighting.
  * Added automatic text truncation that adapts to available space while restoring full text when it fits.
  * Added text display and truncation controls for modding and development.
* **Localization**
  * Added labels and help text for game speed, requested versus actual speed, game quality, and estimated game quality.


lL1l1:
@coderabbitai review

lua/maui/text.lua
```diff
@@ -41,6 +45,78 @@ Text = ClassUI(moho.text_methods, Control) {
...
+        if self._truncationEnabled == enabled then
+            return
+        end
+
+        self._truncationEnabled = enabled
+        if enabled then
+            -- Preserve any existing OnDirty callback before adding _applyTruncation
+            if not self._originalWidthOnDirty then
+                self._originalWidthOnDirty = self.Width.OnDirty
+            end
```
lL1l1: Technically this will fail if truncation is enabled and then OnDirty is changed later. 
You'd need to put the width lazyvar behind a proxy that intercepts modifications to OnDirty to avoid this.
Proxy meaning an empty table with a metatable that uses a function for __newindex to intercept OnDirty. Not sure how the engine interacts with that though, sometimes it rawgets things.

lua/maui/text.lua
```diff
@@ -41,6 +45,84 @@ Text = ClassUI(moho.text_methods, Control) {
...
+
+                if self._originalWidthOnDirty then
+                    self._originalWidthOnDirty(var)
+                end
+
+                self:_applyTruncation()
+            end
+
+            -- Apply truncation immediately so SetText() and SetTruncationEnabled() order doesn't matter
+            self:_applyTruncation()
```
lL1l1, line 112: This causes a circular lazyvar error because `SetTruncationEnabled` is called in `score.lua` before Left/Right/Width are set, and `applyTruncation` calls `Width()`. Ideally `SetTruncationEnabled` shouldn't interact with the layout LazyVars by calling them directly as from a user perspective it is a property of the text, kinda unrelated to the layout, but I am not experienced with UI code that much so I don't have a direct solution. Could you see if something elegant is possible?

lua/maui/text.lua
```diff
@@ -41,6 +45,84 @@ Text = ClassUI(moho.text_methods, Control) {
...
+            end
+
+            self.Width.OnDirty = function(var)
+
+                if self._originalWidthOnDirty then
+                    self._originalWidthOnDirty(var)
+                end
+
+                self:_applyTruncation()
+            end
```
lL1l1, line 109: Technically this will fail if truncation is enabled and then `OnDirty` is changed later.
You'd need to put the width lazyvar behind a proxy that intercepts modifications to OnDirty to avoid this.
Proxy meaning an empty table with a metatable that uses a function for __newindex to intercept OnDirty. Not sure how the engine interacts with that though, sometimes it rawgets things.

This is an optional change though because from what I see in the codebase, `Width.OnDirty` is rarely changed.

lua/maui/text.lua
lL1l1, line 130: @coderabbitai What happens if `SetClipToWidth` is set on a text with truncation enabled? Does it cause an infinite loop because `Width` adjusts upon text advance changing, and text advance changes as text is truncated to match a changed width?

lL1l1 (changes requested):
I added a bunch of commits that include:
- annotation fixes
- style updates for snippets/code/annotations to fit the style of the rest of the repo. Let me know if you disagree or think something was made more unclear.
- Fixing `SetText` argument type to match engine `SetText`
- Forcing `_fullText` to be a string so that `GetText` works like engine `GetText`
- Update the branch with a merge from develop

Error:
The score board doesn't load after the latest changes because truncation is applied immediately upon setting the text, which accesses Width(), but the text's layout is not yet initialized, so it causes a circular lazyvar error.

lua/maui/text.lua
Lightningbulb2: hmm I think I see what you're saying, but correct me if anything is wrong here:

*(I'll use "element" in the context of HTML, hopefully you've seen its syntax)*

The `Control` type defined in `control.lua` is basically the lowest level UI element. It contains all the dimensions, state, and properties for a "standard" element.

Elements like `Text`, `Bitmap`, and `Button` all extend `Control`

So with the current setup, it's not that Text is modifying the layout, but rather the Text ***element*** is managing its own layout while also containing the text.

The way FAF does it would be like if standard HTML was like this:
```HTML

<!--used for alignment and formatting-->
<div width="50%" text-align="center" background-color="lightblue">

    <text Cropping="true" TruncationText="..." Color="red" Width="30">
       YOURWORDSHERE
    </text> 
    
    <Bitmap file="path/to/file/" Width="30">
     
    <Button="path/to/standardButton">

</div>

```

But HTML sort of does what you're suggesting and uses the generic formatting element `<div>` to both contain and style raw text, but also layout many other elements

```HTML

<div width="50%" text-align="center" background-color="lightblue">

   <!-- class properties OR inline definitions ("color: red" for example)-->
    <div class="beautiful-text" color="red">
       YOURWORDSHERE
    </div> 
    
    <Bitmap file="path/to/file/" Width="30">
     
    <Button="path/to/standardButton">
</div>

```


```CSS


// CSS class definition for above text

.beautiful-text {
  width: 250px;             /* 1. Explicit width is required */
  white-space: nowrap;      /* 2. Prevents the text from wrapping to a new line */
  overflow: hidden;         /* 3. Hides the overflowing content */
  text-overflow: ellipsis;  /* 4. Adds the '...' at the end */
}

```

But while HTML is generally designed as a text-first system, I see why the FAF engine kept it as its own element. Things would get cluttered and confusing if `control.lua` was designed to hold and display text by default.

If text starts getting a lot of complex/incompatible features then you would want to split/inherit it into several text types. (Like scrolling animated text for example) HTML has `<H1>` to `<H6>` for header sizes.

---

So while I'm still unsure about the solution to the circular dependency problem, that's the explanation to why it's setup like this.

---

Why does `SetClipToWidth()` seem to have no problem with the lazyvar thing, but mine does?
```LUA
    ---@param clipToWidth boolean
    SetClipToWidth = function(self, clipToWidth)
        if clipToWidth then
            self.Width:Set(function() return self.Right() - self.Left() end)
        else
            self.Width:Set(function() return math.floor(self.TextAdvance()) end)
        end
        self:SetNewClipToWidth(clipToWidth)
    end,
```

Thanks for reviewing and helping me with this!

(I had a fun time writing this! I didn't anticipate my basic HTML experience would help me understand FAF UI xD)

*Written without any AI*

Lightningbulb2: maybe not perfect, but this fixes the lazyvar width() initialization

```LUA
--- Internal function to fit the truncation string inside the max width
_applyTruncation = function(self)

  -- Avoid unnecessary truncation, and no width issues
  if self._fullText == "" then return end

  local initialized, maxWidth = pcall(function() return self.Width() end)
  if not initialized or maxWidth <= 0 then
      self:SetDisplayText(self._fullText)
      return
end
```


and that also makes me realize that we could have optional truncation as a shorthand in `SetClipToWidth()`
```LUA
    --- Adjusts text cropping behavior
    ---@param clipToWidth boolean
    ---@param truncationEnabled? boolean
    ---@param customTruncationSymbol? string
    SetClipToWidth = function(self, clipToWidth, truncationEnabled, customTruncationSymbol)
        if clipToWidth then
            self.Width:Set(function() return self.Right() - self.Left() end)
        else
            self:SetTruncationEnabled(false)
            self.Width:Set(function() return math.floor(self.TextAdvance()) end)
        end

        if truncationEnabled then
            self:SetTruncationEnabled(truncationEnabled)
        end

        if customTruncationSymbol ~= nil then
            self:SetTruncationText(customTruncationSymbol)
        end

        self:SetNewClipToWidth(clipToWidth)
    end,

```
But then I wonder if it should all only be in `SetClipToWidth()` as that is required for truncation anyway.

lua/maui/text.lua
Lightningbulb2: Yeah at least this one won't overwrite another `OnDirty`, but it will be up to someone else to preserve the `truncation OnDirty` when setting their own. Although that could get really frustrating to troubleshoot, but it's sorta a limitation of only having a single `OnDirty` for any given property.

Maybe some sort of helper version that won't overwrite an existing one like `PreservedOnDirty` or `StackOnDirty`

Or a more complex system that allows you to add/remove to a list of named `OnDirty` functions. (I'd rather have a global fix rather than a niche hacky one, but that is probably out of the scope of this)

PR #7175, opened 2026-07-10
Rebalance mercy with higher damage and lower AoE

lL1l1:
## Reasons to rebalance the mercy again:
1. While the current AoE-focused stats and low cost stats have a great risk/reward ratio when targeting engineers (air facs, naval facs) or T1 pgens (T2 if you get 3 through somehow), they aren't very versatile so the timing to use Mercies is very small.
2. People really want the old identity of the mercy in being able to deal tons of damage to ACUs or just as suicide bombers against an incoming experimental (like t1 bombers).
3. The AoE-focused stats were intended to also play a role in dealing large damage against armies, but they failed at that.

[Discord thread](https://discord.com/channels/197033481883222026/1512538645750485264) developing the changes.

The mercy is given 2100 damage over 4 seconds in 3.5 radius instead of 1000 over 9.5 seconds in 7.5 radius.
The cost and HP buffs meant for the old balance are reverted.

How the new stats resolve the listed reasons:
1. The higher damage is more versatile because less mercies are needed to kill valuable targets: 1 mercy can snipe any T2 mex + storages, 2 mercies can snipe T2 pgens (additionally, the 3.5 AoE lets it hit 2 diagonal pgens unlike other fighter/bombers), 4 mercies can kill any t2 land/air HQ or T3 mex.
2. ACU sniping is now viable but is balanced for skill levels. At low skill an ACU might be idle and walk out of the DoT late so they'll suffer a large amount of damage compared to the mercy cost. At high skill an ACU is probably going to be moving around so they'll suffer damage somewhat equal to that of a dodged corsair snipe (2970 mass = 11 mercies = 3 passes of 7 corsairs to kill a base level ACU).
3. The new mercy is probably still going to be bad against armies, although it will feel more impactful since it can finish off units with its high damage, especially if they stand still or get stuck in pathfinding. The versatility in sniping targets will definitely make this less of an issue though.

Spawn 4 mercies and right click an ACU. 
If the ACU is moving, they deal 1050 damage (expected balance based on mass-equivalent of 3 passes from dodged corsairs).
If I move the ACU on reaction to the mercies, then it takes 1225 damage on average (3*350 from 2 mercies, 4*350 from other 2 mercies).

{BBF3B04F-E238-4AD6-B692-179BE0F5F321}

## Additional context
Credit to https://github.com/gapforever2/fa/commit/80b5d62eab4050abba01933ac36012954070c697 for the dmg/radius stats: 

## Checklist

## Summary by CodeRabbit

* **Balance Changes**
  * Rebalanced Mercy with significantly higher Kamikaze damage and a smaller blast radius.
  * Mercy now has lower health, higher construction costs, and a longer build time.
  * Kamikaze damage-over-time effects resolve faster with fewer pulses.
* **Gameplay Improvements**
  * Guided missile spread and impact effects now scale more accurately with their configured damage radius and damage-over-time duration.


Rhaelya (approved):
Looks good to me

PR #7177, opened 2026-07-16
Fix init crash due to string mod version and prettify set mod list logging

lL1l1:
## Issue
The mod "AutoTML" has `version="1.1.1"`, which causes rule and session init files to error because the string format function was using `d` (digit) to format the version, and "1.1.1" cannot be converted to that.

Change the format identifier from `d` to `s` (string).

I noticed that the mods list printing was very long and hard to understand due to debug prints. I copied the style used in the init files over and removed the debug-style `table.print` statements.

run `LOG(string.format("%s", "1.1.1"))` in console and it does not error and produces the correct output.

The mod list output is lined up correctly:
```
info: Mod list set to:
info:         [UI]  "Common Mod Tools"              v01 (zcbf6277-24e3-437a-b968-Common-v1)    by Crotalus
info:         [UI]  "ReUI"                          v06 (reui-1.3.1)                           by 4z0t
info:         [SIM] "2% Particles"                  v01 (ab88d092-fae3-46e2-8224-53a47db7aa59) by CatmanFS
info:         [UI]  "Penguin's Icon Mod"            v02 (3ccc4eca-0cb1-11ec-ac65-a37599393411-02) by EmperorPenguin
info:         [UI]  "Selection Cost UI"             v09 (2018eaac-e8ed-11e4-b02c-1681e6b44c9)  by Eternal-
```

## Checklist

### DIFF
--- a/lua/RuleInit.lua
+++ b/lua/RuleInit.lua
@@ -21,7 +21,7 @@ doscript '/lua/system/debug.lua'
 
 LOG('Active game mods for blueprint loading:')
 for _, mod in __active_mods do
-    LOG(string.format('\t"%-30s v%02d (%-37s by %s', tostring(mod.name) .. '"', tostring(mod.version), tostring(mod.uid) .. ')', tostring(mod.author)))
+    LOG(string.format('\t"%-30s v%02s (%-37s by %s', tostring(mod.name) .. '"', tostring(mod.version), tostring(mod.uid) .. ')', tostring(mod.author)))
 end
 
 doscript '/lua/footprints.lua'
--- a/lua/RuleInit.lua
+++ b/lua/RuleInit.lua
@@ -21,7 +21,12 @@ doscript '/lua/system/debug.lua'
 
 LOG('Active game mods for blueprint loading:')
 for _, mod in __active_mods do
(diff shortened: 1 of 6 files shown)


lua/RuleInit.lua
```diff
@@ -21,7 +21,12 @@ doscript '/lua/system/debug.lua'
 
 LOG('Active game mods for blueprint loading:')
 for _, mod in __active_mods do
-    LOG(string.format('\t"%-30s v%02d (%-37s by %s', tostring(mod.name) .. '"', tostring(mod.version), tostring(mod.uid) .. ')', tostring(mod.author)))
+    LOG(string.format('\t"%-30s v%02s (%-37s by %s'
+        , tostring(mod.name) .. '"'
```
4z0t, line 25: include these trailing characters into format string.

4z0t (changes requested):

lua/RuleInit.lua
lL1l1: That would cause padding to be added inside the quotes instead of after them.

4z0t (approved):

PR #7186, opened 2026-07-22
Fix selection with `Hidden` wrapper not being ignored in commandmode

lL1l1:
Fixes selection done within the `Hidden` wrapper not being ignored in `commandmode.lua`.
`gamemain.lua`'s `SetIgnoreSelection` is used since it also sets the ignore mode in `commandmode.lua` and it lets us avoid an import call every time selection changes, which is especially frequent for hidden selections.

Un-revert the debug logging commit.
Select a unit.
Select the move order, shift click once, then run this command while holding shift:
```
import('/lua/ui/game/selection.lua').Hidden(function()
    UISelectionByCategory('ALLUNITS', false, false, false, false)
    LOG('Selected units', table.getn(GetSelectedUnits()))
end)
```
letting go of shift should cancel command mode and not log anything. Before these changes it didn't.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed selection handling while the selection interface is temporarily hidden.
  * Prevented hidden selections from affecting command mode, unit views, rally points, sounds, and related updates.
  * Ensured previous selections are restored correctly after hidden selection operations.

### DIFF
--- a/changelog/snippets/category.7186.md
+++ b/changelog/snippets/category.7186.md
@@ -1,0 +1,1 @@
+- Your explanation here... [Don't forget to change the category in the filename] (#7186).
--- a/changelog/snippets/category.7186.md
+++ b/changelog/snippets/category.7186.md
@@ -1,1 +1,0 @@
-- Your explanation here... [Don't forget to change the category in the filename] (#7186).
--- a/changelog/snippets/fix.7186.md
+++ b/changelog/snippets/fix.7186.md
@@ -1,0 +1,1 @@
+- Fix selections done within `selection.lua`'s `Hidden` wrapper not being ignored in `EndCommandMode` (#7186).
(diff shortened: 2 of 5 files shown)


4z0t:
@Garanas decided to not make [this](https://github.com/FAForever/fa/blob/da311cdc28f9214a349fd1b3b314df82b76277f5/lua/userInit.lua#L378-L391) code for everyone, so you'll have to @lL1l1 ...

4z0t (changes requested):
Use solution from `IssueBlueprintCommandToUnits`.

4z0t:
No, just fix the one you're touching in this PR

lL1l1:
I refactored it as requested and also noticed that the hidden state would be broken by errors so I wrapped it in pcall. If pcall fails it fixes the hidden state and then errors so the script that called it doesn't continue assuming something in the callback worked.

PR #7191, opened 2026-07-24
Merge with all weapon blueprints that have the given label when using MergeLabel

lL1l1:
When implementing weapon merging, I assumed weapon labels should be unique, but according to @The-Balthazar they're sometimes used for using the same weapon script class.

none

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed weapon blueprint merging so `MergeLabel` now applies to all matching weapon blueprints, rather than only one.
  * Improved consistency when multiple sub-weapons share the same label, ensuring each matching blueprint is processed during loading.

* **Documentation**
  * Clarified blueprint weapon-merging behavior in the relevant documentation and changelog.
  * Documented that labels may match multiple weapon blueprints and that all matching entries are merged.


lua/system/Blueprints.lua
```diff
@@ -1002,8 +1003,12 @@ function MergeWeaponByLabel(baseBp, label, insertPos, newBp)
     end
 
     for i, w in weaponTable do
+        local merged = false
         if w.Label == label then
             weaponTable[i] = BlueprintMerged(w, newBp)
+            merged = true
+        end
+        if merged then
```
speed2CZ: The code will return after a single merge, just like the old version, just with one extra boolean set

speed2CZ (changes requested):

lua/system/Blueprints.lua
lL1l1: That was a careless mistake; it should've been outside the loop so that it doesn't insert at the end of the array if it merged some blueprints.

PR #7193, opened 2026-07-25
Give non-Sera cruisers 0.3 splash to counter aircraft shift-g

lL1l1:
Cruisers need minimal splash so that ASF or gunship shift-g can be countered by defensive AA usage throughout naval T2 stage.
At the same time they need to avoid countering torp bombers too hard.

I added some debug logging and tested to find that 0.3 splash is a good compromise.

Use debug logging commit to see how many units are hit per tick of damage instances. 
Use a UEF cruiser (since it fires 1 projectile per tick) and test against various numbers of hold-fire torp bombers and ASF.

I tested 0.1, 0.3, and 0.5 splash. I liked 0.3 with the following results:
- I got max 2 hits at once for 20 torps that are put into formation about 160 distance away and then right clicked onto the cruiser. The 2 hits happened for the small moment where the torps were flying directly over the cruiser in their attack run.
- I got 1-7 hits at once for 100 ASF that were poorly shift-g micro'd. The occassional tick would hit 25 times, meaning that if the enemy manually attacks the core of the ASF shift-g instead of letting auto-attack hit stray ASF on the outside, then they can deal massive damage. For very strongly shift-g micro'd 100 ASF (using shift-g + dragging around the move order near the ASF so they clump), I was able to see up to 50 of them get hit in 1 tick.
I think these ASF results show that the change will discourage shift-g micro over enemy cruisers, as the snowball risk is too great, and that will re-add the depth of shift-g vs non-shift-g in defense vs offense respectively for the t2 naval stage/area.

## Checklist

## Summary by CodeRabbit

- **Balance Changes**
  - Non-Seraphim cruisers now have a small area-effect radius on their anti-air weapons.
  - This improves their ability to damage tightly grouped air units, especially ASF and gunships, while keeping the impact on torpedo bombers limited.
  - The change applies to the UEF, Aeon, and Cybran cruiser anti-air weapons.


vjt4:
Can T2 Torps go back to 240M if cruisers all get splash back

lL1l1:
It's less than splash than before and I tested it to not be as op as 1 splash so no.

PR #7201, opened 2026-08-04
Introduce map selection localization support.

Yuchenjimmy:
This PR introduces the ability to translate mission names, descriptions. Now the game will detect if a map has LOC entries for name and description, and it will have different outputs based on three conditions. Following are examples showing how it handles a mission name:

1. A map already has proper LOC support == return it normaly.
2. A Skirmish map without LOC support == return LOC('<LOC '..name..'_Name>'..str)
3. A FAF Coop map without LOC support == return LOC('<LOC FAF_Coop_'..name..'_Name>'..str)

1. Skirmish maps show translation correctly.

2. Campaign maps show translation correctly.
Forged Alliance 2026_8_4 下午 11_07_17

3. FAF Coop maps show translation correctly.
Forged Alliance 2026_8_4 下午 11_08_26

4. Maps without translation show original text as expected.
Forged Alliance 2026_8_4 下午 11_08_47

## Additional context
Here are two corresponding pull requests:

1. Introduce the ability to translate mission objectives and descriptions: https://github.com/FAForever/faf-coop-maps/pull/450
2. Corresponding localization strings: https://github.com/FAForever/fa-coop/pull/123

This pull request should have be created an years ago, lazy me.

## Checklist
- [X] Changes are documented in a changelog snippet according to the [guidelines](https://faforever.github.io/fa/development/changelog#format-of-a-snippet).


lua/ui/dialogs/mapselect.lua
```diff
@@ -962,11 +962,41 @@ function SetupOptionsPanel(parent, curOptions)
...
 end
 
+-- Generate LOC entries for scenario names
+function LOCN(str, type)
+    local name = str
+    name = name:gsub(' %- ', '_'):gsub('%-', '_'):gsub(':', ''):gsub("'", ''):gsub('%w+ Mission %d+ ', ''):gsub(' ', '_')
+    if string.find(str, '<LOC') then
+        return LOC(str)
+    elseif type == 'campaign_coop' then
+        return LOC('<LOC FAF_Coop_'..name..'_Name>'..str)
```
speed2CZ, line 972: its not a good idea to hardcode `FAF_Coop_`, it's a convention for custom missions, not a requirement.

Yuchenjimmy: Hi. `FAF_Coop_` is not required in the mission name, folder, `OPERATION_NAME`, or `OPERATION_DESCRIPTION`. The function only checks whether the input already contains `<LOC`. If it does, it uses that value unchanged. If it does not, the function generates a LOC entry, and `FAF_Coop_` is only part of that generated output.
For example, mission “Fort Clarke Assault” has original name `OPERATION_NAME = 'Fort Clarke Assault'`, this function produces `<LOC FAF_Coop_Fort_Clarke_Assault_Name>Fort Clarke Assault`.
Therefore, `FAF_Coop_` is not used to identify or validate a custom mission. It is only added when constructing the output localization key.

speed2CZ:
In what cases is this actually gonna help? Generating the LOC tag won't do anything if there are no translations anyway?

Yuchenjimmy:
> In what cases is this actually gonna help? Generating the LOC tag won't do anything if there are no translations anyway?

Indeed. Just as you said, it will not have any effect until someone decides to help translate it into their own language. This simply provides an opportunity for those who may want to contribute translations in the future. With that in mind, I also created another PR that adds Traditional Chinese translations: [FAForever/fa-coop#123](https://github.com/FAForever/fa-coop/pull/123). It is not finished yet, so I have left it as a draft for now.

There is another PR that directly adds LOC keys for the SC and SCFA campaign maps, which is also linked in the Additional context section above.

speed2CZ:
In the code you've changed, there was only one place where the string was not localized, other 2 already called LOC on it, so what is the tag generation for exactly?

On its own it has no purpose and you're just guessing what the tags in the maps will be. But if the maps include tags, then you dont need to generate them.

Yuchenjimmy:
The existing LOC function calls do not localize plain strings by themselves. For example, `LOC('Fort Clarke Assault')` just returns the same English text because there is no localization key to look up.

Most of co-op maps, including Fort Clarke Assault, provide plain names and descriptions. **Only a small number of them already include LOC keys**, for example:

`OPERATION_NAME = '<LOC FAF_Coop_Havens_Invasion_Name>Haven\'s Invasion'`

The generated tags follow this existing format. If a map already provides a LOC tag, the function preserves it; otherwise, it generates one for the plain metadata.

speed2CZ:
If there are any maps that dont have the loc tags in names and actually have localazation strings avaialble, the tags should be added to the maps. It should not be generated.

If the map doesnt have  any localization, generating tags won't help anything.

Yuchenjimmy:
> If there are any maps that dont have the loc tags in names and actually have localazation strings avaialble, the tags should be added to the maps. It should not be generated.

99% maps don't have loc tags, that's a lot of work. On the other hand, currently it's not required to add loc tags when making a map, which means it needs someone manualy to add them everytime there's a new map. With this PR, translators can just edit strings.lua for every map, without touching the map repo.

> If the map doesnt have any localization, generating tags won't help anything.

Sorry, I don’t understand the connection here. Tags and localization serve different purposes, so a map not being localized doesn’t necessarily mean tags are useless. I will update Traditional Chinese strings, and it just needs somebody else to update other languages.

PR #7205, opened 2026-08-07
Revert SACU rebalance to prep for release

lL1l1:
SACU partial rebal iteration 2 cannot be released because it is untested.
There was no testing because I didn't do any promo work because Rhaelya was completely unavailable for the past few weeks and I wanted him to do a discord announcement before I do anything.

## Summary by CodeRabbit

* **Balance Changes**
  * Rebalanced SACU commanders across all factions, including costs, build times, production, combat abilities, shields, stealth, and upgrade effects.
  * Resource Allocation upgrades no longer affect death-weapon damage.
  * Updated Quantum Gateways with increased durability and revised construction and economy values.
  * Adjusted adjacency bonuses for power generators and mass fabricators.
* **Bug Fixes**
  * Improved shield enhancement behavior and command-unit targeting.
* **Localization**
  * Updated unit and upgrade descriptions across supported languages to reflect current effects and presets.


hotcheese4444:
So the first round was tested and problems found, we have a patch that addresses (or at least tries to address) the issues. And now we're going back to square 1? Will SACUs ever be touched

lL1l1:
It's a huge set of changes that shouldn't be tested for only 1 iteration before release. Yes it sucks to delay it but we just don't have the test framework to bring it to an acceptable level in 1 iteration.
My specific concern is that the 2nd iteration mostly touches high tier rambos while the low tier combatants are potentially still quite strong, and there have been no games trying that strategy extensively as far I know.

We can continue working on the SACUs. This revert PR can be merged or applied to staging/faf to not touch the develop branch.
Test frameworks could be improved to not require players. Currently setting up test scenarios is 100% manual work with bad UI tools.

BlackYps (approved):

PR #7213, opened 2026-08-09
Fix plugin uri search and add missing diagnostic

lL1l1:
hook file paths with non-word characters were not being found, for example economy_mini.lua as used in ecomanager.
There was also a unicode character warning for some hooked file so I disabled the diagnostic.

The hook works when I applied the fix in my mod development process.

## Checklist


speed2CZ (approved):
Do we still need this? The new addon wont use this at all.

Issue #7184, opened 2026-07-21
[Bug]: Aircraft get stuck in (human player's) air staging for unknown reason


lL1l1:
### Describe the Bug

In some games people have noticed that aircraft will dock at air staging and not automatically undock once repaired and refueled like they should. Instead, they have to be manually unloaded using the unload command by clicking on the aircraft in the transport storage display, and the yellow "deploy" command does not work.
This has been a long-standing bug without successes replicating it:
- Planes are not taking off from the airfield ([Discord](https://discord.com/channels/197033481883222026/1528864505625444493))
- Aircraft doesn't auto-release from air stages in the lategame ([Discord](https://discord.com/channels/197033481883222026/1322240211429883935)); replays [23908896](https://replay.faforever.com/23908896) 37:40 sainse, and [26100581](https://replay.faforever.com/26100581) 30:00

https://github.com/user-attachments/assets/1bc53132-f9f6-4972-bbfd-79b1ce9d992a

- Planes getting permanently stuck in air staging beacon ([Discord](https://discord.com/channels/197033481883222026/1205933573798895646)); replay 21820767 19:20 berrysdisciple

### Reproduce the bug

As mentioned, people haven't found a way to replicate the bug yet. 
Some people say it is related to stalling resources.
ETFreeman said that it happened to him when he was using a normal dock command.

In my own testing mass stalls don't do anything because repairing doesn't cost mass, and energy stalls slow down repairing but don't affect unit behavior. Stalling resources does not affect refuel rate.
I was able to replicate a similar situation to the bug by issuing an order that makes aircraft dock (dock, patrol, attack move, guard orders) and then using ctrl+shift+right click to delete the order while aircraft were docked at the air staging, which caused the aircraft to not get out of the air staging automatically after they're done repairing.
My exact test setup was simply spawning 4 T3 mass extractors to stall energy and then using damaged ASF in various scenarios.
To damage the ASF, I used this script:
```lua
ConExecute([[SimLua 
local s = DebugGetSelection()
for _, u in s do
    u:SetHealth(u, u:GetMaxHealth() * 0.01)
end]])
```
The scenarios I've tested were:
1. manually deleting the orders that cause docking while aircraft were docked
[...]

PR #7219, opened 2026-08-12
Fix aircraft stuck on air staging facility after passing through Restoration Field

K-ETFreeman:
## Issue
Partially addresses #7184.

Aircraft get stuck on air staging when their stored HP exceeds current MaxHealth — the engine's auto-undock in `CUnitRefuel::TaskTick` uses strict equality `health == maxHealth`, unreachable when `health > maxHealth`.

The corrupt state persists indefinitely after any MaxHealth buff with `DoNotFill=true` (e.g., Seraphim ACU Restoration Field) expires on a plane that was healed to the buffed max. Replay `26100581` confirms: all 16 stuck planes show `hp=319/maxHp=290`.

Override `OnTransportAttach` on `AirStagingPlatformUnit` to clamp docked plane's HP to current MaxHealth. Scoped to air staging — regular transports (`AirTransportUnit`, `AircraftCarrierUnit`) have separate class hierarchies and are unaffected.

This is a partial fix. The second failure mode in #7184 (leaked attachment from task death — e.g., canceling a patrol order while docked) needs an engine binary patch and is tracked separately.

## Why a Lua workaround instead of an engine fix
The cleanest fix would be a binary patch on `CUnitRefuel::TaskTick @ .text:00621490` — relax the strict equality `health == maxHealth` to `health >= maxHealth`. That hardens the consumer side at the source.

However, binary patches take months to ship (separate repo, separate review process, requires executable redistribution). This Lua fix is proposed as a stopgap so players stop hitting the bug in the meantime. Once the engine `>=` patch lands, this override should be reverted.

Sandbox test with Seraphim ACU + Restoration Field + XSA0102 interceptors:
- Damage interceptors, fly through aura until healed to buffed max (319 HP)
- Leave aura, wait 5+ seconds for buff to expire (stored HP=319, maxHp reverts to 290)
- Dock at air staging
- Pre-fix: planes stuck forever
- Post-fix: planes auto-undock after refuel completes

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Fixed an issue that could leave aircraft stuck on an air staging facility after passing through a Restoration Field.
  - Attached aircraft health is now corrected to its maximum when docking, preventing transport-related failures.


lL1l1 (commented):
I investigated this further and this seems like a more core bug to the MaxHealth Buff used by regen aura. When the buff is removed, it sets the max health below the current health, but the engine keeps that current health value until the next call of AdjustHealth. This causes numerous bugs:
- The health greater than max is used to calculate the health adjustment, and the max health ceiling kicks in after. For example, a unit that gained and then lost 500 max HP will only take 100 damage if hit by a 600 damage projectile.
- Air units get stuck in air staging due to the engine checking health == maxhealth to undock units.
- If the unit does not take damage and regains the max health buff, its HP is restored to the large value even though the buff blueprint specifies that it should not fill the target's health upon application.

For these reasons I'd rather fix the health buff and I don't see a reason for this air staging code to exist.

PR #7223, opened 2026-08-14
Implement feedback for chat rework

### LINKED ISSUE #7238: Can't see my own messages unless chat window is open
Image
Image

When i type in chat the chat window closes and my i can't see if my message was sent or not - only when i open the chat window again i can see my message. shouldn't my messages pop up in chat area like the messages from other players? ir is it intended behavior

lL1l1:
Implementation of the following feedback:
- ✅it hides immediately after closing with enter but in the past I think it showed  for the duration of the "unread message". (Make chat stay around after closing it until it fades away)
  - ✅In general fade time should apply after any time chat appears, for example so you can see what you typed after pressing enter and so that you can wait for a reply. (Make chat stay around after closing it until it fades away)
- ✅Combatant SACUs show up as a shared engineer. You could filter by blueprint selection priority to determine the sacu's role.
- ℹ️Judging by the fact that SSB sends a share message at the same time as the new chat, but it doesn't appear in replays, maybe that's a limitation of the engine that causes missed chats in replays in general.
- ✅ Add option to disable receive units messages
- ✅ Add option to disable receive resources messages
- ✅ The camera link attached to the unit share uses a slow pan and zoom while the manual camera link is instant. That could be a setting. ( ✅The unit share cam link was made instant, ❌there is no setting because the two cam links use very different functions)
-✅ Clicking a cam link without opening the chat dialog with enter doesn't work.
- ✅ Private messaging someone and using "You" to send resources/units could be improved from the perspective of observers. - it is changed to using "%s sent %s X".
  -✅ You could also consider making it a team chat message so the team knows where resources are going.
- ✅Chat feed background extends to the right of the scroll bar instead of to the left
- ✅ Chat window should not close due to idling until after MouseExit from the main chat window + Edit is not focused.
- ❌ Escape key deletes the currently drafted non-empty message instead of immediately losing Edit focus. Rejected: Kinda fine standard UI behavior. Players can use keyboard but arguably its better to have esc instant exit so that they can issue hotkeys asap and use a hotkey while moving their hand to their mouse.
- ℹ️ Discrepency in chat saving to replays: Only people sending messages save to replay. Previously everyone saved messages they received.
- Notify shows up as Local when observing. You could copy some work from [Fix notify messages for observers #6849](<https://github.com/FAForever/fa/pull/6849>). I implemented the PR's changes:
  - ✅Observers no longer save chats they receive to replay, so they don't save the notify data to replay.
  - ✅ Change message sender name from "local" to whoever sent them for obs.
- ✅Live Observers cannot see ally messages, they should be able to.
- ✅Live Observers cannot see private messages, they should be able to.

For chat hiding behavior:
Set Fade time to 5s in options. Bind "Append a short synthetic chat message".
1. Press enter to open chat. Type a message. Press enter to send the message then press enter again to exit the edit box. The chat message should stay on screen instead of instantly disappearing.
2. Send a synthetic message. Then send a normal message. There should be no problems with overlapping chat message lines. Then close the chat window. No messages should disappear (chat fading feed history should be the same as the open chat window's history).

For the rest, testing with 3-player multiplayer lan launch script. Use -tolobby arg to go to lobby and put a player in obs to test the live obs changes.
Didn't test replay features.

## Additional notes
The feed interface is badly design because the chat feed is keeping its own independent chat history and managing their lifetimes and such. I'd rather read the history from the most recent message and render messages based on the system time. More generally, it's completely independent of the chat lines interface even though the two do the same job of displaying relevant lines, just split by chat window's open/close state.

- There's a bug where changing the fade time keeps the current chat feed on screen forever until a new chat message is sent.
- Feed interface doesn't update in reaction to mute option changes due to its separate history (would require refactor or tight coupling: equal complexity projects).

## Checklist

## Bug Fixes
- Improved chat history retention and message fade behavior when the chat window or feed is hidden.
- Added clickable camera links and army names, with faster camera movement and improved link availability.
- Prevented automatic chat closure while typing or hovering over the window.
- Improved observer notifications, private-message visibility, and sender attribution.
- Enhanced resource and unit transfer notifications with clearer sender and recipient details.
- Enhancement notifications now identify the relevant player.

## New Features
- Added options to mute shared-resource and shared-unit messages.
- Added Ctrl-click copying for chat message text.


BlackYps:
what is a synthetic message?

loc/RU/strings_db.lua
```diff
@@ -4339,16 +4339,16 @@ chat_0012="Фильтры сообщений"
...
 chat_resources_overflow_mass="Хранилище заполнено — потеряно %d массы. Постройте больше хранилищ."
 chat_resources_overflow_energy="Хранилище заполнено — потеряно %d энергии. Постройте больше хранилищ."
-chat_units_received_one="%s передал вам юнита."
-chat_units_received_many="%s передал вам %d юнитов."
-chat_engineers_received_one="%s передал вам инженера."
-chat_engineers_received_many="%s передал вам %d инженеров."
+chat_units_received_one="%s передал %s юнита."
+chat_units_received_many="%s передал %s %d юнитов."
+chat_engineers_received_one="%s передал %s инженера."
+chat_engineers_received_many="%s передал %s %d инженеров."
```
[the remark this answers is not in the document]
lL1l1: sounds wrong for the nouns used

lL1l1:
> what is a synthetic message?

It's the term used for the chat debug hotkeys that are implemented in [ChatDebug.lua](https://github.com/FAForever/fa/blob/fix/chat-close-behavior/lua/ui/game/chat/ChatDebug.lua), which append an entry directly using the chat controller to test UI behavior like layout and onframe logic.

Garanas:
> ℹ️ Discrepency in chat saving to replays: Only people sending messages save to replay. Previously everyone saved messages they received.

It's sufficient for only one player to sent it (via a sim callback), the callback is recorded in the replay for all users. The sim callback only exists to record the messages. The default should be to just sent the messages directly, entirely in the UI.

The old approach is still done regardless, because otherwise existing parsers would no longer parse all the chat messages.

Note that messages that only exist in the UI (like those of SSB) do not exist in the replay. Since they are not synced via a callback. 

edit: and thank you for picking up the work. I hope the code is as readable to you as it was to me.

lua/ui/game/chat/ChatInterface.lua
```diff
@@ -143,6 +144,19 @@ local ChatInterface = ClassUI(Window) {
             )
         )
         ---@diagnostic enable: undefined-field
+
+        local oldHandleEvent = Window.HandleEvent
+        ---@param self UIChatInterface
+        ---@param event KeyEvent
+        self.HandleEvent = function(self, event)
```
Garanas: I think you can also redefine the HandleEvent as a function of the class and then call the base class accordingly. Just like hierarchies in for example the unit classes.

lL1l1: good spot; not sure where I looked to get the idea to do it this nonstandard way.

lL1l1:
> It's sufficient for only one player to sent it (via a sim callback), the callback is recorded in the replay for all users. The sim callback only exists to record the messages. The default should be to just sent the messages directly, entirely in the UI.

It is a small security hole in that someone can modify their client to not record messages they send, and then mods would have no way to see anything and they'd have to track down the behavior across reports and start asking for screenshots as evidence.

lL1l1:
Any remaining changes would deal deeply with how observers and players can message when cheats are enabled, which is complicated, so I don't want to do them in this PR which I want finished for a hotfix.

lL1l1:
So I think this PR is finished

BlackYps (approved):

PR #7226, opened 2026-08-15
Add restart session button to escape handler in singleplayer

lL1l1:
Helps developers escape UI lockups where hotkeys don't work.

Reload some UI module that relies on state from other modules. 
Alt-f4 should successfully bring up the escape handler if you have the option that prevents instantly exiting the game turned on. Then you can click the button to restart, letting you continue working on the UI module.

## Checklist


speed2CZ:
Btw. there's a hotkey, by default Ctrl F10 I think for restarting the current session

lL1l1:
Yes but as I mentioned in the description if you break the UI you can break hotkeys. The escape handler is called from the engine so it is  more reliable in those situations.

speed2CZ:
Fair enough, maybe I just dont break the UI enough

speed2CZ (approved):

PR #7231, opened 2026-08-16
Improve callback handling

lL1l1:
Improve callback handling across the codebase by:
- wrapping callbacks in pcalls
- removing callbacks if they fail
- writing appropriate warning messages upon failure
- verifying inputs for added callbacks
- adding annotations

## Checklist


lL1l1:
One consideration to make is that if a callback errors conditionally then it may not be necessary to remove. For example a UI OnBeat callback for an economy overview mod that errors on a specific unit type or unit event, but its UI continues working otherwise.

PR #7235, opened 2026-08-18
Fix lobby crash and launch re-entrancy when kicking observers

juerghegglin:
This PR fixes a UI lifetime crash and launch re-entrancy issue when the host launches a game with observers present while "Allow Observers" is disabled:

1. **`lua/ui/controls/popups/popup.lua`**:
   - Guarded `Popup:Close()` with `not IsDestroyed(self)` to prevent calling native methods on already destroyed UI objects.
   - Guarded `Popup:OnClosed()` with an `if not self._closed then ... end` flag so `EscapeHandler.PopEscapeHandler()` is called at most once per popup.
   - Added `Popup:OnDestroy()` to ensure escape handlers are popped cleanly if a parent group (such as `GUI`) destroys the popup directly.

2. **`lua/ui/uiutil.lua`**:
   - In `QuickDialog.MakeButton()`, set `self._clicked = true` and invoked `self:Disable()` on first press to prevent rapid double-clicks and repeated Enter presses from firing the callback multiple times.
   - Added `if destroyOnCallback and not IsDestroyed(popup) then popup:Close() end` to ensure callbacks that destroy the parent container do not cause `popup:Close()` to execute against an already destroyed popup.

3. **`lua/ui/lobby/lobby.lua`**:
   - Added a `launchInProgress` flag in `lobby.lua`.
   - In `TryLaunch()`, guarded against re-entry and disabled `GUI.launchGameButton` once validation passes and the launch sequence is committed.
   - Reset `launchInProgress = false` (and re-enabled the button) if peer connection verification fails or when `CreateUI()` runs.

- Created an automated unit and regression test suite in `tests/utility/quickdialog_lifecycle_test.lua` (22 assertions passing).
- Verified:
  1. QuickDialog open/close lifecycle and EscapeHandler balance.
  2. Callbacks that destroy the parent container do not throw errors or cause native double-destroy crashes.
  3. Rapid double-clicks and Enter key presses do not invoke duplicate actions.
  4. Lobby `TryLaunch` kick and launch calls execute exactly once.

## Checklist
- [x] Changes are documented in a changelog snippet
- [x] Request 2-3 reviewers from the list of reviewers and their areas of knowledge

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented crashes when closing or destroying popups multiple times.
  * Ensured popup cleanup runs only once, including when closed with Escape.
  * Prevented duplicate actions from rapid repeated dialog-button clicks.
  * Prevented duplicate launch attempts and restored launch controls after failed launches.
  * Improved stability when dialogs or popups are destroyed during callbacks or with observers disabled.

* **Tests**
  * Added regression coverage for popup lifecycles, dialog interactions, repeated clicks, Escape handling, and launch recovery.

### DIFF
--- a/lua/ui/controls/popups/popup.lua
+++ b/lua/ui/controls/popups/popup.lua
@@ -8,6 +8,7 @@ local EscapeHandler = import("/lua/ui/dialogs/eschandler.lua")
 -- and draws a standard background behind its content. You'll probably want to extend it to do
 -- something more involved, or use it as-is if you want to manually assemble your popup UI Group.
 ---@class Popup : Group
+---@field _closed boolean
 Popup = ClassUI(Group) {
 
     --- Create a new popup
@@ -17,6 +18,7 @@ Popup = ClassUI(Group) {
     __init = function(self, GUI, content)
         Group.__init(self, GUI)
         self.content = content
+        self._closed = false
         content:SetParent(self)
         LayoutHelpers.AtLeftTopIn(content, self)
 
@@ -49,6 +51,7 @@ Popup = ClassUI(Group) {
 
         ---- Close when the escape key is pressed.
         EscapeHandler.PushEscapeHandler(function()
+            self._closed = true
             EscapeHandler.PopEscapeHandler()
             self:OnEscapePressed()
         end)
--- a/lua/ui/controls/popups/popup.lua
+++ b/lua/ui/controls/popups/popup.lua
@@ -57,8 +57,10 @@ Popup = ClassUI(Group) {
     --- Close the dialog.
     ---@param self Popup
     Close = function(self)
-        self:OnClosed()
-        self:Destroy()
+        if not IsDestroyed(self) then
+            self:OnClosed()
+            self:Destroy()
+        end
     end,
 
     --- Called when escape is pressed if the dialog is open. Defaults to closing the dialog.
@@ -76,6 +78,16 @@ Popup = ClassUI(Group) {
     --- Called when the dialog is closed via any method.
     ---@param self Popup
     OnClosed = function(self)
-        EscapeHandler.PopEscapeHandler()
+        if not self._closed then
+            self._closed = true
+            EscapeHandler.PopEscapeHandler()
+        end
+    end,
+
+    --- Called when the control is destroyed
+    ---@param self Popup
+    OnDestroy = function(self)
+        self:OnClosed()
+        Group.OnDestroy(self)
     end,
 }
--- a/lua/ui/controls/popups/popup.lua
+++ b/lua/ui/controls/popups/popup.lua
@@ -5,18 +5,18 @@ local Bitmap = import("/lua/maui/bitmap.lua").Bitmap
 local EscapeHandler = import("/lua/ui/dialogs/eschandler.lua")
 
 --- Base class for popups. A popup appears on top of other UI content, darkens the content behind it,
--- and draws a standard background behind its content. You'll probably want to extend it to do
--- something more involved, or use it as-is if you want to manually assemble your popup UI Group.
+--- and draws a standard background behind its content
 ---@class Popup : Group
----@field _closed boolean
----@field _escapeHandler function
 ---@field Trash TrashBag
+---@field content Control
+---@field _closed boolean
+---@field _escapeHandler? function
 Popup = ClassUI(Group) {
 
-    --- Create a new popup
+    --- Creates a new popup
     ---@param self Popup
-    ---@param GUI Control       -- A reference to the lobby's GUI object (dialogs should all be parented off there)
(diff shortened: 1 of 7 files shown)


lua/ui/controls/popups/popup.lua
```diff
@@ -48,34 +53,62 @@ Popup = ClassUI(Group) {
...
+        EscapeHandler.PushEscapeHandler(escapeHandler)
+
+        self.Trash:Add({
+            Destroy = function()
+                if self._escapeHandler then
+                    EscapeHandler.PopEscapeHandler(self._escapeHandler)
+                    self._escapeHandler = nil
+                end
+            end,
+        })
```
lL1l1: Move into class scope as part of Popup.OnDestroy

lL1l1: You also don't need the new self.Trash field after removing this cleanup function.

lua/ui/lobby/lobby.lua
```diff
@@ -2234,7 +2239,21 @@ local function TryLaunch(skipNoObserversCheck)
...
+    launchInProgress = true
+    if GUI and GUI.launchGameButton and not IsDestroyed(GUI.launchGameButton) then
+        GUI.launchGameButton:Disable()
+    end
+
     if not EveryoneHasEstablishedConnections(gameInfo.GameOptions.AllowObservers) then
+        launchInProgress = false
+        if HostUtils and HostUtils.RefreshButtonEnabledness then
+            HostUtils.RefreshButtonEnabledness()
+        elseif GUI and GUI.launchGameButton and not IsDestroyed(GUI.launchGameButton) then
```
lL1l1:
```suggestion
        elseif not IsDestroyed(GUI.launchGameButton) then
```

lua/ui/lobby/lobby.lua
```diff
@@ -2234,7 +2239,21 @@ local function TryLaunch(skipNoObserversCheck)
...
+        return
+    end
+    launchInProgress = true
+    if GUI and GUI.launchGameButton and not IsDestroyed(GUI.launchGameButton) then
+        GUI.launchGameButton:Disable()
+    end
+
     if not EveryoneHasEstablishedConnections(gameInfo.GameOptions.AllowObservers) then
+        launchInProgress = false
+        if HostUtils and HostUtils.RefreshButtonEnabledness then
```
lL1l1:
```suggestion
        if HostUtils.RefreshButtonEnabledness then
```

lua/ui/lobby/lobby.lua
```diff
@@ -2234,7 +2239,21 @@ local function TryLaunch(skipNoObserversCheck)
         HostUtils.KickObservers("GameLaunched")
     end
 
+    if launchInProgress then
+        return
+    end
+    launchInProgress = true
+    if GUI and GUI.launchGameButton and not IsDestroyed(GUI.launchGameButton) then
```
lL1l1: AI usually doesn't pick up that FA just allows indexing nil (returns nil). IsDestroyed also accepts nil and returns true in that case, so we can simplify everything.
```suggestion
    if not IsDestroyed(GUI.launchGameButton) then
```

lL1l1 (commented):
Were you able to replicate the crash without these fixes? Just wondering because nobody has replicated it before in a dev environment.
The changes are fine since adding safeguards for engine-related behavior makes sense, but if the observer issue can't be replicated then I have doubts it would actually fix it. A specific example that comes to mind is that when you try to manually kick observers but they don't get kicked (you can't attempt to kick again unless you move them to a new row in the obs list), the lobby will freeze for the host when they try to start the game.

juerghegglin:
> Were you able to replicate the crash without these fixes? Just wondering because nobody has replicated it before in a dev environment. The changes are fine since adding safeguards for engine-related behavior makes sense, but if the observer issue can't be replicated then I have doubts it would actually fix it. A specific example that comes to mind is that when you try to manually kick observers but they don't get kicked (you can't attempt to kick again unless you move them to a new row in the obs list), the lobby will freeze for the host when they try to start the game.

Yeah I was wondering about that 'observer kick prior to game' bug as well - has this link ever been established? What I did here is I prompted AI to first replicate the bug with a test and only then fix with a regression, but I didn't double check though. I'll take another close look at what we have and get back with a clearer answer. Thanks for the review so far!

juerghegglin:
applied your suggestions and ready for another review.

Just to be clear, I noticed a bug when launching the game that seemed to occure based on my click-through-speed. I noticed that when I'm fast to click through from "Launch Game" to "Kick Observers?" then my lobby would freeze at times and looking at the destroy code there does seem to be the two issues tackled in the PR. So I didn't reproduce the bug locally, only with actual games and the unit+regression tests only test the engine cleanup issues.

juerghegglin:
@coderabbitai resume

PR #7247, opened 2026-08-25
Cleanup utils and add annotations

speed2CZ:
- Added missing annotations to string functions
- `StringJoin` now points to `table.concat`. Since thats exactly what it did.
- `StringStartsWith` points directly to `StringStarts` since one of them is duplicate due oversight of existing function by someone in the past.

## Additional context
The string functions are still bit messy, they should probably all be contained in the global `string` library, same as `table` library is extended in the very same file. But that's probably for another PR.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected string joining to consistently match established behavior.
  * Consolidated duplicate string-prefix functionality while preserving existing compatibility.
  * Improved string utility handling and negative-token selection behavior.

* **Documentation**
  * Added type annotations to string and math utilities, improving editor assistance and developer-facing documentation.

### DIFF
--- a/lua/system/utils.lua
+++ b/lua/system/utils.lua
@@ -656,7 +656,8 @@ function StringSplit(str, sep)
     sep = sep or ":"
     local fields = {}
     local pattern = string.format("([^%s]+)", sep)
-    local _ = str:gsub(pattern, function(c) fields[table.getn(fields)+1] = c end)
+    ---@diagnostic disable-next-line: discard-returns
+    str:gsub(pattern, function(c) fields[table.getn(fields)+1] = c end)
     return fields
 end
 
@@ -717,7 +718,8 @@ end
 ---@return string
 function StringReverse(str)
     local tbl =  {}
-    local _ = str:gsub(".", function(c) table.insert(tbl,c) end)
+    ---@diagnostic disable-next-line: discard-returns
+    str:gsub(".", function(c) table.insert(tbl,c) end)
     tbl = table.reverse(tbl)
     return table.concat(tbl)
 end
--- a/lua/system/utils.lua
+++ b/lua/system/utils.lua
@@ -694,6 +694,7 @@ end
 
 --- Prepends a string with specified symbol or one space
 ---@param str string
+---@param symbol string? # Defaults to `" "`
 ---@return string
 function StringPrepend(str, symbol)
     if not symbol then symbol = ' ' end
--- a/lua/system/utils.lua
+++ b/lua/system/utils.lua
@@ -677,12 +677,14 @@ end
 
 --- Adds comma as thousands separator in specified value
 --- e.g. StringComma(10000) --> 10,000
+---@param value number
 ---@return string
 function StringComma(value)
-    local str = value or 0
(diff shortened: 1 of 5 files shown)


lua/system/utils.lua
```diff
@@ -646,42 +646,38 @@ end)
...
 
---- Returns items as a single string, separated by the delimiter
-function StringJoin(items, delimiter)
-    local str = "";
-    for k,v in items do
-        str = str .. v .. delimiter
-    end
-    return str
-end
+StringJoin = table.concat
```
Mozd-122, line 649: These don't seem to produce identical results in testing:

`print(StringJoin({"test1", "test2", "test3"}, ", "))`
`print(table.concat({"test1", "test2", "test3"}, ", "))`


`test1, test2, test3,`
`test1, test2, test3`

speed2CZ: Yes, but I believe the desired result is without the repetition after the last string.

The only usecase I found it the repo is with space as the joining string for putting together console command and the extra space at the end didnt matter.

lua/system/utils.lua
```diff
@@ -646,42 +646,38 @@ end)
...
----@param sep string
+---@param sep? string Defaults to `:`
 ---@return string[]
 function StringSplit(str, sep)
-    local sep, fields = sep or ":", {}
+    sep = sep or ":"
+    local fields = {}
     local pattern = string.format("([^%s]+)", sep)
-    str:gsub(pattern, function(c) fields[table.getn(fields)+1] = c end)
+    local _ = str:gsub(pattern, function(c) fields[table.getn(fields)+1] = c end)
```
lL1l1: I think it reads better if you just exclude the local variable.
```suggestion
    str:gsub(pattern, function(c) fields[table.getn(fields)+1] = c end)
```

lua/system/utils.lua
```diff
@@ -710,25 +710,37 @@ end
 
 --- Reverses order of letters for specified string
 --- e.g. StringReverse('abc123') --> 321cba
+---@param str string
+---@return string
 function StringReverse(str)
     local tbl =  {}
-    str:gsub(".", function(c) table.insert(tbl,c) end)
+    local _ = str:gsub(".", function(c) table.insert(tbl,c) end)
```
lL1l1:
```suggestion
    str:gsub(".", function(c) table.insert(tbl,c) end)
```

lL1l1 (approved):
I found and added some missing annotations out of the functions touched. Technically this is good to go, but I did leave a comment about the string:gsub style.

lua/system/utils.lua
speed2CZ: It was either what I did or disabling the diagnostic about no discard return type. I dont have strong feeling about either

lL1l1: That diagnostic shouldn't show up because `string.gsub` doesn't have the `---@nodiscard` annotation. Can you confirm that by going to your definition of gsub (press f12 on the function)?

speed2CZ: Yeah, it was a old lua lsp

PR #7250, opened 2026-08-26
Move custom String functions into default string library

speed2CZ:
Ive changed this PR to contain only the changes to strings. Vectors will get a new PR once it's more polished.

## Changes
- All custom `String_` functions are moved to the default `string`. Same pattern as the extended `table`.
- Removed imports of the `utils.lua` file, as this is a global file and doesnt need to be imported. When it does, it is actually redefining the globals.

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added common text operations to the `string` library, including splitting, joining, capitalization, formatting, and prefix/suffix checks.
  * Added a **Restart Session** option to the quit dialog when available.
  * Added access to wreckage positions collected from scenario army groups.
  * Retained compatibility with existing string helper names.

* **Bug Fixes**
  * Fixed scenario setup logic that could prevent faction configuration from completing correctly.

* **Tests**
  * Updated string utility tests to use the standardized API.


lua/system/utils.lua
```diff
@@ -1119,4 +1154,267 @@ function vector_metatable.__mul(a, b)
...
+    ---Returns surface elevation at given position. Takes water into count.
+    ---@param vector Vector
+    ---@return number
+    GetSurfaceHeight = function(vector)
+        return GetSurfaceHeight(vector[1], vector.z or vector[2])
+    end,
+    ---Returns elevation at given position. Ignores water surface.
+    ---@param vector Vector
+    ---@return number
+    GetTerrainHeight = function(vector)
```
4z0t: There is no such function in UI code.

lua/system/utils.lua
```diff
@@ -1119,4 +1154,267 @@ function vector_metatable.__mul(a, b)
...
+            error("Vector:Distance2DTo: invalid argument", 2)
+        end
+
+        local dx, dz = to[1] - from[1], to[3] - from[3]
+        return MathSqrt(dx*dx + dz*dz)
+    end,
+    ---Returns surface elevation at given position. Takes water into count.
+    ---@param vector Vector
+    ---@return number
+    GetSurfaceHeight = function(vector)
```
4z0t: There is no such function in UI code.

4z0t (changes requested):
Overall idea is good, but it lacks scope and finesse. Vector mustn't know about terrain at all. Most of vector functions can be moved into its meta with engine patch where it will be way more performant, including string ones.

4z0t:
I made an issue before https://github.com/FAForever/FA-Binary-Patches/issues/93

speed2CZ:
Ive changed this PR to contain only the changes to strings. Vectors will get a new PR once it's more polished.

PR #7256, opened 2026-08-28
add guard against OnDefeat failures for AI

relent0r:
There are certain niche scenarios when an AIs platoon will not have cplatoon functions available. After analyzing a replay it was found that when an AI ACU was destroyed units around it that were destroyed in the blast had a chance of being part way through the destroy state when the ondefeat function is called. 

Tested based on a replay with the issue present. The following log watch generated.
```
warning: Error running lua script: ...forever\replaydata\gamedata\lua.nx2\lua\simutils.lua(1611): attempt to call method `Stop' (a nil value)
         stack traceback:
             ...forever\replaydata\gamedata\lua.nx2\lua\simutils.lua(1611): in function <...forever\replaydata\gamedata\lua.nx2\lua\simutils.lua:1598>
             ...aforever\replaydata\gamedata\lua.nx2\lua\aibrain.lua(445): in function `SetDefeatStatus'
             ...aforever\replaydata\gamedata\lua.nx2\lua\aibrain.lua(465): in function `OnDefeat'
             ...ua\sim\victorycondition\abstractvictorycondition.lua(406): in function `DefeatForArmy'
             ...a\lua.nx2\lua\sim\victorycondition\unitcondition.lua(62): in function `EvaluateVictoryCondition'
             ...ua\sim\victorycondition\abstractvictorycondition.lua(212): in function <...ua\sim\victorycondition\abstractvictorycondition.lua:210>
```

After adding guards the issue no longer happened and the ondefeat function could complete successfully.

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved AI disable handling during defeat events by safely checking platoon functions before calling them.
  * Prevented errors when optional platoon commands are unavailable.
  * Corrected platoon disband handling so the command runs when supported.
  * Improved stability when AI-controlled armies are defeated.

### DIFF
--- a/changelog/snippets/ai.7256.md
+++ b/changelog/snippets/ai.7256.md
@@ -1,1 +1,1 @@
-- Add guards for platoon function calls in the DisableAI function to stop errors during the OnDefeat call.
+- Fix error when an army is defeated related to missing AI platoon functions (#7256).
--- a/changelog/snippets/ai.7256.md
+++ b/changelog/snippets/ai.7256.md
@@ -1,1 +1,1 @@
-- Add guards for platoon function calls in the DisablAI function to stop errors during the OnDefeat call.
+- Add guards for platoon function calls in the DisableAI function to stop errors during the OnDefeat call.
--- a/changelog/snippets/ai.7256.md
+++ b/changelog/snippets/ai.7256.md
@@ -1,0 +1,1 @@
+- Add guards for platoon function calls in the DisablAI function to stop errors during the OnDefeat call.
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -1437,8 +1437,12 @@ function DisableAI(self)
             if not unit.Dead then
                 local handle = unit.PlatoonHandle
                 if handle and self:PlatoonExists(handle) then
-                    handle:Stop()
-                    handle:PlatoonDisbandNoAssign()
+                    if handle.Stop then
+                        handle:Stop()
+                    end
+                    if handle.PlatoonDisbandNoAssig then
+                        handle:PlatoonDisbandNoAssign()
+                    end
                 end
                 IssueStop({ unit })
                 IssueToUnitClearCommands(unit)
--- a/lua/SimUtils.lua
+++ b/lua/SimUtils.lua
@@ -1440,7 +1440,7 @@ function DisableAI(self)
                     if handle.Stop then
                         handle:Stop()
                     end
-                    if handle.PlatoonDisbandNoAssig then
+                    if handle.PlatoonDisbandNoAssign then
                         handle:PlatoonDisbandNoAssign()
                     end
                 end


lua/SimUtils.lua
```diff
@@ -1437,8 +1437,12 @@ function DisableAI(self)
...
                 local handle = unit.PlatoonHandle
                 if handle and self:PlatoonExists(handle) then
-                    handle:Stop()
-                    handle:PlatoonDisbandNoAssign()
+                    if handle.Stop then
+                        handle:Stop()
+                    end
+                    if handle.PlatoonDisbandNoAssig then
+                        handle:PlatoonDisbandNoAssign()
+                    end
```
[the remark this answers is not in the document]
relent0r: good bot!

lua/SimUtils.lua
```diff
@@ -1437,8 +1437,12 @@ function DisableAI(self)
...
                 local handle = unit.PlatoonHandle
                 if handle and self:PlatoonExists(handle) then
-                    handle:Stop()
-                    handle:PlatoonDisbandNoAssign()
+                    if handle.Stop then
+                        handle:Stop()
+                    end
+                    if handle.PlatoonDisbandNoAssign then
+                        handle:PlatoonDisbandNoAssign()
+                    end
```
lL1l1, line 1445: I read a bit of the platoon code, and it seems like there's some manual `setmetatable` use that could be leaving platoons with an incomplete/wrong platoon class?
For example, `AIPlatoonAdaptiveSilo` has a typo using `AIPlatoonAdaptiveSiloBehavior` instead of `AIPlatoonAdaptiveSilo`, but it's not being used anywhere so nobody caught the issue.
https://github.com/FAForever/fa/blob/41797807d08507e0ad9eb487815213f7c2f3b16c/lua/aibrains/platoons/platoon-adaptive-silo.lua#L249

Could you post the replay and maybe investigate the metatable of the platoon that fails at being destroyed (for example if it is a mod using the adaptive silo)? These functions should never be missing because they're part of the base class. Even if the platoon is from a mod.

changelog/snippets/ai.7256.md
```diff
@@ -0,0 +1 @@
+- Add guards for platoon function calls in the DisableAI function to stop errors during the OnDefeat call.
```
lL1l1: Need to add PR#.
```suggestion
- Add guards for platoon function calls in the DisableAI function to stop errors during the OnDefeat call (#7256).
```
I prefer describing what is fixed instead of how it is fixed since it is more relatable to players who have experienced it.
```suggestion
- Fix error when an army is defeated related to missing AI platoon functions (#7256).
```

lL1l1 (commented):
Technically this works but it doesn't seem like it should be fixed this way.

lua/SimUtils.lua
relent0r: This was based off a replay for my own AI.
https://replay.faforever.com/27584457

I don't disagree that it may have something to do with the state machine platoon implementation. The platoons in question were otherwise functioning fine.

I thought this would be the safest change to make that would be low risk of unforeseen consequences.

I did a dump of the platoon table when troubleshooting as I thought maybe I'd accidentally set it to an empty table, but the resulting dump looked the same as any other platoon and had the c object attached so I was thinking that the platoon was mid was through its own destruction. 

In my testing I only had the issue happen once in say 10 games but a player pointed out that it happened alot more often for them with the difference being that they had a much higher player count.

relent0r: If you have suggestions of I could better troubleshoot it then I'm happy to try find the core reason.

lL1l1: I did some metatable trickery in this branch to track down tracebacks of what changes were done to the platoon that failed:
https://github.com/FAForever/fa/tree/investigate-ai-ondefeat-failures

It's useful so I'll make a PR for it on a more organized branch.

I have included the output log in the branch files. What stands out to me is this final assignment of the failing platoon's metatable to `nil`, which would cause it to not have the necessary commands.
```log
"[283.745270] Assign \"<metatable>\" as nil
from stack traceback:
...\lua\globalinit.lua:18>
...\lua\system\class.lua(802): in function `ChangeState'
...\mods\rngai\lua\ai\statemachines\platoon-base-rng.lua(284): in function `ChangeStateExt'
...\mods\rngai\lua\ai\statemachineutilities.lua(1777): in function `cb'
...\lua\sim\unit.lua(4177): in function `DoUnitCallbacks'
...\lua\sim\unit.lua(4321): in function `DoOnFailedToBuildCallbacks'
...\lua\sim\unit.lua:2936>
...\lua\sim\units\constructionunit.lua:117>
[C]: in function `IssueClearCommands'
...\mods\rngai\lua\ai\statemachines\platoon-engineer-resource.lua(726): in function <...\mods\rngai\lua\ai\statemachines\platoon-engineer-resource.lua:416>"

```

After investigating RNGAI, I saw:
```lua
            unit.PlatoonHandle:ChangeStateExt(unit.PlatoonHandle.PerformBuildTask)
```
Where `PerformBuildTask` is a State.
But that State only exists in `lua\AI\StateMachines\platoon-engineer-utility.lua` `AIPlatoonEngineerBehavior`
Our engineer is clearly using `lua\AI\StateMachines\platoon-engineer-resource.lua` `AIPlatoonEngineerBehavior`,
Despite having the same name, that file's class does not have `PerformBuildTask` and does not inherit from the utility engineer class.
Therefore we change the unit to a `nil` State, which loses the `Stop` and `PlatoonDisbandNoAssign` methods.
`ChangeState` has no nil check, so your `nil` does go directly into the metatable:
https://github.com/FAForever/fa/blob/653be9b4d35f0cf6ed9f513201d18de05513e17e/lua/system/class.lua#L784-L818

relent0r:
Thanks @lL1l1 I hadn't considered that getting one state machine to perform an adhoc build command would then fire a callback thats designed for a different state machine. Makes sense. In that case I guess I can fix this within my AI rather than in this PR.

I guess to avoid this for myself and other people that use the platoon states performing a nil check would also be useful in the base class. Then I could at least make it warn.

PR #7261, opened 2026-08-28
Always clear build templates when ending command mode

lL1l1:
Fixes ending template build with right click carrying over the template into cheat spawn's visuals.

Start template build mode. Cancel with right click. Start cheat spawn mode. It should not show the template that you canceled.

## Additional context
I looked in the engine and documented every instance of the command mode ending with `isCancel = false`. I couldn't understand why templates weren't cleared if it was a right click order. It doesn't make sense to me because it ends the command mode regardless, which loses the build mode capability that is required for templates to work, and if you want to re-enable the build mode capability, then you should re-enable it with a template if you want a template, not have some remainder template carrying over because you pressed right click.

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
  - Build templates are now consistently cleared when command mode ends.
  - Prevented canceled build templates from appearing during unit cheating.
  - Improved cleanup when leaving command mode, including when a command is stopped normally or canceled.

### DIFF
--- a/changelog/snippets/fix.7261.md
+++ b/changelog/snippets/fix.7261.md
@@ -1,1 +1,1 @@
-- Your explanation here... [Don't forget to change the category in the filename] (#7261).
+- Always clear build templates when ending command mode (#7261). This fixes build templates that were canceled by right click appearing when cheating in units.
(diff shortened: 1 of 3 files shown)


speed2CZ:
There are couple of leftover LOG statements, those should probably be commented out?

PR #7270, opened 2026-09-01
Fix more annotations

speed2CZ:
Compilation of annotation fixes Ive done when debugging the new lua ls addon
- add new types
- fixed types
- removed duplicate types
- added type casting when safe
- fixes `@see` references
- cleaned up forward declarations with default false value `local something = false` to just `local something`
- cleaned up invalid args passed into functions
- added local keywords to variables that were missing it
- cleaned up function return types
- removed `return` statements after `error()` calls

Works best with #7267

## Checklist


lL1l1 (approved):
I briefly read all the annotations for style, and the ones I found to be more complex or weird I checked for functionality.

PR #7272, opened 2026-09-02
Refactor formation commands

speed2CZ:
- Created a new file with helper functions for formation commands that orient the formations correctly by angle on each node.
- Refactored platoon and ScenarioFramework to use the new formation commads.
- Attack platoons in campaign now orient formations correctly, so units move through path nodes faster.

## Additional context
Platoons already had helper functions to Issue various formation commands with proper formation orientation, this PR extracts that code into it's own module, so its easy to use from elsewhere. It also removes the code repetition that was present in the existing functions

## Checklist

## Summary by CodeRabbit

- **New Features**
  - Units now face the direction of travel at each route waypoint, improving formation movement.
  - Platoons can use selected formations while moving, patrolling, and attacking along routes.
  - Group movement and attack routes are supported.
  - Campaign attack platoons move through path nodes more efficiently.
  - Route commands can return the commands issued for further handling.

- **Bug Fixes**
  - Empty routes now trigger a warning and exit safely instead of causing an error.

### DIFF
--- a/lua/ScenarioFramework.lua
+++ b/lua/ScenarioFramework.lua
@@ -27,6 +27,7 @@ local SyncVoice = import("/lua/simsyncutils.lua").SyncVoice
 local CategoryToString = import("/lua/sim/categoryutils.lua").ToString
 local Cinematics = import("/lua/cinematics.lua")
 local Game = import("/lua/game.lua")
+local FormationCommands = import("/lua/sim/formationcommands.lua")
 local ScenarioUtils = import("/lua/sim/scenarioutilities.lua")
 local SimCamera = import("/lua/simcamera.lua").SimCamera
 local SimUIVars = import("/lua/sim/simuistate.lua")
@@ -430,83 +431,6 @@ CreateUnitToPositionDistanceTrigger = TriggerFile.CreateUnitToPositionDistanceTr
 CreateUnitToMarkerDistanceTrigger = CreateUnitToPositionDistanceTrigger -- got renamed for some reason
 CreateUnitNearTypeTrigger = TriggerFile.CreateUnitNearTypeTrigger
 
--- platoon functions REQUIRE `squad` to be non-nil when present
-
---- Orders a platoon to move along a route
----@param platoon Platoon
----@param route (MarkerName | Vector)[]
----@param squad? PlatoonSquads
-function PlatoonMoveRoute(platoon, route, squad)
-    for _, node in route do
-        if type(node) == 'string' then
-            node = ScenarioUtils.MarkerToPosition(node)
-        end
-        if squad then
-            platoon:MoveToLocation(node, false, squad)
-        else
-            platoon:MoveToLocation(node, false)
-        end
-    end
-end
-
---- Orders platoon to patrol a route
----@param platoon Platoon
----@param route (MarkerName | Vector)[]
----@param squad? PlatoonSquads
-function PlatoonPatrolRoute(platoon, route, squad)
-    for _, node in route do
-        if type(node) == 'string' then
-            node = ScenarioUtils.MarkerToPosition(node)
-        end
-        if squad then
-            platoon:Patrol(node, squad)
-        else
-            platoon:Patrol(node)
-        end
-    end
-end
-
---- Orders a platoon to attack-move along a route
----@param platoon Platoon
----@param route (MarkerName | Vector)[]
----@param squad? PlatoonSquads
-function PlatoonAttackRoute(platoon, route, squad)
-    for _, node in route do
-        if type(node) == 'string' then
-            node = ScenarioUtils.MarkerToPosition(node)
-        end
-        if squad then
-            platoon:AggressiveMoveToLocation(node, squad)
-        else
-            platoon:AggressiveMoveToLocation(node)
-        end
-    end
-end
-
---- Orders a platoon to move along a chain
----@param platoon Platoon
----@param chain ChainName
----@param squad? PlatoonSquads
-function PlatoonMoveChain(platoon, chain, squad)
-    PlatoonMoveRoute(platoon, ScenarioUtils.ChainToPositions(chain), squad)
-end
-
---- Orders a platoon to patrol along a chain
----@param platoon Platoon
----@param chain ChainName
----@param squad? PlatoonSquads
-function PlatoonPatrolChain(platoon, chain, squad)
-    PlatoonPatrolRoute(platoon, ScenarioUtils.ChainToPositions(chain), squad)
-end
-
---- Orders a platoon to attack-move through a chain
----@param platoon Platoon
----@param chain ChainName
----@param squad? PlatoonSquads
-function PlatoonAttackChain(platoon, chain, squad)
-    PlatoonAttackRoute(platoon, ScenarioUtils.ChainToPositions(chain), squad)
-end
-
 --- Orders a group to patrol along a chain
 ---@param units Unit[]
 ---@param chain ChainName
@@ -529,12 +453,17 @@ function GroupPatrolRoute(units, route)
 end
 
 --- Orders a group to patrol a route in formation
+---
+---`IssueFormPatrol` Does NOT return `SimCommand`
 ---@param units Unit[]
 ---@param chain ChainName
 ---@param formation string
 function GroupFormPatrolChain(units, chain, formation)
-    for _, pos in ScenarioUtils.ChainToPositions(chain) do
-        IssueFormPatrol(units, pos, formation, 0)
+    local path = ScenarioUtils.ChainToPositions(chain)
+    local angles = FormationCommands.GetAnglesForRoute(path)
+
+    for i, pos in ipairs(path) do
+        IssueFormPatrol(units, pos, formation, angles[i])
     end
 end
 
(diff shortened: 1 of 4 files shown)


lua/sim/FormationCommands.lua
```diff
@@ -0,0 +1,98 @@
...
+local baseZeroAngle = Vector(0, 0, 1)
+
+---Converts a table with position to vector if its not a vector already
+---@param pos table<{[1]: number, [2]: number, [3]: number}>|Vector
+local function positionToVector(pos)
+    if pos.x then
+        return pos
+    end
+    return Vector(pos[1], pos[2], pos[3])
+end
```
lL1l1, line 29: You could set the metatable of pos to the meta of Vector instead.
```suggestion
local VectorMt = getmetatable(Vector2(0,0))

---Converts a table with position to vector if its not a vector already
---@param pos table<{[1]: number, [2]: number, [3]: number}>|Vector
local function positionToVector(pos)
    local mt = getmetatable(pos)
    if not mt then
        return setmetatable(pos, VectorMt)
    elseif mt == VectorMt then 
        return pos 
    else
        error("Position has a metatable that is not the Vector metatable")
    end
end
```

I suppose it would break if the vector metatable is modified after importing the file though.

lua/sim/FormationCommands.lua
```diff
@@ -0,0 +1,98 @@
...
+
+--- Returns angles the unit formation should face in `IssueFormation` commands for given `path`.
+---@param path Vector[]
+---@param initialPosition? Vector Position that is used to calculate the first angle of the path. Defaults to last position of the path
+---@return number[] angles List of angles for every position
+function GetAnglesForRoute(path, initialPosition)
+    local count = tableGetn(path)
+    if count == 0 then return {} end
```
lL1l1, line 37:
```suggestion
    if count == 0 then return EmptyTable end
```
If the table is never modified this is ok.

lua/ScenarioFramework.lua
```diff
@@ -430,59 +431,175 @@ CreateUnitToPositionDistanceTrigger = TriggerFile.CreateUnitToPositionDistanceTr
...
+            end
+            local angles = FormationCommands.GetAnglesForRoute(route, platoon:GetSquadPosition(squadName))
+            FormationCommands.UnitsFormationOrder(units, IssueFormMove, route, angles, formation)
         end
     end
 end
 
+--- Orders platoon to patrol a route
+---
+--- `IssueFormPatrol` Does NOT return `SimCommand`
```
lL1l1, line 550: This comment seems out of place. Did you mean that this function does not return `SimCommand` like the other platoon order route functions? In that case, PlatoonMoveRoute needs to return a command too.

lua/ScenarioFramework.lua
```diff
@@ -430,59 +431,175 @@ CreateUnitToPositionDistanceTrigger = TriggerFile.CreateUnitToPositionDistanceTr
...
+    routeToPositions(route)
+
+    local cmd
+    for _, squadName in pairs(squads) do
+        local units = platoon:GetSquadUnits(squadName)
+
+        if not table.empty(units) then
+            if formation == 'NoFormation' then
+                GroupAttackRoute(units, route)
+                return {}
```
lL1l1, line 594:
```suggestion
                return EmptyTable
```

lua/platoon.lua
```diff
@@ -28,6 +29,7 @@ local tableEmpty = table.empty
 local tableSort = table.sort
 local tableCopy = table.copy
 
+local pairs, ipairs = pairs, ipairs
```
lL1l1, line 32: We shouldn't use `pairs`/`ipairs` in code because `ipairs` is 18x slower than an incrementing loop and `pairs` is just slightly less efficient than writing nothing (where engine inserts `next` as the iterator).

lua/platoon.lua
```diff
@@ -4138,176 +4149,70 @@ Platoon = Class(moho.platoon_methods) {
...
-        if not path or not (type(path) == 'table') then
-            error("IssueAggressiveMoveAlongRoute: The path is not a table. For paths with only one node, use { node } as the path.")
-            return { }
-        end
+        formation = formation or self.PlatoonData.UseFormation or 'NoFormation'
 
         if tableEmpty(path) then
-            error("IssueAggressiveMoveAlongRoute: The path is empty.");
-            return { }
+            WARN("Platoon:IssueAggressiveMoveAlongRoute: The path is empty.")
```
lL1l1, line 4196: It would be useful to create a debug component for platoons to have some standardized info/text in the logs, like there is for weapons, projectiles, and units.

lua/platoon.lua
```diff
@@ -4138,176 +4149,70 @@ Platoon = Class(moho.platoon_methods) {
...
+            return {}
```
lL1l1, line 4197:
```suggestion
            return EmptyTable
```

lua/platoon.lua
```diff
@@ -4138,176 +4149,70 @@ Platoon = Class(moho.platoon_methods) {
...
-            error("IssuePatrolAlongRoute: The path is not a table. For paths with only one node, use { node } as the path.")
-            return { }
-        end
+        formation = formation or self.PlatoonData.UseFormation or 'NoFormation'
 
         if tableEmpty(path) then
-            error("IssuePatrolAlongRoute: The path is empty.");
-            return { }
+            WARN("Platoon:IssuePatrolAlongRoute: The path is empty.")
+            return {}
```
lL1l1, line 4162:
```suggestion
            return EmptyTable
```

lua/platoon.lua
```diff
@@ -4316,89 +4221,44 @@ Platoon = Class(moho.platoon_methods) {
...
-            error("IssueMoveAlongRoute: The path is not a table. For paths with only one node, use { node } as the path.")
-            return { }
-        end
+        formation = formation or self.PlatoonData.UseFormation or 'NoFormation'
 
         if tableEmpty(path) then
-            error("IssueMoveAlongRoute: The path is empty.");
-            return { }
+            WARN("Platoon:IssueMoveAlongRoute: The path is empty.")
+            return {}
```
lL1l1, line 4232:
```suggestion
            return EmptyTable
```

lua/platoon.lua
```diff
@@ -4316,89 +4221,44 @@ Platoon = Class(moho.platoon_methods) {
...
-
-        -- pre-compute the angles
-        local angles = { }
-        for k = 1, count do
+        local angles = FormationCommands.GetAnglesForRoute(path, self:GetPlatoonPosition())
+        local units = self:GetPlatoonUnits()
 
-            local curr = path[k - 1]
-            local next = path[k]
+        -- Last command should be aggressive move
```
lL1l1, line 4254: Why does the last command have to be aggressive move?

lua/ScenarioFramework.lua
```diff
@@ -430,59 +431,175 @@ CreateUnitToPositionDistanceTrigger = TriggerFile.CreateUnitToPositionDistanceTr
...
+        if not table.empty(units) then
+            if formation == 'NoFormation' then
+                GroupMoveRoute(units, route)
+                return
+            end
+            local angles = FormationCommands.GetAnglesForRoute(route, platoon:GetSquadPosition(squadName))
+            FormationCommands.UnitsFormationOrder(units, IssueFormMove, route, angles, formation)
         end
     end
 end
```
lL1l1, line 546: Refactor the NoFormation condition as an early exit by using `GroupMoveRoute` on all the units in the platoon instead of by-squad.
```suggestion
function PlatoonMoveRoute(platoon, route, squad, formation)
    formation = formation or platoon:GetFormationFromPlatoonData()
    if formation == 'NoFormation' then
        local units = platoon:GetPlatoonUnits()
        GroupMoveRoute(units, route)
        return
    end
    local squads = FormationCommands.GetSquadsForFormationOrder(squad)
    routeToPositions(route)

    for _, squadName in pairs(squads) do
        local units = platoon:GetSquadUnits(squadName)

        if not table.empty(units) then
            local angles = FormationCommands.GetAnglesForRoute(route, platoon:GetSquadPosition(squadName))
            FormationCommands.UnitsFormationOrder(units, IssueFormMove, route, angles, formation)
        end
    end
end
```

lua/sim/FormationCommands.lua
speed2CZ: Yeah, I have a plan to redo the vector metatable and cleanup all the things using vectors, but thats for another PR

lua/platoon.lua
speed2CZ: Idk, that's how it was set previously, and since this part is used by the regular AI, I kept it the same to not break things.

lua/platoon.lua
speed2CZ: I use it mainly for readability, to clearly distinguish which loop care or dont care about the order of processing things, If it becomes a problem is some part then it can be changed, but generally I dont feel like writing a manual numerical iteration each time. The overall changes in the refactors are usually much higher by saving a lot of computation or skipping many table allocations, than what a ipair costs.

lua/platoon.lua
speed2CZ: I dont know what all the different modded AIs would do with this, so it's safer to return a new table.

lua/ScenarioFramework.lua
speed2CZ: That would change the behaviour as `platoon:MoveToLocation()` will issue move command to each squad. Same for all the other platoon command functions that got replaced.

PR #7290, opened 2026-09-12
Working prototype of custom sacu presets.

RabidPope:
**Description**
This allows you to set desired upgrades for an SACU and add it to the build queue on the gateway.
The enhancements are selected via the same tabs as you would for an SACU when selecting a gateway.
You can then queue it by pressing the button added to the gateway's orders panel.

Note: this demo now contains 4 main factions.

**Testing done on the proposed changes.**
Basic testing that the upgrades show on the sacu.
Repeat and pause buttons still work.
Energy cost and mass scale with what is selected.

Animation

Build time seems to increase accordingly.
Multiple gateways build different units.

Please let me know of any optimisations I can make.

Disclaimer: AI was used to assist in creation.

This exists as a mod on the faf mod vault as "Custom SACU's"

## Summary by CodeRabbit

- **New Features**
  - Added a Quantum Gateway loadout picker for UEF, Aeon, Cybran, and Seraphim SACUs.
  - Added automatic enhancement prerequisite handling and generated loadout combinations.
  - Added support for queuing selected loadouts, including Shift-click to queue five.
  - Integrated loadout controls, icons, queue behavior, and construction UI support across factions.
  - Generated combinations remain hidden from standard construction lists while available through the loadout picker.
- **Bug Fixes**
  - Corrected SACU loadout names and icons displayed in unit and queue views.
- **Documentation**
  - Added a tooltip explaining how to queue SACU loadouts.

### DIFF
--- a/lua/system/blueprints-sacu-combos.lua
+++ b/lua/system/blueprints-sacu-combos.lua
@@ -1,12 +1,5 @@
 --******************************************************************************************************
 --** Cybran SACU loadout combos
---**
---** Generates a hidden EnhancementPreset for every valid URL0301 slot combination
---** that is not already a named factory preset. IDs are stable: enhancement names
---** are sorted before the suffix is built, so clients cannot disagree on the string.
---**
---** Existing role presets (RAS, Engineer, Combat, ...) are left untouched and stay
---** on the Quantum Gateway construct grid.
 --******************************************************************************************************
 
 local TableInsert = table.insert
@@ -18,6 +11,17 @@ local ComboFactionBaseIds = {
 }
 
 local SlotOrder = { 'LCH', 'RCH', 'Back' }
+
+local ShortLabels = {
+    EMPCharge = 'EMP',
+    FocusConvertor = 'Amp',
+    NaniteMissileSystem = 'AA',
+    ResourceAllocation = 'RAS',
+    SelfRepairSystem = 'Nano',
+    StealthGenerator = 'Stealth',
+    CloakingGenerator = 'Cloak',
+    Switchback = 'Fab',
+}
 
 local function IsRemoveEnhancement(name)
     return string.sub(name, -6) == 'Remove'
@@ -131,9 +135,9 @@ end
 local function ComboUnitName(enhancements, bp)
     local labels = {}
     for _, name in enhancements do
-        TableInsert(labels, CleanLabel(name, bp.Enhancements[name]))
+        TableInsert(labels, ShortLabels[name] or CleanLabel(name, bp.Enhancements[name]))
     end
-    return 'SACU (' .. table.concat(labels, ' / ') .. ')'
+    return 'SACU (' .. table.concat(labels, '/') .. ')'
 end
 
 function InjectSacuLoadoutPresets(all_bps)
@@ -167,7 +171,7 @@ function InjectSacuLoadoutPresets(all_bps)
                         if not bp.EnhancementPresets[presetName] then
                             local pretty = ComboUnitName(enhList, bp)
                             bp.EnhancementPresets[presetName] = {
-                                Description = pretty,
+                                Description = bp.Description,
                                 BuildIconSortPriority = 90,
                                 Enhancements = enhList,
                                 HelpText = pretty,
--- a/lua/system/blueprints-sacu-combos.lua
+++ b/lua/system/blueprints-sacu-combos.lua
@@ -120,15 +120,18 @@ local function ComboPresetName(enhancements)
     return 'combo_' .. table.concat(copy, '_')
 end
 
+local function CleanLabel(name, def)
+    local label = name
+    if def and def.Name then
+        label = string.gsub(def.Name, '^<LOC [^>]+>', '')
+    end
+    return label
+end
+
 local function ComboUnitName(enhancements, bp)
     local labels = {}
     for _, name in enhancements do
-        local def = bp.Enhancements[name]
-        local label = name
-        if def and def.Name then
-            label = def.Name
-        end
-        TableInsert(labels, label)
+        TableInsert(labels, CleanLabel(name, bp.Enhancements[name]))
     end
     return 'SACU (' .. table.concat(labels, ' / ') .. ')'
 end
@@ -162,14 +165,15 @@ function InjectSacuLoadoutPresets(all_bps)
                     else
                         local presetName = ComboPresetName(enhList)
                         if not bp.EnhancementPresets[presetName] then
+                            local pretty = ComboUnitName(enhList, bp)
                             bp.EnhancementPresets[presetName] = {
-                                Description = ComboUnitName(enhList, bp),
+                                Description = pretty,
                                 BuildIconSortPriority = 90,
                                 Enhancements = enhList,
-                                HelpText = ComboUnitName(enhList, bp),
+                                HelpText = pretty,
                                 SelectionPriority = 1,
                                 SortCategory = 'SORTOTHER',
-                                UnitName = ComboUnitName(enhList, bp),
+                                UnitName = pretty,
                                 HiddenInBuildMenu = true,
                             }
                             generated = generated + 1
--- a/lua/system/blueprints-sacu-combos.lua
+++ b/lua/system/blueprints-sacu-combos.lua
@@ -1,0 +1,202 @@
+--******************************************************************************************************
+--** Cybran SACU loadout combos
+--**
+--** Generates a hidden EnhancementPreset for every valid URL0301 slot combination
+--** that is not already a named factory preset. IDs are stable: enhancement names
+--** are sorted before the suffix is built, so clients cannot disagree on the string.
+--**
+--** Existing role presets (RAS, Engineer, Combat, ...) are left untouched and stay
+--** on the Quantum Gateway construct grid.
+--******************************************************************************************************
+
+local TableInsert = table.insert
+local TableGetn = table.getn
+local StringLower = string.lower
+
+local ComboFactionBaseIds = {
+    url0301 = true,
+}
+
+local SlotOrder = { 'LCH', 'RCH', 'Back' }
+
+local function IsRemoveEnhancement(name)
+    return string.sub(name, -6) == 'Remove'
+end
+
+local function SortedKeys(hashtable)
+    local keys = {}
+    for key in hashtable do
+        TableInsert(keys, key)
+    end
+    table.sort(keys)
+    return keys
+end
+
+local function EnhancementChain(enhName, slotDefs)
+    local chain = { enhName }
+    local current = enhName
+    local guard = 0
+    while slotDefs[current] and slotDefs[current].Prerequisite and guard < 8 do
+        local pre = slotDefs[current].Prerequisite
+        if not slotDefs[pre] then
+            break
+        end
+        TableInsert(chain, 1, pre)
+        current = pre
+        guard = guard + 1
+    end
+    return chain
+end
+
+local function SlotChoices(slotDefs)
+    local choices = { {} }
+    local names = SortedKeys(slotDefs)
+    for _, name in names do
+        TableInsert(choices, EnhancementChain(name, slotDefs))
+    end
+    return choices
+end
+
+local function GroupBySlot(enhancements)
+    local bySlot = {}
+    for name, def in enhancements do
+        if def.Slot and not IsRemoveEnhancement(name) then
+            bySlot[def.Slot] = bySlot[def.Slot] or {}
+            bySlot[def.Slot][name] = def
+        end
+    end
+    return bySlot
+end
+
+function EnhancementSetKey(list)
+    local copy = {}
+    for _, name in list do
+        TableInsert(copy, name)
+    end
+    table.sort(copy)
+    return string.lower(table.concat(copy, '|'))
+end
+
+local function ExistingPresetKeys(namedPresets)
+    local keys = {}
+    for presetName, preset in namedPresets do
+        if preset.Enhancements then
+            keys[EnhancementSetKey(preset.Enhancements)] = presetName
+        end
+    end
+    return keys
+end
+
+local function AllCombinations(slotChoices)
+    local combos = { {} }
+    for _, slot in SlotOrder do
+        local choices = slotChoices[slot]
+        if choices then
+            local nextCombos = {}
+            for _, prefix in combos do
+                for _, choice in choices do
+                    local merged = {}
+                    for _, name in prefix do
+                        TableInsert(merged, name)
+                    end
+                    for _, name in choice do
+                        TableInsert(merged, name)
+                    end
+                    TableInsert(nextCombos, merged)
+                end
+            end
+            combos = nextCombos
+        end
+    end
+    return combos
+end
+
+local function ComboPresetName(enhancements)
+    local copy = {}
+    for _, name in enhancements do
+        TableInsert(copy, StringLower(name))
+    end
+    table.sort(copy)
+    return 'combo_' .. table.concat(copy, '_')
+end
+
+local function ComboUnitName(enhancements, bp)
+    local labels = {}
+    for _, name in enhancements do
+        local def = bp.Enhancements[name]
+        local label = name
+        if def and def.Name then
+            label = def.Name
+        end
+        TableInsert(labels, label)
+    end
+    return 'SACU (' .. table.concat(labels, ' / ') .. ')'
+end
+
+function InjectSacuLoadoutPresets(all_bps)
+    if not all_bps or not all_bps.Unit then
+        return
+    end
+
+    local generated = 0
+    local skippedNamed = 0
+
+    for id, bp in all_bps.Unit do
+        if ComboFactionBaseIds[id] and bp.Enhancements and bp.EnhancementPresets then
+            local bySlot = GroupBySlot(bp.Enhancements)
+            local slotChoices = {}
+            for _, slot in SlotOrder do
+                if bySlot[slot] then
+                    slotChoices[slot] = SlotChoices(bySlot[slot])
+                end
+            end
+
+            local namedKeys = ExistingPresetKeys(bp.EnhancementPresets)
+            local combos = AllCombinations(slotChoices)
+
+            for _, enhList in combos do
+                if TableGetn(enhList) > 0 then
+                    local setKey = EnhancementSetKey(enhList)
+                    if namedKeys[setKey] then
+                        skippedNamed = skippedNamed + 1
+                    else
+                        local presetName = ComboPresetName(enhList)
+                        if not bp.EnhancementPresets[presetName] then
+                            bp.EnhancementPresets[presetName] = {
+                                Description = ComboUnitName(enhList, bp),
+                                BuildIconSortPriority = 90,
+                                Enhancements = enhList,
+                                HelpText = ComboUnitName(enhList, bp),
+                                SelectionPriority = 1,
+                                SortCategory = 'SORTOTHER',
+                                UnitName = ComboUnitName(enhList, bp),
+                                HiddenInBuildMenu = true,
+                            }
+                            generated = generated + 1
+                        end
+                    end
+                end
+            end
+        end
+    end
+
+    SPEW(string.format('SACU loadout: generated %d hidden Cybran combos, reused %d named presets', generated, skippedNamed))
+end
+
+function MarkHiddenSacuLoadoutPresets(all_bps)
+    if not all_bps or not all_bps.Unit then
+        return
+    end
+
+    for id, bp in all_bps.Unit do
+        local assigned = bp.EnhancementPresetAssigned
+        if assigned and assigned.Name and string.sub(assigned.Name, 1, 6) == 'combo_' then
+            local baseId = assigned.BaseBlueprintId
+            if ComboFactionBaseIds[baseId] then
+                bp.CategoriesHash = bp.CategoriesHash or {}
+                bp.CategoriesHash['SACULOADOUTCOMBO'] = true
+                bp.Categories = table.unhash(bp.CategoriesHash)
+            end
+        end
+    end
+end
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -46,6 +46,13 @@ local prevBuildables = false
 local prevSelection = false
 local prevBuildCategories = false
 
+local function UnitBuildIconId(id)
+    if id and string.sub(id, 1, 13) == 'url0301_combo' then
+        return 'url0301'
+    end
+    return id
+end
+
 -- Flag to indicate if every selected unit is a factory
 local allFactories = nil
 if options.gui_templates_factory ~= 0 then
@@ -579,8 +586,9 @@ function CommonLogic()
 
     controls.secondaryChoices.SetControlToType = function(control, type)
         local function SetIconTextures(control)
-            if DiskGetFileInfo(UIUtil.UIFile('/icons/units/' .. control.Data.id .. '_icon.dds', true)) then
-                control.Icon:SetTexture(UIUtil.UIFile('/icons/units/' .. control.Data.id .. '_icon.dds', true))
+            local iconId = UnitBuildIconId(control.Data.id)
+	    if DiskGetFileInfo(UIUtil.UIFile('/icons/units/' .. iconId .. '_icon.dds', true)) then
+		control.Icon:SetTexture(UIUtil.UIFile('/icons/units/' .. iconId .. '_icon.dds', true))
             else
                 control.Icon:SetTexture(UIUtil.UIFile('/icons/units/default_icon.dds'))
             end
@@ -770,7 +778,7 @@ function CommonLogic()
 
     controls.choices.SetControlToType = function(control, type)
         local function SetIconTextures(control, optID)
-            local id = optID or control.Data.id
+            local id = UnitBuildIconId(optID or control.Data.id)
             if DiskGetFileInfo(UIUtil.UIFile('/icons/units/' .. id .. '_icon.dds', true)) then
                 control.Icon:SetTexture(UIUtil.UIFile('/icons/units/' .. id .. '_icon.dds', true))
             else
--- a/lua/ui/game/construction.lua
+++ b/lua/ui/game/construction.lua
@@ -419,7 +419,11 @@ function CreateTabs(type)
         defaultTabOrder = {t3 = 1, t2 = 2, t1 = 3, t4 = 4} -- T4 is last because only the Novax can build T4 but not T3
     elseif type == 'enhancement' then
         local selection = sortedOptions.selection
-        local enhancements = selection[1]:GetBlueprint().Enhancements
+        local loadout = import('/lua/ui/game/sacuLoadout.lua')
+	local enhancements = selection[1]:GetBlueprint().Enhancements
+	if loadout.IsCybranGatewaySelection(selection) then
+    		enhancements = loadout.GetSacuEnhancements()
+	end
         local enhCommon = import("/lua/enhancementcommon.lua")
         local enhancementPrefixes = {Back = 'b-', LCH = 'la-', RCH = 'ra-'}
         local newTabs = {}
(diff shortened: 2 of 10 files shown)


lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+--- Which of FactionSacuIds' factions (if any) a unit belongs to.
+--- @param unit Unit
+--- @return string|nil faction category, e.g. 'CYBRAN'
+local function GetUnitSacuFaction(unit)
+    for faction in FactionSacuIds do
+        if unit:IsInCategory(faction) then
+            return faction
+        end
+    end
+end
```
lL1l1: Use the module `factions.lua` to interact with factions for mod compatibility.

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+        table.insert(copy, name)
+    end
+    table.sort(copy)
+    return string.lower(table.concat(copy, '|'))
+end
+
+--- Generated combo units share their base SACU's icon (they have no icon of their
+--- own). Used by construction.lua, unitview.lua and unitviewDetail.lua so the
+--- build grid, queue grid and unit-view panel all fall back consistently, for
+--- every faction's combos.
```
lL1l1: Why not set the icon in the unit blueprint generator?

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+    return id
+end
+
+--- True if every unit in the selection is a Quantum Gateway belonging to the
+--- SAME known faction. A mixed-faction selection is rejected outright (rather
+--- than accepted and half-honored) because QueueSelected only ever derives
+--- one preset id, from selection[1]'s faction, and issues it to the whole
+--- selection - a gateway of a different faction would silently reject that
+--- blueprint id and queue nothing, which is the "strange behavior" this
+--- guards against.
```
lL1l1: This would act incorrectly if a mod made SACUs available from non-gateway units or any faction gateway ("All Faction Quantum Gate" mod), or if custom scripts changed construction capabilities mid-match.
You should use the construction rules for a set of units as acquired from `GetUnitCommandData`.

lua/system/blueprints-sacu-combos.lua
```diff
@@ -0,0 +1,247 @@
...
+-- Base SACU blueprint id, one per faction. Kept in sync by hand with
+-- SacuLoadout.FactionSacuIds in lua/ui/game/sacuLoadout.lua (that file can't
+-- import this one - see the comment at the top of it - so the two tables can't
+-- share a single definition).
+local ComboFactionBaseIds = {
+    uel0301 = true, -- UEF
+    ual0301 = true, -- Aeon
+    url0301 = true, -- Cybran
+    xsl0301 = true, -- Seraphim
+}
```
lL1l1: Use a new category added to the blueprint to trigger enhancement autogeneration and manage UI enabling for the enhancement picker.
Categories are synced between sim and UI and cannot be changed in runtime. They're basically an improvement of this table system.
This would be friendlier to modders not having to hook and synchronize all files interacting with your new system.

lua/system/blueprints-sacu-combos.lua
```diff
@@ -0,0 +1,247 @@
...
+    StabilitySuppressant = 'Reacton',
+    SystemIntegrityCompensator = 'Nano',
+    Teleporter = 'Teleport',
+    -- Seraphim
+    DamageStabilization = 'Nano',
+    EngineeringThroughput = 'Fab',
+    EnhancedSensors = 'Sensor',
+    Missile = 'Missile',
+    Overcharge = 'OC',
+}
```
lL1l1: Put these names in the unit blueprint. Also it needs localized versions.

lua/system/blueprints-sacu-combos.lua
```diff
@@ -0,0 +1,247 @@
...
+-- import this one - see the comment at the top of it - so the two tables can't
+-- share a single definition).
+local ComboFactionBaseIds = {
+    uel0301 = true, -- UEF
+    ual0301 = true, -- Aeon
+    url0301 = true, -- Cybran
+    xsl0301 = true, -- Seraphim
+}
+
+local SlotOrder = { 'LCH', 'RCH', 'Back' }
```
lL1l1, line 19: Ideally the system can be managed without hardcoded slots because the slots are in blueprints. The sim works this way but the UI doesn't (because its overkill to make such a dynamic UI when it's a relatively restricted system in the game).

lua/system/blueprints-sacu-combos.lua
```diff
@@ -0,0 +1,247 @@
...
+        TableInsert(keys, key)
+    end
+    table.sort(keys)
+    return keys
+end
+
+local function EnhancementChain(enhName, slotDefs)
+    local chain = { enhName }
+    local current = enhName
+    local guard = 0
```
lL1l1, line 37:
```suggestion
    local depth = 0
```

lua/system/blueprints-sacu-combos.lua
```diff
@@ -0,0 +1,247 @@
...
+    return 'combo_' .. table.concat(copy, '_')
+end
+
+local function CleanLabel(name, def)
+    local label = name
+    if def and def.Name then
+        label = string.gsub(def.Name, '^<LOC [^>]+>', '')
+    end
+    return label
+end
```
lL1l1, line 146: You shouldn't remove LOC tags.

lua/system/blueprints-sacu-combos.lua
```diff
@@ -0,0 +1,247 @@
...
+                            bp.EnhancementPresets[presetName] = {
+                                -- The hand-authored presets on this unit (RAS, Engineer, Combat, ...)
+                                -- set Description to the same short string as UnitName, not the base
+                                -- SCU's own Description ("Support Armored Command Unit") - that's what
+                                -- was showing up appended after the name in build tooltips/info panels.
+                                Description = pretty,
+                                BuildIconSortPriority = 90,
+                                Enhancements = enhList,
+                                HelpText = pretty,
+                                SelectionPriority = 1,
```
lL1l1: Need to distinguish between combat (1) and engineer (3) priority based on enhancement data.

lua/ui/game/construction.lua
```diff
@@ -420,6 +424,11 @@ function CreateTabs(type)
     elseif type == 'enhancement' then
         local selection = sortedOptions.selection
         local enhancements = selection[1]:GetBlueprint().Enhancements
+        local isGatewaySelection = SacuLoadout.IsGatewaySelection(selection)
```
lL1l1, line 427: For mod and scripted map compatibility:
- `GetUnitCommandData(selection)` and filter for SACU build capability.
- `table.empty(EntityCategoryFilterOut(categories.FACTORY, selection))` to check if selection has non-factory units.

lua/ui/game/construction.lua
```diff
@@ -420,6 +424,11 @@ function CreateTabs(type)
...
+        local gatewaySacuId = isGatewaySelection and SacuLoadout.GetSelectionSacuId(selection)
```
lL1l1, line 428: I think this should work: 
```lua
EntityCategoryGetUnitList(buildableCategories * categories.USEBUILDPRESETSAUTO)[1]
```
The category is an example, with the name based off the existing `USEBUILDPRESETS` category. Maybe a new category is not necessary and the system is compatible with any modded preset user?
Also note that it accesses index 1. You could have multiple SACUs you could make upgrades for the UI architecture you use doesnt accomodate that.

lua/ui/game/construction.lua
```diff
@@ -459,7 +468,11 @@ function CreateTabs(type)
                         ---@field UnitID UnitId
 
                         enhTable.ID = enhName
-                        enhTable.UnitID = selection[1]:GetBlueprint().BlueprintId
+                        if isGatewaySelection then
+                            enhTable.UnitID = gatewaySacuId
+                        else
+                            enhTable.UnitID = selection[1]:GetBlueprint().BlueprintId
+                        end
```
lL1l1, line 475: This value is available in the blueprint for existing presets:
https://github.com/FAForever/fa/blob/77ac963d82613137bb0a66e75011aa68453fddef/lua/system/Blueprints.lua#L793
so use
```lua
local bp = selection[1]:GetBlueprint()
enhTable.UnitID = bp.BaseBlueprintId or bp.BlueprintId
```

lua/ui/game/construction.lua
```diff
@@ -2520,6 +2578,14 @@ function OnSelection(buildableCategories, selection, isOldSelection)
         capturingKeys = false
         -- Sorting down units
         local buildableUnits = EntityCategoryGetUnitList(buildableCategories)
+        local visibleBuildableUnits = {}
+        for _, unitId in buildableUnits do
+            local unitBp = __blueprints[unitId]
+            if not (unitBp and unitBp.CategoriesHash and unitBp.CategoriesHash.SACULOADOUTCOMBO) then
```
lL1l1, line 2590:
```suggestion
            if not unitBp.CategoriesHash.SACULOADOUTCOMBO then
```
supcom lua allows indexing nil, and it returns nil. AIs don't pick up on this.

lua/ui/game/construction.lua
```diff
@@ -2611,7 +2677,8 @@ function OnSelection(buildableCategories, selection, isOldSelection)
             end
         end
 
-        if table.getn(selection) == 1 and selection[1]:GetBlueprint().Enhancements then
+        local isGatewaySelection = SacuLoadout.IsGatewaySelection(selection)
+        if (table.getn(selection) == 1 and selection[1]:GetBlueprint().Enhancements) or isGatewaySelection then
```
lL1l1, line 2687: Intersect buildable categories of the selection with the `USEBUILDPRESETSAUTO` category I suggested.

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+
+local function IsRemoveEnhancement(name)
+    return string.sub(name, -6) == 'Remove'
+end
+
+local function EnhancementChain(sacuId, name)
+    local bp = __blueprints[sacuId]
+    local chain = { name }
+    local current = name
+    local guard = 0
```
lL1l1, line 119:
```suggestion
    local depth = 0
```

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
+--******************************************************************************************************
+--** SACU loadout picker
+--** Used by each faction's Quantum Gateway enhancement tab. Clicks choose slots; Queue issues a factory build.
+--******************************************************************************************************
+
+-- Note: deliberately NOT importing blueprints-sacu-combos.lua here. That file is
+-- loaded via doscript() from Blueprints.lua so its functions become real globals;
+-- import()-ing it as well would re-run the whole file into a second, throwaway
+-- module table just to reach EnhancementSetKey. The one function we need is
+-- reproduced locally below instead.
```
lL1l1, line 10: Which function? Move the comment to that function.

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+function Open()
+    QueueSelected(1)
+end
+
+function Toggle()
+    QueueSelected(1)
+end
+
+function Close()
+end
```
lL1l1, line 243: Make sure to remove the debug code.

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+    local sacuId = GetSelectionSacuId(selection)
+    local id = sacuId and FindBlueprintId(sacuId, SelectedEnhancements(sacuId))
+    if not id then
+        print('No blueprint for that loadout')
+        return
+    end
+
+    IssueBlueprintCommand("UNITCOMMAND_BuildFactory", id, count)
+    print('Queued ' .. id)
+end
```
lL1l1, line 231: Let UI code handle the fail/success response. For example, I may want a prettier response than the one from `print`.
As a suggestion, you should add an error sound when it doesn't work:
```lua
PlaySound(Sound({ Bank = 'Interface', Cue = 'UI_Menu_Error_01' }))
```

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+
+--- True if every unit in the selection is a Quantum Gateway belonging to the
+--- SAME known faction. A mixed-faction selection is rejected outright (rather
+--- than accepted and half-honored) because QueueSelected only ever derives
+--- one preset id, from selection[1]'s faction, and issues it to the whole
+--- selection - a gateway of a different faction would silently reject that
+--- blueprint id and queue nothing, which is the "strange behavior" this
+--- guards against.
+function IsGatewaySelection(selection)
```
lL1l1: Annotations are missing

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+    table.sort(copy)
+    return string.lower(table.concat(copy, '|'))
+end
+
+--- Generated combo units share their base SACU's icon (they have no icon of their
+--- own). Used by construction.lua, unitview.lua and unitviewDetail.lua so the
+--- build grid, queue grid and unit-view panel all fall back consistently, for
+--- every faction's combos.
+--- @param id string|nil
+--- @return string|nil
```
lL1l1, line 41: Use `?` to annotate `|nil` like the rest of the repo.

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+function GetSelectionSacuId(selection)
+    if not selection or table.empty(selection) then
+        return nil
+    end
+    local faction = GetUnitSacuFaction(selection[1])
+    return faction and FactionSacuIds[faction]
+end
+
+function GetSacuEnhancements(sacuId)
```
lL1l1, line 104: Missing annotations. 
Variable name implies type `UnitId` (a string) but it is used as a unit blueprint.

lua/ui/game/sacuLoadout.lua
```diff
@@ -0,0 +1,252 @@
...
+                        if string.find(realId, 'combo_') then
+                            return realId
+                        end
+                    end
+                end
+            end
+        end
+    end
+    return found
+end
```
lL1l1, line 212: bp.BaseBlueprintId

lua/ui/game/unitview.lua
```diff
@@ -448,10 +450,17 @@ function UpdateWindow(info)
...
+        -- Generated SACU loadout combos already carry their full description in
+        -- the name itself (e.g. "SACU (Cloak/EMP)") - skip the usual "Name: TechN
+        -- Description" suffix for them so nothing trails after the name.
+        if bp.CategoriesHash and bp.CategoriesHash.SACULOADOUTCOMBO then
+            controls.name:SetText(name)
+        else
+            if name ~= '' then
+                name = name .. ': '
+            end
+            controls.name:SetText(name .. description)
```
lL1l1: Do not hardcode UI like this for a specific unit. This already works fine by setting `name` to an empty string, like in existing presets.

lua/ui/game/unitviewDetail.lua
```diff
@@ -213,6 +214,13 @@ function LOCStr(str)
 end
 
 function GetShortDesc(bp)
+    -- Generated SACU loadout combos already carry their full description in the
+    -- name itself (e.g. "SACU (Cloak/EMP)") - skip the usual "Name: TechN
+    -- Description" suffix for them so nothing trails after the name.
+    if bp.CategoriesHash and bp.CategoriesHash.SACULOADOUTCOMBO then
+        return bp.General.UnitName and LOC(bp.General.UnitName) or ''
+    end
+
```
lL1l1: Do not hardcode ui for specific units like this.

lua/ui/game/unitviewDetail.lua
```diff
@@ -881,7 +889,14 @@ function Show(bp, builderUnit, bpID)
...
 
-    local iconName = GameCommon.GetCachedUnitIconFileNames(bp)
+    local iconBp = bp
+    if bp.BlueprintId then
+        local iconId = SacuLoadout.UnitBuildIconId(bp.BlueprintId)
+        if iconId ~= bp.BlueprintId then
+            iconBp = __blueprints[iconId] or bp
+        end
+    end
+    local iconName = GameCommon.GetCachedUnitIconFileNames(iconBp)
```
lL1l1, line 900: Shouldn't this work fine if the autogenerated bp has `Display.IconName`?

lL1l1: Also why doesn't construction.lua use this function?

Seems like the implementation and usage of this function is incomplete, although it seems quite useful. It was never touched since gpg.

lua/ui/help/tooltips.lua
```diff
@@ -221,6 +221,10 @@ Tooltips = {
         title = "<LOC tooltipui0048>Production Toggle",
         description = "<LOC tooltipui0492>Turn the selected units production capabilities on/off",
     },
+    sacu_loadout_queue = {
+        title = "Queue SACU loadout",
+        description = "Queue a SACU with the upgrades selected on the enhancement tab. Shift-click queues 5.",
```
lL1l1: When finalizing the PR, make sure to add LOC keys for all new tooltips and other UI text elements.

lL1l1 (commented):
Balance:
- Requires an option to disable an enhancement from being queued in the gateway. This would apply to teleport enhancements.
  UI-wise, the enhancement should appear in the gate but be greyed out as unvailable, with a tooltip explaining that the enhancement cannot be built in the gateway.
  Alternatively, the cost of teleport enhancements has to be made higher and its contribution to sacrifice needs to be disabled. The cost increase would be in proportion to the equivalent mass of mass fab income or pgen income discounted due to adjacency with the tele preset.

UI:
I think it would be good to show enhancements in the bottom bar (the queue) like we have with ACUs. So it would show an SACU icon, the enhancements, a separator bar to indicate it is the preset, and then the next unit. 
With such a queue setup, you could create a more consistent queue where you issue a number of SACUs, then pick upgrades (this would restart construction as you are changing the currently built unit), then pick the next unit. It would not require new buttons, it would show what upgrades you have queued in the custom preset, it would allow for modification of queued upgrades instead of needing to re-define the desired preset, and players would have to learn a simpler divider bar instead of a new order button.
This would also elegantly fix the multi-faction/multi-upgradeable-unit issues since you would have a selector for the upgradeable unit in that initial queued unit, off of which you would choose the queued enhancements to determine the preset.

You should make sure the "factory queue preview" option works as intended with the SACU UI.

A lot of annotations are missing.

RabidPope:
@lL1l1 Hi. Thanks for all the feedback and tips.
Made the changes you suggested.
Please let me know if I need to make more fixes if you get the time to check.

About the suggested change in functionality.
Its a great idea, but would rather add that as a separate PR later.
I also dont mind if anyone copies this PR and makes their own changes, since this is mostly just a prototype to get the ball rolling.

lua/system/blueprints-sacu-combos.lua
RabidPope: attempts to resolve this text with LOC() instead of stripping the
tag (so combo labels would actually translate) crashed blueprint loading.
LOC is apparently not yet available as a global this early in the load
order?

lua/system/blueprints-sacu-combos.lua
RabidPope: Done. Some priorities don't make sense, like stealth. Others are guesswork on my part.

lua/system/blueprints-sacu-combos.lua
RabidPope: Done. Dont have a non english client to test this yet though.

lua/ui/game/sacuLoadout.lua
RabidPope: Done.
About the suggested change in functionality.
Its a great idea, but would rather add that as a separate PR later.

lua/system/blueprints-sacu-combos.lua
RabidPope: Not sure what would be the best way to do it.

lua/ui/game/construction.lua
RabidPope: @lL1l1 Do you mean multiple gateways building different upgrade selections?

lua/ui/game/sacuLoadout.lua
RabidPope: Updated.

lua/system/blueprints-sacu-combos.lua
RabidPope: Updated

lua/ui/game/sacuLoadout.lua
RabidPope: I changed this and pushed it but it is not reflectin in the pull request. Thit is a copy from my branch.

local function FindBlueprintId(sacuId, enhancements)
    if table.empty(enhancements) then
        return sacuId
    end

    local want = EnhancementSetKey(enhancements)
    local found = nil
    for id, bp in __blueprints do
        if bp and bp.BaseBlueprintId == sacuId and bp.EnhancementPresetAssigned then
            local assigned = bp.EnhancementPresetAssigned
            if assigned.Enhancements and EnhancementSetKey(assigned.Enhancements) == want then
                local realId = bp.BlueprintId or id
                found = realId
                if type(realId) == 'string' and string.find(realId, 'combo_') then
                    return realId
                end
            end
        end
    end
    return found
end

lua/system/blueprints-sacu-combos.lua
```diff
@@ -6,56 +6,18 @@ local TableInsert = table.insert
...
-    xsl0301 = true, -- Seraphim
-}
+-- Any unit blueprint carrying this category is treated as a loadout-eligible
+-- base SACU: InjectSacuLoadoutPresets below generates its enhancement combo
+-- presets, and MarkHiddenSacuLoadoutPresets tags those generated combos with
+-- SACULOADOUTCOMBO. The four vanilla SACUs (uel0301/ual0301/url0301/xsl0301)
+-- carry it via their own .bp files. A mod adding a new faction, or extending
+-- an existing unit, opts in the same way - by adding SACULOADOUTBASE to that
+-- unit's Categories (or AddCategories) - without hooking or editing this file
+-- or lua/ui/game/sacuLoadout.lua at all.
```
lL1l1: When this is done, the comments would have to be cleaned up to remove the AI's habit of writing memory into the documentation. For example here it talks about "without hooking or editing".
Also in general the comments are quite long and repeat info somewhat.
Here it could just be `-- Category that triggers generation of loadouts for the unit`. or "category enabling loadout capability for the unit"

lua/system/blueprints-sacu-combos.lua
lL1l1: Yeah it isn't available because there is no doscript Localization.lua like there is in globalinit.lua (maybe because it would load the huge loc database into ruleinit lua state).

I think you may be able to create multiple localization directives inside a string using the "directives" feature:
https://github.com/FAForever/fa/blob/749d78d662cd7267f1608d2843a3c22b4abdb089/lua/system/Localization.lua#L111-L123
So the final unit description should be:
```lua
LoadoutUnit.Description = "{i lockey1} {i lockey2} {i lockey3}"
```
where the loc keys are the keys for the installed enhancements' short names.

lua/ui/game/sacuLoadout.lua
```diff
@@ -56,60 +30,86 @@ end
...
+        -- every loadout preset (hand-authored or our generated combos) is a
+        -- clone of it and carries that field.
+        if bp and not bp.EnhancementPresetAssigned then
+            if baseSacuId and baseSacuId ~= id then
+                return nil
+            end
+            baseSacuId = id
         end
-        faction = unitFaction
     end
```
lL1l1, line 69: Get rid of subcommander category as it does not determine the loadout ability.
Use loadout base category to get the base unit immediately.
```lua
local buildableBaseUnits = EntityCategoryGetUnitList(buildableCategories * categories.SACULOADOUTBASE)
if table.getn(buildableBaseUnits) > 1 then return nil end -- mimicking the `baseSacuId ~= id` check
return buildableBaseUnits[1]
```

lua/ui/game/unitviewDetail.lua
```diff
@@ -136,11 +136,19 @@ function ShowEnhancement(bp, bpID, iconID, iconPrefix, userUnit)
...
+            -- Shows the gateway-disabled message instead of the description.
+            CreateLines(View.Description, {{color = UIUtil.fontColor,
+                lines = WrapText(LOC('<LOC sacu_gateway_disabled_tooltip>This cannot be built in from the gateway.'), View.Description.Value[1].Width(), function(text)
+                    return View.Description.Value[1]:GetStringAdvance(text)
+                end)}})
+        else
+            -- If enhancement of preset, then remove extension. (ual0301_Engineer -> ual0301)
+            if string.find(bpID, '_') then
+                bpID = string.sub(bpID, 1, string.find(bpID, "_[^_]*$")-1)
+            end
```
lL1l1: Use blueprint.BaseId

lua/system/blueprints-sacu-combos.lua
```diff
@@ -0,0 +1,275 @@
...
+function InjectSacuLoadoutPresets(all_bps)
+    if not all_bps or not all_bps.Unit then
+        return
+    end
+
+    local generated = 0
+    local skippedNamed = 0
+
+    for id, bp in all_bps.Unit do
+        if bp.CategoriesHash and bp.CategoriesHash[BaseCategory] and bp.Enhancements and bp.EnhancementPresets then
```
lL1l1:
```suggestion
        if bp.CategoriesHash[BaseCategory] and bp.Enhancements and bp.EnhancementPresets then
```

lua/ui/game/sacuLoadout.lua
lL1l1: I must've misunderstood the function at first, because my comment doesn't make sense. I think that's because the function name is bad (what blueprint id is it actually finding?).

With that out of the way, I think you shouldn't need to iterate on __blueprints if the bp id is generated from enhancements. You could have a shared function that generates a bp id from a list of enhancement names and a bp.

lua/ui/game/sacuLoadout.lua
```diff
@@ -202,51 +202,39 @@ local function FindBlueprintId(sacuId, enhancements)
...
 function QueueSelected(count)
     count = count or 1
     local selection = GetSelectedUnits() or {}
     if not IsGatewaySelection(selection) then
-        print('Select a Quantum Gateway first')
-        return
+        return false, 'Select a Quantum Gateway first'
     end
 
     local sacuId = GetSelectionSacuId(selection)
```
lL1l1: Now `GetSelectionSacuId` is called twice: once directly and second through IsGatewaySelection

lua/ui/game/sacuLoadout.lua
```diff
@@ -202,51 +202,39 @@ local function FindBlueprintId(sacuId, enhancements)
...
+--- @param count? integer
+--- @return boolean success
+--- @return string idOrReason the queued blueprint id on success, else a failure reason
 function QueueSelected(count)
     count = count or 1
     local selection = GetSelectedUnits() or {}
     if not IsGatewaySelection(selection) then
-        print('Select a Quantum Gateway first')
-        return
+        return false, 'Select a Quantum Gateway first'
```
lL1l1: "Selection cannot build any units with loadouts" would better generalize for other factory structures

RabidPope:
@lL1l1 
 ---@param op string 
 ---@param ident string 
 ---@return string 
 local function LocSubFn(op, ident) 
     if op == 'i' then 
         local loc = loc_table[ident] 
         if loc then 
             return LocExpand(loc) 
         else 
             WARN("missing localization key for include: " .. ident) 
             return "{unknown key: " .. ident .. '}' 
         end 
        
Will this work for the short names added in this PR (they're not in strings_db.lua yet), or does something need to change first?

PR #6888, opened 2025-07-17
Extend command line single player launch capabilities

Garanas:
## Description of the proposed changes

Fixes various things:

- Documents and annotate the majority of the launching a skirmish via the command line interface.
- Assign the default value of all lobby options by default.
- Add support for `/gameoptions` argument to define lobby options.
- Add support for (a new) `/gamebots` argument to define AI army configurations.
- Add support for (a new) `/gamemods` argument to define what (sim) mods to load.
- Add support for (a new) `/runWithTheWind` argument to run the simulation as fast as possible.
- Fix an issue with the scoreboard for missing the `FullShare` option.

And introduces a new powershell script as inspired by the script made by @lL1l1. It serves as an example on how to launch a bot session.

## Testing done on the proposed changes

It enables a setup such as this:

https://github.com/user-attachments/assets/7ca47d64-c390-4bd7-ac1b-176c048325ad

The AI is running, but the AI itself is not functional yet. The procedure is described at:

- https://github.com/Garanas/fa-average-joe-ai?tab=readme-ov-file#launch-the-game

Where you can launch the game from Visual Studio Code using command line arguments. 

## Additional context

I am not aware of this functionality being used anywhere except for the AI tournament functionality of Softles. That setup overrides this file however. The existing launch script uses the autolobby functionality. Therefore I'm not that afraid of backwards compatibility issues.

## Checklist

- [x] Changes are annotated, including comments where useful
- [x] Changes are documented in a changelog snippet according to the [guidelines](https://faforever.github.io/fa/development/changelog#format-of-a-snippet).
- [ ] Request 2-3 reviewers from the [list of reviewers and their areas of knowledge](https://github.com/FAForever/fa/blob/deploy/fafdevelop/CONTRIBUTING.md#reviewers).

Some of these changes are inspired by the work of @HardlySoftly , see also attached his scripts for the automated AI tournaments.

[FAF-AI-Autorun-main.zip](https://github.com/user-attachments/files/21297656/FAF-AI-Autorun-main.zip)

PR #7294, opened 2026-09-14
Feature: Lua configurable command line game launches

PreciseBump38:
Adds the ability to launch a command line session using a lua config file.  Native lua launch configs avoid having to do lots of command line argument parsing, meaning lots of config can happen without masses of parsing code and a very messy command line.

Example usage is included at the bottom.

The code was developed with AI assistance, but this PR is pure human rambling.

## Motivation
I'm doing bits of SCFA AI development and want the ability to run test games repeatably / easily.  For example, this should enable much easier running of AI tournaments.

I have extra bits of harness on top of this (e.g. python scripts to generate configs and parse logs), but felt like the core single player launch changes might be useful to contribute back to the main FA repo.  I'll probably publish the whole harness separately later once I'm happy with it (I want to do a whole web-ui for running batches of games, splitting out different logging streams for exploration/review, one-click reruns of particular games that had odd things in the logs or unexpected results, etc).

Should have no effect beyond the initial game launch - so should have no impact on how most people use the game.  I'm doing AI development off the back of these changes and have ironed out issues as they've come up.

## Additional Change Context
- I've added a logging line to SafeQuit() - that's so that tooling (such as I am hanging off this already) can tell the difference between a normal game end and manual window closure.

## Discussion Points
- How should this be documented for people to use?  I've included a lua snippet below to show how it works, but that's obviously not in the codebase.  Open to ideas on where/how to document effectively; for my own modding I have dedicated test/config directories, but I'm not 100% on where/if people want example configs in this repo.
- I'd kind of like to update the default game options in the file, at least to bump the unit cap to 1000.  Not included for now as it isn't relevant to the feature.
- Are there any features that other people want and could be added while we're here?

## Todo
- [x] Document changes in a snippet
- [ ] Approval / see if anybody else even wants this
- [ ] Cover off discussion points
- [ ] Get some reviews

## Example Usage
Here's an example of the command line that could be used:
`ForgedAlliance.exe` /init init.lua /RunWithTheWind /nomovie /nobugreport /map blah /autorun /mods/my_ai/test/quick_match.lua /log match.log`

And a corresponding quick_match.lua config:
```lua
-- Mounted anywhere the engine can reach it, e.g. /mods/my_ai/test/quick_match.lua
config = {
    mapName = "SCMP_007",

    -- Same default GameOptions as a normal /map launch; just override what you want
    scenario = {
        Victory = "demoralization",
        CheatsEnabled = "true",
    },

    armies = {
        -- No aikey => the human slot (at most one).
        { spawn = 1, faction = 1 },
        -- Everything else is an AI. `data` is accessible in-game via AutorunGetAIBrainData(aiBrain)
        { spawn = 2, aikey = "rush", faction = 2, data = { difficulty = "hard" } },
    },

    -- Tooling for stopping the sim at configured times (real-time or by number of ticks)
    sim = {
        targetSpeed = 10,
        maxGameSeconds = 900,
    },

    -- Easy injection of scripts in the sim context (once the sim starts)
    script = "/mods/my_ai/test/quick_match_hooks.lua",

    -- Session-wide data, accessible in-game via AutorunGetGlobalData()
    data = { logLevel = "verbose" },
}
```

Note that the `/map blah` arg is a required placeholder to trigger the SinglePlayerLaunch.lua methods at all from the engine.  The code checks this arg and warns if it resolves to something legitimate that is then overridden from the config (or uses the argument if it isn't specified in the config file).

## Summary by CodeRabbit

- **New Features**
  - Added support for launching fully configured single-player sessions with `/autorun <configPath>`.
  - Autorun configurations can define maps, scenarios, mods, army teams, player limits, and optional game-time or real-time limits.
  - Added session-start automation hooks for custom setup and first-tick actions.
  - Autorun scripts can access configured army-specific and session-wide data.

- **Bug Fixes**
  - Added logging when leaving an active non-replay session.


lua/SinglePlayerLaunch.lua
```diff
@@ -391,3 +409,220 @@ function StartCommandLineSession(mapName, isPerfTest)
...
+
+--- Waits for the session to actually start, then applies an optional game
+--- speed and ends the session once a configured game-time and/or real-time
+--- limit is reached. No-op if none of the three are set.
+---@param targetSpeed? number
+---@param maxGameSeconds? number
+---@param maxRealSeconds? number
+local function ApplySimOptions(targetSpeed, maxGameSeconds, maxRealSeconds)
+    local N = 100
+    local start = GetSystemTimeSeconds()
```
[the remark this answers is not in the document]
PreciseBump38: Not sure about this one so I'm going to leave it for now.  Given the simulation play rate isn't deterministic (and is highly configurable with these changes) then I sort of want a control over real times including the map loads.  If a very short test is desirable and map loading shouldn't be taken into account then I can always use the maxGameSeconds as a control.

lua/ui/dialogs/eschandler.lua
```diff
@@ -19,6 +19,7 @@ function SafeQuit()
         ForkThread(function ()
             ConExecute('ren_oblivion true')
             ConExecute('ren_ui false')
+            LOG("SafeQuit: player quit an active session")
```
lL1l1:
```suggestion
```

lua/SinglePlayerLaunch.lua
```diff
@@ -128,13 +129,17 @@ end
 
 
 --- Gets the scenario file from a map name if it isn't a scenario file already.
+--- Pass quiet=true to get nil instead of an error when it doesn't resolve.
```
lL1l1:
```suggestion
---
--- Pass quiet=true to get nil instead of an error when it doesn't resolve.
```

lua/SinglePlayerLaunch.lua
```diff
@@ -391,3 +409,227 @@ function StartCommandLineSession(mapName, isPerfTest)
...
+    local armies = scenario.Configurations.standard.teams[1].armies
+    local numColors = table.getn(import("/lua/gamecolors.lua").GameColors.PlayerColors)
+    local humanFound = false
+    for _, armyConfig in config.armies do
+        local armyIndex = armyConfig.spawn
+        if (not armyIndex) or (armyIndex <= 0) or (armyIndex > table.getn(armies)) then
+            error("Invalid army spawn "..tostring(armyIndex))
+        end
+        if sessionInfo.teamInfo[armyIndex] then
+            error("Duplicate army spawn "..tostring(armyIndex))
```
lL1l1: Could specify which index of the config's table is duplicated with what other index.

lua/SinglePlayerLaunch.lua
```diff
@@ -391,3 +409,227 @@ function StartCommandLineSession(mapName, isPerfTest)
...
+    ModUtils.SetSelectedMods(mods)
+
+    sessionInfo.teamInfo = {}
+    local armies = scenario.Configurations.standard.teams[1].armies
+    local numColors = table.getn(import("/lua/gamecolors.lua").GameColors.PlayerColors)
+    local humanFound = false
+    for _, armyConfig in config.armies do
+        local armyIndex = armyConfig.spawn
+        if (not armyIndex) or (armyIndex <= 0) or (armyIndex > table.getn(armies)) then
+            error("Invalid army spawn "..tostring(armyIndex))
```
lL1l1: Error string should contain the expected value

lua/SinglePlayerLaunch.lua
```diff
@@ -359,10 +367,20 @@ local function SetupCommandLineSkirmish(scenario, isPerfTest)
...
 --- Called by the engine using the `/map <mapPath>` launch arg
 ---@param mapName FileName
 ---@param isPerfTest any
 function StartCommandLineSession(mapName, isPerfTest)
+    -- Fully divert into a lua scripted session on `/autorun <config>`. All
+    -- other command line arguments are disregarded, `mapName` is just passed
+    -- so a warning can be given if it doesn't match a config provided value.
+    if HasCommandLineArg(configuredSessionCommandTrigger) then
+        return StartConfiguredSession(mapName)
+    end
```
lL1l1: Starting the config session from here means that it is dependent on /map being enabled. You can change `uimain.lua` `StartFrontEndUI` instead to make it not reliant on that. That function runs every time the main menu is reached though, so you could try it in `SetupUI` or just add a variable that lets the behavior only happen once.

lua/SinglePlayerLaunch.lua
```diff
@@ -391,3 +409,227 @@ function StartCommandLineSession(mapName, isPerfTest)
...
+    local configArg = GetCommandLineArg(configuredSessionCommandTrigger, 1)
+    local configLocation = configArg and configArg[1]
+    if not configLocation then
+        error("No config location specified, check your "..configuredSessionCommandTrigger.." argument")
+    end
+    if string.sub(configLocation, 1, 1) ~= "/" then
+        configLocation = "/config/"..configLocation
+    end
+    _ALERT("Loading game configuration from:", configLocation)
+    local config = import(configLocation).config
```
lL1l1: You should create a class annotation for the config script, so that all the functionality is visible in one place.

lua/SinglePlayerLaunch.lua
lL1l1: Also you could move the config session code to a new file so that it can be called from all the necessary contexts. This would help extend the feature to the local multiplayer code for example (although that requires messing with the autolobby and lobby so it may be tricky and outside the scope of this PR).

lua/simInit.lua
```diff
@@ -350,6 +350,35 @@ function BeginSession()
...
+        local scriptFile = import(ScenarioInfo.autorunScript)
+        if scriptFile.OnBeginSession then
+            scriptFile.OnBeginSession(ScenarioInfo)
+        end
+        if scriptFile.OnFirstGameTick then
+            ForkThread(function()
+                WaitTicks(1)
+                scriptFile.OnFirstGameTick(ScenarioInfo)
+            end)
+        end
```
lL1l1: What's the use case for the first tick function?

lL1l1 (commented):
You may find #6888 to have some feature you missed. That PR was also extending the session launch options but I like your structure of saveable launch config files over very long launch arguments.

I think it would be interesting to look at the lobby's preset system (`presets.lua` file) and see how much it overlaps with configs and if they could work together.

